PR #1720, opened 2024-02-22
Fix: temporal activities timeout

TBonnin:
Set more appropriate timeout for temporal activities Faster timeout for actions and webhooks as well as setting scheduleToStart timeouts

Syncs: 1h enqueued, 24h to execute
Actions: 1m enqueued, 15mins to execute
Webhook: 1h enqueued, 15mins to execute

https://linear.app/nango/issue/NAN-388/[outage]-temporal-set-appropriate-timeouts-for-temporal-activities

### DIFF
--- a/packages/jobs/lib/workflows.ts
+++ b/packages/jobs/lib/workflows.ts
@@ -2,17 +2,49 @@ import { CancellationScope, proxyActivities, isCancellation } from '@temporalio/
 import type * as activities from './activities.js';
 import type { WebhookArgs, ContinuousSyncArgs, InitialSyncArgs, ActionArgs } from './models/worker';
 
-const DEFAULT_TIMEOUT = '24 hours';
-const MAXIMUM_ATTEMPTS = 3;
+const SYNC_TIMEOUT = '24h';
+const SYNC_MAX_ATTEMPTS = 3;
+const ACTION_TIMEOUT = '15m';
+const ACTION_MAX_ATTEMPTS = 1; // no retry
+const WEBHOOK_TIMEOUT = '15m';
+const WEBHOOK_MAX_ATTEMPTS = 3;
 
-const { cancelActivity, reportFailure, routeSync, scheduleAndRouteSync, runAction, runWebhook } = proxyActivities<typeof activities>({
-    startToCloseTimeout: DEFAULT_TIMEOUT,
-    scheduleToCloseTimeout: DEFAULT_TIMEOUT,
+const { routeSync, scheduleAndRouteSync } = proxyActivities<typeof activities>({
+    // 1 hour to start so syncs are not evicted from the queue too soon
+    // 24 hours to complete to allow for long running syncs
+    scheduleToStartTimeout: '1h',
+    startToCloseTimeout: SYNC_TIMEOUT,
     retry: {
-        initialInterval: '5m',
-        maximumAttempts: MAXIMUM_ATTEMPTS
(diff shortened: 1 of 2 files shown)


packages/jobs/lib/workflows.ts
```diff
@@ -2,18 +2,40 @@ import { CancellationScope, proxyActivities, isCancellation } from '@temporalio/
...
+const { runWebhook } = proxyActivities<typeof activities>({
+    scheduleToStartTimeout: '30m',
+    startToCloseTimeout: WEBHOOK_TIMEOUT,
+    retry: {
+        maximumAttempts: WEBHOOK_MAX_ATTEMPTS
+    }
+});
+
+const { cancelActivity, reportFailure } = proxyActivities<typeof activities>({
+    startToCloseTimeout: '30s'
```
khaliqgant: Just confirming that this was tested against cancelling a sync and that this won't interfere?

packages/jobs/lib/workflows.ts
```diff
@@ -2,18 +2,40 @@ import { CancellationScope, proxyActivities, isCancellation } from '@temporalio/
...
+const { runAction } = proxyActivities<typeof activities>({
+    scheduleToStartTimeout: '1m',
+    startToCloseTimeout: ACTION_TIMEOUT,
+    retry: {
+        maximumAttempts: ACTION_MAX_ATTEMPTS
+    }
+});
+
+const { runWebhook } = proxyActivities<typeof activities>({
+    scheduleToStartTimeout: '30m',
```
khaliqgant: Why 30minutes for a webhook?

khaliqgant (approved):
Looks good, just a few clarifying q's

packages/jobs/lib/workflows.ts
TBonnin: no reason. I changed it to 1h, like syncs

packages/jobs/lib/workflows.ts
TBonnin: tested cancellation locally and it worked

packages/jobs/lib/workflows.ts
```diff
@@ -2,18 +2,50 @@ import { CancellationScope, proxyActivities, isCancellation } from '@temporalio/
...
 import type { WebhookArgs, ContinuousSyncArgs, InitialSyncArgs, ActionArgs } from './models/worker';
 
-const DEFAULT_TIMEOUT = '24 hours';
-const MAXIMUM_ATTEMPTS = 3;
+const SYNC_TIMEOUT = '24h';
+const SYNC_MAX_ATTEMPTS = 3;
+const ACTION_TIMEOUT = '15m';
+const ACTION_MAX_ATTEMPTS = 1; // no retry
+const WEBHOOK_TIMEOUT = '15m';
+const WEBHOOK_MAX_ATTEMPTS = 3;
```
bodinsamuel, line 10: imo all constants like should always be configurable via process.env

bodinsamuel (approved):

packages/jobs/lib/workflows.ts
TBonnin: good idea but I am worried about losing the capability of tracking such changes. If we had some IaC it would be better and even then that would require to check 2 codebases to figure out which is the current values

bodinsamuel: we need to centralize our usage of env var that's for sure, it was just a general statement over this pattern

PR #1725, opened 2024-02-23
persist: add total records count and size metrics

TBonnin:
## Describe your changes
We want to track total number of records persisted and total size so we can track change in usage pattern


khaliqgant (approved):

bodinsamuel (approved):
yes !

TBonnin:
Tests are not passing but it seems unrelated to my PR. Merging anyway

PR #1727, opened 2024-02-23
fix(db): add search path

bodinsamuel:
## Describe your changes

Start of a long effort on removing the schema everywhere and also fixing errors that could happen (reported on [slack](https://nango-community.slack.com/archives/C04ABR352H0/p1708685209570519?thread_ts=1708685072.874629&cid=C04ABR352H0))

- Add Search path force Postgres to look for table in this schema instead of public
- Test with one function


khaliqgant (approved):
Works well, but before merging can you remove the changes in the environmentService

PR #1728, opened 2024-02-23
[nan-428] fix legacy scripts and bad migration query

khaliqgant:
## Describe your changes
NAN-428

### DIFF
--- a/packages/shared/lib/services/sync/config/config.service.ts
+++ b/packages/shared/lib/services/sync/config/config.service.ts
@@ -20,6 +20,8 @@ const convertSyncConfigToStandardConfig = (syncConfigs: extendedSyncConfig[]): S
         models: {}
     };
 
+    let isV1 = false;
+
     for (const syncConfig of syncConfigs) {
         if (!syncConfig) {
             continue;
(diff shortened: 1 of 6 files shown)


packages/shared/lib/services/sync/config/config.service.ts
```diff
@@ -39,6 +41,15 @@ const convertSyncConfigToStandardConfig = (syncConfigs: extendedSyncConfig[]): S
 
         const syncName = syncConfig.sync_name;
 
+        const endpoint =
+            !syncConfig.endpoints_object || syncConfig?.endpoints_object?.length === 0
+                ? null
+                : syncConfig.endpoints_object.map((endpoint) => `${endpoint.method} ${endpoint.path}`);
+
+        if (!endpoint || endpoint.length === 0) {
+            isV1 = true;
```
bodinsamuel, line 50: no endpoint necessarily means v1? I had the impression it was available on v1 too

bodinsamuel (commented):
There is still errors on the tests but looks good overall

bodinsamuel (approved):

packages/shared/lib/services/sync/config/config.service.ts
khaliqgant: Good question. I just checked and if it is found to be v1 the endpoint functionality isn't available.

PR #1729, opened 2024-02-23
Revamp followups

khaliqgant:
## Describe your changes
Fixes downloading of private templates

### DIFF
--- a/packages/shared/lib/services/sync/config/config.service.ts
+++ b/packages/shared/lib/services/sync/config/config.service.ts
@@ -97,7 +97,9 @@ const convertSyncConfigToStandardConfig = (syncConfigs: extendedSyncConfig[]): S
 
     const { success, error, response: standardConfig } = convertV2ConfigObject(nangoConfig);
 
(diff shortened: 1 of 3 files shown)


packages/shared/lib/services/sync/config/config.service.ts
```diff
@@ -94,7 +95,9 @@ const convertSyncConfigToStandardConfig = (syncConfigs: extendedSyncConfig[]): S
         }
     }
 
-    const { success, response: standardConfig } = convertV2ConfigObject(nangoConfig);
+    const { success, error, response: standardConfig } = convertV2ConfigObject(nangoConfig);
+
+    logger.error(`Error in converting sync config to standard config: ${error}`);
```
bodinsamuel: there is no if to check if there is an error

khaliqgant: Thanks! fixed

bodinsamuel (approved):

PR #1730, opened 2024-02-23
[NAN-425] webapp revamp endless loading loop on activities

khaliqgant:
## Describe your changes
Refactor activity page

NAN-425

### DIFF
--- a/packages/webapp/src/pages/Activity.tsx
+++ b/packages/webapp/src/pages/Activity.tsx
@@ -169,9 +169,6 @@ export default function Activity() {
                 const id = parseInt(activityLogId);
                 setExpandedRow(id);
 
-                // remove query param env from the url without updating the push state
-                navigate(location.pathname + '?' + queryString.stringify({ ...queryParams, env: null }), { replace: true });
-
                 // wait 1 second before scrolling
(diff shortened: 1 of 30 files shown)


packages/webapp/src/pages/Activity.tsx
```diff
@@ -458,21 +335,32 @@ export default function Activity() {
         }
     }
 
+    if (error || logActivitiesError) {
+        requestErrorToast();
```
bodinsamuel: it shouldn't not be triggered by the rendering, otherwise it will pop each time you render the view.
Screenshot 2024-02-23 at 15 12 38

bodinsamuel (commented):
We need to start splitting components because it's hard to review right now

packages/webapp/src/pages/Activity.tsx
khaliqgant: Yup, agreed, this is still a WIP -- b/c I can't reproduce the infinite loop issue on my local

khaliqgant:
> We need to start splitting components because it's hard to review right now

Definitely, this page has gotten out of control

bodinsamuel (approved):

Luckydamian1 (commented):
###

PR #1731, opened 2024-02-23
fix(nango_sync_endpoints):relation doesn't exist

hassan254-prog:
## Describe your changes
Fix error `_nango_sync_endpoints` relation does not exist in the database


bodinsamuel (approved):
Thanks

PR #1734, opened 2024-02-23
Nan 426 other followups

khaliqgant:
## Describe your changes
- Fix hosted temporal error when looking up flows
- Fix missing status for `ERROR`
- FIx flattening of output when there is a multi model return

NAN-426

### DIFF
--- a/packages/cli/scripts/v1-v2.js
+++ b/packages/cli/scripts/v1-v2.js
@@ -57,7 +57,11 @@ function convertYAML(inputYAML) {
                     delete data.integrations[integration][taskName];
                 } else if (task.runs) {
                     change.type = 'sync';
-                    task.output = task.returns ? task.returns[0] : null;
+                    if (Array.isArray(task.returns)) {
+                        task.output = task.returns.length === 1 ? task.returns[0] : task.returns;
+                    } else {
+                        task.output = task.returns;
+                    }
                     change.changes.push({
                         action: `Set output from returns for ${taskName}`,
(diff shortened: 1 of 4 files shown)


packages/cli/scripts/v1-v2.js
```diff
@@ -57,7 +57,11 @@ function convertYAML(inputYAML) {
                     delete data.integrations[integration][taskName];
                 } else if (task.runs) {
                     change.type = 'sync';
-                    task.output = task.returns ? task.returns[0] : null;
+                    if (Array.isArray(task.returns)) {
+                        task.output = task.returns.length === 1 ? task.returns[0] : task.returns;
```
bodinsamuel, line 61: why would we care about returning `[0]` if we can return `task.returns` anyway?
I feel like the fix is `task.output = task.returns`

bodinsamuel (approved):

packages/cli/scripts/v1-v2.js
khaliqgant: Well, output shouldn't be a single array element but rather a single element. Otherwise if it is multi model then only should it be an array

khaliqgant: Talked about on this notion I believe: https://www.notion.so/nangohq/Nov-2023-nango-yaml-Revamp-4755d2717a0142dfb9a447e37eb91a5e?pvs=4

bodinsamuel: ah I didn't know there was a diff between single model vs multi model 👌🏻

PR #1735, opened 2024-02-23
feat(ui): merge with npm workspace

bodinsamuel:
## Describe your changes

Fixes NAN-436

- Make `webapp` be part of the npm workspace
It's handy for `npm install`, deduplicate packages, and finally fix eslint
- Fix the only thing breaking on my end by adding missing param `placeholder={null}` to the modals
- Simplify `npm run dev:watch` to have only one compilation
Let me know if you really rely on `watch:prettier` or one thing that was removed
- Fix missing `runner` reference in jobs tsconfig.json

--- 

In term of packages it's more or less the same as before:

## Before

729M  node_modules/
418M  packages/webapp/node_modules/

## After
1.0G	 node_modules/
68M  packages/webapp/node_modules/

✅ Tested on staging


khaliqgant (approved):
Pulled locally and works as expected 🙌

PR #1742, opened 2024-02-26
[NAN-437] Make retryOn configurable

khaliqgant:
## Describe your changes
Allow users to set a custom retryOn policy in addition to the defaults. Complain on compilation if not used correctly.

NAN-437


packages/cli/lib/services/parser.service.ts
```diff
@@ -101,6 +102,31 @@ class ParserService {
...
+                        if (!retriesPropertyFound && retryOnPropertyFound) {
+                            const lineNumber = path.node.loc?.start.line as number;
+                            console.log(
+                                chalk.red(
+                                    `Usage of 'retryOn' without 'retries' at "${filePath}:${lineNumber}". 'retryOn' should not be used if 'retries' is not set.`
+                                )
+                            );
+                            retryOnUsedCorrectly = false;
+                        }
+                    }
```
bodinsamuel, line 129: zod might be a better fit to validate this kind of properties

packages/server/lib/controllers/proxy.controller.ts
```diff
@@ -92,7 +93,8 @@ class ProxyController {
                 headers,
                 baseUrlOverride,
                 decompress: decompress === 'true' ? true : false,
-                method: method.toUpperCase() as HTTP_VERB
+                method: method.toUpperCase() as HTTP_VERB,
+                retryOn: retryOn && Array.isArray(retryOn) ? retryOn.map(Number) : null
```
bodinsamuel: it's already mapped as Number so I think it can be as simple as that, no?
```suggestion
                retryOn,
```

bodinsamuel (approved):
Looks good

packages/server/lib/controllers/proxy.controller.ts
khaliqgant: Yup, thanks. Updated with https://github.com/NangoHQ/nango/pull/1742/commits/8377b14dcf7555fdc788a75f11d9ceea4f6ce8c5 in addition to documentation

packages/cli/lib/services/parser.service.ts
khaliqgant: Yeah, most likely. An improvement to make down the line I'd say

docs-v2/reference/sdks/node.mdx
```diff
@@ -829,12 +829,15 @@ const res = await nango.get({
...
             </ResponseField>
-            <ResponseField name="data" type="unkown">
+            <ResponseField name="data" type="unknown">
                 The body of the request.
             </ResponseField>
             <ResponseField name="retries" type="number">
                 The number of retries in case of failure (with exponential back-off). Optional, default 0.
             </ResponseField>
+            <ResponseField name="retryOn" type="number[]">
+                Array of additional status codes to retry a request in addition to the 5xx and 429
```
bastienbeurier, line 839: @khaliqgant you told me there was more cases, right? Like `ECONNRESET` `ETIMEDOUT` `ECONNABORTED`

bastienbeurier: And thank you for updating the docs!

khaliqgant: Yeah, true @bastienbeurier. I'll add that into the documentation as well

PR #1743, opened 2024-02-26
fix(connection): empty creds when deleting

bodinsamuel:
## Describe your changes

Fixes NAN-431

- Remove creds and associated fields when we delete a connection
- Add missing empty creds type


khaliqgant (approved):

packages/shared/lib/services/connection.service.ts
```diff
@@ -546,7 +546,7 @@ class ConnectionService {
                 environment_id,
                 deleted: false
             })
-            .update({ deleted: true, deleted_at: new Date() });
+            .update({ deleted: true, credentials: {}, credentials_iv: null, credentials_tag: null, deleted_at: new Date() });
```
khaliqgant, line 549: Any reason an empty object instead of null for credentials?

bodinsamuel: `credentials` column is not nullable unfortunately

PR #1744, opened 2024-02-26
[nan-437] add in econn errors

khaliqgant:
## Describe your changes
Documentation update


packages/cli/package.json
```diff
@@ -26,7 +26,7 @@
     "dependencies": {
         "@babel/traverse": "^7.22.5",
         "@inquirer/prompts": "^2.3.0",
-        "@nangohq/shared": "^0.37.26",
+        "@nangohq/shared": "^0.38.2",
```
bodinsamuel, line 29: we really need to fix that

PR #1745, opened 2024-02-26
fix(sync): off-load data deletion

bodinsamuel:
## Describe your changes

Fixes NAN-432

When deleting a connection or an integration, only soft deletes connections/syncs, the rest will be deleted by the cron. Ideally we would have a queue to listen and not bother with the weirdness of this script, but we don't, so we try to find everything that should be deleted but that it's not deleted.
I'm not entirely confident about the performance, I put some low LIMIT to avoid any issues.

- Stop hard deleting data when deleting connections/provider/syncs
- Add a cron that will swipe the data after the fact by selecting sync_id in the related table that should not exists

## Test

- Create a sync and run it
- Check `data_records`  
  - 1️⃣ You should see your records
- Delete the connection or the provider
- Check `data_records`
  - 2️⃣ You should see your records
- Wait 20minutes (or change the timing of the cron :D)
- Check `data_records` 
  - 3️⃣ You should **NOT** see your records


packages/shared/lib/services/sync/data/records.service.ts
```diff
@@ -549,3 +545,35 @@ export async function getRecordsByExternalIds(external_ids: string[], nango_conn
...
+    return db.knex
+        .select<Pick<Sync, 'id'>[]>('syncs.id')
+        .from('_nango_sync_data_records AS rec')
+        .join('_nango_syncs AS syncs', 'syncs.id', '=', 'rec.sync_id')
+        .where('syncs.deleted', true)
+        .groupBy('syncs.id')
+        .limit(limit);
+}
+
+export async function deleteRecordsBySyncIdSafe(syncId: string, limit: number = 5000): Promise<void> {
```
khaliqgant: I'm not sure what `Safe` means in this context. Can a better name be used here?

khaliqgant (changes requested):
Method name request inline. It would be great to test this on staging before going to prod.

packages/shared/lib/services/sync/data/records.service.ts
bodinsamuel: it was just to emphasis that it was okay to use vs the regular delete, but I can remove the Safe entirely since there is no alternative

bodinsamuel:
I'll test in staging but there is nothing to delete 👌🏻 
Note: there is currently 355K of undeleted records in prod, so it should have some impact at the beginning

packages/shared/lib/services/sync/data/records.service.ts
bodinsamuel: Done ✅

bodinsamuel:
So there was no data but a few jobs to delete, all good 🙏🏻 
Screenshot 2024-02-28 at 17 02 57

khaliqgant (approved):

PR #1746, opened 2024-02-26
fix(db): remove schema()

bodinsamuel:
## Describe your changes

Follow up of #1727 

Remove all call to schema(), that is now not useful


khaliqgant (approved):
Works as expected, thanks for the followup

PR #1753, opened 2024-02-27
ci: test local image

bodinsamuel:


bodinsamuel:
Can't make it work, github actions is really painful to debug

PR #1755, opened 2024-02-28
[nan-439] fix model output

khaliqgant:
## Describe your changes
Display model inputs correctly

NAN-439


packages/shared/lib/services/nango-config.service.ts
```diff
@@ -101,7 +101,10 @@ function getFieldsForModel(modelName: string, config: NangoConfig): { name: stri
         return null;
     }
 
-    const modelData = config.models[modelName] || config.models[`${modelName.slice(0, -1)}`];
+    // if it is an array of models, we still need to be able to recognize it
+    const strippedModelName = modelName.replace(/[\W_]+/g, '');
```
bodinsamuel: I don't get this regex what's the input like?

bodinsamuel (approved):
Looks okay, I'm missing a bit the context of the fix here

packages/shared/lib/services/nango-config.service.ts
khaliqgant: It should replace `Foo[]` to be `Foo` so the model can be found correctly. I updated the regex in either case so it is clearer and does just that.

bodinsamuel: ah okay I forgot about this array syntax, makes sense 👌🏻

PR #1756, opened 2024-02-28
fix(lint): enable type checker

bodinsamuel:
## Describe your changes

Fixes NAN-454 

- Enable Typescript type checking linter
It makes the linter a little slower but catch way more issues. 

- Ran `--fix` on the packages 
It mostly removed unnecessary typings


.eslintrc
```diff
@@ -2,37 +2,53 @@
...
-      // Previously warn but now errors
-      "@typescript-eslint/no-non-null-assertion": "warn",
-      "@typescript-eslint/no-explicit-any": "warn"
+        "no-console": 1, // Means warning
+        "prettier/prettier": 2,
+        "@typescript-eslint/no-inferrable-types": "off",
+        // Previously warn but now errors
+        "@typescript-eslint/no-non-null-assertion": "warn",
+        "@typescript-eslint/no-explicit-any": "warn",
+        // To reenable as error progressively
```
bodinsamuel, line 36: To avoid adding errors everywhere I put those as warn, but those are relevant errors that we want to fix

bodinsamuel:
Can I have a final review on this please? @khaliqgant @TBonnin  ☺️

khaliqgant (approved):
Looks good, shouldn't fix be run regularly -- aka part of the build or precommit phase?

bodinsamuel:
> Looks good, shouldn't fix be run regularly -- aka part of the build or precommit phase?

We could at precommit like we do with prettier, I think it's also up to each developer to configure their IDE so that eslint runs along the edition. And at some point we will enforce eslint in the CI (hopefully)

.eslintrc
```diff
@@ -2,37 +2,53 @@
...
+        "@typescript-eslint/require-await": "warn",
+        "@typescript-eslint/no-floating-promises": "warn",
```
TBonnin, line 38: ❤️ ❤️ ❤️

PR #1760, opened 2024-02-28
[NAN-450] standardize api reference scripts tabs empty states

khaliqgant:
## Describe your changes
Copy update for empty states. Runs prettier for FE and apply prettier changes for the FE.

NAN-450

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,8 +15,8 @@
     "scripts": {
         "create:migration": "cd packages/shared/lib/db && knex migrate:make $1 --esm --knexfile ./knexfile.cjs",
         "undo:migration": "cd packages/shared/lib/db && knex migrate:rollback --esm --knexfile ./knexfile.cjs",
-        "prettier-format": "prettier --config .prettierrc \"./**/*.ts\" --write",
-        "prettier-watch": "onchange './**/*.ts' -- prettier --write {{changed}}",
+        "prettier-format": "prettier --config .prettierrc \"./**/*.{ts,tsx}\" --write",
+        "prettier-watch": "onchange './**/*.{ts,tsx}' -- prettier --write {{changed}}",
         "lint": "eslint . --ext .ts,.tsx",
         "lint:fix": "eslint . --ext .ts,.tsx --fix",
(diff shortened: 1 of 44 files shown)


package.json
```diff
@@ -15,8 +15,8 @@
     "scripts": {
         "create:migration": "cd packages/shared/lib/db && knex migrate:make $1 --esm --knexfile ./knexfile.cjs",
         "undo:migration": "cd packages/shared/lib/db && knex migrate:rollback --esm --knexfile ./knexfile.cjs",
-        "prettier-format": "prettier --config .prettierrc \"./**/*.ts\" --write",
-        "prettier-watch": "onchange './**/*.ts' -- prettier --write {{changed}}",
+        "prettier-format": "prettier --config .prettierrc \"./**/*.{ts,tsx}\" --write",
+        "prettier-watch": "onchange './**/*.{ts,tsx}' -- prettier --write {{changed}}",
```
bodinsamuel, line 19: Are you using those commands yourself? because prettier is already handled by eslint

packages/webapp/src/index.tsx
```diff
@@ -14,16 +14,16 @@ const options = {
...
+    integrations: [
+        new Sentry.BrowserTracing({
+            tracePropagationTargets: [/^https:\/\/app.nango\.dev\/api/]
+        }),
+        new Sentry.Replay()
+    ],
+    tracesSampleRate: 0.1,
+    replaysSessionSampleRate: 0.1,
+    replaysOnErrorSampleRate: 0.3
 });
```
bodinsamuel, line 27: Sentry is instantiated two times, here and in App.tsx, is it normal?

bodinsamuel (approved):
Finally, so satisfying 🙏🏻 🤩

package.json
khaliqgant: I ran `prettier-format` manually to get these updates because my local wasn't formatting previously on watch.

packages/webapp/src/index.tsx
khaliqgant: This was what I saw based on the docs. In the `App.tsx` routingInstrumentation is setup and in `index.tsx` tracing is setup via `tracePropagationTargets`

bodinsamuel: okay, why not 👍🏻

PR #1763, opened 2024-02-29
[nan-409] move flyout to the right

khaliqgant:
## Describe your changes
Moves flyout to the left so it doesn't get cut off with a small screen

image

NAN-409


bodinsamuel (approved):
it's also cut on the bottom for me, but I think it should just be handled by a component in the future 👌🏻

bodinsamuel (approved):

PR #1764, opened 2024-02-29
sdk proxy: memoize connection to avoid request on every proxy call

TBonnin:
## Describe your changes
Right now each call to `nango.proxy` in scripts makes a request to the api to fetch the connection.
This change caches the connection for 1 minute to avoid all those unnecessary api requests.

The `shouldRefreshCredentials` logic used in connection service is quite complex and requires accessing the database so we are not using it in the sdk but only cache the connection for one minute (happy to discuss this arbitrary value), which is good enough to avoid calling /connection/... on every proxy call without running into expired credentials

https://linear.app/nango/issue/NAN-411/reduce-the-number-of-calls-made-to-connection-when-proxying-from

### DIFF
--- a/packages/shared/lib/sdk/sync.ts
+++ b/packages/shared/lib/sdk/sync.ts
@@ -243,6 +243,8 @@ export class NangoAction {
 
     public ActionError = ActionError;
 
+    private memoizedConnection: { connection: Connection; timestamp: number } | undefined;
+
     constructor(config: NangoProps) {
         if (config.activityLogId) {
             this.activityLogId = config.activityLogId;
@@ -334,12 +336,12 @@ export class NangoAction {
         if (this.dryRun) {
             return this.nango.proxy(config);
         } else {
-            const proxyConfig = this.proxyConfig(config);
-            const connection = await this.nango.getConnection(proxyConfig.providerConfigKey, proxyConfig.connectionId);
+            const connection = await this.getConnection();
             if (!connection) {
                 throw new Error(`Connection not found using the provider config key ${this.providerConfigKey} and connection id ${this.connectionId}`);
             }
 
+            const proxyConfig = this.proxyConfig(config);
             const { response, activityLogs: activityLogs } = await proxyService.route(proxyConfig, {
                 existingActivityLogId: this.activityLogId as number,
                 connection,
@@ -416,7 +418,12 @@ export class NangoAction {
 
     public async getConnection(): Promise<Connection> {
         this.exitSyncIfAborted();
-        return this.nango.getConnection(this.providerConfigKey as string, this.connectionId as string);
+        if (!this.memoizedConnection || Date.now() - this.memoizedConnection.timestamp > 60000) {
+            const connection = await this.nango.getConnection(this.providerConfigKey as string, this.connectionId as string);
+            this.memoizedConnection = { connection, timestamp: Date.now() };
+            return connection;
+        }
+        return this.memoizedConnection.connection;
     }
 
     public async setMetadata(metadata: Record<string, any>): Promise<AxiosResponse<void>> {
--- a/packages/shared/lib/sdk/sync.ts
+++ b/packages/shared/lib/sdk/sync.ts
@@ -226,6 +226,8 @@ interface EnvironmentVariable {
     value: string;
 }
 
+const MEMOIZED_CONNECTION_TTL = 60000;
+
 export class NangoAction {
     private nango: Nango;
     private attributes = {};
@@ -418,7 +420,7 @@ export class NangoAction {
 
     public async getConnection(): Promise<Connection> {
         this.exitSyncIfAborted();
-        if (!this.memoizedConnection || Date.now() - this.memoizedConnection.timestamp > 60000) {
+        if (!this.memoizedConnection || Date.now() - this.memoizedConnection.timestamp > MEMOIZED_CONNECTION_TTL) {
             const connection = await this.nango.getConnection(this.providerConfigKey as string, this.connectionId as string);
             this.memoizedConnection = { connection, timestamp: Date.now() };
             return connection;
--- a/packages/shared/lib/sdk/sync.unit.test.ts
+++ b/packages/shared/lib/sdk/sync.unit.test.ts
@@ -6,10 +6,43 @@ import configService from '../services/config.service.js';
 import type { CursorPagination, LinkPagination, OffsetPagination } from '../models/Proxy.js';
 import { NangoAction } from './sync.js';
 import { isValidHttpUrl } from '../utils/utils.js';
+import proxyService from '../services/proxy.service.js';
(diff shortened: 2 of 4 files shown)


khaliqgant:
> without running into expired credentials

But isn't it still possible to run into expired credentials in this time frame? What happens if the credentials do expire in that 60 second window?

TBonnin:
> > without running into expired credentials
> 
> But isn't it still possible to run into expired credentials in this time frame? What happens if the credentials do expire in that 60 second window?

That would mean that last `getConnection` returned a connection with credentials expiring in less than 60s. If I read the `shouldRefreshCredentials` function properly in connection service, this cannot happen. Am I wrong?

packages/shared/lib/sdk/sync.unit.test.ts
```diff
@@ -6,12 +6,45 @@ import configService from '../services/config.service.js';
...
+    let nangoAction: NangoAction;
+    let nango: Nango;
+    beforeEach(async () => {
+        nangoAction = new NangoAction({
+            secretKey: '***',
+            providerConfigKey: 'github',
+            connectionId: 'connection-1'
+        });
+        nango = new Nango({ secretKey: '***' });
+        (await import('@nangohq/node')).Nango.prototype.getConnection = vi.fn().mockReturnValue({ credentials: {} });
```
bodinsamuel, line 27: You can mock `nango` instance directly? this stuff is always so messy

packages/shared/lib/sdk/sync.ts
```diff
@@ -416,7 +418,12 @@ export class NangoAction {
 
     public async getConnection(): Promise<Connection> {
         this.exitSyncIfAborted();
-        return this.nango.getConnection(this.providerConfigKey as string, this.connectionId as string);
+        if (!this.memoizedConnection || Date.now() - this.memoizedConnection.timestamp > 60000) {
```
bodinsamuel: I would make it a constant

packages/shared/lib/sdk/sync.unit.test.ts
```diff
@@ -6,12 +6,45 @@ import configService from '../services/config.service.js';
...
+
+    it('memoizes connection', async () => {
+        await nangoAction.proxy({ endpoint: '/issues' });
+        await nangoAction.proxy({ endpoint: '/issues' });
+        expect(nango.getConnection).toHaveBeenCalledTimes(1);
+    });
+    it('get connection if memoized connection is too old', async () => {
+        await nangoAction.proxy({ endpoint: '/issues' });
+        const later = Date.now() + 61000;
+        vi.spyOn(Date, 'now').mockReturnValue(later);
```
bodinsamuel, line 42: Does the mocking reset after this? if not, it could break other following tests.

bodinsamuel (commented):
Looks good 👍🏻 

I would be in favor of idempotency as much as we could, right now if somebody modifies the connection during a sync it could suddenly modify the script by accident. So if we can cache forever that would be better imo.

Regarding the refresh, you could abstract the `tokenExpirationCondition` in `shouldRefreshCredentials` to get the time of expiration.
And change to: 
```
this.memoizedConnection = { connection, expiresAt: this.getExpiration(connection) }
```

packages/shared/lib/sdk/sync.unit.test.ts
TBonnin: I was being consistent with the other tests in this file but I agree it is ugly

TBonnin:
> Looks good 👍🏻
> 
> I would be in favor of idempotency as much as we could, right now if somebody modifies the connection during a sync it could suddenly modify the script by accident. So if we can cache forever that would be better imo.

Credentials can expires so we must be able re-fetch. If the concept of credentials and connection were separated we would keep the connection and only refresh the credentials but that's not the case

> 
> Regarding the refresh, you could abstract the `tokenExpirationCondition` in `shouldRefreshCredentials` to get the time of expiration. And change to:
> 
> ```
> this.memoizedConnection = { connection, expiresAt: this.getExpiration(connection) }
> ```

That was my initial idea but `tokenExpirationCondition` depends on `providerConfig` which is read from the database.

packages/shared/lib/sdk/sync.unit.test.ts
TBonnin: yep 
```
afterEach(() => {
        vi.clearAllMocks();
    });
```

packages/shared/lib/sdk/sync.unit.test.ts
TBonnin: also nango is a private field of NangoAction so I am not sure I can mock it directly

khaliqgant:
> > > without running into expired credentials
> > 
> > 
> > But isn't it still possible to run into expired credentials in this time frame? What happens if the credentials do expire in that 60 second window?
> 
> That would mean that last `getConnection` returned a connection with credentials expiring in less than 60s. If I read the `shouldRefreshCredentials` function properly in connection service, this cannot happen. Am I wrong?



> > > without running into expired credentials
> > 
> > 
> > But isn't it still possible to run into expired credentials in this time frame? What happens if the credentials do expire in that 60 second window?
> 
> That would mean that last `getConnection` returned a connection with credentials expiring in less than 60s. If I read the `shouldRefreshCredentials` function properly in connection service, this cannot happen. Am I wrong?

Ah ok gotcha, nope – should be good then.

khaliqgant (approved):

bodinsamuel (approved):
> That was my initial idea but tokenExpirationCondition depends on providerConfig which is read from the database.

Not sure I understood everything, looks good anyway 👌🏻

TBonnin:
> > That was my initial idea but tokenExpirationCondition depends on providerConfig which is read from the database.
> 
> Not sure I understood everything, looks good anyway 👌🏻

Part of the `tokenExpirationCondition` logic depends on the `providerConfig` which is fetched from here https://github.com/NangoHQ/nango/blob/9d4da515905116cf4129f98c16c2b72100afa3f8/packages/shared/lib/services/config.service.ts#L112

bastienbeurier:
@TBonnin qq: does the caching work when calling [GET /connection/id](https://docs.nango.dev/reference/api/connection/get) externally? Or it's only for the proxy and scripts?

Impressive drop in server calls by the way, good job!!

TBonnin:
@bastienbeurier caching is only in the nango sdk, when calling `nango.proxy()` from scripts

bastienbeurier:
> @bastienbeurier caching is only in the nango sdk, when calling `nango.proxy()` from scripts

Got it, thanks! It'd be nice to add this for fetching credentials exteranlly as well, since this is an unlimited feature. But no rush!

TBonnin:
We can only cache because it happens client side (nango sdk). Caching server side would mean caching unencrypted credentials and we probably don't want to do that too much.

PR #1766, opened 2024-02-29
fix(datadog): correctly track runner

bodinsamuel:
## Describe your changes

Fixes NAN-493

It's been two weeks of investigation to find why the runner was not being properly tracked in Datadog.
Turns out, it was just missing env variables 😶 

- Add missing env var when we create runner
How do we update every runners? 🥲
- Remove import of tracer in favor of `dd-trace` to avoid accidental re-init 

--- 

I have manually updated staging, so you can check in Datadog rn

Screenshot 2024-03-01 at 17 35 34


khaliqgant (approved):

packages/jobs/lib/crons/autoIdleDemo.ts
```diff
@@ -14,7 +14,7 @@ import {
     logger
 } from '@nangohq/shared';
 import { SpanTypes } from '@nangohq/shared';
-import tracer from '../tracer.js';
```
TBonnin, line 17: I got the runner specific changes but why is this import change everywhere necessary?

bodinsamuel: When we import tracer.js we init the tracer, because it's a singleton it's not necessary to import this tracer specifically. However because it's a singleton and we import files from others folder we had a double init before, to avoid this mistake again I think it's safer to not export the tracer.

PR #1767, opened 2024-02-29
Tbonnin/package version

TBonnin:
## Describe your changes

We ran several times into a situation where `shared` version was not
upgraded properly when publishing a new version.
I (Thomas) suspect the -w option might be buggy and causes some of the
packages not to upgrade the shared version properly.
This commit gets rid of the -w option in favor of navigating into the
package folder and running npm install there directly.
I am not sure this would fix our problem but it is worth trying imho

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -27425,7 +27425,7 @@
             "version": "1.0.0",
             "license": "SEE LICENSE IN LICENSE FILE IN GIT REPOSITORY",
             "dependencies": {
-                "@nangohq/shared": "^0.38.0",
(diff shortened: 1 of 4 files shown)


khaliqgant (approved):
🤞🏿

bodinsamuel (commented):
Locally it works, so I would suspect it comes from the CI itself 
Can you add `set -x` so we have a log of the command and maybe a `git diff` at the end of script?

Otherwise, there are libraries for release that would have solved this. Like semantic-release

PR #1768, opened 2024-02-29
[NAN-478] update runner logic to allow to connect to a remote instance

khaliqgant:
## Describe your changes
* Add a remote instance runner so jobs doesn't need to spawn a process. In some cases I saw that a sync failed because spawning a process took too long. For enterprise customers they can setup a runner instance and so in that case jobs will connect to that remote instance instead. Customers can set this using the `RUNNER_SERVICE_URL` env variable
* Best practice is to use an AWS role for connecting to a bucket so this update allows customers to configure bucket connection using the role as long as they have `AWS_REGION` and `AWS_BUCKET_NAME` set

NAN-478

### DIFF
--- a/packages/jobs/lib/runner/remote.runner.ts
+++ b/packages/jobs/lib/runner/remote.runner.ts
@@ -1,0 +1,30 @@
+import type { Runner } from './runner.js';
+import { RunnerType } from './runner.js';
+import { getRunnerClient } from '@nangohq/nango-runner';
+
+export class RemoteRunner implements Runner {
+    public client: any;
+    public runnerType: RunnerType = RunnerType.Remote;
+    constructor(
+        public readonly id: string,
+        public readonly url: string
+    ) {
+        this.client = getRunnerClient(this.url);
+    }
+
+    async suspend(): Promise<void> {
+        console.warn('cannot suspend a remote runner');
+    }
+
+    toJSON() {
+        return { runnerType: this.runnerType, id: this.id, url: this.url };
+    }
+
+    static fromJSON(obj: any): RemoteRunner {
+        throw new Error(`'fromJSON(${obj})' not implemented`);
+    }
+
+    static async getOrStart(runnerId: string): Promise<RemoteRunner> {
+        return new RemoteRunner(runnerId, process.env['RUNNER_SERVICE_URL'] || 'http://nango-runner');
+    }
+}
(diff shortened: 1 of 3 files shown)


packages/jobs/lib/runner/remote.runner.ts
```diff
@@ -0,0 +1,30 @@
...
+    async suspend(): Promise<void> {
+        console.warn('cannot suspend a remote runner');
+    }
+
+    toJSON() {
+        return { runnerType: this.runnerType, id: this.id, url: this.url };
+    }
+
+    static fromJSON(obj: any): RemoteRunner {
+        throw new Error(`'fromJSON(${obj})' not implemented`);
```
bodinsamuel, line 24: Will probably output `[object object]`

packages/jobs/lib/runner/remote.runner.ts
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    static async getOrStart(runnerId: string): Promise<RemoteRunner> {
+        return new RemoteRunner(runnerId, process.env['RUNNER_SERVICE_URL'] || 'http://nango-runner');
```
bodinsamuel, line 28: it should throw if no URL imo

bodinsamuel (approved):
Can't test this it right now but looks okay 👌🏻

packages/jobs/lib/runner/runner.ts
```diff
@@ -48,7 +50,15 @@ export async function getOrStartRunner(runnerId: string): Promise<Runner> {
         } catch (err) {}
     }
     const isRender = process.env['IS_RENDER'] === 'true';
-    const runner = isRender ? await RenderRunner.getOrStart(runnerId) : await LocalRunner.getOrStart(runnerId);
+    let runner: Runner;
+    if (isEnterprise()) {
```
TBonnin, line 54: I would check for the existence of the env var since this is the prerequesite condition to make remote runner work

khaliqgant: Which env variable?
- `NANGO_ENTERPRISE`? --> if that isn't set this will just go to false
- `RUNNER_SERVICE_URL` --> defaults to the kubernetes address https://github.com/NangoHQ/nango-helm-charts/blob/main/charts/nango/values.yaml#L41

packages/jobs/lib/runner/remote.runner.ts
khaliqgant: commented on this here: https://github.com/NangoHQ/nango/pull/1768/files#r1508667681

packages/jobs/lib/runner/runner.ts
TBonnin: sorry I am talking about `RUNNER_SERVICE_URL`

PR #1770, opened 2024-03-01
fix(integrations): update base_url for gong-oauth

hassan254-prog:
## Describe your changes
Gong recently made changes to their API. Gong-oauth now uses `api_base_url_for_customer`, which varies for each customer, as its `base_url` for proxy requests. This parameter is returned in the response of `generate-customer-token` when exchanging code for an access token.


bastienbeurier (approved):

khaliqgant:
@hassan254-prog should this still be in draft?

hassan254-prog:
Robin had suggested I place it in draft just in case there is more work to it.

tconlin:
@hassan254-prog Will this automatically work for all customers? Will there ever be a situation where a sync will fail because a customer does not yet have `api_base_url_for_customer` defined?

hassan254-prog:
@tconlin , according to this [thread](https://help.gong.io/hc/en-us/articles/13944551222157-Create-an-app-for-Gong), whenever the code is exchanged for an access token during authorization, as part of data residency support, the `api_base_url_for_customer` will be generated and returned in response. Have you gotten the chance to test this?

tconlin:
@rguldener Can we bring this out of drafts?

rguldener:
> @rguldener Can we bring this out of drafts?

Sure! I only suggested draft initially because there might have been more work needed. Feel free to move this forward.

packages/shared/providers.yaml
```diff
@@ -753,7 +755,7 @@ gong-oauth:
     disable_pkce: true
     token_request_auth_method: basic
     proxy:
-        base_url: https://api.gong.io
+        base_url: ${connectionConfig.api_base_url_for_customer}
```
bastienbeurier: I'm concerned this will break existing connections that don't have the `api_base_url_for_customer` field populated in the connection config. 

Is there a way to fall back to `https://api.gong.io`? 

Would something like `${connectionConfig.api_base_url_for_customer || 'https://api.gong.io}` work?

bastienbeurier (changes requested):

packages/shared/providers.yaml
hassan254-prog: @bastienbeurier , I will do some test on this and see if the platform will handle such case. Incase of any changes to the platform, I will add them to this pull request. Thank you.

bastienbeurier: For reference, I checked that no one was using Gong in prod at the moment.

TBonnin (approved):
is it still the case that nobody uses gong, because that's a breaking change. otherwise look good to me

hassan254-prog:
@TBonnin , I don't have much context on this, but the last time we checked with @bastienbeurier , it was okay to merge. Could you please check if the number has changed before merging? Thank you!

TBonnin:
> @TBonnin , I don't have much context on this, but the last time we checked with @bastienbeurier , it was okay to merge. Could you please check if the number has changed before merging? Thank you!

I see 12 accounts using `gong-auth` in prod environments (5 have connections). We are gonna need a strategy so we don't break their integrations cc @bastienbeurier for visibility

packages/shared/providers.yaml
bastienbeurier: It looks like people are using this in prod now. Is there a way to have a fallback system? 

Something like `${connectionConfig.api_base_url_for_customer || 'https://api.gong.io'}`

bastienbeurier:
> > @TBonnin , I don't have much context on this, but the last time we checked with @bastienbeurier , it was okay to merge. Could you please check if the number has changed before merging? Thank you!
> 
> I see 12 accounts using `gong-auth` in prod environments (5 have connections). We are gonna need a strategy so we don't break their integrations cc @bastienbeurier for visibility

I also see production usage for it. 

@hassan254-prog @TBonnin @khaliqgant Is there a way to have a fallback system to avoid a breaking change?

Something like ${connectionConfig.api_base_url_for_customer || 'https://api.gong.io'}

TBonnin:
> Something like ${connectionConfig.api_base_url_for_customer || '[https://api.gong.io'}](https://api.gong.io'%7D)

I don't think we support this syntax in providers.yaml @khaliqgant might be able to confirm.

Could we populate the connection config for existing gong-auth connections with the default url? Might require synchronization between doing that and releasing this change though.

khaliqgant:
> > Something like ${connectionConfig.api_base_url_for_customer || '[https://api.gong.io'}](https://api.gong.io'%7D)
> 
> I don't think we support this syntax in providers.yaml @khaliqgant might be able to confirm.

No we don't but the idea was brought up in the past but decided against it to not have too much code like syntax in yaml which might prove difficult to maintain.

packages/shared/providers.yaml
```diff
@@ -1123,7 +1125,7 @@ gong-oauth:
     disable_pkce: true
     token_request_auth_method: basic
     proxy:
-        base_url: https://api.gong.io
+        base_url: ${connectionConfig.api_base_url_for_customer} || https://api.gong.io
```
bodinsamuel, line 1129: connectionConfig seems to be camelCase, I assume we can't rewrite properties in `token_response_metadata`?

bodinsamuel (approved):

packages/shared/providers.yaml
hassan254-prog: Yes currently we can't rewrite these properties.

packages/shared/providers.yaml
```diff
@@ -1124,7 +1126,7 @@ gong-oauth:
...
```
khaliqgant, line 1129: @bodinsamuel unless I missed something this won’t work without https://github.com/NangoHQ/nango/pull/1873?

bodinsamuel: ah wasn't aware of this, thanks 👌🏻

PR #1771, opened 2024-03-01
[NAN-461] records pagination default ordering should be based on

khaliqgant:
## Describe your changes
Records pagination should be based on `updated_at` instead of `created_at`

NAN-461


bodinsamuel (approved):
I still think it's a breaking change that should be announced to customers. 
Other than that you should add an index similar to `idx_nango_records_composite` (do not drop the old one until later)

khaliqgant:
Whoops. Did not mean to merge this PR. Will revert this!

PR #1775, opened 2024-03-01
[NAN-461] records pagination default ordering should be based on

khaliqgant:
## Describe your changes
Records pagination should be based on `updated_at` instead of `created_at`

NAN-461

Note: see https://github.com/NangoHQ/nango/pull/1792 to test all the pagination updates in aggregate

### DIFF
--- a/packages/shared/lib/db/migrations/20240301111115_records_updated_at_composite.cjs
+++ b/packages/shared/lib/db/migrations/20240301111115_records_updated_at_composite.cjs
@@ -1,4 +1,3 @@
-const tableName = '_nango_sync_data_records';
 
 exports.config = { transaction: false };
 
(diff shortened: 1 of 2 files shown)


khaliqgant:
Blocked by documentation and messaging cc/ @bastienbeurier

packages/shared/lib/db/migrations/20240301111115_records_updated_at_composite.cjs
```diff
@@ -0,0 +1,11 @@
+const tableName = '_nango_sync_data_records';
```
bodinsamuel:
```suggestion
```

bodinsamuel (approved):

PR #1776, opened 2024-03-01
Fix/cron delete syncs perf

bodinsamuel:
## Describe your changes

This cron has proven much more complicated than anticipated. It's not easy to say to PG finds me what should be deleted but not is not. Basically it has to do a seq-scan on all the tables, which is not good. 

- Revert to a naive approach, select every sync that has recently been deleted and iterate over them. 
That means more useless work but at least it should be much faster

- Use advisory lock to prevent other cron to trigger at the same time

This change the fact if the cron didn't work for a moment it will not catch up, but since it's impossible I prefer to focus on the best case scenario until we have a queue to manage jobs.

### DIFF
--- a/packages/jobs/lib/crons/deleteSyncsData.ts
+++ b/packages/jobs/lib/crons/deleteSyncsData.ts
@@ -1,8 +1,8 @@
 import * as cron from 'node-cron';
-import { errorManager, ErrorSourceEnum, logger, MetricTypes, softDeleteJobs, softDeleteSchedules, telemetry, syncDataService, db } from '@nangohq/shared';
+import { errorManager, ErrorSourceEnum, logger, MetricTypes, softDeleteSchedules, telemetry, syncDataService, db } from '@nangohq/shared';
 import tracer from '../tracer.js';
 
-const limitJobs = 5;
+// const limitJobs = 100;
 const limitSchedules = 100;
 const limitSyncs = 10;
 const limitRecords = 1000;
@@ -10,7 +10,7 @@ const limitRecords = 1000;
 export async function deleteSyncsData(): Promise<void> {
     /**
      * Clean data from soft deleted syncs.
-     * This cron is to be remove at some point, we need a queue to delete specific provider/connection/sync
+     * This cron needs to be removed at some point, we need a queue to delete specific provider/connection/sync
      */
     cron.schedule('*/20 * * * *', async () => {
         const start = Date.now();
@@ -39,12 +39,14 @@ export async function exec(): Promise<void> {
         }
 
         // -----
+        // It simply is not possible right now, table is too big
+
         // Soft delete jobs
-        let countJobs = 0;
-        do {
-            countJobs = await softDeleteJobs(limitJobs);
-            logger.info(`[deleteSyncs] soft deleted ${countJobs} jobs`);
-        } while (countJobs >= limitJobs);
+        // let countJobs = 0;
+        // do {
+        //     countJobs = await softDeleteJobs(limitJobs);
+        //     logger.info(`[deleteSyncs] soft deleted ${countJobs} jobs`);
+        // } while (countJobs >= limitJobs);
 
         // -----
         // Soft delete schedules
--- a/packages/jobs/lib/crons/deleteSyncsData.ts
+++ b/packages/jobs/lib/crons/deleteSyncsData.ts
@@ -1,8 +1,8 @@
 import * as cron from 'node-cron';
-import { errorManager, ErrorSourceEnum, logger, MetricTypes, softDeleteJobs, softDeleteSchedules, telemetry, syncDataService } from '@nangohq/shared';
+import { errorManager, ErrorSourceEnum, logger, MetricTypes, softDeleteJobs, softDeleteSchedules, telemetry, syncDataService, db } from '@nangohq/shared';
 import tracer from '../tracer.js';
 
-const limitJobs = 100;
+const limitJobs = 5;
 const limitSchedules = 100;
 const limitSyncs = 10;
 const limitRecords = 1000;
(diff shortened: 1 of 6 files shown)


packages/jobs/lib/crons/deleteSyncsData.ts
```diff
@@ -26,30 +37,40 @@ export async function deleteSyncsData(): Promise<void> {
...
-    logger.info(`[deleteSyncs] found ${syncs.length} syncs for records`);
-    for (const sync of syncs) {
-        logger.info(`[oldActivity] deleting syncId: ${sync.id}`);
-        await syncDataService.deleteRecordsBySyncId(sync.id!, limitRecords);
-    }
+    await db.knex.transaction(async (trx) => {
+        // Because it's slow and create deadlocks
+        // we need to acquire a Lock that prevents any other duplicate cron to execute the same thing
+        const { rows } = await trx.raw(`SELECT pg_try_advisory_xact_lock(?);`, [123456789]);
+        if (!rows || rows.lengt <= 0 || rows[0].pg_try_advisory_xact_lock === false) {
```
TBonnin:
```suggestion
        if (!rows || rows.length <= 0 || rows[0].pg_try_advisory_xact_lock === false) {
```

packages/shared/lib/services/sync/sync.service.ts
```diff
@@ -739,3 +739,10 @@ export async function findPausableDemoSyncs(): Promise<PausableSyncs[]> {
 
     return syncs;
 }
+
+export async function findRecentlyDeletedSync(): Promise<{ id: string }[]> {
+    const q = db.knex.from('_nango_syncs').select('_nango_syncs.id').where(db.knex.raw("_nango_syncs.deleted_at >  NOW() - INTERVAL '6h'"));
```
TBonnin: does that mean if the cron doesn't run for more than 6h we will end up missing some syncs to delete?

packages/jobs/lib/crons/deleteSyncsData.ts
```diff
@@ -26,30 +37,40 @@ export async function deleteSyncsData(): Promise<void> {
...
+            do {
+                countJobs = await softDeleteJobs({ syncId: sync.id, limit: limitJobs });
+                logger.info(`[deleteSyncs] soft deleted ${countJobs} jobs`);
+            } while (countJobs >= limitJobs);
+
+            // -----
+            // Soft delete schedules
+            let countSchedules = 0;
+            do {
+                countSchedules = await softDeleteSchedules({ syncId: sync.id, limit: limitSchedules });
```
TBonnin, line 66: should we discard failures and keep looping if an error happens?

packages/jobs/lib/crons/deleteSyncsData.ts
bodinsamuel: ah good catch, that's what you get with `any` 😠 thanks

packages/shared/lib/services/sync/sync.service.ts
bodinsamuel: yes as I said in the desc, it's currently not possible to achieve this in an efficient manner. In the best case it will run every 20 minutes so all good but not bullet proof. If you have suggestion I'm taking it.

packages/jobs/lib/crons/deleteSyncsData.ts
bodinsamuel: we could but that would requires more logic with try/catching and stuff. I don't know in advance how much I need to delete because running `COUNT(*)` is also super slow, so if I keep it running with errors it would go forever. It's safer like this and if any errors happen it should be fixed, wdyt?

packages/shared/lib/services/sync/sync.service.ts
TBonnin: 6 hours is not a lot for us to notice if the cron start to fail silently. The only thing I can think of is having some sort of queue :D

TBonnin (approved):

packages/jobs/lib/crons/deleteSyncsData.ts
TBonnin: I agree that explicitly failing is not a bad idea. However we also only have 6 hours to fix an issue or we risk to miss some deleted syncs.

khaliqgant (approved):
Having logs on this is nice but it would be nice to have some additional higher level (analytics) data to 
a) not have to go to the logs to check
b) have an idea of sync churn
c) make sure the cron and deletion logic is working as expected

code itself looks good but what are your thoughts on this?

bodinsamuel:
> a) not have to go to the logs to check
> b) have an idea of sync churn
> c) make sure the cron and deletion logic is working as expected

I can log more metrics yes, this cron is just a byproduct of DELETE /config and DELETE /sync but having the details can be interesting.

khaliqgant:
> I can log more metrics yes, this cron is just a byproduct of DELETE /config and DELETE /sync but having the details can be interesting.

Not blocking but could be a nice improvement IMO

packages/shared/lib/services/sync/sync.service.ts
bodinsamuel: I mean I started this change with 24h but there has been about 200 deleted syncs in that range, it's a lot of unnecessary calls every 20minutes. I'm totally okay to put back 24h if you think it's worth it? ☺️

TBonnin: your call. you have been looking at this problem way more than me

bodinsamuel:
Added some metrics, let me know if that's alright ☺️

packages/shared/lib/services/sync/sync.service.ts
bodinsamuel: I think I'll delete manually what's left and see how it behaves in prod. Hopefully no further iteration is necessary

TBonnin (approved):

PR #1777, opened 2024-03-01
[nan-467] add updatedAfter | updated_after param

khaliqgant:


packages/shared/lib/utils/error.ts
```diff
@@ -517,6 +517,11 @@ export class NangoError extends Error {
                 this.message = 'The sync is not currently running so cannot be cancelled';
                 break;
 
+            case 'incorrect_param':
+                this.status = 400;
+                this.message = `The parameter ${this.payload['incorrect']} is incorrect. Please ${this.payload['correct']}.`;
```
TBonnin: what about `The param 'INCORRECT' is invalid. Do you mean 'CORRECT'?`

packages/shared/lib/services/notification/webhook.service.ts
```diff
@@ -145,6 +149,7 @@ class WebhookService {
                 deleted: 0
             },
             syncType,
+            updatedAfter: dayjs(now).toDate().toISOString(),
```
TBonnin, line 152: what does dayjs bring that cannot be done with Date?

TBonnin (approved):

packages/shared/lib/utils/error.ts
khaliqgant: Updated with https://github.com/NangoHQ/nango/pull/1777/commits/26d6d1e272416dd8e77f21a50fe95f0c2c782fc1

packages/shared/lib/services/notification/webhook.service.ts
khaliqgant: Consistency with how the records.service is handling it is all really.

PR #1778, opened 2024-03-01
[nan-468] add cursor to metadata

khaliqgant:


packages/shared/lib/services/sync/data/records.service.ts
```diff
@@ -278,7 +278,7 @@ export async function getDataRecords(
                 END as last_action`),
             'json as record'
         );
-        result = encryptionManager.decryptDataRecords(result, 'record') as unknown as DataRecordWithMetadata[];
+        result = encryptionManager.decryptDataRecordsAndAddCursor(result, 'record') as unknown as DataRecordWithMetadata[];
```
TBonnin: as mentioned in the wrong PR :) I am not a huge fan of adding the cursor in the encryption manager. It has nothing to do with encryption and I think the cursor logic should live here in records.service

khaliqgant: Completely agree, but I didn't want to add another iteration loop in the records service. Could use a generator I suppose

bodinsamuel: or simply make a generic loop where you call both things? I mean you can create a generic function for it, that would still split the concerns and not repeat code
```ts
for (cont res of result) {
  const record = encryptionManager.decryptDataRecord(res)
  record.cursor = addCursor(res);
}
```

khaliqgant: Yup, and I guess I could also add make the update you just mentioned in Slack too 😅 https://nangohq.slack.com/archives/C067NRBH1LL/p1709626758131279

docs-v2/spec.yaml
```diff
@@ -803,6 +803,10 @@ paths:
                                                             type: string
                                                             description: The timestamp at which Nango last detected a change to this record
                                                             example: "2023-09-18T15:20:35.941305+00:00"
+                                                        cursor:
+                                                            type: string
+                                                            description: The current cursor of the record. Use this to fetch records updated after this record
```
bodinsamuel, line 808: is this really `>` or `>=`? Looking at the code, I feel like it's not after this record but "at this record" which is a detail but change things when you consume

bodinsamuel (approved):
Minor comments 👍🏻

packages/shared/lib/services/sync/data/records.service.ts
bodinsamuel: yes indeed that would be a great place; let's just not mix both work but if you do create a generic function I can rewrite the json operation there 👌🏻

docs-v2/spec.yaml
khaliqgant: It is correctly stated as after this record:

```
const decodedCursorValue = Buffer.from(cursorValue, 'base64').toString('ascii');
const [cursorSort, cursorId] = decodedCursorValue.split('||');

query = query.where((builder) =>
                builder.where('updated_at', '>', cursorSort).orWhere((builder) => builder.where('updated_at', '=', cursorSort).andWhere('id', '>', cursorId))
            );
```

packages/shared/lib/services/sync/data/records.service.ts
khaliqgant: Updated

docs-v2/spec.yaml
bodinsamuel: ah okay I didn't saw the query, thanks 👌🏻

packages/shared/lib/services/sync/data/records.service.ts
```diff
@@ -440,19 +441,27 @@ export async function getAllDataRecords(
...
 
-        const customerResult = result.map((item) => item.record);
+        const customerResult = result.map((item: RawDataRecordResult) => {
+            const decryptedRecord = encryptionManager.decryptDataRecord(item);
+            if (!decryptedRecord) {
+                return decryptedRecord;
+            }
+            const nextCursor = item.record._nango_metadata.last_modified_at.toString();
+            const id = item.id;
+            const encodedCursorValue = Buffer.from(`${nextCursor}||${id}`).toString('base64');
```
TBonnin, line 455: nitpick: delimiter can be a single char. That's always one char that won't be needed to be sent over the network :)

TBonnin (approved):

packages/shared/lib/services/sync/data/records.service.ts
```diff
@@ -440,19 +441,27 @@ export async function getAllDataRecords(
...
         );
 
-        const result = encryptionManager.decryptDataRecords(rawResult, 'record') as unknown as SyncDataRecord[];
-
         if (result.length === 0) {
             return { success: true, error: null, response: { records: [], next_cursor: nextCursor } };
         }
 
-        const customerResult = result.map((item) => item.record);
+        const customerResult = result.map((item: RawDataRecordResult) => {
```
bodinsamuel:
```suggestion
        const customerResult = result.map((item) => {
```

packages/shared/lib/utils/encryption.manager.ts
```diff
@@ -272,6 +272,32 @@ class EncryptionManager {
...
+            return dataRecord;
+        }
+
+        const { encryptedValue, iv, authTag } = record;
+
+        const decryptedString = this.decrypt(encryptedValue, iv, authTag);
+
+        const updatedRecord = {
+            ...JSON.parse(decryptedString)
+        };
```
bodinsamuel:
```suggestion
        } as RawDataRecordResult;
```

packages/shared/lib/utils/encryption.manager.ts
```diff
@@ -272,6 +272,32 @@ class EncryptionManager {
...
+
+        if (record['_nango_metadata']) {
+            updatedRecord['_nango_metadata'] = record['_nango_metadata'];
```
bodinsamuel, line 302: Seems like it should be this no?
```suggestion
            updatedRecord.record['_nango_metadata'] = record['_nango_metadata'];
```

packages/shared/lib/services/sync/data/records.service.ts
khaliqgant: Updated

packages/shared/lib/utils/encryption.manager.ts
khaliqgant: No, but I did make some improvements to the types so it is clearer with https://github.com/NangoHQ/nango/pull/1778/commits/ee85d254e59866a3a91f5c2c5d9d2088fd6dd304

PR #1780, opened 2024-03-03
fix(integration):stripe-app access token refresh on expire

hassan254-prog:
## Describe your changes
Nango was finding it hard to refresh `stripe-app`  access token upon expiration. This was because upon exchanging a code for an access token, there is no `expire_at` or `expire_in` value in the response. However, according to the [docs](https://docs.stripe.com/stripe-apps/api-authentication/oauth#refresh-access-token), the access token expires after an hour, which is equivalent to 3600 seconds. I added that value during the creation of the Stripe app token. This will help facilitate automatic refresh on access token expiration.

### DIFF
--- a/packages/shared/lib/clients/provider.client.ts
+++ b/packages/shared/lib/clients/provider.client.ts
@@ -22,6 +22,8 @@ class ProviderClient {
             case 'facebook':
             case 'tiktok-ads':
             case 'tiktok-accounts':
+            case 'stripe-app':
+            case 'stripe-app-sandbox':
                 return true;
             default:
                 return false;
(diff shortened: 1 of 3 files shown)


packages/shared/lib/clients/provider.client.ts
```diff
@@ -192,6 +200,70 @@ class ProviderClient {
...
+            if (response.status === 200 && response.data) {
+                return {
+                    access_token: response.data['access_token'],
+                    livemode: response.data['livemode'],
+                    refresh_token: response.data['refresh_token'],
+                    scope: response.data['scope'],
+                    stripe_publishable_key: response.data['stripe_publishable_key'],
+                    stripe_user_id: response.data['stripe_user_id'],
+                    token_type: response.data['token_type'],
+                    expires_in: 3600
```
khaliqgant: Can we make this a const and put this at the top of the file in the event it ever changes we can easily change it?

packages/shared/lib/clients/provider.client.ts
```diff
@@ -192,6 +200,70 @@ class ProviderClient {
...
```
khaliqgant: use const as mentioned below

packages/shared/lib/clients/provider.client.ts
hassan254-prog: Sure, I'll refactor both methods by placing `const expires_in` at the top of the file. This way, it will be easier to locate and modify if needed in the future. Thank you.

packages/shared/lib/clients/provider.client.ts
```diff
@@ -217,7 +290,7 @@ class ProviderClient {
                 };
             }
 
-            throw new NangoError('tiktok_token_request_error', response.data);
+            throw new NangoError('tiktok_token_request_error');
```
khaliqgant: Why this change? Seems unrelated

hassan254-prog: Reverted the change.

PR #1781, opened 2024-03-03
feat(SecretInput):handle null optionalvalue gracefully

hassan254-prog:
## Describe your changes
Ensure that the `SecretInput` component properly handles null values for the `optionalvalue` prop by providing fallbacks, preventing React warnings about passing null values to the input element's value prop.


packages/webapp/src/components/ui/input/SecretInput.tsx
```diff
@@ -13,13 +13,17 @@ type SecretInputProps = Omit<JSX.IntrinsicElements['input'], 'defaultValue'> & {
...
+) {
     const [isSecretVisible, setIsSecretVisible] = useState(false);
 
-    const [changedValue, setChangedValue] = useState(props.defaultValue);
+    const [changedValue, setChangedValue] = useState(defaultValue);
 
-    const value = props.optionalvalue === null ? '' : props.optionalvalue || changedValue;
-    const updateValue = props.setoptionalvalue || setChangedValue;
+    // Handle null value in optionalvalue
+    const value = optionalvalue ?? changedValue;
```
bodinsamuel: SecretInputProps never define `null` as accepted value, I wonder where it was happening and if it's allowed can you modify this type? (and maybe the other type too)

hassan254-prog: Just updated the data types for these. Thanks @bodinsamuel for this.

bodinsamuel: I still can't pass `optionalValue={null}` => `Type 'null' is not assignable to type 'string | undefined'.`

hassan254-prog: Should now work ☺️.Thanks

bodinsamuel (approved):
👌🏻

PR #1782, opened 2024-03-03
feat(webapp):improve field values on refresh

hassan254-prog:
## Describe your changes
Improve access token and refresh token fields to display a small message when refreshing tokens and update the values with the new tokens afterward. Revoked these displayed access tokens as well.
improve_oauth2

### DIFF
--- a/packages/webapp/src/pages/Connection/Authorization.tsx
+++ b/packages/webapp/src/pages/Connection/Authorization.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react';
+import { useState } from 'react';
 import { Prism } from '@mantine/prism';
 import { Loading } from '@geist-ui/core';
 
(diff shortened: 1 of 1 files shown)


packages/webapp/src/pages/Connection/Authorization.tsx
```diff
@@ -16,6 +17,15 @@ interface AuthorizationProps {
 
 export default function Authorization(props: AuthorizationProps) {
     const { connection, forceRefresh, loaded } = props;
+    const [refreshing, setRefreshing] = useState(false);
+
+    useEffect(() => {}, [refreshing]);
```
bodinsamuel: What's the use of this useEffect? ☺️

bodinsamuel (approved):
Looks good 👍🏻 
Personal opinion: usually we put a loader in the button instead (not sure what's Nango view about this)

packages/webapp/src/pages/Connection/Authorization.tsx
hassan254-prog: You are right, thanks for catching that, useEffect has no effect here ☺️. I was actually thinking instead of having refreshing in the SecretInput we can just change Loading icon to a spinner . What do you think?

bodinsamuel: imo I think it's better yes but I'm not the one deciding here so I can't say for sure :D

khaliqgant (approved):
Thanks!

PR #1784, opened 2024-03-04
show settings if cloud or enterprise

khaliqgant:
## Describe your changes
Enterprise should be able to access user and team settings

NAN-495

### DIFF
--- a/packages/webapp/src/components/LeftNavBar.tsx
+++ b/packages/webapp/src/components/LeftNavBar.tsx
@@ -13,7 +13,7 @@ import {
 } from '@heroicons/react/24/outline';
 
 import { useStore } from '../store';
-import { isCloud } from '../utils/utils';
+import { isCloud, isEnterprise } from '../utils/utils';
 import { useSignout } from '../utils/user';
 
 export enum LeftNavBarItems {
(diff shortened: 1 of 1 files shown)


packages/webapp/src/components/LeftNavBar.tsx
```diff
@@ -220,7 +220,7 @@ export default function LeftNavBar(props: LeftNavBarProps) {
                                 <span className="items-center w-32 text-center text-gray-400 justify-center text-left text-sm truncate">{email}</span>
                             </div>
                             <EllipsisHorizontalIcon className="flex h-5 w-5 ml-3 text-gray-400 cursor-pointer" />
-                            {isCloud() && showUserSettings && (
+                            {(isCloud() || isEnterprise()) && showUserSettings && (
```
TBonnin, line 223: can isCloud and isEnterprise be included in the `showUserSettings` logic?

khaliqgant: `showUserSettings` is just a useState prop so there is actual no logic there just a boolean. The entire logic piece could be abstracted out potentially but I don't see much of an advantage there.

TBonnin: showUserSettings alone is then not enough to determine if user settings should or should not be shown, which seems weird to me because that the main purpose of this boolean state. You are the react expert though so up to you

TBonnin (approved):

PR #1785, opened 2024-03-04
[nan-439] match syncs based on the sync id + model combo to prevent multiple dropdowns being open

khaliqgant:
## Describe your changes
In a multi model scenario the sync id will be the same so the dropdown might show more than once. Use a combination of the model return which has to be unique and the sync id to correctly identify the sync

NAN-439

### DIFF
--- a/packages/webapp/src/pages/Connection/Syncs.tsx
+++ b/packages/webapp/src/pages/Connection/Syncs.tsx
@@ -212,10 +212,7 @@ export default function Syncs(props: SyncsProps) {
                                         )}
                                     </div>
                                     <div className="relative interact-with-sync">
-                                        <EllipsisHorizontalIcon
-                                            className="flex h-5 w-5 cursor-pointer"
-                                            onClick={() => toggleDropdown(hashSync(sync))}
-                                        />
+                                        <EllipsisHorizontalIcon className="flex h-5 w-5 cursor-pointer" onClick={() => toggleDropdown(hashSync(sync))} />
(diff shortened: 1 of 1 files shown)


packages/webapp/src/pages/Connection/Syncs.tsx
```diff
@@ -208,8 +208,11 @@ export default function Syncs(props: SyncsProps) {
                                         )}
                                     </div>
                                     <div className="relative interact-with-sync">
-                                        <EllipsisHorizontalIcon className="flex h-5 w-5 cursor-pointer" onClick={() => toggleDropdown(sync.id)} />
-                                        {openDropdownId === sync.id && (
+                                        <EllipsisHorizontalIcon
+                                            className="flex h-5 w-5 cursor-pointer"
+                                            onClick={() => toggleDropdown(`${sync.id}${JSON.stringify(sync.models)}`)}
+                                        />
+                                        {openDropdownHash === `${sync.id}${JSON.stringify(sync.models)}` && (
```
TBonnin: can we have a hash function or even maybe a constant to document what's this hash is?

TBonnin (approved):

packages/webapp/src/pages/Connection/Syncs.tsx
khaliqgant: Thanks, updated with https://github.com/NangoHQ/nango/pull/1785/commits/f0415abe26dff5ce2401cc5676f1c1e4b1cd244a

PR #1788, opened 2024-03-04
[nan-501] add updateLogs trigger

khaliqgant:
## Describe your changes
Adds an additional trigger to ensure we have log messages to match to the parent log. With a log ravamp this should be changed so this is hopefully temporary-ish logic.

NAN-501

### DIFF
--- a/packages/webapp/src/pages/Activity.tsx
+++ b/packages/webapp/src/pages/Activity.tsx
@@ -84,6 +84,7 @@ export default function Activity() {
     const [expandedRow, setExpandedRow] = useState(-1);
     const [limit] = useState(20);
     const [logIds, setLogIds] = useState<number[]>([]);
+    const [updateLogs, setUpdateLogs] = useState(false);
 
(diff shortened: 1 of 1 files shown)


packages/webapp/src/pages/Activity.tsx
```diff
@@ -127,19 +128,24 @@ export default function Activity() {
...
-                return { ...activity, messages: messages ? messages.reverse() : [] };
-            });
-            setActivities(updatedActivities as ActivityResponse[]);
+    const { error: logActivitiesError } = useSWR(
+        () => (logIds.length > 0 ? `/api/v1/activity-messages?updateLogs=${updateLogs}&logIds=${logIds.join(',')}` : null),
+        {
+            onSuccess: (data: ActivityMessageResponse) => {
+                const updatedActivities = activities?.map((activity) => {
+                    const messages = data[activity.id];
+                    console.log(messages);
```
bodinsamuel:
```suggestion
```

bodinsamuel (approved):

PR #1790, opened 2024-03-04
[nan-464] fix precision issue

khaliqgant:
## Describe your changes
The main use of dayjs here was to make sure the timestamp is valid. If it is then use the timestamp passed in to maintain precision.

NAN-464

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12349,6 +12349,11 @@
             "version": "1.11.7",
             "license": "MIT"
         },
+        "node_modules/dayjs-plugin-utc": {
+            "version": "0.1.2",
+            "resolved": "https://registry.npmjs.org/dayjs-plugin-utc/-/dayjs-plugin-utc-0.1.2.tgz",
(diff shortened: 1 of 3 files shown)


bodinsamuel (commented):
Note if we use `dayjs` only for date validation we could just use `new Date()` and check for `Invalid Date`
Also this fix breaks the normalization, for example an user could just pass `2024` or `feb, 24` which I don't think will work with PG

khaliqgant:
> 2024

Thanks! Meant to keep this in draft as it is related to the db replication epic with https://github.com/NangoHQ/nango/pull/1775 and others. Still a WIP with the others.

khaliqgant:
> > 2024
> 
> Thanks! Meant to keep this in draft as it is related to the db replication epic with #1775 and others. Still a WIP with the others.

Updated with https://github.com/NangoHQ/nango/pull/1790/commits/938fbf36785d2b5895eb18f4694c469fc7e45091 to use iso 8601

TBonnin (approved):

PR #1792, opened 2024-03-04
Nango to db replication improvements

khaliqgant:
DO NOT MERGE

- This is a meta PR to test out the [Nango-to-DB replication improvements](https://linear.app/nango/project/nango-to-db-replication-improvements-7c920dad09a2) epic

## Describe your changes
Combines:
* https://github.com/NangoHQ/nango/pull/1790 -- fixes precision issue
* https://github.com/NangoHQ/nango/pull/1775 -- default pagination should be on `updated_at`
* https://github.com/NangoHQ/nango/pull/1777 -- add `updated_after` parameter

### DIFF
--- a/packages/server/lib/controllers/sync.controller.ts
+++ b/packages/server/lib/controllers/sync.controller.ts
@@ -122,7 +122,7 @@ class SyncController {
     // to deprecate
     public async getRecords(req: Request, res: Response, next: NextFunction) {
         try {
-            const { model, delta, offset, limit, sort_by, order, filter, include_nango_metadata } = req.query;
+            const { model, delta, updated_after, offset, limit, sort_by, order, filter, include_nango_metadata } = req.query;
             const environmentId = getEnvironmentId(res);
             const connectionId = req.get('Connection-Id') as string;
             const providerConfigKey = req.get('Provider-Config-Key') as string;
@@ -136,7 +136,7 @@ class SyncController {
                 providerConfigKey,
                 environmentId,
                 model as string,
-                delta as string,
+                String(delta || updated_after),
                 offset as string,
                 limit as string,
                 sort_by as string,
--- a/packages/shared/lib/services/sync/data/records.service.ts
+++ b/packages/shared/lib/services/sync/data/records.service.ts
@@ -343,7 +343,7 @@ export async function getAllDataRecords(
                 model
             })
             .orderBy([
-                { column: 'created_at', order: 'asc' },
+                { column: 'updated_at', order: 'asc' },
                 { column: 'id', order: 'asc' }
             ]);
 
@@ -358,7 +358,7 @@ export async function getAllDataRecords(
             }
 
             query = query.where((builder) =>
-                builder.where('created_at', '>', cursorSort).orWhere((builder) => builder.where('created_at', '=', cursorSort).andWhere('id', '>', cursorId))
+                builder.where('updated_at', '>', cursorSort).orWhere((builder) => builder.where('updated_at', '=', cursorSort).andWhere('id', '>', cursorId))
             );
         }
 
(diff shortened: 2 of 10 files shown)


packages/server/lib/controllers/sync.controller.ts
```diff
@@ -159,17 +159,31 @@ class SyncController {
...
 
+            if (updatedAfter) {
+                const error = new NangoError('incorrect_param', { incorrect: 'updatedAfter', correct: 'updated_after' });
+
+                errorManager.errResFromNangoErr(res, error);
+                return;
+            }
+
+            if (next_cursor) {
+                const error = new NangoError('incorrect_param', { incorrect: 'next_cursor', correct: 'cursor' });
```
khaliqgant, line 175: A customer recently shot themself in the foot with this: https://nangohq.slack.com/archives/C06GHQPLTAN/p1709405405531569

docs-v2/integrate/guides/receive-webhooks-from-nango.mdx
```diff
@@ -28,7 +28,8 @@ Webhooks from Nango are POST requests with the following body:
     "model": "<string>",
     "responseResults": { "<DataModel>": { "added": 123, "updated": 123, "deleted": 123 } },
     "syncType": "INITIAL" | "INCREMENTAL",
-    "queryTimeStamp": "<timestamp>", // Null for initial sync runs.
+    "updatedAfter": "<timestamp>", // Null for initial sync runs.
+    "queryTimeStamp": "<timestamp>", // DEPRECATED: (use updatedAfter) null for initial sync runs.
```
TBonnin, line 32: are we also adding the cursor in the webhook payload?

packages/shared/lib/services/sync/data/records.service.ts
```diff
@@ -373,19 +376,17 @@ export async function getAllDataRecords(
...
                 const error = new NangoError('invalid_timestamp');
 
                 return { success: false, error, response: null };
             }
 
-            const timeToDate = time.toDate();
-
-            const utcString = timeToDate.toUTCString();
-            query = query.andWhere('updated_at', '>=', utcString);
+            const formattedDelta = time.toISOString();
```
TBonnin, line 388: since we are moving away from `detla` can we name those const differently.
Ex: time -> updatedAfterTime, and formattedDelta => UpdateAfterTimeFormatted

packages/shared/lib/utils/encryption.manager.ts
```diff
@@ -245,6 +248,10 @@ class EncryptionManager {
 
             if (!record.encryptedValue) {
                 decryptedDataRecords.push(dataRecord);
+                const nextCursor = record['_nango_metadata']['last_modified_at'] as unknown as string;
+                const id = dataRecord['id'];
+                const encodedCursorValue = Buffer.from(`${nextCursor}||${id}`).toString('base64');
+                record['_nango_metadata']['cursor'] = encodedCursorValue;
```
TBonnin, line 254: I am not a huge fan of doing this in the encryption manager since it has nothing to do with encryption. IMHO this should live in `data.service`

khaliqgant: Same, but I didn't want to add another iteration loop in the records service. Could use a generator I suppose

docs-v2/integrate/guides/receive-webhooks-from-nango.mdx
khaliqgant: No, don't think so. The scope of the [issue](https://linear.app/nango/issue/NAN-468/return-the-cursor-in-the-record-metadata) says we don't need to return it as part of the webhook.

PR #1795, opened 2024-03-04
fix(manifest):fix property names

hassan254-prog:
## Describe your changes
Fix properties `name` and `short_name` on app manifest.


khaliqgant (approved):
Thanks!

PR #1796, opened 2024-03-05
Update asana.mdx

kishenpateldotwork:
## Describe your changes
Had issue with using Asana OAuth, then found this ticket with support providing the solution so wanted to update your docs as well: https://forum.asana.com/t/asana-app-client-id-not-valid-for-initiating-oauth2/404312


khaliqgant (approved):
Thanks!

PR #1797, opened 2024-03-05
[nan-439] allow no output for script

khaliqgant:
## Describe your changes
If an action has not output it would throw an error in the webapp

NAN-439

### DIFF
--- a/packages/webapp/src/utils/utils.tsx
+++ b/packages/webapp/src/utils/utils.tsx
@@ -322,6 +322,9 @@ export const parseInput = (flow: Flow) => {
 };
 
 export function generateResponseModel(models: NangoSyncModel[], output: string, isSync: boolean): Record<string, any> {
+    if (!output) {
+        return {};
+    }
     const model = models.find((model) => model.name === output);
     const jsonResponse = generateExampleValueForProperty(model as NangoSyncModel);
     if (!isSync) {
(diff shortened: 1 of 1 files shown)


packages/webapp/src/utils/utils.tsx
```diff
@@ -322,6 +322,9 @@ export const parseInput = (flow: Flow) => {
 };
 
 export function generateResponseModel(models: NangoSyncModel[], output: string, isSync: boolean): Record<string, any> {
```
bodinsamuel:
```suggestion
export function generateResponseModel(models: NangoSyncModel[], output: string | undefined, isSync: boolean): Record<string, any> {
```
The type is incorrect then, and that also means Flow's type is also wrong if `output`(returns) is optional

khaliqgant: Updated https://github.com/NangoHQ/nango/pull/1797/commits/71e47d002478a281ed8b820919d457903efdf529

output is optional already actually https://github.com/NangoHQ/nango/blob/master/packages/webapp/src/types.ts#L219

bodinsamuel: EndpointReference.tsx is using `returns` not `output` (I don't know the difference between those 2 properties)

bodinsamuel (approved):

packages/webapp/src/utils/utils.tsx
khaliqgant: `returns` is the v1 legacy property and `output` is the v2 property.

PR #1799, opened 2024-03-05
fix: jobs vulnerability

TBonnin:
jobs depends on the `api` package to make request to render api This package suffers from a vulnerability of one of its dependency, specifically the lodash.setWith package which is actually deprecated. 
There is a PR open for `api` to use full lodash instead of per method packages (which are deprecated) but it has not been merged yet. https://github.com/readmeio/api/pull/859

This commit replaces the `api` package used to generate a render sdk from their openapi spec by a home-made RenderAPI class (40 lines of code) 

https://linear.app/nango/issue/NAN-453/[credal]-fix-jobs-vulnerability


TBonnin:
Screenshot 2024-03-05 at 15 21 01

packages/cli/package.json
```diff
@@ -31,6 +31,7 @@
         "ajv-errors": "^3.0.0",
         "axios": "^1.2.0",
         "byots": "^5.0.0-dev.20221103.1.34",
```
bodinsamuel, line 33: unrelated, do we use this `byots` package? it seems not, or at least not imported 🤔

packages/cli/package.json
```diff
@@ -31,6 +31,7 @@
...
+        "chalk": "^5.3.0",
```
TBonnin, line 34: unrelated but chalk was not declared as CLI dependency

packages/jobs/lib/runner/render.runner.ts
```diff
@@ -68,17 +103,17 @@ export class RenderRunner implements Runner {
                     image: { ownerId: ownerId, imagePath: `nangohq/nango-runner:${imageTag}` },
                     serviceDetails: { env: 'image' },
                     envVars: [
-                        { key: 'NODE_ENV', value: process.env['NODE_ENV'] },
-                        { key: 'NANGO_CLOUD', value: process.env['NANGO_CLOUD'] },
+                        { key: 'NODE_ENV', value: process.env['NODE_ENV'] || NodeEnv.Dev },
+                        { key: 'NANGO_CLOUD', value: process.env['NANGO_CLOUD'] || 'true' },
```
bodinsamuel, line 71: I would default to false, because it would change the local development behavior
```suggestion
                        { key: 'NANGO_CLOUD', value: process.env['NANGO_CLOUD'] || 'false' },
```

packages/jobs/lib/runner/render.runner.ts
```diff
@@ -1,15 +1,50 @@
...
 import { ProxyAppRouter, getRunnerClient } from '@nangohq/nango-runner';
-import { getEnv, getPersistAPIUrl } from '@nangohq/shared';
-import api from 'api';
+import { AxiosResponse, NodeEnv, getEnv, getPersistAPIUrl } from '@nangohq/shared';
 import tracer from 'dd-trace';
+import axios, { AxiosInstance } from 'axios';
 
-const render = api('@render-api/v1.0#aiie8wizhlp1is9bu');
-render.auth(process.env['RENDER_API_KEY']);
+export class RenderAPI {
```
bodinsamuel: nitpicking, class could be in a separate file

bodinsamuel (commented):
Nice 🚀 Few comments but overall good!
I guess you already have tested this in staging?

packages/cli/package.json
TBonnin: good question. It is a bit out of the scope of this PR. I only happen to touch CLI package.json because it wouldn't compile without explicitly set chalk as a dependency but I am sure there is a  some more cleanup to do

packages/jobs/lib/runner/render.runner.ts
TBonnin: render runners are only used in the context of nango cloud and never locally

packages/jobs/lib/runner/render.runner.ts
TBonnin: sure

packages/jobs/lib/runner/render.runner.ts
bodinsamuel: ah true 👌🏻

bodinsamuel (approved):

packages/cli/package.json
khaliqgant: Agree of out of scope but we used `byots` in an earlier iteration of the compilation of the typescript files for the cli, but now use `ts-node` instead.

khaliqgant (approved):

PR #1800, opened 2024-03-05
[nan-477] hide interactive demo on enterprise and don't create example integration

khaliqgant:
## Describe your changes
1. Hide the interactive demo if not cloud
2. Default landing page for cloud is getting-started, otherwise is the `${env}/integrations`
3. Only allow access to getting started page if cloud
4. Only show sidebar getting-started page if cloud

NAN-477


packages/webapp/src/App.tsx
```diff
@@ -69,7 +69,7 @@ const App = () => {
             return url.pathname;
         }
 
-        return env === 'dev' ? '/dev/getting-started' : '/prod/integrations';
+        return env === 'dev' && isCloud() ? '/dev/getting-started' : `/${env}/integrations`;
```
TBonnin, line 72: `dev AND isCloud`. Is that even possible?

packages/webapp/src/components/LeftNavBar.tsx
```diff
@@ -154,7 +154,7 @@ export default function LeftNavBar(props: LeftNavBarProps) {
                         </div>
                     )}
                     <div className="space-y-1">
-                        {env === 'dev' && (
+                        {isCloud() && env === 'dev' && (
```
TBonnin, line 157: same question here

packages/webapp/src/App.tsx
khaliqgant: Dev environment 
image

on our cloud

TBonnin: got it.

TBonnin (approved):

packages/webapp/src/components/LeftNavBar.tsx
bodinsamuel: I didn't reviewed this sorry, but that means we can't access the getting started locally. It's probably okay but I'm not sure it was the goal :D

khaliqgant: No worries. It wasn't the goal but was necessary to properly lock the getting-started page down. The only reason it was so permissive before was to make it easy to work on locally. Locally already as it is we can't access the team page without editing some code in the `App.tsx` so I felt this was ok

PR #1803, opened 2024-03-06
fix(getting-started): remove some code snippets, re-up local testing

bodinsamuel:
## Describe your changes

Fixes NAN-469

- Remove unnecessary code snippets
- Reup the possibility to access getting started locally with a global state


packages/webapp/src/App.tsx
```diff
@@ -55,6 +55,12 @@ const App = () => {
     const SentryRoutes = Sentry.withSentryReactRouterV6Routing(Routes);
     const env = useStore((state) => state.cookieValue);
     const signout = useSignout();
+    const setShowGettingStarted = useStore((state) => state.setShowGettingStarted);
+    const showGettingStarted = useStore((state) => state.showGettingStarted);
+
+    useEffect(() => {
+        setShowGettingStarted(env === 'dev' && (isCloud() || isLocal()));
```
khaliqgant, line 62: why a useeffect here? What issue were you seeing?

khaliqgant (changes requested):
Question about useEffect inline

packages/webapp/src/App.tsx
bodinsamuel: It's just to setup the global state so we don't have to rewrite the condition multiple times, and keep it in sync with environment change. would you have done something differently?

khaliqgant: ah ok gotcha -- to keep things DRY. Woks for me.

khaliqgant (approved):

packages/webapp/src/App.tsx
```diff
@@ -55,6 +55,12 @@ const App = () => {
     const SentryRoutes = Sentry.withSentryReactRouterV6Routing(Routes);
     const env = useStore((state) => state.cookieValue);
     const signout = useSignout();
+    const setShowGettingStarted = useStore((state) => state.setShowGettingStarted);
+    const showGettingStarted = useStore((state) => state.showGettingStarted);
+
+    useEffect(() => {
```
TBonnin, line 61: should we use `useState` instead?
```
const [showGettingStarted, setShowGettingStarted] = useState(env === 'dev' && (isCloud() || isLocal()));
```

bodinsamuel: The same condition is reused in an other file, so it's less ideal

PR #1804, opened 2024-03-06
fix(eslint): improve webapp reporting

bodinsamuel:
## Describe your changes

WebApp was linted as a NodeJS process 

- Use proper browser env and tsconfig for WebApp
- Overrides some rules


.eslintrc
```diff
@@ -45,10 +49,44 @@
...
+            },
+            "rules": {
+                "no-console": "off",
+                "@typescript-eslint/no-misused-promises": [
+                    "warn",
+                    {
+                        "checksVoidReturn": {
+                            "arguments": false,
+                            "attributes": false
+                        }
```
bodinsamuel, line 86: I ignored those errors because every `onClick={}` were complaining about async being used. I feel like too many warnings is not great and this is not really helping the codebase, wdyt?

TBonnin (approved):

PR #1807, opened 2024-03-06
[enterprise] check for usage of s3 for remote files

khaliqgant:
## Describe your changes
Enterprise customers can use s3 or store integration files locally.


bodinsamuel (approved):
Feels like we would struggle less with env var `NANGO_STORAGE='s3' | 'local'`

TBonnin (approved):

khaliqgant:
> Feels like we would struggle less with env var `NANGO_STORAGE='s3' | 'local'`

Yeah, probably! Will add that in at a later date!

PR #1808, opened 2024-03-06
fix(ui): move and pin devDependencies

bodinsamuel:
## Describe your changes

Fixes NAN-519

- Move everything to `devDeps`
We install everything the UI as prod dependencies, which is irrelevant since the build does not require any dependencies.
Also if we go for a unified Docker it will reduce the install size by a large amount. 

- Pin dependencies
Open to debate. Having a caret means you are okay with transparent minor update, which in NPM ecosystem is very optimistic. Unless it's a library, in which case you do want a caret, pinning creates a stable reproductible build.
See recommendation https://docs.renovatebot.com/dependency-pinning/#so-whats-best

- Removed `jest` as we are not using it


bodinsamuel:
Tested in staging ✅ Bundle is also slightly less big (-20ko)

khaliqgant:
Will the CI install devDependencies when we build out the webapp as part of the `build:prod` command?

bodinsamuel:
> Will the CI install devDependencies when we build out the webapp as part of the build:prod command?

Currently build:prod does not specify anything differently than the others except the .env vars ☺️

khaliqgant (approved):

TBonnin:
the caret would make sense in a world full of unicorns where library new version (even patch ones) never breaks. This world doesn't exist unfortunately.

TBonnin (approved):

PR #1810, opened 2024-03-06
Add docs about cursor-based syncing

bastienbeurier:


docs-v2/integrate/guides/receive-webhooks-from-nango.mdx
```diff
@@ -28,8 +28,7 @@ Webhooks from Nango are POST requests with the following body:
     "model": "<string>",
     "responseResults": { "<DataModel>": { "added": 123, "updated": 123, "deleted": 123 } },
     "syncType": "INITIAL" | "INCREMENTAL",
-    "updatedAfter": "<timestamp>", // Null for initial sync runs.
-    "queryTimeStamp": "<timestamp>", // DEPRECATED: (use updatedAfter) null for initial sync runs.
```
bodinsamuel, line 32: We should probably remove `queryTimeStamp` from our code then no?

docs-v2/integrate/guides/sync-data-from-an-api.mdx
```diff
@@ -84,20 +81,17 @@ const result = await nango.listRecords({
...
 
 </Tabs>
 
-<Tip>
- If you consume the API (vs. the SDK), you can use the standard endpoint (above) or a unique generated endpoint described in the _API Reference_ tab of your integration in the Nango UI.
-</Tip>
+This returns an array of records conforming to the specified data model. 
 
-This returns an array of records conforming to the specified data model. Each record contains useful metadata.
+Each record contains useful metadata, including the cursor that you used to fetch the latest updates.
```
bodinsamuel:
```suggestion
Each record contains useful metadata, including the cursor that you use to fetch the latest updates. 
```

docs-v2/reference/sdks/node.mdx
```diff
@@ -690,7 +690,7 @@ const records = await nango.listRecords<ModelName>({
 ```
 
 <Warning>
-`nango.getRecords()` is deprecated and  will be removed in future releases as it does not support efficient pagination. Please use `nango.listRecords()` detailed below.
+`nango.getRecords()` is deprecated and will be removed in future releases as it does not support efficient pagination. Please use `nango.listRecords()` detailed below.
```
bodinsamuel, line 693: Note we could add `// @deprecated` to the node-client to make dev aware of this

bodinsamuel (approved):
looks good

docs-v2/integrate/guides/receive-webhooks-from-nango.mdx
bastienbeurier: The problem is that it would be a breaking change! Our plan was to 1. deprecate 2. warm about removal + log usage 3. remove entirely.

docs-v2/reference/sdks/node.mdx
bastienbeurier: Done!

docs-v2/integrate/guides/sync-data-from-an-api.mdx
bastienbeurier: Thanks!

docs-v2/integrate/guides/receive-webhooks-from-nango.mdx
bodinsamuel: yeah sorry, what I meant is that you removed it from the doc so it felt like all those work had been done

PR #1812, opened 2024-03-06
use updatedAfter instead of modified after

khaliqgant:
## Describe your changes
Decided that `modifiedAfter` is clearer than `updatedAfter`


TBonnin (approved):
looks good other than the compilation error

PR #1813, opened 2024-03-06
[NAN-510] return the connection(s) associated with the webhook forward

khaliqgant:
## Describe your changes
When a webhook is forwarded include the connectionIds. Since a connection can be created multiple times under a different name it is always plural even though in most cases it will be a single element in the array. Also adds in special handling for Slack.

NAN-510


packages/shared/lib/integrations/scripts/webhook/internal-nango.ts
```diff
@@ -11,22 +11,20 @@ import telemetry, { LogTypes } from '../../../utils/telemetry.js';
...
+        connectionIdentifier: string,
+        propName?: string
+    ): Promise<{ connectionIds: string[] }>;
 }
 
 export const internalNango: InternalNango = {
     getWebhooks: async (environment_id, nango_config_id) => {
         return await getSyncConfigsByConfigIdForWebhook(environment_id, nango_config_id);
     },
-    executeScriptForWebhooks: async (integration, body, webhookType, connectionIdentifier, propName) => {
```
khaliqgant, line 21: Moved this so that even if there are no webhook handlers for the incoming webhook we can still report associated connectionIds.

packages/shared/lib/services/notification/webhook.service.ts
```diff
@@ -317,6 +321,7 @@ class WebhookService {
 
         const headers = {
             ...nangoHeaders,
+            'Source-Payload-Content-Type': webhookOriginalHeaders['content-type'],
```
khaliqgant: Not sold on this name, but felt this is useful so that users will know what the providers content type was for the original payload. For providers like slack it isn't json.

khaliqgant:
The thing I don't like about this is the difference in interfaces with regards to the `connectionId` vs. `connectionIds`:

```
export interface NangoForwardWebhookBody {
    from: string;
    type: WebhookType.FORWARD;
    connectionIds: string[]; --> potential for confusion
    providerConfigKey: string;
    payload: any;
}

export interface NangoSyncWebhookBody {
    from: string;
    type: WebhookType.SYNC;
    connectionId: string;
    providerConfigKey: string;
    syncName: string;
    model: string;
    responseResults: SyncResult;
    syncType: SyncType;
    modifiedAfter: string | null;
    queryTimeStamp: string | null;
}
```

Thoughts @bastienbeurier @TBonnin @bodinsamuel ?

bastienbeurier:
> The thing I don't like about this is the difference in interfaces with regards to the `connectionId` vs. `connectionIds`:
> 
> ```
> export interface NangoForwardWebhookBody {
>     from: string;
>     type: WebhookType.FORWARD;
>     connectionIds: string[]; --> potential for confusion
>     providerConfigKey: string;
>     payload: any;
> }
> 
> export interface NangoSyncWebhookBody {
>     from: string;
>     type: WebhookType.SYNC;
>     connectionId: string;
>     providerConfigKey: string;
>     syncName: string;
>     model: string;
>     responseResults: SyncResult;
>     syncType: SyncType;
>     modifiedAfter: string | null;
>     queryTimeStamp: string | null;
> }
> ```
> 
> Thoughts @bastienbeurier @TBonnin @bodinsamuel ?

Why do we have multiple connection IDs here?

khaliqgant:
As [Gretel pointed](https://linear.app/nango/issue/NAN-226/[gretel]-webhook-improvements) out a webhook can have multiple connections associated with it.

> We don’t have a specific use-case for this, but we have been supporting this in our past integrations because it’s not impossible (a customer with access to some workspaces in the sources, could want to have more than one Gretel workspace, and connect their account to multiple Gretel workspaces).

packages/shared/lib/integrations/scripts/webhook/github-app-oauth-webhook-routing.ts
```diff
@@ -36,7 +36,7 @@ export default async function route(nango: Nango, integration: ProviderConfig, h
         await handleCreateWebhook(integration, body);
     }
 
-    await nango.executeScriptForWebhooks(integration, body, 'installation.id', 'installation_id');
+    return await nango.executeScriptForWebhooks(integration, body, 'installation.id', 'installation_id');
```
TBonnin: you can remove the await if we the promise is going to be returned

packages/shared/lib/integrations/scripts/webhook/github-app-webhook-routing.ts
```diff
@@ -27,5 +27,5 @@ export default async function route(nango: Nango, integration: ProviderConfig, h
         }
     }
 
-    await nango.executeScriptForWebhooks(integration, body, 'installation.id', 'installation_id');
+    return await nango.executeScriptForWebhooks(integration, body, 'installation.id', 'installation_id');
```
TBonnin: same here

packages/shared/lib/integrations/scripts/webhook/hubspot-webhook-routing.ts
```diff
@@ -37,10 +40,15 @@ export default async function route(nango: Nango, integration: ProviderConfig, h
...
+                if (response && response.connectionIds?.length > 0) {
+                    connectionIds = connectionIds.concat(response.connectionIds);
+                }
             }
         }
+
+        return { connectionIds };
     } else {
-        await nango.executeScriptForWebhooks(integration, body, 'subscriptionType', 'portalId');
+        return await nango.executeScriptForWebhooks(integration, body, 'subscriptionType', 'portalId');
```
TBonnin: same here

packages/shared/lib/integrations/scripts/webhook/internal-nango.ts
```diff
@@ -63,6 +61,12 @@ export const internalNango: InternalNango = {
             );
         }
 
+        const syncConfigsWithWebhooks = await internalNango.getWebhooks(integration.environment_id, integration.id as number);
+
+        if (syncConfigsWithWebhooks.length <= 0) {
+            return { connectionIds: connections?.map((connection) => connection.connection_id) || [] };
+        }
+
```
TBonnin: this block can go after the `if(!connections....` below, saving a call to `getWebhhoks` if no connections

packages/shared/lib/services/notification/webhook.service.ts
TBonnin: Should it be prefix with `x-nango` like the other headers?
what about `x-nango-source-content-type` for the name?

TBonnin (approved):

bastienbeurier:
> As [Gretel pointed](https://linear.app/nango/issue/NAN-226/%5Bgretel%5D-webhook-improvements) out a webhook can have multiple connections associated with it.
> 
> > We don’t have a specific use-case for this, but we have been supporting this in our past integrations because it’s not impossible (a customer with access to some workspaces in the sources, could want to have more than one Gretel workspace, and connect their account to multiple Gretel workspaces).

Does it make sense to fire multiple webhooks then? I agree that having an array of connection IDs is unusual, and might feel weird in the happy path.

packages/shared/lib/integrations/scripts/webhook/github-app-oauth-webhook-routing.ts
bodinsamuel: it's really up to each and everyone here, on my side I prefer to always await ahah 
a rule that I feels works when reviewing coding style: if there is an eslint rule for that, then we pick a solution and no more debate; otherwise it still can be suggested but as personal opinion

bodinsamuel:
> Does it make sense to fire multiple webhooks then? I agree that having an array of connection IDs is unusual, and might feel weird in the happy path.

I think it make sense to have one webhook with multiple `connectionId`. 
However if we think longer term, maybe some people will ask us to pass connection level data with the webhook, and then suddenly it's a mess to send in one payload.

bodinsamuel (approved):
Looks good to me if we stick with this solution 👌🏻

khaliqgant:
@bodinsamuel @TBonnin opted for multiple webhooks per connection with https://github.com/NangoHQ/nango/pull/1813/commits/4cc340ddae85301b1a389d579300da350089dd02

Can you check this out again please?

TBonnin (approved):

PR #1814, opened 2024-03-07
do not use jsonb_set in postgreSQL

TBonnin:
## Describe your changes
Appending _nango_metadata to the record json in postgres tends to be slow when json is big
This commit remove the call to jsonb_set in the query and manipulate the json in js land to add _nango_metadata child attributes

https://linear.app/nango/issue/NAN-515/fetching-big-records-is-slow

### DIFF
--- a/packages/shared/lib/models/Sync.ts
+++ b/packages/shared/lib/models/Sync.ts
@@ -216,23 +216,22 @@ export type CustomerFacingDataRecord = {
     _nango_metadata: RecordMetadata;
 } & Record<string, any> & { id: string | number };
 
-export interface EncryptedRecord {
+export type EncryptedRawRecord = {
     iv: string;
     authTag: string;
     encryptedValue: string;
-}
+};
 
-export type EncryptedInternalDataRecord = {
-    _nango_metadata: RecordMetadata;
-} & EncryptedRecord;
+export type UnencryptedRawRecord = Record<string, any> & { id: string | number };
 
-export type GetRecordsResponse = { records: CustomerFacingDataRecord[] | DataRecordWithMetadata[]; next_cursor?: string | null } | null;
-
-export interface RawDataRecordResult {
+export type RawDataRecordResult = {
     id: string | number;
-    record: CustomerFacingDataRecord | EncryptedInternalDataRecord;
-}
+    record: UnencryptedRawRecord | EncryptedRawRecord;
+} & RecordMetadata;
 
+export type GetRecordsResponse = { records: CustomerFacingDataRecord[]; next_cursor?: string | null } | null;
+
+// TO DEPRECATE
 export type RecordWrapCustomerFacingDataRecord = { record: CustomerFacingDataRecord }[];
 
 export interface DataRecord extends Timestamps {
@@ -256,13 +255,14 @@ export interface DataRecord extends Timestamps {
 export type LastAction = 'ADDED' | 'UPDATED' | 'DELETED' | 'added' | 'updated' | 'deleted';
 
 interface RecordMetadata {
-    first_seen_at: Date;
-    last_modified_at: Date;
+    first_seen_at: string;
+    last_modified_at: string;
     last_action: LastAction;
-    deleted_at: Date | null;
+    deleted_at: string | null;
     cursor: string;
 }
 
+// DEPRECATED
 export interface DataRecordWithMetadata extends RecordMetadata {
     record: object;
 }
--- a/packages/shared/lib/services/sync/data/mocks.ts
+++ b/packages/shared/lib/services/sync/data/mocks.ts
@@ -3,15 +3,64 @@ import { createSyncSeeds } from '../../../db/seeders/sync.seeder.js';
 import { createSyncJobSeeds } from '../../../db/seeders/sync-job.seeder.js';
 import { formatDataRecords } from './records.service.js';
 import type { DataResponse } from '../../../models/Data.js';
+import connectionService from '../../connection.service.js';
+import * as DataService from './data.service.js';
+import type { Connection } from '../../../models/Connection.js';
+
+export async function upsertRecords(n: number): Promise<{ connection: Connection; model: string }> {
+    const activityLogId = 1;
+    const environmentId = 1;
+    const environmentName = 'mock-records';
+    const toInsert = generateInsertableJson(n);
+    const {
+        response: { response: records },
+        meta: { modelName, nangoConnectionId }
+    } = await createRecords(toInsert, environmentName);
+    if (!records) {
+        throw new Error('Failed to format records');
+    }
+    const connection = await connectionService.getConnectionById(nangoConnectionId);
+    if (!connection) {
+        throw new Error(`Connection '${nangoConnectionId}' not found`);
+    }
+    const chunkSize = 1000;
+    for (let i = 0; i < records.length; i += chunkSize) {
+        const { error, success } = await DataService.upsert(
+            records.slice(i, i + chunkSize),
+            '_nango_sync_data_records',
+            'external_id',
+            nangoConnectionId,
+            modelName,
+            activityLogId,
+            environmentId
+        );
+        if (!success) {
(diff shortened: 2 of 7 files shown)


packages/shared/lib/models/Sync.ts
```diff
@@ -256,13 +255,14 @@ export interface DataRecord extends Timestamps {
 export type LastAction = 'ADDED' | 'UPDATED' | 'DELETED' | 'added' | 'updated' | 'deleted';
 
 interface RecordMetadata {
-    first_seen_at: Date;
-    last_modified_at: Date;
+    first_seen_at: string;
+    last_modified_at: string;
     last_action: LastAction;
-    deleted_at: Date | null;
+    deleted_at: string | null;
```
TBonnin, line 261: those were never date since they are coming from pg as json encoded date.

packages/shared/lib/models/Sync.ts
```diff
@@ -256,13 +255,14 @@ export interface DataRecord extends Timestamps {
...
     cursor: string;
 }
 
+// DEPRECATED
 export interface DataRecordWithMetadata extends RecordMetadata {
     record: object;
 }
```
TBonnin, line 268: only use in the deprecated records function so it should be removed when we finally delete the deprecated record fetching method

packages/shared/lib/services/sync/data/records.service.integration.test.ts
```diff
@@ -1,56 +1,59 @@
...
+        const timestampRegex = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{6}\+\d{2}:\d{2}$/;
+        expect(response?.records[0]?.['_nango_metadata']).toMatchObject({
+            first_seen_at: expect.stringMatching(timestampRegex),
+            last_modified_at: expect.stringMatching(timestampRegex),
+            last_action: 'ADDED',
+            deleted_at: null,
+            cursor: expect.stringMatching(/^[A-Za-z0-9+/]+={0,2}$/) // base64 encoded string
+        });
+        expect(response?.next_cursor).toBe(null); // no next page
+    });
```
TBonnin, line 37: adding a test to make sure the format of metadata isn't modified with those changes

packages/shared/lib/services/sync/data/records.service.ts
```diff
@@ -421,57 +421,66 @@ export async function getAllDataRecords(
             }
         }
 
-        let nextCursor = null;
-
-        const result: RawDataRecordResult[] = await query.select(
+        const rawResults: RawDataRecordResult[] = await query.select(
+            // PostgreSQL stores timestamp with microseconds precision
+            // however, javascript date only supports milliseconds precision
+            // we therefore convert timestamp to string (using to_json()) in order to avoid precision loss
```
TBonnin, line 427: basically we cannot query pg timestamp column into a js Date without losing the microseconds precision (which then make the cursor logic wrong) so all dates must be json encoded string

packages/shared/lib/services/sync/data/records.service.ts
```diff
@@ -512,26 +521,6 @@ export function verifyUniqueKeysAreUnique(data: DataResponse[], optionalUniqueKe
...
-
-    const result = encryptionManager.decryptDataRecords(encryptedRecord, 'json');
-
-    if (!result || result.length === 0) {
-        return null;
-    }
-
-    return result[0] as unknown as SyncDataRecord;
-}
-
```
TBonnin, line 534: unrelated but not used anywhere

packages/shared/lib/models/Sync.ts
khaliqgant: Should update the types in the node-client too to match these: https://github.com/NangoHQ/nango/blob/master/packages/node-client/lib/index.ts#L83

khaliqgant (changes requested):
Looks good and thanks for adding tests. I have a [follow up](https://linear.app/nango/issue/NAN-513/add-tests-to-the-listrecrods-call-to-solidify-behavior) to add more tests that will hopefully assist when you do the records rework. Feedback inline about the node-client

packages/shared/lib/models/Sync.ts
TBonnin: fix in 7c906107

TBonnin:
@khaliqgant can I have a second look. I fix the type in the node client and did some code cleanup

khaliqgant (approved):
nice

packages/shared/lib/models/Sync.ts
bodinsamuel: knex is doing some transformation so I guess it was kinda true :D

packages/shared/lib/models/Sync.ts
bodinsamuel, line 268:
```suggestion
/**
 * @deprecated
 */
export interface DataRecordWithMetadata extends RecordMetadata {
    record: object;
}
```

packages/shared/lib/services/sync/data/mocks.ts
```diff
@@ -3,15 +3,64 @@ import { createSyncSeeds } from '../../../db/seeders/sync.seeder.js';
...
+        meta: { modelName, nangoConnectionId }
+    } = await createRecords(toInsert, environmentName);
+    if (!records) {
+        throw new Error('Failed to format records');
+    }
+    const connection = await connectionService.getConnectionById(nangoConnectionId);
+    if (!connection) {
+        throw new Error(`Connection '${nangoConnectionId}' not found`);
+    }
+    const chunkSize = 1000;
```
bodinsamuel, line 26: personal opinion: code without line break is a bit harder to read in the long run

packages/shared/lib/models/Sync.ts
TBonnin: not even. knex was fine with those fields defined as date but at runtime `typeof` would be `string` 😡

packages/shared/lib/services/sync/data/records.service.integration.test.ts
```diff
@@ -1,56 +1,59 @@
...
+        const { success, response, error } = await RecordsService.getAllDataRecords(
+            connection.connection_id,
+            connection.provider_config_key,
+            connection.environment_id,
+            model
+        );
+        expect(success).toBe(true);
+        expect(error).toBe(null);
+        expect(response?.records.length).toBe(n);
+        const timestampRegex = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{6}\+\d{2}:\d{2}$/;
```
bodinsamuel, line 28: I had created an helper in vitest to achieve something similar without copy pasting

```ts
const dateReg = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{3}Z$/;

expect.extend({
  toBeIsoDate: (received: any) => {
    if (!dateReg.test(received)) {
      return {
        message: () => `expected ${received} to be an ISO Date`,
        pass: false,
      };
    }
    return { pass: true, message: () => '' };
  },
});
```

packages/shared/lib/services/sync/data/records.service.integration.test.ts
```diff
@@ -1,56 +1,59 @@
...
-                modelName, // model
+            const result = await RecordsService.getAllDataRecords(
+                connection.connection_id,
+                connection.provider_config_key,
+                connection.environment_id,
+                model,
                 undefined, // delta
                 limit, // limit
                 undefined, // filter
                 cursor // cursor
```
bodinsamuel, line 55: json params would make this better :D

bodinsamuel (approved):
Looks good 🚀 
Thanks for doing that, did you noticed something to make you jump on it?

packages/shared/lib/services/sync/data/records.service.integration.test.ts
TBonnin: I  extended expect at first as well but since it was actually shorter to use stringMatching I went for that

PR #1815, opened 2024-03-07
[nan-520] if more than 25 hours then let the incremental sync run

khaliqgant:
## Describe your changes
Situation came up for two customers. One self hosted there was issues in their setup and so the initial sync didn't complete so was still in a running state. If an `INITIAL` sync has been running for more than 25 hours (1 hour grace period) then allow incremental syns to run since we can assume it is stuck

NAN-520 

### DIFF
--- a/packages/shared/lib/services/sync/job.service.ts
+++ b/packages/shared/lib/services/sync/job.service.ts
@@ -6,6 +6,8 @@ import { Job as SyncJob, SyncStatus, SyncType, SyncResultByModel } from '../../m
 
 const SYNC_JOB_TABLE = dbNamespace + 'sync_jobs';
 
+const SYNC_TIMEOUT_HOURS = 25;
+
 export const createSyncJob = async (
     sync_id: string,
     type: SyncType,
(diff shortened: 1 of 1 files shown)


packages/shared/lib/services/sync/job.service.ts
```diff
@@ -165,7 +167,11 @@ export const isInitialSyncStillRunning = async (sync_id: string): Promise<boolea
         })
         .first();
 
-    if (result) {
+    // if it has been running for more than 24 hours then we should assume it is stuck
+    const moreThan24Hours =
+        result && result.updated_at ? new Date(result.updated_at).getTime() < new Date().getTime() - SYNC_TIMEOUT_HOURS * 60 * 60 * 1000 : false;
```
bodinsamuel, line 172: but I thought temporal would send us the timeout?

khaliqgant: The type of failure here is outside of the temporal context, but yes for a straight forward timeout we handle it and change the job status: https://github.com/NangoHQ/nango/blob/master/packages/jobs/lib/activities.ts#L385 and as seen in [datadog](https://us3.datadoghq.com/logs?query=environment%3Aproduction%20tags%3Aflow_job_failure%20&agg_t=count&cols=host%2Cservice&fromUser=false&index=%2A&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=timeseries&from_ts=1707208584262&to_ts=1709800584262&live=true)

bodinsamuel: initial sync is managed by temporal right? I would expect both regular workflow and this workflow to send timeout error so we don't have to have those conditions

khaliqgant: It is, but there are failures that temporal won't know about aka the jobs instance crashing

bodinsamuel: if jobs is crashing then temporal should send a timeout to the server no? Maybe I miss-understand something but if it works for regular sync it should here too

PR #1816, opened 2024-03-07
fix(integration): teamtailor integration sync template

hassan254-prog:
## Describe your changes
Reduce the page size for the example sync template.

### DIFF
--- a/integration-templates/teamtailor/teamtailor-candidates.ts
+++ b/integration-templates/teamtailor/teamtailor-candidates.ts
@@ -11,7 +11,7 @@ export default async function fetchData(nango: NangoSync): Promise<void> {
                 link_path_in_response_body: 'links.next',
                 limit_name_in_request: 'page[size]',
                 response_path: 'data',
-                limit: 100
+                limit: 30
             }
         };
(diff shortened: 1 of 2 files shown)


integration-templates/teamtailor/teamtailor-candidates.ts
```diff
@@ -11,7 +11,7 @@ export default async function fetchData(nango: NangoSync): Promise<void> {
                 link_path_in_response_body: 'links.next',
                 limit_name_in_request: 'page[size]',
                 response_path: 'data',
-                limit: 100
+                limit: 30
```
bodinsamuel, line 14: is there any linear issue associated with this change? 
if so can you think about linking them in your PR desc, that would be great to add some more context ☺️

bodinsamuel (approved):

khaliqgant (approved):

integration-templates/teamtailor/teamtailor-candidates.ts
hassan254-prog: Thanks @bodinsamuel for approving this 🙏. I will do that with the upcoming pull requests ☺️.

PR #1820, opened 2024-03-07
[nan-448] allow enterprise to sign up with a known token

khaliqgant:
## Describe your changes
1. Enterprise customer sets `NANGO_ADMIN_INVITE_TOKEN` to a known value
2. On creation of their Nango instance they navigate to `{customer}.nango.com/signup/{known-token}`
3. They enter their name and password, get directed to the signin page
4. Signin

Also disables the signup page for enterprise
Adds more permissive route access for local

NAN-448


packages/shared/lib/services/user.service.ts
```diff
@@ -136,6 +137,18 @@ class UserService {
     async getInvitedUserByToken(token: string): Promise<InviteUser | null> {
         const date = new Date();
 
+        if (isEnterprise() && process.env['NANGO_ADMIN_INVITE_TOKEN'] === token) {
+            return {
```
TBonnin, line 141: can we return null instead of a empty user?

packages/webapp/src/App.tsx
```diff
@@ -140,17 +140,17 @@ const App = () => {
                         <Route path="/:env/project-settings" element={<ProjectSettings />} />
                     </Route>
                     <Route path="/auth-link" element={<AuthLink />} />
-                    {(isCloud() || isEnterprise()) && (
+                    {(isCloud() || isEnterprise() || isLocal()) && (
```
TBonnin, line 143: when is this condition not true anymore now?

bodinsamuel (approved):

packages/shared/lib/services/user.service.ts
khaliqgant: Well, if we return null then the path would think there is no actual user so it is an empty user to get around that.

packages/webapp/src/App.tsx
khaliqgant: hosted

packages/shared/lib/services/user.service.ts
TBonnin: ok. can we add a comment to explain that in the code. I am sure if I read it again in 3 months I would have forgotten about it

packages/webapp/src/App.tsx
TBonnin: we are gonna need an enum at some point to track all different type of installations :D

TBonnin (approved):

packages/shared/lib/services/user.service.ts
bodinsamuel: I think once again, a good env var (or a computed env var) would be easier to follow. Ultimately those are all feature flags that could be centralized.

khaliqgant:
Thanks both. I'll make a Linear issue around flag consolidation to clean up this enterprise logic.

khaliqgant:
> Thanks both. I'll make a Linear issue around flag consolidation to clean up this enterprise logic.

https://linear.app/nango/issue/NAN-533/clean-up-isenterprise-logic-in-favor-of-enumenv-variable

PR #1822, opened 2024-03-07
[nan-324] check for pnpm

khaliqgant:
## Describe your changes
User is using `pnpm` and can't upgrade properly. Detect pnpm and call the correct upgrade action

NAN-324


packages/cli/lib/utils.ts
```diff
@@ -168,9 +168,16 @@ export async function upgradeAction(debug = false) {
...
+        const packagePath = getPackagePath();
+        const usePnpm = path.resolve(packagePath, '..').includes('.pnpm');
+
+        const args = usePnpm
+            ? locallyInstalled
+                ? ['add', `nango@${latestVersion}`]
+                : ['add', '-g', `nango@${latestVersion}`]
+            : locallyInstalled
+              ? ['install', '--no-audit', '--save', `nango@${latestVersion}`]
+              : ['install', '-g', '--no-audit', `nango@${latestVersion}`];
```
TBonnin: a double `if` is easier to read imho. 

I wish typescript has pattern matching

khaliqgant: yeah, I loved it when I was writing scala.

Updated with https://github.com/NangoHQ/nango/pull/1822/commits/e221defc153298fc2139e2eaa26e055da022e0b2

PR #1648, opened 2024-02-09
feat(activity):connection and integration filters in activities

hassan254-prog:
## Describe your changes
Integrations and connections filters now fetch from activities. Also fixed the issue where filtered result differs in `offset` value from the one requested to the server.
activity

## Checklist before requesting a review (skip if just adding/editing APIs & templates)
- [ ] I added tests, otherwise the reason is: 
- [ ] I added observability, otherwise the reason is:
- [ ] I added analytics, otherwise the reason is:

PR #1825, opened 2024-03-08
Khaliq/nan 525 upfort logs filter bug

khaliqgant:
## Describe your changes
Integration filter should use the provider config key instead of the provider. This bug was introduced by https://github.com/NangoHQ/nango/pull/1648

NAN-525

### DIFF
--- a/packages/shared/lib/services/activity/activity.service.ts
+++ b/packages/shared/lib/services/activity/activity.service.ts
@@ -248,7 +248,7 @@ export async function getTopLevelLogByEnvironment(
     return logs || [];
 }
 
-export async function activityFilter(environment_id: number, filterColumn: 'connection_id' | 'provider'): Promise<string[]> {
+export async function activityFilter(environment_id: number, filterColumn: 'connection_id' | 'provider_config_key'): Promise<string[]> {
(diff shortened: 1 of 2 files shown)


packages/shared/lib/services/activity/activity.service.ts
```diff
@@ -261,7 +266,7 @@ export async function activityFilter(environment_id: number, filterColumn: 'conn
 
     const distinctValues: string[] = logs
         .map((log: Record<string, string>) => log[filterColumn] as string)
-        .filter((value: string | undefined): value is string => typeof value === 'string');
+        .filter((value: string | undefined): value is string => typeof value === 'string' && value !== '');
```
bodinsamuel: could be done is sql no? `where <> ''`

bodinsamuel (approved):

packages/shared/lib/services/activity/activity.service.ts
khaliqgant: Indeed, updated with https://github.com/NangoHQ/nango/pull/1825/commits/5f9dbd2066bb0472fd69e4d1cf823767083753a8

PR #1826, opened 2024-03-08
fix(deps): pin aws-sdk

bodinsamuel:
## Describe your changes

Fixes NAN-530

AWS upgraded their sdk with a broken version, it's usually better to have a `^` for libraries but in that case it's a side-effect and this library is not used by the CLI. Ultimately, the CLI should not import this package at all but in the meantime it's safer to pin back to something that was working.

- Pin dependency to last working version

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -154,588 +154,1024 @@
                 "tslib": "^1.11.1"
             }
         },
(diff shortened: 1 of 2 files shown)


khaliqgant (approved):
Thanks Samuel

TBonnin (approved):

PR #1827, opened 2024-03-08
[nan-526] for OAauth2 force an expiration unless explicitly stated

khaliqgant:
## Describe your changes
A customer imported connections and didn't import the expiration date. Then one day the connections stopped working because the system never know when to refresh their access token. Unless a provider provides tokens that never expire force on import an expires_at or expires_in parameter

NAN-526

### DIFF
--- a/packages/server/lib/controllers/connection.controller.ts
+++ b/packages/server/lib/controllers/connection.controller.ts
@@ -535,7 +535,7 @@ class ConnectionController {
                     return;
                 }
 
-                if (!parsedExpiresAt && noExpiration && noExpiration !== true) {
+                if (!parsedExpiresAt && noExpiration !== true) {
                     errorManager.errRes(res, 'missing_expires_at');
                     return;
                 }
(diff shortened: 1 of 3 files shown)


TBonnin:
what does importing a connection means? using the api?

khaliqgant:
> what does importing a connection means? using the api?

Yeah exactly: http://localhost:3033/reference/api/connection/post

packages/server/lib/controllers/connection.controller.ts
```diff
@@ -534,6 +534,12 @@ class ConnectionController {
                     errorManager.errRes(res, 'missing_access_token');
                     return;
                 }
+
+                if (!parsedExpiresAt && noExpiration && noExpiration !== true) {
```
TBonnin: why the `&& noExpiration`? isn't the last part enough?

TBonnin (approved):

packages/server/lib/controllers/connection.controller.ts
khaliqgant: Thanks, removed with https://github.com/NangoHQ/nango/pull/1827/commits/964a99c105175d9931dc57d9daf042728cade61e

PR #1829, opened 2024-03-08
Algolia api support

hassan254-prog:
## Describe your changes

- Added configuration for algolia in providers.yaml
- Added algolia to docs-v2.
- Added algolia pages for Supported API groups in mint.json.
- Added algolia SVG file in template-logos folder.
- Consolidate algolia docs pages
- Added an action template that adds a single record contact to an index.

## Test
This provider has been used successfully in local development to connect to Algolia and successfully call those endpoints after authorizing using Nango.

The documentation update was reviewed using `npm run docs` and verifying on localhost.

### DIFF
--- a/docs-v2/integrations/integration-templates/algolia.mdx
+++ b/docs-v2/integrations/integration-templates/algolia.mdx
@@ -12,7 +12,7 @@ sidebarTitle: 'Algolia'
 </Card>
 
 <Card title="Get the Algolia template"
-      href="https://github.com/NangoHQ/nango/tree/master/integration-templates/teamtailor"
+      href="https://github.com/NangoHQ/nango/tree/master/integration-templates/algolia"
       icon="github">
     Get the latest version of the Algolia integration template from GitHub
 </Card>


integration-templates/algolia/algolia-create-contacts.ts
```diff
@@ -0,0 +1,25 @@
...
+            data: postData
+        });
+
+        return {
+            createdAt: resp.data.createdAt,
+            taskID: resp.data.taskID,
+            objectID: resp.data.objectID
+        };
+    } catch (error: any) {
+        throw new Error(`Error in runAction: ${error.message}`);
```
khaliqgant: use nango.ActionError

integration-templates/algolia/algolia-create-contacts.ts
```diff
@@ -0,0 +1,25 @@
+import type { NangoAction, AlgoliaContact, AlgoliaCreateContactInput } from './models';
+
+export default async function runAction(nango: NangoAction, input: AlgoliaCreateContactInput): Promise<AlgoliaContact> {
+    const endpoint = `/1/indexes/contacts`;
+
+    try {
+        const postData = {
```
khaliqgant, line 13: validate these inputs

packages/shared/providers.yaml
```diff
@@ -61,6 +61,16 @@ airtable:
     proxy:
         base_url: https://api.airtable.com
     docs: https://docs.nango.dev/integrations/all/airtable
+algolia:
+    categories:
+        - other
```
khaliqgant: Feel like a better category can be used here. `search` is the obvious one but unsure if others would exist in this category.

hassan254-prog: I was actually thinking of that, we can add a new category and then later as I backfill scripts, I can add other providers to the search category. What do you think about that?

integration-templates/algolia/algolia-create-contacts.ts
hassan254-prog: This endpoint doesn't have any required fields. If you pass an empty object as your input, a new objectID is created. Do we still need to validate these fields?

khaliqgant: We should have at least one identifier go into the system. AN empty object doesn't seem like much use.

packages/shared/providers.yaml
khaliqgant: That works!

hassan254-prog: Great I will do that.

integration-templates/algolia/algolia-create-contacts.ts
hassan254-prog: That works, I will add input name validation.

packages/shared/providers.yaml
```diff
@@ -61,6 +61,16 @@ airtable:
     proxy:
         base_url: https://api.airtable.com
     docs: https://docs.nango.dev/integrations/all/airtable
+algolia:
+    categories:
+        - search
+    auth_mode: API_KEY
+    proxy:
+        base_url: https://${connectionConfig.APP_ID}.algolia.net
```
bodinsamuel, line 69: Note that not all customers have this url, on specific scenarios but should be fine rn

integration-templates/algolia/algolia-create-contacts.ts
```diff
@@ -0,0 +1,33 @@
...
+            email: input.email
+        };
+        const resp = await nango.post({
+            endpoint: endpoint,
+            data: postData
+        });
+
+        return {
+            createdAt: resp.data.createdAt,
+            taskID: resp.data.taskID,
```
bodinsamuel, line 25: note that the API gives back a TaskID because the indexing is not synchronous. I guess we don't care for the template but it could be something users should be aware of.

integration-templates/algolia/algolia-create-contacts.ts
```diff
@@ -0,0 +1,33 @@
...
+
+    const endpoint = `/1/indexes/contacts`;
+
+    try {
+        const postData = {
+            name: input.name,
+            company: input.company,
+            email: input.email
+        };
+        const resp = await nango.post({
```
bodinsamuel, line 18: Note: the Algolia API in itself is not fully resilient; the sdks contains a little bit of logic to target different servers in case of downtime. We probably don't care but for power user they will need to load the `algoliasearch` npm package somehow

docs-v2/integrations/all/algolia.mdx
```diff
@@ -0,0 +1,32 @@
...
+---
+
+API configuration: [`algolia`](https://nango.dev/providers.yaml)
+
+## Features
+
+| Features | Status |
+| - | - |
+| [Auth (API Key)](/integrate/guides/authorize-an-api) | ✅ |
+| [Sync data](/integrate/guides/sync-data-from-an-api) | ✅ |
```
bodinsamuel, line 13: Are those really ✅?

docs-v2/integrations/integration-templates/algolia.mdx
```diff
@@ -0,0 +1,21 @@
...
+## Get started with the Algolia template
+
+<Card title="How to use integration templates"
+      href="/understand/concepts/templates"
+      icon="book-open">
+    Learn how to use integration templates in Nango
+</Card>
+
+<Card title="Get the Algolia template"
+      href="https://github.com/NangoHQ/nango/tree/master/integration-templates/teamtailor"
```
bodinsamuel:
```suggestion
      href="https://github.com/NangoHQ/nango/tree/master/integration-templates/algolia"
```

bodinsamuel (approved):
Looks good, since it's just an action template the notes are just "FYI" but nothing to change imo

docs-v2/integrations/integration-templates/algolia.mdx
hassan254-prog: Thank you Samuel 🙏.This was just a simple sync script to test the integration.

PR #1830, opened 2024-03-08
[nan-374] support union string literal in nango.yaml

khaliqgant:
## Describe your changes
Usage:
nango.yaml
```
gender: m|f
```

NAN-374

### DIFF
--- a/packages/cli/lib/services/model.service.ts
+++ b/packages/cli/lib/services/model.service.ts
@@ -142,12 +142,12 @@ class ModelService {
 
             if (tsType.includes('|')) {
                 const types = tsType.split('|');
-                const isEnum = types.every((type) => !JAVASCRIPT_PRIMITIVES.includes(type.trim()));
+                const hasStringLiteral = types.some((type) => !JAVASCRIPT_PRIMITIVES.includes(type.trim()));
 
-                if (isEnum) {
+                if (hasStringLiteral) {
                     const enumValues = tsType
                         .split('|')
-                        .map((e) => `'${e.trim()}'`)
+                        .map((e) => (JAVASCRIPT_PRIMITIVES.includes(e.trim()) ? e.trim() : `'${e.trim()}'`))
                         .join(' | ');
                     tsType = enumValues;
                 }
--- a/packages/cli/lib/services/model.service.ts
+++ b/packages/cli/lib/services/model.service.ts
@@ -1,7 +1,7 @@
 import fs from 'fs';
 import yaml from 'js-yaml';
 import chalk from 'chalk';
-import type { NangoConfig, NangoModel, NangoIntegration, NangoIntegrationData } from '@nangohq/shared';
+import { JAVASCRIPT_PRIMITIVES, NangoConfig, NangoModel, NangoIntegration, NangoIntegrationData } from '@nangohq/shared';
 import { SyncConfigType, nangoConfigFile } from '@nangohq/shared';
 import { printDebug, getNangoRootPath } from '../utils.js';
 import { TYPES_FILE_NAME, NangoSyncTypesFileLocation } from '../constants.js';
@@ -139,6 +139,17 @@ class ModelService {
             if (hasUndefined) {
                 tsType = `${tsType} | undefined`;
             }
+
+            const isEnum = tsType.includes('|') && !JAVASCRIPT_PRIMITIVES.includes(tsType);
+
+            if (isEnum) {
+                const enumValues = tsType
+                    .split('|')
+                    .map((e) => `'${e.trim()}'`)
+                    .join(' | ');
+                tsType = enumValues;
+            }
+
             return tsType;
         } else {
             try {
--- a/packages/cli/lib/services/model.service.ts
+++ b/packages/cli/lib/services/model.service.ts
@@ -140,14 +140,17 @@ class ModelService {
                 tsType = `${tsType} | undefined`;
             }
 
-            const isEnum = tsType.includes('|') && !JAVASCRIPT_PRIMITIVES.includes(tsType);
+            if (tsType.includes('|')) {
+                const types = tsType.split('|');
+                const isEnum = types.every((type) => !JAVASCRIPT_PRIMITIVES.includes(type.trim()));
 
-            if (isEnum) {
-                const enumValues = tsType
-                    .split('|')
-                    .map((e) => `'${e.trim()}'`)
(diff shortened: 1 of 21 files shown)


packages/cli/lib/services/model.service.ts
```diff
@@ -139,6 +139,20 @@ class ModelService {
             if (hasUndefined) {
                 tsType = `${tsType} | undefined`;
             }
+
+            if (tsType.includes('|')) {
+                const types = tsType.split('|');
+                const isEnum = types.every((type) => !JAVASCRIPT_PRIMITIVES.includes(type.trim()));
```
TBonnin: I am gonna be a bit pedantic but `|` represents union type not enum

TBonnin: are you sure about the negation. Isn't it a union type if every element are primitives?

packages/cli/lib/services/model.service.ts
```diff
@@ -139,6 +139,20 @@ class ModelService {
             if (hasUndefined) {
                 tsType = `${tsType} | undefined`;
             }
+
+            if (tsType.includes('|')) {
```
TBonnin, line 143: Does supporting union types make the specific handling of undefined and null above unecessary?

packages/cli/lib/services/model.service.ts
```diff
@@ -139,6 +139,20 @@ class ModelService {
...
+                const isEnum = types.every((type) => !JAVASCRIPT_PRIMITIVES.includes(type.trim()));
+
+                if (isEnum) {
+                    const enumValues = tsType
+                        .split('|')
+                        .map((e) => `'${e.trim()}'`)
+                        .join(' | ');
+                    tsType = enumValues;
+                }
+            }
```
TBonnin, line 154: I think this logic could be generalized even if tsType doesn't include `|`. if not an union, after splitting we would operate on a single element list. It also has the benefit of handling stuff like boolean vs bool, ...
```
return tsTypes
   .split('|')
   .map((e) => {
      let res = e;
      switch (e.trim()) {
         case 'boolean':
         case 'bool':
            res = 'boolean';
       ....
      }
      return `'${res}'`
   })
   .join('|')
```

packages/cli/lib/services/model.service.ts
bodinsamuel: +1 on union. Also the yaml should be an array of string no?
```yaml
gender: 
  - female
  - male
```

khaliqgant: "Enum" was a misnomer here, but really shorthand for what was trying to be achieved here:

the nango yaml would be this:

```
MyModel:
   gender: male|female
```

generated ts file
```
  gender: 'male' | 'female'
```

So just a string literal union type. 

> Isn't it a union type if every element are primitives?

I'm not sure what you mean here? In this case we only want to characterize this as a string literal if it doesn't contain any primitives

> Also the yaml should be an array of string no?

When translated to typescript that wouldn't give us exactly what we want, but maybe I'm misunderstanding what you mean here?

packages/cli/lib/services/model.service.ts
khaliqgant: I think my earlier comment clarifies this -- but let me know if not!

packages/cli/lib/services/model.service.ts
khaliqgant: We're not explicitly handling union types other than the enum (string literal union type) here and I'd like to stay away from that in the context of this PR.

khaliqgant:
Rereading your comments @TBonnin and @bodinsamuel i realize that the PR wasn’t clear enough in its intent. 

1. Union types are already supported in the nango.yaml in that we just pass the type through directly to typescript. 
2. We can’t directly support enums because they would no longer be a pure type and we can’t use relative imports in the sync scripts. I called it an enum because that was the client called it and this was the closest we can get at the time. As mentioned in my comment though it is just a string literal union type that this PR adds support for

packages/cli/lib/services/model.service.ts
bodinsamuel: enums in json schema are array of possibles values, so I was thinking it could work if we do the same in yaml
https://json-schema.org/understanding-json-schema/reference/enum

avoid false positive and parsable natively

khaliqgant: Gotcha. I'm not sure how this would translate as we aren't directly using JSON schema to parse the yaml and then we still wouldn't be able to use a typescript enum regardless. See [comment](https://github.com/NangoHQ/nango/pull/1830#issuecomment-1989428859)

bodinsamuel:
I was thinking this, but maybe I'm missing something:

```patch
diff --git a/packages/cli/lib/services/model.service.ts b/packages/cli/lib/services/model.service.ts
index 5290a835..d3e1b612 100644
--- a/packages/cli/lib/services/model.service.ts
+++ b/packages/cli/lib/services/model.service.ts
@@ -141,6 +141,9 @@ class ModelService {
             }
             return tsType;
         } else {
+            if (Array.isArray(rawField)) {
+                return rawField.map((val) => JSON.stringify(val)).join(' | ');
+            }
             try {
                 const nestedFields = Object.keys(rawField)
                     .map((fieldName: string) => `  ${fieldName}: ${this.getFieldType(rawField[fieldName] as string | NangoModel, debug, modelName)};`)
```

## Input
```yaml
integrations:
  [...]
models:
  GithubIssueDemo:
    id: integer
    gender:
      - female
      - male
```

## Output
```ts
export interface GithubIssueDemo {
  id: number;
  gender: "female" | "male";
}
```

TBonnin:
I have indeed misunderstood the PR, sorry about that. I like @bodinsamuel solution tough, so we don't implement some special syntax for string literal union type

khaliqgant:
@bodinsamuel it seems consensus is if we can't do 

```
gender: 'male' | 'female'
```
which isn't valid yaml, that

```
gender: male | female
```

is preferred. To harden this behavior and parsing I added in tests and also support for
```
gender: male | female | null
```

which will become
```ts
gender: 'male' | 'female' | null;
```

in ts as expected.

Let me know thoughts and happy to discuss this further!

bodinsamuel (approved):
Let's move forward on this 👌🏻 
and thanks for the added tests ☺️

khaliqgant:
cc/ @bastienbeurier more docs to add 😅

PR #1832, opened 2024-03-08
feature(integrations): Add Xero and Netsuite to integration templates

tconlin:
## Describe your changes
Added simplified version of xero and netsuite integration work to templates, both actions and syncs

https://linear.app/nango/issue/NAN-539/add-xero-and-netsuite-to-integration-templates


TBonnin:
this is based on the work we did for n/a. tbh I would prefer to see those integrations stabilize (aka: used for real use cases) before we add them to the built-in list of integrations

PR #1834, opened 2024-03-11
fix(flows): add github issues demo example flow

bodinsamuel:
## Describe your changes

Contributes to NAN-522

For the getting started revamp (aka interactive demo) we want to fetch data from a known and controlled source. 

- Add new flow dedicated for the demo
I'm not entirely sure it's the way to go, but I feel like deployPreBuiltFlow is looking at this and since it's looking for S3 right now I need this on master first.


integration-templates/github/github-issues-demo-action.ts
```diff
@@ -0,0 +1,24 @@
...
+            body: `This Issue was created automatically using Nango's Action system.
+
+Action can be triggered with an HTTP call.
+Take a look at our [Documentation](https://docs.nango.dev/integrate/guides/perform-workflows-with-an-api)`,
+            labels: ['automatic']
+        }
+    });
+
+    return {
+        status: res.status === 201
```
TBonnin: why not returning the origin status instead of a boolean?

packages/shared/flows.yaml
```diff
@@ -404,6 +404,15 @@ integrations:
...
         scopes: repo
+      github-issues-demo:
+        runs: every 5 minutes
+        auto_start: false
+        sync_type: full
+        output: GithubIssueDemo
+        scopes: public_repo
+        description: |
+          Fetches GitHub issues from our showcase repository.
+        endpoint: /github/demo-issues
```
TBonnin, line 415: I know GET is the default method but I would recommend we always specify it explicitely.

packages/shared/flows.yaml
```diff
@@ -2189,10 +2206,10 @@ integrations:
         firstname: string
         lastname: string
         headline: string
-        account:
+        account: &ref_0
```
TBonnin, line 2209: where does this magic come from?

TBonnin (approved):

packages/shared/flows.yaml
bodinsamuel: I just ran the script, honestly unsure :D

khaliqgant: Looks like an underlying issue with the `__extends` logic when dealing with nested objects which should be fixed ASAP if this is happening

integration-templates/github/github-issues-demo-action.ts
bodinsamuel: yeah right

packages/shared/flows.yaml
bodinsamuel: fixed

packages/shared/flows.yaml
bodinsamuel: is this really an issue tho?

khaliqgant: Not blocking this PR, but it could negatively impact the API Reference display -- would have to check

PR #1839, opened 2024-03-11
[nan-529] fix bug in syncstatus endpoint

khaliqgant:
## Describe your changes
* Adds `track_deletes` to the nango.yaml schema
* Fixes issue with mismatch of the UI vs. the API when a sync is running
* Remove duplicated syncs when a connection id is passed to the status endpoint 

NAN-529

### DIFF
--- a/packages/server/lib/controllers/sync.controller.ts
+++ b/packages/server/lib/controllers/sync.controller.ts
@@ -549,7 +549,8 @@ class SyncController {
             const environmentId = getEnvironmentId(res);
 
             if (syncNames === '*') {
-                syncNames = await getSyncsByProviderConfigKey(environmentId, provider_config_key as string).then((syncs) => syncs.map((sync) => sync.name));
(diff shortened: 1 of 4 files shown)


packages/server/lib/controllers/sync.controller.ts
```diff
@@ -549,7 +549,8 @@ class SyncController {
             const environmentId = getEnvironmentId(res);
 
             if (syncNames === '*') {
-                syncNames = await getSyncsByProviderConfigKey(environmentId, provider_config_key as string).then((syncs) => syncs.map((sync) => sync.name));
+                const syncs = await getSyncsByProviderConfigKey(environmentId, provider_config_key as string);
+                syncNames = syncs.filter((sync) => (connection_id ? sync.connection_id === connection_id : true)).map((sync) => sync.name);
```
bodinsamuel: would be great to filter this in the query instead

bodinsamuel (approved):

PR #1840, opened 2024-03-12
chore: unified dockerfile

bodinsamuel:
> [!NOTE] 
> This is a proposal, open to debate.

## Context

Fixes NAN-596

We currently maintain 4 Dockerfiles, and need to build 7 in total. I feel like it's a lot to maintain. As we grow we will add more images and more folder in packages which can results in more stuff being build for nothing and errors.

I would love to reduce this maintenance to 3 images (`staging`, `prod`, `enterprise`) and down to 1 if possible. Right now we need many versions also because we pass dedicated env vars to the Webapp build which is something we could do differently (e.g: serving those via API calls)

## Proposal

- Build one unified Docker image `nangohq/nango` 

It's currently 148MB (compressed) [hub.docker.com](https://hub.docker.com/repository/docker/nangohq/nango/tags?page=1&ordering=last_updated)  vs `nangohq/nango` is 145MB. So the difference is already minimal enough that it doesn't really matter.

NB: the image is private rn

NB: I disabled build for `entreprise`  for the moment. I feel like it would be a waste of time and if we manage to get rid of the Env Vars situation, the single image could serve this scenario. 

## Migration

If we agree the migration path is not urgent. We just need to take the time to do it when we want. 
E.g for jobs:
- changing the image to `nangohq/jobs:production` to `nangohq/nango:prod-<hash>`
- adding a `Docker Command`, e.g: `node packages/jobs/dist/app.js`
- When we deploy, instead of triggering a new deploy on the same tag, we can trigger a deploy with the new image url 

NB: this should also solve our docker caching issue

### Should we keep the other images?

In the long-term I feel like it's not really useful, I think what other projects are doing is perfect with only 2-3 images:
- project/image_community
- project/image_entreprise
- project/image_private (for us)


bodinsamuel:
Talked with @bastienbeurier we agreed on this solution, as long as we communicate and document the change.
We will keep building all others images until the migration is done on our side then deprecate their usage officially.

bodinsamuel:
Can I have a review on this @khaliqgant @TBonnin? thanks ☺️

.github/workflows/build-image-reusable.yaml
```diff
@@ -0,0 +1,60 @@
...
+        inputs:
+            if:
+                description: 'Whether to run this job'
+                required: false
+                default: true
+                type: boolean
+            name:
+                required: true
+                type: string
+            key_secret_sentry:
```
khaliqgant: This reads oddly to me.

`sentry_secret_key`
`posthog_secret_key`

Dockerfile
```diff
@@ -0,0 +1,87 @@
+# ------------------
+# New tmp image
+# ------------------
+FROM node:18.19.1-bullseye-slim AS tmp
+
+# Setup the app WORKDIR
+WORKDIR /app/tmp
+
+# Copy and install dependencies separately from the app's code
+# To leverage Docker's cache when no dependency has change
```
khaliqgant:
```suggestion
# To leverage Docker's cache when no dependency has changed
```

khaliqgant:
Thanks Samuel. I think this makes sense, but the work here should be captured in a project within our ticketing system. I say that to make sure we encompass the impacted pieces of this change as a checklist of sorts.

- migration path
- how our customers affected (those who are self hosting and those enterprise)
- helm charts needed change: https://github.com/NangoHQ/nango-helm-charts

bodinsamuel:
Good point I created a Linear project https://linear.app/nango/project/unified-dockerfile-1a4273be8741

.github/workflows/build-image-reusable.yaml
bodinsamuel: it's because it's not the key itself but the key of `secrets` object. I'll rename to something more explicit

khaliqgant (approved):

Dockerfile
```diff
@@ -0,0 +1,87 @@
+# ------------------
+# New tmp image
+# ------------------
+FROM node:18.19.1-bullseye-slim AS tmp
```
TBonnin: nitpick: tmp being a overused term, what about calling the initial image `builder` or something like this

Dockerfile
```diff
@@ -0,0 +1,87 @@
...
+
+ARG image_env
+ARG git_hash
+
+ENV PORT=8080
+ENV NODE_ENV=production
+ENV IMAGE_ENV $image_env
+ENV GIT_HASH $git_hash
+
+EXPOSE 8080
```
TBonnin, line 87: do we currently use port 80. That would require an infra change if yes

scripts/build_docker.sh
```diff
@@ -0,0 +1,62 @@
...
+  --platform linux/amd64 \
+  --build-arg image_env=${ENV} \
+  --build-arg git_hash=${GIT_HASH} \
+  --build-arg posthog_key=${SENTRY_KEY} \
+  --build-arg sentry_key=${POSTHOG_KEY} \
+  --cache-from type=gha \
+  --cache-to type=gha,mode=max \
+  --file ../Dockerfile \
+  $tags \
+  ../
```
TBonnin, line 62: did you run https://www.shellcheck.net/ or something similar for this script?

TBonnin (approved):

Dockerfile
bodinsamuel: sure good point

scripts/build_docker.sh
bodinsamuel: I have it on my vscode yes, there are only suggestions about quotes I'll add them to be sure

Dockerfile
bodinsamuel: locally we use 300x 
on render we pass port in an env var, e.g:`NANGO_JOBS_PORT`
I don't think it will be much more complicated than changing the env var but my test on staging will tell us

PR #1841, opened 2024-03-12
[nan-544] allow optional params to be passed in and used

khaliqgant:
## Describe your changes
See the community reported issue: https://nango-community.slack.com/archives/C04ABR352H0/p1710132351607969

The user is triggering an action from a sync in an effort to abstract out logic. This used to work until optional parameters were not allowed to be passed into a getConnection call with the changes in https://github.com/NangoHQ/nango/pull/1764/files#diff-f52f0d61cd6802499f5a31aca89d3aec3df278a751065390c02b6bbdb24c0840L337-R341

This update allows optional params to be passed into a getConnection call. The ideal case is that the caching of a connection call is keyed by a `providerConfigKey` and `connectionId` combination but I feel that is an edge case that is out of scope of the immediate user reported issue and this update resolves their blocking issue.

NAN-544

### DIFF
--- a/packages/shared/lib/sdk/sync.ts
+++ b/packages/shared/lib/sdk/sync.ts
@@ -338,7 +338,9 @@ export class NangoAction {
         if (this.dryRun) {
             return this.nango.proxy(config);
         } else {
-            const connection = await this.getConnection();
+            const { connectionId, providerConfigKey } = config;
+            console.log(config);
+            const connection = await this.getConnection(providerConfigKey, connectionId);
             if (!connection) {
                 throw new Error(`Connection not found using the provider config key ${this.providerConfigKey} and connection id ${this.connectionId}`);
             }
@@ -418,10 +420,14 @@ export class NangoAction {
         return this.nango.getToken(this.providerConfigKey as string, this.connectionId as string);
     }
 
-    public async getConnection(): Promise<Connection> {
+    public async getConnection(optionalProviderConfigKey?: string, optionalConnectionId?: string): Promise<Connection> {
         this.exitSyncIfAborted();
+
+        const providerConfigKey = optionalProviderConfigKey || this.providerConfigKey;
+        const connectionId = optionalConnectionId || this.connectionId;
+
         if (!this.memoizedConnection || Date.now() - this.memoizedConnection.timestamp > MEMOIZED_CONNECTION_TTL) {
-            const connection = await this.nango.getConnection(this.providerConfigKey as string, this.connectionId as string);
+            const connection = await this.nango.getConnection(providerConfigKey as string, connectionId as string);
             this.memoizedConnection = { connection, timestamp: Date.now() };
             return connection;
         }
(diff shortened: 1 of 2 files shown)


packages/shared/lib/sdk/sync.ts
```diff
@@ -338,7 +338,9 @@ export class NangoAction {
         if (this.dryRun) {
             return this.nango.proxy(config);
         } else {
-            const connection = await this.getConnection();
+            const { connectionId, providerConfigKey } = config;
+            console.log(config);
```
bodinsamuel:
```suggestion
```

packages/shared/lib/sdk/sync.ts
```diff
@@ -418,10 +420,14 @@ export class NangoAction {
...
     }
 
-    public async getConnection(): Promise<Connection> {
+    public async getConnection(optionalProviderConfigKey?: string, optionalConnectionId?: string): Promise<Connection> {
         this.exitSyncIfAborted();
+
+        const providerConfigKey = optionalProviderConfigKey || this.providerConfigKey;
+        const connectionId = optionalConnectionId || this.connectionId;
+
         if (!this.memoizedConnection || Date.now() - this.memoizedConnection.timestamp > MEMOIZED_CONNECTION_TTL) {
```
bodinsamuel: are you sure it doesn't break memoizedConnection? 
passing a parameter does not change the fact that the connection can be memoized and the wrong one will be returned no?

packages/shared/lib/sdk/sync.ts
```diff
@@ -413,19 +434,27 @@ export class NangoAction {
...
+            !this.memoizedConnections.get(`${providerConfigKey}${connectionId}`) ||
+            Date.now() - this.memoizedConnections.get(`${providerConfigKey}${connectionId}`)!.timestamp > MEMOIZED_CONNECTION_TTL
+        ) {
+            const connection = await this.nango.getConnection(providerConfigKey as string, connectionId as string);
+            this.memoizedConnections.set(`${providerConfigKey}${connectionId}`, { connection, timestamp: Date.now() });
             return connection;
         }
-        return this.memoizedConnection.connection;
+
+        return this.memoizedConnections.get(`${providerConfigKey}${connectionId}`)!.connection;
```
bodinsamuel: Just a coding style nitpick, it's kinda hard to read, I would store this variable instead of computing it 4 times.

```ts
const couple = `${providerConfigKey}${connectionId}`;
const exists = this.memoizedConnections.get(couple);
if (!exists || Date.now() - exists.timestamp) {
  [...]
  set(couple, { connection });
}

return exists.connection;
```

bodinsamuel (approved):
👍🏻

packages/shared/lib/sdk/sync.ts
khaliqgant: Nice suggestion, updated with https://github.com/NangoHQ/nango/pull/1841/commits/f69f07cdf093d9df8ea8492e69d850dbdc17c6a0

Thanks.

PR #1844, opened 2024-03-13
fix(ui): use hooks for meta/env/user

bodinsamuel:
## Describe your changes

Fixes NAN-558
Contributes to NAN-516

We were spamming a bit /meta /env /user because we were using straight http call in the components. 
This is tentative to start refactoring those calls using SWR and hooks. I also need to add stuff to `/meta` for the new interactive demo and it was a good time to have something stable to do it later.

- Start exporting some types from the Server
We are duplicating types from server to webapp, which creates discrepancy. Start small by exposing `GetMeta` to see if the forms is fitting.

- Introduce `useEnvironment` `useMeta` `useUser` 
This should remove the biggest offender that fetch on every pages. This also reduce the boilerplate needed on some pages, but could be reduced further by removing some useless setState.


bodinsamuel:
~There is a weird things happening to the dom, let me check that before reopening~ nevermind it was my browser cache

packages/webapp/src/App.tsx
```diff
@@ -100,6 +100,11 @@ const App = () => {
         >
             <SWRConfig
                 value={{
+                    refreshInterval: 15 * 60000,
```
khaliqgant, line 103: Can this be overridden on the activity page to a shorter value? It has been quite nice to have refreshes on that page . Temporal does a 15 second refresh FWIW

image

khaliqgant:
Clicking through the nav and the activity page just doesn't load:


https://github.com/NangoHQ/nango/assets/1724137/d83e33b4-26c0-4947-9faa-15b993de4d03

packages/webapp/package.json
```diff
@@ -33,6 +33,7 @@
         "@heroicons/react": "2.0.18",
         "@mantine/prism": "5.10.5",
         "@nangohq/frontend": "^0.39.3",
+        "@nangohq/server": "file:../server",
```
khaliqgant, line 36: nice

packages/webapp/src/App.tsx
bodinsamuel: it's configurable at the call level yes. Activity page is not using SWR yet no?

bodinsamuel: nevermind it does, changing that 👍🏻

khaliqgant: Thanks. That page was the entire reason I added in SWR last minute FWIW

bodinsamuel:
> Clicking through the nav and the activity page just doesn't load:

Indeed, thanks for raising, there was an issue with the fetcher cache, it was not triggering `onSuccess` so I had to change the logic to something more hooky. Let me know

bodinsamuel:
@khaliqgant it's in staging

PR #1849, opened 2024-03-14
fix(demo): hide after complete

bodinsamuel:
## Describe your changes

Fixes NAN-516

- Hide demo after completion
This will only happen after refresh. I was going to animate it but definitely not worth the time at this stage.

- Change the way we handle 404 and redirection
Because I needed to wait for meta to be ready before making a redirection (I don't want to redirect to the demo if we already completed it) I needed to have a hook instead of a function. I split NotFound, Homepage, and handled the different scenario there. 

---

NB: I noticed this bug https://linear.app/nango/issue/NAN-568/fix-env-in-url-is-not-synced-to-cookie


khaliqgant (approved):
code looks good. question on behavior
1. What happens on the not found?
2. Is there a way to reset the demo to be able to do it again? Know Robin previously wanted a way to reset a demo

bodinsamuel:
>What happens on the not found?

I still redirect to either integrations or interactive demo, if that's the question.

> Is there a way to reset the demo to be able to do it again? Know Robin previously wanted a way to reset a demo

After full completion, the button goes to the user menu bottom left, and you can double click the `Demo` in the title of the page to reset the progress ☺️ It won't put it back to the menu though.

khaliqgant:
> > What happens on the not found?
> 
> I still redirect to either integrations or interactive demo, if that's the question.
> 
> > Is there a way to reset the demo to be able to do it again? Know Robin previously wanted a way to reset a demo
> 
> After full completion, the button goes to the user menu bottom left, and you can double click the `Demo` in the title of the page to reset the progress ☺️ It won't put it back to the menu though.

Got it, thanks

PR #1850, opened 2024-03-14
fix(jobs): do not wait for temporal

bodinsamuel:
## Describe your changes

After last eslint fix, I "broke" this call by adding the await but it's actually not returning


bodinsamuel:
Merging because nobody is there and I need to deploy :D

khaliqgant:
👍

PR #1855, opened 2024-03-14
fix(demo): feedback

bodinsamuel:
## Describe your changes

Fixes NAN-566

- Name of the action was not descriptive enough, changed to `github-create-demo-issue`
- Missing updated wording in Action (my bad @khaliqgant I got confused with my local setup and I was sure to have pushed it)
- Filter the sync records on the label to avoid pulling everything
- Unpause sync to run regular schedule
- Show records by default
- Use proper issue URL in sync


packages/shared/flows.yaml
```diff
@@ -1,2756 +1,2756 @@
 integrations:
```
bodinsamuel, line 1: the big change is because it has been formatted by prettier before

bodinsamuel:
The change in Action and Sync, are hardly testable we need to change the way we pull templates

khaliqgant (approved):

PR #1858, opened 2024-03-14
404 page

tconlin:
## Describe your changes
- Add 404 Page and handle routing from Connections page and Integrations page
- Add padding to dashboard content
- Updated empty state working on connections page

https://linear.app/nango/issue/NAN-570/design-404-page
https://linear.app/nango/issue/NAN-567/wording-on-empty-states


packages/webapp/src/pages/Connection/Show.tsx
```diff
@@ -60,8 +60,9 @@ export default function ShowIntegration() {
 
         const getConnections = async () => {
             const res = await getConnectionDetailsAPI(connectionId, providerConfigKey, false);
-
-            if (res?.status === 200) {
+            if (res?.status === 400) {
+                navigate('/404');
```
bodinsamuel: you don't want to redirect to a 404, you want to keep the URL so the user can know that the ressource they accessed is 404. 
To achieve this, you can remove the route and instead store a state to do something like this in the component
```tsx
if (is404) {
  return <NotFound />
}
```

packages/webapp/src/pages/PageNotFound.tsx
```diff
@@ -0,0 +1,21 @@
+import DefaultLayout from '../layout/DefaultLayout';
+import { Link } from 'react-router-dom';
+
+export default function PageNotFound() {
+    return (
+        <>
```
bodinsamuel: Unnecessary <>

bodinsamuel (commented):
Thanks for handling that, just a small comments on the implem

.env-example
bodinsamuel, line 1: Why did you rename this file?

packages/webapp/src/pages/Connection/Show.tsx
tconlin: Implemented this

.env-example
tconlin: mistake, fixed

packages/webapp/src/App.tsx
```diff
@@ -120,6 +121,7 @@ const App = () => {
                             <Route path="/reset-password/:token" element={<ResetPassword />} />
                         </>
                     )}
+                    <Route path="/404" element={<PageNotFound />} />
```
bodinsamuel:
```suggestion

```

bodinsamuel (approved):
🚀

bodinsamuel (approved):

bodinsamuel:
Sorry about the delay Taylor. Don't hesitate to ping on slack or ask the others too 👍🏻

PR #1865, opened 2024-03-18
feat: logs storage v2

bodinsamuel:
## Describe your changes

Fixes NAN-580

> [!NOTE]
> This is the implementation only, it's not used anywhere to limit PR size and possible issues at deploy

- Introduce ElasticSearch in the stack
Locally that means a single node in Docker, and also in tests. Integration tests are a bit slower to start because of that. 

- Create client, model and helpers
This is a "in theory" implementation. It should cover most use cases (and it's tested) but actual usage might require some fine tuning.

## ElasticSearch details

- One index with everything (Operation and Messages) 
Because we want a unified logs UI. It's a bit like in Datadog traces, everything is a span, the main Span just don't have parent. It's the same here, Operation don't have `parentId` otherwise they share the exact same properties.

- No partitioning
For the moment I'll go for one index, mostly because we don't have that much data, I don't want to over-engineer this feature. Also Operation and Messages can spread for many days and it might be too naive to split by days.
Retention can still easily be achieved by using `deleteByQuery` 

```ts
const ctx = await getOperationContext(operationPayload);
// == const activityLogId = await createActivityLog();

ctx.info('This is an info');
ctx.success();
// == createActivityLogMessageAndEnd
```


packages/logs/lib/vitest.d.ts
```diff
@@ -0,0 +1,14 @@
+export * from 'vitest';
```
bodinsamuel, line 1: For some reason I can't reuse it from shared :/

packages/cli/docker/docker-compose.yaml
```diff
@@ -103,5 +104,26 @@ services:
         networks:
             - nango
```
khaliqgant, line 105: unrelated to this PR but we should probably drop support for the ability to run nango from the cli.

packages/logs/lib/client.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * ------ Logs
+     */
+    async log(data: MessageRowInsert): Promise<void> {
+        await createMessage(getFormattedMessage({ ...data, parentId: this.id }));
+    }
+
+    async debug(message: string, meta: MessageMeta = null): Promise<void> {
+        await this.log({ type: 'log', level: 'debug', message, meta, source: 'nango' });
```
khaliqgant: source should be a variable. Would think maybe for enterprise customers this value would be different.

packages/logs/lib/env.ts
```diff
@@ -0,0 +1,3 @@
+import { parseEnvs, ENVS } from '@nangohq/utils/dist/environment/parse.js';
+
+export const envs = parseEnvs(ENVS.required({ NANGO_LOGS_ES_URL: true, NANGO_LOGS_ES_USER: true, NANGO_LOGS_ES_PWD: true }));
```
khaliqgant, line 3: nice

khaliqgant (commented):
[Dockerfile](https://github.com/NangoHQ/nango/blob/master/Dockerfile) should be edited? 

Assume more fine grained schema settings for string search (tokenizers etc) will be added later if needed?

packages/logs/lib/client.ts
bodinsamuel: source `nango` means internal system, better name is welcome, maybe `internal`?
log generated from sync/actions will be inserted with `log()`

bodinsamuel:
>[Dockerfile](https://github.com/NangoHQ/nango/blob/master/Dockerfile) should be edited?
> Assume more fine grained schema settings for string search (tokenizers etc) will be added later if needed?

Good point, I added both. The analyser is rather simple for the moment [standard](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-standard-analyzer.html). but should be enough to get started to achieve full-text search

packages/cli/docker/docker-compose.yaml
bodinsamuel: I'm not even sure what is it for?

packages/logs/lib/client.ts
TBonnin: I like `internal` better than `nango`

packages/logs/lib/models/messages.ts
```diff
@@ -0,0 +1,129 @@
...
+ * List operations
+ */
+export async function listOperations(opts: { limit: number }): Promise<ListOperations> {
+    const res = await client.search<MessageRow>({
+        index: indexMessages.index,
+        size: opts.limit,
+        sort: [{ createdAt: 'desc' }, '_score'],
+        track_total_hits: true,
+        query: {
+            // @ts-expect-error I don't get the error
```
TBonnin, line 36: what's the error?

packages/logs/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "name": "@nangohq/logs",
+    "version": "1.0.0",
+    "type": "module",
+    "main": "./dist/index.js",
+    "types": "./dist/index.js",
+    "scripts": {},
+    "repository": {
+        "type": "git",
+        "url": "git+https://github.com/NangoHQ/nango.git",
+        "directory": "packages/jobs"
```
TBonnin:
```suggestion
        "directory": "packages/logs"
```

TBonnin (approved):

packages/logs/lib/client.ts
bodinsamuel: okay perfect then

packages/logs/package.json
bodinsamuel: good catch thanks

packages/logs/lib/models/messages.ts
bodinsamuel: The query is valid and it works but the typings is not happy, could not find any way to make it pass

bodinsamuel:
Do you have a final comment on this PR @khaliqgant? ☺️

khaliqgant (approved):

PR #1867, opened 2024-03-18
fix: prevent race condition when updating job results

TBonnin:
locking the job row to prevent concurrent call to UpdateSyncJobResult to overwrite each other

https://linear.app/nango/issue/NAN-577/persist-handle-concurrent-access


packages/shared/lib/services/sync/job.service.ts
```diff
@@ -89,47 +89,49 @@ export const updateLatestJobSyncStatus = async (sync_id: string, status: SyncSta
...
-        const incomingResult = result[model];
-        const finalResult = {
-            ...existingResult,
-            [model]: {
-                added: Number(added) + Number(incomingResult?.added),
-                updated: Number(updated) + Number(incomingResult?.updated)
-            }
-        };
+    return db.knex.transaction(async (trx) => {
+        const { result: existingResult } = await schema().from<SyncJob>(SYNC_JOB_TABLE).select('result').where({ id }).first().forUpdate();
```
TBonnin: using transaction and `select for update` to lock the row.

bodinsamuel (commented):
Looks good 👌🏻 
You have a failing test, I'm not sure it's related but let me know ☺️

TBonnin:
> Looks good 👌🏻 You have a failing test, I'm not sure it's related but let me know ☺️

Not sure what's going on with the test. The `select` times out now it is in a transaction

TBonnin:
> > Looks good 👌🏻 You have a failing test, I'm not sure it's related but let me know ☺️
> 
> Not sure what's going on with the test. The `select` times out now it is in a transaction.

My bad. I hadn't followed the _'DO NOT mock the database'_ rule and I got bitten. Removed the db mocking in [2018920](https://github.com/NangoHQ/nango/pull/1867/commits/201892077a459d67de4f25f1122812cfbdbdf630)

packages/persist/lib/server.integration.test.ts
```diff
@@ -0,0 +1,207 @@
...
+    Sync,
+    Job as SyncJob,
+    SyncType,
+    SyncStatus,
+    Environment,
+    db
+} from '@nangohq/shared';
+
+describe('Persist API', () => {
+    const port = 3096;
```
bodinsamuel, line 22: might be cool to use https://www.npmjs.com/package/get-port

packages/persist/lib/server.integration.test.ts
```diff
@@ -0,0 +1,207 @@
...
+                        softDelete: true
+                    }),
+                    headers: {
+                        'Content-Type': 'application/json'
+                    }
+                }
+            );
+            expect(response.status).toEqual(400);
+            const respBody = (await response.json()) as any[];
+            expect(respBody[0]['errors']['issues'][0]['path'][0]).toEqual('records');
```
bodinsamuel: not related to this pr: Could be interesting to test the full body with toMatchSnapshot or toMatchObject since we don't have a lot of tests regarding this

packages/persist/lib/server.integration.test.ts
```diff
@@ -0,0 +1,207 @@
...
+                    activityLogId: seed.activityLogId,
+                    lastSyncDate: new Date(),
+                    trackDeletes: false
+                }),
+                headers: {
+                    'Content-Type': 'application/json'
+                }
+            }
+        );
+        expect(response.status).toEqual(201);
```
bodinsamuel, line 153: not related to this PR: this test do not actually check if the records were deleted

packages/persist/lib/server.integration.test.ts
```diff
@@ -0,0 +1,207 @@
...
+                    lastSyncDate: new Date(),
+                    trackDeletes: false,
+                    softDelete: true
+                }),
+                headers: {
+                    'Content-Type': 'application/json'
+                }
+            }
+        );
+        expect(response.status).toEqual(201);
```
bodinsamuel, line 181: same comment

packages/persist/lib/server.integration.test.ts
```diff
@@ -0,0 +1,207 @@
...
+    });
+    if (!activityLogId) throw new Error('Activity log not created');
+    const connectionRes = await connectionService.upsertConnection(`conn-test`, `provider-test`, 'google', {} as AuthCredentials, {}, env.id, 0);
+    const connectionId = connectionRes[0]?.id;
+    if (!connectionId) throw new Error('Connection not created');
+
+    const connection = (await connectionService.getConnectionById(connectionId)) as Connection;
+    if (!connection) throw new Error('Connection not found');
+
+    const sync = await createSync(connectionId, 'sync-test');
```
bodinsamuel, line 206: not related to this pr: the fact that we try/catch in models or don't throw makes those boilerplate so unnecessary huge 😩

packages/persist/lib/server.integration.test.ts
```diff
@@ -0,0 +1,207 @@
...
+        env,
+        activityLogId,
+        connection,
+        sync,
+        syncJob
+    };
+};
+
+const clearDb = async () => {
+    await db.knex.raw(`DROP SCHEMA nango CASCADE`);
```
bodinsamuel, line 222: we should invest in tests helpers to avoid repeating those stuff

packages/shared/lib/services/sync/job.service.ts
```diff
@@ -89,47 +89,49 @@ export const updateLatestJobSyncStatus = async (sync_id: string, status: SyncSta
...
-        const incomingResult = result[model];
-        const finalResult = {
-            ...existingResult,
-            [model]: {
-                added: Number(added) + Number(incomingResult?.added),
-                updated: Number(updated) + Number(incomingResult?.updated)
-            }
-        };
+    return db.knex.transaction(async (trx) => {
+        const { result: existingResult } = await trx.from<SyncJob>(SYNC_JOB_TABLE).select('result').forUpdate().where({ id }).first();
```
bodinsamuel, line 93: love that the fix is so small <3

bodinsamuel (approved):
All good ✅ 
Took the opportunity to express some concerns but not related to this PR ☺️

khaliqgant (approved):

TBonnin:
thank you @bodinsamuel. All your points are valid (I have updated the error matching test) and I will improve the tests as part of the track_delete and records migration

PR #1868, opened 2024-03-18
fix(deps): upgrade knex

bodinsamuel:
## Describe your changes

Fixes NAN-594

- Upgrade and pin knex


packages/jobs/package.json
```diff
@@ -15,6 +15,7 @@
         "directory": "packages/jobs"
     },
     "dependencies": {
+        "@nangohq/nango-jobs": "file:../jobs",
```
TBonnin: depends on itself? 😛

packages/jobs/package.json
```diff
@@ -29,7 +30,7 @@
         "dotenv": "^16.0.3",
         "fs-extra": "^11.1.1",
         "js-yaml": "^4.1.0",
-        "knex": "^2.4.2",
+        "knex": "3.1.0",
```
TBonnin, line 32: it is a major upgrade. Are we fully confident upgrading doesn't break anythin?

packages/jobs/package.json
bodinsamuel: ah jesus it was a test, thanks

packages/jobs/package.json
bodinsamuel: the main breaking change is the drop of Node16 and some small types fixes 
https://github.com/knex/knex/releases/tag/3.0.0

But I would love to test on staging before merging

bodinsamuel:
Looks good on staging 👌🏻

TBonnin (approved):

PR #1871, opened 2024-03-18
fix(eslint): pass #6

bodinsamuel:
## Describe your changes

- Frontend eslint was not configured correctly
- Manually fix some Eslint errors


packages/server/lib/clients/oauth1.client.ts
```diff
@@ -76,44 +72,37 @@ export class OAuth1Client {
...
-                        resolve(parsedFull);
+                function (error, data, _response) {
+                    if (error) {
+                        reject(error);
+                        return;
                     }
+
+                    const queryParams = new URLSearchParams(typeof data === 'string' ? data : data?.toString());
+
+                    resolve(Object.fromEntries(queryParams));
```
bodinsamuel: Can you double check if I fixed correctly? ☺️

packages/server/lib/utils/web-socket-error.ts
```diff
@@ -3,116 +3,114 @@ export interface WSErr {
     message: string;
 }
 
-export class WSErrBuilder {
```
bodinsamuel, line 6: Everything could be replaced by actual `class [..] extend Error`

packages/server/lib/clients/auth.client.ts
```diff
@@ -19,89 +19,84 @@
...
+        session({
+            secret: process.env['NANGO_ADMIN_KEY'] || 'nango',
+            resave: false,
+            saveUninitialized: false,
+            store: sessionStore,
+            name: 'nango_session',
+            unset: 'destroy',
+            cookie: { maxAge: 30 * 24 * 60 * 60 * 1000, secure: false },
+            rolling: true
+        })
```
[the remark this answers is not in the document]
bodinsamuel: any reason we have set `secure: false`?

packages/cli/lib/services/compile.service.ts
```diff
@@ -73,7 +74,7 @@ class CompileService {
                     continue;
                 }
                 const result = compiler.compile(fs.readFileSync(filePath, 'utf8'), filePath);
-                const jsFilePath = filePath.replace(/\/[^\/]*$/, `/dist/${path.basename(filePath.replace('.ts', '.js'))}`);
+                const jsFilePath = filePath.replace(/\/[^/]*$/, `/dist/${path.basename(filePath.replace('.ts', '.js'))}`);
```
TBonnin, line 77: how was it working before? or maybe it wasn't?

packages/persist/lib/server.ts
```diff
@@ -39,7 +39,7 @@ server.post(
             msg: z.string()
         })
     }),
-    persistController.saveActivityLog
+    persistController.saveActivityLog.bind(persistController)
```
TBonnin, line 42: why is bind necessary here?

bodinsamuel: it's not necessary to make it work but because it's in a class `this` reference can be mistakenly overrode.
overall our usage of class is probably unnecessary

packages/cli/lib/services/compile.service.ts
bodinsamuel: I think it's just ignored by the regex compiler hence why it's unnecessary, cf: https://eslint.org/docs/latest/rules/no-useless-escape

packages/persist/lib/server.ts
TBonnin: what would you recommend instead of using a class?

packages/server/lib/clients/oauth1.client.ts
TBonnin: can we add tests to validate it with code?

packages/persist/lib/server.ts
bodinsamuel: straight up function, if we had state it would be debatable but it's not the case and I would argue keeping state between api calls is dangerous

packages/server/lib/clients/oauth1.client.ts
bodinsamuel: I can extract this logic yes

packages/persist/lib/server.ts
TBonnin: agreed. In this specific case the class in only used as a namespace mechanism

packages/server/lib/clients/oauth1.client.ts
```diff
@@ -137,6 +124,10 @@ export class OAuth1Client {
...
         const url = new URL(this.authConfig.authorization_url);
         const params = new URLSearchParams(queryParams);
-        return `${url}?${params.toString()}`;
+        return `${url.href}?${params.toString()}`;
     }
 }
+
+export function extractQueryParams(data: string | Buffer | undefined) {
+    return new URLSearchParams(typeof data === 'string' ? data : data?.toString());
+}
```
TBonnin, line 133: why not add `Object.fromEntries` as part of the logic here?

bodinsamuel: good point

TBonnin (approved):

PR #1872, opened 2024-03-18
[NAN-578] Implement Google Social Login/Signup

khaliqgant:
## Describe your changes
Adds in google auth using workos

image

NAN-578


packages/server/lib/controllers/auth.controller.ts
```diff
@@ -260,6 +272,99 @@ class AuthController {
             next(error);
         }
     }
+
+    async getHostedLogin(req: Request, res: Response, next: NextFunction) {
+        try {
+            const provider = req.query['provider'] as string;
```
bodinsamuel, line 322: would be great to have some validation/allowlist

packages/server/lib/controllers/auth.controller.ts
```diff
@@ -260,6 +272,99 @@ class AuthController {
...
+
+            if (!workos) {
+                errorManager.errRes(res, 'workos_not_configured');
+                return;
+            }
+
+            const body = req.body;
```
bodinsamuel: same comment

packages/server/lib/controllers/auth.controller.ts
```diff
@@ -260,6 +272,99 @@ class AuthController {
...
+                state: body ? Buffer.from(JSON.stringify(body)).toString('base64') : ''
+            });
+
+            res.send({ url: oAuthUrl });
+        } catch (err) {
+            next(err);
+        }
+    }
+
+    async createAccountIfNotInvited(state: string, name: string): Promise<number | null> {
```
bodinsamuel: this is not a controller method I think

packages/server/lib/controllers/auth.controller.ts
```diff
@@ -260,6 +272,99 @@ class AuthController {
...
+        const account = await environmentService.createAccount(`${name}'s Organization`);
+        if (!account) {
+            throw new NangoError('account_creation_failure');
+        }
+        return account.id;
+    }
+
+    async loginCallback(req: Request, res: Response, next: NextFunction) {
+        try {
+            const { code, state } = req.query;
```
bodinsamuel, line 396: same

packages/server/lib/controllers/auth.controller.ts
```diff
@@ -260,6 +272,99 @@ class AuthController {
...
+                return;
+            }
+
+            const { user: authorizedUser } = await workos.userManagement.authenticateWithCode({
+                clientId: process.env['WORKOS_CLIENT_ID'] || '',
+                code: code as string
+            });
+
+            let user: User | null = null;
+            const existingUser = await userService.getUserByEmail(authorizedUser.email);
```
bodinsamuel, line 413: what about suspended user?

packages/webapp/src/pages/InviteSignup.tsx
```diff
@@ -82,84 +80,81 @@ export default function InviteSignup() {
...
+                            <div className="border-t border-gray-600 flex-grow mr-7"></div>
+                            <span className="text-dark-500">or continue with</span>
+                            <div className="border-t border-gray-600 flex-grow ml-7"></div>
                         </div>
+                        <GoogleButton
+                            text="Sign up with Google"
+                            invitedAccountID={invitedAccountID}
+                            token={token}
+                            setServerErrorMessage={setServerErrorMessage}
+                        />
```
bodinsamuel: When I click this button it's trigger the form and show an error in the email

packages/webapp/src/components/ui/button/Auth/Google.tsx
```diff
@@ -0,0 +1,82 @@
+interface Props {
```
bodinsamuel, line 1: I was the impression that `/ui` folder was for Generic reusable component, this one is really a one of. (my pov is that everything is related to ui anyway so it's either an unnecessary hierarchy level or it's specific for one scenario)

packages/server/lib/controllers/auth.controller.ts
```diff
@@ -26,6 +28,16 @@ export interface WebUser {
     name: string;
 }
 
+interface InviteAccountState {
+    accountId: number;
+    token: string;
+}
+
+let workos: WorkOS | null = null;
+if (process.env['WORKOS_API_KEY']) {
```
bodinsamuel: missing in `.env.example`

packages/server/lib/controllers/auth.controller.ts
```diff
@@ -260,6 +272,99 @@ class AuthController {
...
+
+            if (!workos) {
+                errorManager.errRes(res, 'workos_not_configured');
+                return;
+            }
+
+            const body = req.body;
+
+            const oAuthUrl = workos?.userManagement.getAuthorizationUrl({
+                clientId: process.env['WORKOS_CLIENT_ID'] || '',
```
bodinsamuel, line 335: missing in `.env.example`

packages/webapp/src/pages/ResetPassword.tsx
```diff
@@ -39,23 +39,17 @@ export default function Signin() {
...
-                                        </label>
-                                    </div>
-                                </div>
                                 <div className="mt-1">
                                     <input
                                         id="password"
                                         name="password"
                                         type="password"
+                                        placeholder="Password"
                                         autoComplete="current-password"
```
bodinsamuel: unrelated, this field does not work with password manager, I think it needs `new-password`

packages/webapp/src/pages/InviteSignup.tsx
khaliqgant: I'm not able to reproduce this / not fully understanding what you mean. Can you clarify or maybe provide a quick screen cast?

packages/server/lib/controllers/auth.controller.ts
khaliqgant: I'm not sure what scenario you mean here? A user is suspended from a particular organization so they should be able to still sign in under their existing account if they have one -- or make a new one if they don't.

packages/server/lib/controllers/auth.controller.ts
khaliqgant: Updated

khaliqgant:
@bodinsamuel Thanks for the review. I updated the PR based on some of your feedback. The other items I didn't see as blocking but let me know if you feel strongly about any items I didn't address and I can add more color on why I didn't adjust it based on your feedback.

packages/server/lib/controllers/auth.controller.ts
TBonnin: I would have the server throw and not start if workOS (or anything else actually) is not configured properly. Instead of checking if workOS is not null at request time

packages/server/lib/controllers/auth.controller.ts
```diff
@@ -26,6 +28,36 @@ export interface WebUser {
...
+    token: string;
+}
+
+let workos: WorkOS | null = null;
+if (process.env['WORKOS_API_KEY']) {
+    workos = new WorkOS(process.env['WORKOS_API_KEY']);
+}
+
+const createAccountIfNotInvited = async (state: string, name: string): Promise<number | null> => {
+    const parsedAccount = (state ? JSON.parse(Buffer.from(state, 'base64').toString('ascii')) : null) as InviteAccountState | null;
```
TBonnin: Cannot JSON.parse throw an exception? What should happen in this case?

packages/server/lib/controllers/auth.controller.ts
```diff
@@ -26,6 +28,36 @@ export interface WebUser {
...
+
+    if (accountId !== null) {
+        const token = parsedAccount?.token;
+        const validToken = await userService.getInvitedUserByToken(token as string);
+        if (validToken) {
+            await userService.markAcceptedInvite(token as string);
+        }
+
+        return accountId;
+    }
```
TBonnin, line 80: can this be?
```
    const res: Result<InviteAccountState> = parseState(state) // returns a valid InviteAccountState or a ResultErr
    if (isOk(res)) {
        const validToken = await userService.getInvitedUserByToken(token);
        if (validToken) {
            await userService.markAcceptedInvite(parsedAccount.token);
        }
        return parsedAccount.accountId ;
    }
```

packages/server/lib/controllers/auth.controller.ts
```diff
@@ -260,6 +292,79 @@ class AuthController {
...
     }
+
+    getHostedLogin(req: Request, res: Response, next: NextFunction) {
+        try {
+            const provider = req.query['provider'] as string;
+
+            if (!workos) {
+                errorManager.errRes(res, 'workos_not_configured');
+                return;
+            }
```
TBonnin, line 372: see my comment above. This is not a concern for the user and I think it would be better to crash the server at startup

packages/server/lib/controllers/auth.controller.ts
TBonnin: should server not start if this is missing?

packages/server/lib/controllers/auth.controller.ts
```diff
@@ -260,6 +292,79 @@ class AuthController {
...
+
+            if (existingUser) {
+                user = existingUser;
+            } else {
+                const name =
+                    authorizedUser.firstName || authorizedUser.lastName
+                        ? `${authorizedUser.firstName || ''} ${authorizedUser.lastName || ''}`
+                        : authorizedUser.email.split('@')[0];
+
+                const accountId = await createAccountIfNotInvited(state as string, name as string);
```
TBonnin: should state and name be validated prior to make sure they are acceptable string. That would also avoid having to cast them

packages/server/lib/controllers/auth.controller.ts
```diff
@@ -260,6 +292,79 @@ class AuthController {
             next(error);
         }
     }
+
+    getHostedLogin(req: Request, res: Response, next: NextFunction) {
```
TBonnin: not sure what `Hosted` referred to here?

khaliqgant: Hosted = WorkOs

packages/server/lib/controllers/auth.controller.ts
khaliqgant: WorkOS is only required for Cloud so don't think it should crash here.

packages/server/lib/controllers/auth.controller.ts
khaliqgant: sure, updated

packages/server/lib/controllers/auth.controller.ts
khaliqgant: Added

packages/webapp/src/components/ui/button/Auth/Google.tsx
```diff
@@ -24,11 +24,13 @@ export default function GoogleButton({ text, setServerErrorMessage, invitedAccou
 
         if (invitedAccountID && token) {
             postBody.body = JSON.stringify({
-                accountId: invitedAccountID,
-                token
+                accountId: invitedAccountID
             });
         }
-        const res = await fetch('/api/v1/hosted/signup?provider=GoogleOAuth', postBody);
+        console.log(invitedAccountID, token);
```
bodinsamuel:
```suggestion

```

packages/server/lib/controllers/auth.controller.ts
bodinsamuel: I find our use of Hosted rather confusing, everything is hosted by default.
Is it enterprise, cloud or community? if `hosted = enterprise + cloud` maybe we can come up with a naming more representative like "managed". 
Plus we use `hosted` for self-hosted in the Docker images right now, which is the opposite

packages/server/lib/controllers/auth.controller.ts
bodinsamuel: I add the impression that user could only be part of one org, so if they only have one, they get suspended and they login what are they seeing?

packages/server/lib/server.ts
```diff
@@ -152,6 +154,12 @@ if (AUTH_ENABLED) {
     app.route('/api/v1/reset-password').put(rateLimiterMiddleware, authController.resetPassword.bind(authController));
 }
 
+if (HOSTED_AUTH_ENABLED) {
```
bodinsamuel: same comment about naming

bodinsamuel (commented):
Minor comments 👌🏻

khaliqgant:
> I add the impression that user could only be part of one org, so if they only have one, they get suspended and they login what are they seeing?

If a user is suspended they're essentially locked out of Nango. Handling this seems out of scope of this as it is a known limitation of the invite flow. Let me know if you disagree

packages/server/lib/controllers/auth.controller.ts
TBonnin: what I mean is this is a configuration issue that would better detected as soon as possible (aka: when the app starts) rather than deferred to whenever a request is being received. 
For sure for non-cloud setup this configuration can be ignored.

I prefer never to deployed a misconfigured app than getting a bug report by an end-user

khaliqgant: Yup, I gotcha, but who are we protecting here? Developers running the application locally? Cloud will have the necessary env variables.

TBonnin: > Yup, I gotcha, but who are we protecting here? 

We protecting our future self

> Cloud will have the necessary env variables.

They will, until one day the infra changes and configuration is somehow forgotten or messed up.
In general trying to detect error as soon as possible is a good principle to follow.

Not a blocker though.

packages/webapp/src/pages/InviteSignup.tsx
bodinsamuel: https://github.com/NangoHQ/nango/assets/1637651/cecd46e2-d24f-43cd-9d81-a2baba79e2f4

khaliqgant: Ah yes, thanks. This should be resolved with a push I made today. Can you confirm that this doesn't happen?

bodinsamuel: deploying the latest to staging and it works, thanks 👌🏻

bodinsamuel:
Tested one last time on staging, all good 👍🏻

PR #1873, opened 2024-03-18
feat(proxy): add fallback to proxy base_url

hassan254-prog:
## Describe your changes
Continuing from this, https://github.com/NangoHQ/nango/pull/1770#discussion_r1526495126. We need to have a fallback incase `api_base_url_for_customer` field is not populated in the connection configuration. I have also added a unit test for this..


packages/shared/lib/services/proxy.service.unit.test.ts
```diff
@@ -420,6 +420,46 @@ describe('Proxy service Construct URL Tests', () => {
         expect(url).toBe('https://myinstanceurl.com/api/test');
     });
 
+    it('Should handle Proxy base URL interpolation where connection configuration param is present', () => {
+        const config = {
+            template: {
+                auth_mode: AuthModes.OAuth2,
+                proxy: {
+                    base_url: '${connectionConfig.api_base_url_for_customer} || https://api.gong.io'
```
bodinsamuel, line 433: I don't think I have the full context but I would not go this way and add a new attributes to avoid this pseudo-js syntax. I'll let the others weigh on this

TBonnin:
how many customers use the gong integration in prod. If number is small, can we coordinate with them to update their connection instead of making the providers definition and code more complicated?

hassan254-prog:
Since we have no one using Gong in production, we can successfully close this follow-up PR.

khaliqgant (approved):

PR #1874, opened 2024-03-19
fix(eslint): pass #7

bodinsamuel:
## Describe your changes

- Add `eslint-plugin-import` to catch unspecified extension, missing packages and others stuff. 
This has caught a few issues here in there already.

- Enable `@typescript-eslint/consistent-type-imports`, it's important to let Typescript removes types from the imports/exports. Not mixing make them explicit and can be automatically hidden from IDE
Note: that I haven't fixed all errors to avoid a too big PR but it's will be autofixed with our subsequent PRs, nothing to do.


packages/webapp/src/utils/user.tsx
```diff
@@ -1,6 +1,6 @@
 import storage, { LocalStorageKeys } from '../utils/local-storage';
 import { useLogoutAPI } from '../utils/api';
-import { useNavigate } from 'react-router';
+import { useNavigate } from 'react-router-dom';
```
khaliqgant, line 3: This is done in quite a few other places, importing from react-router vs. react-router-dom

Probably a search and replace would be a quick win for this
```
import { useNavigate } from 'react-router';
```

khaliqgant (approved):

packages/webapp/src/utils/user.tsx
bodinsamuel: I have removed all of them 👍🏻 ☺️

khaliqgant: Nice!

PR #1876, opened 2024-03-19
[nan-600] pass in a generic

khaliqgant:
## Describe your changes
Allows a user to pass a generic to the ActionError, like so:

```
import type { NangoAction } from './models';

interface CustomError {
    code: number;
    key1: string;
}

export default async function runAction(nango: NangoAction): Promise<undefined> {
    await nango.get({
        baseUrlOverride: 'http://localhost:3009',
        endpoint: '/'
    });

    const { ActionError } = nango;

    throw new ActionError<CustomError>({
        code: 12345,
        key1: 'value1'
    });
}

```

NAN-600

### DIFF
--- a/packages/shared/lib/sdk/sync.ts
+++ b/packages/shared/lib/sdk/sync.ts
@@ -204,11 +204,11 @@ interface Connection {
     credentials: AuthCredentials;
 }
 
-export class ActionError extends Error {
+export class ActionError<T = Record<string, unknown>> extends Error {
     type: string;
     payload?: Record<string, unknown>;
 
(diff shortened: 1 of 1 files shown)


bodinsamuel (approved):
Not sure what it solves?
They could also create a custom error with `Custom extends ActionError`

khaliqgant:
> Not sure what it solves? They could also create a custom error with `Custom extends ActionError`

Not much, just allows for added specificity and they could bring that type into their nango-integrations from somewhere else. I think the user path is easier than extending ActionError.

PR #1884, opened 2024-03-20
fix(db): default migration folder

bodinsamuel:
## Describe your changes

When testing the new Docker image I realised we have hardcoded the migration folder in the docker for some reason instead of having a working default.

- Fix the default


bodinsamuel:
changing some additional stuff

bodinsamuel:
Too much has changed to reup this

PR #1885, opened 2024-03-20
[nan-616] if no connection, still forward

khaliqgant:
## Describe your changes
If no connection still forward webhook

NAN-616

### DIFF
--- a/packages/shared/lib/services/notification/webhook.service.ts
+++ b/packages/shared/lib/services/notification/webhook.service.ts
@@ -1,4 +1,5 @@
-import axios, { AxiosError } from 'axios';
+import type { AxiosError } from 'axios';
+import axios from 'axios';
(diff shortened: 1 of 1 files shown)


packages/shared/lib/services/notification/webhook.service.ts
```diff
@@ -291,6 +294,10 @@ class WebhookService {
             return;
         }
 
+        if (!connectionIds || connectionIds.length === 0) {
+            await this.forwardHandler(environment_id, providerConfigKey, '', provider, payload, webhookOriginalHeaders);
```
bodinsamuel, line 298: To avoid mistakes later

```suggestion
            await this.forwardHandler(environment_id, providerConfigKey, '', provider, payload, webhookOriginalHeaders);
            return;
```

PR #1887, opened 2024-03-20
fix(shared): proper ts rootDir

bodinsamuel:
## Describe your changes

Fixes NAN-617

- rootDir was not pointing to lib
It currently output everything to dist/lib which is inconvenient but also makes the paths unreliable since it changes from ts to js

- Fix package.json loading (and cache it)
This package.json is loaded on every call to `/meta`. 
And it was spamming errors in tests


khaliqgant (approved):
Thanks for digging on this. Would just want to test this on staging including a full cli deploy etc just to make sure nothing unexpected happens.

TBonnin (approved):
Thank you so much. I have been wanted to fix this for a very long time

bodinsamuel:
Looks good on staging ✅

PR #1889, opened 2024-03-20
fix: integration should be unique per environment

TBonnin:
The unique constraint for integration is for unique_key, environment_id and deleted_at which can be null. However null values are not considered equal by default by postgres. 🤯 

from postgres doc at https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-UNIQUE-CONSTRAINTS
> By default, two null values are not considered equal in this
comparison. That means even in the presence of a unique constraint it is possible to store duplicate rows that contain a null value in at least one of the constrained columns. This behavior can be changed by adding the clause NULLS NOT DISTINCT

With this constraint, editing a integration with a name that already exists shows an error

https://linear.app/nango/issue/NAN-615/user-managed-to-create-two-integrations-with-the-same-integration-id


khaliqgant (approved):
nice

bodinsamuel (approved):
There is the same issue in Connections I think

TBonnin:
> There is the same issue in Connections I think

good point. modifying the constraint for connections table as well

PR #1891, opened 2024-03-21
fix(cli): glob change of behavior

bodinsamuel:
## Describe your changes

Followup of #1874 

When I did this eslint pass, I found undeclared usage of packages. It was the case for glob, so I had to pick a version. After testing, I picked the last one. 
While `nango cli dev` worked,  it turns out it had a slightly different logic than `nango deploy dev` which hide a change in behavior.

- Share a single unified way to listFiles
- Add tests

## How to test?

- npm run dw
- `node ../nango/packages/cli/dist/index.js dev`
- modify a sync
- `node ../nango/packages/cli/dist/index.js deploy dev`
- check that your file are still there and compiled

### DIFF
--- a/packages/cli/lib/cli.ts
+++ b/packages/cli/lib/cli.ts
@@ -299,7 +299,7 @@ export const tscWatch = async (debug = false) => {
         if (filePath === nangoConfigFile) {
             return;
         }
-        compileFile(listFile(`./${filePath}`));
+        compileFile(listFile(filePath));
     });
 
     watcher.on('unlink', (filePath: string) => {
--- a/packages/cli/lib/cli.ts
+++ b/packages/cli/lib/cli.ts
@@ -5,7 +5,6 @@ import yaml from 'js-yaml';
 import chalk from 'chalk';
 import chokidar from 'chokidar';
 import * as tsNode from 'ts-node';
-import { glob } from 'glob';
 import ejs from 'ejs';
 import * as dotenv from 'dotenv';
 import { spawn } from 'child_process';
@@ -18,6 +17,8 @@ import configService from './services/config.service.js';
 import modelService from './services/model.service.js';
 import parserService from './services/parser.service.js';
 import { NangoSyncTypesFileLocation, TYPES_FILE_NAME, exampleSyncName } from './constants.js';
+import type { ListedFile } from './services/compile.service.js';
+import { listFile, listFiles } from './services/compile.service.js';
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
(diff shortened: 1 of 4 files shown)


packages/cli/lib/services/compile.service.ts
```diff
@@ -89,5 +86,29 @@ class CompileService {
...
+    if (!filePath.startsWith('./')) {
+        filePath = `./${filePath}`;
+    }
+    return {
+        inputPath: filePath,
+        outputPath: filePath.replace(/\/[^/]*$/, `/dist/${path.basename(filePath.replace('.ts', '.js'))}`),
+        baseName: path.basename(filePath, '.ts')
+    };
+}
+export function listFiles({ cwd, syncName }: { cwd?: string; syncName?: string | undefined }): ListedFile[] {
```
khaliqgant: This should have a more specific name in that it lists only typescript files.

packages/cli/lib/cli.ts
```diff
@@ -313,43 +314,38 @@ export const tscWatch = async (debug = false) => {
     watcher.on('change', (filePath: string) => {
         if (filePath === nangoConfigFile) {
             // config file changed, re-compile each ts file
-            const integrationFiles = glob.sync(`./*.ts`);
+            const integrationFiles = listFiles({});
```
khaliqgant: weird interface to pass an empty object to a function

bodinsamuel: yep true

khaliqgant (changes requested):
Few comments

packages/cli/lib/services/compile.service.ts
bodinsamuel: changed 👌🏻

bodinsamuel:
Changed naming to something more explicit, + found an other issue with the compiler

packages/cli/lib/cli.ts
```diff
@@ -314,20 +314,22 @@ export const tscWatch = async (debug = false) => {
...
-        compileFile(listFile(filePath));
+        compileFile(getFileToCompile(filePath));
     });
 
-    const compiler = tsNode.create({
-        skipProject: true, // when installed locally we don't want ts-node to pick up the package tsconfig.json file
-        compilerOptions: JSON.parse(tsconfig).compilerOptions
-    });
     function compileFile(file: ListedFile) {
+        // This needs to be re-declared each time
```
khaliqgant, line 327: What was the issue here?

bodinsamuel: If you store the compiler, files won't be recompiled after an other change. I guess there is some cache or whatever (unrelated to glob)

khaliqgant: gotcha, yeah, think I ran into this issue recently. Thanks.

khaliqgant (approved):
Haven't tested directly but code changes look good.

bodinsamuel:
Thanks, would love a test when you can tomorrow ☺️ Hopefully this is the right fix

PR #1892, opened 2024-03-21
Refactor track_deletes

TBonnin:
## Describe your changes

Refactoring track_deletes logic so we don't need an extra snapshot tables.
The logic is now simple: 
- `upsert` adds/updates records (or set `deletedAt`), including updating the sync_job_id that serves as a generationId
- when the sync is finish, if `track_deletes == true` then every records without a `deletedAt` that doesn't have the same `sync_job_id` as the current one is marked as deleted

https://linear.app/nango/issue/NAN-599/track-deletes-refactoring


packages/jobs/lib/crons/autoIdleDemo.integration.test.ts
```diff
@@ -33,32 +33,31 @@ describe('Auto Idle Demo', () => {
         });
 
         const connName = nanoid();
-        const env = await environmentService.createEnvironment(0, envName);
         await configService.createProviderConfig({
             unique_key: DEMO_GITHUB_CONFIG_KEY,
             provider: 'github',
-            environment_id: env!.id,
```
TBonnin, line 40: unrelated to track_deletes: making environment id not optional, same from sync and syncJob below

packages/shared/lib/models/Data.ts
```diff
@@ -1,14 +1,12 @@
 export interface DataResponse {
     id?: string;
-    [index: string]: unknown | undefined | string | number | boolean | Record<string, string | boolean | number | unknown>;
+    [index: string]: null | undefined | object | string | number | boolean | Record<string, string | boolean | number>;
```
TBonnin, line 3: unrelated but the linter was yelling at me

packages/shared/lib/models/Data.ts
```diff
@@ -1,14 +1,12 @@
...
 }
 
 export interface UpsertSummary {
     addedKeys: string[];
     updatedKeys: string[];
     deletedKeys?: string[];
-    affectedInternalIds: string[];
-    affectedExternalIds: string[];
```
TBonnin, line 11: not used anywhere

packages/shared/lib/services/sync/run.service.ts
```diff
@@ -401,25 +401,17 @@ export default class SyncRun {
     async finishSync(models: string[], syncStartDate: Date, version: string, totalRunTime: number, trackDeletes?: boolean): Promise<void> {
         let i = 0;
         for (const model of models) {
+            let deletedKeys: string[] = [];
             if (!this.isWebhook && trackDeletes) {
```
TBonnin, line 408: @khaliqgant I have kept the condition here but I am not sure why `!this.isWebhook` is necessary

khaliqgant: Because a webhook and a sync with track deletes share the same config

packages/shared/lib/services/sync/job.service.ts
```diff
@@ -17,15 +17,18 @@ export const createSyncJob = async (
...
     run_id?: string
 ): Promise<Pick<SyncJob, 'id'> | null> => {
-    const job: SyncJob = {
+    let job = {
         sync_id,
         type,
         status,
         job_id
-    };
+    } as { sync_id: string; type: SyncType; status: SyncStatus; job_id: string; run_id?: string };
```
bodinsamuel: Seems more correct to do:
```ts
    let job: { sync_id: string; type: SyncType; status: SyncStatus; job_id: string; run_id?: string } = {
        sync_id,
        type,
        status,
        job_id
    };
```

packages/shared/lib/services/sync/data/records.service.ts
```diff
@@ -63,17 +52,7 @@ export const formatDataRecords = (
...
-        if (softDelete) {
-            if (record[deletedAtKey]) {
-                external_deleted_at = dayjs(record[deletedAtKey] as string).toDate();
-            } else {
-                external_deleted_at = new Date();
-            }
-        }
-
-        formattedRecords[i] = {
+        const formattedRecord = {
```
bodinsamuel:
```suggestion
        const formattedRecord: SyncDataRecord = {
```

packages/shared/lib/services/sync/data/records.service.ts
```diff
@@ -83,14 +62,15 @@ export const formatDataRecords = (
             sync_id: syncId,
             sync_job_id,
             external_is_deleted: softDelete,
-            external_deleted_at,
             pending_delete: false
-        };
+        } as SyncDataRecord;
```
bodinsamuel:
```suggestion
        };
```

packages/shared/lib/services/sync/data/records.service.ts
```diff
@@ -536,36 +516,45 @@ export async function getRecordsByExternalIds(external_ids: string[], nango_conn
...
+        .from<SyncDataRecord>(RECORDS_TABLE)
+        .where({
+            nango_connection_id: connectionId,
+            model,
+            sync_id: syncId,
+            external_is_deleted: false
+        })
+        .whereNot({
+            sync_job_id: generation
+        })
```
bodinsamuel, line 551: I'm wondering about the performance of this query. There is no index currently mapping those conditions, do you have a way to run an explain in prod with a large sync?

packages/shared/lib/services/sync/run.service.integration.test.ts
```diff
@@ -0,0 +1,356 @@
...
+            const newRecords = [
+                { id: '1', name: 'A' },
+                { id: '3', name: 'c' }
+            ];
+            const expectedResult = { added: 1, updated: 1, deleted: 0 };
+            const records = await verifySyncRun(rawRecords, newRecords, trackDeletes, expectedResult);
+
+            records
+                .filter((record) => record.id == 1)
+                .forEach((record) => {
```
bodinsamuel: Note that with this method if `records` is empty or not returning the correct set of records no `expect` will fail

bodinsamuel (commented):
Works well, minor comments 👌🏻 
I have one test that I think failed, but I don't know enough the system to be sure it's not what we want.
Run a sync three times:
- 1. Add record id: 1 ✅
- 2. Delete record id: 1 ✅
- 3. Add back record id: 1 ❌
The record column `external_deleted_at` is still filled with a date

---

I guess the next steps after a few weeks is to drop the table and `pending_delete` column?

TBonnin:
> Works well, minor comments 👌🏻 I have one test that I think failed, but I don't know enough the system to be sure it's not what we want. Run a sync three times:
> 
>     * 1. Add record id: 1 ✅
> 
>     * 2. Delete record id: 1 ✅
> 
>     * 3. Add back record id: 1 ❌
>          The record column `external_deleted_at` is still filled with a date 
Good catch. will fix. Question though: should the creation date of a deleted records that is being brought back be reset or not. ie: should it be considered as added again or simply updated?

TBonnin:
> I guess the next steps after a few weeks is to drop the table and `pending_delete` column?

The table will be dropped after the migration and the new table is not going to have a pending_delete column

packages/shared/lib/services/sync/data/records.service.ts
bodinsamuel: Discussed in Slack, it seems alright but If not, an index like this ` nango_connection_id, model, sync_id, job_id where external_is_deleted = false` could work

packages/shared/lib/db/migrations/20240322141605_undelete_records_reset_createdat.cjs
```diff
@@ -0,0 +1,37 @@
...
+            RETURNS TRIGGER AS $$
+            BEGIN
+                IF OLD.external_deleted_at IS NOT NULL AND NEW.external_deleted_at IS NULL THEN
+                    NEW.created_at = NOW();
+                    NEW.updated_at = NOW();
+                    NEW.external_is_deleted = FALSE;
+                END IF;
+                RETURN NEW;
+            END;
+            $$ LANGUAGE plpgsql;
```
bodinsamuel, line 17: There is no way to have this logic in the code? I'm really not fond of sql logic, it's hard to debug and easy to forget

bodinsamuel (approved):
Looks good, very nice improvement 🙏🏻

packages/shared/lib/db/migrations/20240322141605_undelete_records_reset_createdat.cjs
TBonnin: I am also not a huge fan of the fact it is easy to forget. The alternative would be to fetch all existing records and compare before merging

khaliqgant (approved):
Works as expected from my tests, thanks!

PR #1893, opened 2024-03-21
record id should not change between script executions

TBonnin:
## Describe your changes

Currently the record internal id is modified every time a record is updated which is counter intuitive and makes debugging harder

With this commit a record id is now a uuid v5 generated from the unique composite key (connection, model, external id)

### DIFF
--- a/packages/shared/lib/services/sync/data/records.service.ts
+++ b/packages/shared/lib/services/sync/data/records.service.ts
@@ -33,6 +33,13 @@ export const formatDataRecords = (
     trackDeletes = false,
     softDelete = false
 ): ServiceResponse<SyncDataRecord[]> => {
+    // hashing unique composite key (connection, model, external_id)
+    // to generate stable record ids across script executions
+    const stableId = (rawRecord: DataResponse): string => {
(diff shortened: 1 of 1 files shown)


bodinsamuel (approved):

packages/shared/lib/services/sync/data/records.service.ts
```diff
@@ -33,6 +33,13 @@ export const formatDataRecords = (
     trackDeletes = false,
     softDelete = false
 ): ServiceResponse<SyncDataRecord[]> => {
+    // hashing unique composite key (connection, model, external_id)
+    // to generate stable record ids across script executions
+    const stableId = (rawRecord: DataResponse): string => {
+        const namespace = uuid.v5(`${nango_connection_id}${model}`, uuid.NIL);
+        return uuid.v5(`${nango_connection_id}${model}${rawRecord.id}`, namespace);
```
TBonnin, line 40: having a namespace hierarchy is not mandatory, we could use a constant root namespace. Let me know what you think?

bodinsamuel: could be just nango_connection_id yes, otherwise I'm fine with this

PR #1895, opened 2024-03-22
[nan-621] output # of records

khaliqgant:
## Describe your changes
Ouput a summary of records that would be added, updated, or deleted by the integration script

NAN-621

### DIFF
--- a/packages/cli/lib/services/dryrun.service.ts
+++ b/packages/cli/lib/services/dryrun.service.ts
@@ -143,7 +143,10 @@ class DryRunService {
             stubbedMetadata = rawStubbedMetadata;
         }
 
-        const logMessages: string[] = [];
+        const logMessages = {
+            counts: { updated: 0, added: 0, deleted: 0 },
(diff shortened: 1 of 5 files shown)


packages/cli/lib/services/dryrun.service.ts
```diff
@@ -143,7 +143,10 @@ class DryRunService {
             stubbedMetadata = rawStubbedMetadata;
         }
 
-        const logMessages: string[] = [];
+        const logMessages = {
+            counts: { updated: 0, added: 0, deleted: 0 },
```
bodinsamuel, line 147: not sure I would have mixed those things, they seems a bit unrelated?

bodinsamuel (approved):

packages/cli/lib/services/dryrun.service.ts
khaliqgant: Just oganized this way to make the view concise

image

PR #1903, opened 2024-03-25
(feat) [nan-591] insert into bigQuery

khaliqgant:
## Describe your changes
Add in BigQuery data ingestion from sync, actions, and webhooks

NAN-591


TBonnin:
Can we start making dedicated packages for this kind of functionality instead of adding it into `shared`.  Maybe inside `/packages/internals/` to signifiy they will never be published?

packages/shared/lib/clients/big-query.client.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    private async createTable() {
+        const table = this.client.dataset(this.datasetName).table(`${getEnv()}_${this.tableName}`);
+        const [exists] = await table.exists();
+        if (!exists) {
+            await table.create({
+                schema: {
+                    fields: [
+                        { name: 'executionType', type: 'STRING' },
+                        { name: 'connectionId', type: 'STRING' },
```
bastienbeurier: @khaliqgant which connection ID is this? The one picked by the user, or the primary key from the db? We probably want both.

bastienbeurier:
I wonder if there's a way to flag or exclude test data. 

Random idea: accounts created with a @nango.dev email are flagged as test accounts in our main DB & excluded from tracking?

packages/shared/lib/clients/big-query.client.ts
khaliqgant: Added `internalConnectionId` which is the primary key from the db

bastienbeurier: Thank you!

packages/shared/lib/services/sync/run.service.ts
```diff
@@ -158,7 +159,7 @@ export default class SyncRun {
         if (!nangoConfig) {
             const message = `No ${this.isAction ? 'action' : 'sync'} configuration was found for ${this.syncName}.`;
             if (this.activityLogId) {
-                await this.reportFailureForResults(message);
+                await this.reportFailureForResults(message, 0);
```
bodinsamuel: I would be nice to use more json object, because when I read the code here it's impossible to know what 0 means

packages/shared/lib/services/sync/run.service.ts
```diff
@@ -557,9 +564,25 @@ export default class SyncRun {
             },
             `syncId:${this.syncId}`
         );
+
+        await BigQueryClient.insert({
```
bodinsamuel: if this insert is already try/catch I would say we don't care about await this here
```suggestion
        void BigQueryClient.insert({
```

packages/shared/package.json
```diff
@@ -19,6 +19,7 @@
     "dependencies": {
         "@aws-sdk/client-s3": "3.348.0",
         "@datadog/datadog-api-client": "^1.16.0",
+        "@google-cloud/bigquery": "^7.5.1",
```
bodinsamuel: can we pin dependency? every google sdk are auto generated and they are know to introduce small breaking changes

packages/shared/lib/clients/big-query.client.ts
```diff
@@ -0,0 +1,72 @@
...
+                        { name: 'accountId', type: 'INTEGER' },
+                        { name: 'scriptName', type: 'STRING' },
+                        { name: 'scriptType', type: 'STRING' },
+                        { name: 'environmentId', type: 'INTEGER' },
+                        { name: 'providerConfigKey', type: 'STRING' },
+                        { name: 'status', type: 'STRING' },
+                        { name: 'syncId', type: 'STRING' },
+                        { name: 'content', type: 'STRING' },
+                        { name: 'runTime', type: 'INTEGER' }, // in seconds
+                        { name: 'timestamp', type: 'INTEGER' }
```
bodinsamuel: what is timestamp? date of creation of the row, or start of the sync or end of the sync?

bodinsamuel (commented):
Minor comments, looks good otherwise 👌🏻

packages/shared/lib/clients/big-query.client.ts
```diff
@@ -0,0 +1,72 @@
...
+type Row = Record<string, string | number | undefined>;
+
+class BigQueryClient {
+    private client: BigQuery;
+    private datasetName: string;
+    private tableName: string;
+
+    constructor() {
+        this.client = new BigQuery();
+        this.datasetName = 'dataset';
```
TBonnin: let's call it `raw`. Right now we would be querying `raw` tables directly but soon we are going to want to aggregate, clean, process those dataset tagles into something that can be displayed in dashboard, etc...

packages/shared/lib/clients/big-query.client.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    private async initializeTable() {
+        await this.createDataSet();
+        await this.createTable();
+    }
+
+    public async insert(data: Row) {
+        try {
+            if (isCloud()) {
+                await this.initializeTable();
```
TBonnin: can we do this once. Otherwise every insert is going to check if table exists first

packages/shared/lib/clients/big-query.client.ts
```diff
@@ -0,0 +1,72 @@
...
+
+class BigQueryClient {
+    private client: BigQuery;
+    private datasetName: string;
+    private tableName: string;
+
+    constructor() {
+        this.client = new BigQuery();
+        this.datasetName = 'dataset';
+        this.tableName = 'script_runs';
```
TBonnin: Should the table name be a parameter of insert. I imagine we might want to have more tables in the future

packages/shared/lib/clients/big-query.client.ts
```diff
@@ -0,0 +1,72 @@
...
+                        { name: 'connectionId', type: 'STRING' },
+                        { name: 'accountId', type: 'INTEGER' },
+                        { name: 'scriptName', type: 'STRING' },
+                        { name: 'scriptType', type: 'STRING' },
+                        { name: 'environmentId', type: 'INTEGER' },
+                        { name: 'providerConfigKey', type: 'STRING' },
+                        { name: 'status', type: 'STRING' },
+                        { name: 'syncId', type: 'STRING' },
+                        { name: 'content', type: 'STRING' },
+                        { name: 'runTime', type: 'INTEGER' }, // in seconds
```
TBonnin:
```suggestion
                        { name: 'runTimeSecs', type: 'INTEGER' },
```

packages/shared/lib/clients/big-query.client.ts
TBonnin:
```suggestion
                        { name: 'createdAt', type: 'INTEGER' }
```

packages/shared/lib/clients/big-query.client.ts
```diff
@@ -0,0 +1,72 @@
...
+            });
+        }
+    }
+
+    private async initializeTable() {
+        await this.createDataSet();
+        await this.createTable();
+    }
+
+    public async insert(data: Row) {
```
TBonnin: can we have a more specific type like `RunScriptRow` to avoid passing malformed object that don't match the schema. Ideally we could even generate the BQ schema from this type but for now that's fine if we have duplication I think

packages/shared/lib/services/sync/run.service.ts
```diff
@@ -24,7 +24,28 @@ import type { Environment } from '../../models/Environment';
 import type { Metadata } from '../../models/Connection';
 import * as recordsService from './data/records.service.js';
 
+interface BigQueryClientInterface {
+    insert(row: RunScriptRow): void;
+}
+
+interface RunScriptRow {
```
khaliqgant, line 31: Don't love duplicated types but given our setup this is needed for now

packages/data-ingestion/package.json
```diff
@@ -0,0 +1,23 @@
+{
+    "name": "@nangohq/data-ingestion",
```
bodinsamuel, line 2: Nice to see the first split, I think the name is debatable but no strong opinion (e.g: bigquery, bg, billing, analytics, etc.)

packages/jobs/lib/activities.ts
```diff
@@ -23,11 +23,18 @@ import {
     getLastSyncDate
 } from '@nangohq/shared';
 import { getLogger } from '@nangohq/utils/dist/logger.js';
+import { BigQueryClient } from '@nangohq/data-ingestion/dist/index.js';
```
bodinsamuel, line 26:
```suggestion
import { BigQueryClient } from '@nangohq/data-ingestion';
```

packages/data-ingestion/lib/index.ts
```diff
@@ -0,0 +1,104 @@
...
+
+class BigQueryClient {
+    private client: BigQuery;
+    private datasetName: string;
+    private tableName: string;
+
+    constructor({ datasetName, tableName }: { datasetName: string; tableName: string }) {
+        this.client = new BigQuery();
+        this.tableName = tableName;
+        this.datasetName = datasetName;
```
bodinsamuel, line 32: Might be interesting to throw on isCloud if the env var is empty

packages/data-ingestion/lib/index.ts
```diff
@@ -0,0 +1,104 @@
...
+    scriptName: string;
+    scriptType: string;
+    environmentId: number;
+    providerConfigKey: string;
+    status: string;
+    syncId: string;
+    content: string;
+    runTimeInSeconds: number;
+    createdAt: number;
+}
```
bodinsamuel, line 22: not a blocker: the schema is not linked to the type, which could lead to desync
```ts

const fields: { name: keyof RunScriptRow; type: 'STRING' | 'INTEGER' }[] = [
    { name: 'accountId', type: 'STRING' },
    { name: 'connectionId', type: 'STRING' }
];
```

bodinsamuel (commented):
Looks good (can't really test the code) 👌🏻

Maybe we can use optionalDependencies to avoid code duplication, so it would work locally with Typescript and there will be no ref about the type in the code after compilation. 

Also it's missing on tsconfig.build.json

packages/shared/lib/services/sync/run.service.ts
bodinsamuel: Thanks, so much better 🙏🏻

packages/data-ingestion/lib/index.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    public async insert(data: RunScriptRow, tableName?: string) {
+        const table = tableName || this.tableName;
+        try {
+            if (isCloud) {
+                await this.client.dataset(this.datasetName).table(table).insert(data);
+            }
+            if (isLocal) {
+                logger.info(`Data would be inserted into BigQuery type ${JSON.stringify(data, null, 2)}`);
+            }
```
TBonnin: that's for debugging I guess?

packages/shared/lib/services/sync/run.service.ts
```diff
@@ -24,7 +24,28 @@ import type { Environment } from '../../models/Environment';
...
+    providerConfigKey: string;
+    status: string;
+    syncId: string;
+    content: string;
+    runTimeInSeconds: number;
+    createdAt: number;
+}
+
 interface SyncRunConfig {
+    bigQueryClient: BigQueryClientInterface;
```
TBonnin: can this be optional so we don't even need to pass it in the tests or any situation it is no required?

packages/data-ingestion/lib/index.ts
```diff
@@ -0,0 +1,104 @@
...
+                        { name: 'content', type: 'STRING' },
+                        { name: 'runTimeInSeconds', type: 'FLOAT' },
+                        { name: 'createdAt', type: 'INTEGER' }
+                    ]
+                }
+            });
+        }
+    }
+
+    public async insert(data: RunScriptRow, tableName?: string) {
```
TBonnin, line 89: Since we are not enforcing type and schema matching, should insert simply accept any record<string, string|number>?

packages/data-ingestion/lib/index.ts
khaliqgant: Yeah, can remove it if you think it is noise?

packages/shared/lib/services/sync/run.service.ts
khaliqgant: done

packages/data-ingestion/lib/index.ts
khaliqgant: Won't work as then there would be a type mismatch between the two.

packages/jobs/lib/activities.ts
khaliqgant: Would rather keep it explicit with the file extension I think.

packages/data-ingestion/lib/index.ts
khaliqgant: yeah, will leave this out though for now

packages/data-ingestion/lib/index.ts
khaliqgant: Removed

TBonnin (approved):

PR #1905, opened 2024-03-25
Improve scripts reference docs

bastienbeurier:


docs-v2/reference/scripts.mdx
```diff
@@ -56,37 +56,343 @@ Read more about [integration scripts](/understand/concepts/scripts) to understan
...
+### GET requests
 
-### HTTP requests
+Makes an HTTP GET request inside an integration script.
 
-HTTP requests in integration scripts are made with the proxy ([concept]((/understand/concepts/proxy)) / [step-by-step guide](/integrate/guides/proxy-requests-to-an-api) / [reference](/reference/sdks/node#proxy)).
+```js
+const res = await nango.get({
+    endpoint: '/endpoint',
+    baseUrlOverride: 'https://base-url.com'
```
TBonnin: why do we mention baseUrlOverride (here and in the example below)? It is a optional param that is defined below. I would keep the example as simple as possible imho

docs-v2/reference/scripts.mdx
```diff
@@ -56,37 +56,343 @@ Read more about [integration scripts](/understand/concepts/scripts) to understan
...
+            </ResponseField>
+            <ResponseField name="decompress" type="boolean">
+                Override the decompress option when making requests. Optional, defaults to false
+            </ResponseField>
+            <ResponseField name="responseType" type="'arraybuffer' | 'blob' | 'document' | 'json' | 'text' | 'stream'">
+                The type of the response. 
+            </ResponseField>
+        </Expandable>
+    </ResponseField>
+</Expandable>
```
TBonnin: would it make sense to have a section describing the fields common to all the methods instead of repeating them?

docs-v2/reference/scripts.mdx
```diff
@@ -56,37 +56,343 @@ Read more about [integration scripts](/understand/concepts/scripts) to understan
...
+            </ResponseField>
+            <ResponseField name="headers" type="Record<string, string>">
+                The headers of the request.
+            </ResponseField>
+            <ResponseField name="params" type="Record<string, string | number>">
+                The query parameters of the request.
+            </ResponseField>
+            <ResponseField name="data" type="unknown">
+                The body of the request.
+            </ResponseField>
```
TBonnin: I don't think `get` has a data field

docs-v2/reference/scripts.mdx
```diff
@@ -56,37 +56,343 @@ Read more about [integration scripts](/understand/concepts/scripts) to understan
...
+            <ResponseField name="headers" type="Record<string, string>">
+                The headers of the request.
+            </ResponseField>
+            <ResponseField name="params" type="Record<string, string | number>">
+                The query parameters of the request.
+            </ResponseField>
+            <ResponseField name="data" type="unkown">
+                The body of the request.
+            </ResponseField>
```
TBonnin, line 92: same a `get`. `delete`doesn't take a body

docs-v2/reference/scripts.mdx
```diff
@@ -56,37 +56,343 @@ Read more about [integration scripts](/understand/concepts/scripts) to understan
 
 Integration scripts expose a helper object (`NangoSync` for sync scripts, `NangoAction` for action scripts), which allows to interact with external APIs & Nango more easily. 
 
+# HTTP requests
+
 <Tip>
-The script helper object shares some methods, but not all, with the [Node backend SDK](/reference/sdks/node), which is why this reference will link to the Node SDK reference a lot.
+Note that all HTTP requests benefit from automatic credential injection. Because scripts are executed in the context of a specific integration & connection, Nango can automatically retrieve  & refresh the relevant API credentials.
```
khaliqgant:
```suggestion
Note that all HTTP requests benefit from automatic credential injection. Because scripts are executed in the context of a specific integration & connection, Nango can automatically retrieve & refresh the relevant API credentials.
```

docs-v2/reference/scripts.mdx
```diff
@@ -56,37 +56,343 @@ Read more about [integration scripts](/understand/concepts/scripts) to understan
...
+<Tip>
+    The response object is an [Axios response](https://axios-http.com/docs/res_schema).
+</Tip>
+</Expandable>
+
+### PATCH requests
+
+Makes an HTTP PATCH request inside an integration script.
+
+Triggers an action for a connection.
```
bodinsamuel: I'm not sure I get this sentence in this context

docs-v2/reference/scripts.mdx
```diff
@@ -56,37 +56,343 @@ Read more about [integration scripts](/understand/concepts/scripts) to understan
...
+<Tip>
+    The response object is an [Axios response](https://axios-http.com/docs/res_schema).
 </Tip>
+</Expandable>
+
+### DELETE requests
+
+Makes an HTTP DELETE request inside an integration script.
+
+Triggers an action for a connection.
```
bodinsamuel: same

bodinsamuel (approved):
Minor comments 👌🏻

docs-v2/reference/scripts.mdx
```diff
@@ -99,29 +405,205 @@ await nango.log("This is a log.");
...
+    </ResponseField>
+</Expandable>
+
+**Response**
+
+Empty response.
+
+### Edit connection metadata
+
+Edit custom metadata for the connection. Only overrides specified properties, not the entire metadata.
```
khaliqgant: Is it clear that it overrides and also adds properties potentially?

```
await nango.updateMetadata({ 
   'CUSTOM_KEY1': 'CUSTOM_VALUE1',
   'new-prop': 'new-value
});

```

docs-v2/reference/scripts.mdx
bastienbeurier: Done!

docs-v2/reference/scripts.mdx
bastienbeurier: Good idea, done!

docs-v2/reference/scripts.mdx
bastienbeurier: If I'm not mistaken, it's not standard but technically possible.

docs-v2/reference/scripts.mdx
bastienbeurier: Edited!

docs-v2/reference/scripts.mdx
bastienbeurier: Removed

docs-v2/reference/scripts.mdx
bastienbeurier: Removed

PR #1906, opened 2024-03-26
[nan-638] add in credentials override and scopes override

khaliqgant:
## Describe your changes
Allow a user to override oauth credentials and scopes. Show in the UI only for netsuite

NAN-638

### DIFF
--- a/packages/frontend/lib/index.ts
+++ b/packages/frontend/lib/index.ts
@@ -99,9 +99,14 @@ export default class Nango {
     public auth(
         providerConfigKey: string,
         connectionId: string,
-        options?: (ConnectionConfig | BasicApiCredentials | ApiKeyCredentials | AppStoreCredentials) & AuthOptions
+        options?: (ConnectionConfig | OptionalOAuthCredentials | BasicApiCredentials | ApiKeyCredentials | AppStoreCredentials) & AuthOptions
     ): Promise<AuthResult> {
-        if (options && 'credentials' in options && Object.keys(options.credentials).length > 0) {
+        if (
+            options &&
+            'credentials' in options &&
+            (!('oauth_client_id' in options.credentials) || !('oauth_client_secret' in options.credentials)) &&
+            Object.keys(options.credentials).length > 0
+        ) {
             const credentials = options.credentials as BasicApiCredentials | ApiKeyCredentials;
             const { credentials: _, ...connectionConfig } = options as ConnectionConfig;
 
@@ -311,6 +316,14 @@ export default class Nango {
                 query.push(`user_scope=${connectionConfig.user_scope.join(',')}`);
             }
 
+            if (connectionConfig.credentials) {
+                const credentials = connectionConfig.credentials;
+                if ('oauth_client_id' in credentials && 'oauth_client_secret' in credentials) {
+                    query.push(`credentials[oauth_client_id]=${credentials.oauth_client_id}`);
+                    query.push(`credentials[oauth_client_secret]=${credentials.oauth_client_secret}`);
+                }
+            }
+
             for (const param in connectionConfig.authorization_params) {
                 const val = connectionConfig.authorization_params[param];
                 if (typeof val === 'string') {
@@ -330,7 +343,12 @@ interface ConnectionConfig {
     hmac?: string;
     user_scope?: string[];
     authorization_params?: Record<string, string | undefined>;
-    credentials?: BasicApiCredentials | ApiKeyCredentials | AppStoreCredentials;
+    credentials?: OptionalOAuthCredentials | BasicApiCredentials | ApiKeyCredentials | AppStoreCredentials;
+}
+
+interface OptionalOAuthCredentials {
+    oauth_client_id: string;
+    oauth_client_secret: string;
 }
 
 interface BasicApiCredentials {
--- a/packages/server/lib/controllers/oauth.controller.ts
+++ b/packages/server/lib/controllers/oauth.controller.ts
@@ -16,7 +16,7 @@ import type {
     TemplateOAuth2 as ProviderTemplateOAuth2,
     OAuthSession,
     OAuth1RequestTokenResult,
-    AuthCredentials,
+    OAuth2Credentials,
     ConnectionConfig
 } from '@nangohq/shared';
 import {
@@ -95,6 +95,7 @@ class OAuthController {
             const callbackUrl = await getOauthCallbackUrl(environmentId);
             const connectionConfig = req.query['params'] != null ? getConnectionConfig(req.query['params']) : {};
             const authorizationParams = req.query['authorization_params'] != null ? getAdditionalAuthorizationParams(req.query['authorization_params']) : {};
+            const overrideCredentials = req.query['credentials'] != null ? getAdditionalAuthorizationParams(req.query['credentials']) : {};
 
             if (connectionId == null) {
                 await createActivityLogMessageAndEnd({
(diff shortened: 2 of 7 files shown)


packages/webapp/src/types.ts
```diff
@@ -134,20 +134,26 @@ export interface Connection {
...
 }
 
 export interface BasicApiCredentials {
-    [key: string]: string;
     username: string;
     password: string;
 }
 
 export interface ApiKeyCredentials {
-    [key: string]: string;
```
bastienbeurier, line 147: Isn't this going to break some things? Don't we have weird cases of API key auth where users need to pass in more fields?

packages/webapp/src/pages/Connection/Create.tsx
```diff
@@ -294,11 +330,29 @@ export default function IntegrationCreate() {
...
         }
 
+        let oauthCredentialsString = '';
+
+        if (integration?.authMode === AuthModes.OAuth2 && optionalOAuthClientId && optionalOAuthClientSecret) {
+            oauthCredentialsString = `
+    credentials: {
+        oauth_client_id: '${optionalOAuthClientId}',
+        oauth_client_secret: '${optionalOAuthClientSecret}'
+}`;
```
bodinsamuel: display: it's missing a line break here (or in the snippet)

packages/webapp/src/pages/Connection/Create.tsx
```diff
@@ -395,6 +449,54 @@ nango.${integration?.authMode === AuthModes.None ? 'create' : 'auth'}('${integra
                                 </div>
                             )}
 
+                            {integration?.provider === 'netsuite' && (
```
bodinsamuel, line 457: there is no other way to detect this?

packages/shared/lib/models/Auth.ts
```diff
@@ -129,6 +129,12 @@ export interface OAuth2Credentials extends CredentialsCommon {
 
     refresh_token?: string;
     expires_at?: Date | undefined;
+
+    config_override?: {
+        client_id?: string;
+        client_secret?: string;
+        scopes?: string | string[];
```
bodinsamuel: nitpicking but is it possible to only allow `string[]`?

packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -989,6 +1018,40 @@ class OAuthController {
...
+                        client_secret: connectionConfig['oauth_client_secret']
+                    }
+                };
+
+                connectionConfig = Object.keys(session.connectionConfig).reduce((acc: Record<string, string>, key: string) => {
+                    if (key !== 'oauth_client_id' && key !== 'oauth_client_secret') {
+                        acc[key] = connectionConfig[key] as string;
+                    }
+                    return acc;
+                }, {});
```
bodinsamuel, line 1035: maybe it's just me but I have hard time reading reducer, I don't really understand what's going on

packages/server/lib/controllers/connection.controller.ts
```diff
@@ -558,6 +558,26 @@ class ConnectionController {
...
+                }
+
+                if (req.body['oauth_client_secret']) {
+                    oAuthCredentials.config_override = {
+                        ...oAuthCredentials.config_override,
+                        client_secret: req.body['oauth_client_secret']
+                    };
+                }
+
+                if (req.body['oauth_scopes']) {
```
bodinsamuel, line 574: does those three do not need to be set altogether? 
like what happen if we only set the secret or only the scopes? from what I understand I would expect to be everything or nothing

bodinsamuel (commented):
Minor comments, the code regarding auth is really hard to read but seems good

packages/frontend/lib/index.ts
```diff
@@ -99,9 +99,14 @@ export default class Nango {
     public auth(
         providerConfigKey: string,
         connectionId: string,
-        options?: (ConnectionConfig | BasicApiCredentials | ApiKeyCredentials | AppStoreCredentials) & AuthOptions
+        options?: (ConnectionConfig | OptionalOAuthCredentials | BasicApiCredentials | ApiKeyCredentials | AppStoreCredentials) & AuthOptions
```
TBonnin: what about `OverrideOAuthCredentials` instead, to make it clear they would override the integration ones

packages/frontend/lib/index.ts
```diff
@@ -311,6 +316,14 @@ export default class Nango {
                 query.push(`user_scope=${connectionConfig.user_scope.join(',')}`);
             }
 
+            if (connectionConfig.credentials) {
+                const credentials = connectionConfig.credentials;
+                if ('oauth_client_id' in credentials && 'oauth_client_secret' in credentials) {
+                    query.push(`credentials[oauth_client_id]=${credentials.oauth_client_id}`);
+                    query.push(`credentials[oauth_client_secret]=${credentials.oauth_client_secret}`);
```
TBonnin, line 323: same here what about making it clearer with `override_oauth_client_id` and `override_oauth_client_secret` keys?

packages/server/lib/controllers/oauth.controller.ts
TBonnin: I don't mind `reduce` personally but it would be way more readable in typescript if the initial value was the first parameter instead of the last one after the function. Sometimes I have a hard time understanding language creators/maintainers.

packages/webapp/src/types.ts
khaliqgant: This is just the webapp types so it won't break anything that a user is doing.

packages/webapp/src/pages/Connection/Create.tsx
khaliqgant: In the display this is what I see which is expected
image

Can you clarify what the issue you're seeing is?

packages/webapp/src/pages/Connection/Create.tsx
khaliqgant: Not without putting in some more logic in the providers.yaml. IMO this works for now and it is only one provider that we're adding support for ATM.

packages/server/lib/controllers/oauth.controller.ts
khaliqgant: I would have rather just done
```
delete connectionConfig.oauth_client_id;
delete connectionConfig.oauth_client_secret;
```
but eslint was yelling at me.

packages/server/lib/controllers/connection.controller.ts
khaliqgant: The three do not need to be set altogether, no. Scopes can be edited independently. A user could theoretically just override the client_id and not the secret which is why I separated them out.

packages/shared/lib/models/Auth.ts
khaliqgant: Updated. It is actually just a `string`. It multiple scopes it is comma separated

packages/frontend/lib/index.ts
khaliqgant: Opted for `OAuthCredentialsOverride`

packages/webapp/src/pages/Connection/Create.tsx
bodinsamuel: it's a detail,  `} })` line 5

packages/frontend/lib/index.ts
khaliqgant: Don't think this is necessary here if the interface is `OAuthCredentialsOverride` as this is just what we push to our server so it is just an internal name

packages/server/lib/controllers/oauth.controller.ts
bodinsamuel: ah yes I see, I think it's fair to disable some eslint errors when we don't find them relevant. 
And if it's often the case, it's always good to revisit the list of activated rules 👌🏻

packages/server/lib/controllers/connection.controller.ts
bodinsamuel: okay got it, thanks for the clarification 👌🏻

bodinsamuel (approved):

packages/webapp/src/pages/Connection/Create.tsx
khaliqgant: Ah gotcha. Updated with https://github.com/NangoHQ/nango/pull/1906/commits/1eff44a4ff4624531b20da2146a775ab84c953d8:

image
image
image
image

TBonnin (approved):

PR #1907, opened 2024-03-26
fix: batchDelete returns incorrect sync result

TBonnin:
## Describe your changes
Sync can return incorrect deletion results. 
Fixing by getting directly getting the modified rows when `soft_delete = true`

https://linear.app/nango/issue/NAN-641/batchdelete-doesnt-return-the-correct-result

### DIFF
--- a/packages/shared/lib/services/sync/data/data.service.ts
+++ b/packages/shared/lib/services/sync/data/data.service.ts
@@ -35,13 +35,18 @@ export async function upsert(
     try {
         const encryptedRecords = encryptionManager.encryptDataRecords(recordsWithoutDuplicates);
 
-        await schema().from(RECORDS_TABLE).insert(encryptedRecords).onConflict(['nango_connection_id', 'external_id', 'model']).merge();
+        const externalIds = await schema()
+            .from<DataRecord>(RECORDS_TABLE)
+            .insert(encryptedRecords)
+            .onConflict(['nango_connection_id', 'external_id', 'model'])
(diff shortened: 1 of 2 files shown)


packages/shared/lib/services/sync/run.service.integration.test.ts
```diff
@@ -164,7 +164,7 @@ describe('Running sync', () => {
                 { id: '1', name: 'a' },
                 { id: '2', name: 'b' }
             ];
-            const expectedResult = { added: 0, updated: 0, deleted: 0 };
+            const expectedResult = { added: 0, updated: 0, deleted: 2 };
```
TBonnin, line 167: I had the wrong test 🤦

bodinsamuel (approved):

packages/shared/lib/services/sync/data/data.service.ts
```diff
@@ -35,13 +35,18 @@ export async function upsert(
...
+            .insert(encryptedRecords)
+            .onConflict(['nango_connection_id', 'external_id', 'model'])
+            .merge()
+            .returning('external_id');
 
         if (softDelete) {
             return {
                 success: true,
                 summary: {
-                    deletedKeys: [...addedKeys, ...updatedKeys],
```
TBonnin, line 44: addedKeys would be empty, since it is checking if the ids is already in the db. UpdatedKeys is also empty if the record hash hasn't changed

PR #1908, opened 2024-03-26
fix(api): get env by secretKey

bodinsamuel:
## Describe your changes

Fixes NAN-642

SecretKey is encrypted meaning we can't select an env by secretKey. The problem has been hidden by the fact that we maintain a in-memory cache, but new users could suffer from this. 

- Add a new column to store the hashed version of the `secretKey`
Backfilled everything so we don't have a transition period

- Add a flag to enable or disable local cache
- Time the function in datadog so we can track the performance of it
- Bonus: remove `getByAccountIdAndEnvironment` that was duplicated by `getById`
- Bonus: add a way to properly serialize Error (we have a lot of those in the code but they don't actually work)


packages/shared/lib/services/environment.service.ts
```diff
@@ -547,4 +542,12 @@ class EnvironmentService {
     }
 }
 
+export async function hashSecretKey(key: string) {
+    if (!ENCRYPTION_KEY) {
+        return key;
+    }
+
+    return (await pbkdf2(key, ENCRYPTION_KEY, 310000, 32, 'sha256')).toString('base64');
```
TBonnin, line 553: curious to know how you picked the 31000 iterations and 32 key length value?

packages/shared/lib/services/environment.service.ts
```diff
@@ -20,11 +20,16 @@ interface EnvironmentAccount {
 type EnvironmentAccountSecrets = Record<string, EnvironmentAccount>;
 
 export const defaultEnvironments = ['prod', 'dev'];
+const CACHE_ENABLED = !(process.env['NANGO_CACHE_ENV_KEYS'] === 'false');
```
TBonnin, line 23: we have a feature flag util (reading keys from redis)
```
const useCache = await featureFlags.isEnabled('use-env-secret-cache', 'global', true);
```
If you want to be able to enable/disable it live you would need to keep adding to the cache and check the flag only when retrieving the data https://github.com/NangoHQ/nango/pull/1908/files#diff-f3d9e285186619d6dfde44bdbbb720119863b18a3a80a41025f4a9c83e13705aL97

packages/shared/lib/services/environment.service.ts
bodinsamuel: copied from our password hash function

packages/shared/lib/services/environment.service.ts
bodinsamuel: Indeed, I wasn't aware of this feature.
I think, reading this flag from redis at each call would === caching the secrets to Redis, defeating the purpose of this :D I'd leave it like this unless you really find it better

packages/shared/lib/db/migrations/20240326141028_hash-key.cjs
```diff
@@ -0,0 +1,36 @@
...
+const pbkdf2 = utils.promisify(crypto.pbkdf2);
+const ENCRYPTION_KEY = process.env['NANGO_ENCRYPTION_KEY'];
+
+function decrypt(enc, iv, authTag) {
+    const decipher = crypto.createDecipheriv('aes-256-gcm', Buffer.from(ENCRYPTION_KEY, 'base64'), Buffer.from(iv, 'base64'));
+    decipher.setAuthTag(Buffer.from(authTag, 'base64'));
+    let str = decipher.update(enc, 'base64', 'utf8');
+    str += decipher.final('utf8');
+    return str;
+}
```
bodinsamuel, line 14: Because those file are .cjs we can't really import our own code. It's either work locally but not in test, or the opposite :/

packages/shared/lib/services/environment.service.ts
TBonnin: If you are worried about skewing the telemetry. You can check the flag outside of the tracking and pass the boolean to `getAccountIdAndEnvironmentIdBySecretKey`. Up to you though

TBonnin (approved):

khaliqgant (approved):

bodinsamuel:
I added the encryption in test because there was a few edge case not possible to test without it and it's better to be closer to production. Can you re-review and test on your side please? ☺️ 
NB: work on staging

khaliqgant (approved):
Works as expected

PR #1911, opened 2024-03-27
(feat) [NAN-638] updated and docs

khaliqgant:
## Describe your changes
Updates name for the oauth overrides to be more explicit.
- Fix creating a new connection to show right after creation
- Fix deleting a connection to not show right after deletion

NAN=638

### DIFF
--- a/docs-v2/reference/sdks/frontend.mdx
+++ b/docs-v2/reference/sdks/frontend.mdx
@@ -139,6 +139,14 @@ const result = nango.auth('<INTEGRATION-ID>', '<CONNECTION-ID>', {
           <ResponseField name="password" type="string">
             For Basic authorization, pass in the user's password.
           </ResponseField>
+
+          <ResponseField name="oauth_client_id_override" type="string">
+            For OAuth2 override the integration's client id value
+          </ResponseField>
+
+          <ResponseField name="oauth_client_secret_override" type="string">
+            For OAuth2 override the integration's client secret value
+          </ResponseField>
         </Expandable>
       </ResponseField>
     </Expandable>
--- a/docs-v2/reference/sdks/frontend.mdx
+++ b/docs-v2/reference/sdks/frontend.mdx
@@ -141,11 +141,11 @@ const result = nango.auth('<INTEGRATION-ID>', '<CONNECTION-ID>', {
           </ResponseField>
 
           <ResponseField name="oauth_client_id_override" type="string">
-            For OAuth2 override the integration's client id value
+            For OAuth 2, override the integration's client ID with a connection-level client ID. This is useful when your users bring their own OAuth 2 app (e.g. Netsuite).
           </ResponseField>
 
           <ResponseField name="oauth_client_secret_override" type="string">
-            For OAuth2 override the integration's client secret value
+            For OAuth 2, override the integration's client secret with a connection-level client secret. This is useful when your users bring their own OAuth 2 app (e.g. Netsuite).
           </ResponseField>
         </Expandable>
       </ResponseField>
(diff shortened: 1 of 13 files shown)


docs-v2/spec.yaml
```diff
@@ -363,6 +363,15 @@ paths:
                                 no_expiration:
                                     type: boolean
                                     description: (OAuth2, optional) If the provider gives access tokens that don't expire, pass in `true` to avoid an import validation error.
+                                oauth_client_id_override:
+                                    type: string
+                                    description: (OAuth2, optional) Override the integration's OAuth client id
+                                oauth_client_id_secret_override:
+                                    type: string
+                                    description: (OAuth2, optional) Override the integration's OAuth client secret
+                                oauth_scopes_override:
```
bastienbeurier: Why is it an explicit param? Shouldn't there be a connection config that can be passed to the import connection endpoint?

docs-v2/reference/sdks/frontend.mdx
```diff
@@ -139,6 +139,14 @@ const result = nango.auth('<INTEGRATION-ID>', '<CONNECTION-ID>', {
           <ResponseField name="password" type="string">
             For Basic authorization, pass in the user's password.
           </ResponseField>
+
+          <ResponseField name="oauth_client_id_override" type="string">
+            For OAuth2 override the integration's client id value
```
bastienbeurier: Nit: I would be slightly more explicit. 

"For OAuth 2, override the integration's client ID with a connection-level client ID. This is useful when your users bring their own OAuth 2 app (e.g. Netsuite)."

docs-v2/reference/sdks/frontend.mdx
```diff
@@ -139,6 +139,14 @@ const result = nango.auth('<INTEGRATION-ID>', '<CONNECTION-ID>', {
...
+          </ResponseField>
+
+          <ResponseField name="oauth_client_secret_override" type="string">
+            For OAuth2 override the integration's client secret value
```
bastienbeurier: Same thing, I would be more explicit.

docs-v2/spec.yaml
```diff
@@ -363,6 +363,15 @@ paths:
                                 no_expiration:
                                     type: boolean
                                     description: (OAuth2, optional) If the provider gives access tokens that don't expire, pass in `true` to avoid an import validation error.
+                                oauth_client_id_override:
+                                    type: string
+                                    description: (OAuth2, optional) Override the integration's OAuth client id
+                                oauth_client_id_secret_override:
```
bodinsamuel:
```suggestion
                                oauth_client_secret_override:
```

packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -1018,40 +1027,45 @@ class OAuthController {
                 };
             }
 
-            if (connectionConfig['oauth_client_id'] && connectionConfig['oauth_client_secret']) {
+            if (connectionConfig['oauth_client_id_override']) {
```
bodinsamuel, line 1030: it would great to type connectionConfig and BaseConnection["connection_config"] in the future

bodinsamuel (approved):

packages/server/lib/controllers/oauth.controller.ts
khaliqgant: Yeah, might be hard to nail down but would be better for sure.

docs-v2/spec.yaml
khaliqgant: Updated

docs-v2/reference/sdks/frontend.mdx
khaliqgant: Updated

docs-v2/reference/sdks/frontend.mdx
khaliqgant: Updated

PR #1913, opened 2024-03-27
(fix) [NAN-644] refactor logger and some environment detection logic out of

khaliqgant:
## Describe your changes
Add ~`internals`~ `utils` package and move logger and some environment detection logic there

NAN-644


packages/internals/lib/environment/detection.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+export function isTest(): boolean {
+    return Boolean(process.env['CI'] !== undefined || process.env['VITEST']);
+}
+
+export function isBasicAuthEnabled() {
+    return !isCloud() && process.env['NANGO_DASHBOARD_USERNAME'] && process.env['NANGO_DASHBOARD_PASSWORD'];
+}
+
```
khaliqgant: Debatable if these two below should be here or not.

packages/shared/lib/utils/encryption.manager.ts
```diff
@@ -10,6 +10,8 @@ import type { RawDataRecordResult, DataRecord, DataRecordWithMetadata, RecordWra
 import db from '../db/database.js';
 import util from 'util';
 
+const { logger } = new Logger('Encryption.Manager');
+
 interface DataRecordJson {
     encryptedValue: string;
     [key: string]: any;
```
TBonnin, line 17: I am tempted to also move `encryption.manager` into this new package. After all it was in utils before. The question is how to we deal with types?

TBonnin (commented):
Reiterating that I think this package would be better called `utils`. And personally I would create all those internal packages (I feel we are going to have a bunch) into a subdirectory of `/packages` (for example: `libs` or `internals` to differentiate the end products (cloud services, published packages, etc...) to the stuff only used internally

packages/internals/lib/environment/detection.ts
```diff
@@ -0,0 +1,62 @@
...
+    if (isStaging()) {
+        return NodeEnv.Staging;
+    } else if (isProd()) {
+        return NodeEnv.Prod;
+    } else {
+        return NodeEnv.Dev;
+    }
+}
+
+export function isStaging() {
```
bodinsamuel: I would go for static env var, there is no benefits in making function calls all the time

packages/internals/lib/logger.ts
```diff
@@ -1,20 +1,22 @@
...
-const nangoLogFormat = winston.format.printf((info) => {
-    return `${info['timestamp']} [${info['level'].toUpperCase()}] ${info['message']}`;
-});
+const nangoLogFormat = (service = '') => {
+    return winston.format.printf((info) => {
+        return `${info['timestamp']} [${info.level.toUpperCase()}]${service ? ` [${service}] ` : ''}${info.message}`;
+    });
+};
 
 class NangoLogger {
```
bodinsamuel: I think I want to question our use of Class, most of the time they seems to hold no state, here it's just a function.
```ts
export function getLogger(server?: string) {
  winston.createLogger({
      levels: winston.config.syslog.levels,
      format: winston.format.combine(winston.format.timestamp(), nangoLogFormat),
      format: winston.format.combine(winston.format.timestamp(), nangoLogFormat(service)),
      transports: [new winston.transports.Console({ level: process.env['LOG_LEVEL'] || 'info' })]
  });
}
export const logger = getLogger();
```

packages/internals/package.json
```diff
@@ -0,0 +1,22 @@
...
+    },
+    "keywords": [],
+    "repository": {
+        "type": "git",
+        "url": "git+https://github.com/NangoHQ/nango.git",
+        "directory": "packages/internals"
+    },
+    "license": "SEE LICENSE IN LICENSE FILE IN GIT REPOSITORY",
+    "dependencies": {
+        "winston": "^3.8.2"
```
bodinsamuel: can we pin the dependency?

packages/internals/package.json
```diff
@@ -0,0 +1,22 @@
+{
+    "name": "@nangohq/nango-internals",
+    "version": "1.0.0",
+    "description": "Internal modules for use in Nango applications",
+    "type": "module",
+    "main": "dist/index.js",
+    "typings": "dist/index.d.ts",
+    "scripts": {
+        "build": "rimraf ./dist && tsc"
```
bodinsamuel: I don't think this command is necessary, the main tsconfig should build it

packages/internals/lib/logger.ts
```diff
@@ -1,20 +1,22 @@
...
             levels: winston.config.syslog.levels,
-            format: winston.format.combine(winston.format.timestamp(), nangoLogFormat),
+            format: winston.format.combine(winston.format.timestamp(), nangoLogFormat(service)),
             transports: [new winston.transports.Console({ level: process.env['LOG_LEVEL'] || 'info' })]
         });
     }
 }
 
-export default new NangoLogger().logger;
+export default NangoLogger;
```
bodinsamuel: I would also like to discuss our use of export default. 
I find it a bit inconvenient to find import across the codebase, import name is not enforced making it harder to have a consistent way of naming things. Plus if we need to export more then we have a mix of default + named.
For example here, vscode can't autocomplete Logger and only find NangoLogger which would differ from what you have wrote.

bodinsamuel (commented):
Nice starts, very glad we are moving on this.
I have a few comments, starting from almost scratch is a good time to start with alignment too 👍🏻 

I think it's also missing two things:
- Nodemon listen to internals/dist on all related packages, so it restart when modified
- Our many Dockerfiles are not updated and I'm actually surprised it passed the CI

packages/persist/nodemon.json
```diff
@@ -1,6 +1,6 @@
 {
-  "watch": ["lib", "../shared/lib", "../../.env"],
-  "ext": "ts,json",
-  "ignore": ["lib/**/*.test.ts"],
-  "exec": "tsc && tsx -r dotenv/config lib/app.ts dotenv_config_path=./../../.env"
-}
\ No newline at end of file
+    "watch": ["lib", "../shared/lib", "../utils/lib", "../../.env"],
```
bodinsamuel: note it would be cool to wait /dist not /lib because it reloads too fast for Typescript to compile and pick up the actual changes.

packages/shared/lib/services/notification/slack.service.ts
```diff
@@ -344,11 +344,11 @@ class SlackService {
         let payloadContent = '';
 
         if (connectionCount === 0) {
-            payloadContent = `[Resolved] *${syncName}* (${syncType.toLowerCase()}) failed. Read <${getBasePublicUrl()}/${envName}/activity?activity_log_id=${originalActivityLogId}|logs>.`;
+            payloadContent = `[Resolved] *${syncName}* (${syncType.toLowerCase()}) failed. Read <${baseUrl}/${envName}/activity?activity_log_id=${originalActivityLogId}|logs>.`;
         } else {
             const count = connectionCount;
             const connection = count === 1 ? 'connection' : 'connections';
-            payloadContent = `*${syncName}* (${syncType.toLowerCase()}) is failing for ${count} ${connection}. Read <${getBasePublicUrl()}/${envName}/activity?activity_log_id=${originalActivityLogId}|logs>.`;
+            payloadContent = `*${syncName}* (${syncType.toLowerCase()}) is failing for ${count} ${connection}. Read <${baseUrl}/${envName}/activity?activity_log_id=${originalActivityLogId}|logs>.`;
```
bodinsamuel: is this fix correct? seems like it should use `basePublicUrl`

packages/persist/nodemon.json
khaliqgant: Should probably do the same for shared as well then right?

bodinsamuel: yes ☺️

packages/shared/lib/services/notification/slack.service.ts
khaliqgant: Thanks, updated

packages/persist/nodemon.json
khaliqgant: Done with https://github.com/NangoHQ/nango/pull/1913/commits/2b63567aa07f7ea4c1a1927002783ff9aae364cb

packages/internals/package.json
khaliqgant: Updated

packages/internals/lib/logger.ts
khaliqgant: Updated as discussed

.github/workflows/integration-flows.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
               run: |
                   # Build, install CLI and verify it can run
                   npm install
-                  npm run build -w @nangohq/node -w @nangohq/shared -w nango
+                  npm run build -w @nangohq/node -w @nangohq/utils -w @nangohq/shared -w nango
```
bodinsamuel, line 31: I think we could just `npm run ts-build`

bodinsamuel (approved):
🚀 Amazing!

TBonnin (approved):

.github/workflows/integration-flows.yaml
bodinsamuel: I think you forgot to change that, or put back a build command in utils/package.json, but I think using ts-build is safer

khaliqgant: Building everything seemed like overkill rather than just building what we need exactly. 

I’ll add back in the specific build command which was there originally in utils. I think it’s handy to have specific build commands in each package regardless

khaliqgant: https://github.com/NangoHQ/nango/pull/1916

PR #1915, opened 2024-03-27
fix(env): pin Postgres

bodinsamuel:
## Describe your changes

Fixes NAN-648

Tests and docker-compose are not running in the same version as the Production.

- Pin Postgres version
- Prepare integration tessts to run more than one containers

NB: locally you might need to rm your data if you use the docker-compose


tests/setup.ts
```diff
@@ -23,10 +26,20 @@ export const setup = async () => {
     process.env['NANGO_DB_NAME'] = 'postgres';
     process.env['NANGO_DB_MIGRATION_FOLDER'] = './packages/shared/lib/db/migrations';
     process.env['TELEMETRY'] = 'false';
-};
+}
+
+export async function setup() {
+    await Promise.all([setupPostgres()]);
```
TBonnin, line 33: promise.All of one promise?

bodinsamuel: There will be ElasticSearch soon enough (It's just a PR split from my big PR)

tests/setup.ts
```diff
@@ -1,18 +1,21 @@
...
-export const setup = async () => {
-    const container: TestContainer = new GenericContainer('postgres');
-    startedContainer = await container
-        .withEnvironment({ POSTGRES_USER: 'postgres' })
-        .withEnvironment({ POSTGRES_PASSWORD: 'nango_test' })
-        .withEnvironment({ POSTGRES_DB: 'postgres' })
+async function setupPostgres() {
+    const container = new PostgreSqlContainer('postgres:15.5-alpine');
+    const pg = await container
+        .withDatabase('postgres')
```
khaliqgant: variable for this since it is repeated with `NANGO_DB_NAME`

bodinsamuel: good point 👌🏻

khaliqgant (approved):
🔥

PR #1916, opened 2024-03-28
[nan-644] add back in build command

khaliqgant:
## Describe your changes
Add back in specific build command

NAN-644

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,33 +1,33 @@
-name: "[Release] NPM publish"
+name: '[Release] NPM publish'
 
 on:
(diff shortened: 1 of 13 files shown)


.github/workflows/publish.yaml
```diff
@@ -1,33 +1,33 @@
...
-                description: "Version to publish"
+                description: 'Version to publish'
                 required: true
-                default: "0.0.0"
+                default: '0.0.0'
 
 jobs:
     npm-publish:
-        if: github.ref == 'refs/heads/master'
+        #if: github.ref == 'refs/heads/master'
```
bodinsamuel, line 14: to revert

PR #1917, opened 2024-03-28
[nan-644] remove utils dep for shared

khaliqgant:
## Describe your changes
remove dep on utils in shared for now


bodinsamuel (approved):
Looks good 👍🏻

PR #1918, opened 2024-03-28
[nan-644] pack dependency to install it

khaliqgant:
## Describe your changes
Sine the CLI uses shared, the dependencies that shared has need to be accessible. By using [bundleDependncies](https://docs.npmjs.com/cli/v10/configuring-npm/package-json#bundledependencies) we can pack the `utils` package and then shared will use that bundle to resolve the dependency. Unfortunately, it makes our publish script a bit more complicated.

1. Pack the `utils` package
2. Change the shared dependency in utils from `file:../utils` to the packed file

The other option is to _just_ publish utils...


packages/utils/package.json
```diff
@@ -5,6 +5,7 @@
     "type": "module",
     "main": "dist/index.js",
     "typings": "dist/index.d.ts",
+    "bundleDependencies": true,
```
khaliqgant, line 8: This is needed I believe to include the dependencies of utils, like `winston`

bodinsamuel (commented):
As I said earlier, not a fan because it add a problem on top of the problem, and we don't know the behavior of this feature. Also if we bundle `utils` why not bundling `shared` directly?
Tomorrow we will maybe split the DB or the SDK or some services from shared in their own packages and I'm sure we'll forget stuff. (I understand we need a fix but I'm not sure we need to rush)

### Alternatives I see:

1/ Even if it's shitty I would go the dumb solution of copy pasting stuff from shared to CLI and slowly moving back to something better. 

2/ Create a new packages that only contains what's needed for the CLI and that is imported in the CLI and Shared (and do not requires utils). Slightly more work but avoid most code duplication.

TBonnin:
> As I said earlier, not a fan because it add a problem on top of the problem, and we don't know the behavior of this feature. Also if we bundle `utils` why not bundling `shared` directly? Tomorrow we will maybe split the DB or the SDK or some services from shared in their own packages and I'm sure we'll forget stuff. (I understand we need a fix but I'm not sure we need to rush)
> ### Alternatives I see:
> 
> 1/ Even if it's shitty I would go the dumb solution of copy pasting stuff from shared to CLI and slowly moving back to something better.
> 
> 2/ Create a new packages that only contains what's needed for the CLI and that is imported in the CLI and Shared (and do not requires utils). Slightly more work but avoid most code duplication.

I think the end goal is to organized our codebase in smaller modules and avoid having a massive shared package. On top of that we probably don't want to publish all those internal modules. To achieve this we must be able to vendor those modules and bundled them into the packages that are using them. 
npm `BundleDependencies` seems to be a way to achieve this (even though I am annoyed that we have to pack the dependencies ourselves). If there is a better solution to vendor our internal modules I am happy to do it. @bodinsamuel 

As you mentioned @bodinsamuel, we can vendor `shared` directly into CLI but the idea behind this PR (vendoring utils) is a good first step that is necessary anyway if we want to have smaller internal modules

bodinsamuel:
> I think the end goal is to organized our codebase in smaller modules and avoid having a massive shared package

My point is that we will probably create plenty of small packages but CLI will still need almost all of them because `services` folder will need all of them anyway. So today, we add `utils` to the bundle but tomorrow we will need to add more and it won't solve anything for the CLI. 
Forgot to mention, I'm fine with publishing utils too but it has the same issue (= we'll need to publish all other packages until shared is entirely free of references).

---

The actual solution is to list everything the CLI import and make them "other packages free" then move everything else. So that's why I was saying copy pasting code or creating a new package that would reverse the move (easier said than done I agree).

We can go for this right now; but since the immediate issue was already solved by copy pasting the logger, then I don't think we need to rush for a second short-term fix.

TBonnin:
tl;dr: this PR objective is not to trim CLI dependencies. It is about introducing a method to bundle internal modules

This PR is not addressing the massive CLI dependencies problem I think. This PR starts addressing the challenge of packaging internal modules so we can start splitting the codebase (and shared in particular) into smaller modules. 

You are right the actual solution to deal with CLI dependencies is to list everything the CLI import and make them into their own package. Once we have those dependencies isolated into a/some smaller module(s) we will need to bundle them using the method introduced in this PR for example.

TBonnin (approved):

khaliqgant:
@bodinsamuel let me know if you want to discuss this more tomorrow. If not and you're OK with this, I'll test this by pushing a beta version with this publish logic to verify that it works as expected with the cli.

bodinsamuel (approved):
sure

khaliqgant:
Seems like this is actually blocked by https://github.com/npm/cli/issues/3466

bodinsamuel:
nb: I removed publish/verify from the required CI jobs to avoid blocking other PR

khaliqgant:
@bodinsamuel this [test](https://github.com/NangoHQ/nango/actions/runs/8680958310/job/23802673495?pr=1918) is failing due to the node v20 issue vs. 18 issue I mentioned the other day.

scripts/publish.sh
```diff
@@ -28,6 +28,17 @@ fi
 
 npm install
 
+# Utils
+mkdir -p "$GIT_ROOT_DIR/packages/shared/vendor"
+pushd "$GIT_ROOT_DIR/packages/utils"
+npm install --workspaces=false
```
khaliqgant, line 34: with workspaces=false this works with packing the utils package correctly as verified by https://github.com/NangoHQ/nango/actions/runs/8680958308/job/23802673428?pr=1918

TBonnin (approved):

packages/shared/package.json
```diff
@@ -9,6 +9,9 @@
         "build": "rimraf dist && tsc",
         "prepublishOnly": "npm install && npm run build"
     },
+    "bundleDependencies": [
+        "@nangohq/utils"
```
sfortowa, line 13: So if I need to add an other one I can just add it here?

sfortowa (approved):
Looks good, let's see 👌

bodinsamuel (approved):
Sorry I reviewed with my test account ahah 
Looks good :D

PR #1919, opened 2024-03-28
fix(utils): parse env with zod

bodinsamuel:
## Context

Fixes NAN-662

Currently we are using `process.env` everywhere but it's untyped and always a pain to deal with `any`. The idea with this function is that each service can call it when they launch and combine it with their own set of required env vars. 
For example in logs I need Elasticsearch env but I can't make it globally required. (I'm sure it will be handy for @TBonnin too)
You can find an usage example here: https://github.com/NangoHQ/nango/blob/1085b3149239da8c2bd050fffc1021e39c34f061/packages/logs/lib/env.ts#L1-L4

The main drawback is that it can't be used for floating constants like we have in `detection.ts`. I guess **we need to decide** wether it's fine to have this function run directly or make the detection constants also an object in a function. I will say it's a bit less convenient for sure, but with less side-effects
```ts
function initDetection(envs) {
  return { isCloud: envs.NANGO_CLOUD }
}
export const { isCloud } = initDetection(initGlobalEnv());
```

## Changes

- Expose a way to parse process.env with Zod (not exhaustive)

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -74,8 +74,5 @@ DEFAULT_GITHUB_CLIENT_SECRET=
 WORKOS_API_KEY=
 WORKOS_CLIENT_ID=
 
-# Encryption key for secrets / records stored in database
-NANGO_ENCRYPTION_KEY=
-
 # Enable or disable environment keys cache
 NANGO_CACHE_ENV_KEYS=
--- a/.env.example
+++ b/.env.example
@@ -76,3 +76,6 @@ WORKOS_CLIENT_ID=
 
 # Enable or disable environment keys cache
 NANGO_CACHE_ENV_KEYS=
+
+# Key to send email
+MAILGUN_API_KEY=
--- a/package-lock.json
+++ b/package-lock.json
@@ -34222,7 +34222,8 @@
             "version": "1.0.0",
             "license": "SEE LICENSE IN LICENSE FILE IN GIT REPOSITORY",
             "dependencies": {
-                "winston": "3.8.2"
+                "winston": "3.8.2",
+                "zod": "3.22.4"
             },
             "devDependencies": {}
         },
--- a/package-lock.json
+++ b/package-lock.json
@@ -7120,12 +7120,38 @@
                 "linux"
             ]
         },
+        "node_modules/@rollup/rollup-linux-powerpc64le-gnu": {
+            "version": "4.13.2",
+            "resolved": "https://registry.npmjs.org/@rollup/rollup-linux-powerpc64le-gnu/-/rollup-linux-powerpc64le-gnu-4.13.2.tgz",
+            "integrity": "sha512-zvXvAUGGEYi6tYhcDmb9wlOckVbuD+7z3mzInCSTACJ4DQrdSLPNUeDIcAQW39M3q6PDquqLWu7pnO39uSMRzQ==",
+            "cpu": [
+                "ppc64le"
+            ],
+            "dev": true,
+            "optional": true,
+            "os": [
+                "linux"
+            ]
+        },
         "node_modules/@rollup/rollup-linux-riscv64-gnu": {
             "version": "4.13.0",
             "resolved": "https://registry.npmjs.org/@rollup/rollup-linux-riscv64-gnu/-/rollup-linux-riscv64-gnu-4.13.0.tgz",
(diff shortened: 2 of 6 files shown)


.env.example
```diff
@@ -74,8 +74,5 @@ DEFAULT_GITHUB_CLIENT_SECRET=
 WORKOS_API_KEY=
 WORKOS_CLIENT_ID=
 
-# Encryption key for secrets / records stored in database
-NANGO_ENCRYPTION_KEY=
-
```
bodinsamuel, line 82: I didn't saw it was already in the file at the beginning

TBonnin:
It is nice from a typing point of view but doesn't address the required env vars problem. Would it make more sense to have each package do their own validation to ensure the required vars are defined. It is not this much code

bodinsamuel:
> It is nice from a typing point of view but doesn't address the required env vars problem. Would it make more sense to have each package do their own validation to ensure the required vars are defined. It is not this much code

Agreed that what I said in the PR desc, right now I have listed mostly optional env, except maybe the DB env. Im trying to avoid having multiple sources of truth and implementation, but I think I have found a solution that could work.

```ts
const ENVS = {
    // Database
    NANGO_DATABASE_URL: z.string().url().optional(),
    NANGO_DB_PORT: z.coerce.number().default(5432),
    NANGO_ENCRYPTION_KEY: z.string().optional()
};
type EnvKeys = keyof typeof ENVS;

export function initEnvs<const TKey extends Partial<Record<EnvKeys, true>>>(required: TKey) {
    const tmp = { ...ENVS };

    const schema = z.object(tmp).required(required);

    const res = schema.safeParse(process.env);
    if (!res.success) {
        throw new Error(`Missing or invalid env vars: ${envErrorToString(res.error.issues)}`);
    }

    return res.data;
}
```
```ts
# usage
const envs = initEnvs({ NANGO_ENCRYPTION_KEY: true });
```

wdyt?

TBonnin:
Looks good. what about the param to be `initEnvs(option: {required: string[]} = { required: []})` so using it doesn't require the zod syntax with the unecessary true: `initEnvs({required: ['NANGO_ENCRYPTION_KEY', ...]})`. Or a straight up array.

bodinsamuel:
The syntax is a bit annoying but doesn't require anything else; the alternative requires a bit more typescript magic. I'm okay with both 

```ts
export function initGlobalEnv<const TKey extends EnvKeys[]>(required: TKey) {
    const tmp = { ...ENVS };

    const schema = z.object(tmp).required(
        required.reduce(
            (prev, curr) => {
                prev[curr as TKey[0]] = true;
                return prev;
            },
            {} as Record<TKey[0], true>
        )
    );

    const res = schema.safeParse(process.env);
    if (!res.success) {
        throw new Error(`Missing or invalid env vars: ${envErrorToString(res.error.issues)}`);
    }

    return res.data;
}

const envs = initGlobalEnv(['NANGO_DATABASE_URL']);
envs.NANGO_ENCRYPTION_KEY;
envs.NANGO_DATABASE_URL;
```

TBonnin:
> The syntax is a bit annoying but doesn't require anything else; the alternative requires a bit more typescript magic.
I imagine. I proposed the object with the required field so it is clearer that all envs are being loaded with the ones being specified as being required.

At the end it is up to you. I would even propose a more radical interface where you would load the env vars as follow 
```
Envs().load()
// or
Envs().required(['a', 'b']).load()
```

I guess the implementation would look like this

```
function Envs(): EnvsLoader {
    return new EnvsLoader()
}

class EnvsLoader {
   private schema = z.object({
        NANGO_DATABASE_URL: ...,
        NANGO_DB_PORT: ...,
        NANGO_ENCRYPTION_KEY: ...
        ...
    })

    constructor() {}

    public required(required: string[]): EnvsLoader {
        // ...
        this.schema.required(...)
        return this;
    }

    public load() {
        const res = schema.safeParse(process.env)
        ...
        return data;
    }
}
```

As I said, up to you. All you have proposed is good imho

bodinsamuel:
So after all missing env vars in the schema, Typescript was not happy about the types, it was too big for the compilation. Not entirely sure why so I changed a bit the API, something more simple but that works. We can improve later imo

TBonnin (approved):

PR #1921, opened 2024-03-28
fix(ui): sync cookie and url

bodinsamuel:
## Describe your changes

Fixes NAN-568

When accessing an env by URL that was different than the cookie, the cookie would take over. But it's deceptive/dangerous if someone is sharing an URL and you start modifying stuff on the wrong env. 

- Change `/meta` to not require the cookie because we need the list of correct available envs
- Sync the URL with the cookie (the URL always win)
- If the env is incorrect is not available in `meta.envs` then we change the cookie and display a 404
We don't redirect because it would still be deceptive for the user.
 
- Other fixes related to type and simplification

## Test

https://github.com/NangoHQ/nango/assets/1637651/b809f196-dff3-43fd-a156-7d0febd72748


packages/webapp/src/components/PrivateRoute.tsx
```diff
@@ -1,19 +1,33 @@
...
     const setStoredEnvs = useStore((state) => state.setEnvs);
     const setBaseUrl = useStore((state) => state.setBaseUrl);
     const setEmail = useStore((state) => state.setEmail);
     const setDebugMode = useStore((state) => state.setDebugMode);
+    const setCookieValue = useStore((state) => state.setCookieValue);
+
+    useEffect(() => {
+        // sync path with cookie
+        const pathSplit = location.pathname.split('/');
+        if (env !== pathSplit[1]) {
```
khaliqgant: logically I know that there will always be an element at index position 1, but this kind of access still makes me nervous

khaliqgant (approved):
Non blocking comment inline

packages/webapp/src/components/PrivateRoute.tsx
bodinsamuel: hehe yes I can add some safety

bodinsamuel:
There was an issue in staging, trying to fix it before merging

bodinsamuel:
Working in staging, I'll let you recheck PrivateRoute.tsx 👍🏻

bodinsamuel:
Can I have a review on this @khaliqgant @TBonnin please? ☺️

khaliqgant (approved):

PR #1923, opened 2024-03-28
[nan-649] redirect on error

khaliqgant:
## Describe your changes
Redirect the user if they cancel the google auth

NAN-649


TBonnin:
what if a real problem happens? They will just be redirected without have any info. no?

bodinsamuel (approved):

khaliqgant:
> what if a real problem happens? They will just be redirected without have any info. no?

Thanks, this is OK for now from a product perspective (spoke with Bastien).

PR #1926, opened 2024-03-29
[nan-660] add personio support using API and OAuth client credentials

khaliqgant:
## Describe your changes
Add support as a rather urgent client request

NAN-660


packages/shared/lib/models/Auth.ts
```diff
@@ -136,6 +137,16 @@ export interface OAuth2Credentials extends CredentialsCommon {
...
 
+export interface OAuth2ClientCredentials extends CredentialsCommon {
+    type: AuthModes.OAuth2CC;
+    token: string;
+
+    expires_at?: Date | undefined;
+
+    client_id: string;
+    client_secret: string;
+}
```
bodinsamuel, line 148: This honestly looks like the `config_override` you just did 😅

packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -82,7 +82,7 @@ class OAuthController {
 
         try {
             if (!wsClientId) {
-                analytics.track(AnalyticsTypes.PRE_WS_OAUTH, accountId);
+                analytics.track(AnalyticsTypes.PRE_OAUTH2_CC_AUTH, accountId);
```
bodinsamuel: seems not related to oauth2_cc but regular oauth

bodinsamuel (approved):
I must say I don't get the difference between regular oauth and this one. The obscure name doesn't help :D 
If you can provide more context that would be great

packages/server/lib/controllers/oauth.controller.ts
khaliqgant: Yup. Leftover from a different direction I started to take. Will fix. Thanks!

docs-v2/integrations/all/personio.mdx
```diff
@@ -0,0 +1,34 @@
...
+-   [How to register an Application](https://developer.personio.de/docs/getting-started-with-the-personio-api)
+-   [OAuth-related docs](https://developer.personio.de/reference/auth)
+-   [Web API docs (their REST API)](https://developer.personio.de/reference/introduction)
+
+<Tip>Need help getting started? Get help in the [community](https://nango.dev/slack).</Tip>
+
+## API gotchas
+
+-   Personio offers a separate recruiting API that is API key based and is listed under `personio-recruiting`
+-   The values for the Partner-ID Personio-App can be defined by you, but should represent your company name and application name. For example `MY_APPLICATION`, `ACME` are valid values.
```
bastienbeurier, line 32: Probably worth mentioning that there's 2 APIs, with different auth types, and that one of them is the client credentials flow that authorize an app vs. a user.

khaliqgant:
> I must say I don't get the difference between regular oauth and this one. The obscure name doesn't help :D If you can provide more context that would be great

OAuth_CC stands for Client Credentials: https://auth0.com/docs/get-started/authentication-and-authorization-flow/client-credentials-flow

PR #1927, opened 2024-03-29
[nan-591] lookup account and env name

khaliqgant:
## Describe your changes
Add in additional props

NAN-591


khaliqgant:
Note that the schema in big query needs to be updated

bodinsamuel (approved):

PR #1928, opened 2024-03-29
fix(eslint): pass #9

bodinsamuel:
## Describe your changes

- Run automated fix `npx eslint --fix`
- Added `.js` on type import (unfortunately it's a known limitation from the package)
- Sync our usage of `ms` (types conflict)
- Bonus: naming the testcontainers was nice but it can conflict if you relaunch the process very fast


packages/webapp/package.json
```diff
@@ -53,7 +53,7 @@
         "env-cmd": "10.1.0",
         "js-cookie": "3.0.5",
         "lodash": "4.17.21",
-        "ms": "2.1.3",
+        "ms": "3.0.0-canary.1",
```
TBonnin, line 56: canary?

tests/setup.ts
```diff
@@ -13,7 +14,7 @@ async function setupPostgres() {
         .withUsername(user)
         .withPassword(password)
         .withExposedPorts(5432)
-        .withName('pg-test')
+        .withName(`pg-test-${randomUUID()}`)
```
TBonnin, line 17: I ran into this one last week. Thank you for fixing

TBonnin (approved):

packages/webapp/package.json
bodinsamuel: it's the only version published that has types, we already use it in `shared`

PR #1930, opened 2024-04-01
[nan-668] add verification bash script to publish packages

khaliqgant:
## Describe your changes
The goal with this PR is to mimic a publish process of the cli with the code that is in review to ensure a working cli that is in a remote source. This is accomplished by building and publishing the packages to the Github registry and finally using the cli to verify that it works expected. This will protect against instances where a faulty cli is published. This will be especially useful to verify that this PR: https://github.com/NangoHQ/nango/pull/1918 works as expected with packaging the utils package. Additionally, as we refactor the cli to move code out of shared and reduce the package size of the cli this will ensure things still work as expected.

NAN-668

### DIFF
--- a/.github/workflows/cli-verification.yaml
+++ b/.github/workflows/cli-verification.yaml
@@ -45,9 +45,10 @@ jobs:
               env:
                   NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
               run: |
-                  npx nango version
                   GIT_HASH=$(git rev-parse HEAD)
-                  cd ../ && mkdir nango-cli-test && cd nango-cli-test
+                  mkdir nango-cli-test && cd nango-cli-test
                   npm init -y
                   npm -g install @nangohq/cli@0.0.1-$GIT_HASH
                   nango version --debug
+                  nango init --debug
+                  nango generate --debug
--- a/.github/workflows/cli-verification.yaml
+++ b/.github/workflows/cli-verification.yaml
@@ -41,7 +41,9 @@ jobs:
                   NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
               run: |
                   GIT_HASH=$(git rev-parse HEAD)
+                  echo "attempting install"
                   npm install @nangohq/shared@0.0.1-$GIT_HASH
+                  echo "building"
                   npm run build
                   npm version 0.0.1-$GIT_HASH --workspaces-update=false
                   npm publish --ignore-scripts
--- a/.github/workflows/cli-verification.yaml
+++ b/.github/workflows/cli-verification.yaml
@@ -32,16 +32,7 @@ jobs:
               shell: bash
               run: |
                   GIT_HASH=$(git rev-parse --short HEAD)
-                  npm run build:hosted
-                  # publish all before bumping to an alternate version
-                  npm publish -w packages/node-client
-                  npm i
-                  npm publish -w packages/shared
-                  npm i
-                  npm publish -w packages/cli
-                  npm i
-                  npm publish -w packages/frontend
-                  bash ./scripts/publish.sh 0.0.1-$GIT_HASH true
+                  bash ./scripts/publish.sh 0.0.1-$GIT_HASH
             - name: Install the cli from the github package registry
               env:
                   NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/cli-verification.yaml
+++ b/.github/workflows/cli-verification.yaml
@@ -31,12 +31,12 @@ jobs:
                   NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
               shell: bash
               run: |
-                  GIT_HASH=$(git rev-parse --short HEAD)
+                  GIT_HASH=$(git rev-parse HEAD)
                   bash ./scripts/publish.sh 0.0.1-$GIT_HASH
             - name: Install the cli from the github package registry
               env:
                   NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
               run: |
-                  GIT_HASH=$(git rev-parse --short HEAD)
+                  GIT_HASH=$(git rev-parse HEAD)
                   npm install nango@0.0.1-$GIT_HASH
                   npx nango version --debug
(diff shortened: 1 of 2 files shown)


.github/workflows/cli-verification.yaml
```diff
@@ -0,0 +1,41 @@
...
+                  node-version: '18.x'
+                  registry-url: 'https://npm.pkg.github.com'
+                  scope: '@nangohq'
+            - name: Publish npm packages to the github registry
+              env:
+                  NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+              shell: bash
+              run: |
+                  GIT_HASH=$(git rev-parse --short HEAD)
+                  bash ./scripts/publish.sh 0.0.1-$GIT_HASH true
```
TBonnin: I would personally prefer the script to handle this very specific version rather than having a bypass mechanism that make it possible to misuse the script.

scripts/publish.sh
```diff
@@ -14,23 +14,29 @@ function bump_and_npm_publish {
...
 VERSION=$1
+BY_PASS_VERSION_CHECK=$2
 
 # ensure version is of format x.y.z
-if [[ ! "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
-    echo "VERSION '$VERSION' is not of format x.y.z"
-    exit 1
+if [[ "$BY_PASS_VERSION_CHECK" != "true" ]]; then
+    # ensure version is of format x.y.z
+    if [[ ! "$VERSION" =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
```
TBonnin: See my comment above. I would use the following regex `^([0-9]+\.[0-9]+\.[0-9]+|0\.0\.1-[0-9a-fA-F]{40})$` and not have a bypass

.github/workflows/cli-verification.yaml
bodinsamuel: same ☝🏻 but I'm also wondering why we publish and then call the script? Seems dangerous or unnecessary

.github/workflows/cli-verification.yaml
```diff
@@ -0,0 +1,42 @@
...
+              run: |
+                  GIT_HASH=$(git rev-parse HEAD)
+                  bash ./scripts/publish.sh 0.0.1-$GIT_HASH
+            - name: Install the cli from the github package registry
+              env:
+                  NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+              run: |
+                  GIT_HASH=$(git rev-parse HEAD)
+                  npm install nango@0.0.1-$GIT_HASH
+                  npx nango version --debug
```
bodinsamuel, line 57: If the goal is to test the cli I think I would prefer we don't publish each commit to npm, 
`build + npm install -g packages/cli` should work no?

khaliqgant: Thanks. The goal is to check the packaging of the cli package to a remote source -- in this case github registry. By packaging I mean the changes in https://github.com/NangoHQ/nango/pull/1918

If I install the cli via `npm install -g packages/cli` I don't think that'll mimic the same process as I would like to test that npm is doing when we upload a packed dependency.

khaliqgant: Added more info to the PR description

.github/workflows/cli-verification.yaml
```diff
@@ -0,0 +1,57 @@
...
+              run: |
+                  GIT_HASH=$(git rev-parse HEAD)
+                  bash ./scripts/publish.sh 0.0.1-$GIT_HASH
+            - name: Publish the cli privately under the correct scope
+              working-directory: packages/cli
+              env:
+                  NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+              run: |
+                  jq '.name = "@nangohq/cli"' package.json > temp.json && mv temp.json package.json
+                  npm publish --access public
```
TBonnin, line 45: isn't the cli published by running the `publish.sh` script in the previous step? why do we need to publish it again?

khaliqgant: It attempts to publish it but fails because it is not scoped under our organization `@nangohq` which is why I rename the package and publish it separately. 

https://github.com/NangoHQ/nango/blob/f9448fcca5d5d6da60630536f2633b4c7a523800/.github/workflows/cli-verification.yaml#L39-L45

.github/workflows/cli-verification.yaml
```diff
@@ -0,0 +1,57 @@
...
+            - name: Install the cli from the github package registry
+              env:
+                  NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+              run: |
+                  GIT_HASH=$(git rev-parse HEAD)
+                  mkdir nango-cli-test && cd nango-cli-test
+                  npm init -y
+                  npm install @nangohq/cli@0.0.1-$GIT_HASH
+                  npx nango version --debug
```
bodinsamuel, line 57: I would at least test the output

.github/workflows/cli-verification.yaml
bodinsamuel: why not doing that directly then?
Also doesn't it will publish @nangohq/node and @nangohq/shared every time for nothing?

khaliqgant: > why not doing that directly then?

Easier to reuse the script which bumps the references and packages correctly.

> Also doesn't it will publish @nangohq/node and @nangohq/shared every time for nothing?

Shared needs to be published, node is published again and not used as is frontend but there is no harm in publishing those packages.

.github/workflows/cli-verification.yaml
khaliqgant: Don't see much gain there

bodinsamuel: if you don't test how can you be sure it's the right version being tested? also not all errors will throw process.exit(1)

khaliqgant: > also not all errors will throw process.exit(1)

Is fair. Updated with https://github.com/NangoHQ/nango/pull/1930/commits/9706481358d8fa15231bbe2004a2525421087e68

bodinsamuel (approved):

TBonnin:
It is a bit of an hacky solution but since it is contained in a single github actions I am fine with it. We can improve it in the future if the need arises

khaliqgant:
> It is a bit of hacky solution but since it is contained in a single github actions I am fine with it. We can improve it in the future if the need arises

Agree it is a bit _(read: very)_ hacky but also agree if it causes problems we can improve or rip it out even. The hope is that it saves us in the future. Time will tell if that is actually the case.

PR #1933, opened 2024-04-01
add records package (records migration)

TBonnin:
## Describe your changes

Adding the `records`package to interact with nango users records. This package introduces a new table which will be hosted on a different database.
For now the package is not being used, other than running the db migration. Next step will be to use this package to double write the records in both the old table and the new one.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -18,6 +18,7 @@ COPY packages/shared/package.json ./packages/shared/package.json
 COPY packages/webapp/package.json ./packages/webapp/package.json
 COPY packages/data-ingestion/package.json ./packages/data-ingestion/package.json
 COPY packages/logs/package.json ./packages/logs/package.json
+COPY packages/records/package.json ./packages/records/package.json
 COPY package*.json  ./
 
 # Install every dependencies
--- a/packages/records/lib/db/migrations/20240327151027_create_records_table.ts
+++ b/packages/records/lib/db/migrations/20240327151027_create_records_table.ts
@@ -1,0 +1,95 @@
+import type { Knex } from 'knex';
+
+const TABLE = 'records';
+const PARTITION_COUNT = 256;
+
+function partitionTable(i: number): string {
+    return `${TABLE}_p${i}`;
+}
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.transaction(async (trx) => {
+        // TABLE
+        await trx.raw(`
+            CREATE TABLE "${TABLE}" (
+            id uuid NOT NULL,
+            external_id character varying(255) NOT NULL,
+            json jsonb,
+            data_hash character varying(255) NOT NULL,
+            connection_id integer NOT NULL,
+            model character varying(255) NOT NULL,
+            created_at timestamp with time zone NOT NULL DEFAULT CURRENT_TIMESTAMP,
+            updated_at timestamp with time zone NOT NULL DEFAULT CURRENT_TIMESTAMP,
+            deleted_at timestamp with time zone,
+            sync_id uuid,
+            sync_job_id integer,
+            CONSTRAINT ${TABLE}_connection_id_external_id_model UNIQUE (connection_id, external_id, model)
+            ) PARTITION BY HASH (connection_id, model)
+        `);
+        for (let i = 0; i < PARTITION_COUNT; i++) {
+            await trx.raw(`
+                CREATE TABLE "${partitionTable(i)}" PARTITION OF "${TABLE}"
+                FOR VALUES WITH (MODULUS ${PARTITION_COUNT}, REMAINDER ${i});
+            `);
+        }
+        // TRIGGERS
+        await trx.raw(`
+             CREATE OR REPLACE FUNCTION ${TABLE}_undelete()
+             RETURNS TRIGGER AS $$
+             BEGIN
+                 IF OLD.deleted_at IS NOT NULL AND NEW.deleted_at IS NULL THEN
+                     NEW.created_at = NOW();
+                     NEW.updated_at = NOW();
+                 END IF;
+                 RETURN NEW;
+             END;
+             $$ LANGUAGE plpgsql;
+        `);
+        await trx.raw(`
+                     CREATE TRIGGER ${TABLE}_undelete_trigger
+                     BEFORE UPDATE ON ${TABLE}
+                     FOR EACH ROW
+                     EXECUTE FUNCTION ${TABLE}_undelete();
+        `);
+        await knex.raw(`
+            CREATE OR REPLACE FUNCTION ${TABLE}_updated_at()
+            RETURNS TRIGGER AS $$
+            BEGIN
+                IF OLD.data_hash IS DISTINCT FROM NEW.data_hash THEN
+                    NEW.updated_at = NOW();
+                END IF;
+
+                RETURN NEW;
+            END;
+          $$ LANGUAGE plpgsql;
+        `);
+        await knex.raw(`
+            CREATE TRIGGER ${TABLE}_updated_at_trigger
+            BEFORE UPDATE ON ${TABLE}
+            FOR EACH ROW
+            EXECUTE FUNCTION ${TABLE}_updated_at();
+        `);
+        // INDEXES
+        await knex.schema.alterTable(TABLE, function (table) {
+            table.index(['connection_id', 'model', 'external_id']);
+            table.index(['connection_id', 'model', 'updated_at', 'id']);
+            table.index('sync_id');
+            table.index('sync_job_id');
+            table.unique(['connection_id', 'model', 'id']);
+        });
+    });
+}
+
+export async function down(knex: Knex): Promise<void> {
+    // TABLE
+    // INDEXES are dropped automatically
+    await knex.raw(`DROP TABLE IF EXISTS "${TABLE}"`);
+    for (let i = 0; i < PARTITION_COUNT; i++) {
+        await knex.raw(`DROP TABLE IF EXISTS "${partitionTable(i)}"`);
+    }
+    // TRIGGERS
+    await knex.raw(`DROP TRIGGER IF EXISTS ${TABLE}_undelete_trigger ON ${TABLE};`);
+    await knex.raw(`DROP FUNCTION IF EXISTS ${TABLE}_undelete();`);
+    await knex.raw(`DROP TRIGGER IF EXISTS ${TABLE}_updated_at_trigger ON ${TABLE};`);
+    await knex.raw(`DROP FUNCTION IF EXISTS ${TABLE}_updated_at();`);
+}
--- a/packages/records/lib/env.ts
+++ b/packages/records/lib/env.ts
@@ -1,0 +1,8 @@
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { ENVS, parseEnvs } from '@nangohq/utils/dist/environment/parse.js';
+
+export const envs = parseEnvs(ENVS.required({ RECORDS_DATABASE_URL: true }));
+
+export const filename = fileURLToPath(import.meta.url);
+export const dirname = path.dirname(path.join(filename, '../../'));
--- a/packages/records/lib/helpers/uniqueKey.ts
+++ b/packages/records/lib/helpers/uniqueKey.ts
@@ -1,0 +1,43 @@
+import type { FormattedRecord } from '../types.js';
+
+export function getUniqueId(record: FormattedRecord): string {
+    return record.external_id;
+}
+
+export function verifyUniqueKeysAreUnique(records: FormattedRecord[]): { nonUniqueKeys: Set<string> } {
+    const idMap = new Set<string>();
+    const nonUniqueKeys = new Set<string>();
+
+    for (const record of records) {
+        const id = getUniqueId(record);
+        if (idMap.has(id)) {
+            nonUniqueKeys.add(id);
+        } else {
+            idMap.add(id);
+        }
+    }
+
+    return { nonUniqueKeys };
+}
+
+export function removeDuplicateKey(records: FormattedRecord[]): { records: FormattedRecord[]; nonUniqueKeys: string[] } {
+    // TODO:
+    // for (const nonUniqueKey of nonUniqueKeys) {
+    //     await createActivityLogMessage({
+    //         level: 'error',
+    //         environment_id,
+    //         activity_log_id: activityLogId,
+    //         content: `There was a duplicate key found: ${nonUniqueKey}. This record will be ignore in relation to the model ${model}.`,
+    //         timestamp: Date.now()
+    //     });
+    // }
+
+    const { nonUniqueKeys } = verifyUniqueKeysAreUnique(records);
+    const seen = new Set();
+    const recordsWithoutDuplicates = records.filter((record) => {
+        const key = getUniqueId(record);
+        return seen.has(key) ? false : seen.add(key);
+    });
+
+    return { records: recordsWithoutDuplicates, nonUniqueKeys: Array.from(nonUniqueKeys) };
+}
--- a/packages/records/lib/index.ts
+++ b/packages/records/lib/index.ts
@@ -1,0 +1,2 @@
+export * from './db/migrate.js';
+export * as records from './models/records.js';
--- a/packages/records/lib/models/records.integration.test.ts
+++ b/packages/records/lib/models/records.integration.test.ts
@@ -1,0 +1,175 @@
+import { expect, describe, it, beforeAll, afterEach } from 'vitest';
+import dayjs from 'dayjs';
+import * as uuid from 'uuid';
+import { migrate } from '../db/migrate.js';
+import { RECORDS_TABLE } from '../constants.js';
+import { db } from '../db/client.js';
+import * as Records from '../models/records.js';
+import { formatRecords } from '../helpers/format.js';
+import type { UnencryptedRecordData } from '../types.js';
+import type { UpsertSummary } from '../types.js';
+import { isErr } from '@nangohq/shared';
+
+describe('Records service', () => {
+    beforeAll(async () => {
+        await migrate();
+    });
+
+    afterEach(async () => {
+        await db(RECORDS_TABLE).truncate();
+    });
+
+    it('Should write records', async () => {
+        const connectionId = 1;
+        const model = 'my-model';
+        const syncId = '00000000-0000-0000-0000-000000000000';
+        const records = [
+            { id: '1', name: 'John Doe' },
+            { id: '1', name: 'John Doe' },
+            { id: '2', name: 'Jane Doe' },
+            { id: '3', name: 'Max Doe' },
+            { id: '4', name: 'Mike Doe' }
+        ];
+        const res = await upsertRecords(records, connectionId, model, syncId, 1);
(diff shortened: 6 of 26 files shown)


packages/records/lib/db/migrations/20240327151027_create_records_table.ts
```diff
@@ -0,0 +1,104 @@
...
+            EXECUTE FUNCTION ${TABLE}_updated_at();
+        `);
+        // INDEXES
+        await knex.schema.alterTable(TABLE, function (table) {
+            table.index(['connection_id', 'model']);
+            table.index(['connection_id', 'model', 'external_id']);
+            table.index(['connection_id', 'model', 'updated_at']);
+            table.index('sync_id');
+            table.index('sync_job_id');
+            table.unique(['connection_id', 'model', 'id']);
```
TBonnin, line 78: @khaliqgant @bodinsamuel can you double check that I have not forgotten required indexes. 
Note: the `records` package doesn't support the legacy functions to fetch records.

packages/records/lib/models/records.integration.test.ts
```diff
@@ -0,0 +1,179 @@
...
+import dayjs from 'dayjs';
+import * as uuid from 'uuid';
+import { migrate } from '../db/migrate.js';
+import { RECORDS_TABLE } from '../constants.js';
+import { db } from '../db/client.js';
+import * as Records from '../models/records.js';
+import { formatRecords } from '../helpers/format.js';
+import type { UnencryptedRecordData } from '../types.js';
+import type { UpsertSummary } from '../types.js';
+import { isErr } from '@nangohq/shared';
```
TBonnin, line 11: my goal is to move `Result` to `@nangohq/utils` so records doesn't have to depend on `shared`

packages/records/lib/models/records.ts
```diff
@@ -0,0 +1,417 @@
...
+    ReturnedRecord,
+    UnencryptedRecord,
+    UpsertSummary
+} from '../types.js';
+import { decryptRecordData, decryptRecords, encryptDataRecords } from '../utils/encryption.js';
+import { RECORDS_TABLE } from '../constants.js';
+import { removeDuplicateKey, getUniqueId } from '../helpers/uniqueKey.js';
+import { logger } from '../utils/logger.js';
+import { resultErr, resultOk } from '@nangohq/shared';
+import type { Result } from '@nangohq/shared';
```
TBonnin, line 18: my goal is to move `Result` to `@nangohq/utils` so records doesn't have to depend on `shared`

packages/records/lib/models/records.ts
```diff
@@ -0,0 +1,417 @@
+import dayjs from 'dayjs';
```
TBonnin, line 1: `models/records.ts` contains all the functions that are reading/writing from/to the records table

packages/records/lib/vitest.d.ts
```diff
@@ -0,0 +1,14 @@
...
+    toBeIsoDate: () => TR;
+    toBeIsoDateTimezone: () => TR;
+    toBeUUID: () => TR;
+}
+
+declare module 'vitest' {
+    export interface Assertion<T = any> extends CustomMatchers<T> {}
+    export interface AsymmetricMatchersContaining extends CustomMatchers {}
+    export interface ExpectStatic<T = any> extends CustomMatchers<T> {}
+}
```
TBonnin, line 14: Tried to use the one from `shared` but couldn't make it work. @bodinsamuel I saw you had the same issue with the `logs` package.

packages/shared/lib/utils/encryption.manager.ts
```diff
@@ -35,19 +35,19 @@ class EncryptionManager {
...
+    public encrypt(str: string): [string, string, string] {
         const iv = crypto.randomBytes(12);
         const cipher = crypto.createCipheriv(this.algo, Buffer.from(this.key!, this.encoding), iv);
         let enc = cipher.update(str, 'utf8', this.encoding);
         enc += cipher.final(this.encoding);
         return [enc, iv.toString(this.encoding), cipher.getAuthTag().toString(this.encoding)];
     }
 
-    private decrypt(enc: string, iv: string, authTag: string): string {
+    public decrypt(enc: string, iv: string, authTag: string): string {
```
TBonnin, line 50: I am planning to refactor encryption manager so it only exposes the `encrypt` and `decrypt` primitives and the logic to encrypt/decrypt environment/connection/records live in their respective package/service, like it is done in this PR

Dockerfile
```diff
@@ -17,6 +17,7 @@ COPY packages/server/package.json ./packages/server/package.json
 COPY packages/shared/package.json ./packages/shared/package.json
 COPY packages/webapp/package.json ./packages/webapp/package.json
 COPY packages/data-ingestion/package.json ./packages/data-ingestion/package.json
+COPY packages/records/package.json ./packages/shared/package.json
```
khaliqgant:
```suggestion
COPY packages/records/package.json ./packages/records/package.json
```

packages/records/lib/models/records.ts
```diff
@@ -0,0 +1,417 @@
...
+import { removeDuplicateKey, getUniqueId } from '../helpers/uniqueKey.js';
+import { logger } from '../utils/logger.js';
+import { resultErr, resultOk } from '@nangohq/shared';
+import type { Result } from '@nangohq/shared';
+
+dayjs.extend(utc);
+
+const BATCH_SIZE = 1000;
+
+export async function getRecords(
```
khaliqgant: Agree with feedback that @bodinsamuel made in another PR that maybe an object is easier to reason about and manage esp on internal only calls like this.

packages/records/tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
+    "extends": "../../tsconfig.json",
+    "compilerOptions": {
+        "rootDir": "lib",
+        "outDir": "dist"
+    },
+    "references": [
+        {
+            "path": "../shared"
```
khaliqgant, line 9: add utils

packages/server/package.json
```diff
@@ -25,6 +25,7 @@
         "@hapi/boom": "^10.0.1",
         "@nangohq/shared": "^0.39.15",
         "@nangohq/utils": "file:../utils",
+        "@nangohq/records": "file:../records",
```
khaliqgant, line 28: server tsconfig also needs updating

packages/records/lib/models/records.ts
TBonnin: I don't remember @bodinsamuel comment. Can you refresh my memory? :D

khaliqgant: Sorry, it wasn't made in this PR, but rather here: https://github.com/NangoHQ/nango/pull/1903#discussion_r1541202375

TBonnin: ah got it you mean for the parameters. Yes that make sense

packages/records/lib/db/migrations/20240327151027_create_records_table.ts
```diff
@@ -0,0 +1,104 @@
...
+        `);
+        await knex.raw(`
+            CREATE TRIGGER ${TABLE}_updated_at_trigger
+            BEFORE UPDATE ON ${TABLE}
+            FOR EACH ROW
+            EXECUTE FUNCTION ${TABLE}_updated_at();
+        `);
+        // INDEXES
+        await knex.schema.alterTable(TABLE, function (table) {
+            table.index(['connection_id', 'model']);
```
bodinsamuel: I think this index is unnecessary since it's already covered by the two after

packages/records/lib/env.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const schema = z.object({
+    RECORDS_DB_HOST: z.string().optional().default('localhost'),
+    RECORDS_DB_PORT: z.coerce.number().optional().default(5432),
+    RECORDS_DB_USER: z.string().optional().default('nango'),
+    RECORDS_DB_NAME: z.string().optional().default('nango'),
+    RECORDS_DB_PASSWORD: z.string().optional().default('nango'),
+    RECORDS_DB_SSL: z.enum(['true', 'false']).optional().default('false')
+});
+export const envs = schema.parse(process.env);
```
bodinsamuel: you can use the utils env now that it's merged ☺️

packages/records/lib/env.ts
```diff
@@ -0,0 +1,16 @@
...
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const schema = z.object({
+    RECORDS_DB_HOST: z.string().optional().default('localhost'),
+    RECORDS_DB_PORT: z.coerce.number().optional().default(5432),
+    RECORDS_DB_USER: z.string().optional().default('nango'),
+    RECORDS_DB_NAME: z.string().optional().default('nango'),
+    RECORDS_DB_PASSWORD: z.string().optional().default('nango'),
+    RECORDS_DB_SSL: z.enum(['true', 'false']).optional().default('false')
```
bodinsamuel: imo we are using those many env vars for the main db but an URL is enough

packages/records/lib/vitest.d.ts
bodinsamuel: yeah painful, I don't understand why

packages/records/lib/helpers/uniqueKey.ts
```diff
@@ -0,0 +1,48 @@
...
+    // TODO:
+    // for (const nonUniqueKey of nonUniqueKeys) {
+    //     await createActivityLogMessage({
+    //         level: 'error',
+    //         environment_id,
+    //         activity_log_id: activityLogId,
+    //         content: `There was a duplicate key found: ${nonUniqueKey}. This record will be ignore in relation to the model ${model}.`,
+    //         timestamp: Date.now()
+    //     });
+    // }
```
bodinsamuel, line 33: imo we currently put activity logs everywhere but they should not be in helpers

packages/records/lib/helpers/uniqueKey.ts
```diff
@@ -0,0 +1,48 @@
...
+            if (!nonUniqueKeys.includes(id.toString())) {
+                nonUniqueKeys.push(id.toString());
+            }
+        } else {
+            idMap[id] = true;
+        }
+    }
+
+    return { isUnique, nonUniqueKeys };
+}
```
bodinsamuel: Seems like `isUnique` is never used in the rest of the codebase. Also you could use Set for cleaner/faster dedup (not suggesting change)

```ts
export function verifyUniqueKeysAreUnique(records: FormattedRecord[]): { nonUniqueKeys: Set<string> } {
    const idMap = new Set<string>();
    const nonUniqueKeys = new Set<string>();

    for (const record of records) {
        const id = getUniqueId(record);
        if (idMap.has(id)) {
            nonUniqueKeys.add(id);
        } else {
            idMap.add(id);
        }
    }

    return { nonUniqueKeys };
}

File: packages/records/lib/db/migrations/20240327151027_create_records_table.ts (hunk 1 above)
Comment (bodinsamuel) [reply packages/records/lib/db/migrations/20240327151027_create_records_table.ts:78]: Looking at the current indexes, it is missing:
- id
- nango_connection_id, model, updated_at, id
- external_id

File: packages/records/lib/db/migrations/20240327151027_create_records_table.ts [hunk 16]
```diff
@@ -0,0 +1,104 @@
...
+            table.index(['connection_id', 'model', 'external_id']);
+            table.index(['connection_id', 'model', 'updated_at']);
+            table.index('sync_id');
+            table.index('sync_job_id');
+            table.unique(['connection_id', 'model', 'id']);
+        });
```
bodinsamuel, line 79: I'm surprised you can add index on the main table, from what I recall you had to add index on each partition instead

edit: nevermind it's working 👌🏻

packages/records/lib/db/migrations/20240327151027_create_records_table.ts
```diff
@@ -0,0 +1,104 @@
...
+    for (let i = 0; i < PARTITION_COUNT; i++) {
+        await knex.raw(`DROP TABLE IF EXISTS "${partitionTable(i)}"`);
+    }
+    // TRIGGERS
+    await knex.raw(`DROP TRIGGER IF EXISTS ${TABLE}_undelete_trigger ON ${TABLE};`);
+    await knex.raw(`DROP FUNCTION IF EXISTS ${TABLE}_undelete();`);
+    await knex.raw(`DROP TRIGGER IF EXISTS ${TABLE}_updated_at_trigger ON ${TABLE};`);
+    await knex.raw(`DROP FUNCTION IF EXISTS ${TABLE}_updated_at();`);
+    // INDEXES
+    await knex.schema.alterTable(TABLE, function (table) {
```
bodinsamuel: I don't think you can alter table after you dropped it

packages/records/package.json
```diff
@@ -0,0 +1,31 @@
+{
+    "name": "@nangohq/nango-records",
+    "version": "1.0.0",
+    "type": "module",
+    "main": "dist/index.js",
```
bodinsamuel, line 5: it does not exists

packages/server/lib/server.ts
```diff
@@ -38,6 +38,7 @@ import { getLogger } from '@nangohq/utils/dist/logger.js';
 import { getGlobalOAuthCallbackUrl, environmentService, getPort, errorManager, getWebsocketsPath, packageJsonFile } from '@nangohq/shared';
 import oAuthSessionService from './services/oauth-session.service.js';
 import migrate from './utils/migrate.js';
+import { migrate as migrateRecords } from '@nangohq/records/dist/db/migrate.js';
```
bodinsamuel: Can we have a barrel file instead? same issue as `utils` typescript can't autosuggest this file in vscode

packages/records/lib/db/migrations/20240327151027_create_records_table.ts
```diff
@@ -0,0 +1,104 @@
...
+            CREATE TABLE "${TABLE}" (
+            id uuid NOT NULL,
+            external_id character varying(255) NOT NULL,
+            json jsonb,
+            data_hash character varying(255) NOT NULL,
+            connection_id integer NOT NULL,
+            model character varying(255) NOT NULL,
+            created_at timestamp with time zone NOT NULL DEFAULT CURRENT_TIMESTAMP,
+            updated_at timestamp with time zone NOT NULL DEFAULT CURRENT_TIMESTAMP,
+            deleted_at timestamp with time zone,
```
bodinsamuel, line 23: Not sure if it's a good idea but can we take this opportunity to remove the timezone from this table?

packages/records/lib/models/records.ts
bodinsamuel: can be done in a subsequent PR if that brings too much change imo

packages/server/lib/server.ts
bodinsamuel: also I would not ship this PR with this, so you can merge right now and try later.
and it's missing in migrate.ts

bodinsamuel (commented):
Looks good 🚀

packages/records/lib/db/migrations/20240327151027_create_records_table.ts
TBonnin: good point

packages/records/lib/db/migrations/20240327151027_create_records_table.ts
TBonnin: > Looking at the current indexes, it is missing:
> 
>     * id
The idea is that id should be unique but because it is a partition table `unique constraint on partitioned table must include all partitioning columns` which is cover by `table.unique(['connection_id', 'model', 'id'])` 
> 
>     * nango_connection_id, model, updated_at, id
Fixed L76 to include `id` that I had forgotten
> 
>     * external_id
Which query would benefit from a `index(external_id)`?

packages/records/lib/models/records.ts
TBonnin: done

packages/records/lib/db/migrations/20240327151027_create_records_table.ts
TBonnin: 🤦

packages/server/lib/server.ts
TBonnin: good point about not shipping

packages/records/lib/db/migrations/20240327151027_create_records_table.ts
TBonnin: why remove the timezone?

packages/server/lib/server.ts
TBonnin: also added a barrel file

packages/records/lib/db/migrations/20240327151027_create_records_table.ts
bodinsamuel: imo it's too complicated to handle date with timezone and provides no advantages, no big monstrous bug but some stuff will slightly differ from time to time which is inconvenient. 
e.g: If the script is running during Daylight Saving Time change, if the scripts are running in different timezones etc. 
Unless I'm missing something it's at the very least not useful

packages/records/lib/db/migrations/20240327151027_create_records_table.ts
bodinsamuel: > * id

I was mostly thinking about when we query by ID not uniqueness.

> Which query would benefit from a index(external_id)?

In theory none, but it was used by the query planner in the last month to yield 500K rows

packages/records/lib/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './db/migrate.js';
+export * as Records from './models/records.js';
```
bodinsamuel: I think it's common to only consider PascalCase for class, wdyt?

bodinsamuel (approved):
🚀 
Some minor comments still open but not blocking

packages/records/lib/db/migrations/20240327151027_create_records_table.ts
```diff
@@ -0,0 +1,95 @@
...
+            BEFORE UPDATE ON ${TABLE}
+            FOR EACH ROW
+            EXECUTE FUNCTION ${TABLE}_updated_at();
+        `);
+        // INDEXES
+        await knex.schema.alterTable(TABLE, function (table) {
+            table.index(['connection_id', 'model', 'external_id']);
+            table.index(['connection_id', 'model', 'updated_at', 'id']);
+            table.index('sync_id');
+            table.index('sync_job_id');
```
khaliqgant, line 77: * ~`data_hash`~ --> this appears to be dropped
* `nango_connection_id, mode, external_id` --> haven't checked if that is used / needed

packages/records/lib/db/migrations/20240327151027_create_records_table.ts
TBonnin: since I am not sure if it is a good idea or not I am gonna keep it like it is. Timezones is a tricky topic and I am not sure there is one perfect solution

packages/records/lib/index.ts
TBonnin: I personally don't like wildcard import name with lowercase. Ok it is not a class but it is basically a namespace and using lowercase doesn't make it clear that is not a instance of a class or a function. 
If the convention in js land is to use lowercase I will change it though

TBonnin:
I am gonna merge as is. The `records` package is not used yet.
Regarding the indexes, I think I am covering all queries but anyway I will double check and monitor postgres slow queries when enabling double write. I will also put `getRecords` behind a flag so I can confirm there is no gap in term of indexes

PR #1934, opened 2024-04-01
add: script to migrate records

TBonnin:
## Describe your changes
Adding script to migrate records

### DIFF
--- a/scripts/one-off/records-migration/package.json
+++ b/scripts/one-off/records-migration/package.json
@@ -6,7 +6,8 @@
     "scripts": {
         "test": "echo \"Error: no test specified\" && exit 1",
         "build": "tsc -b tsconfig.json .",
-        "migrate": "npm run build && node ./dist/migrate.js"
+        "migrate": "npm run build && node ./dist/migrate.js",
+        "verify": "npm run build && node ./dist/verify.js"
     },
     "author": "",
     "dependencies": {
--- a/scripts/one-off/records-migration/package.json
+++ b/scripts/one-off/records-migration/package.json
@@ -1,0 +1,15 @@
+{
+    "name": "records-migration",
+    "version": "1.0.0",
+    "description": "",
+    "type": "module",
+    "scripts": {
+        "test": "echo \"Error: no test specified\" && exit 1",
+        "build": "tsc -b tsconfig.json .",
+        "migrate": "npm run build && node ./dist/migrate.js"
+    },
+    "author": "",
+    "dependencies": {
+        "knex": "^3.1.0"
+    }
+}
--- a/scripts/one-off/records-migration/README.md
+++ b/scripts/one-off/records-migration/README.md
@@ -1,5 +1,9 @@
-# Run
+# How to run
 
 Update `SOURCE_DB_URL` and `TARGET_DB_URL` into [./migrate.ts](migrate.ts)
 
-Run `npm run migrate`
+Run:
+```
+npm install
+npm run migrate
+```
--- a/scripts/one-off/records-migration/README.md
+++ b/scripts/one-off/records-migration/README.md
@@ -1,0 +1,5 @@
+# Run
+
+Update `SOURCE_DB_URL` and `TARGET_DB_URL` into [./migrate.ts](migrate.ts)
+
+Run `npm run migrate`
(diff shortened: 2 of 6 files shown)


scripts/one-off/records-migration/package.json
```diff
@@ -0,0 +1,15 @@
+{
```
bodinsamuel, line 1: it's usually fair to have a `scripts/package.json` so that it can benefits more than one

scripts/one-off/records-migration/main.ts
```diff
@@ -0,0 +1,171 @@
...
+            .where((builder) => {
+                if (checkpoint.lastCreatedAt && checkpoint.lastId) {
+                    builder.where(sourceKnex.raw(`(created_at, id) > (?, ?)`, [checkpoint.lastCreatedAt, checkpoint.lastId]));
+                }
+            })
+            .orderBy([
+                { column: 'created_at_raw', order: 'asc' },
+                { column: 'id', order: 'asc' }
+            ])
+            .limit(batchSize)) as unknown as TargetRecord[];
```
bodinsamuel: no need to change for a script but to avoid this you need to set `select<TargetRecord[]>`

scripts/one-off/records-migration/main.ts
```diff
@@ -0,0 +1,171 @@
...
+        // function that execute another function returning its result + the time it took to execute
+        const lastRow = records[records.length - 1];
+        if (lastRow) {
+            checkpoint = { lastCreatedAt: lastRow.created_at, lastId: lastRow.id };
+            await saveCheckpoint(checkpoint);
+            console.log(
+                `${res.length} more rows migrated (total: ${endWrite - startRead}, read: ${endRead - startRead}ms, write: ${endWrite - startWrite}ms). Checkpoint: ${JSON.stringify(checkpoint)}. `
+            );
+        } else {
+            console.warn('Could not save checkpoint');
```
bodinsamuel: I would kill the script :p

scripts/one-off/records-migration/main.ts
```diff
@@ -0,0 +1,171 @@
+import Knex from 'knex';
+import fs from 'fs';
```
bodinsamuel: nb: you can directly import `fs/promises` when you need it

bodinsamuel (approved):
Looks good, I haven't run it but the logic seems good 🙏🏻. 
I have multiple feedback/question, nothing blocking:

- How do we handle deletion? 
- That means this script needs to be run multiple times very close to the switch date otherwise it won't be fully sync, right?
- Maybe it's over-engineering: we could parallelize per connection/sync that would (in theory) make the script faster and could also catch deletion (=> a deleted connection = we drop the records)

TBonnin:
> Looks good, I haven't run it but the logic seems good 🙏🏻. I have multiple feedback/question, nothing blocking:
> 
>     * How do we handle deletion?
> 
>     * That means this script needs to be run multiple times very close to the switch date otherwise it won't be fully sync, right?
> 


Short answer: we don't handle deletion at migration. My reasoning is that orphan records caused by race conditions are not causing any problems functionally speaking. They will be taking unnecessary space on storage though.
I am planning to clean them up then once the migration is complete and running.

>     * Maybe it's over-engineering: we could parallelize per connection/sync that would (in theory) make the script faster and could also catch deletion (=> a deleted connection = we drop the records)

That's a good idea. It would make the checkpointing and error handling more complicated though. I will look into it to see if the added complexity is worth it

Thank you for the feedback

bodinsamuel (approved):

bodinsamuel:
> That's a good idea. It would make the checkpointing and error handling more complicated though. I will look into it to see if the added complexity is worth it

Yep, a little bit more logic that would require to paginate connection/syncs then copy associated records. Either by page or with a queue. I have used this package to avoid rewriting logic like this https://caolan.github.io/async/v3/docs.html#queue you have a `concurrency` that helps managing the processing smoothly.
Considering it will probably take many days maybe it's worth it but not entirely sure either so feel free to merge

khaliqgant (approved):
Curious about how you tested this and what is the plan for running this on staging?

TBonnin:
> Curious about how you tested this and what is the plan for running this on staging?

I have tested using prod replica as the source and localhost as target.
I will first run the migration in staging when enabling double writes

PR #1936, opened 2024-04-02
fix(demo): server side tracking

bodinsamuel:
## Describe your changes

Tried to be smart about server side tracking because calls can be retried and some do not have dedicated endpoint (3 and 6), but some events are sent out of order making it hard to track on Posthog.

- Use more naive approach to server side tracking

### DIFF
--- a/packages/server/lib/controllers/onboarding.controller.ts
+++ b/packages/server/lib/controllers/onboarding.controller.ts
@@ -343,7 +343,7 @@ class OnboardingController {
 
             await updateOnboardingProgress(status.id, progress);
             if (progress === 3 || progress === 6) {
-                void analytics.track(AnalyticsTypes[`DEMO_${progress as 0}`], account.id, { user_id: user.id });
+                void analytics.track(AnalyticsTypes[`DEMO_${progress}`], account.id, { user_id: user.id });
             }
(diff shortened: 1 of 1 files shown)


packages/server/lib/controllers/onboarding.controller.ts
```diff
@@ -327,15 +332,19 @@ class OnboardingController {
...
             if (!status) {
                 res.status(404).send({ message: 'no_onboarding' });
                 return;
             }
 
-            await updateOnboardingProgress(status.id, req.body.progress);
-            void analytics.track(AnalyticsTypes[`DEMO_${req.body.progress as 0}`], account.id, { user_id: user.id });
+            await updateOnboardingProgress(status.id, progress);
+            if (progress === 3 || progress === 6) {
+                void analytics.track(AnalyticsTypes[`DEMO_${progress as 0}`], account.id, { user_id: user.id });
```
khaliqgant: `progress as 0` ?

bodinsamuel: it's just for typescript sake but actually I can remove this now that I narrow the type number

khaliqgant (approved):

PR #1941, opened 2024-04-02
fix: use destructured object on some functions

bodinsamuel:
## Describe your changes

Fixes NAN-675

In the process of migration to logs v2, I need to pass the new ctx to a lot of functions and because we use classic parameters everywhere it's a bit messy especially when there are some optional parameters too.
I didn't modified everything to avoid a too large PR.

- Use destructured object parameters where it would be painful to add a new param


packages/server/lib/controllers/proxy.controller.ts
```diff
@@ -180,36 +188,63 @@ class ProxyController {
...
+        environment_id,
+        isSync,
+        isDryRun
+    }: {
+        res: Response;
+        method: HTTP_VERB;
+        configBody: ApplicationConstructedProxyConfiguration;
+        activityLogId: number;
+        environment_id: number;
+        isSync?: boolean | undefined;
```
TBonnin, line 205: do we the undefined when using `?`

TBonnin (approved):

packages/server/lib/controllers/proxy.controller.ts
bodinsamuel: when using strict mode typescript makes a distinction between a inexistant value `?` vs a value pointing to `undefined`. It enforces object shape but can be annoying

PR #1942, opened 2024-04-02
fix(docker): change Postgres version for public docker-compose

bodinsamuel:
## Describe your changes

~I honestly don't know but seems to fix it.~
Also I realised I have pinned Postgres but if you are not part of this repo and do rely on the docker-compose.yml then it will scream at you that your data is not the same version as the docker image, so we have to leave those one as v16 unfortunately.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -19,7 +19,7 @@ jobs:
               uses: actions/checkout@v3
 
             - name: Build the docker-compose stack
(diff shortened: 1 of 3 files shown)


bodinsamuel:
Okay I have tested elsewhere it seems it's working again without this fix, probably just github fail.
The PR is still relevant for Postgres version

TBonnin (approved):

khaliqgant (approved):

PR #1943, opened 2024-04-02
runner: report high memory usage

TBonnin:
## Describe your changes
The goal is to report the runner high memory usage so users can understand better when a script is failing.
This is not a perfect solution, for example if memory spikes and runner crashes within a check interval no activity log would be emitted.

@bastienbeurier this is how it looks like
Screenshot 2024-04-02 at 20 51 34

https://linear.app/nango/issue/NAN-665/make-runner-out-of-memory-error-more-explicit

### DIFF
--- a/packages/jobs/lib/server.ts
+++ b/packages/jobs/lib/server.ts
@@ -35,7 +35,7 @@ function healthProcedure() {
 function idleProcedure() {
     return publicProcedure.input(z.object({ runnerId: z.string().nonempty(), idleTimeMs: z.number() })).mutation(async ({ input }) => {
         const { runnerId, idleTimeMs } = input;
-        logger.info(`[IDLE]: runner '${runnerId}' has been idle for ${idleTimeMs}ms. Suspending...`);
+        logger.info(`[runner ${runnerId}]: idle for ${idleTimeMs}ms. Suspending...`);
         await suspendRunner(runnerId);
         return { status: 'ok' };
     });
--- a/packages/runner/lib/monitor.ts
+++ b/packages/runner/lib/monitor.ts
@@ -1,0 +1,150 @@
+import os from 'os';
+import fs from 'fs';
+import { stringifyError, type NangoProps } from '@nangohq/shared';
+import * as superjson from 'superjson';
+import { fetch } from 'undici';
+import { getLogger } from '@nangohq/utils/dist/logger.js';
+
+const MEMORY_WARNING_PERCENTAGE_THRESHOLD = 75;
+const logger = getLogger('Runner');
+
+export class RunnerMonitor {
+    private runnerId: string;
+    private tracked: Map<number, NangoProps> = new Map<number, NangoProps>();
+    private jobsServiceUrl: string = '';
+    private persistServiceUrl: string = '';
+    private idleMaxDurationMs = parseInt(process.env['IDLE_MAX_DURATION_MS'] || '') || 0;
+    private lastIdleTrackingDate = Date.now();
+    private lastMemoryReportDate: Date | null = null;
+    private idleInterval: NodeJS.Timeout | null = null;
+    private memoryInterval: NodeJS.Timeout | null = null;
+
+    constructor({ runnerId, jobsServiceUrl, persistServiceUrl }: { runnerId: string; jobsServiceUrl: string; persistServiceUrl: string }) {
+        this.runnerId = runnerId;
+        this.jobsServiceUrl = jobsServiceUrl;
+        this.persistServiceUrl = persistServiceUrl;
+        if (this.jobsServiceUrl.length > 0) {
+            this.memoryInterval = this.checkMemoryUsage();
+            this.idleInterval = this.checkIdle();
+        }
+        process.on('SIGTERM', this.onExit.bind(this));
+    }
+
+    private onExit(): void {
+        if (this.idleInterval) {
+            clearInterval(this.idleInterval);
+        }
+        if (this.memoryInterval) {
+            clearInterval(this.memoryInterval);
+        }
+    }
+
+    track(nangoProps: NangoProps): void {
+        if (nangoProps.syncJobId) {
+            this.lastIdleTrackingDate = Date.now();
+            this.tracked.set(nangoProps.syncJobId, nangoProps);
+        }
+    }
+
+    untrack(nangoProps: NangoProps): void {
+        if (nangoProps.syncJobId) {
+            this.tracked.delete(nangoProps.syncJobId);
+        }
+    }
+
+    private checkMemoryUsage(): NodeJS.Timeout {
+        // eslint-disable-next-line @typescript-eslint/no-misused-promises
+        return setInterval(async () => {
+            const rss = process.memoryUsage().rss;
+            const total = getTotalMemoryInBytes();
+            const memoryUsagePercentage = (rss / total) * 100;
+            if (memoryUsagePercentage > MEMORY_WARNING_PERCENTAGE_THRESHOLD) {
+                await this.reportHighMemoryUsage(memoryUsagePercentage);
+            }
+        }, 1000);
+    }
+
+    private async reportHighMemoryUsage(memoryUsagePercentage: number): Promise<void> {
+        // only report if it has been more than 30 seconds since the last report
+        if (this.lastMemoryReportDate) {
+            const now = new Date();
+            const diffInSecs = (now.getTime() - this.lastMemoryReportDate.getTime()) / 1000;
+            if (diffInSecs < 30) {
+                return;
+            }
+        }
+        this.lastMemoryReportDate = new Date();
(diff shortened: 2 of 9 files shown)


packages/runner/lib/server.ts
```diff
@@ -60,7 +61,14 @@ function runProcedure() {
...
         .mutation(async ({ input }): Promise<RunnerOutput> => {
             const { nangoProps, code, codeParams } = input;
-            return await exec(nangoProps, input.isInvokedImmediately, input.isWebhook, code, codeParams);
+            if (nangoProps.activityLogId && nangoProps.environmentId) {
+                Usage.track(nangoProps.activityLogId, nangoProps.environmentId);
+            }
+            const res = await exec(nangoProps, input.isInvokedImmediately, input.isWebhook, code, codeParams);
+            if (nangoProps.activityLogId) {
+                Usage.untrack(nangoProps.activityLogId);
+            }
```
bodinsamuel, line 56: no try/catch/finally means this value could float for a long time

packages/runner/lib/usage.ts
```diff
@@ -0,0 +1,57 @@
...
+import { createActivityLogMessage } from '@nangohq/shared';
+
+interface Tracked {
+    activityLogId: number;
+    environmentId: number;
+}
+
+const MEMORY_WARNING_PERCENTAGE_THRESHOLD = 70;
+let lastMemoryReportDate: Date | null = null;
+const tracked: Map<number, Tracked> = new Map<number, Tracked>();
```
bodinsamuel: While I believe our usage of Class is not always clear, I think it was a good time to avoid floating state imo

packages/runner/lib/usage.ts
```diff
@@ -0,0 +1,57 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-misused-promises
+    setInterval(async () => {
+        const rss = process.memoryUsage().rss;
+        const total = os.totalmem();
+        const memoryUsagePercentage = (rss / total) * 100;
+
+        if (memoryUsagePercentage > MEMORY_WARNING_PERCENTAGE_THRESHOLD) {
+            await report(memoryUsagePercentage);
+        }
+    }, 1000);
```
bodinsamuel: A floating setInterval will block a graceful shutdown, it needs to be stored and closed properly

packages/runner/lib/server.ts
```diff
@@ -5,6 +5,7 @@ import type { Request, Response, NextFunction } from 'express';
 import timeout from 'connect-timeout';
 import type { NangoProps, RunnerOutput } from '@nangohq/shared';
 import { getLogger } from '@nangohq/utils/dist/logger.js';
+import * as Usage from './usage.js';
```
khaliqgant: Any reason why this is capitalized? I think usage is a quite generic name and wouldn’t give a clue to what it is doing at first glance

khaliqgant (approved):
Non blocking feedback

packages/runner/lib/usage.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+    lastMemoryReportDate = new Date();
+    for (const { environmentId, activityLogId } of tracked.values()) {
+        await createActivityLogMessage({
+            level: 'warn',
+            environment_id: environmentId,
+            activity_log_id: activityLogId,
+            content: `Memory usage of nango scripts is high: ${memoryUsagePercentage.toFixed(2)}% of the total available memory.`,
+            timestamp: Date.now()
+        });
```
bodinsamuel: I'm not sure how it works since Runner do not have database access

TBonnin: 🤦

TBonnin:
@khaliqgant @bodinsamuel thank you for your feedback. I have refactored the PR quite a bit. Would you mind giving it another look?

packages/runner/lib/monitor.ts
```diff
@@ -0,0 +1,156 @@
...
+const logger = getLogger('Runner');
+
+export class RunnerMonitor {
+    private runnerId: string;
+    private tracked: Map<number, NangoProps> = new Map<number, NangoProps>();
+    private baseUrl: string = '';
+    private idleMaxDurationMs = parseInt(process.env['IDLE_MAX_DURATION_MS'] || '') || 0;
+    private lastIdleTrackingDate = Date.now();
+    private lastMemoryReportDate: Date | null = null;
+    private idleInterval: NodeJS.Timeout | null = null;
```
TBonnin, line 19: a bit out of scope but I renamed this entire logic `RunnerMonitor` and moved the idle logic into it, since it consist in monitoring the runner idling time

packages/runner/lib/monitor.ts
```diff
@@ -0,0 +1,156 @@
...
+        return 0;
+    }
+}
+
+function getTotalMemoryInBytes(): number {
+    // when running inside a container, os.totalmem() returns the total memory of the system, not the memory limit of the container
+    // see: https://github.com/nodejs/node/issues/51095
+    // process.constrainedMemory() is supposed to return the memory limit of the container but it doesn't work on Render
+    // so we need to use a workaround to get the memory limit of the container on Render
+    return process.constrainedMemory() || getRenderTotalMemoryInBytes() || os.totalmem();
```
TBonnin, line 131: This is incredibly fragile but I don't have a better solution to make it work in Render since `os.totalmem` doesn't return the container memory limit but the vm's one :( Let me know what you think

packages/runner/lib/monitor.ts
```diff
@@ -0,0 +1,156 @@
...
+        }
+    }
+}
+
+function getRenderTotalMemoryInBytes(): number {
+    try {
+        const output = fs.readFileSync('/sys/fs/cgroup/memory.max', 'utf-8');
+        const memoryLimitInBytes = parseInt(output.trim(), 10);
+        return memoryLimitInBytes;
+    } catch (_error) {
```
bodinsamuel:
```suggestion
    } catch {
```

bodinsamuel: (note you could also just log it for debug purposes)

packages/runner/lib/monitor.ts
```diff
@@ -0,0 +1,156 @@
...
+                    Accept: 'application/json',
+                    'Content-Type': 'application/json'
+                },
+                body: superjsonEncodedBody
+            });
+            if (res.status !== 200) {
+                logger.error(`Error reporting '${superjsonEncodedBody}' to '${url}': ${JSON.stringify(await res.json())}`);
+            }
+        } catch (err) {
+            logger.error(`Error reporting '${superjsonEncodedBody}' to '${url}': ${JSON.stringify(err)}`);
```
bodinsamuel: JSON.stringify error doesn't work, I have added an util for that

```suggestion
            logger.error(`Error reporting '${superjsonEncodedBody}' to '${url}': ${stringifyError(err)}`);
```

packages/jobs/lib/server.ts
```diff
@@ -17,7 +18,8 @@ const publicProcedure = t.procedure;
 
 const appRouter = router({
     health: healthProcedure(),
-    idle: idleProcedure()
+    idle: idleProcedure(),
+    activityLogMessage: activityLogMessageProcedure()
```
bodinsamuel: I'm surprised it goes back to Jobs and not persist? I felt like `/environment/:environmentId/log` was kind of already there for that

bodinsamuel (approved):
Works great locally 👍🏻 
Minor comments

khaliqgant (approved):

packages/jobs/lib/server.ts
TBonnin: Totally. I am an idiot. Switching to use the existing persist endpoint

bodinsamuel (approved):
Looks very nice 🥰 sorry for all the feedbacks and thanks for the patience!

I have an other, running locally I get this: raises two questions why it throwed and why logger do not have `warn` 🤔 
edit: I have checked, we have configured the logger with syslog.levels which has `warning` not `warn`. Type is not good unfortunately.
```ts
[2]         logger.warn(`Error reading memory limit from ${memoryMaxFile}: ${stringifyError(error)}`);
[2]                ^
[2]
[2] TypeError: logger.warn is not a function
[2]     at getRenderTotalMemoryInBytes (file:///Users/samuelbodin/code/nango/packages/runner/dist/monitor.js:114:16)
[2]     at getTotalMemoryInBytes (file:///Users/samuelbodin/code/nango/packages/runner/dist/monitor.js:123:43)
[2]     at Timeout._onTimeout (file:///Users/samuelbodin/code/nango/packages/runner/dist/monitor.js:52:27)
```
Might not exists in macOs
```ts
  syscall: 'open',
[2]   code: 'ENOENT',
[2]   path: '/sys/fs/cgroup/memory.max'
```

TBonnin:
@bodinsamuel I have decided to remove the logging. It is super noisy when running locally

PR #1944, opened 2024-04-02
fix

bodinsamuel:
## Describe your changes

While implementing Logs V2:

- Fix wrong uuid used in webhook
- Query full account object (with a single query) because I'll need it to log properly
I don't have a way to test this feature, I'm not sure what it is actually :D

### DIFF
--- a/packages/server/lib/controllers/config.controller.ts
+++ b/packages/server/lib/controllers/config.controller.ts
@@ -284,9 +284,9 @@ class ConfigController {
             }
 
             const config = await configService.getProviderConfig(providerConfigKey, environmentId);
-            const environmentUuid = await environmentService.getAccountUUIDFromEnvironment(environmentId);
+            const environment = await environmentService.getById(environmentId);
 
-            if (config == null) {
+            if (!config || !environment) {
                 errorManager.errRes(res, 'unknown_provider_config');
                 return;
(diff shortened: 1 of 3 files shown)


packages/server/lib/controllers/config.controller.ts
```diff
@@ -284,9 +284,9 @@ class ConfigController {
             }
 
             const config = await configService.getProviderConfig(providerConfigKey, environmentId);
-            const environmentUuid = await environmentService.getAccountUUIDFromEnvironment(environmentId);
+            const environment = await environmentService.getById(environmentId);
```
bodinsamuel, line 287: not entirely sure if it never worked or we just got lucky?

khaliqgant: What do you mean? Are you seeing an incorrect value with how it was previously?

bodinsamuel: we were getting the accountUUID and not environmentUUID

khaliqgant: Ah ok I see, thanks. That value isn't used but rather the frontend uses the `webhook_receive_url` from the environment.

khaliqgant (approved):

PR #1945, opened 2024-04-03
feat(logs): implem v2 logging

bodinsamuel:
## Context

Contributes to NAN-582

Since we started talking about that a lot of stuff have fluctuated. So here is a recap:
- When we want to log something we create an `Operation` that will contain the state and the main data (i.e: account, env, sync, connection, user)
- Each log message has the exact same schema with `parentId` filled with the `Operation`'s id
- logs and operation are stored in the same index

## Changes

- Double write of logs to Elasticsearch storage

The write is currently controlled by the env var `NANGO_LOGS_ENABLED` to be able to disable everything if something goes wrong. The main calls to ES have been try/catched temporarily to also avoid any issue. 

It should cover all logs across the codebase, but it's missing a lot of context data (i.e: account, env name, sync, connection) that were not added to limit the size of this PR. 

## How to test?

- Set the env
```sh
# (false will log to console only, true will log to ES)
NANGO_LOGS_ENABLED=false | true 
```
- Launch the backend
```sh
npm run dw
npm run dwa
```
- Launch a sync / action, create a connection, cancel a sync, deploy integrations
- Go to  http://localhost:9500/messages/_search?sort=createdAt:desc
Its not very readable but you should find all the logs there.

## Next steps

- Complete the missing data
- New API endpoint
- Implement the new UI
- Switch to new UI
- Remove activityLogId param wherever it's now useless (it's still going to be needed at some place where it's serialized)
- Remove old library


bodinsamuel:
Well I can't pack logs because it depends on utils that is also packed. So I'll let you review but we need to find a solution
Check the step before the one failing: https://github.com/NangoHQ/nango/actions/runs/8754509051/job/24026432058?pr=1945

scripts/publish.sh
```diff
@@ -30,14 +39,12 @@ npm ci
...
 # pack utils and install it in shared
 mkdir -p "$GIT_ROOT_DIR/packages/shared/vendor"
-pushd "$GIT_ROOT_DIR/packages/utils"
-jq '.bundleDependencies = true' package.json > temp.json && mv temp.json package.json
-npm install --workspaces=false
-npm run build
-npm pack --pack-destination "$GIT_ROOT_DIR/packages/shared/vendor"
-popd
+vendor "utils"
+vendor "logs"
```
bodinsamuel: this step doesn't work

packages/logs/lib/utils.ts
```diff
@@ -1,3 +1,8 @@
 import { getLogger } from '@nangohq/utils';
 
+// Try to satisfies CLI pack
+export type { Logger } from 'winston';
```
bodinsamuel: Because CLI is packed, the type is not portable and can't compile

packages/logs/lib/client.ts
```diff
@@ -45,33 +98,125 @@ export class LogContext {
...
+        if (this.dryRun) {
+            return;
+        }
+
+        try {
+            await setTimeouted({ id: this.id });
+        } catch (err) {
+            // TODO: reup throw
+            logger.error(`failed_to_set_timeout ${stringifyError(err)}`);
+        }
```
TBonnin: noticed this pattern is coming back multiple time. What about something like 
```
async do(f: () => void, consoleMsg: string, errorMsg: string): Promise<void> {
        if (this.logToConsole) {
            logger.info(consoleMsg);
        }
        if (this.dryRun) {
            return;
        }

        try {
            f();
        } catch (err) {
            // TODO: reup throw
            logger.error(errorMsg);
        }
}
```

packages/logs/lib/env.ts
```diff
@@ -1,3 +1,5 @@
 import { parseEnvs, ENVS } from '@nangohq/utils';
 
-export const envs = parseEnvs(ENVS.required({ NANGO_LOGS_ES_URL: true, NANGO_LOGS_ES_USER: true, NANGO_LOGS_ES_PWD: true }));
+// Because this file is being loaded by the CLI and runner it's not possible to required anything :(
```
TBonnin: runner shouldn't log anything directly. no?

packages/logs/lib/models/helpers.ts
```diff
@@ -44,3 +54,21 @@ export function getFormattedMessage(data: Partial<MessageRow>): MessageRow {
...
+    CANCEL: 'cancel'
+} as const;
+
+export const oldLevelToNewLevel = {
+    debug: 'debug',
+    info: 'info',
+    warn: 'warn',
+    error: 'error',
+    verbose: 'debug',
+    silly: 'debug',
```
TBonnin, line 64: 🙈

packages/persist/lib/controllers/persist.controller.ts
```diff
@@ -57,6 +58,9 @@ class PersistController {
             },
             false
         );
+        const logCtx = getExistingOperationContext({ id: String(activityLogId) });
```
TBonnin: when does one know when to use `getExistingOperationContext` vs `getOperationContext`? when one has a `activityLogId`?

packages/logs/lib/client.ts
```diff
@@ -45,33 +98,125 @@ export class LogContext {
...
+export interface OperationContextData extends FormatMessageData {
+    start?: boolean;
+}
+
+/**
+ * Create an operation and return a context
+ */
+export async function getOperationContext(
+    data: OperationRowInsert,
+    { start, account, user, environment }: SetRequired<OperationContextData, 'account'>,
```
TBonnin: why is environment not required? is it not something we always want/have?

packages/server/lib/controllers/connection.controller.ts
```diff
@@ -136,13 +149,19 @@ class ConnectionController {
...
                     environment_id: environment.id,
                     auth_mode: template?.auth_mode,
                     content: `Token manual refresh fetch was successful for ${providerConfigKey} and connection ${connectionId} from the web UI`,
                     timestamp: Date.now()
                 });
+                const logCtx = await getOperationContext(
+                    { id: String(activityLogId), operation: { type: 'token' }, message: 'Get connection web' },
+                    { account: response.account, environment: response.environment }
+                );
+                await logCtx.info('Token manual refresh fetch was successful');
```
TBonnin: are we not losing the providerConfigKey and connectionId info?

packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -136,6 +146,8 @@ class OAuthController {
                         timestamp: Date.now(),
                         content: error.message
                     });
+                    await logCtx.error(error.message);
+                    await logCtx.failed();
```
TBonnin, line 150: would it be nice to be able to chain things like `logCtx.error(...).failed()`?

packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -1505,9 +1650,11 @@ class OAuthController {
                         },
                         session.provider,
                         activityLogId,
-                        { initiateSync, runPostConnectionScript }
+                        { initiateSync, runPostConnectionScript },
+                        logCtx
                     );
                 }
+                await logCtx.success();
```
TBonnin, line 1661: what's bother me with current logs implementation, which we are not addressing, is that it is very easy to forget to mark the activity/operation as successful or failed

packages/logs/lib/types/messages.ts
```diff
@@ -1,12 +1,56 @@
-export type LogLevel = 'trace' | 'debug' | 'info' | 'warn' | 'error';
```
TBonnin, line 1: might be nice to document the different type and possible values. For example I am not sure what is `MessageType` used for and what do the values `log` | `http` means

packages/shared/lib/clients/sync.client.ts
```diff
@@ -289,6 +301,8 @@ class SyncClient {
                     syncData: JSON.stringify(syncData)
                 }
             });
+            // await logCtx.error('Failed to init sync', {error: err});
+            // await logCtx.failed();
```
TBonnin, line 306: commented?

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -75,6 +77,10 @@ export async function deploy(
     });
 
     const activityLogId = await createActivityLog(log);
+    const logCtx = await getOperationContext(
+        { id: String(activityLogId), operation: { type: 'deploy', action: 'custom' }, message: 'Deploying custom syncs' },
+        { account: { id: accountId }, environment: { id: environment_id } }
+    );
```
TBonnin, line 83: you probably have tried but I have to ask since one of challenge of dealing with the codebase is the tangled dependencies: is it impossible to inject the `getOperationContext` or any logs high level function into `shared` so it doesn't have to required and loaded by any component using shared that doesn't also need `logs`?

TBonnin (commented):
Thank you for the hard work. Looks very good. 
I have a few Qs but we can talk on Monday

packages/logs/lib/env.ts
bodinsamuel: It doesn't but load shared and thus logs 😢

packages/persist/lib/controllers/persist.controller.ts
bodinsamuel: Yes exactly, I can merge both if that's confusing

packages/logs/lib/client.ts
bodinsamuel: yeah good point 👍🏻

packages/server/lib/controllers/oauth.controller.ts
bodinsamuel: could be great but since both are promise the syntax would be painful 😞 
```ts
await (await logCtx.error('Missing HMAC in query params')).failed();
```

packages/logs/lib/types/messages.ts
bodinsamuel: I'll try to add more comments 👌🏻

packages/shared/lib/clients/sync.client.ts
bodinsamuel: Yeah, I didn't changed broken logic but added the comment to not forget. This one of the places where the activity logs were not updated in the catch

packages/server/lib/controllers/connection.controller.ts
bodinsamuel: this is repeated information, it's already stored in the logOperation (but I'll revert actually because I haven't fixed all of them)

packages/logs/lib/client.ts
bodinsamuel: why not yes 👌🏻

packages/server/lib/controllers/oauth.controller.ts
bodinsamuel: yep agreed, there is no easy way to do unless you wrap everything with a big function which is annoying and would increase the change related by a huge amount. 
However the spec has evolved and since Bastien wants something closer to an auditlog I will probably end up provisioning the logContext at the API level and thus wrap the endpoints transparently. BUT it's not going to be that easy because some endpoints returning does not mean the activity is done (i.e: oauth login)

If you have an other idea I'm definitely open.

packages/shared/lib/services/sync/config/deploy.service.ts
bodinsamuel: it's not impossible, but requires a lot more work and it's weird to pass just a function since it's a not singleton. However I could make a class LogContextGetter (or whatever name) to make it clear.

Since I'm stuck with the CLI, I might try

packages/persist/lib/controllers/persist.controller.ts
TBonnin: it is not confusing when you remember both exist but it might be an good idea to simplify so there is always one function to use. It might make the signature too complex though (`getOperationContext` signature is already quite a mouthful)

packages/server/lib/controllers/oauth.controller.ts
TBonnin: what about using `then` 
```
await logCtx.error('Missing HMAC in query params')).then((ctx) => ctx.failed)
```

I am nitpicking here feel free to ignore me :)

packages/server/lib/controllers/oauth.controller.ts
TBonnin: yeah let's discuss

packages/persist/lib/controllers/persist.controller.ts
bodinsamuel: I have changed everything to inject the dependency, let me know 👍🏻

packages/logs/lib/client.integration.test.ts
```diff
@@ -55,17 +64,51 @@ describe('client', () => {
...
+        expect(spyMsg).not.toHaveBeenCalled();
+        expect(spyLog).toHaveBeenCalledTimes(2);
+    });
+
+    it('should respect logToConsole=false', async () => {
+        const spy = vi.spyOn(logger, 'info');
+
+        // Create operation
+        expect(spy).not.toHaveBeenCalled();
+        const ctx = await logContextGetter.create(operationPayload, { account, environment }, { dryRun: true, logToConsole: false });
```
TBonnin, line 99: why this one has `logToConsole` false and the one above is `true`

packages/logs/lib/env.ts
TBonnin: is it still true?

packages/shared/lib/clients/sync.client.ts
```diff
@@ -30,8 +30,8 @@ import telemetry, { LogTypes } from '../utils/telemetry.js';
 import errorManager, { ErrorSourceEnum } from '../utils/error.manager.js';
 import { NangoError } from '../utils/error.js';
 import type { RunnerOutput } from '../models/Runner.js';
-import { isTest, isProd, getLogger, metrics } from '@nangohq/utils';
-import { isErr, resultOk, type Result, resultErr } from '@nangohq/utils';
+import type { LogContext, LogContextGetter } from '@nangohq/logs';
```
TBonnin, line 33: so is it ok to still depends on `logs` just for types? It doesn't load anything else from `logs`?

bodinsamuel: that's what I'm trying to figure out, in theory it should be okay because we don't ship devDeps

packages/logs/lib/client.integration.test.ts
bodinsamuel: Im testing that logToConsole is working as intended (with the spy just above)

packages/shared/lib/clients/sync.client.ts
bodinsamuel: Seems to work if we trust the CI 🙌🏻

TBonnin (approved):

packages/logs/lib/env.ts
bodinsamuel: good point it's not anymore

bodinsamuel:
Latest addition: 
- dependency injection to avoid installing logs in shared as a prod deps
- using opensearch locally too, with the appropriate js client. Elasticsearch js client refuse to connect to an opensearch instance

I will test staging asap and merge if no other issues

packages/server/lib/controllers/oauth.controller.ts
bodinsamuel: I feel like it's more confusing, ideally the calls would buffer until the last chain to avoid this like knex query builder

bodinsamuel:
Everything good

PR #1946, opened 2024-04-03
[nan-680] add is_capped and admin only api to edit

khaliqgant:
## Describe your changes
Add a way to track customers and manage with a simple API. This is the first step of resource capping as we need to know which customers to cap.

NAN-680

### DIFF
--- a/packages/shared/lib/db/migrations/20240403100610_add_account_status_info.cjs
+++ b/packages/shared/lib/db/migrations/20240403100610_add_account_status_info.cjs
@@ -2,14 +2,12 @@ const tableName = '_nango_accounts';
 
 exports.up = async function (knex, _) {
     await knex.schema.alterTable(tableName, function (table) {
-        table.boolean('is_paying').defaultTo(false);
-        table.string('subscription_type');
+        table.boolean('is_capped').defaultTo(true);
     });
 };
 
 exports.down = async function (knex, _) {
     await knex.schema.table(tableName, function (table) {
-        table.dropColumn('is_paying');
-        table.dropColumn('subscription_type');
+        table.dropColumn('is_capped');
     });
 };
--- a/packages/shared/lib/db/migrations/20240403100610_add_account_status_info.cjs
+++ b/packages/shared/lib/db/migrations/20240403100610_add_account_status_info.cjs
@@ -1,0 +1,15 @@
+const tableName = '_nango_accounts';
+
+exports.up = async function (knex, _) {
+    await knex.schema.alterTable(tableName, function (table) {
+        table.boolean('is_paying').defaultTo(false);
+        table.string('subscription_type');
+    });
+};
+
+exports.down = async function (knex, _) {
+    await knex.schema.table(tableName, function (table) {
+        table.dropColumn('is_paying');
+        table.dropColumn('subscription_type');
+    });
+};
--- a/packages/shared/lib/services/account.service.ts
+++ b/packages/shared/lib/services/account.service.ts
@@ -33,6 +33,20 @@ class AccountService {
         }
     }
 
+    async getAccountByUUID(uuid: string): Promise<Account | null> {
+        try {
+            const result = await db.knex.select('*').from<Account>(`_nango_accounts`).where({ uuid }).first();
+            return result || null;
+        } catch (e) {
+            errorManager.report(e, {
+                source: ErrorSourceEnum.PLATFORM,
+                operation: LogActionEnum.DATABASE
+            });
+
+            return null;
+        }
+    }
+
     async getAccountAndEnvironmentIdByUUID(targetAccountUUID: string, targetEnvironment: string): Promise<{ accountId: number; environmentId: number } | null> {
         const account = await db.knex.select('id').from<Account>(`_nango_accounts`).where({ uuid: targetAccountUUID });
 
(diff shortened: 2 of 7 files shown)


packages/shared/lib/services/account.service.ts
```diff
@@ -33,6 +33,20 @@ class AccountService {
         }
     }
 
+    async getAccountByUUID(uuid: string): Promise<Account | null> {
+        try {
+            const result = await db.knex.select('*').from<Account>(`_nango_accounts`).where({ uuid }).first();
```
bodinsamuel: Just my personal opinion, maybe @TBonnin has some thoughts too:
That would be great for us to start having model functions that are a bit more pure (= less side effects). The problem with early try/catch it needs a lot of boilerplate to handle and does not actually make the error reporting much smoother since we never know if everything was try/catch correctly. 
It makes the code looks like go (lengthy functions with a lot of `if err != nil`) and it's especially annoying in tests imo
 
One good thing, it allows more precise error reporting (vs a global try/catch that needs more fine-tuning) but we could also just re-throw the error wrapped.

packages/shared/lib/db/migrations/20240403100610_add_account_status_info.cjs
```diff
@@ -0,0 +1,13 @@
+const tableName = '_nango_accounts';
+
+exports.up = async function (knex, _) {
+    await knex.schema.alterTable(tableName, function (table) {
+        table.boolean('is_capped').defaultTo(true);
```
bodinsamuel, line 5: Open question:
I don't think it's really future proof, I'm not even sure what does that means in practice. Bastien will manually set this flag for people abusing the system or not paying?

imo future proof does not require much more work: a json column `plan` with something like 
`{ connection: { max: <number> }, syncs: { max: number }, records: { maxSize: number, maxTotalSize: number} }`
That would accommodate more use cases from: free tier, trial, different plan, custom entreprise, etc. Especially that we already know that it will happen for records storage and we'll need a place to store those info.

bodinsamuel (approved):
Approved on principle, but I have some open questions

packages/shared/lib/db/migrations/20240403100610_add_account_status_info.cjs
TBonnin: I also vote for a `plan: json` column. Even if it means an object with a single boolean field for now. I can guarantee than very soon we will need different values and not only boolean but threshold values like @bodinsamuel described. Changing the type of the column will be annoying

packages/shared/lib/services/account.service.ts
TBonnin: 200% agree. Side effects make refactoring more painful and tests more complicated because of the extra dependencies. I would recommend that whenever we can we push error handling, logging, etc.. to the most outer layer possible. model function for instance can return a `Promise<Result<T>>`. The error being handled by the controller calling the model

bodinsamuel: tbh I think the `Result<>` we put in place has the same drawbacks I mentioned earlier. I think I just find that `throw new Error` is the most simple and effective way to handle this. At least for model functions. 

We don't use a lot of transaction but in our code base we already have a lot of intermediate state that needs to be handled at each `if (error)` (e.g: activity logs) and that's painful to handle without a try/catch.

packages/shared/lib/db/migrations/20240403100610_add_account_status_info.cjs
khaliqgant: `is_capped` is simply if a check if a user is paying or not and also allows us to allow high priority trial customers get the full feature set. It is purposely simple and flexible. 

I disagree regarding the `json` column. I think making it a json column is anticipating something in which we have no idea what we're going to want and making for a check that is not as optimal as it can be given known parameters/spec in the meantime.

packages/shared/lib/services/account.service.ts
khaliqgant: Removed the try catch

packages/shared/lib/db/migrations/20240403100610_add_account_status_info.cjs
TBonnin: every single customer plan model I have encountered quickly evolves into a complex set of features. Go with the boolean column if you prefer but I bet you will owe @bodinsamuel and I a beer relatively soon. 😆

khaliqgant: 😂

I definitely agree that the customer plan model will get more complex as we grow. I just would prefer we modify what we have to account for the next iteration explicitly. Happy to provide a beer in either case though 🍻

bodinsamuel: it was an open suggestion, I don't think my solution was perfect but I do believe more planning would not hurt

khaliqgant: The planning for this was done as part of the spec of the [issue](https://linear.app/nango/issue/NAN-680/implement-way-to-upgradedowngrade-customers) & [project brief](https://linear.app/nango/document/resource-capping-v0-project-brief-749be67127a9) itself.

PR #1947, opened 2024-04-03
fix(utils): add barrel file

bodinsamuel:
## Describe your changes

Fixes NAN-687

Since I didn't find any working alternatives and I haven't had any other suggestion I'm proposing this.

- Add a standard barrel file in `utils`


packages/utils/lib/errors.ts
```diff
@@ -0,0 +1,15 @@
...
+export function errorToObject(err: unknown) {
+    return serializeError(err);
+}
+
+/**
+ * Transform any Error or primitive to a string
+ */
+export function stringifyError(err: unknown) {
+    return JSON.stringify(serializeError(err));
+}
```
bodinsamuel, line 15: Note: I'm adding that but don't import it until we fix the packaging of the cli

khaliqgant (approved):

PR #1950, opened 2024-04-03
(feat) [NAN-677] enforce cap when calling nangoauth

khaliqgant:
## Describe your changes
Add check when creating a new connection and importing a connection to see if the account has reached the cap.

NAN-680


packages/shared/lib/services/sync/config/config.service.ts
```diff
@@ -569,6 +569,54 @@ export async function getSyncConfigsWithConnectionsByEnvironmentId(environment_i
     return result;
 }
 
+export async function getSyncConfigsWithConnectionsByEnvironmentIdAndProviderConfigKey(
+    providerConfigKey: string,
+    environment_id: number
+): Promise<({ connections: { connection_id: string }[] } & SyncConfig & ProviderConfig)[]> {
+    const result = await schema()
```
khaliqgant: Happy to get another pair of eyes on this query to improve it.

packages/shared/lib/utils/error.ts
```diff
@@ -559,6 +559,13 @@ export class NangoError extends Error {
                 this.message = `Missing an account name for account login/signup.`;
                 break;
 
+            case 'resource_capped':
+                this.status = 400;
+                // TODO docs link
```
khaliqgant, line 564: Pending from Bastien.

packages/shared/lib/services/sync/config/config.service.ts
bodinsamuel: I guess you only want connection with syncs? I feel like it's harder to explain to customers but whatever :D 
Isn't something like this enough? I'm not sure why you selected that much stuff from the table

```sql
SELECT
	COUNT(1)
FROM
	_nango_connections
JOIN _nango_syncs ON _nango_syncs.nango_connection_id = _nango_connections.id AND _nango_syncs.deleted = FALSE
WHERE
	environment_id = 2
	AND provider_config_key = 'github-demo'
	AND _nango_syncs.deleted = FALSE;
```

and if it's only 3 connections ever
```sql
SELECT
	COUNT(1)
FROM
	_nango_connections
WHERE
	environment_id = 2
	AND provider_config_key = 'github-demo'
	AND _nango_syncs.deleted = FALSE;
```

packages/shared/lib/hooks/hooks.ts
```diff
@@ -11,10 +11,40 @@ import integrationPostConnectionScript from '../integrations/scripts/connection/
...
+
+    const scriptConfigs = await getSyncConfigsWithConnectionsByEnvironmentIdAndProviderConfigKey(providerConfigKey, environmentId);
+
+    let reachedCap = false;
+
+    if (scriptConfigs.length > 0) {
+        for (const script of scriptConfigs) {
+            const { connections } = script;
+
+            if (connections.length >= 3) {
```
bodinsamuel: Could be a constant for clarity

packages/shared/lib/hooks/hooks.ts
```diff
@@ -11,10 +11,40 @@ import integrationPostConnectionScript from '../integrations/scripts/connection/
...
+    environmentId
+}: {
+    providerConfigKey: string | undefined;
+    environmentId: number;
+}): Promise<boolean> => {
+    if (!providerConfigKey) {
+        return false;
+    }
+
+    const scriptConfigs = await getSyncConfigsWithConnectionsByEnvironmentIdAndProviderConfigKey(providerConfigKey, environmentId);
```
bodinsamuel: opinion: not a huge fan of overly descriptive function, we have the parameters so we already know that info. `getConnectionsCount()` is more useful imo and directly state its raison d'être

packages/shared/lib/utils/error.ts
```diff
@@ -559,6 +559,13 @@ export class NangoError extends Error {
...
+                this.message =
+                    'Free accounts cannot have > 3 connections for integrations with active scripts. Upgrade or deactivate the scripts to create more connections (https://docs.nango.dev/relevant-link).';
```
bodinsamuel: Not a good idea to document moving part in the error message (people rely on error message to automate stuff and we often forget to update them).
Something like `You have reached maximum number of connections. Check our documentation to know more (https://docs.nango.dev/relevant-link).` create far less problem in the long-run

bodinsamuel (commented):
small review, not tested

packages/server/lib/middleware/resource-capping.middleware.ts
```diff
@@ -0,0 +1,37 @@
...
+    const { success, error, response } = await getEnvironmentAndAccountId(res, req);
+
+    if (!success || response === null) {
+        errorManager.errResFromNangoErr(res, error);
+        return;
+    }
+
+    const { accountId, environmentId } = response;
+
+    const account = await accountService.getAccountById(accountId);
```
TBonnin, line 19: we are making more queries to the db on each request by adding a middleware which is doing something similar to the `access.middleware`, ie: fetching the account and environment from the db. Can we fetch the `account.is_capped` as part of the `access.middleware` and save it into the `res.locals` like we do with `accountId` and `environmentId` so it can be access down the line by controllers?

packages/shared/lib/hooks/hooks.ts
```diff
@@ -11,10 +11,40 @@ import integrationPostConnectionScript from '../integrations/scripts/connection/
 import webhookService from '../services/notification/webhook.service.js';
 import { SpanTypes } from '../utils/telemetry.js';
 import { isCloud, isLocal, isEnterprise } from '../utils/temp/environment/detection.js';
+import { getSyncConfigsWithConnectionsByEnvironmentIdAndProviderConfigKey } from '../services/sync/config/config.service.js';
 import type { Result } from '../utils/result.js';
 import { resultOk, resultErr } from '../utils/result.js';
 import { NangoError } from '../utils/error.js';
 
+export const connectionCreationStartCapCheck = async ({
```
TBonnin, line 24: do we need a new function? Can we not call `syncService.getSyncsByProviderConfigKey` and count the connections?

packages/server/lib/middleware/resource-capping.middleware.ts
khaliqgant: I like the idea in theory, but the issue is that the value of `is_capped` can change via the api and then the value stored in the `res.locals` is stale and potentially inaccurate. Since the user changing the `is_capped` setting is different than the user who has the capping nor not.

packages/shared/lib/hooks/hooks.ts
khaliqgant: Updated to `getSyncConfigsWithConnections`

packages/shared/lib/utils/error.ts
khaliqgant: Updated to be less specific

packages/server/lib/middleware/resource-capping.middleware.ts
bodinsamuel: `res.locals` is only stored for the current request if that's your worry

packages/server/lib/controllers/connection.controller.ts
```diff
@@ -527,6 +528,21 @@ class ConnectionController {
...
+                return;
+            }
+
+            if (account.is_capped && provider_config_key) {
+                const isCapped = await connectionCreationStartCapCheckHook({ providerConfigKey: provider_config_key, environmentId });
+                if (isCapped) {
+                    errorManager.errRes(res, 'resource_capped');
+                    return;
+                }
+            }
```
bodinsamuel, line 544: this is already checked in the middleware right?
tbf I would only keep this check here since it's really specific to this use case

packages/server/lib/middleware/resource-capping.middleware.ts
khaliqgant: Ah yes, but still this is making it more complicated than it needs since right now the only controller it goes through is the `connection.controller` when it is imported. If it went into the `access.middleware` instead of the `capping.middleware` then the logic to check the cap would need to go through:
- `oauth.controller`
- `unauth.controller`
- `apiAuth.controller`
- `appStoreController`

packages/shared/lib/hooks/hooks.ts
khaliqgant: That method would be good, but we need both to check against syncs and actions.

packages/server/lib/middleware/resource-capping.middleware.ts
```diff
@@ -0,0 +1,37 @@
...
+    if (account.is_capped && providerConfigKey) {
+        const isCapped = await connectionCreationStartCapCheckHook({ providerConfigKey, environmentId });
+        if (isCapped) {
+            errorManager.errRes(res, 'resource_capped');
+            return;
+        }
+    }
+
+    next();
+};
```
bodinsamuel, line 37: This will be executed to all query with ?providerConfigKey in the URL? a bit unnecessary imo if we already cap at creation and it will break current users GET requests.

packages/shared/lib/services/sync/config/config.service.ts
khaliqgant:
```
SELECT
	COUNT(1)
FROM
	_nango_connections
JOIN _nango_syncs ON _nango_syncs.nango_connection_id = _nango_connections.id AND _nango_syncs.deleted = FALSE
WHERE
	environment_id = 2
	AND provider_config_key = 'github-demo'
	AND _nango_syncs.deleted = FALSE;
```

Thanks, that query would work nicely but at some point I would like to move away from storing the `provider_config_key` in the connections table and instead rely on the `config_id` there which relates back to the `_nango_configs` table. As a result of the `provider_config_key` being stored in the connections table a user can't edit a provider_config_key if there are connections are attached to it. Happy to discuss that more if you want?

packages/shared/lib/hooks/hooks.ts
khaliqgant: Updated.

packages/shared/lib/hooks/hooks.ts
```diff
@@ -11,10 +11,42 @@ import integrationPostConnectionScript from '../integrations/scripts/connection/
...
+            const { connections } = script;
+
+            if (connections.length >= CONNECTIONS_WITH_SCRIPTS_CAP_LIMIT) {
+                reachedCap = true;
+                break;
+            }
+        }
+    }
+
+    return reachedCap;
```
bodinsamuel: No need to store this value imo
```suggestion
  for (const script of scriptConfigs) {
      if (script.connections.length >= CONNECTIONS_WITH_SCRIPTS_CAP_LIMIT) {
          return true;
      }
  }
  
  return false;
```

packages/server/lib/controllers/connection.controller.ts
khaliqgant: This method doesn't go through the middleware, only the calls that are attempting to creating a connection using the public auth key go through that middleware.

packages/server/lib/middleware/resource-capping.middleware.ts
khaliqgant: No, it will just get executed for all calls that are creating a connection using their public key auth. 

```
app.route('/oauth/connect/:providerConfigKey').get(apiPublicAuth, oauthController.oauthRequest.bind(oauthController));
app.route('/oauth2/auth/:providerConfigKey').post(apiPublicAuth, oauthController.oauth2RequestCC.bind(oauthController));
app.route('/api-auth/api-key/:providerConfigKey').post(apiPublicAuth, apiAuthController.apiKey.bind(apiAuthController));
app.route('/api-auth/basic/:providerConfigKey').post(apiPublicAuth, apiAuthController.basic.bind(apiAuthController));
app.route('/app-store-auth/:providerConfigKey').post(apiPublicAuth, appStoreAuthController.auth.bind(appStoreAuthController));
app.route('/unauth/:providerConfigKey').post(apiPublicAuth, unAuthController.create.bind(unAuthController));
```

packages/shared/lib/hooks/hooks.ts
khaliqgant: Thanks. Until the values are set via the API the cap won't fire yet. Once we put it in place I'll update the code to return early instead of using a local variable. https://github.com/NangoHQ/nango/pull/1950/commits/8656bc9ded5d6c79b5f80fe28fac75b879042db5

packages/shared/lib/services/sync/config/config.service.ts
bodinsamuel: This give you all connections with active syncs or actions for a provider.
```sql 
SELECT
	conn.id
FROM
	_nango_configs AS configs
	JOIN _nango_connections AS conn ON configs.id = conn.config_id
		AND conn.deleted = FALSE
	JOIN _nango_sync_configs AS syncs ON syncs.nango_config_id = configs.id
		AND syncs.deleted = FALSE
		AND syncs.active = TRUE
WHERE
	configs.environment_id = 2
        AND configs.provider_config_key = 'github-demo'
	AND configs.deleted = FALSE
GROUP BY
	conn.id
HAVING
	COUNT(syncs.id) > 0;
```

then `res.length >= 3`

packages/server/lib/middleware/resource-capping.middleware.ts
bodinsamuel: oh okay it's a bit messy but got it, sorry for the confusion 👍🏻

packages/server/lib/controllers/connection.controller.ts
bodinsamuel: yeah okay, got confused since we have 3 apis 👌🏻

packages/shared/lib/services/sync/config/config.service.ts
khaliqgant: Thanks, this doesn't give me what I want either 

```
SELECT
	conn.id
FROM
	nango._nango_configs AS configs
	JOIN nango._nango_connections AS conn ON configs.id = conn.config_id
		AND conn.deleted = FALSE
	JOIN nango._nango_sync_configs AS syncs ON syncs.nango_config_id = configs.id
		AND syncs.deleted = FALSE
		AND syncs.active = TRUE
WHERE
	configs.environment_id = 2
        AND configs.unique_key = 'unauthenticated'
	AND configs.deleted = FALSE
GROUP BY
	conn.id
HAVING
	COUNT(syncs.id) > 0;
```

gives me one row 
image

image

bodinsamuel: Yeah indeed, I realized that `_nango_sync_configs` are per config not per connection and `_nango_syncs` are per connection but do not contain actions :/ 
I guess you got the right one then!

khaliqgant:
@bodinsamuel are you good with this one? A few other tasks rely on this one so is semi blocking but not a huge issue.

bodinsamuel (approved):
I'm sure we can improve it later 👌🏻

PR #1951, opened 2024-04-04
fix(eslint): pass #10

bodinsamuel:
## Describe your changes

- Fix await/void promises
- Fix some unhandled errors

Down to **4 errors** (and **7385 warnings** 🙈), we can now think about adding it to the CI

### DIFF
--- a/packages/shared/lib/clients/provider.client.ts
+++ b/packages/shared/lib/clients/provider.client.ts
@@ -11,8 +11,10 @@ import axios from 'axios';
 import qs from 'qs';
 import { parseTokenExpirationDate, isTokenExpired } from '../utils/utils.js';
 import { NangoError } from '../utils/error.js';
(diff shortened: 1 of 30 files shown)


packages/shared/lib/clients/provider.client.ts
```diff
@@ -11,8 +11,10 @@ import axios from 'axios';
 import qs from 'qs';
 import { parseTokenExpirationDate, isTokenExpired } from '../utils/utils.js';
 import { NangoError } from '../utils/error.js';
+import { getLogger } from '../utils/temp/logger.js';
 
 const stripeAppExpiresIn = 3600;
+const logger = getLogger('provider');
```
khaliqgant:
```suggestion
const logger = getLogger('Provider.Client');
```

PR #1953, opened 2024-04-04
[NAN-676] enforce cap when activating a script

khaliqgant:
## Describe your changes
Builds on https://github.com/NangoHQ/nango/pull/1950

NAN-676


packages/shared/lib/constants.ts
```diff
@@ -1,2 +1,4 @@
 export const SYNC_TASK_QUEUE = 'nango-syncs';
 export const WEBHOOK_TASK_QUEUE = 'nango-webhooks';
+//export const CONNECTIONS_WITH_SCRIPTS_CAP_LIMIT = 3;
+export const CONNECTIONS_WITH_SCRIPTS_CAP_LIMIT = Infinity;
```
TBonnin, line 4: is that on purpose?

packages/webapp/src/pages/Integration/components/EnableDisableSync.tsx
```diff
@@ -96,9 +113,26 @@ export default function EnableDisableSync({
...
+                setModalTitleColor('text-white');
+                setModalTitle('You’ve reached your connections limit!');
+                setModalContent(
+                    `Scripts are a paid feature. You can only use them with 3 connections or less.
+                    Upgrade or delete some connections to activate this script.`
+                );
+                setModalOkButtonTitle('Upgrade');
+                setModalCancelButtonTitle('Learn more');
+                setModalOkButtonLink('https://nango.dev/chat');
+                setModalCancelButtonLink('https://docs.nango.dev/REPLACE-ME');
```
TBonnin, line 127: is it going to be replace later?

TBonnin (commented):
question about the comment out limit constant but looks good otherwise

packages/shared/lib/constants.ts
khaliqgant: Yes, very much so. We don't want to cap until we can set the account information correctly via the API

packages/webapp/src/pages/Integration/components/EnableDisableSync.tsx
khaliqgant: Yup, launch blocker to replace.

packages/shared/lib/services/connection.service.ts
```diff
@@ -990,6 +991,25 @@ class ConnectionService {
         }
     }
 
+    public async shouldCapUsage({ providerConfigKey, environmentId }: { providerConfigKey: string | undefined; environmentId: number }): Promise<boolean> {
```
bodinsamuel: Can it really be undefined?

TBonnin (approved):

packages/shared/lib/services/connection.service.ts
khaliqgant: I'll have to look again at it. It probably _shouldn't be_

khaliqgant: Thanks, updated.

packages/shared/lib/constants.ts
bodinsamuel: Just thought about it during the night (could not sleep). We also have to think about enterprise/community 👌🏻

PR #1956, opened 2024-04-04
[nan-683] track last fetched time for a sync

khaliqgant:
## Describe your changes
Track last fetched date for syncs. Curious @TBonnin how this jives with the work you're doing?

NAN-683

### DIFF
--- a/packages/shared/lib/models/Sync.ts
+++ b/packages/shared/lib/models/Sync.ts
@@ -38,6 +38,7 @@ export interface Sync extends TimestampsAndDeleted {
         nanos?: number;
     };
     frequency: string | null;
+    last_fetched_at?: Date | null;
 }
 
 export interface Action extends TimestampsAndDeleted {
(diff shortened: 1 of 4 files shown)


packages/shared/lib/models/Sync.ts
```diff
@@ -38,6 +38,7 @@ export interface Sync extends TimestampsAndDeleted {
         nanos?: number;
     };
     frequency: string | null;
+    last_fetched_at?: Date | null;
```
bodinsamuel: It probably impact the creation type but a column can't be undefined.
```suggestion
    last_fetched_at: Date | null;
```

bodinsamuel (approved):

packages/shared/lib/models/Sync.ts
khaliqgant: Updated with https://github.com/NangoHQ/nango/pull/1956/commits/0d24d66585f7f380d1dbd9ba996aa84d5b4b464c. Yeah, forces us to be a bit more explicit

PR #1959, opened 2024-04-05
fix(proxy): dryRun optional activityLogId

bodinsamuel:
## Describe your changes

Fixes NAN-699

Most functions called by proxy didn't not know that activityLogId was maybe null (when isDryRun = true)

- Fix types
- Fix missing await


packages/shared/lib/services/proxy.service.ts
```diff
@@ -258,23 +258,21 @@ class ProxyService {
...
-                        environment_id,
-                        activity_log_id: activityLogId,
-                        timestamp: Date.now(),
-                        content
-                    }
-                ];
+                activityLogs.push({
+                    level: 'error',
+                    environment_id,
+                    activity_log_id: activityLogId as number, // We don't care if it's empty or not
```
khaliqgant: Can you add a comment explaining why?

khaliqgant (changes requested):
Small nit for explanation about a comment

packages/shared/lib/services/proxy.service.ts
bodinsamuel: sure 👍🏻

packages/shared/lib/services/proxy.service.ts
```diff
@@ -341,7 +339,7 @@ class ProxyService {
             activityLogs.push({
                 level: 'error',
                 environment_id,
-                activity_log_id: activityLogId,
+                activity_log_id: activityLogId as number, // In DryRun this value can be empty
```
TBonnin, line 342: would it make sense to actually ensure activityLogId is never empty? Probably related question: does the `id` field of activityLog need to be an optional?

TBonnin (approved):

packages/shared/lib/services/proxy.service.ts
bodinsamuel: yep I agree with you, I initially did this fix. But it was a mess because there is a lot of place where activityLogId=undefined means we don't want to log on purpose, others where it was forgotten (or at least it's not clear why it's not there), and this specific case where we want to log but not to the DB. 
The migration will hopefully clarify a bit this

bodinsamuel: (Im re-reading myself) to be more clear, we don't have an `activityLogId` in dryRun, so if I want to make it required there, I need to "invent" one which could lead to other issues since other dev could assume we do have a real one

PR #1960, opened 2024-04-05
[NAN-698] Add the ability to call a sync from an action

khaliqgant:
## Describe your changes
This allows a user to have a coordinating action that tells sync when to start based on input from other syncs. As a proof of concept example:

- coordinating action
```
import type { NangoAction } from './models';

export default async function runAction(nango: NangoAction, input?: string): Promise<boolean> {
    await nango.log('starting action');
    await nango.log(input);
    const order = [
        'first-sync',
        'second-sync'
    ];

    if (!input && order[0] && typeof order[0] === 'string') {
        await nango.log('running first sync');
        await nango.triggerSync(nango.providerConfigKey, nango.connectionId, order[0] as string)
    } else {
        if (!order.includes(input as string)) {
            throw new Error(`Invalid input: ${input}`);
        }

        const index = order.indexOf(input as string);
        const nextIndex = index + 1;

        if (order[nextIndex] && typeof order[nextIndex] === 'string') {
            await nango.log('running next sync');
            await nango.triggerSync(nango.providerConfigKey, nango.connectionId, order[nextIndex] as string);
        }
    }
    return true;
}
```
- first-sync
```
import type { NangoSync } from './models';

    await nango.log('First sync ran');
    await nango.triggerAction(nango.providerConfigKey, nango.connectionId, 'sync-coordinator', 'first-sync')
}

```
- second-sync
```
import type { NangoSync } from './models';

    await nango.log('Second sync ran');
}
```

NAN-698

### DIFF
--- a/packages/jobs/lib/runner/local.runner.ts
+++ b/packages/jobs/lib/runner/local.runner.ts
@@ -51,7 +51,7 @@ export class LocalRunner implements Runner {
                 env: {
                     ...process.env,
                     RUNNER_ID: runnerId,
-                    IDLE_MAX_DURATION_MS: '60000'
+                    IDLE_MAX_DURATION_MS: '0'
                 }
             });
 
--- a/packages/runner/lib/exec.ts
+++ b/packages/runner/lib/exec.ts
@@ -80,7 +80,11 @@ export async function exec(
                     throw new Error(`Default exports is not a function but a ${typeof scriptExports.default}`);
                 }
                 if (isAction) {
-                    return await scriptExports.default(nango, codeParams);
+                    let inputParams = codeParams;
+                    if (typeof codeParams === 'object' && Object.keys(codeParams).length === 0) {
+                        inputParams = undefined;
+                    }
+                    return await scriptExports.default(nango, inputParams);
                 } else {
                     return await scriptExports.default(nango);
                 }
--- a/packages/shared/lib/services/environment.service.ts
+++ b/packages/shared/lib/services/environment.service.ts
@@ -136,7 +136,7 @@ class EnvironmentService {
             .select<Account>('_nango_accounts.*')
             .from(TABLE)
             .join('_nango_accounts', '_nango_accounts.id', '_nango_environments.account_id')
-            .where({ id: environment_id })
+            .where('_nango_environments.id', environment_id)
             .first();
 
         return result || null;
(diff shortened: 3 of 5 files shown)


packages/jobs/lib/runner/local.runner.ts
```diff
@@ -51,7 +51,7 @@ export class LocalRunner implements Runner {
                 env: {
                     ...process.env,
                     RUNNER_ID: runnerId,
-                    IDLE_MAX_DURATION_MS: '60000'
+                    IDLE_MAX_DURATION_MS: '0'
```
khaliqgant, line 54: Unsure why we were idling on the local runner @TBonnin? Maybe as part of testing?

packages/runner/lib/exec.ts
```diff
@@ -80,7 +80,11 @@ export async function exec(
                     throw new Error(`Default exports is not a function but a ${typeof scriptExports.default}`);
                 }
                 if (isAction) {
-                    return await scriptExports.default(nango, codeParams);
```
khaliqgant, line 83: even if the input in the request was empty we were sending an empty object which was confusing and makes for a bad experience for a user of an action IMO

packages/shared/lib/services/environment.service.ts
```diff
@@ -136,7 +136,7 @@ class EnvironmentService {
             .select<Account>('_nango_accounts.*')
             .from(TABLE)
             .join('_nango_accounts', '_nango_accounts.id', '_nango_environments.account_id')
-            .where({ id: environment_id })
+            .where('_nango_environments.id', environment_id)
```
khaliqgant, line 139: @bodinsamuel noticed this happening in APM often 
image

bodinsamuel: arg nice catch thanks

bodinsamuel (approved):

packages/jobs/lib/runner/local.runner.ts
TBonnin: because it mimics how it actually work in production and since starting a runner locally is super fast I didn't see any harm having the idling functionality enabled. Did it cause any issue for you? Feel free to set an longer threshold if you prefer

TBonnin (approved):

packages/jobs/lib/runner/local.runner.ts
khaliqgant: No issue, just was very noisy when I was trying to look at the logs on my local and was new behavior so was a bit surprising. Let me know if you have any issue with this being removed.

TBonnin: It might be useful to detect if the feature stops working but I understand it adds some noise. Your call, I don't mind either way

PR #1963, opened 2024-04-05
Adds People Data Labs Auth

jwd-dev:
## Describe your changes
Adds peopledatalabs as an integration. Super simple, just an API key. 

### DIFF
--- a/docs-v2/integrations/all/peopledatalabs.mdx
+++ b/docs-v2/integrations/all/peopledatalabs.mdx
@@ -27,6 +27,6 @@ API configuration: [`peopledatalabs`](https://nango.dev/providers.yaml)
 
 ## API gotchas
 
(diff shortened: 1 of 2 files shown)


docs-v2/integrations/all/peopledatalabs.mdx
```diff
@@ -0,0 +1,32 @@
...
+## Getting started
+
+-   [Generate an API token in your People Data Labs account](https://dashboard.peopledatalabs.com/)
+-   [People Data Labs API docs](https://docs.peopledatalabs.com/docs/introduction)
+
+<Tip>Need help getting started? Get help in the [community](https://nango.dev/slack).</Tip>
+
+## API gotchas
+
+- People Data Labs uses API_KEY auth mode with query param: Param `api_token` in the request url to access different endpoints.
```
TBonnin:
```suggestion
- People Data Labs uses API_KEY auth mode with query parameter `api_key` in the request url.
```

PR #1966, opened 2024-04-08
fix wave accounting token url

sxtfv:
## Describe your changes
Fixing Wave Accounting configuration

Exchange code with url https://api.waveapps.com/oauth2/token - 405
image

Exchange the same code with url https://api.waveapps.com/oauth2/token/ - 200
image


bodinsamuel (approved):
Thanks @sxtfv for the fix!

Issue #1970, opened 2024-04-10
[node-client] nango.listRecords does not work woth modifiedAfter


descampsk:
Hello,

There is an issue with the node client and listRecords.

This code : 

```ts
const result = await this.nango.listRecords({
        providerConfigKey,
        connectionId,
        model,
        modifiedAfter,
      });
```

gets all records and not only these which are modified after a certain date.

This code 

```ts
const result = await this.nango.listRecords({
        providerConfigKey,
        connectionId,
        model,
        modifiedAfter: '2024-04-09T15:44:21.555Z',
        delta: 'true',
      });
```

returns a 400 error.

```json
[...]

PR #1971, opened 2024-04-10
fix(node-client): fix list records with modifiedAfter

descampsk:
## Describe your changes

Use delta and modified_after instead of modifiedAfter

Fix https://github.com/NangoHQ/nango/issues/1970

### DIFF
--- a/packages/node-client/lib/index.ts
+++ b/packages/node-client/lib/index.ts
@@ -416,7 +416,7 @@ export class Nango {
         const { connectionId, providerConfigKey, model, delta, modifiedAfter, limit, filter, cursor } = config;
         validateSyncRecordConfiguration(config);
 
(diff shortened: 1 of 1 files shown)


bodinsamuel:
hey @descampsk 
Thanks a lot for taking the time to report and providing a fix. We are taking over your PR just to merge it quickly. 

For clarification, `delta` is deprecated and was a date not a boolean. We just renamed delta to modified_after for clarification, that's why they were interchangeable.

bodinsamuel:
Fixed by #1972

descampsk:
hey @bodinsamuel 

Thanks but we still have to have delta set which is not indicated in the documentation https://docs.nango.dev/integrate/guides/sync-data-from-an-api#fetch-the-latest-data.

Should we have `${(delta || modifiedAfter) ? modified_after=${modifiedAfter || delta}` instead ?

bodinsamuel:
You are right indeed, we missed this details! Thanks for raising 🙏🏻

bodinsamuel:
The new version `0.39.16` has been released 👍🏻

PR #1977, opened 2024-04-10
fix(webhooks): do not await send

bodinsamuel:
## Describe your changes

Fixes NAN-721

> [!NOTE]  
> This is a proposal

When we create a new connection we send a webhook synchronously which blocks the response until the webhook is sent. 
It's debatable wether or not this should be in the critical path. I guess it's handy to have it sync but it's annoying if the backend receiving the webhook is down or slow. The ideal fix is to put a queue / dedicated webhook service

If you think this behavior is wanted/fine, then we can close this PR.

- Do not await webhooks in the critical path


TBonnin:
I have mixed feelings. I prefer when things fail explicitly. Would an error be shown if it is asynchronous but fails?

bodinsamuel:
> I have mixed feelings. I prefer when things fail explicitly. Would an error be shown if it is asynchronous but fails?

The problem is the retry / exponential backoff. Currently it never really fail unless you wait 10minutes for the response, so it's not frankly better. The error will still be shown in the ActivityLog if that's your question.

> If the webhook has to retry when sending does that pause block the rest of the connection creation?

Nope, currently the webhook is sent at the end, so the connection is created and then we wait.

-- 

I'm also undecided about this, the decision is either we optimize for best case or worst case scenario.

TBonnin:
if the error is shown in the activity logs I think it is ok to not wait.

TBonnin (approved):

bodinsamuel:
Discovered it could lead to white page when creating a new connection too, merging this

PR #1979, opened 2024-04-11
[nan-716] separate into jobs and publish packages so it doesn't block the version bump

khaliqgant:
## Describe your changes
Publish relevant packages to the github registry so the versions can later be bumped

NAN-716


.github/workflows/cli-verification.yaml
```diff
@@ -28,29 +28,66 @@ jobs:
...
+                  always-auth: true
             - name: Install the cli from the github package registry
               env:
                   NODE_AUTH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
               run: |
-                  GIT_HASH=$(git rev-parse HEAD)
-                  mkdir nango-cli-test && cd nango-cli-test
                   npm init -y
-                  npm install @nangohq/cli@0.0.1-$GIT_HASH
+                  npm install nango@npm:@nangohq/cli@0.0.1-$GIT_HASH
```
khaliqgant, line 90: We need to install it as an alias so it can be found when we try and run it.

.github/workflows/cli-verification.yaml
```diff
@@ -28,29 +28,66 @@ jobs:
...
+                  function check_and_publish {
+                    PACKAGE_PATH=$1
+                    PACKAGE_NAME=$(node -p "require('./${PACKAGE_PATH}/package.json').name")
+                    PACKAGE_VERSION=$(node -p "require('./${PACKAGE_PATH}/package.json').version")
+                    if npm view ${PACKAGE_NAME}@${PACKAGE_VERSION} ; then
+                      echo "Version ${PACKAGE_VERSION} of ${PACKAGE_NAME} already exists. Skipping publish."
+                    else
+                      echo "Publishing ${PACKAGE_NAME}@${PACKAGE_VERSION}..."
+                      pushd ./${PACKAGE_PATH}
+                      npm publish --ignore-scripts
```
bodinsamuel, line 44: I don't see `--ignore-scripts` in the allowed flag, what does it do?

bodinsamuel (approved):

.github/workflows/cli-verification.yaml
khaliqgant: Doesn’t run the prepublish and postbuild scripts since we don’t need them

bodinsamuel: the prepublish does the build no, if we skip I wonder what we are publishing?

khaliqgant: The publishing here is more a formality so that the later the npm version command doesn’t fail.

PR #1980, opened 2024-04-11
jobs: fix heartbeat

TBonnin:
runningScripts used to be of type Record<string, context> but was changed to a Map in https://github.com/NangoHQ/nango/commit/8a5520e228cc9cde059c20278c0183c12a71d696 It caused the `Object.keys(this.runningScripts)` to always return an empty array, which means no heartbeat have benn sent in 2 months

https://linear.app/nango/issue/NAN-711/sync-frequency-issue

### DIFF
--- a/packages/jobs/lib/integration.service.ts
+++ b/packages/jobs/lib/integration.service.ts
@@ -5,6 +5,9 @@ import { createActivityLogMessage, localFileService, remoteFileService, NangoErr
 import type { Runner } from './runner/runner.js';
 import { getOrStartRunner, getRunnerId } from './runner/runner.js';
 import tracer from 'dd-trace';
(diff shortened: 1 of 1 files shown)


packages/jobs/lib/integration.service.ts
```diff
@@ -210,16 +213,17 @@ class IntegrationService implements IntegrationServiceInterface {
 
     private sendHeartbeat() {
         setInterval(() => {
-            Object.keys(this.runningScripts).forEach((syncId) => {
-                const scriptObject = this.runningScripts.get(syncId);
-
-                if (!scriptObject) {
-                    return;
+            this.runningScripts.forEach((script, syncId) => {
```
TBonnin, line 216: this line is the fix. The rest is error handling/logging

khaliqgant (approved):

PR #1983, opened 2024-04-11
allow empty client_secret for github app

khaliqgant:
## Describe your changes
Currently prevents any new github app's being created


bodinsamuel:
You pushed GitHub app inside this PR, should we review it?

khaliqgant:
> You pushed GitHub app inside this PR, should we review it?

Nope. That was a mistake this morning! Ill remove it when I land

bodinsamuel (approved):

PR #1987, opened 2024-04-15
fix(api): stop using env cookie

bodinsamuel:
## Describe your changes

Fixes NAN-729

We were using cookie as a way to automically detect env, because cookies are shared across tab this could create issues when switching env in an other tab and doing a request in an other.

- Remove use of cookie env, all (almost) path now take `env` as a query parameter

Tested on staging but please test on your side since it's touching every pages and endpoints (edit, deleted and stuff)


packages/server/lib/utils/utils.ts
```diff
@@ -1,41 +1,52 @@
...
 
-    const environmentAndAccount = await environmentService.getAccountAndEnvironmentById(user.account_id, currentEnvironment);
+    return resultOk(user);
+}
 
+export async function getUserAccountAndEnvironmentFromSession(
+    req: Request<any, any>
+): Promise<ServiceResponse<{ user: User; account: Account; environment: Environment }>> {
+    const getUser = await getUserFromSession(req);
+    if (!getUser.ok) {
```
TBonnin: maybe we should agree on an unique way to check for failure. For example always using `isErr` or `isOk`. I think the consistency will make code easier to read on the long term

khaliqgant (changes requested):
Trying to create a connection results in 
image

bodinsamuel:
> Looks like the frontend sdk calls need to set an env as well.

There is a deeper issue here, because it shouldn't be the case. There is some shared code that shouldn't shared across different auth strategy

bodinsamuel:
Can I get a review again? @khaliqgant @TBonnin

TBonnin (approved):

packages/server/lib/utils/utils.ts
bodinsamuel: ah yeah, it just works like this so I don't always think about the helpers

khaliqgant (approved):
Code looks good. I haven’t had a chance to pull again locally to test but feel free to merge if you want. If not I’ll test again this evening

packages/server/lib/utils/utils.ts
bodinsamuel: Fixed*

PR #1988, opened 2024-04-15
fix(npm): use ci instead of install

bodinsamuel:
## Describe your changes

Fixes NAN-734

`npm ci` ([doc](https://docs.npmjs.com/cli/v9/commands/npm-ci)) is useful for automated environment, it prevent any accidental upgrade and exit the process if the lockfile has to be modified.

- Use `npm ci` in all scripts


TBonnin (approved):

bodinsamuel:
There is a broken behavior in npm apparently, not sure where it failed but I had to rollback stuff related to publishing the cli.
also an implicit behavior `npm publish` calls all the `prepublish` and prepare for the workspace, not just the current directory

TBonnin (approved):

PR #1989, opened 2024-04-15
fix(node): use version 20

bodinsamuel:
## Describe your changes

Fixes NAN-732

Node20 is the LTS since october, it's now mostly stable and will provide new API and perf upgrade.
The codebase "should" stay compatible with 18 until April 2025 (we are still running the tests in 18 but we are not doing matrix tests with everything). We have to be careful with new API only available in v20 or we can also decide to entirely drop support for v18. However I think it's safer to support 18 for a longtime because of the node-client and cli.

- Use Node20 by default
- Upgrade related github actions package


.github/workflows/cli-verification.yaml
```diff
@@ -78,7 +78,7 @@ jobs:
         steps:
             - uses: actions/setup-node@v4
               with:
-                  node-version: '18.x'
+                  node-version: '20.12.2'
```
bodinsamuel, line 81: for some reason using .nvmrc was causing issues

bodinsamuel:
I will test in staging before merging *

khaliqgant (approved):

bodinsamuel:
Tested on staging, everything is alright, I'll let it run for the night to see if it crashes on its own

TBonnin (approved):

bodinsamuel:
Everything work on staging (deployed all the infra and tested syncs)

PR #1993, opened 2024-04-15
[NAN-681] Allow custom syncs to be deactivated

khaliqgant:
## Describe your changes
Allow custom scripts to be deactivated from the UI without breaking devs workflow by making a script default to be enabled. If a script is deactivated and is then deployed it'll stay deactivated. 

NAN-681


packages/webapp/src/types.ts
```diff
@@ -122,6 +122,7 @@ export enum AuthModes {
 
 export interface Connection {
     id: number;
+    connection_id: string;
```
khaliqgant, line 125: 😒

TBonnin: it is very confusing to have `connectionId` and `connection_id`

packages/server/lib/controllers/config.controller.ts
```diff
@@ -355,22 +366,38 @@ class ConfigController {
...
-                    const filteredSyncs = syncs.filter((sync) => !enabledSyncs.some((enabledSync) => enabledSync.name === sync.name));
-                    const filteredActions = actions.filter((action) => !enabledActions.some((enabledAction) => enabledAction.name === action.name));
-
-                    unEnabledFlows.syncs = filteredSyncs;
-                    unEnabledFlows.actions = filteredActions;
+                const enabledAndDisabledFlows = await getConfigWithEndpointsByProviderConfigKey(environmentId, providerConfigKey);
+                const publicFlows: StandardNangoConfig = availableFlowsForProvider as StandardNangoConfig;
+                const disabledFlows = publicFlows;
+
+                if (availableFlows && enabledAndDisabledFlows && publicFlows) {
```
TBonnin: `availableFlows` doesn't seem to be used in the if block

packages/server/lib/controllers/config.controller.ts
```diff
@@ -355,22 +366,38 @@ class ConfigController {
...
+                        (acc, sync: NangoSyncConfig) => (sync.enabled ? acc.enabledSyncs.push(sync) : acc.disabledSyncs.push(sync), acc),
+                        { enabledSyncs: [], disabledSyncs: [] }
+                    );
+                    const { enabledActions, disabledActions } = actions.reduce<ActionConfigs>(
+                        (acc, action: NangoSyncConfig) => (action.enabled ? acc.enabledActions.push(action) : acc.disabledActions.push(action), acc),
+                        { enabledActions: [], disabledActions: [] }
+                    );
+
+                    enabledAndDisabledFlows.syncs = enabledSyncs;
+                    enabledAndDisabledFlows.actions = enabledActions;
```
TBonnin: `enabledAndDisabledFlows` name is confusing: at first it contains all flows but here we mutate it to only have enabled syncs/actions

packages/server/lib/controllers/config.controller.ts
```diff
@@ -355,22 +366,38 @@ class ConfigController {
...
+
+                    const filteredSyncs = publicSyncs.filter(
+                        (sync) => ![...enabledSyncs, ...disabledSyncs].some((enabledSync) => enabledSync.name === sync.name)
+                    );
+                    const filteredActions = publicActions.filter(
+                        (action) => ![...enabledActions, ...disabledActions].some((enabledAction) => enabledAction.name === action.name)
+                    );
+
+                    disabledFlows.syncs = [...filteredSyncs, ...disabledSyncs];
+                    disabledFlows.actions = [...filteredActions, ...disabledActions];
```
TBonnin: exctracting this block into its own function with maybe a comment might make it easier to understand. Maybe it is just me but I had to look at it for a long time to fully grasp what's going on here

packages/shared/lib/services/nango-config.service.ts
```diff
@@ -257,6 +257,18 @@ const parseModelInEndpoint = (endpoint: string, allModelNames: string[], inputMo
     return { success: true, error: null, response: inputModel };
 };
 
+const getEnabledStatus = (script: NangoIntegrationDataV2, isPublic: boolean, preBuilt: boolean): boolean => {
```
TBonnin: what about calling this function `isEnabled`?

packages/webapp/src/types.ts
khaliqgant: Agree completely. Started digging into the resolution and turned out to be a little bit of a rabbit hole

TBonnin: why do we need it though? I am sure we are eventually going to run into bugs because of it

packages/server/lib/server.ts
```diff
@@ -210,8 +209,10 @@ app.route('/api/v1/syncs').get(webAuth, syncController.getSyncs.bind(syncControl
 app.route('/api/v1/sync/:syncId/frequency').put(webAuth, syncController.updateFrequency.bind(syncController));
 app.route('/api/v1/flows').get(webAuth, flowController.getFlows.bind(syncController));
 app.route('/api/v1/flow/deploy/pre-built').post(webAuth, flowController.deployPreBuiltFlow.bind(flowController));
+app.route('/api/v1/flow/deploy/pre-built').post(webAuth, flowController.deployPreBuiltFlow.bind(flowController));
```
bodinsamuel: bad c/c?

packages/shared/lib/services/nango-config.service.ts
```diff
@@ -366,7 +378,10 @@ export function convertV2ConfigObject(config: NangoConfigV2, showMessages = fals
                     webhookSubscriptions = [sync['webhook-subscriptions'] as string];
                 }
             }
+            const is_public = (isPublic !== undefined ? isPublic : sync.is_public === true) as boolean;
+            const pre_built = (isPublic !== undefined ? isPublic : sync.pre_built === true) as boolean;
```
bodinsamuel: It was like that before but it's masking that isPublic can be null

packages/shared/lib/services/nango-config.service.ts
```diff
@@ -465,14 +481,18 @@ export function convertV2ConfigObject(config: NangoConfigV2, showMessages = fals
             }
 
             const scopes = action?.scopes || action?.metadata?.scopes || [];
+            const is_public = (isPublic !== undefined ? isPublic : action.is_public === true) as boolean;
+            const pre_built = (isPublic !== undefined ? isPublic : action.pre_built === true) as boolean;
```
bodinsamuel: same comment

packages/shared/lib/services/sync/config/config.service.ts
```diff
@@ -473,7 +475,19 @@ export async function getSyncConfigByParams(
...
-    await schema().from<SyncConfig>(TABLE).where({ id, deleted: false }).update({ active: false, deleted: true, deleted_at: new Date() });
+    await schema().from<SyncConfig>(TABLE).where({ id, deleted: false }).update({
+        active: false,
+        deleted: true,
+        deleted_at: new Date()
+    });
+}
+
+export async function disableScriptConfig(id: number, environment_id: number): Promise<void> {
+    await schema().from<SyncConfig>(TABLE).where({ id, environment_id, deleted: false }).update({ enabled: false });
```
bodinsamuel: if you provide an id, the other conditions are useless

packages/shared/lib/services/sync/config/config.service.ts
```diff
@@ -473,7 +475,19 @@ export async function getSyncConfigByParams(
...
+}
+
+export async function enableScriptConfig(id: number, environment_id: number): Promise<void> {
+    await schema().from<SyncConfig>(TABLE).where({ id, environment_id, deleted: false }).update({ enabled: true });
```
bodinsamuel: same comment

packages/shared/lib/models/Sync.ts
```diff
@@ -125,6 +126,7 @@ export interface SlimSync {
     sync_id?: string | null;
     providerConfigKey?: string;
     connections?: number;
+    enabled?: boolean;
```
bodinsamuel, line 129: The problem with mixing typing here is that it's optional at creation but not when you query

bodinsamuel (approved):

packages/shared/lib/models/Sync.ts
khaliqgant: This is for a `SlimSync` where it is optional and it isn't queried for but rather is a constructed object where `enabled` isn't always needed.

packages/server/lib/controllers/config.controller.ts
khaliqgant: Done with https://github.com/NangoHQ/nango/pull/1993/commits/3dbb6110054b7df5b935a995cf17b5764773c1f8

packages/shared/lib/services/nango-config.service.ts
khaliqgant: Done with https://github.com/NangoHQ/nango/pull/1993/commits/d3a38c96d98ec76ea5b024a9a687418896f66451

packages/shared/lib/services/nango-config.service.ts
khaliqgant: Updated with https://github.com/NangoHQ/nango/pull/1993/commits/ff41ddb750763966dbdbb237e91d4285429605b0

packages/shared/lib/services/nango-config.service.ts
khaliqgant: Updated with https://github.com/NangoHQ/nango/pull/1993/commits/ff41ddb750763966dbdbb237e91d4285429605b0

packages/webapp/src/types.ts
khaliqgant: Depending on the API it sometimes sends `connectionId` and sometimes `connection_id`. I agree it should be fixed but unsure if it is in the scope of this PR. Let me know if you think it is a blocker for this PR @bodinsamuel @TBonnin

TBonnin (approved):

packages/webapp/src/types.ts
TBonnin: up to you @khaliqgant  but I would prefer if we fix this as soon as possible

packages/shared/lib/models/Sync.ts
bodinsamuel: It is used in `getSyncConfigsByProviderConfigKey` which now output enabled by default

bodinsamuel (approved):

packages/webapp/src/types.ts
khaliqgant: FYI updated with https://github.com/NangoHQ/nango/pull/1993/commits/986a22b09d21b8749d2fb9773ed6b610831fafa9

PR #1994, opened 2024-04-15
fix(api): only enable session for web

bodinsamuel:
## Describe your changes

Fixes NAN-739

Session has been enabled by default on all router, it has no big impact since most user connecting would have no session. But if you happened to be logged to app.nango.dev you could end up in the wrong branch of the code (imo there is too much shared code with implicit behavior but that's an other issue). 

`req.user` is always defined if you have a cookie wether you are in web api or api key api
https://github.com/NangoHQ/nango/blob/d0733880116932098c3fb1d938f6a556f03b3bc5/packages/shared/lib/utils/utils.ts#L296-L315)

It was mostly fine until the cap check (only a few endpoints are impacted), because of the call to `getEnvironmentAndAccountId`. So a query with a session could end up trying to create a connection to a wrong/random env (`utils.ts:318`) leading to a white page.
And also break my PR that is enforcing `env` in query param #1987 (it was only enforced on session router which lead to this investigation)

That's why I'm trying to push for code split in the APIs because it's very easy to mix stuff

- Use different router to split the Web API and the Public API
Mid-term fix would be to create a file per API but to limit the issues with sharing logic like this I think we should split by package entirely (even if we still end up with just one express API loading everything)

## Test

We don't have a strong setup for e2e API test, so I suggest to run it locally and test creating a connection.


packages/server/lib/server.ts
```diff
@@ -157,90 +154,93 @@ if (MANAGED_AUTH_ENABLED) {
 }
 
 // Webapp routes (session auth).
-app.route('/api/v1/meta').get(webAuth, environmentController.meta.bind(environmentController));
```
TBonnin, line 160: signup/login/etc... routes above are not part of the web router?

TBonnin (approved):

packages/server/lib/server.ts
bodinsamuel: oh yes good catch

PR #1995, opened 2024-04-16
[NAN-697] load a script from a directory that is the provider config

khaliqgant:
## Describe your changes
Allow nested directories to be used in a nango-integrations directory. There will be follow ups to update the `init` command and also add a migrate command.

Allows for
```
nango-integrations/
├── nango.yaml               
└── hubspot/
    ├── syncs/
    │   └── contacts.ts      
    └── actions/
        └── create-contact.ts
```

and is backwards compatible. 

NAN-697


TBonnin (approved):

packages/cli/lib/sync.unit.test.ts
```diff
@@ -451,4 +469,40 @@ describe('generate function tests', () => {
...
+
+        const success = await compileService.run({ debug: true });
+
+        await fs.promises.rm('./hubspot', { recursive: true, force: true });
+        await fs.promises.rm('./github', { recursive: true, force: true });
+        await fs.promises.rm('./dist', { recursive: true, force: true });
+        await fs.promises.rm('./nango.yaml', { force: true });
+        await fs.promises.rm('./models.ts', { force: true });
+
+        expect(success).toBe(true);
```
bodinsamuel, line 488: there is no verification that nested file have been indeed compiled no?

bodinsamuel (approved):
Working locally 👍🏻 
Might be good to actually document this alternative somewhere

packages/cli/lib/sync.unit.test.ts
khaliqgant: No, but if not compiled successfully, success won't be true.

khaliqgant:
A follow up to this will be to allow a sync/action name to be duplicated within different providers.
```
hubspot:
   syncs:
     contacts:

salesforce:
  syncs:
    contacts:
```

khaliqgant:
> Working locally 👍🏻 Might be good to actually document this alternative somewhere

Yeah there is a specific documentation task along with a few others for the migration command etc

PR #1996, opened 2024-04-16
run records database migration

TBonnin:
## Describe your changes

Setup records db migration so records table will be created. I have also setup the env var in staging and production

Next step will be double writing records in this table in addition to writing them in current table

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -27,6 +27,7 @@ services:
             - NANGO_DB_SSL=${NANGO_DB_SSL}
             - NANGO_DB_POOL_MIN=${NANGO_DB_POOL_MIN}
             - NANGO_DB_POOL_MAX=${NANGO_DB_POOL_MAX}
+            - RECORDS_DATABASE_URL=postgres://${NANGO_DB_USER}:${NANGO_DB_PASSWORD}@${NANGO_DB_HOST}:5432/${NANGO_DB_NAME}$([[ "${NANGO_DB_SSL}" == "true" ]] && echo "?sslmode=require")
             - SERVER_PORT=${SERVER_PORT}
             - NANGO_SERVER_URL=${NANGO_SERVER_URL:-http://localhost:3003}
             - NANGO_DASHBOARD_USERNAME=${NANGO_DASHBOARD_USERNAME}
--- a/packages/records/lib/db/migrations/20240327151027_create_records_table.ts
+++ b/packages/records/lib/db/migrations/20240327151027_create_records_table.ts
@@ -22,8 +22,7 @@ export async function up(knex: Knex): Promise<void> {
             updated_at timestamp with time zone NOT NULL DEFAULT CURRENT_TIMESTAMP,
             deleted_at timestamp with time zone,
             sync_id uuid,
-            sync_job_id integer,
-            CONSTRAINT ${TABLE}_connection_id_external_id_model UNIQUE (connection_id, external_id, model)
+            sync_job_id integer
             ) PARTITION BY HASH (connection_id, model)
         `);
         for (let i = 0; i < PARTITION_COUNT; i++) {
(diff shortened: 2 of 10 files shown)


docker-compose.yaml
```diff
@@ -27,6 +27,7 @@ services:
             - NANGO_DB_SSL=${NANGO_DB_SSL}
             - NANGO_DB_POOL_MIN=${NANGO_DB_POOL_MIN}
             - NANGO_DB_POOL_MAX=${NANGO_DB_POOL_MAX}
+            - RECORDS_DATABASE_URL=postgres://${NANGO_DB_USER}:${NANGO_DB_PASSWORD}@${NANGO_DB_HOST}:5432/${NANGO_DB_NAME}$([[ "${NANGO_DB_SSL}" == "true" ]] && echo "?sslmode=require")
```
TBonnin, line 30: not sure it is necessary since hosted is only use for oauth

packages/records/lib/db/migrations/20240327151027_create_records_table.ts
```diff
@@ -23,7 +23,6 @@ export async function up(knex: Knex): Promise<void> {
             deleted_at timestamp with time zone,
             sync_id uuid,
             sync_job_id integer,
-            CONSTRAINT ${TABLE}_connection_id_external_id_model UNIQUE (connection_id, external_id, model)
```
TBonnin, line 26: was defined twice. Here and https://github.com/NangoHQ/nango/pull/1996/files#diff-b8bc3ca28e369e9030e8fb23d563c153a1fc8b88c11d08083b1ed2ed3a57c445R73

bodinsamuel: the one you highlighted is not UNIQUE tho

docker-compose.yaml
bodinsamuel, line 30: Could have been easier in the code. Also it's not documented in .env.example

packages/server/lib/server.ts
```diff
@@ -76,6 +77,7 @@ const upload = multer({ storage: multer.memoryStorage() });
 // operator should run migrate.ts once before starting the service.
 if (NANGO_MIGRATE_AT_START === 'true') {
     await migrate();
+    await migrateRecords();
```
bodinsamuel, line 80: missing in migrate.ts

TBonnin: not sure what migrate.ts is for but yeah I will add it there

docker-compose.yaml
TBonnin: I didn't document it because it it is all based on other env vars that are documented. Also as mentioned I am not sure it is necessary since hosted is only used for oauth

packages/records/lib/db/migrations/20240327151027_create_records_table.ts
TBonnin: Good point. I have explicitly used unique below

packages/server/lib/server.ts
bodinsamuel: https://github.com/NangoHQ/nango/pull/663

packages/records/lib/db/migrations/20240327151027_create_records_table.ts
bodinsamuel: There is a syntax error but otherwise looks good now 👍🏻

bodinsamuel (approved):
Let's go 🚀 
Can you just add records/dist in the nodemon.json

TBonnin:
> Let's go 🚀 Can you just add records/dist in the nodemon.json

do you mean adding a `packages/records/nodemon.json`?

bodinsamuel:
> do you mean adding a packages/records/nodemon.json?

No sorry I wasn't clear, in packages/server/nodemon.json you need to add records to the watchlist otherwise it won't reload on change

PR #1997, opened 2024-04-16
fix(integration): use random id on conflict

bodinsamuel:
## Describe your changes

Fixes NAN-747

When creating an integration we insert the exact name then we append an "autoincremented id". However the only way we have is to count the integrations with the same name, it's unreliable since names can be changed or if we delete something else than the last one (the count is going down but not the max id)
- `1, 2, 3`  -> next is 4 ✅
- `2, 3` -> next is 3 ❌

Parsing the name could be unreliable too and adding an auto-incremented column over-engineered for just a name.

- Use a random id if it's not the first integration with that name
- Reuse id generation

Screenshot 2024-04-16 at 16 41 50


packages/utils/lib/id.ts
```diff
@@ -0,0 +1,9 @@
+import { customAlphabet } from 'nanoid';
+
+/**
+ * Nanoid without special char to use in URLs
+ */
+export const alphabet = '346789ABCDEFGHJKLMNPQRTUVWXYabcdefghijkmnpqrtwxyz';
+export const minSize = 8;
```
TBonnin: is minSize ever used anywhere?

TBonnin (approved):

packages/utils/lib/id.ts
bodinsamuel: ah good catch, it was a leftover 👌🏻

PR #1999, opened 2024-04-16
Move Result and core Encryption logic to packages/utils

TBonnin:
## Describe your changes
For convenience reason `records` package was depending on `shared` for `Result` and `EncryptionManager`. To implement records writing we need `shared` to depend on `records` to mark records as deleted when `track_delete = true`, which would create a dependency cycle between `shared` and `records`. 
This PR is moving `Result` and the core of the `Encryption` logic to utils so `records` package doesn't depend on `shared` anymore. And anyway no library package should depends on `shared`


packages/shared/lib/utils/encryption.manager.ts
```diff
@@ -19,40 +18,15 @@ interface DataRecordJson {
...
+if (!ENCRYPTION_KEY) {
+    throw new Error('NANGO_ENCRYPTION_KEY environment variable is required.');
+}
 
-class EncryptionManager {
-    private key: string | undefined;
-    private algo: CipherGCMTypes = 'aes-256-gcm';
-    private encoding: BufferEncoding = 'base64';
-    private encryptionKeyByteLength = 32;
+class EncryptionManager extends Encryption {
```
TBonnin, line 22: ideally all functions in this file should be moved in the model they refer to. I kept it for simplicity sake, I didn't want to go down the rabbit hole of refactoring a huge chunk of the codebase

bodinsamuel (approved):
Looks good 💪🏻 you need to remove shared from records tsconfig.json

PR #2000, opened 2024-04-17
Write records to new database

TBonnin:
This PR implements the double write to the new records database. No reading from this db is done yet.

### DIFF
--- a/packages/cli/lib/services/dryrun.service.ts
+++ b/packages/cli/lib/services/dryrun.service.ts
@@ -8,6 +8,7 @@ import { parseSecretKey, printDebug, hostport, getConnection, getConfig } from '
 import configService from './config.service.js';
 import compileService from './compile.service.js';
 import integrationService from './local-integration.service.js';
+import type { RecordsServiceInterface } from '@nangohq/shared/lib/services/sync/run.service.js';
 
 interface RunArgs extends GlobalOptions {
     sync: string;
@@ -148,8 +149,23 @@ class DryRunService {
             messages: []
         };
 
+        // dry-run mode does not read or write to the records database
+        // so we can safely mock the records service
+        const recordsService: RecordsServiceInterface = {
+            markNonCurrentGenerationRecordsAsDeleted: (
+                _connectionId: number,
+                _model: string,
+                _syncId: string,
+                _generation: number
+                // eslint-disable-next-line @typescript-eslint/require-await
+            ): Promise<string[]> => {
+                return Promise.resolve([]);
+            }
+        };
+
         const syncRun = new syncRunService({
             integrationService,
+            recordsService,
             writeToDb: false,
             nangoConnection,
             provider,
--- a/packages/persist/lib/controllers/persist.controller.ts
+++ b/packages/persist/lib/controllers/persist.controller.ts
@@ -1,5 +1,6 @@
 import type { NextFunction, Request, Response } from 'express';
 import type { LogLevel, DataResponse, DataRecord, UpsertResponse } from '@nangohq/shared';
+import { records as recordsService, format as recordsFormatter, type FormattedRecord, type UnencryptedRecordData } from '@nangohq/records';
 import {
     createActivityLogMessage,
     errorManager,
@@ -14,7 +15,9 @@ import {
 } from '@nangohq/shared';
 import tracer from 'dd-trace';
 import type { Span } from 'dd-trace';
-import { resultErr, resultOk, isOk, type Result } from '@nangohq/utils';
+import { getLogger, resultErr, resultOk, isOk, isErr, type Result } from '@nangohq/utils';
+
+const logger = getLogger('PersistController');
 
 type persistType = 'save' | 'delete' | 'update';
 type RecordRequest = Request<
@@ -68,8 +71,18 @@ class PersistController {
             params: { environmentId, nangoConnectionId, syncId, syncJobId },
             body: { model, records, providerConfigKey, connectionId, activityLogId }
         } = req;
-        const persist = async (dataRecords: DataRecord[]) => {
-            return await dataService.upsert(dataRecords, nangoConnectionId, model, activityLogId, environmentId, false);
+        const persist = async (records: FormattedRecord[], legacyRecords: DataRecord[]) => {
+            recordsService
+                .upsert(records, nangoConnectionId, model, false)
+                .then((res) => {
+                    if (isErr(res)) {
+                        throw res.err;
+                    }
+                })
+                .catch((reason) => {
+                    logger.error(`Failed to save records: ${reason}`);
+                });
+            return await dataService.upsert(legacyRecords, nangoConnectionId, model, activityLogId, environmentId, false);
         };
         const result = await PersistController.persistRecords({
             persistType: 'save',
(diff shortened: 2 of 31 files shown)


TBonnin:
@khaliqgant do I need to pack and install `records` like `utils` in publish.sh? https://github.com/NangoHQ/nango/blob/4a40221b7544a7b12780b5d1bbc3b794ec08eee3/scripts/publish.sh#L33-L40

packages/records/lib/models/records.ts
```diff
@@ -10,7 +10,7 @@ import type {
     UnencryptedRecord,
     UpsertSummary
 } from '../types.js';
-import { decryptRecordData, decryptRecords, encryptDataRecords } from '../utils/encryption.js';
+import { decryptRecord, decryptRecords, encryptRecords } from '../utils/encryption.js';
```
TBonnin, line 13: unrelated: just making functions name more consistent

khaliqgant:
> @khaliqgant do I need to pack and install `records` like `utils` in publish.sh?
> 
> https://github.com/NangoHQ/nango/blob/4a40221b7544a7b12780b5d1bbc3b794ec08eee3/scripts/publish.sh#L33-L40

Yeah, based on the fact that it is a dependency of shared. Also need to dynamically add bundleDependencies (see https://github.com/NangoHQ/nango/pull/2001/files#diff-c63759f58a022cf694046a961470c6e28c042d8f221646bc5ddbb5fa959dda75R34) to records based on the `nanoid` issue @bodinsamuel found.

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -148,8 +149,23 @@ class DryRunService {
...
+                _connectionId: number,
+                _model: string,
+                _syncId: string,
+                _generation: number
+                // eslint-disable-next-line @typescript-eslint/require-await
+            ): Promise<string[]> => {
+                return [];
+            }
+        };
+
```
TBonnin, line 165: In order to avoid having `shared` depends on `records` I have deciding to use dependency injection for the 2 cases where `records` must be accessed in `shared`.
One is in `run.service` to mark as deleted records when `track_delete = true`. The second one is in `sync.client` to delete all records when a FULL sync is triggered.

Here for dryrun, since `writeToDb` is false I am passing a mock records service, otherwise I pass the real one (see below in jobs)

packages/shared/package.json
```diff
@@ -65,6 +65,7 @@
         "npm": ">=6.14.11"
     },
     "devDependencies": {
+        "@nangohq/records": "file:../records",
```
TBonnin, line 68: I couldn't entirely avoid making `records` a dependency of `shared` because some `run.service` integration end-to-end tests are accessing the db but at least it is only a dev dependency

packages/records/lib/utils/encryption.ts
```diff
@@ -2,12 +2,20 @@ import type { EncryptedRecordData, FormattedRecord, UnencryptedRecord, Unencrypt
 import { Encryption } from '@nangohq/utils';
 import { envs } from '../env.js';
 
+function getEncryption(): Encryption {
+    const encryptionKey = envs.NANGO_ENCRYPTION_KEY;
+    if (!encryptionKey) {
+        throw new Error('NANGO_ENCRYPTION_KEY is not set');
+    }
+    return new Encryption(encryptionKey);
```
bodinsamuel, line 10: Could be done in the construct I guess

packages/shared/lib/clients/sync.client.ts
```diff
@@ -46,6 +46,10 @@ const OVERLAP_POLICY: ScheduleOverlapPolicy = ScheduleOverlapPolicy.BUFFER_ONE;
 
 const namespace = process.env['TEMPORAL_NAMESPACE'] || 'default';
 
+export interface RecordsServiceInterface {
+    deleteRecordsBySyncId({ syncId }: { syncId: string }): Promise<{ totalDeletedRecords: number }>;
+}
```
bodinsamuel, line 51: Sync it's a type it could live inside records too

packages/records/lib/env.ts
```diff
@@ -2,7 +2,7 @@ import path from 'node:path';
 import { fileURLToPath } from 'node:url';
 import { ENVS, parseEnvs } from '@nangohq/utils';
 
-export const envs = parseEnvs(ENVS.required({ RECORDS_DATABASE_URL: true, NANGO_ENCRYPTION_KEY: true }));
+export const envs = parseEnvs(ENVS.required({ RECORDS_DATABASE_URL: true }));
```
khaliqgant, line 5: Thanks

packages/shared/tsconfig.json
```diff
@@ -4,13 +4,6 @@
...
     },
-    "references": [
-        {
-            "path": "../node-client"
-        },
-        {
-            "path": "../utils"
-        }
-    ],
+    "references": [{ "path": "../node-client" }, { "path": "../utils" }, { "path": "../records" }],
```
bodinsamuel, line 7: we are fighting each other on .json files but prettier should have kicked in here

packages/shared/lib/services/sync/run.service.ts
```diff
@@ -456,12 +463,22 @@ export default class SyncRun {
...
+                deletedKeys = await legacyRecordsService.markNonCurrentGenerationRecordsAsDeleted(
                     this.nangoConnection.id as number,
                     model,
                     this.syncId as string,
                     this.syncJobId as number
                 );
+                this.recordsService
+                    .markNonCurrentGenerationRecordsAsDeleted(this.nangoConnection.id as number, model, this.syncId as string, this.syncJobId as number)
+                    .catch((e) => {
+                        console.error(`Error marking non current generation records as deleted:`, e, {
```
khaliqgant: logger.error ?

packages/persist/lib/controllers/persist.controller.ts
```diff
@@ -126,8 +149,18 @@ class PersistController {
...
+            recordsService
+                .update(records, nangoConnectionId, model)
+                .then((res) => {
+                    if (isErr(res)) {
+                        throw res.err;
+                    }
+                })
+                .catch((reason) => {
+                    logger.error(`Failed to update records: ${reason}`);
+                });
```
bodinsamuel, line 162: imo we should be consistent between async and then/catch, it's confusing to mix the syntax here

bodinsamuel (approved):
Looks good 🙏🏻

khaliqgant (approved):
Couple of non blocking comments

packages/records/lib/utils/encryption.ts
TBonnin: The thing is the constructor is also used by `shared/lib/utils/encryption.manager.ts` and doing so will change its behavior. So for now it is specific to records

packages/persist/lib/controllers/persist.controller.ts
TBonnin: that's only temporary for the double write. Once migration is done, I will cleanup the code and stick with `await`

packages/shared/lib/clients/sync.client.ts
TBonnin: `shared` doesn't depend on `records` :D

packages/shared/lib/services/sync/run.service.ts
TBonnin: https://github.com/NangoHQ/nango/pull/2000/commits/658061dd6b6f87bd2df5e9ae63964666172c089c

packages/shared/tsconfig.json
TBonnin: what's the correct format?

bodinsamuel: the one before your change, lint-staged should have handled this 🤔

packages/shared/lib/clients/sync.client.ts
bodinsamuel: as a devDeps yes so it would have worked for types :D

PR #2005, opened 2024-04-17
fix: type of count is a string

bodinsamuel:
## Describe your changes

- Type is a string apparently bc of max_safe_integer in JS


TBonnin (approved):
That's definitely not intuitive. 🤷

PR #2012, opened 2024-04-18
fix(utils): move metrics

bodinsamuel:
## Describe your changes

Fixes NAN-764

I need metrics for logs, since shared is already dependent of logs I can't have a cyclic import so I took the opportunity to move part of the telemetry file but without side effects (no self instantiation) 

- Move metrics library to `utils`


TBonnin (approved):
we have to be careful to ensure `utils` is not becoming the new `shared`

bodinsamuel:
> we have to be careful to ensure utils is not becoming the new shared

Yep agreed, I just don't think we will have enough material for a full package regarding telemetry.
Also `shared` is big because it contains utils and implementation, I think it's fair if we end up moving all utils inside utils imo. If we find another common denominator (i.e: types, or whatever) happy to make another package ☺️

PR #2014, opened 2024-04-18
docker compose: set valid default value for RECORDS_DATABASE_URL env var

TBonnin:
RECORDS_DATABASE_URL must be a valid url. Setting it to the nango-db container as default


khaliqgant (approved):

bodinsamuel (approved):
ah yes good compromise 👍🏻

PR #2015, opened 2024-04-18
fetch new records and compare with legacy ones (still returning legacy)

TBonnin:
## Describe your changes
Next step in the records migration: reading from new db. First, we read both legacy and new db, comparing the output and logging discrepancy, still returning the legacy output.

Fetching new records is behind a feature flag so it can easily be enabled/disabled

Once we have confirmed there is no discrepancy, I will make a PR to return the output of new db

### DIFF
--- a/packages/server/lib/controllers/sync.controller.ts
+++ b/packages/server/lib/controllers/sync.controller.ts
@@ -220,18 +220,28 @@ class SyncController {
                 if (isErr(newResponse)) {
                     logger.error('Error fetching records from new records service', newResponse.err);
                 } else {
-                    if (JSON.stringify(response) !== JSON.stringify(newResponse.res)) {
-                        logger.error(
-                            `[RECORDS MIGRATION] Differences between legacy and new records: ${JSON.stringify(response)} <<<>>> ${JSON.stringify(newResponse.res)}`,
-                            {
-                                connectionId,
-                                model,
-                                modifiedAfter: delta || modified_after,
-                                limit,
-                                filter,
-                                cursor
-                            }
-                        );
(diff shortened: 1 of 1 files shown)


packages/server/lib/controllers/sync.controller.ts
```diff
@@ -188,11 +195,51 @@ class SyncController {
...
+                    filter: filter as LastAction,
+                    cursor: cursor as string
+                });
+
+                if (isErr(newResponse)) {
+                    logger.error('Error fetching records from new records service', newResponse.err);
+                } else {
+                    if (JSON.stringify(response) !== JSON.stringify(newResponse.res)) {
+                        logger.error(
+                            `[RECORDS MIGRATION] Differences between legacy and new records: ${JSON.stringify(response)} <<<>>> ${JSON.stringify(newResponse.res)}`,
```
TBonnin: It is going to produce very long log lines but it will be very handy for debugging. Since it is behind a flag it can be disabled if too noisy

bodinsamuel: There is a possibility that dates will often conflict no? updated_at, created_at are set by the DB if I'm not mistaken

bodinsamuel (approved):

packages/server/lib/controllers/sync.controller.ts
TBonnin: good point. As discussed, I modified the comparison logic to ignore the metadata: https://github.com/NangoHQ/nango/pull/2015/commits/de9c21919e7963da37419ada987544f75d31e870

PR #2016, opened 2024-04-19
[nan-756] allow to be truly dry

khaliqgant:
## Describe your changes
A sync triggered by an action during a dry run wasn't truly dry. This resolves that.

NAN-756


packages/shared/lib/sdk/sync.ts
```diff
@@ -620,8 +641,17 @@ export class NangoAction {
         return this.nango.triggerAction(providerConfigKey, connectionId, actionName, input) as T;
```
bodinsamuel, line 641: What about triggerAction? If not possible to dryRun can we just log something instead?

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -220,6 +244,10 @@ class DryRunService {
                 }
```
bodinsamuel, line 255: Since you are buffering the output, the prompt will pop before any messages has been printed I think

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -166,6 +183,7 @@ class DryRunService {
         const syncRun = new syncRunService({
             integrationService,
             recordsService,
+            dryRunService: new DryRunService(environment, true),
```
TBonnin, line 196: Sounds weird that a function in `DryRunService` creates a new instance of DryRunService. No?

packages/shared/lib/sdk/sync.ts
khaliqgant: This is something I've mentioned to Bastien quite a few times and is a larger product question and outside the scope of this issue IMO.

packages/cli/lib/services/dryrun.service.ts
khaliqgant: Not sure I follow -- I'm seeing this work as expected.

packages/cli/lib/services/dryrun.service.ts
khaliqgant: Maybe? Do you see this as blocking? Can refactor a bit if you feel strongly about it.

packages/shared/lib/sdk/sync.ts
bodinsamuel: this is an issue because you can wait for the result?

bodinsamuel (approved):

packages/shared/lib/sdk/sync.ts
khaliqgant: Yeah, it is an issue mainly because the whole premise of an action is to interact with an external API.

packages/cli/lib/services/dryrun.service.ts
TBonnin: why not passing `this`?

khaliqgant: because the `returnOutput` instance variable would be different in that case, otherwise we could pass `this`

TBonnin: it is confusing to have two levels where `.run` is being called. It is not blocking though. up to you

packages/shared/lib/sdk/sync.ts
bodinsamuel: make sense

packages/cli/lib/services/dryrun.service.ts
bodinsamuel: yes sorry I thought you removed the console.log 🤦🏻

PR #2017, opened 2024-04-19
records migration: return records from new database

TBonnin:
I won't merge this PR today, monitoring the fetch and compare logs for now but I will appreciate a review.
It is behind a feature flag anyway.

## Describe your changes
Returns records from the new database if flag is enabled (`new-records-return:global`)

### DIFF
--- a/packages/server/lib/controllers/onboarding.controller.ts
+++ b/packages/server/lib/controllers/onboarding.controller.ts
@@ -162,24 +162,19 @@ class OnboardingController {
                 payload.progress = 3;
             }
 
-            let getRecords: Result<CustomerFacingDataRecord[], Error> = resultErr('failed_to_get_records');
+            let getRecords: Result<CustomerFacingDataRecord[], Error>;
             const shouldReturnNewRecords = await featureFlags.isEnabled('new-records-return', 'global', false);
             if (shouldReturnNewRecords) {
-                const { error, response: connection } = await connectionService.getConnection(connectionId, DEMO_GITHUB_CONFIG_KEY, environment.id);
-                if (!error && connection) {
-                    const newGetRecords = await recordsService.getRecords({
-                        connectionId: connection.id as number,
-                        model: DEMO_MODEL
-                    });
-                    if (isOk(newGetRecords)) {
-                        getRecords = resultOk(newGetRecords.res.records);
-                    }
-                }
+                const newGetRecords = await recordsService.getRecords({
+                    connectionId: connectionExists.id,
+                    model: DEMO_MODEL
+                });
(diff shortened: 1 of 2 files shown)


packages/server/lib/controllers/onboarding.controller.ts
```diff
@@ -161,12 +162,31 @@ class OnboardingController {
                 payload.progress = 3;
             }
 
-            const getRecords = await syncDataService.getAllDataRecords(connectionId, DEMO_GITHUB_CONFIG_KEY, environment.id, DEMO_MODEL);
-            if (!getRecords.success) {
+            let getRecords: Result<CustomerFacingDataRecord[], Error> = resultErr('failed_to_get_records');
+            const shouldReturnNewRecords = await featureFlags.isEnabled('new-records-return', 'global', false);
+            if (shouldReturnNewRecords) {
+                const { error, response: connection } = await connectionService.getConnection(connectionId, DEMO_GITHUB_CONFIG_KEY, environment.id);
```
bodinsamuel: Connection is already checked in `checkIfConnectionExists` it returns the id

packages/server/lib/controllers/onboarding.controller.ts
```diff
@@ -161,12 +162,31 @@ class OnboardingController {
                 payload.progress = 3;
             }
 
-            const getRecords = await syncDataService.getAllDataRecords(connectionId, DEMO_GITHUB_CONFIG_KEY, environment.id, DEMO_MODEL);
-            if (!getRecords.success) {
+            let getRecords: Result<CustomerFacingDataRecord[], Error> = resultErr('failed_to_get_records');
```
bodinsamuel: It's hiding the actual error (if any) but maybe it's alright during the transition?

packages/server/lib/controllers/onboarding.controller.ts
```diff
@@ -26,10 +26,11 @@ import {
     createActivityLog,
```
bodinsamuel, line 26: tbf you could migrate the onboarding now

bodinsamuel (commented):
looks okay, but didn't tested locally 👌🏻

packages/server/lib/controllers/onboarding.controller.ts
TBonnin: we don't do anything with the actual error but yeah fixed in 9e577400

packages/server/lib/controllers/onboarding.controller.ts
TBonnin: fixed in 9e577400

bodinsamuel (approved):

PR #2019, opened 2024-04-19
fix(ui): handle long entity names

bodinsamuel:
## Describe your changes

Fixes NAN-674

- Handle long endpoint name
- Handle long integration name
- Handle long connection name

Nb: we are mixing table and display flex which is confusing and do not play well for some css, I changed to use div where necessary

Screenshot 2024-04-19 at 17 12 31
Screenshot 2024-04-19 at 17 12 25
---
Screenshot 2024-04-19 at 17 12 37
Screenshot 2024-04-19 at 17 12 16

Screenshot 2024-04-19 at 17 15 02


TBonnin (approved):
tested locally. looks good

PR #2020, opened 2024-04-19
fix: 'localeCompare is not a function' error

TBonnin:
records id can be a number or a string. Using < and > that applies to both instead of localeCompare that only works for string

### DIFF
--- a/packages/server/lib/controllers/sync.controller.ts
+++ b/packages/server/lib/controllers/sync.controller.ts
@@ -229,13 +229,13 @@ class SyncController {
                             const { _nango_metadata: _ignored, ...rest } = r;
                             return rest;
                         })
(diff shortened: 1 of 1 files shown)


packages/server/lib/controllers/sync.controller.ts
```diff
@@ -229,13 +229,13 @@ class SyncController {
                             const { _nango_metadata: _ignored, ...rest } = r;
                             return rest;
                         })
-                        .sort(({ id: a }, { id: b }) => (a as string).localeCompare(b as string));
+                        .sort(({ id: a }, { id: b }) => (a > b ? 1 : a < b ? -1 : 0));
```
bodinsamuel, line 232: you could have use String() but works too

bodinsamuel (approved):

PR #2023, opened 2024-04-19
fix: batchDelete counts already deleted records or non-existing ones as deleted

TBonnin:
## Describe your changes
Upsert summary would count already deleted records or non-existing ones as deleted

https://linear.app/nango/issue/NAN-727/sync-logs-always-return-all-deleted-records-even-when-they-were

Tested locally

### DIFF
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -194,41 +194,24 @@ export async function upsert(records: FormattedRecord[], connectionId: number, m
         );
     }
 
-    const addedKeys: string[] = [];
-    const updatedKeys: string[] = [];
+    let summary: UpsertSummary = { addedKeys: [], updatedKeys: [], deletedKeys: [], nonUniqueKeys };
     try {
-        const externalIds: { external_id: string }[] = [];
         await db.transaction(async (trx) => {
             for (let i = 0; i < recordsWithoutDuplicates.length; i += BATCH_SIZE) {
                 const chunk = recordsWithoutDuplicates.slice(i, i + BATCH_SIZE);
-                addedKeys.push(...(await getAddedKeys(chunk, connectionId, model)));
-                updatedKeys.push(...(await getUpdatedKeys(chunk, connectionId, model)));
+                const chunkSummary = await getUpsertSummary(chunk, connectionId, model, nonUniqueKeys, softDelete);
+                summary = {
+                    addedKeys: [...summary.addedKeys, ...chunkSummary.addedKeys],
+                    updatedKeys: [...summary.updatedKeys, ...chunkSummary.updatedKeys],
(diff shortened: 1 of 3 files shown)


packages/records/lib/models/records.ts
```diff
@@ -355,49 +338,61 @@ export async function markNonCurrentGenerationRecordsAsDeleted(connectionId: num
...
- * Compute Added Keys
- * @desc for any incoming payload use the provided unique to check against the rows
- * in the database and return the keys that are not in the database
- *
+ * getUpdatedKeys
+ * @desc returns a list of the keys that exist in the records tables but have a different data_hash
  */
-async function getAddedKeys(response: FormattedRecord[], connectionId: number, model: string): Promise<string[]> {
-    const keys: string[] = response.map((record: FormattedRecord) => getUniqueId(record));
+async function getUpdatedKeys(records: FormattedRecord[], connectionId: number, model: string): Promise<string[]> {
```
TBonnin, line 344: consolidating the logic in a single function since there were called together anyway. It allows not having to make a extra query to figure out which keys are going to be deleted. It is the same `knownKeys` query previously used in `getAddedKeys`

packages/shared/lib/services/sync/data/data.service.ts
```diff
@@ -33,11 +33,12 @@ export async function upsert(
 
     const addedKeys = await getAddedKeys(recordsWithoutDuplicates, nangoConnectionId, model);
     const updatedKeys = await getUpdatedKeys(recordsWithoutDuplicates, nangoConnectionId, model);
+    const deletedKeys = await getDeletedKeys(recordsWithoutDuplicates, nangoConnectionId, model);
```
TBonnin, line 36: for legacy records I went the easy route, creating a new function. Given this code is going away very soon I didn't feel like refactoring things to avoid making an extra query.

packages/records/lib/models/records.ts
```diff
@@ -355,49 +338,61 @@ export async function markNonCurrentGenerationRecordsAsDeleted(connectionId: num
...
         .whereIn('external_id', keys)
-        .whereNotIn(['external_id', 'data_hash'], keysWithHash)) as unknown as string[];
+        .pluck('external_id');
 
-    return rowsToUpdate;
+    if (softDelete) {
+        return {
+            addedKeys: [],
+            updatedKeys: [],
+            deletedKeys: nonDeletedKeys,
```
bodinsamuel, line 383: this part is confusing, I don't get how a select `deleted_at:null` can be used for deletedKeys

bodinsamuel (approved):

packages/records/lib/models/records.ts
TBonnin: because this query happens before the deletion so all all known records that are not deleted yet (`select ... from ... where deleted is null and external_id in (list of keys)`) will then be deleted

PR #2024, opened 2024-04-21
[NAN-697] allow repeat script names

khaliqgant:
## Describe your changes
With the CLI organization sync and action names should be able to be repeated. 


models.ts
```diff
@@ -0,0 +1,415 @@
+export interface Contact {
```
bodinsamuel: Should not be here

packages/shared/lib/models/NangoConfig.ts
```diff
@@ -109,6 +109,11 @@ export type NangoSyncEndpoint = {
     [key in HTTP_VERB]?: string;
 };
 
+export enum LayoutMode {
+    ROOT = 'root',
+    NESTED = 'nested'
+}
```
bodinsamuel: personal opinion: enums are most of the time unnecessary, can make splitting code harder since it exports runtime code and force you to import one more thing. I don't see the value over `type LayoutMode = 'root' | 'nested'`

packages/cli/lib/services/compile.service.ts
```diff
@@ -71,7 +71,17 @@ class CompileService {
...
         for (const file of integrationFiles) {
             try {
-                const providerConfiguration = config.find((config) => [...config.syncs, ...config.actions].find((sync) => sync.name === file.baseName));
+                const isNested =
+                    file.inputPath.split('/')[file.inputPath.split('/').length - 2] === 'syncs' ||
+                    file.inputPath.split('/')[file.inputPath.split('/').length - 2] === 'actions';
+
+                let providerConfiguration;
+                if (isNested) {
+                    const providerConfigKey = file.inputPath.split('/')[file.inputPath.split('/').length - 3];
```
bodinsamuel: you are splitting the inputPath three times a row, could be nicer to store this

packages/shared/lib/services/file/local.service.ts
```diff
@@ -6,16 +6,24 @@ import archiver from 'archiver';
...
 const __dirname = dirname(__filename);
 
 class LocalFileService {
-    public getIntegrationFile(syncName: string, setIntegrationPath?: string | null) {
+    public getIntegrationFile(syncName: string, providerConfigKey: string, setIntegrationPath?: string | null) {
         try {
-            const filePath = setIntegrationPath ? `${setIntegrationPath}/dist/${syncName}.${SYNC_FILE_EXTENSION}` : this.resolveIntegrationFile(syncName);
-            const realPath = fs.realpathSync(filePath);
+            const filePath = setIntegrationPath ? `${setIntegrationPath}dist/${syncName}.${SYNC_FILE_EXTENSION}` : this.resolveIntegrationFile(syncName);
+            const fileNameWithProviderConfigKey = filePath.replace(`.${SYNC_FILE_EXTENSION}`, `-${providerConfigKey}.${SYNC_FILE_EXTENSION}`);
```
bodinsamuel, line 19: I don't get this, it will look for `mysync-github.ts`?

packages/cli/lib/services/compile.service.ts
```diff
@@ -71,7 +71,17 @@ class CompileService {
...
+                    providerConfiguration = config.find((config) => config.providerConfigKey === providerConfigKey);
+                } else {
+                    providerConfiguration = config.find((config) => [...config.syncs, ...config.actions].find((sync) => sync.name === file.baseName));
+                }
```
TBonnin, line 84: would storing the result of `file.inputPath.split('/')` and avoid the repetition make the code more readable? ex:

```
const pathSegments = file.inputPath.split('/');
const typeFolder = pathSegments[pathSegments.length - 2];
const isNested = typeFolder === 'syncs' || typeFolder === 'actions';

if (isNested) {
  const providerConfigKey = pathSegments[pathSegments.length - 3];
  providerConfiguration = config.find((config) => config.providerConfigKey === providerConfigKey);
} else {
...
```

TBonnin (approved):

packages/shared/lib/services/file/local.service.ts
khaliqgant: Yes, exactly. This way syncs and actions can have the same name across different providers.

packages/cli/lib/services/compile.service.ts
khaliqgant: Updated

packages/shared/lib/models/NangoConfig.ts
khaliqgant: Fair, updated.

models.ts
khaliqgant: Thanks! Updated.

packages/cli/lib/services/compile.service.ts
khaliqgant: Thanks, updated

bodinsamuel (approved):

PR #2026, opened 2024-04-22
fix(release): unify build 

bodinsamuel:
## Describe your changes

> [!NOTE]
> This is a proposal

Each package has its own slightly different `npm run build` command. It is confusing and actually redundant with the global build command (except for the CLI but hopefully we will be able to remove that at some point). And we will add more packages so we want a something simple that works

- Proposing to reduce to one command  `npm run ts-build`
- Remove unnecessary commands


package.json
```diff
@@ -23,7 +23,7 @@
         "prettier-watch": "onchange './**/*.{ts,tsx}' -- prettier --write {{changed}}",
         "lint": "eslint . --ext .ts,.tsx",
         "lint:fix": "eslint . --ext .ts,.tsx --fix",
-        "ts-build": "tsc -b tsconfig.build.json && npm run postbuild -ws --if-present",
+        "ts-build": "tsc -b tsconfig.build.json && npm run -w nango postbuild",
```
TBonnin, line 26: is it not the same?

bodinsamuel: yeah it is, I just want to clarify it's a one of and the other packages should not rely on it + it's explicit

bodinsamuel:
Do you see any blocker with this @TBonnin ? I want to try with log v2

TBonnin (approved):

PR #2030, opened 2024-04-22
remove legacy records code

TBonnin:
## Describe your changes
Records have been migrated to their own datastore. All logic to interact with records is now in `records` package.

Note: I haven't included the deletion of the `_nango_sync_data_records` table in this PR. First because I am not sure of the impact of deleting a big table like this and I also figure we might want to keep the data around a little bit

https://linear.app/nango/issue/NAN-528/remove-legacy-getrecords-endpoint


packages/server/lib/jobs/index.ts
```diff
@@ -1,8 +0,0 @@
-import { isCloud } from '@nangohq/utils';
-import { encryptionManager } from '@nangohq/shared';
-
-export async function encryptDataRecords(): Promise<void> {
-    if (isCloud) {
-        await encryptionManager.encryptAllDataRecords();
-    }
-}
```
TBonnin, line 8: doesn't seem to be used anywhere. No sure what was it for? @khaliqgant

packages/shared/lib/services/sync/run.service.integration.test.ts
```diff
@@ -209,12 +218,18 @@ describe('SyncRun', () => {
...
 });
 
+const initDb = async () => {
+    await database.multipleMigrations();
+    await migrateRecords();
+};
+
 const clearDb = async () => {
     await db.knex.raw(`DROP SCHEMA nango CASCADE`);
+    // TODO: clear records???
```
TBonnin, line 227: I don't really want to expose records.db outside of the package, especially if it is only used in tests. any idea?

packages/server/lib/jobs/index.ts
khaliqgant: Safe to remove. It was there to encrypt records that weren't encrypted at the beginning of the sync work last year.

packages/shared/lib/services/sync/run.service.integration.test.ts
bodinsamuel: you exposed a migrate function, you can expose a clear function imo

bodinsamuel (approved):
All good 🚀

packages/shared/lib/services/sync/run.service.integration.test.ts
TBonnin: done in [bdf531b](https://github.com/NangoHQ/nango/pull/2030/commits/bdf531b92d32dbc39ce544d770202a0e7174b9b3)

bodinsamuel (approved):

bodinsamuel (approved):

PR #2032, opened 2024-04-23
Add guide to handle large datasets

bastienbeurier:


docs-v2/customize/guides/advanced/handle-large-datasets.mdx
```diff
@@ -4,6 +4,123 @@ sidebarTitle: 'Handle large datasets'
...
+        await nango.batchSave(mapUsers(res.data.members), 'User');
+    }
+}
+```
+
+# Incremental syncing
+
+The preferred method for syncing larger datasets is to fetch only the incremental changes from the external API. This method is known as an ["incremental" sync](https://docs.nango.dev/understand/concepts/syncs#incremental-syncs). 
+
+Nango automatically stores the timestamp of the last sync execution start. You can use this timestamp to instruct the external API to send only the changes since that date. That way, you only receive and persist the modified records in the Nango cache.
```
bodinsamuel, line 48: I would add a mention to `lastSyncDate` prop. It's the key element to achieve incremental syncing and it's not entirely clear how to achieve that by just reading

bodinsamuel (approved):
Massive

docs-v2/customize/guides/advanced/handle-large-datasets.mdx
bastienbeurier: Agreed, I'll update!

PR #2033, opened 2024-04-23
Mention lastSyncDate in docs

bastienbeurier:


bodinsamuel (approved):
🙏🏻

PR #2034, opened 2024-04-23
fix(logs): add more context data

bodinsamuel:
## Describe your changes

Fixes NAN-582

After merging there was a lot of `TODO` left regarding context data. This PR fix most of them, we are still missing the `name`s of all the IDs we are storing but I'll fix that at UI implementation time.

- Add missing context data
- Store activityLogId in `oauth_session` instead of in the log
Not sure why it wasn't done like this before, it's much easier and does not pollute the operation with a tmp field. Maybe there was a good reason though, let me know.

- Reuse `logCtx` appropriately when in try/catch situation


packages/server/lib/controllers/proxy.controller.ts
```diff
@@ -372,15 +370,6 @@ class ProxyController {
...
     }
 
-    /**
-     * Get
-     * @param {Response} res Express response object
-     * @param {NextFuncion} next callback function to pass control to the next middleware function in the pipeline.
-     * @param {HTTP_VERB} method
-     * @param {string} url
-     * @param {ApplicationConstructedProxyConfiguration} config
-     */
```
bodinsamuel, line 382: Those comments were outdated

packages/shared/lib/services/notification/webhook.service.ts
```diff
@@ -302,45 +302,58 @@ class WebhookService {
...
+
+    async forward({
+        integration,
+        account,
+        connectionIds,
+        payload,
+        webhookOriginalHeaders,
+        logContextGetter
+    }: {
+        integration: Config;
```
bodinsamuel, line 318: I replaced the floating values with `integration` easier to Type and provide more info

packages/shared/lib/services/notification/webhook.service.ts
```diff
@@ -302,45 +302,58 @@ class WebhookService {
...
-        providerConfigKey: string,
-        connectionId: string,
-        provider: string,
-        payload: Record<string, any> | null,
-        webhookOriginalHeaders: Record<string, string>,
-        logContextGetter: LogContextGetter
-    ) {
-        const { send, environmentInfo } = await this.shouldSendWebhook(environment_id, { forward: true });
+    async forwardHandler({
+        integration,
```
bodinsamuel, line 342: same

packages/shared/lib/services/sync/run.service.ts
```diff
@@ -742,8 +742,7 @@ export default class SyncRun {
             content
         });
         await this.logCtx?.error(content);
-        // TODO: fix this
-        if (content === 'The script was cancelled successfully') {
+        if (isCancel) {
```
bodinsamuel, line 762: This is not the best fix, but anything else was too much change

packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -233,16 +232,14 @@ class OAuthController {
...
+                webSocketClientId: wsClientId,
+                activityLogId: String(activityLogId)
             };
 
             if (userScope) {
                 session.connectionConfig['user_scope'] = userScope;
             }
 
-            await updateSessionIdActivityLog(activityLogId as number, session.id);
-            // TODO: handle that
```
bodinsamuel, line 244: This is the change related to DB

packages/server/lib/controllers/apiAuth.controller.ts
```diff
@@ -48,12 +49,13 @@ class ApiAuthController {
...
 
         const activityLogId = await createActivityLog(log);
-        const logCtx = await logContextGetter.create(
-            { id: String(activityLogId), operation: { type: 'auth' }, message: 'Authorization API Key' },
-            { account: { id: accountId }, environment: { id: environmentId } }
-        );
 
+        let logCtx: LogContext | undefined;
         try {
+            logCtx = await logContextGetter.create(
```
TBonnin, line 55: if `create` fails no logs would be added but the error will be bubbling up. Not sure that is what we want since logging is very much internal

bodinsamuel: Right now it's safe, it's in the try/catch here, and currently in the client it's all try/catched too. Happy to discuss wether or not we should make it throw globally but right now should be good (unless I missed something) ☺️

TBonnin: I am not discussing the safety of it but the behavior if context creation fails. In this case no logs would be saved obviously but the context creation error will bubble up, which I am not sure it is what we want. Not a big deal though

TBonnin (approved):

packages/server/lib/controllers/apiAuth.controller.ts
bodinsamuel: ah yeah sorry I get it, so right now it's catched anyway in the `create()` function so no worries about that.
ultimately I wanted to remove it, but it's true that we could say it's not a critical piece and could leave it as a silent error

PR #2041, opened 2024-04-24
feat: hn demo

bodinsamuel:
## Describe your changes

fixes NAN-788

- Introduce a temporary frontend endpoint to automatically create an account and redirect to the interactive demo. 
It detects if we are connected to avoid creating new accounts for nothing otherwise it will create an account with a random id and email `demo-<id>@example.com`.
- Banner is displayed when connected to a demo account 
- Bonus: fix the broken script to compile

---

Screenshot 2024-04-24 at 12 38 03


bodinsamuel:
Merging since there is nobody available today, still feel free to drop feedback if any

PR #2045, opened 2024-04-25
fix(records): transaction not reusing opened connection

bodinsamuel:
## Describe your changes

Contributes to NAN-796

There is discrepancy is the new records storage, we have a lot of error in production related to pool exhaustion which is probably due do the fact that the transactions were not reusing the connection correctly.  Not entirely sure it's the full fix but maybe

- Fix transaction usage
- Still increase the pool since we have a better DB

```logs
2024-04-25T08:58:41.431Z [ERROR] [records] Failed to upsert records to table records.
Model: ***, Nango Connection ID: ***.
Attempted to insert/update/delete: 100 records
KnexTimeoutError: Knex: Timeout acquiring a connection. The pool is probably full. Are you missing a .transacting(trx) call?
2024-04-25T08:58:41.431Z [ERROR] [PersistController] Failed to save records: Error: Failed to upsert records to table records.
```

### DIFF
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -15,6 +15,7 @@ import { RECORDS_TABLE } from '../constants.js';
 import { removeDuplicateKey, getUniqueId } from '../helpers/uniqueKey.js';
 import { logger } from '../utils/logger.js';
 import { resultErr, resultOk, type Result } from '@nangohq/utils';
+import type { Knex } from 'knex';
 
 dayjs.extend(utc);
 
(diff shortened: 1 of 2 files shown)


packages/records/lib/models/records.ts
```diff
@@ -199,7 +200,7 @@ export async function upsert(records: FormattedRecord[], connectionId: number, m
         await db.transaction(async (trx) => {
             for (let i = 0; i < recordsWithoutDuplicates.length; i += BATCH_SIZE) {
                 const chunk = recordsWithoutDuplicates.slice(i, i + BATCH_SIZE);
-                const chunkSummary = await getUpsertSummary(chunk, connectionId, model, nonUniqueKeys, softDelete);
+                const chunkSummary = await getUpsertSummary(chunk, connectionId, model, nonUniqueKeys, softDelete, trx);
```
bodinsamuel, line 203: I have been beaten by this so many times I should have caught it in review 😫

khaliqgant:
🙏

PR #2048, opened 2024-04-25
Add demo video to docs

bastienbeurier:


docs-v2/introduction.mdx
```diff
@@ -53,6 +53,10 @@ Nango's flexibility ensures it supports any API integration:
 
 But remember, Nango can work with **any API and any use-case**!
 
+# 📺 Demo video
+
```
bodinsamuel, line 57: Nice, would be cool to link the sample app here maybe?
also in the sample app repo replacing the static image by this video

bodinsamuel: not sure why mintlify didn't trigger here

bodinsamuel (approved):

docs-v2/introduction.mdx
bastienbeurier: Agreed, I'll link the video in the sample app. 

About the link to the sample app, are you proposing to add it somewhere in the introduction page? It's already prominently at the top level of the docs, so IMO as discoverable as we can make it, but let me know what you think.

bodinsamuel: yeah true about the menu 👌🏻

PR #2055, opened 2024-04-26
feat: types package

bodinsamuel:
## Context

> [!NOTE]
> This is a proposal

We need more and more to share types between many packages, back-end, front-end et clients. Currently the shared folder is contains most types but it also contains a lot of side effect and code in general. 

I propose we create a dedicated package --that we publish-- that will contain only types (i.e: db types, api types) that can be reused without fear of any side effects and can be installed as a devdeps.
It would allow us to link `server` with `webapp` but also types the `node-client` removing all discrepancies and copy pasting.

This is a common strategy in TS ecosystem, most libraries do that to some extents 
- GitHub: https://github.com/octokit/types.ts
- AWS: https://github.com/aws/aws-sdk-js-v3/tree/main/packages/types
- Sentry: https://github.com/getsentry/sentry-javascript/tree/develop/packages/types
- Trigger.dev: https://github.com/triggerdotdev/trigger.dev/blob/main/packages/core/src/schemas/api.ts

## Changes

Fixes NAN-810

- New packages `@nangohq/types` 

- One example of how we can link everything with one endpoint `GET /onboarding`


packages/types/lib/api.ts
```diff
@@ -0,0 +1,60 @@
+export interface ApiError<TCode extends string> {
```
TBonnin, line 1: This api types are not being used yet. right?

TBonnin (approved):
❤️

packages/types/lib/api.ts
bodinsamuel: no actually I forgot to use it, thanks!

PR #2058, opened 2024-04-26
fix(eslint): pass #11

bodinsamuel:
## Describe your changes

Continue to fix eslint rules, adding strict-type-checked found a lot of errors just about everywhere. Most of them are autofixed but I fixed some manually.


package.json
```diff
@@ -76,11 +76,7 @@
         "testcontainers": "^9.12.0",
         "tsx": "^4.6.2",
         "typescript": "^5.3.3",
-        "vitest": "^0.33.0"
-    },
-    "dependencies": {
-        "@babel/parser": "^7.22.5",
-        "@babel/traverse": "^7.22.5"
```
bodinsamuel, line 84: I think they were here for the CLI at the very beggining, probably a leftover but I prefer to have a confirmation

packages/server/lib/controllers/config.controller.ts
```diff
@@ -471,17 +471,13 @@ class ConfigController {
...
-            } else {
-                throw new NangoError('provider_config_creation_failure');
-            }
+            void analytics.track(AnalyticsTypes.CONFIG_CREATED, accountId, { provider });
+            res.status(200).send({
+                config: {
+                    unique_key: result.unique_key,
+                    provider
+                }
+            });
```
bodinsamuel, line 480: createEmptyProviderConfig does not return null but throw

packages/cli/package.json
```diff
@@ -68,7 +68,7 @@
         "@types/promptly": "^3.0.2",
         "babel-loader": "^9.1.2",
         "esbuild": "^0.17.19",
-        "vitest": "^1.4.0",
+        "vitest": "^0.33.0",
```
bodinsamuel, line 71: I downgraded vitest because the main one is 0.33.0 and it will validate some types/methods but not run with it, which can lead to confusion

packages/server/lib/middleware/ratelimit.middleware.ts
```diff
@@ -43,11 +42,16 @@ export const rateLimiterMiddleware = (req: Request, res: Response, next: NextFun
...
+        .catch((rateLimiterRes: unknown) => {
+            if (rateLimiterRes instanceof RateLimiterRes) {
+                res.setHeader('Retry-After', Math.floor(rateLimiterRes.msBeforeNext / 1000));
+                setXRateLimitHeaders(rateLimiterRes);
+                logger.info(`Rate limit exceeded for ${key}. Request: ${req.method} ${req.path})`);
+                res.status(429).send('Too Many Requests');
+                return;
+            }
+
+            res.status(500).send('Too Many Requests');
```
TBonnin: should it be `429: Too Many requests` or `500: some other error`?

package.json
khaliqgant: Confirmed, left over from before.

packages/server/lib/middleware/ratelimit.middleware.ts
bodinsamuel: oops yes thanks

TBonnin (approved):

PR #2062, opened 2024-04-29
fix(logs): cron delete old logs

bodinsamuel:
## Describe your changes

Fixes NAN-794

- Cron: delete old logs (15 default retention)


TBonnin (commented):
Can deleting old logs be done by dropping old indexes?

khaliqgant (approved):
Looks good but curious about Thomas' question as well.

bodinsamuel:
Logs are currently stored in a single index, I didn't wanted to over engineer this rn and we often dont know how long a sync run so it would make the query a little more complexe to handle

TBonnin:
> Logs are currently stored in a single index, I didn't wanted to over engineer this rn and we often dont know how long a sync run so it would make the query a little more complexe to handle

I think you can define a wildcard pattern to query over multiple indexes. This way we can have daily indexes (`logs-20240430`, `logs-20240501`, etc...) with a `logs-*` pattern. Couple that with an index lifetime management rules to drop indexes after 14 days.

bodinsamuel:
I messed up this branch :| closing

PR #2063, opened 2024-04-29
feat(logs): POST /logs/search

bodinsamuel:
## Describe your changes

Fixes NAN-846

> [!NOTE]
> This is the proposed change for new API routes, discussed last week

- **New route** `POST /api/v1/logs/search` 
POST because it will require a lot of params that are easier to send with a json body. 

- **Integration test**
This is the first e2e API test. To achieve it I needed an auth method not tied to a session, so when it's in test mode we are using SecretKey auth, let me know if that's alright.

- Add async wrapper to catch errors
Note that it won't be useful anymore after express 5 but it's still a long way before release


packages/logs/lib/models/messages.ts
```diff
@@ -26,19 +27,24 @@ export async function createMessage(row: MessageRow): Promise<void> {
 /**
  * List operations
  */
-export async function listOperations(opts: { limit: number }): Promise<ListOperations> {
+export async function listOperations(opts: { accountId: number; environmentId?: number; limit: number }): Promise<ListOperations> {
+    const q: opensearchtypes.QueryDslQueryContainer = {
```
khaliqgant: Never a fan of single letter variables. If you call it `query` you can then just do

```
body: { query }
```

later

packages/logs/lib/models/messages.ts
```diff
@@ -1,10 +1,11 @@
 import { client } from '../es/client.js';
-import type { MessageRow } from '../types/messages.js';
+import type { MessageRow, OperationRow } from '@nangohq/types';
```
khaliqgant, line 2: Can you explain the reasoning for calling this directory `models`?

khaliqgant (approved):
Small knit and question

docker-compose.yaml
```diff
@@ -46,7 +46,7 @@ services:
             - '${SERVER_PORT:-3003}:${SERVER_PORT:-3003}'
         depends_on:
             - nango-db
-            - nanog-opensearch
```
TBonnin, line 49: sounds like a pokemon name. I like it

packages/server/lib/controllers/v1/logs/searchLogs.ts
```diff
@@ -0,0 +1,41 @@
...
+
+const validation = z
+    .object({
+        limit: z.number().optional().default(100)
+    })
+    .strict();
+
+export const searchLogs = asyncWrapper<SearchLogs>(async (req, res) => {
+    if (!envs.NANGO_LOGS_ENABLED) {
+        res.status(422).send({ error: { code: 'feature_disabled' } });
```
TBonnin: 404. no?

packages/logs/lib/models/messages.ts
bodinsamuel: yeah good point 👍🏻

packages/server/lib/controllers/v1/logs/searchLogs.ts
```diff
@@ -0,0 +1,41 @@
...
+        return;
+    }
+
+    const emptyQuery = requireEmptyQuery(req, { withEnv: true });
```
TBonnin, line 19: not sure why `requireEmptyQuery` is a function where below we do the validation of the body inline, or do you mean to use `requireEmptyBody` below?

packages/server/lib/routes.ts
```diff
@@ -28,21 +28,26 @@ import passport from 'passport';
...
-const webAuth = AUTH_ENABLED
+let webAuth = AUTH_ENABLED
     ? [passport.authenticate('session'), authMiddleware.sessionAuth.bind(authMiddleware), rateLimiterMiddleware]
     : isBasicAuthEnabled
       ? [passport.authenticate('basic', { session: false }), authMiddleware.basicAuth.bind(authMiddleware), rateLimiterMiddleware]
       : [authMiddleware.noAuth.bind(authMiddleware), rateLimiterMiddleware];
 
+if (isTest) {
+    webAuth = apiAuth;
+}
```
TBonnin, line 50: is it for convenience so we don't have to do web auth in the tests?

packages/logs/lib/models/messages.ts
bodinsamuel: it's the data accessors, e.g all access to DBs or Opensearch are here. I initially planned to have more but in the end it's a single model so less useful

TBonnin (approved):

packages/server/lib/controllers/v1/logs/searchLogs.ts
bodinsamuel: yeah I wasn't sure because 404 means it the URL is invalid (basically), it's kind of invalid but also not and doesn't convey the same initial message. Happy to put 404

packages/server/lib/routes.ts
bodinsamuel: yes exactly, so we are "closer" to true e2e tests. That's what motivated me to have the unified context, so there is no difference between the auth

packages/server/lib/controllers/v1/logs/searchLogs.ts
TBonnin: 404 means resource doesn't exist so it fits quite well I think. The error message conveys that is a feature that can be enabled

packages/server/lib/routes.ts
TBonnin: sounds good. Can you add a comment before the if, so in 6months I remember why we have done that

packages/server/lib/controllers/v1/logs/searchLogs.ts
bodinsamuel: `requireEmptyQuery` is a utils to avoid all the endpoints that have no query params to rewrite this logic. 
the body is not empty and specific to this endpoint so no shared logic to reuse.
Don't know if that makes sense

packages/server/lib/routes.ts
bodinsamuel: yeah good point

packages/server/lib/controllers/v1/logs/searchLogs.ts
bodinsamuel: changed 👌🏻

packages/server/lib/routes.ts
bodinsamuel: changed

PR #2064, opened 2024-04-29
fix(sync): maximum call stack for large dataset

bodinsamuel:
## Describe your changes

Fixes NAN-819

- Fix maximum call stack error happening when trying to `batch*()` 
Didn't knew spread syntax could cause this

## Test

```ts
import type { NangoSync, SimpleLarge } from '../../models';

  await nango.log('started');
    const records: SimpleLarge[] = [];

    for (let index = 0; index < 1_000_000; index++) {
      records.push({ id: index });
    }

    await nango.batchSave(records, 'SimpleLarge');
}
```


TBonnin (approved):
TIL

PR #2067, opened 2024-04-29
retry deadlock_detected error when upserting records

TBonnin:
## Describe your changes

Postgres doc advises to retry deadlock failures
https://www.postgresql.org/docs/current/mvcc-serialization-failure-handling.html

This commit also moves `retry` to `packages/utils`

Tested Locally

### DIFF
--- a/packages/utils/lib/retry.ts
+++ b/packages/utils/lib/retry.ts
@@ -1,15 +1,16 @@
 interface RetryConfig {
     maxAttempts: number;
     delayMs: number | ((attempt: number) => number);
+    retryIf: (error: Error) => boolean;
 }
 
 export async function retry<T>(fn: () => T, config: RetryConfig): Promise<T> {
-    const { maxAttempts, delayMs } = config;
+    const { maxAttempts, delayMs, retryIf } = config;
     for (let attempt = 1; attempt <= maxAttempts; attempt++) {
         try {
             return fn();
         } catch (error) {
-            if (attempt < maxAttempts) {
+            if (attempt < maxAttempts && retryIf(error as Error)) {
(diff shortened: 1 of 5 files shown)


packages/utils/lib/retry.ts
```diff
@@ -1,15 +1,16 @@
...
-    const { maxAttempts, delayMs } = config;
+    const { maxAttempts, delayMs, retryIf } = config;
     for (let attempt = 1; attempt <= maxAttempts; attempt++) {
         try {
             return fn();
         } catch (error) {
-            if (attempt < maxAttempts) {
+            if (attempt < maxAttempts && retryIf(error as Error)) {
                 const delay = typeof delayMs === 'number' ? delayMs : delayMs(attempt);
                 await new Promise((resolve) => setTimeout(resolve, delay));
```
bodinsamuel, line 15: side note: we have access to timers/promise that allows us to do `await setTimeout(100)`

bodinsamuel (approved):
👌🏻 Do we have an idea of why? I still don't fully get where it would deadlock. I fear repeating will just throw 3 more errors if the problem is inside the transaction

packages/utils/lib/retry.ts
TBonnin: no sure I follow

bodinsamuel: just saying there is a better syntax but it's not part of this PR

TBonnin:
> 👌🏻 Do we have an idea of why? I still don't fully get where it would deadlock. 

I found some instances of multiple requests received at the same time by `persist`. That doesn't explain all occurrences of the error though so.

> I fear repeating will just throw 3 more errors if the problem is inside the transaction

Same. If problem persists after I merge this we will have more info about it

PR #2069, opened 2024-04-30
[NAN-751] Add migrate command to make migration easier

khaliqgant:
## Describe your changes
Adds a cli migration command `nango migrate-to-directories` to easily change to the nested cli directory structure

NAN-751

### DIFF
--- a/packages/cli/lib/services/migration.service.ts
+++ b/packages/cli/lib/services/migration.service.ts
@@ -33,6 +33,9 @@ export const v1toV2Migration = async (loadLocation: string): Promise<void> => {
 };
 
 async function createDirectory(dirPath: string): Promise<void> {
+    if (fs.existsSync(dirPath)) {
+        return;
+    }
     try {
         await fs.promises.mkdir(dirPath, { recursive: true });
         console.log(chalk.green(`Created directory at ${dirPath}.`));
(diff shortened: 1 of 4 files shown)


packages/cli/lib/services/migration.service.ts
```diff
@@ -0,0 +1,132 @@
...
+        }
+        return;
+    }
+    try {
+        await fs.promises.mkdir(dirPath, { recursive: true });
+        if (debug) {
+            printDebug(`Created directory at ${dirPath}.`);
+        }
+    } catch (error) {
+        console.error(chalk.red(`There was an issue creating the directory at ${dirPath}.`), error);
```
bodinsamuel: this seems to be a critical operation and if it fails everything else will fail. imo it should not be try/catch or stop the processing

packages/cli/lib/services/migration.service.ts
```diff
@@ -0,0 +1,132 @@
...
+async function updateModelImport(filePath: string, debug = false): Promise<void> {
+    try {
+        const data = await fs.promises.readFile(filePath, 'utf8');
+        const updatedData = data.replace(/from '\.\/models/, "from '../../models");
+        await fs.promises.writeFile(filePath, updatedData, 'utf8');
+        if (debug) {
+            printDebug(`Updated imports in ${filePath}.`);
+        }
+    } catch (error) {
+        console.error(chalk.red(`There was an issue updating the imports in ${filePath}.`), error);
```
bodinsamuel, line 74: same

packages/cli/lib/services/migration.service.ts
```diff
@@ -0,0 +1,132 @@
...
+    }
+    try {
+        await fs.promises.rename(source, destination);
+        if (debug) {
+            printDebug(`Moved file from ${source} to ${destination}.`);
+        }
+
+        return true;
+    } catch (error) {
+        console.error(chalk.red(`There was an issue moving the file from ${source} to ${destination}.`), error);
```
bodinsamuel: same

bodinsamuel (approved):
Works great locally 👌🏻 
Maybe a safer way would be to copy file and only delete the original at the end

packages/cli/lib/services/migration.service.ts
khaliqgant: Updated.

packages/cli/lib/services/migration.service.ts
khaliqgant: Updated.

packages/cli/lib/services/migration.service.ts
khaliqgant: This isn't as critical so left this in the try, catch

PR #2071, opened 2024-04-30
[nan-749] default to nested when running nango init

khaliqgant:
## Describe your changes
When setting up a `nango-integrations` directory using the `nango init` command use the updated nested directory structure. The `nango generate` command will now create files using the nested directory structure. An associated documentation task should be merged before this to reduce confusion. 

NAN-749


packages/cli/lib/cli.ts
```diff
@@ -164,7 +165,17 @@ export const generate = async (debug = false, inParentDirectory = false) => {
...
-                fs.writeFileSync(`${dirPrefix}/${name}.ts`, stripped);
+                if (layout_mode === 'root') {
+                    fs.writeFileSync(`${dirPrefix}/${name}.ts`, stripped);
+                } else {
+                    if (!fs.existsSync(`${dirPrefix}/${providerConfigKey}`)) {
+                        fs.mkdirSync(`${dirPrefix}/${providerConfigKey}`);
+                    }
+                    if (!fs.existsSync(`${dirPrefix}/${providerConfigKey}/${type}s`)) {
+                        fs.mkdirSync(`${dirPrefix}/${providerConfigKey}/${type}s`);
+                    }
```
TBonnin: I think `mkdirSync` has a `recursive` argument to create a full path without having to create each folder individually

packages/cli/lib/services/migration.service.ts
```diff
@@ -65,7 +65,7 @@ async function moveFile(source: string, destination: string, debug = false): Pro
 async function updateModelImport(filePath: string, debug = false): Promise<void> {
     try {
         const data = await fs.promises.readFile(filePath, 'utf8');
-        const updatedData = data.replace(/from '\.\/models/, "from '../../models");
+        const updatedData = data.replace(/from (["'])\.\/models/g, 'from $1../../models');
```
TBonnin: why not simply `data.replace(/\.\/models/g, '../../models');` so we don't have to worry about what kind of quotes are being used?

TBonnin (approved):

packages/cli/lib/cli.ts
khaliqgant: Thanks, updated

packages/cli/lib/services/migration.service.ts
khaliqgant: Good call, updated, thanks!

PR #2074, opened 2024-04-30
fix(persist): truncate big logs

bodinsamuel:
## Describe your changes

Fixes NAN-818

Large log can break the UI and make indexing very slow.

- Enforce maximum payload limit (conservative value to avoid breaking too many syncs)
- Truncate log after ~10kb
Which should be enough for customers to log what they want and split their logs if needed. I didn't wanted to parse+batch myself but we could with little bit more effort
- Fix server response when `express.json` errors

## Tests

```ts
import type { NangoSync, SimpleLargeLog } from '../../models';

  await nango.log('started');
  const logs: SimpleLargeLog[] = [];

  for (let index = 0; index < 1_000; index++) {
    logs.push({ id: index });
  }

  await nango.log(logs);
}

```

## Results 
Truncated
Screenshot 2024-04-30 at 16 32 49

Erroring
Screenshot 2024-04-30 at 16 32 54


packages/persist/lib/server.ts
```diff
@@ -61,18 +61,24 @@ const validateRecordsRequest = validateRequest({
     })
 });
 const recordPath = '/environment/:environmentId/connection/:nangoConnectionId/sync/:syncId/job/:syncJobId/records';
-server.post(recordPath, validateRecordsRequest, persistController.saveRecords.bind(persistController));
-server.delete(recordPath, validateRecordsRequest, persistController.deleteRecords.bind(persistController));
-server.put(recordPath, validateRecordsRequest, persistController.updateRecords.bind(persistController));
+server.post(recordPath, express.json({ limit: '100mb' }), validateRecordsRequest, persistController.saveRecords.bind(persistController));
```
khaliqgant: Variable for 100mb

TBonnin (approved):

khaliqgant (approved):

PR #2076, opened 2024-04-30
persist: add authentication

TBonnin:
persist is now verifying that the environment of a request matches the secretKey, making it harder to impersonate runners

https://linear.app/nango/issue/NAN-689/add-secret-key-authentication-for-persist-api

### DIFF
--- a/packages/persist/lib/controllers/persist.controller.ts
+++ b/packages/persist/lib/controllers/persist.controller.ts
@@ -28,7 +28,7 @@ type RecordRequest = Request<
         syncId: string;
         syncJobId: number;
     },
-    any,
+    void,
     {
         model: string;
         records: Record<string, any>[];
@@ -36,12 +36,12 @@ type RecordRequest = Request<
         connectionId: string;
         activityLogId: number;
     },
-    any
+    void
 >;
 
 class PersistController {
     public async saveActivityLog(
-        req: Request<{ environmentId: number }, any, { activityLogId: number; level: LogLevel; msg: string }, any>,
+        req: Request<{ environmentId: number }, void, { activityLogId: number; level: LogLevel; msg: string }, void>,
         res: Response,
         next: NextFunction
     ) {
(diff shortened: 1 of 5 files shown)


packages/persist/lib/server.ts
```diff
@@ -26,6 +27,8 @@ server.use((req: Request, res: Response, next: NextFunction) => {
     }
 });
 
+server.use('/environment/:environmentId/*', authMiddleware);
```
TBonnin, line 30: zod middleware doesn't play well with other middlewares. I tried `server.post(..., [authMiddleware, zodMiddleware], ...)` but got typing errors. 
Therefore using `use()` with a wildcard. Let me know if you can think of a better solution

packages/persist/lib/controllers/persist.controller.ts
```diff
@@ -28,20 +28,20 @@ type RecordRequest = Request<
...
+    void
 >;
 
 class PersistController {
     public async saveActivityLog(
-        req: Request<{ environmentId: number }, any, { activityLogId: number; level: LogLevel; msg: string }, any>,
+        req: Request<{ environmentId: number }, void, { activityLogId: number; level: LogLevel; msg: string }, void>,
         res: Response,
         next: NextFunction
     ) {
```
TBonnin, line 47: unrelated to the PR. Solving some linter warnings

packages/shared/lib/sdk/sync.ts
```diff
@@ -533,7 +533,11 @@ export class NangoAction {
         const response = await persistApi({
             method: 'POST',
             url: `/environment/${this.environmentId}/log`,
+            headers: {
+                Authorization: `Bearer ${this.nango.secretKey}`
+            },
             data: {
+                secretKey: this.nango.secretKey,
```
khaliqgant: Why sending in the body too?

TBonnin: oups. A leftover of my first pass. Fixing

khaliqgant (approved):

PR #2077, opened 2024-05-01
[nan-748] use same method to watch and compile

khaliqgant:
## Describe your changes
When using `nango dev` the compiled file didn't have the `providerConfigKey` appended to the file in the `dist` directory. Consolidate methods so the compile service is used to maintain consistency.

Move public integration templates into nested `syncs` or `actions` directories

NAN-748


TBonnin (approved):

bodinsamuel (approved):
Works well 🚀

PR #2079, opened 2024-05-01
records: return new upsert result + retry upsert

TBonnin:
## Describe your changes

This PR contains 2 commits:
- intermediary step before removing the legacy records logic: still double writing but now returning the result of the upsert to new db
- retrying the entire upsert transaction didn't solve the deadlock_detected error. I am still not sure what's going on, making an attempt at only retrying the upsert query inside the transaction

### DIFF
--- a/packages/persist/lib/controllers/persist.controller.ts
+++ b/packages/persist/lib/controllers/persist.controller.ts
@@ -1,7 +1,7 @@
 import type { NextFunction, Request, Response } from 'express';
-import type { LogLevel, DataResponse, DataRecord, UpsertResponse } from '@nangohq/shared';
+import type { LogLevel, DataResponse, DataRecord } from '@nangohq/shared';
 import { records as recordsService, format as recordsFormatter } from '@nangohq/records';
-import type { FormattedRecord, UnencryptedRecordData } from '@nangohq/records';
+import type { FormattedRecord, UnencryptedRecordData, UpsertSummary } from '@nangohq/records';
 import {
     createActivityLogMessage,
     errorManager,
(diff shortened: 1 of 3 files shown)


packages/records/lib/models/records.ts
```diff
@@ -433,3 +428,11 @@ async function getRecordsByExternalIds(external_ids: string[], connection_id: nu
 
     return result;
 }
+
+function shouldRetryUpsert(error: Error): boolean {
+    if ('code' in error) {
+        const errorCode = (error as { code: string }).code;
+        return errorCode === '40P01'; // deadlock_detected
```
khaliqgant: nit: seems a bit random to have this here

packages/persist/lib/controllers/persist.controller.ts
```diff
@@ -117,17 +110,10 @@ class PersistController {
...
-                .then((res) => {
-                    if (isErr(res)) {
-                        throw res.err;
-                    }
-                })
-                .catch((err: unknown) => {
-                    logger.error(`Failed to delete records: ${stringifyError(err)}`);
-                });
-            return await dataService.upsert(legacyRecords, nangoConnectionId, model, activityLogId, environmentId, true, logCtx);
+            const newUpsert = recordsService.upsert(records, nangoConnectionId, model, true);
```
khaliqgant, line 113: Unclear what the `true` is doing/signaling here. Feel like this is left over from code I probably wrote...

khaliqgant (approved):
Non blocking nits

packages/persist/lib/controllers/persist.controller.ts
TBonnin: it is `softDelete`. Passing a struct with named fields would be better. I will make a PR

packages/records/lib/models/records.ts
TBonnin: I had it inlined in `upsert` originally. I am happy to put it back there if you prefer

khaliqgant: If there will be other error types that we should retry on then I would leave it here. If this is a one off then would inline it. Either way is fine to me tbh

TBonnin: back to being inlined. thx

packages/persist/lib/controllers/persist.controller.ts
TBonnin: addressed in https://github.com/NangoHQ/nango/pull/2080

PR #2090, opened 2024-05-02
fix(api): setup for e2e tests

bodinsamuel:
## Describe your changes

Contributes to NAN-846

> [!NOTE]
> This is setup PR none of the code is used right now but I want to avoid big PR

- Split `server.ts`
Routers needs to be accessible without all the side-effects to be able to load them inside e2e tests

- Add more seeders

- Add test utils that contains a type safe HTTP client

## Tests

*Autocomplete paths*
Screenshot 2024-05-02 at 19 24 08

*Narrow available Methods*
Screenshot 2024-05-02 at 19 24 36

*Type guard errors*
Screenshot 2024-05-02 at 19 32 10

*Type guard success*
Screenshot 2024-05-02 at 19 31 59


packages/server/lib/server.ts
```diff
@@ -86,184 +42,6 @@ if (NANGO_MIGRATE_AT_START === 'true') {
...
-
-// -------
-// Websocket
-const server = http.createServer(app);
-const wss = new WebSocketServer({ server, path: getWebsocketsPath() });
-
-wss.on('connection', async (ws: WebSocket) => {
-    await publisher.subscribe(ws);
-});
-
```
TBonnin, line 266: you said in the PR description that none of the changes are used yet but server/server.ts is modified though. Am I missing something?

TBonnin (approved):

packages/server/lib/server.ts
bodinsamuel: yeah I meant it's changes for e2e tests but it's not used in tests yet

packages/server/lib/routers.ts
```diff
@@ -0,0 +1,227 @@
+import bodyParser from 'body-parser';
```
khaliqgant: Nice, I greatly prefer having a separate file for this too. I typically call it routes which I prefer to routers but non blocking preference.

khaliqgant (approved):
🏆

packages/server/lib/routers.ts
bodinsamuel: yeah make sense, I'll rename it to routes 👌🏻

PR #2093, opened 2024-05-03
[NAN-628] track capping events and clean up logic based on updated specs

khaliqgant:
## Describe your changes
Add tracking events and update spec. 

Rules:
* If integration with > 3 connections and an active script: cannot activate script + deploy allowed, and automatically disabled that script
* If integration with = 3 connections + active scripts: cannot create more connection

NAN-628


packages/shared/lib/services/connection.service.ts
```diff
@@ -1028,8 +1028,9 @@ class ConnectionService {
             return false;
         }
 
-        if (connections.length >= CONNECTIONS_WITH_SCRIPTS_CAP_LIMIT) {
+        if (connections.length > CONNECTIONS_WITH_SCRIPTS_CAP_LIMIT) {
```
khaliqgant, line 1031: A user can have 3 connections with scripts but not more.

packages/shared/lib/hooks/hooks.ts
```diff
@@ -38,6 +41,9 @@ export const connectionCreationStartCapCheck = async ({
 
             if (connections && connections.length >= CONNECTIONS_WITH_SCRIPTS_CAP_LIMIT) {
```
khaliqgant, line 42: If a user has 3 connections already they can't have more.

packages/shared/lib/hooks/hooks.ts
```diff
@@ -16,15 +16,18 @@ import type { Result } from '@nangohq/utils';
...
 
 export const connectionCreationStartCapCheck = async ({
     providerConfigKey,
-    environmentId
+    environmentId,
+    creationType
 }: {
     providerConfigKey: string | undefined;
     environmentId: number;
+    creationType: string;
```
bodinsamuel:
```suggestion
    creationType: "create" | "import";
```

bodinsamuel (approved):

TBonnin (approved):

PR #2094, opened 2024-05-03
[NAN-842] implement unused token refresh

khaliqgant:
## Describe your changes
Adds a cron to refresh stale credentials. Stale credentials in this case are connections that have not been fetch in the past 7 days. With the current number of credentials in prod that meet this criteria with a limit of 500 this will take about 24 days to refresh all of them based on the cron just running once a day. 

Note that this is starting off relatively conservative to track load and also monitor metrics. 

Logic here is intentionally light initially to build upon with what we spoke about in the meeting the other day (with notifications, ability to set refresh interval etc.)

NAN-842


packages/shared/lib/services/connection.service.ts
```diff
@@ -446,6 +446,27 @@ class ConnectionService {
         return result;
     }
 
+    public async getOldConnections({ days, limit }: { days: number; limit: number }): Promise<NangoConnection[]> {
+        const dateThreshold = new Date();
+        dateThreshold.setDate(dateThreshold.getDate() - days);
+
+        const result = await db
+            .knex('_nango_connections')
+            .join('_nango_configs', '_nango_connections.config_id', '_nango_configs.id')
```
bodinsamuel, line 455: not entirely sure it's needed since we should delete connections too, but I would add `and nango_configs.deleted = false` just in case

packages/shared/lib/services/connection.service.ts
```diff
@@ -446,6 +446,27 @@ class ConnectionService {
...
+            .join('_nango_environments', '_nango_connections.environment_id', '_nango_environments.id')
+            .select('connection_id', '_nango_connections.environment_id', 'unique_key as provider_config_key', 'account_id')
+            .where('last_fetched_at', '<', dateThreshold)
+            .orWhere('last_fetched_at', null)
+            .andWhere('_nango_connections.deleted', false)
+            .limit(limit);
```
bodinsamuel, line 461: I think we will constantly pull the same 500 connections if none of them are fetched.
Maybe a new column `last_credentials_refresh_at` could help

packages/jobs/lib/crons/refreshTokens.ts
```diff
@@ -0,0 +1,58 @@
...
+
+        if (typeof account_id !== 'number') {
+            logger.error(`[refreshTokens] connectionId: ${connection_id} is missing account_id`);
+            continue;
+        }
+
+        logger.info(`[refreshTokens] refreshing token for connectionId: ${connection_id}, accountId: ${account_id}`);
+
+        try {
+            await connectionService.getConnectionCredentials(account_id, environment_id, connection_id, provider_config_key, logContextGetter);
```
bodinsamuel: If you don't pass activity_log_id (and a logCtx) nothing will be log wether it fails or succeed, is it fine for the moment?

bodinsamuel (commented):
Looks good, just a few comments maybe not-blocking 👍🏻

packages/shared/lib/services/connection.service.ts
khaliqgant: Yeah, I had that same thought but ended up not implementing it, but can add it in.

packages/jobs/lib/crons/refreshTokens.ts
khaliqgant: Yeah that is fine for the moment in this initial version.

packages/jobs/lib/crons/refreshTokens.ts
```diff
@@ -0,0 +1,58 @@
...
+            metrics.increment(metrics.Types.REFRESH_TOKENS_SUCCESS);
+        } catch (err) {
+            logger.error(`[refreshTokens] failed to refresh token for connectionId: ${connection_id}`);
+            logger.error(err);
```
TBonnin: I would log a single error
```suggestion
            logger.error(`[refreshTokens] failed to refresh token for connectionId: ${connection_id} ${stringifyError(err)}`);
```

packages/shared/lib/services/connection.service.ts
TBonnin: does the script or the `getConnectionCredentials` function update the `last_fetched_at`. If not, the same connections will be selected again

khaliqgant: getConnection updates it

TBonnin: so it is fine then, the script is going to iterate through all the connections not recently fetched recently and update the `last_fetched_at`

bodinsamuel: indeed 👍🏻

khaliqgant: > Yeah, I had that same thought but ended up not implementing it, but can add it in.

so my comment here was referring to connections where a refresh fails. Those failing connections will always have a last_fetched_at that doesn’t change because the connection refresh actually doesn’t succeed

khaliqgant: UPDATE: just checked the logic of getConnectionCredentials and it calls getConnection first and that updates last_fetched_at and then it tries to refresh the credentials if needed. So the above scenario I mentioned shouldn’t happen.

khaliqgant:
TODO: The cron invocation actually needs to move into the server as this was asked for by someone in the community who is not running syncs which means they wouldn't actually be running the jobs service. Additionally this feature will benefit users of just the OAuth side of things.

TBonnin:
> TODO: The cron invocation actually needs to move into the server as this was asked for by someone in the community who is not running syncs which means they wouldn't actually be running the jobs service. Additionally this feature will benefit users of just the OAuth side of things.

jobs is likely to disappear so all crons will have to move somewhere

PR #2096, opened 2024-05-03
fix(Provider): proxy is optional

bodinsamuel:
## Describe your changes

Following the eslint upgrade it now removes unnecessary `?` so it broke my setup locally. Proxy is optional because of `unauthenticated` provider.

- Set `proxy` as optional

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -63,7 +63,7 @@
         "@typescript-eslint/no-base-to-string": "warn",
         "@typescript-eslint/restrict-plus-operands": "warn",
         "@typescript-eslint/consistent-type-exports": "error",
-        "@typescript-eslint/no-unnecessary-condition": "warn",
+        "@typescript-eslint/no-unnecessary-condition": "off", // Our types are not good enough right now
         "@typescript-eslint/only-throw-error": "warn",
         "react/prop-types": "off",
         "@typescript-eslint/no-unused-vars": [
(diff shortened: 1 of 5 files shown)


khaliqgant:
My sense is that the types aren’t always completely correct and so removing the optional ? is going to break more things. The linter gives us false confidence to delete ? willy nilly — so to speak

bodinsamuel:
> My sense is that the types aren’t always completely correct and so removing the optional ? is going to break more things. > The linter gives us false confidence to delete ? willy nilly — so to speak

we can disable the rule for the moment, wdyt?

khaliqgant:
> > My sense is that the types aren’t always completely correct and so removing the optional ? is going to break more things. > The linter gives us false confidence to delete ? willy nilly — so to speak
> 
> we can disable the rule for the moment, wdyt?

Yeah I think so. Don’t want it to bite us

bodinsamuel:
Disabled 👌🏻  but we should aim to have better types, this kind of unnecessary condition also makes the code slower and shaky

khaliqgant (approved):

khaliqgant:
> Disabled 👌🏻 but we should aim to have better types, this kind of unnecessary condition also makes the code slower and shaky

100% agree

khaliqgant:
> My sense is that the types aren’t always completely correct and so removing the optional ? is going to break more things. The linter gives us false confidence to delete ? willy nilly — so to speak

Case in point: https://github.com/NangoHQ/nango/pull/2099 as reported https://nango-community.slack.com/archives/C04ABR352H0/p1714974911347219

PR #2097, opened 2024-05-05
fix(api): unified context

bodinsamuel:
## Context

Fixes NAN-837
Contributes to NAN-846

Current context of the API :
- 5 ways to auth (secret key, public key, session, basic auth, no auth), all of them provide them a slightly different context. 
- Some endpoints enforce environment/account and some others don't
- Some endpoints are not tied to environment (user endpoints)
- Hardly e2e testable because we expect everything to be in session for the dashboard endpoints which would require a cookie

## Proposed changes

- **All auth methods populate `res.locals`** when possible with a unified information payload (account + env)
When not possible to populate the endpoints are made aware they won't receive this payload with `never`

- **Remove all manual calls** and everything that depends on `req.session`
Some are still hidden in nested functions calls that environmentId or accountId 

- **Remove caching**
The caching caused us multiple issues, since we are now getting fully fledge objects (vs id+secret) it increases the potential issues. I think our infra should be able to withstand this load. I added missing index and iterate on this if needed.
The db calls are [0.05ms-1ms[, I expect the volume to be about 2/qps 

> [!WARNING]
> We recently secured Persist with this method, that could benefits from being cached. However I realized it also prevents a secretKey rotation to stop an invalid sync. With this PR it could introduce slowness for sure but also prevents accidental "forever" caching. A better solution would be distributed cache just for the secretKey check or a different function that only checks the key without fetching anything else. Let me know if you want further investment regarding this matter.

## How to test?

Locally we have noAuth by default which can lead to false negative. You should try with authenabled
```diff
diff --git a/packages/utils/lib/environment/detection.ts b/packages/utils/lib/environment/detection.ts
index 784c6769..44346a81 100644
--- a/packages/utils/lib/environment/detection.ts
+++ b/packages/utils/lib/environment/detection.ts
@@ -13,7 +13,7 @@ export const isTest = Boolean(process.env['CI'] !== undefined || process.env['VI
 export const isBasicAuthEnabled = !isCloud && process.env['NANGO_DASHBOARD_USERNAME'] && process.env['NANGO_DASHBOARD_PASSWORD'];
 export const isHosted = !isCloud && !isLocal && !isEnterprise;

-export const AUTH_ENABLED = isCloud || isEnterprise;
+export const AUTH_ENABLED = true || isCloud || isEnterprise;
```

Also try to test secretKey endpoints like `GET /records` and connect to an integration


packages/server/lib/routes.ts
```diff
@@ -159,7 +159,7 @@ web.route('/api/v1/integration/new').post(webAuth, configController.createEmptyP
 web.route('/api/v1/integration/:providerConfigKey').delete(webAuth, configController.deleteProviderConfig.bind(connectionController));
 web.route('/api/v1/integration/:providerConfigKey/connections').get(webAuth, configController.getConnections.bind(connectionController));
 
-web.route('/api/v1/provider').get(configController.listProvidersFromYaml.bind(configController));
+web.route('/api/v1/provider').get(webAuth, configController.listProvidersFromYaml.bind(configController));
```
bodinsamuel, line 162: Unsure why it wasn't protected

packages/shared/lib/services/environment.service.ts
```diff
@@ -156,51 +96,56 @@ class EnvironmentService {
...
+                    // This key is set dynamically and does not exists in database
+                    const env = await db.knex.select<Pick<Environment, 'account_id'>>('account_id').from<Environment>(TABLE).where({ name: envName }).first();
+                    if (!env) {
+                        return null;
+                    }
+
+                    return this.getAccountAndEnvironment({ accountId: env.account_id, envName });
                 }
             }
         }
```
TBonnin, line 119: is this function logic the same  as  `getAccountAndEnvironmentBySecretKey` but with a different key?

packages/server/lib/server.ts
```diff
@@ -39,7 +39,6 @@ if (NANGO_MIGRATE_AT_START === 'true') {
     logger.info('Not migrating database');
 }
 
-await environmentService.cacheSecrets();
```
TBonnin, line 43: I imagine caching secrets was done for a good reason. Is this reason no longer relevant?

TBonnin (commented):
Amazing. It simplify the code a lot and save a bunch of requests. I will let @khaliqgant review/approve since he has way more context about why some things like caching were implemented the way they are in the first place.

Regarding your comment about Persist and caching @bodinsamuel, I prefer to avoid caching unless absolutely necessary because caching is complicated ;) So I think we are good for now.

packages/server/lib/server.ts
bodinsamuel: As stated in the PR desc, caching would now cache not just the ID but the full objects. I'm not entirely confident to do that because of everything that could modify those two objects. I feel like it was done before all our performance improvements, and specifically because secretKey was encrypted (we would have to pull everything from the DB to find the unencrypted key). But now that we query by hash it's not really relevant. 
I do wonder about the added ms, I'm not worrying about API it has fairly low usage, but for persist it might be more costly.


I'll let @khaliqgant complete with more context

packages/shared/lib/services/environment.service.ts
bodinsamuel: yes, I initially planned on refactoring this but reverted while trying to keep the unrelated change to a minimum. I'm sure there is a method that could do both

khaliqgant (approved):
Looks good. I didn't have time to test it locally but will pull it once merged and check normal operations.

> But now that we query by hash it's not really relevant.
I do wonder about the added ms, I'm not worrying about API it has fairly low usage, but for persist it might be more costly.

Would be good to leverage APM to see if there is an uptick in response times. Agreed regarding querying by hashing -- makes the caching of the secretKey less necessary so this change is good 👍

bodinsamuel:
okay let's try 🤞🏻

PR #2105, opened 2024-05-06
[nan-842] update refresh logic based on spec change

khaliqgant:
## Describe your changes
Updated specs for the refresh token logic. 1 day and time cap on 23 hour limit on the cron.

NAN-842


bodinsamuel (approved):

packages/server/lib/refreshTokens.ts
```diff
@@ -28,15 +28,23 @@ export function refreshTokens(): void {
 }
 
 export async function exec(): Promise<void> {
+    const startTime = Date.now();
+    const maxDuration = 23 * 60 * 60 * 1000;
+
     logger.info('[refreshTokens] starting');
 
-    const staleConnections = await connectionService.getOldConnections({ days: 7, limit: 500 });
+    const staleConnections = await connectionService.getOldConnections({ days: 1 });
```
TBonnin: can we still have a limit (even if very high)? Just to make sure any refresh token problem don't cascade into making this call returns a huge number of connections

TBonnin (approved):
would it make sense to run the cron more often (instance of just once a day). Since it is running once a day, a token be stale for almost 48h. Also it will help catching problem way more rapidely.

packages/server/lib/refreshTokens.ts
khaliqgant: Point taken about not returning a huge number of connections. I think I'd rather chunk this actually, will update this tomorrow. Thanks

khaliqgant:
Made some adjustments to use a lock, run every 45 minutes with a limit of 250 connections at a time. Removed time limit since we are now using a lock. This lock also prevents multiple instances of the cron running on the server @bodinsamuel @TBonnin

packages/server/lib/refreshTokens.ts
```diff
@@ -30,28 +30,39 @@ export function refreshTokens(): void {
...
 
-    const staleConnections = await connectionService.getOldConnections({ days: 7, limit: 500 });
+    const today = new Date();
+    const lockKey = parseInt(`${today.getFullYear()}${today.getMonth() + 1}${today.getDate()}`);
 
-    logger.info(`[refreshTokens] found ${staleConnections.length} stale connections`);
+    // Lock to prevent multiple instances of this cron job from running at the same time
+    await db.knex.transaction(async (trx) => {
+        const { rows } = await trx.raw<{ rows: { pg_try_advisory_xact_lock: boolean }[] }>(`SELECT pg_try_advisory_xact_lock(?);`, [lockKey]);
+        if (!rows || rows.length <= 0 || !rows[0]!.pg_try_advisory_xact_lock) {
```
TBonnin: can probably be simplify by using optional chaining `if(!rows?.[0]?.pg_try_advisory_xact_lock)

packages/server/lib/refreshTokens.ts
```diff
@@ -1,13 +1,13 @@
...
+import { db, errorManager, ErrorSourceEnum, connectionService } from '@nangohq/shared';
 import { stringifyError, getLogger, metrics } from '@nangohq/utils';
 import tracer from 'dd-trace';
 import { logContextGetter } from '@nangohq/logs';
 
 const logger = getLogger('Server');
 
 export function refreshTokens(): void {
-    cron.schedule('0 6 * * *', () => {
+    cron.schedule('*/45 * * * *', async () => {
```
TBonnin, line 10: I am curious to know why you picked 45 mins

packages/server/lib/refreshTokens.ts
```diff
@@ -30,28 +30,39 @@ export function refreshTokens(): void {
...
-            logger.error(`[refreshTokens] connectionId: ${connection_id} is missing account_id`);
-            continue;
-        }
+        for (const staleConnection of staleConnections) {
+            const { connection_id, environment_id, provider_config_key, account_id } = staleConnection;
+
+            if (typeof account_id !== 'number') {
+                logger.error(`[refreshTokens] connectionId: ${connection_id} is missing account_id`);
+                continue;
+            }
```
TBonnin: How can a connection not have an account id?

packages/shared/lib/services/connection.service.ts
```diff
@@ -446,19 +446,24 @@ class ConnectionService {
         return result;
     }
 
-    public async getOldConnections({ days, limit }: { days: number; limit: number }): Promise<NangoConnection[]> {
+    public async getOldConnections({ days, limit }: { days: number; limit?: number }): Promise<NangoConnection[]> {
```
TBonnin: why optional?

TBonnin (approved):

packages/server/lib/refreshTokens.ts
khaliqgant: I was trying to use the `NangoConnection` type which had `account_id` as potentially undefined. I updated the method to have an account_id that is set and remove this check with https://github.com/NangoHQ/nango/pull/2105/commits/8596099ad95ab9925fc7e81af0b9731d99741475

packages/shared/lib/services/connection.service.ts
khaliqgant: Left over from the previous iteration. Updated with https://github.com/NangoHQ/nango/pull/2105/commits/8596099ad95ab9925fc7e81af0b9731d99741475

packages/server/lib/refreshTokens.ts
khaliqgant: Arbitrarily. But it was a good middle ground between 30 and 1 hour. With 45 minutes it'll take about 2 days to go through the connections. I want to see how long an iteration takes and can tweak it further if needed.

packages/server/lib/refreshTokens.ts
khaliqgant: Updated

PR #2111, opened 2024-05-07
fix(sdk): add more caching

bodinsamuel:
## Describe your changes

Fixes NAN-783

Not a fan of caching, however when running a sync it should be deterministic and not spam our own backend for no reason. I believe we should fetch all the data beforehand but in the meantime we can cache more stuff.

- **Cache metadata**
It's only a part of connection, there is no reason for those to be unsync. nb: cache is cleared after updating the metadata

- **Cache integration**
It's only used to get webhook_url however we have spike of usage (we have no way to know if they come from us or outside the backend)


TBonnin (approved):

khaliqgant (approved):
Looks good, just verifying that you tested this quite a bit as you know caching issues often come back to bite.

bodinsamuel:
Tested with 
```ts


  await nango.log('started');
  const conn = await nango.getConnection();
  await nango.log(conn);

  const meta1 = await nango.getMetadata();
  await nango.log(meta1);

  await nango.setMetadata({ foo: new Date().toISOString() });

  const meta2 = await nango.getMetadata();
  await nango.log(meta2);

  const conn2 = await nango.getConnection();
  await nango.log(conn2);
}
```

Screenshot 2024-05-10 at 17 51 54




Works as expected and even fix a bug where connection was actually staled (metadata inside the object was not up to date because of the cache)

PR #2112, opened 2024-05-07
fix: latency and type error

bodinsamuel:
## Context

Fixes NAN-894
Fixes NAN-893

With the latest deploy we saw an increase of latency that does not show up in metrics at all (only the global latency increase but no other metrics is showing the same increase). Which lead me to believe its a pool acquisition problem. Basically we are doing more queries so we exhaust the pool more rapidly, the queries are still fast but wait time is increased.

- Increase default pool size
I could have increase the env var, but I think it reflects our expectation.

- Track pool usage
It's a bit more tricky, let me know if you have a better way to do it. 

- Fix match of undefined
It's a followup of the eslint PR that introduced the `unnecessary-condition` rules. Unfortunately `Template` type was not good at all.

### DIFF
--- a/packages/shared/lib/db/database.ts
+++ b/packages/shared/lib/db/database.ts
@@ -1,6 +1,7 @@
 import knex from 'knex';
 import type { Knex } from 'knex';
-import { retry } from '@nangohq/utils';
+import { metrics, retry } from '@nangohq/utils';
+import type { Pool } from 'tarn';
 
 export function getDbConfig({ timeoutMs }: { timeoutMs: number }): Knex.Config {
     return {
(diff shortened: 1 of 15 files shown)


packages/shared/lib/db/database.ts
```diff
@@ -31,6 +32,41 @@ export class KnexDatabase {
...
+        }
+
+        const pool = this.knex.client.pool as Pool<any>;
+        const acquisitionMap = new Map<number, number>();
+
+        setInterval(() => {
+            metrics.gauge(metrics.Types.DB_POOL_USED, pool.numUsed());
+            metrics.gauge(metrics.Types.DB_POOL_FREE, pool.numFree());
+            metrics.gauge(metrics.Types.DB_POOL_WAITING, pool.numPendingAcquires());
+        }, 1000);
```
bodinsamuel, line 50: It's not really precise, but 1s should give us a trend without spamming the metric, lmk

TBonnin (approved):

PR #2113, opened 2024-05-07
chore(persist): add error details in auth.middleware

TBonnin:
## Describe your changes
Trying to debug `unauthorized` error when running in Enterprise mode

### DIFF
--- a/packages/persist/lib/middleware/auth.middleware.ts
+++ b/packages/persist/lib/middleware/auth.middleware.ts
@@ -1,5 +1,6 @@
 import type { Request, Response, NextFunction } from 'express';
 import { environmentService } from '@nangohq/shared';
+import { stringifyError } from '@nangohq/utils';
 
 export const authMiddleware = (req: Request, res: Response, next: NextFunction) => {
     const authorizationHeader = req.get('authorization');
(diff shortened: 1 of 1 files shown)


khaliqgant (approved):

packages/persist/lib/middleware/auth.middleware.ts
```diff
@@ -25,11 +26,11 @@ export const authMiddleware = (req: Request, res: Response, next: NextFunction)
...
             if (!result || result.environment.id !== environmentId) {
-                throw new Error('Unauthorized');
+                throw new Error('Unauthorized: cannot find matching environment');
             }
             next();
         })
-        .catch(() => {
-            res.status(401).json({ error: 'Unauthorized' });
+        .catch((error: unknown) => {
+            res.status(401).json({ error: `Unauthorized ${stringifyError(error)}` });
```
bodinsamuel: it's a detail but it will print `Unauthorized: Unauthorized`

bodinsamuel (approved):

packages/persist/lib/middleware/auth.middleware.ts
TBonnin: good point. fixed.

bodinsamuel (approved):

PR #2114, opened 2024-05-07
fix(auth): add span and cache 🙈

bodinsamuel:
## Describe your changes

I bet the cache was useless, turns it wasn't BUT not for the same reason. When I introduced secretKey hash I picked the best hashing algo to not compromise on security, I forgot that its unbearably slow even on modern hardware. So each query to the public api would take 500ms-2s just to hash the query. The rest was smooth.

- **Add very minimalistic cache** to store hashing results
I could have gone for Redis, but I don't think we are sure to make it a mandatory part of the infra. It does not compromise on security nor caching too much since we still do the query. If an hacker manage to dump or access this memory, then we have bigger problems anyway. Using a less CPU intensive hashing algo could be a liability in security so I'm a bit stuck in between and it seems caching is the recommended way to go around those issues ([interesting read](https://tedspence.com/caching-strategies-for-authentication-8346a040234d))

- Add span on auth methods
That will help us see the auth time in Traces, that was the main thing missing during our debugging session.

### DIFF
--- a/packages/server/lib/middleware/access.middleware.ts
+++ b/packages/server/lib/middleware/access.middleware.ts
@@ -70,7 +70,7 @@ export class AccessMiddleware {
 
     async publicKeyAuth(req: Request, res: Response<any, RequestLocals>, next: NextFunction) {
         const active = tracer.scope().active();
-        const span = tracer.startSpan('secretKeyAuth', {
+        const span = tracer.startSpan('publicKeyAuth', {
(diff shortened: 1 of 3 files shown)


khaliqgant (approved):
back to caching 😅

bodinsamuel:
yeah it's unfortunate, but this time it's by design ahah

PR #2115, opened 2024-05-07
[nan-864] refactor action and sync building

khaliqgant:
## Describe your changes
Started working on a data validation logic for a customer that ultimately was decided against, but did this refactor which was beneficial. With this a `StandardNangoConfig` now includes custom model definitions with the model definitions:

```
# before
"models": [
          {
            "name": "Contact",
            "fields": [
              {
                "name": "id",
                "type": "string"
              },
              {
                "name": "created_at",
                "type": "string"
              },
              {
                "name": "updated_at",
                "type": "string"
              },
              {
                "name": "first_name",
                "type": "string"
              },
              {
                "name": "last_name",
                "type": "string"
              },
              {
                "name": "phone",
                "type": "string"
              },
              {
                "name": "email",
                "type": "string"
              },
              {
                "name": "issueDelete",
                "type": "GithubIssue"
              }
            ]
          }
]

# after
"models": [
          {
            "name": "Contact",
            "fields": [
              {
                "name": "id",
                "type": "string"
              },
              {
                "name": "created_at",
                "type": "string"
              },
              {
                "name": "updated_at",
                "type": "string"
              },
              {
                "name": "first_name",
                "type": "string"
              },
              {
                "name": "last_name",
                "type": "string"
              },
              {
                "name": "phone",
                "type": "string"
              },
              {
                "name": "email",
                "type": "string"
              },
              {
                "name": "issueDelete",
                "type": "GithubIssue"
              }
            ]
          },
          {
            "name": "GithubIssue",
            "fields": [
              {
                "name": "issue",
                "type": "string"
              }
            ]
          }
        ]
```


packages/shared/lib/services/nango-config.service.ts
```diff
@@ -299,240 +296,343 @@ export function convertV2ConfigObject(config: NangoConfigV2, showMessages = fals
...
+            error: builtSyncError,
+            response: builtSyncs
+        } = buildSyncs({ syncs, allModelNames, config, providerConfigKey, showMessages, isPublic, allModels, allEndpoints });
 
-            if (sync.input) {
-                const modelFields = getFieldsForModel(sync.input, config) as { name: string; type: string }[];
-                inputModel.name = sync.input;
-                inputModel.fields = modelFields;
-            }
+        if (!builtSyncSuccess || !builtSyncSuccess) {
```
TBonnin: twice the same condition :D

packages/shared/lib/services/nango-config.service.ts
```diff
@@ -299,240 +296,343 @@ export function convertV2ConfigObject(config: NangoConfigV2, showMessages = fals
...
+            success: builtActionSuccess,
+            error: builtActionError,
+            response: builtActions
+        } = buildActions({ actions, allModelNames, config, providerConfigKey, showMessages, isPublic, allModels, allEndpoints });
 
-                    if (sync.output && Array.isArray(sync.output) && sync.output?.length > 1) {
-                        const error = new NangoError('endpoint_output_mismatch', syncName);
-                        return { success: false, error, response: null };
-                    }
+        if (!builtActionSuccess || !builtActionSuccess) {
```
TBonnin: same here

packages/shared/lib/services/nango-config.service.ts
```diff
@@ -299,240 +296,343 @@ export function convertV2ConfigObject(config: NangoConfigV2, showMessages = fals
...
+    name,
+    type
+}: {
+    integrationData: NangoIntegrationDataV2;
+    allModels: string[];
+    allModelNames: string[];
+    config: NangoConfigV2;
+    name: string;
+    type: 'sync' | 'action';
+}): ServiceResponse<NangoSyncModel[]> {
```
TBonnin, line 353: can we use `Result` from now on for all new functions that can fail?

packages/shared/lib/services/nango-config.service.ts
```diff
@@ -299,240 +296,343 @@ export function convertV2ConfigObject(config: NangoConfigV2, showMessages = fals
...
+    const models: NangoSyncModel[] = [];
+    if (integrationData.output) {
+        const integrationDataReturns = Array.isArray(integrationData.output) ? integrationData.output : [integrationData.output];
+        for (const model of integrationDataReturns) {
+            if (!allModels.includes(model)) {
+                if (!isJsOrTsType(model)) {
```
TBonnin: A single `if` with 2 conditions might be easier to read 
```
if(!allModels.includes(model) && !isJsOrTsType(model)) {
```

packages/shared/lib/services/nango-config.service.ts
```diff
@@ -299,240 +296,343 @@ export function convertV2ConfigObject(config: NangoConfigV2, showMessages = fals
...
+                    allModels.push(model);
+                }
+            } else {
+                if (type === 'sync') {
```
TBonnin: `else if {`?

packages/shared/lib/services/nango-config.service.ts
```diff
@@ -299,240 +296,343 @@ export function convertV2ConfigObject(config: NangoConfigV2, showMessages = fals
...
+                    const error = new NangoError('duplicate_model', { model, name, type: 'sync' });
+                    return { success: false, error, response: null };
                 }
             }
+            const modelFields = getFieldsForModel(model, config) as { name: string; type: string }[];
+            models.push({ name: model, fields: modelFields });
```
TBonnin: Should we still push to `models` when `modelFields` is null

packages/shared/lib/services/nango-config.service.ts
```diff
@@ -299,240 +296,343 @@ export function convertV2ConfigObject(config: NangoConfigV2, showMessages = fals
...
 
-            const scopes = sync?.scopes || sync?.metadata?.scopes || [];
+            if (modelFields) {
+                const subModels = modelFields.filter((field) => allModelNames.some((m) => m.includes(field.type.replace(/\[\]/g, ''))));
 
-            if (!sync?.runs && showMessages) {
-                console.log(chalk.yellow(`No runs property found for sync "${syncName}". Defaulting to every day.`));
+                if (subModels.length > 0) {
```
TBonnin: no need for this `if`. The for loop won't iterate if empty

packages/shared/lib/services/nango-config.service.ts
```diff
@@ -299,240 +296,343 @@ export function convertV2ConfigObject(config: NangoConfigV2, showMessages = fals
...
+                    for (const subModel of subModels) {
+                        const subModelFields = getFieldsForModel(subModel.type, config) as { name: string; type: string }[];
+                        const subModelName = subModel.type.replace(/\[\]/g, '');
+                        models.push({ name: subModelName, fields: subModelFields });
```
TBonnin, line 384: same question about `subModelFields` being possibly nil

packages/shared/lib/services/nango-config.service.ts
```diff
@@ -299,240 +296,343 @@ export function convertV2ConfigObject(config: NangoConfigV2, showMessages = fals
...
-                endpoints = assignEndpoints(actionEndpoint, 'POST', false, showMessages);
-                if (actionEndpoint?.includes('{') && actionEndpoint.includes('}')) {
-                    const { success, error, response } = parseModelInEndpoint(actionEndpoint, allModelNames, inputModel, config);
-                    if (!success || !response) {
-                        return { success, error, response: null };
-                    }
-                    inputModel = response;
-                }
+        if (action.input) {
+            if (action.input.includes('{') && action.input.includes('}')) {
```
TBonnin, line 569: you can use the optional chaining operator and get rid of the first `if` I think: `action.input?...`

TBonnin (approved):

PR #2116, opened 2024-05-07
refactor: better Result types

TBonnin:
- Making Result types returns an instance of the interface so we can use dot notation `res.isOk()/res.isErr()` instead of `isOk(res)/isErr(res)`, avoiding to import the functions every time.
- It also allows us to extends Ok and Err with functions like `.unwrap()` that is very useful in tests. I also needed a `.map()` function recently.
- Shortening `resultOk/resultErr` to `Ok/Err`. imho result prefix doesn't add much and since Result is now used across the codebase adding more context isn't adding much value.
- Renaming `ok.res` to `ok.value` (all the `res.res` were a bit weird) and `err.err` to `err.error`

### DIFF
--- a/packages/persist/lib/controllers/persist.controller.ts
+++ b/packages/persist/lib/controllers/persist.controller.ts
@@ -7,7 +7,7 @@ import tracer from 'dd-trace';
 import type { Span } from 'dd-trace';
 import { logContextGetter, oldLevelToNewLevel } from '@nangohq/logs';
 import type { Result } from '@nangohq/utils';
-import { resultErr, resultOk, isOk, isErr, metrics, stringifyError } from '@nangohq/utils';
+import { Err, Ok, metrics, stringifyError } from '@nangohq/utils';
 
 type persistType = 'save' | 'delete' | 'update';
 type RecordRequest = Request<
@@ -84,10 +84,10 @@ class PersistController {
             softDelete: false,
             persistFunction: persist
         });
-        if (isOk(result)) {
+        if (result.isOk()) {
             res.status(201).send();
         } else {
-            next(new Error(`'Failed to save records': ${result.err.message}`));
+            next(new Error(`'Failed to save records': ${result.error.message}`));
         }
     }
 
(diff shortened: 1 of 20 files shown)


packages/utils/lib/result.ts
```diff
@@ -1,26 +1,51 @@
-export interface ResultRes<T> {
-    ok: true;
-    res: T;
+export interface Left<T, E extends Error> {
```
TBonnin, line 5: sticking with Left/Right convention used by functional programming languages

khaliqgant (approved):

packages/persist/lib/controllers/persist.controller.ts
```diff
@@ -204,33 +204,33 @@ class PersistController {
...
                 environment_id: environmentId,
                 activity_log_id: activityLogId,
-                content: `There was an issue with the batch ${persistType}. ${formatting.err.message}`,
+                content: `There was an issue with the batch ${persistType}. ${formatting.error.message}`,
                 timestamp: Date.now()
             });
-            await logCtx.error('There was an issue with the batch', { error: formatting.err, persistType });
-            const res = resultErr(`Failed to ${persistType} records ${activityLogId}`);
+            await logCtx.error('There was an issue with the batch', { error: formatting.error, persistType });
+            const err = Error(`Failed to ${persistType} records ${activityLogId}`);
```
bodinsamuel:
```suggestion
            const err = new Error(`Failed to ${persistType} records ${activityLogId}`);
```

packages/persist/lib/controllers/persist.controller.ts
```diff
@@ -204,33 +204,33 @@ class PersistController {
...
-            span.setTag('error', res.err).finish();
-            return res;
+            span.setTag('error', err).finish();
+            return Err(err);
         }
 
         const syncConfig = await getSyncConfigByJobId(syncJobId);
         if (syncConfig && !syncConfig?.models.includes(model)) {
-            const res = resultErr(`The model '${model}' is not included in the declared sync models: ${syncConfig.models}.`);
+            const err = Error(`The model '${model}' is not included in the declared sync models: ${syncConfig.models}.`);
```
bodinsamuel:
```suggestion
            const err = new Error(`The model '${model}' is not included in the declared sync models: ${syncConfig.models}.`);
```

bodinsamuel (approved):
Seems convenient 👍🏻 
(still not a huge fan of an uppercase function name)

TBonnin:
> Seems convenient 👍🏻 (still not a huge fan of an uppercase function name)

It is not a function, it's a constructor (sort of) 😉

PR #2117, opened 2024-05-08
debug(persist): differentiate environment not found and not matching in auth middleware

TBonnin:
Still investigating the persist auth error when running nango enterprise.
The commit differentiate the two different errors when environment is not matching in persist auth middleware


TBonnin:
closing. The cause of the issue has been discovered. The `hashSecretKey` of the environment was empty.

PR #2119, opened 2024-05-08
fix: ensure secret_key_hashed is updated when needed

TBonnin:
secret_key_hashed must be set and kept up to date when secret key or NANGO_ENCRYPTION_KEY rotate

https://linear.app/nango/issue/NAN-895/secret-key-hashed-was-empty-for-customer-running-enterprise

### DIFF
--- a/packages/shared/lib/utils/encryption.manager.ts
+++ b/packages/shared/lib/utils/encryption.manager.ts
@@ -7,6 +7,7 @@ import type { Environment } from '../models/Environment.js';
 import type { EnvironmentVariable } from '../models/EnvironmentVariable.js';
 import type { Connection, ApiConnection, StoredConnection } from '../models/Connection.js';
 import db from '../db/database.js';
+import { hashSecretKey } from '../services/environment.service.js';
 
 const logger = getLogger('Encryption.Manager');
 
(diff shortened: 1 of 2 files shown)


khaliqgant (approved):
thanks

bodinsamuel (commented):
I was currently fixing that thanks.
I did not mixed encryption and this because it's not the same thing really, but fair enough

packages/shared/lib/utils/encryption.manager.ts
```diff
@@ -20,14 +21,15 @@ class EncryptionManager extends Encryption {
         return Boolean(this?.key && this.key.length > 0);
     }
 
-    public encryptEnvironment(environment: Environment) {
+    public async encryptEnvironment(environment: Environment) {
         if (!this.shouldEncrypt()) {
             return environment;
```
bodinsamuel, line 26: when no encryption bc of this condition we are now not filling `secret_key_hashed`

TBonnin: you are right. I must leave it there https://github.com/NangoHQ/nango/pull/2119/files#diff-f3d9e285186619d6dfde44bdbbb720119863b18a3a80a41025f4a9c83e13705aL240

bodinsamuel: did you merge without changing?

TBonnin: I am an idiot. I made the change and didn't push it. 🤦‍♂️

PR #2120, opened 2024-05-08
[NAN-851] connection syncs improvements

khaliqgant:
## Describe your changes
Sync connection page improvements. 
Use consistent method to get the schedule. Remove schedule discrepancy logic

## Connection Page Improvements:
- Should be inactive while waiting for a success/failure response 
- Should show a loading state while waiting for a success/failure response 
- Should show a toast in all cases (success & failure) 
- The page should update the sync table accordingly (without having to refresh) — page refreshes every 10 seconds
- if triggering job: update to running THEN completed/failed accordingly (ideally without refreshing)

# Bug Fixes
* Sync connection UI shows a sync is paused when it shouldn't be
* Next sync time is often missing

We were using the `listSchedules` which didn't give reliable schedule information. I noticed when testing that the API sync status endpoint gave back reliable information when the UI didn't so the UI now uses the same temporal API of `describeSchedule` to reliably get the schedule state and the future run times.

NAN-851


packages/shared/lib/services/sync/sync.service.ts
```diff
@@ -283,26 +284,14 @@ export const getSyncs = async (nangoConnection: Connection): Promise<(Sync & { s
...
-        } else if (!schedule?.info?.paused && sync.schedule_status === SyncStatus.PAUSED) {
-            sync = {
-                ...sync,
-                schedule_status: SyncStatus.RUNNING
-            };
-            await updateScheduleStatus(schedule_id, SyncCommand.UNPAUSE, null, nangoConnection.environment_id);
+        const schedule = await syncClient?.describeSchedule(schedule_id);
+        let futureActionTimes: number[] = [];
+        if (schedule) {
+            futureActionTimes = schedule?.info?.futureActionTimes?.map((long) => long.seconds?.toNumber()) as number[];
```
TBonnin: would the value of `futureActionTimes` be `undefined` if `.info` or `.futureActionTimes` is undefined?

packages/shared/lib/services/sync/sync.service.ts
```diff
@@ -283,26 +284,14 @@ export const getSyncs = async (nangoConnection: Connection): Promise<(Sync & { s
...
-                schedule_status: SyncStatus.PAUSED
-            };
-            await updateScheduleStatus(schedule_id, SyncCommand.PAUSE, null, nangoConnection.environment_id);
-        } else if (!schedule?.info?.paused && sync.schedule_status === SyncStatus.PAUSED) {
-            sync = {
-                ...sync,
-                schedule_status: SyncStatus.RUNNING
-            };
-            await updateScheduleStatus(schedule_id, SyncCommand.UNPAUSE, null, nangoConnection.environment_id);
+        const schedule = await syncClient?.describeSchedule(schedule_id);
```
TBonnin, line 290: Make me think our scheduler is going to need this functionality :)

packages/webapp/src/pages/Connection/Show.tsx
```diff
@@ -151,7 +141,7 @@ We could not retrieve and/or refresh your access token due to the following erro
         return <PageNotFound />;
     }
 
-    if (!loaded || !syncLoaded) {
+    if (!loaded || syncLoading) {
```
TBonnin, line 144: wha's the difference between `!loaded` and `syncLoading`

TBonnin (approved):

packages/webapp/src/pages/Connection/Show.tsx
khaliqgant: Loaded is the page frame and meta information like the connection and provider and syncLoading is the actual sync data

packages/shared/lib/services/sync/sync.service.ts
khaliqgant: Yup, indeed

packages/shared/lib/services/sync/sync.service.ts
khaliqgant: Updated this to check for these properties.

TBonnin: you can just default to `|| []` so you don't have to check

khaliqgant: Linter complains when I try to update to this, going to leave as is for now -- unless you have any issues

PR #2124, opened 2024-05-09
feat: [nan-851] improve query to use the nango_config_id

khaliqgant:
## Describe your changes
Reduce the complexity of the code since we now have the `_nango_connections.config_id` value.

I went through the production database to ensure that value is always populated. In some cases it was (for about 86 records) and I manually updated the `config_id` to point to a correct value using the `environment_id` and `provider_config_key` to match. There were two instances where that value is NULL but those connections were created in 2023 and had no matching config

NAN-851


khaliqgant:
@bodinsamuel I know you recently went through this query, so it would be great to get another pair of eyes on this. From my testing the query time is reduced further by using the `config_id`

packages/shared/lib/services/sync/sync.service.ts
```diff
@@ -269,8 +258,13 @@ export const getSyncs = async (nangoConnection: Connection): Promise<(Sync & { s
         .join(SYNC_SCHEDULE_TABLE, function () {
             this.on(`${SYNC_SCHEDULE_TABLE}.sync_id`, `${TABLE}.id`).andOn(`${SYNC_SCHEDULE_TABLE}.deleted`, '=', db.knex.raw('FALSE'));
         })
+        .join(SYNC_CONFIG_TABLE, `${SYNC_CONFIG_TABLE}.sync_name`, `${TABLE}.name`)
```
bodinsamuel: This is bringing back the issue I fixed in #2029

khaliqgant: Not exactly since we're now querying against the `config_id` @ https://github.com/NangoHQ/nango/pull/2124/files#diff-7be9b1d2fee1ed263243cfe652dbb029abdc165631a55b0677119bf8fb3a743cR265

I traced back my local history and was able to recreate the issue fixed in #2029 locally and under the same conditions that problem doesn't appear with this adjustment that makes the query a bit simpler and easier to read IMO

Let me know if you see any issues with this change locally.

bodinsamuel: ah yes didn't saw the where, if you can just add the condition in the join then (like above), its confusing but also change the query plan

khaliqgant: I'm not sure what you mean. There is no join that can happen other than the sync_name

bodinsamuel: Sorry that wasn't clear. Putting the join conditions outside the where can change the query plan (depending on some context) and it makes it harder to see what is exactly we are selecting, e.g: I read too fast here and thought we were joining on name only
```ts
.join(SYNC_SCHEDULE_TABLE, function () {
  this.on(`${SYNC_CONFIG_TABLE}.sync_name`, `${TABLE}.name`)
    .andOn(`${SYNC_CONFIG_TABLE}.deleted `, '=', db.knex.raw('FALSE')). [...]
})

khaliqgant: Updated with https://github.com/NangoHQ/nango/pull/2124/commits/011dc1d6826174aadc3039a583f920b033030745

bodinsamuel (approved):

PR #2125, opened 2024-05-09
[nan-903] add optional integration-id

khaliqgant:
## Describe your changes
If a user has repeated sync names then we (nango) might not know how to resolve the script to run in a dry run

```
integrations:
  provider-1:
   syncs:
     foo:
  provider-2:
    syncs:
      foo:
```

This allows a user to pass in an optional flag of `integration-id` to identify `provider-1` or `provider-2`

```
nango dryrun foo unique-connection --integration-id provider-2
```

NAN-903

### DIFF
--- a/packages/cli/lib/services/dryrun.service.ts
+++ b/packages/cli/lib/services/dryrun.service.ts
@@ -30,7 +30,7 @@ class DryRunService {
 
         this.returnOutput = returnOutput;
     }
-    public async run(options: RunArgs, optionalEnvironment?: string, debug = false): Promise<string | void> {
+    public async run(options: RunArgs, optionalEnvironment?: string, optionalProviderConfigKey?: string, debug = false): Promise<string | void> {
         let syncName = '';
         let connectionId, suppliedLastSyncDate, actionInput, rawStubbedMetadata;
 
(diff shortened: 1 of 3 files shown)


packages/cli/lib/services/dryrun.service.ts
```diff
@@ -30,7 +30,7 @@ class DryRunService {
 
         this.returnOutput = returnOutput;
     }
-    public async run(options: RunArgs, optionalEnvironment?: string, debug = false): Promise<string | void> {
+    public async run(options: RunArgs, optionalEnvironment?: string, optionalProviderConfigKey?: string, debug = false): Promise<string | void> {
```
TBonnin: why are `optionalEnvironment` and `optionalProviderConfigKey` not part of `RunArgs`?

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -75,11 +75,15 @@ class DryRunService {
...
-        const providerConfigKey = config.find((config) => [...config.syncs, ...config.actions].find((sync) => sync.name === syncName))?.providerConfigKey;
+        let providerConfigKey = optionalProviderConfigKey;
 
         if (!providerConfigKey) {
-            console.log(chalk.red(`Provider config key not found, please check that the provider exists for this sync name: ${syncName}`));
-            return;
+            providerConfigKey = config.find((config) => [...config.syncs, ...config.actions].find((sync) => sync.name === syncName))?.providerConfigKey;
+
+            if (!providerConfigKey) {
+                console.log(chalk.red(`Provider config key not found, please check that the provider exists for this sync name: ${syncName}`));
```
TBonnin: How does the user `check that the provider exists`?

packages/cli/lib/services/dryrun.service.ts
khaliqgant: Good question. When I was implementing I thought RunArgs was global to the CLI but it is just local to dry run service. In any case, updated with https://github.com/NangoHQ/nango/pull/2125/commits/4ab52803213d736e0b0e75067e55ff99511d16c9

packages/cli/lib/services/dryrun.service.ts
khaliqgant: Added some copy with https://github.com/NangoHQ/nango/pull/2125/commits/4ab52803213d736e0b0e75067e55ff99511d16c9

TBonnin (approved):
sidenote: I would love if all dryRun params could be named including connection and sync name because I always forgot in which order there are supposed to be passed.

khaliqgant:
> sidenote: I would love if all dryRun params could be named including connection and sync name because I always forgot in which order there are supposed to be passed.

Yeah same. That is a big change though as dry run is a very often used command cc/ @bastienbeurier

PR #2126, opened 2024-05-09
feat(db): add ability to change schema name

bodinsamuel:
## Describe your changes

Contributes NAN-900

- Allow to change the default schema name
For testing purpose, local dev, and also to be able to have multiple install in the same DB. @khaliqgant I believe it could also match your local setup.
I need this to be able to run some specific integration tests


TBonnin:
Does this only apply to the main db or should we do the same for `records` as well?

bodinsamuel:
Good question, I did it while debugging the main db but could be interesting to have it for all db. Let me change that

bodinsamuel:
Can I have a review on this please? ☺️ @TBonnin @khaliqgant

TBonnin (approved):

PR #2128, opened 2024-05-09
[nan-905] fix revert key and allow public keys to be able to be rotated

khaliqgant:
## Describe your changes
Rotateable might not be set. Fixes incorrect route introduced by https://github.com/NangoHQ/nango/pull/1987/files#diff-0199d9c354e383165f793ac468f1114e575c42728bf2a704dce1bb153e9c8288L275

NAN-905


bodinsamuel (approved):
Thanks

PR #2129, opened 2024-05-09
fix: show error details from persist in activity logs

TBonnin:
## Describe your changes
Without the error details it is very hard for customers/us to debug any issue with batchSave/Delete/Update.

https://linear.app/nango/issue/NAN-902/expose-error-details-from-persist-into-activity-logs

### DIFF
--- a/packages/shared/lib/sdk/sync.ts
+++ b/packages/shared/lib/sdk/sync.ts
@@ -566,7 +566,7 @@ export class NangoAction {
 
         if (response.status > 299) {
             logger.error(`Request to persist API (log) failed: errorCode=${response.status} response='${JSON.stringify(response.data)}'`, this.stringify());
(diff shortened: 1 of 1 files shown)


khaliqgant (approved):

packages/shared/lib/sdk/sync.ts
```diff
@@ -566,7 +566,7 @@ export class NangoAction {
 
         if (response.status > 299) {
             logger.error(`Request to persist API (log) failed: errorCode=${response.status} response='${JSON.stringify(response.data)}'`, this.stringify());
-            throw new Error(`Failed to log: ${'error' in response.data ? response.data.error : JSON.stringify(response.data)}`);
+            throw new Error(`Failed to log: ${JSON.stringify(response.data)}`);
```
TBonnin, line 569: simplifying and being consistent with the other ones. I think it is fine if we display the entire response: `{error: message}` instead of just `message`

PR #2130, opened 2024-05-09
[nan-862] add level: 'warn' | 'info' | 'error' so it can be filtered

khaliqgant:
## Describe your changes
Add level property for filtering

NAN-862


TBonnin (commented):
how does that work? is `isError` a ddog logs api field?

khaliqgant:
> how does that work? is `isError` a ddog logs api field?

It isn't anything specific to datadog but rather just allows another tag to filter by. From what I saw in the datadog documentation there isn't an easy way to set an error label without changing the pipeline the log streams to.

TBonnin (commented):
can we have a `level` field instead so we can pass `error | info | warn`?

khaliqgant:
> can we have a `level` field instead so we can pass `error | info | warn`?

Yup, probably a better idea. I'll adjust, thanks.

khaliqgant:
> > can we have a `level` field instead so we can pass `error | info | warn`?
> 
> Yup, probably a better idea. I'll adjust, thanks.

Updated with https://github.com/NangoHQ/nango/pull/2130/commits/aa996499ece6a02dc1a1d7eb0ec549000580fe21

TBonnin (commented):
I would also enforce the acceptable values with something like:
```
    public async log(
        eventId: string,
        message: string,
        operation: string,
        context: Record<string, string> & { level?: 'info' | 'error' | 'warn' } = {},
        additionalTags = ''
    ) {
        const additionalProperties = {
            ...context,
            level: context.level || 'info'
        };
        const params: v2.LogsApiSubmitLogRequest = {
            body: [
                {
                    ddsource: 'web',
                    ddtags: `${eventId}, environment:${process.env['NODE_ENV']}, ${additionalTags}`,
                    message,
                    service: operation,
                    additionalProperties: additionalProperties
                }
            ]
        };

        await this.logInstance?.submitLog(params);
    }

```

PR #2132, opened 2024-05-10
feat(orchestration): introducing the scheduler

TBonnin:
This is the first stage of the orchestration v2, implementing the scheduler.
As agreed the first version of the scheduler doesn't support recurring tasks, only tasks that are scheduled immediately.

`scheduler` package is not used yet but it is big enough to warrant a review. There is still a few TODOs, which I will address in subsequent PRs

https://linear.app/nango/document/[brief]-orchestration-v2-4c5d1f1e104b#94b7a156-de22-4ce2-8a77-9cfcfe1fcece


package.json
```diff
@@ -11,6 +11,7 @@
         "packages/records",
         "packages/server",
         "packages/runner",
+        "packages/scheduler",
```
bodinsamuel, line 14: missing in ./Dockerfile and tsconfig.docker.json  (we need to find a way to detect/automate this)

packages/scheduler/lib/db/migrations/20240506105059_initial_scheduler_models.ts
```diff
@@ -0,0 +1,43 @@
...
+                output jsonb NULL,
+                terminated boolean
+            );
+        `);
+        // TODO: add indexes
+    });
+}
+
+export async function down(knex: Knex): Promise<void> {
+    await knex.raw(`DROP TABLE IF EXISTS "tasks"`);
```
bodinsamuel:
```suggestion
    await knex.raw(`DROP TABLE IF EXISTS "${TASKS_TABLE}"`);
```

packages/scheduler/lib/db/migrations/20240506105059_initial_scheduler_models.ts
```diff
@@ -0,0 +1,43 @@
...
+                'FAILED',
+                'EXPIRED',
+                'CANCELLED'
+            );
+        `);
+        await trx.raw(`
+            CREATE TABLE ${TASKS_TABLE} (
+                id uuid PRIMARY KEY,
+                name varchar(255) NOT NULL,
+                payload jsonb NOT NULL,
```
bodinsamuel: any reason to use jsonb and not just json?

packages/scheduler/package.json
```diff
@@ -0,0 +1,27 @@
...
+    "keywords": [],
+    "repository": {
+        "type": "git",
+        "url": "git+https://github.com/NangoHQ/nango.git",
+        "directory": "packages/scheduler"
+    },
+    "dependencies": {
+        "@nangohq/utils": "file:../utils",
+        "knex": "3.1.0",
+        "uuidv7": "^0.6.3"
```
bodinsamuel:
```suggestion
        "uuidv7": "0.6.3"
```

packages/scheduler/lib/vitest.d.ts
```diff
@@ -0,0 +1,14 @@
+export * from 'vitest';
```
bodinsamuel: Since I synced the vitest package version you should be able to point this in the tsconfig instead of copying

packages/scheduler/lib/models/tasks.ts
```diff
@@ -0,0 +1,249 @@
...
+        const transition = validTaskStateTransitions.find((t) => t.from === from && t.to === to);
+        if (transition) {
+            return Ok(transition);
+        } else {
+            return Err(new Error(`Invalid state transition from ${from} to ${to}`));
+        }
+    }
+};
+
+interface DbTask {
```
bodinsamuel, line 43: maybe belong in types.ts?

packages/scheduler/lib/monitor.worker.ts
```diff
@@ -0,0 +1,103 @@
...
+            }
+        });
+    }
+
+    async start(): Promise<void> {
+        logger.info('Starting monitor...');
+        // eslint-disable-next-line no-constant-condition
+        while (!this.cancelled) {
+            await this.expires();
+            await new Promise((resolve) => setTimeout(resolve, this.tickIntervalMs));
```
bodinsamuel: Suggestion to use native timers

```ts
import setTimeout from 'node:timers/promises'

await setTimeout(this.tickIntervalMs);
```

packages/scheduler/lib/types.ts
```diff
@@ -0,0 +1,34 @@
+import type { TaskProps } from './models/tasks';
+
+type JsonValue = string | number | boolean | null;
+type JsonArray = JsonValue[] | JsonObject[];
+export interface JsonObject {
+    [key: string]: JsonValue | JsonObject | JsonArray;
+}
+
+export const taskStates = ['CREATED', 'STARTED', 'SUCCEEDED', 'FAILED', 'EXPIRED', 'CANCELLED'] as const;
```
bodinsamuel: I'll be careful of exporting const inside a types file, this can produce side effects when importing in other stuff

packages/scheduler/lib/monitor.worker.ts
```diff
@@ -0,0 +1,103 @@
...
+            throw new Error('MonitorWorker should be instantiated in the main thread');
+        }
+    }
+
+    start(): void {
+        this.worker.postMessage('start');
+    }
+
+    stop(): void {
+        this.worker.postMessage('stop');
```
bodinsamuel: stop (or another function) should clean up the listener, the ref and terminate otherwise it will block the graceful termination of the process.

packages/scheduler/lib/monitor.worker.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    stop(): void {
+        logger.info('Stopping monitor...');
+        this.cancelled = true;
+    }
+
+    async expires(): Promise<void> {
+        const expired = await tasks.expiresIfTimeout();
+        if (expired.isErr()) {
+            logger.error(`Error expiring tasks: ${stringifyError(expired.error)}`);
```
bodinsamuel, line 94: that why I'm not a fan of this way of doing things in nodejs, looking at `expiresIfTimeout` you are already logging an error so it will double log

packages/scheduler/lib/monitor.worker.ts
```diff
@@ -0,0 +1,103 @@
...
+export class MonitorWorker {
+    private worker: Worker;
+    constructor() {
+        if (isMainThread) {
+            const url = new URL('../dist/monitor.js', import.meta.url);
+            if (!fs.existsSync(url)) {
+                throw new Error(`Monitor script not found at ${url}`);
+            }
+
+            this.worker = new Worker(new URL('../dist/monitor.js', import.meta.url));
```
bodinsamuel:
```suggestion
            this.worker = new Worker(url);
```

packages/scheduler/lib/monitor.worker.ts
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    async expires(): Promise<void> {
+        const expired = await tasks.expiresIfTimeout();
+        if (expired.isErr()) {
+            logger.error(`Error expiring tasks: ${stringifyError(expired.error)}`);
+        } else {
+            const taskIds = expired.value.map((t) => t.id);
+            if (taskIds.length > 0 && !this.cancelled) {
+                this.send({ ids: taskIds });
```
bodinsamuel, line 98:
```suggestion
                await this.send({ ids: taskIds });
```

bodinsamuel: or remove async from send*

packages/scheduler/lib/scheduler.ts
```diff
@@ -0,0 +1,198 @@
...
+    // TODO: scheduling recurring tasks
+
+    private monitor: MonitorWorker | null = null;
+    private onCallbacks: Record<TaskState, (task: Task) => void>;
+
+    /**
+     * Scheduler
+     * @constructor
+     * @param on - Callbacks for task state transitions
+     * @returns Scheduler
```
bodinsamuel, line 20: I would be very cautious about jsdoc, unless we add an eslint package to enforce their consistency it's not going to be maintained properly

packages/scheduler/lib/scheduler.ts
```diff
@@ -0,0 +1,198 @@
...
+     *        SUCCEEDED: (task: Task) => console.log(`Task ${task.id} succeeded`),
+     *        FAILED: (task: Task) => console.log(`Task ${task.id} failed`),
+     *        EXPIRED: (task: Task) => console.log(`Task ${task.id} expired`),
+     *        CANCELLED: (task: Task) => console.log(`Task ${task.id} cancelled`)
+     *    }
+     * });
+     */
+    constructor({ on }: { on: Record<TaskState, (task: Task) => void> }) {
+        this.onCallbacks = on;
+        if (isMainThread) {
```
bodinsamuel, line 35: should throw if this if is relevant no?

packages/scheduler/lib/scheduler.integration.test.ts
```diff
@@ -0,0 +1,140 @@
...
+        await clearDb();
+    });
+
+    it('mark task as SUCCEEDED', async () => {
+        const task = await scheduleTask(scheduler);
+        (await scheduler.dequeue({ groupKey: task.groupKey, limit: 1 })).unwrap();
+        const succeeded = (await scheduler.succeed({ taskId: task.id, output: { foo: 'bar' } })).unwrap();
+        expect(succeeded.state).toBe('SUCCEEDED');
+    });
+    it('should retry failed task if max retris is not reached', async () => {
```
bodinsamuel:
```suggestion
    it('should retry failed task if max retries is not reached', async () => {
```

bodinsamuel (commented):
Massive 👏🏻 
Did not had time to test everything, but here is my first batch of comments

packages/scheduler/lib/models/tasks.integration.test.ts
```diff
@@ -0,0 +1,172 @@
...
+            })
+        ).unwrap();
+        expect(task).toMatchObject({
+            id: expect.any(String) as string,
+            name: 'Test Task',
+            payload: { foo: 'bar' },
+            groupKey: 'groupA',
+            retryMax: 3,
+            retryCount: 1,
+            startsAfter: expect.any(Date) as Date,
```
bodinsamuel: we could add a special expect with the correct type like I did in shared/lib/vitest.d.ts

packages/scheduler/lib/models/tasks.integration.test.ts
```diff
@@ -0,0 +1,172 @@
...
+        .then((t) => t.unwrap());
+}
+
+async function startTask(props?: Partial<tasks.TaskProps>): Promise<Task> {
+    return createTask(props).then(async (t) => (await tasks.transitionState({ taskId: t.id, newState: 'STARTED' })).unwrap());
+}
+
+function rndString(): string {
+    return (Math.random() + 1).toString(36).substring(2, 5);
+}
```
bodinsamuel, line 172: every functions could be in test.helpers.ts for clarity

packages/scheduler/lib/models/tasks.ts
```diff
@@ -0,0 +1,249 @@
...
+    } catch (err: unknown) {
+        return Err(new Error(`Error updating task ${taskId}: ${stringifyError(err)}`));
+    }
+}
+
+export async function transitionState({ taskId, newState, output }: { taskId: string; newState: TaskState; output?: JsonObject }): Promise<Result<Task>> {
+    if (newState === 'SUCCEEDED' && !output) {
+        return Err(new Error(`Output is required when state = '${newState}'`));
+    }
+    const task = await db.from<DbTask>(TASKS_TABLE).where('id', taskId).first();
```
bodinsamuel: could use get() here

packages/scheduler/lib/models/tasks.ts
```diff
@@ -0,0 +1,249 @@
...
+                    'id',
+                    trx
+                        .select('id')
+                        .from<DbTask>(TASKS_TABLE)
+                        .where({ group_key: groupKey, state: 'CREATED' })
+                        .where('starts_after', '<=', db.fn.now())
+                        .orderBy('created_at')
+                        .limit(limit)
+                        .forUpdate()
+                        .skipLocked()
```
bodinsamuel: There is no global dequeue right now, how it will be managed? 
One while loop per groupKey?

--- 
old comment, nevermind I didn't pay attention of groupKey being passed as param

I would use `RANK OVER()`, 1 groupKey can fill the fullfill `limit`, bringing down every other customers which rank over could handle by default

packages/scheduler/lib/models/tasks.ts
```diff
@@ -0,0 +1,249 @@
...
+            return Ok(sorted);
+        });
+    } catch (err: unknown) {
+        return Err(new Error(`Error dequeuing tasks for group key '${groupKey}': ${stringifyError(err)}`));
+    }
+}
+
+export async function expiresIfTimeout(): Promise<Result<Task[]>> {
+    try {
+        return db.transaction(async (trx) => {
```
bodinsamuel: Any reason to use a transaction here? you are doing everything in one query

bodinsamuel (commented):
Finished reading, I'll let you reply and adjust what you feel relevant. Ping me when you want a second review ☺️

package.json
TBonnin: yep. I always forgot those

TBonnin: yep. Always forgot those

packages/scheduler/lib/db/migrations/20240506105059_initial_scheduler_models.ts
TBonnin: no strong reason behind this choice. `json` could do as well, I don't think we are gonna need to query or update the json content. I am happy to switch to `json` if you feel strongly about it

packages/scheduler/lib/scheduler.ts
TBonnin: It is a risk indeed. Would you prefer no jsdoc?

packages/scheduler/lib/scheduler.ts
TBonnin: yes good idea

packages/scheduler/lib/db/migrations/20240506105059_initial_scheduler_models.ts
bodinsamuel: not strongly, it just takes more space than necessary

packages/scheduler/lib/models/tasks.ts
TBonnin: it is purely internal and should not be expose outside of this file I think.

packages/scheduler/lib/types.ts
TBonnin: actually I don't want `taskStates` to be exposed outside of this package. Moved it to tasks.ts

packages/scheduler/lib/monitor.worker.ts
TBonnin: not sure I follow. There is no logging in `expiresIfTimeout` or in any `tasks` function, errors are always returned

packages/scheduler/lib/models/tasks.ts
bodinsamuel: Discussed, for the moment it will be a global groupKey so no issue

packages/scheduler/lib/monitor.worker.ts
bodinsamuel: yep, misread completely my bad

packages/scheduler/lib/models/tasks.integration.test.ts
TBonnin: moved `vitest.d.ts` to utils and done what you mentioned

packages/scheduler/lib/models/tasks.integration.test.ts
TBonnin: I'll move then if I ever need them somewhere else

packages/scheduler/lib/scheduler.ts
TBonnin: I am keeping the comments for now.

packages/scheduler/lib/db/migrations/20240506105059_initial_scheduler_models.ts
TBonnin: changed to `json`

TBonnin:
I addressed all your comments @bodinsamuel

bodinsamuel (approved):
Let's go 🚀

packages/scheduler/lib/models/tasks.ts
```diff
@@ -0,0 +1,246 @@
...
+                        builder
+                            .orWhere({ state: 'STARTED' })
+                            .andWhere(db.raw(`last_heartbeat_at + heartbeat_timeout_secs * INTERVAL '1 seconds' < CURRENT_TIMESTAMP`));
+                        builder
+                            .orWhere({ state: 'STARTED' })
+                            .andWhere(db.raw(`last_state_transition_at + started_to_completed_timeout_secs * INTERVAL '1 seconds' < CURRENT_TIMESTAMP`));
+                    })
+                    .forUpdate()
+                    .skipLocked()
+                    .debug(true)
```
khaliqgant, line 236: Intentional I presume to leave this in while under development @TBonnin ?

khaliqgant:
Looks good! Tests 💯

packages/scheduler/lib/models/tasks.ts
TBonnin: no actually it is some dev leftover. I will remove it

PR #2133, opened 2024-05-10
test(encryption): add integration test for db encryption

bodinsamuel:
## Describe your changes

- Test `encryptDatabaseIfNeeded` 
Only check environment because that what I was trying to fix

## How to test?

- Removing `NANGO_ENCRYPTION_KEY` from your .env should throw `Error: A previously set NANGO_ENCRYPTION_KEY has been removed from your environment variables.`
- Changing `NANGO_ENCRYPTION_KEY` should throw `Rotation of NANGO_ENCRYPTION_KEY is not supported.`


packages/shared/lib/utils/encryption.manager.ts
```diff
@@ -201,31 +201,57 @@ class EncryptionManager extends Encryption {
...
+        if (status === 'not_started') {
+            logger.info('🔐 Encryption key has been set. Encrypting database...');
+            await this.saveDbConfig({ encryption_key_hash: encryptionKeyHash, encryption_complete: false });
+        }
+        if (status === 'incomplete') {
+            logger.info('🔐 Previously started database encryption is incomplete. Continuing encryption of database...');
+        }
+
+        await this.encryptDatabase();
+        await this.saveDbConfig({ encryption_key_hash: encryptionKeyHash, encryption_complete: true });
```
TBonnin, line 254: unrelated to your changes but all the queries in `encryptDatabase` and `saveDbConfig` should probably be inside a transaction.

TBonnin (approved):

packages/shared/lib/utils/encryption.manager.ts
bodinsamuel: At first glance yes, I'm just wondering about the size of the transaction if you have a large db. I think the script assume it can fail at any time and try everything from the start every time which is nice imo

PR #2135, opened 2024-05-10
feat(webapp): [nan-839] add secondary url

khaliqgant:
## Describe your changes
Add the ability to add a secondary webhook url. I purposely treated it as a another column to reduce complexity and to set a hard limit of the number of webhooks we allow customers to send to. Two should be the max IMO.

Also fixes environment name instead of the incorrect `account`

NAN-839


TBonnin:
what's the purpose of a secondary webhook url?

khaliqgant:
> what's the purpose of a secondary webhook url?

Request from customer https://nangohq.slack.com/archives/C05QZ7A2MPB/p1713926064382129

TBonnin:
I personally don't think we should support this. Why can't the customer do the routing on their side. They still gonna need to have some logic to filter based on the connection. This filtering could route to their correct destination.

packages/shared/lib/db/migrations/20240509171724_add_webhook_secondary.cjs
```diff
@@ -0,0 +1,11 @@
+exports.up = async function (knex, _) {
+    return knex.schema.alterTable('_nango_environments', function (table) {
+        table.text('webhook_url_secondary');
```
bodinsamuel, line 3: I think I would make it an array, because if we have 2 why not three or more

packages/server/lib/controllers/environment.controller.ts
```diff
@@ -80,7 +80,7 @@ class EnvironmentController {
             const environmentVariables = await environmentService.getEnvironmentVariables(environment.id);
 
             res.status(200).send({
-                account: { ...environment, env_variables: environmentVariables, host: baseUrl, uuid: account.uuid, email: user.email }
+                environment: { ...environment, env_variables: environmentVariables, host: baseUrl, uuid: account.uuid, email: user.email }
```
bodinsamuel, line 83: I think neither are true, account and env are mixed (and email?). If we really want to refactor, it would be great to have two or more objects

bodinsamuel (commented):
Works as intended (at least the UI and API) 🙏🏻 

I think it's one of those ideas where it's super simple in theory and full of complexity in practice. In a cloud install, you will need 2 wehbooks url (one for your own staging and one for a dev) and maybe 3 because there are actually two devs and maybe 4 because you also need zapier, etc.
And there is no way to differentiate webhooks, so when a dev is playing the others url are also receiving stuff that might not exists on their own backend leading to bug or confusion. 
Not easy, I think it kinda works like this so I'll approve after the few comments but it can only be a short-term solution imo

packages/server/lib/controllers/environment.controller.ts
khaliqgant: Environment is more correct IMO, but agree it should be refactored into separate objects at some point.

packages/shared/lib/db/migrations/20240509171724_add_webhook_secondary.cjs
khaliqgant: Fair, but the goal was to limit this to two.

khaliqgant:
Thanks @TBonnin -- understand your perspective here.

PR #2136, opened 2024-05-10
feat(metrics): [NAN-664] track more usage metrics in datadog

khaliqgant:
## Describe your changes
Add additional metric tracking

NAN-664


packages/utils/lib/telemetry/metrics.ts
```diff
@@ -1,41 +1,45 @@
 import tracer from 'dd-trace';
 
 export enum Types {
```
khaliqgant, line 3: alphabetize

packages/server/lib/controllers/connection.controller.ts
```diff
@@ -259,9 +260,14 @@ class ConnectionController {
             const providerConfigKey = req.query['provider_config_key'] as string;
             const returnRefreshToken = req.query['refresh_token'] === 'true';
             const instantRefresh = req.query['force_refresh'] === 'true';
+            const isSync = (req.get('Nango-Is-Sync') as string) === 'true';
 
             const action = LogActionEnum.TOKEN;
 
+            if (!isSync) {
+                metrics.increment(metrics.Types.GET_CONNECTION, 1, { accountId });
```
TBonnin, line 268: does that create a metric per account? If yes are we ok about the cost increase? cc @bastienbeurier 
Same question about the other ones below

bodinsamuel (approved):
Looks good, same comment as @TBonnin

packages/server/lib/controllers/connection.controller.ts
khaliqgant: Yes. The description of the task states:

> Because we pay per number of metrics, we should track 0-1 dimensions for each metric. I would only track accountID.

PR #2141, opened 2024-05-10
fix(release): publish correctly to github

bodinsamuel:
## Describe your changes

- Fix commit to GitHub
Turns out we don't have access to the full repo by default, have to fetch and checkout the proper branch and setup an author. You can check the commit it created on this branch https://github.com/NangoHQ/nango/pull/2141/commits/4ab23d2861881734717397cd2df41118847c9717


scripts/gitrelease.mjs
```diff
@@ -14,10 +14,24 @@ if (tagExists.stdout !== '') {
 }
 
 const releaseMessage = `chore(release): ${nextVersion} [skip ci]`;
+
+echo`Checkout out branch`;
+await $`git fetch`;
```
TBonnin: we can probably speed things up by only fetching the branch we are interested in

TBonnin (approved):
small comment about git fetch but looks good otherwise

scripts/gitrelease.mjs
bodinsamuel: indeed!

PR #2145, opened 2024-05-13
feat(api): [NAN-793] bulk metadata update api

khaliqgant:
## Describe your changes
Update the set and update metadata endpoints to accept an array of connection ids. If any invalid connection is passed in at any point in the array the entire operation is bailed and no metadata is set.

NAN-793


docs-v2/mint.json
```diff
@@ -522,7 +522,9 @@
                                 "reference/api/connection/get",
                                 "reference/api/connection/post",
                                 "reference/api/connection/set-metadata",
+                                "reference/api/connection/bulk-set-metadata",
```
bastienbeurier: Do we need new endpoints for this? Can we just support both a single value and an array in the existing routes?

docs-v2/reference/sdks/node.mdx
```diff
@@ -459,6 +459,50 @@ await nango.setMetadata('<INTEGRATION-ID>', 'CONNECTION-ID', { 'CUSTOM_KEY1': 'C
 **Response**
 
 Empty response.
+<Expandable>
```
bastienbeurier: Same thing, maybe this can be part of the docs of the existing method?

docs-v2/mint.json
khaliqgant: That what I would have liked to do but we can't without deprecating this current method. This is because the `connectionId` is passed as part of the URL: 

```
curl --location --request PATCH 'http://localhost:3003/connection/to-override/metadata?provider_config_key=google-calendar' \
--data '{
    "myMetadata": "bar"
}'
```

docs-v2/reference/sdks/node.mdx
khaliqgant: See above comment

packages/shared/lib/services/connection.service.ts
```diff
@@ -492,6 +496,13 @@ class ConnectionService {
         return newMetadata;
     }
 
+    public async bulkUpdateMetada(connections: Connection[], metadata: Metadata): Promise<void> {
```
TBonnin: can we not refactor `updateMetadata` to accept a array of connection instead of adding another public method?

packages/shared/lib/services/connection.service.ts
```diff
@@ -477,6 +477,10 @@ class ConnectionService {
             .update({ metadata });
     }
 
+    public async bulkReplaceMetadata(ids: number[], metadata: Metadata) {
```
TBonnin: can we refactor `replaceMetadata` to accept a array of connections?

docs-v2/reference/sdks/node.mdx
```diff
@@ -459,6 +459,50 @@ await nango.setMetadata('<INTEGRATION-ID>', 'CONNECTION-ID', { 'CUSTOM_KEY1': 'C
...
+**Response**
+
+<Expandable>
+```json
+{
+    "connection_ids": "<string[]>",
+    "metadata": {
+        "CUSTOM_KEY1": "CUSTOM_VALUE1"
+    }
+}
```
bodinsamuel, line 472: What if I want to bulk but with different metadata?
I think we could do something like this:
```sh
curl -XPOST /connection/metadata
{ 
  metadatas: [ 
    { connectionId: string, metadata: Record<string, any> },
    { connectionId: string, metadata: Record<string, any> } 
  ]
}
```

docs-v2/reference/sdks/node.mdx
```diff
@@ -487,6 +531,50 @@ await nango.updateMetadata('<INTEGRATION-ID>', 'CONNECTION-ID', { 'CUSTOM_KEY1':
...
+        The custom metadata to store in the connection.
+    </ResponseField>
+</Expandable>
+
+**Response**
+
+<Expandable>
+```json
+{
+    "connection_ids": "<string[]>",
```
bodinsamuel: I feel like it's clearer like, wdyt?
```suggestion
    "connection_ids": ["<string>"],
```

docs-v2/reference/sdks/node.mdx
```diff
@@ -487,6 +531,50 @@ await nango.updateMetadata('<INTEGRATION-ID>', 'CONNECTION-ID', { 'CUSTOM_KEY1':
 **Response**
 
 Empty response.
+<Expandable>
+```json
+{
+    "CUSTOM_KEY1": "CUSTOM_VALUE1"
+}
```
bodinsamuel: I don't get why we should answer the metadata?

packages/server/lib/controllers/connection.controller.ts
```diff
@@ -506,6 +507,93 @@ class ConnectionController {
...
+                    const error = new NangoError('unknown_connections_bailed', { connectionId, providerConfigKey, environmentName: environment.name });
+                    errorManager.errResFromNangoErr(res, error);
+
+                    return;
+                }
+
+                validConnections.push(connection);
+            }
+
+            await connectionService.bulkUpdateMetada(validConnections, metadata);
```
bodinsamuel: could have been a transaction instead of recreating a new method

packages/server/lib/controllers/connection.controller.ts
```diff
@@ -506,6 +507,93 @@ class ConnectionController {
         }
     }
 
+    async bulkSetMetadata(req: Request, res: Response<any, Required<RequestLocals>>, next: NextFunction) {
+        try {
+            const environment = res.locals['environment'];
+            const providerConfigKey = (req.params['provider_config_key'] as string) || (req.get('Provider-Config-Key') as string);
```
bodinsamuel: I don't see how it could be in the params?

docs-v2/spec.yaml
```diff
@@ -559,6 +562,96 @@ paths:
                                 properties:
                                     message:
                                         type: string
+    /connection/metadata/bulk:
+        post:
+            description: Set custom metadata for the connections.
+            parameters:
+                - name: Provider-Config-Key
+                  in: header
```
bodinsamuel:
```suggestion
                  in: query
```

docs-v2/spec.yaml
```diff
@@ -559,6 +562,96 @@ paths:
...
+                            schema:
+                                type: object
+                                properties:
+                                    message:
+                                        type: string
+        patch:
+            description: Edit custom metadata for the connections (only overrides specified properties, not the entire metadata).
+            parameters:
+                - name: Provider-Config-Key
+                  in: header
```
bodinsamuel:
```suggestion
                  in: query
```

packages/shared/lib/services/connection.service.ts
khaliqgant: Sure: https://github.com/NangoHQ/nango/pull/2145/commits/294aa89bed2aa5984e8072081dbb096eb986040e

packages/shared/lib/services/connection.service.ts
khaliqgant: Sure: https://github.com/NangoHQ/nango/pull/2145/commits/294aa89bed2aa5984e8072081dbb096eb986040e

docs-v2/reference/sdks/node.mdx
khaliqgant: Seems out of scope of the issue.

docs-v2/reference/sdks/node.mdx
khaliqgant: Yeah, updated with https://github.com/NangoHQ/nango/pull/2145/commits/9d3557546cfc71fb5dfd2782a2decca09a5105b7

docs-v2/reference/sdks/node.mdx
khaliqgant: Serves as confirmation of successful receipt

packages/server/lib/controllers/connection.controller.ts
khaliqgant: Yeah, but wanted to be more explicit here though with the connection validity and.

packages/server/lib/controllers/connection.controller.ts
khaliqgant: I don't either 😂. Updated with https://github.com/NangoHQ/nango/pull/2145/commits/982847dcca41db7ced7a3cc83234a9adaba80526

docs-v2/spec.yaml
khaliqgant: https://github.com/NangoHQ/nango/pull/2145/commits/19af738044a332b4741e4b99dec2e72c66e88cce

docs-v2/spec.yaml
khaliqgant: https://github.com/NangoHQ/nango/pull/2145/commits/19af738044a332b4741e4b99dec2e72c66e88cce

docs-v2/mint.json
bastienbeurier: Got it. In this case, I would aim to make this endpoint the new standard, and later deprecate the other one, so we keep a minimal number of endpoints. 

In order to do so, I would remove `bulk` from the endpoint and have something like: 
`PATCH https://api.nango.dev/connection/update-metadata`
`POST https://api.nango.dev/connection/set-metadata`

What do you think?

TBonnin: that's logical. It is mirroring the endpoints to modify metadata for a single connection

```
// single connection
POST /connection/:connectionId/metadata
PATCH /connection/:connectionId/metadata

// multiple connection
POST /connection/metadata
PATCH /connection/metadata
```

packages/server/lib/controllers/connection.controller.ts
bodinsamuel: I think you might have modified the wrong endpoint 
edit: nevermind, you just forgot this one

bodinsamuel (approved):
👌🏻 
It would have been great to use the new pattern of creating routes, just to have some payload validation and test but not blocking

khaliqgant:
```
# POST: set metadata
curl --location '/connection/metadata?provider_config_key=unauthenticated' \
--data '{
    "metadata": {
        "my_metaddata": "no"
    },
    "connection_id": ["one", "two-connections"]
}'

# PATCH update metadata
curl --location --request PATCH '/connection/metadata?provider_config_key=unauthenticated' \
--data '{
    "metadata": {
        "updated-metadata": "meta"
    },
    "connection_id": ["one"]
}'
```

The older metadata endpoints still exist but now exist in the documentation marked as legacy.

One more pass is requested to the interface and implementation @bastienbeurier @TBonnin @bodinsamuel

docs-v2/spec.yaml
```diff
@@ -486,6 +486,96 @@ paths:
                                 properties:
                                     message:
                                         type: string
+    /connection/metadata:
+        post:
+            description: Set custom metadata for the connection or connections. If updating multiple connections pass in an array of connection ids instead of a string.
+            parameters:
+                - name: provider_config_key
```
bastienbeurier: Should we keep a query param in a POST? I feel is not standard and it's an opportunity to change it.

docs-v2/spec.yaml
```diff
@@ -486,6 +486,96 @@ paths:
                                 properties:
                                     message:
                                         type: string
+    /connection/metadata:
```
bastienbeurier, line 489: I like the simplicity of the endpoint, but a slight concern is that only the HTTP method (POST vs PATCH) differ when updating vs. overriding metadata. Users could make a mistake more easily and override metadata when they intended to update it, which can be disastrous.

But this might be over-defensive.

docs-v2/spec.yaml
```diff
@@ -486,6 +486,96 @@ paths:
...
+                        application/json:
+                            schema:
+                                type: object
+                                properties:
+                                    message:
+                                        type: string
+        patch:
+            description: Edit custom metadata for the connection or connections (only overrides specified properties, not the entire metadata). If updating multiple connections pass in an array of connection ids instead of a string.
+            parameters:
+                - name: provider_config_key
```
bastienbeurier: Same here about the query apram

docs-v2/spec.yaml
khaliqgant: This is exactly how it is now FWIW so this isn't a departure. I'd like to think we have enough warnings in the docs and a developer should take note of the HTTP method when performing operations.

bastienbeurier: Ah got it, yes then there is not reason to change!

docs-v2/reference/sdks/node.mdx
```diff
@@ -458,14 +461,26 @@ await nango.setMetadata('<INTEGRATION-ID>', 'CONNECTION-ID', { 'CUSTOM_KEY1': 'C
 
 **Response**
 
-Empty response.
+<Expandable>
+```json
+{
+    "connection_id": "<string | [string]>",
```
bodinsamuel, line 467:
```suggestion
    "connectionId": "<string | [string]>",
```

docs-v2/reference/sdks/node.mdx
```diff
@@ -447,8 +450,8 @@ await nango.setMetadata('<INTEGRATION-ID>', 'CONNECTION-ID', { 'CUSTOM_KEY1': 'C
         The integration ID of the connection.
     </ResponseField>
 
-    <ResponseField name="connectionId" type="string" required>
-        The connection ID.
+    <ResponseField name="connectionId" type="string | string[]" required>
```
bodinsamuel, line 453: I feel like it's easier to have one input/output shape no matter what, but just my opinion
connectionIds in input, and output with one or many connectionId

docs-v2/reference/sdks/node.mdx
```diff
@@ -486,7 +501,16 @@ await nango.updateMetadata('<INTEGRATION-ID>', 'CONNECTION-ID', { 'CUSTOM_KEY1':
 
 **Response**
 
-Empty response.
+<Expandable>
+```json
+{
+    "connection_id": "<string | [string]>",
```
bodinsamuel, line 508:
```suggestion
    "connectionId": "<string | [string]>",
```

packages/server/lib/controllers/connection.controller.ts
```diff
@@ -508,9 +509,99 @@ class ConnectionController {
...
+
+            const { connection_id: connectionIdArg, metadata } = body;
+
+            const connectionIds = Array.isArray(connectionIdArg) ? connectionIdArg : [connectionIdArg];
+
+            const validConnections: Connection[] = [];
+
+            for (const connectionId of connectionIds) {
+                const { success, error: yo, response: connection } = await connectionService.getConnection(connectionId, providerConfigKey, environment.id);
+                console.log(yo);
```
bodinsamuel:
```suggestion

```

bodinsamuel (approved):

docs-v2/reference/sdks/node.mdx
bodinsamuel: (unless the Response field above is wrong)

khaliqgant:
> 👌🏻 It would have been great to use the new pattern of creating routes, just to have some payload validation and test but not blocking

Added with https://github.com/NangoHQ/nango/pull/2145/commits/81943a633f531c69d04faa8272937a7a75ccefab

docs-v2/spec.yaml
khaliqgant: Updated to a POST

docs-v2/spec.yaml
khaliqgant: Updated

docs-v2/reference/sdks/node.mdx
khaliqgant: The response field is for the actual argument into the node sdk client which makes sense to have camel case, so `connection_id` is correct here since that is what the HTTP call returns

docs-v2/reference/sdks/node.mdx
khaliqgant: Point taken, but will leave as is -- discussed this with Bastien.

packages/server/lib/controllers/connection.controller.ts
khaliqgant: Removed, thanks

docs-v2/reference/sdks/node.mdx
khaliqgant: See comment https://github.com/NangoHQ/nango/pull/2145#discussion_r1604431831

packages/shared/lib/services/connection.service.ts
```diff
@@ -483,12 +481,11 @@ class ConnectionService {
...
-    public async updateMetadata(connection: Connection, metadata: Metadata): Promise<Metadata> {
-        const existingMetadata = await this.getMetadata(connection);
-        const newMetadata = { ...existingMetadata, ...metadata };
-        await this.replaceMetadata(connection, newMetadata);
-
-        return newMetadata;
+    public async updateMetadata(connections: Connection[], metadata: Metadata): Promise<void> {
+        for (const connection of connections) {
+            const newMetadata = { ...connection.metadata, ...metadata };
+            await this.replaceMetadata([connection.id as number], newMetadata);
```
TBonnin: would have been nice to have all the queries as part of a transaction. Here we might potentially have some out of sync Metadata if one or more query fails somehow

TBonnin (approved):

packages/shared/lib/services/connection.service.ts
khaliqgant: Updated with https://github.com/NangoHQ/nango/pull/2145/commits/25669dc4c4d26e89079a106842d37dfd8ac514b6

PR #2148, opened 2024-05-13
fix(action): logs ActionError in activities

bodinsamuel:
## Describe your changes

Fixes NAN-926

We were not logging those error, only passed them to the API call which makes the debugging rather painful. Maybe it was intentional not sure.

- Logs ActionError to the activities 
To be cautious I only log the one with `action_script_runtime_error` 

Screenshot 2024-05-13 at 17 26 37

### DIFF
--- a/scripts/gitrelease.mjs
+++ b/scripts/gitrelease.mjs
@@ -26,7 +26,7 @@ echo`Adding file`;
 await $`git add -A package.json package-lock.json packages/**/package.json CHANGELOG.md`;
 
 echo`Creating commit`;
-await $`git -c user.name="Release Bot" -c user.email="contact@nango.dev" commit --allow-empty --author="Release Bot <actions@contact@nango.dev>" -m ${releaseMessage} `;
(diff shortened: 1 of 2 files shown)


scripts/gitrelease.mjs
```diff
@@ -26,7 +26,7 @@ echo`Adding file`;
 await $`git add -A package.json package-lock.json packages/**/package.json CHANGELOG.md`;
 
 echo`Creating commit`;
-await $`git -c user.name="Release Bot" -c user.email="contact@nango.dev" commit --allow-empty --author="Release Bot <actions@contact@nango.dev>" -m ${releaseMessage} `;
+await $`git -c user.name="Release Bot" -c user.email="contact@nango.dev" commit --allow-empty --author="Release Bot <contact@nango.dev>" -m ${releaseMessage} `;
```
bodinsamuel, line 29: unrelated but noticed a bad c/c

TBonnin (approved):

khaliqgant (approved):

PR #2149, opened 2024-05-13
feat(jobs): [nan-919] reconcile temporal schedules

khaliqgant:
## Describe your changes
Add in a dirty reconciliation cron for temporal schedule mismatches. this is temporary until we rip out temporal completely. Also cleans up some logic from the refresh token cron.

NAN-919


packages/jobs/lib/crons/reconcileTemporalSchedules.ts
```diff
@@ -0,0 +1,103 @@
+import * as cron from 'node-cron';
+import { telemetry, LogTypes, LogActionEnum, errorManager, ErrorSourceEnum, SyncClient, db, getRunningSchedules } from '@nangohq/shared';
+import { getLogger, metrics } from '@nangohq/utils';
+import tracer from 'dd-trace';
+
+const logger = getLogger('Jobs.TemporalSchedules');
+
+export function reconcileTemporalSchedules(): void {
+    cron.schedule('*/30 * * * *', async () => {
```
TBonnin: I would remove way more often if possible

packages/jobs/lib/crons/reconcileTemporalSchedules.ts
```diff
@@ -0,0 +1,103 @@
...
+            for (const schedule of runningSchedules) {
+                const { schedule_id, sync_id } = schedule;
+
+                logger.info(`[reconcileTemporalSchedules] reconciling scheduleId: ${schedule_id}, syncId: ${sync_id}`);
+
+                try {
+                    const syncSchedule = await syncClient.describeSchedule(schedule_id);
+
+                    if (syncSchedule && syncSchedule.schedule?.state?.paused) {
+                        const temporalClient = syncClient.getClient();
```
TBonnin, line 64: isn't the sync client already defined above https://github.com/NangoHQ/nango/pull/2149/files#diff-5c90205949c03793a595f68f75775f682f30135635833e27a49f590ba120accbR28

packages/jobs/lib/crons/reconcileTemporalSchedules.ts
```diff
@@ -0,0 +1,103 @@
...
+                        const scheduleHandle = temporalClient?.schedule.getHandle(schedule_id);
+
+                        if (scheduleHandle && !schedule_id.includes('demo')) {
```
TBonnin: what if a customer has `demo` in the name of the sync?

packages/jobs/lib/crons/reconcileTemporalSchedules.ts
```diff
@@ -0,0 +1,103 @@
...
+                    logger.error(`[reconcileTemporalSchedules] failed to reconcile scheduleId: ${schedule_id}, syncId: ${sync_id}`);
+                    metrics.increment(metrics.Types.RENCONCILE_TEMPORAL_SCHEDULES_FAILED);
+                }
+            }
+
+            const lastSchedule = runningSchedules[runningSchedules.length - 1];
+
+            if (lastSchedule && typeof lastSchedule.id === 'number' && lastSchedule.id === lastId) {
+                break;
+            }
```
TBonnin, line 93: not sure I following why this if is necessary

packages/server/lib/refreshTokens.ts
```diff
@@ -31,7 +31,7 @@ export async function exec(): Promise<void> {
     logger.info('[refreshTokens] starting');
 
     const today = new Date();
-    const lockKey = parseInt(`${today.getFullYear()}${today.getMonth() + 1}${today.getDate()}`);
+    const lockKey = parseInt(`1${today.getFullYear()}${today.getMonth() + 1}${today.getDate()}`);
```
TBonnin: why this `1`?

packages/jobs/lib/crons/reconcileTemporalSchedules.ts
khaliqgant: This is the underlying temporal raw client.

packages/jobs/lib/crons/reconcileTemporalSchedules.ts
khaliqgant: Edge case that I think we can live with. There were some instances where interactive demo syncs were getting paused and we actually want those to stay paused.

packages/jobs/lib/crons/reconcileTemporalSchedules.ts
khaliqgant: Would you? I don't see it happening that frequently and IMO 30 minutes is pretty often to pull in 3K schedules

packages/jobs/lib/crons/reconcileTemporalSchedules.ts
khaliqgant: safe guard against infinite loop

packages/server/lib/refreshTokens.ts
khaliqgant: avoid potential collisions with any other lock using the date

packages/jobs/lib/crons/reconcileTemporalSchedules.ts
TBonnin: Ideally I would prefer not to run such cron but if we really need to, I think 30mins is also potentially how much time a sync that is supposed to be running will stay wrongly paused, which is quite a long time. Are you worried about Temporal API not being able to handle 2k `describeSchedule`? Can we use `listSchedules()` instead?

packages/jobs/lib/crons/reconcileTemporalSchedules.ts
TBonnin: can we be more specific? Aren't the interactive demo schedule id starting with something like `nango-syncs.github-issues-demo...`

packages/jobs/lib/crons/reconcileTemporalSchedules.ts
```diff
@@ -0,0 +1,103 @@
...
+        }
+        metrics.duration(metrics.Types.RENCONCILE_TEMPORAL_SCHEDULES, Date.now() - start);
+    });
+}
+
+export async function exec(): Promise<void> {
+    logger.info('[reconcileTemporalSchedules] starting');
+
+    const today = new Date();
+    const lockKey = parseInt(`2${today.getFullYear()}${today.getMonth() + 1}${today.getDate()}`);
```
TBonnin: is the lock to prevent multiple instance of the cron to execute at the same time? What if the last cron of a day takes a long time or never finishes?

packages/jobs/lib/app.ts
```diff
@@ -25,6 +26,7 @@ try {
     cronAutoIdleDemo();
     deleteOldActivityLogs();
     deleteSyncsData();
+    reconcileTemporalSchedules();
```
bodinsamuel, line 29: Can we make this optional because temporal does not exist in oauth install

packages/jobs/lib/crons/reconcileTemporalSchedules.ts
bodinsamuel: I added this on my cron to avoid multiple instances conflict yes. Not sure I get the last part of the message

packages/server/lib/refreshTokens.ts
bodinsamuel: pg_try_advisory_xact_lock key should be an int so it's highly restrictive, but we could have used redis lock

bodinsamuel (approved):

packages/jobs/lib/crons/reconcileTemporalSchedules.ts
TBonnin: the key is the date right? so 2 cron instances started on different day could run at the same time, no?

bodinsamuel: good point indeed, on my side I used a constant number. Ideally here we should crc32 a constant string

packages/jobs/lib/app.ts
khaliqgant: This is running on `jobs` which assumes that temporal is running.

packages/jobs/lib/crons/reconcileTemporalSchedules.ts
khaliqgant: Updated with https://github.com/NangoHQ/nango/pull/2149/commits/675bc605697e650fa71a99b63b439051408a2171

packages/jobs/lib/crons/reconcileTemporalSchedules.ts
khaliqgant: Updated to `nango-syncs.issues-demo`

packages/utils/lib/string.ts
```diff
@@ -0,0 +1,9 @@
+export function stringToHash(str: string): number {
+    let hash = 0;
+    for (let i = 0; i < str.length; i++) {
+        const char = str.charCodeAt(i);
+        hash = (hash << 5) - hash + char;
+        hash |= 0;
+    }
+    return hash;
```
TBonnin, line 8: curious: what is this hashing algo called?

TBonnin (approved):

PR #2151, opened 2024-05-14
fix(temporal): should not be loaded for oauth install

bodinsamuel:
## Describe your changes

Fixes NAN-931

- Fix env detection for `isHosted` 

- Breaking change: Throw when we try to instantiate temporal without an env var
If we default the URL we have no way to know if it was a mistake or not. And I believe checking `isHosted` is not an appropriate check, we don't want to prevent people self-hosting a fully fledge infra


packages/jobs/lib/temporal.ts
```diff
@@ -34,9 +34,14 @@ export class Temporal {
...
 
+        const temporalAddress = process.env['TEMPORAL_ADDRESS'];
+        if (!temporalAddress) {
+            throw new Error('TEMPORAL_ADDRESS missing from env var');
+        }
+
         try {
             const connection = await NativeConnection.connect({
-                address: process.env['TEMPORAL_ADDRESS'] || 'localhost:7233',
+                address: temporalAddress,
```
TBonnin, line 44: does that mean I now need the env var for my local setup? :)

packages/utils/lib/environment/detection.ts
```diff
@@ -3,12 +3,13 @@ import { NodeEnv, localhostUrl } from './constants.js';
...
 
+export const isDocker = process.env['SERVER_RUN_MODE'] === 'DOCKERIZED';
 export const isStaging = process.env['NODE_ENV'] === NodeEnv.Staging;
 export const isProd = process.env['NODE_ENV'] === NodeEnv.Prod;
 export const isCloud = process.env['NANGO_CLOUD']?.toLowerCase() === 'true';
-export const isLocal = baseUrl === localhostUrl;
 export const isEnterprise = process.env['NANGO_ENTERPRISE']?.toLowerCase() === 'true';
-export const isDev = process.env['NODE_ENV'] === NodeEnv.Dev;
+export const isDev = !isDocker && process.env['NODE_ENV'] === NodeEnv.Dev;
+export const isLocal = !isCloud && !isEnterprise && !isDocker && isDev;
```
TBonnin: hard to see if those changes have any side effect. Those conditions are starting to be quite complicated

TBonnin (approved):

packages/utils/lib/environment/detection.ts
bodinsamuel: yep, that's why I would love for us to use feature flags instead

packages/jobs/lib/temporal.ts
bodinsamuel: yes, that's the main proposal :D 
happy to discuss if it doesn't make sense

khaliqgant:
Which line/file was the actual fix for the NAN-931 issue?

bodinsamuel:
> Which line/file was the actual fix for the NAN-931 issue?

The fix in hooks.ts

packages/jobs/lib/temporal.ts
TBonnin: I am ok with that. We have some defaults that are aimed at making our local dev life easier but they make detecting bad config in prod more silent.

khaliqgant: Yeah, not a problem on my end either

packages/utils/lib/environment/detection.ts
khaliqgant: I don't understand the conditions fully tbh. Unclear about `isLocal`, `isDev` -- isDev contains isDocker and isLocal contains isDocker and isDev 🤯 . I think it might be worth going to feature flags now even instead of this convoluted logic

bodinsamuel: I copied isDocker to isDev just in case, but actually I'll remove this constant since it's not used in the code.
There is not an easy way to be sure we are in a local install so it's just "not everything" :D

khaliqgant (approved):
The .env.example should be updated to have the `TEMPORAL_ADDRESS`

PR #2152, opened 2024-05-14
fix(ui): env switching issue

bodinsamuel:
## Describe your changes

Fixes NAN-933

- Use navigate instead of pushHistory that was not playing well with react state
- Properly sync and reset query params with react-router-dom util
It's not perfect yet, when dealing with query params in react they should be considered the source of truth and flow top-down instead of setting state then query params otherwise it can lead to desync during re-render

### DIFF
--- a/packages/webapp/src/pages/Logs/Show.tsx
+++ b/packages/webapp/src/pages/Logs/Show.tsx
@@ -1,6 +1,4 @@
 export const Show: React.FC = () => {
-    console.log('coucou');
-
     return (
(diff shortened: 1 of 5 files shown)


packages/webapp/src/pages/Logs/Show.tsx
```diff
@@ -1,6 +1,4 @@
 export const Show: React.FC = () => {
-    console.log('coucou');
```
TBonnin, line 2: 👋 🙈

TBonnin (approved):

PR #2156, opened 2024-05-14
feat(logs): GET /logs/operations/:operationId

bodinsamuel:
## Describe your changes

Fixes NAN-927

- **Add route `GET /logs/operations/:operationId`** 
This will allow us to query a single operation and display it in standalone in the UI


packages/logs/lib/client.ts
```diff
@@ -62,7 +62,7 @@ export class LogContext {
         await this.log({ type: 'log', level: 'warn', message, meta, source: 'internal' });
     }
 
-    async error(message: string, meta: (MessageMeta & { error?: unknown }) | null = null): Promise<void> {
+    async error(message: string, meta: (MessageMeta & { error?: unknown; err?: never; e?: never }) | null = null): Promise<void> {
```
TBonnin, line 65: what's `e`?

packages/server/lib/utils/tests.ts
```diff
@@ -7,20 +7,29 @@ import { getServerPort } from '@nangohq/shared';
 
 import { app } from '../routes.js';
 
+function uriParamsReplacer(tpl: string, data: Record<string, any>) {
+    return tpl.replace(/:([a-zA-Z]+)/g, function (_, name) {
+        return data[name];
+    });
```
TBonnin: if we want to avoid regexes 
```
function uriParamsReplacer(tpl: string, data: Record<string, any>) {
            let res = tpl
            for (const [key, value] of Object.entries(data)) {
                res = res.replace(`:${key}`, value);
            }
            return res;
});
```

TBonnin (approved):
It is funny. For the orchestration project, I am making heavy use of the `Endpoint` interface you introduced recently to define a type safe HTTP API and I came up with sightly different but similar solutions. I would be able to share soon, we can then decide what to keep and what to throw away. Looking forward to it

packages/logs/lib/client.ts
bodinsamuel: I just want to prevent us to pass `err` or `e` instead of `error`

packages/server/lib/utils/tests.ts
bodinsamuel: true, it's more readable

bodinsamuel:
Note: changed the path to /operations/:operationId I realized it could conflict with future routes

> For the orchestration project,

you are not using trpc this time?

PR #2157, opened 2024-05-15
fix(oauth): [nan-934] fix public key lookup

khaliqgant:
## Describe your changes
When creating a slack connection the hmac key needs to match the key of the Nango admin account not the current user's hmac account. 

This bug was introduced with https://github.com/NangoHQ/nango/pull/2097

NAN-934

### DIFF
--- a/packages/server/lib/controllers/environment.controller.ts
+++ b/packages/server/lib/controllers/environment.controller.ts
@@ -126,9 +126,22 @@ class EnvironmentController {
             const nangoAdminUUID = NANGO_ADMIN_UUID;
             const env = 'prod';
             const info = await accountService.getAccountAndEnvironmentIdByUUID(nangoAdminUUID as string, env);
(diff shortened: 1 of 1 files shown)


packages/server/lib/controllers/environment.controller.ts
```diff
@@ -126,9 +126,22 @@ class EnvironmentController {
             const nangoAdminUUID = NANGO_ADMIN_UUID;
             const env = 'prod';
             const info = await accountService.getAccountAndEnvironmentIdByUUID(nangoAdminUUID as string, env);
-            const digest = await hmacService.digest(info?.environmentId as number, integration_key, connectionId as string);
 
-            res.status(200).send({ hmac_digest: digest, public_key: res.locals['environment'].public_key, integration_key });
```
khaliqgant, line 131: The public_key here was wrong and was what caused issues

bodinsamuel (approved):

PR #2158, opened 2024-05-15
feat(cli): [nan-918] exit deploy if not everything compiled successfully

khaliqgant:
## Describe your changes
On compilation error stop the deploy process. This is to prevent data loss if there is an issue with any script as the current flow would delete any scripts that had a compilation error.

NAN-918

### DIFF
--- a/packages/cli/lib/services/deploy.service.ts
+++ b/packages/cli/lib/services/deploy.service.ts
@@ -35,7 +35,12 @@ class DeployService {
             printDebug(`Environment is set to ${environmentName}`);
         }
 
-        await compileAllFiles({ debug });
+        const successfulCompile = await compileAllFiles({ debug });
+
+        if (!successfulCompile) {
+            console.log(chalk.red('Compilation was not fully successful. Please make sure all files compile before deploying'));
+            return;
+        }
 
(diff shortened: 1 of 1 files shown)


packages/cli/lib/services/deploy.service.ts
```diff
@@ -35,7 +35,12 @@ class DeployService {
             printDebug(`Environment is set to ${environmentName}`);
         }
 
-        await compileAllFiles({ debug });
+        const successfulCompile = await compileAllFiles({ debug });
+
+        if (!successfulCompile) {
+            console.log(chalk.red('Compilation was not fully successful. Please make sure all files compile before deploying'));
+            return;
```
bodinsamuel: this kind of errors should probably process.exit(1) otherwise it won't break the CI

bodinsamuel (approved):

packages/cli/lib/services/deploy.service.ts
khaliqgant: Agreed, updated with https://github.com/NangoHQ/nango/pull/2158/commits/8a56cd08968b95164e945c21d007b51d097332de

PR #2159, opened 2024-05-15
fix(webapp): [nan-943] fix API reference for multi models

khaliqgant:
## Describe your changes
Multi model sync API reference assumed the first index for the endpoint and model documentation. This makes it dynamic based on the endpoint clicked on.

NAN-943


bodinsamuel (approved):
Could have been a Route instead imo

PR #2160, opened 2024-05-15
fix(action): debug empty response

bodinsamuel:
## Describe your changes

There is an ongoing issue where a customer receive a 500 when running an action but without any log showing

- Log more stuff (to revert)
- Fix `res.send` was used to output the body but could fail if the body was a number, express would interpret that as a wrong status code


packages/server/lib/controllers/sync.controller.ts
```diff
@@ -430,6 +430,7 @@ class SyncController {
                 await logCtx.failed();
                 span.finish();
 
+                next(actionResponse.error);
```
bodinsamuel, line 433: maybe I'm crazy but this if never worked?

TBonnin: that would be a good job for a compiler/linter to tell you an handler doesn't call next or send ;)

TBonnin (approved):

PR #2161, opened 2024-05-15
fix(api): handle invalid json payload

bodinsamuel:
## Describe your changes

Fixes NAN-945

- **Handle invalid json** payload with a 400 instead of a cryptic 500

### DIFF
--- a/packages/server/lib/routes.ts
+++ b/packages/server/lib/routes.ts
@@ -33,6 +33,7 @@ import { errorManager } from '@nangohq/shared';
 import tracer from 'dd-trace';
 import { searchLogs } from './controllers/v1/logs/searchLogs.js';
 import { getOperation } from './controllers/v1/logs/getOperation.js';
+import type { ApiError } from '@nangohq/types';
 
 export const app = express();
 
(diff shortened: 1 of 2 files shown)


packages/server/lib/routes.ts
```diff
@@ -233,6 +234,11 @@ app.use(staticSite);
 
 // -------
 // Error handling.
-app.use((e: any, req: Request, res: Response, _: any) => {
-    errorManager.handleGenericError(e, req, res, tracer);
+app.use((err: any, req: Request, res: Response<ApiError<'invalid_json'>>, _: any) => {
+    if (err instanceof SyntaxError && 'body' in err && 'type' in err && err.type === 'entity.parse.failed') {
```
bodinsamuel, line 238: really brittle https://github.com/expressjs/express/issues/4065

TBonnin: 😢 
why did you added the `&& 'type' in err && err.type === 'entity.parse.failed'` to the condition compared to the solution in the express repo issue?

TBonnin (approved):

packages/server/lib/routes.ts
bodinsamuel: SyntaxError is a generic error so anything can throw that (it's really a bad package), at least it can't be confused with anything else like this
https://stackoverflow.com/a/58136032

khaliqgant (approved):

PR #2163, opened 2024-05-15
fix(temporal): handle error outside the scripts

bodinsamuel:
## Context

Fixes NAN-929

There are multiple combined errors that lead to a painful investigation:
- Temporal is wrapping all function in `activities.ts` and try/catch them to report failure but because we added an other layer of try/catch, all tasks are marked as successful. This prevent retry (which is a good thing or not I don't even know anymore) but most importantly it broke the expected return in sync.client.ts
- Sync.client.ts expected `actionHandler` to always have a shape of `{success, error, response}` which was not the case as mentioned. Natively Temporal is re-throwing at the place you wait for a workflow but because of our try/catch it was just returning false (**note**: all our workflows are broken this way but only action and webhooks are awaited like this, I only fixed action because we are replacing everything soon)
- The error should have at least been sent to Datadog, but because of the foreign Temporal object it wasn't stringified properly
- The error that started this investigation `model_schema` in db being broken somehow for this customer (it was missing `field`), I could not reproduce, and a redeploy fixed it.

### Addition Notes
- All our workflows are somewhat broken because we have only partial try/catch (@TBonnin you probably need to think about that with your work)
- In a sync, if any errors happen before the activity log is created (`activities.ts:241`) we have absolutely no trace of a sync execution except in datadog, it should probably be the first step in any task execution. And there are a lot of stuff that can fail.
- Error reporting is unmanageable, we have multiple telemetry.log, errorManager, activities reporting different stuff at different place, in `jobs` and in `server`. We should have --ideally-- one single report in a global try/catch.
- Because we don't have a proper sentry setup, and our Datadog setup only reports Error in span, it's simply impossible to be warned automatically of those errors. 

## Changes

- Correctly stringify the error that is sent to Datadog
- Correctly output foreign error in action 
- Fix double headers sent sync.controller
- Fix errResFromNangoErr not handling error being null
- Remove temp logging

## Test

- Throw an error in `activities.ts:runAction`
- Trigger an action

Screenshot 2024-05-15 at 18 55 07

Screenshot 2024-05-15 at 18 56 09

### DIFF
--- a/packages/server/lib/controllers/sync.controller.ts
+++ b/packages/server/lib/controllers/sync.controller.ts
@@ -425,12 +425,11 @@ class SyncController {
                 return;
             } else {
                 span.setTag('nango.error', actionResponse.error);
-                errorManager.errResFromNangoErr(res, actionResponse.error);
                 await logCtx.error('Failed to trigger action', { error: actionResponse.error });
                 await logCtx.failed();
-                span.finish();
 
-                next(actionResponse.error);
+                errorManager.errResFromNangoErr(res, actionResponse.error);
+                span.finish();
                 return;
             }
         } catch (err) {
--- a/packages/shared/lib/utils/error.manager.ts
+++ b/packages/shared/lib/utils/error.manager.ts
@@ -127,10 +127,12 @@ class ErrorManager {
         if (err) {
             logger.error(`Response error: ${JSON.stringify({ statusCode: err.status, type: err.type, payload: err.payload, message: err.message })}`);
             if (!err.message) {
-                res.status(err.status || 500).send({ type: err.type, payload: err.payload });
+                res.status(err.status || 500).send({ type: err.type || 'unknown_error', payload: err.payload });
             } else {
                 res.status(err.status || 500).send({ error: err.message, type: err.type, payload: err.payload });
             }
+        } else {
+            res.status(500).json({ error: { code: 'unknown_empty_error' } });
         }
     }
 
--- a/packages/jobs/lib/activities.ts
+++ b/packages/jobs/lib/activities.ts
@@ -23,7 +23,7 @@ import {
     getLastSyncDate
 } from '@nangohq/shared';
 import { records as recordsService } from '@nangohq/records';
-import { getLogger, env, stringifyError } from '@nangohq/utils';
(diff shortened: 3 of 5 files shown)


packages/server/lib/controllers/sync.controller.ts
```diff
@@ -425,12 +425,11 @@ class SyncController {
                 return;
             } else {
                 span.setTag('nango.error', actionResponse.error);
-                errorManager.errResFromNangoErr(res, actionResponse.error);
                 await logCtx.error('Failed to trigger action', { error: actionResponse.error });
                 await logCtx.failed();
-                span.finish();
 
-                next(actionResponse.error);
```
bodinsamuel, line 433: completely missed that errorManager.errResFromNangoErr was already sending the json, honestly confusing, hopefully with the refactor of the API it's going to be smoother

packages/shared/lib/utils/error.manager.ts
```diff
@@ -127,10 +127,12 @@ class ErrorManager {
...
             logger.error(`Response error: ${JSON.stringify({ statusCode: err.status, type: err.type, payload: err.payload, message: err.message })}`);
             if (!err.message) {
-                res.status(err.status || 500).send({ type: err.type, payload: err.payload });
+                res.status(err.status || 500).send({ type: err.type || 'unknown_error', payload: err.payload });
             } else {
                 res.status(err.status || 500).send({ error: err.message, type: err.type, payload: err.payload });
             }
+        } else {
+            res.status(500).json({ error: { code: 'unknown_empty_error' } });
         }
```
bodinsamuel, line 136: err being null was never handled hanging forever

khaliqgant:
> All our workflows are somewhat broken because we have only partial try/catch (@TBonnin you probably need to think about that with your work)

Can you show an example of this / explain further?

khaliqgant (approved):

TBonnin (approved):

bodinsamuel:
> Can you show an example of this / explain further?

I'm referring to the fact that not everything that throws is try/catch so not all paths are returning the same output and not all paths have the correct error handling. Broken is strong word, but inconsistent at the very least

PR #2167, opened 2024-05-16
fix(getSyncs): should not join on action

bodinsamuel:
## Describe your changes

Fixes NAN-959

When you have an a sync and an action with the same name the join is invalid. Not sure why we don't have a column `sync_config_id` in `_nango_syncs`


khaliqgant (approved):
> Not sure why we don't have a column sync_config_id in _nango_syncs

We should. I can't see any reason not to. Unsure how difficult it would be to backfill this value.

PR #2168, opened 2024-05-16
feat(scripts): [nan-973] slack integration for each environment

khaliqgant:
## Describe your changes
Allow a slack integration to be created at each env level. CLI fixes

NAN-973

### DIFF
--- a/packages/cli/lib/utils.ts
+++ b/packages/cli/lib/utils.ts
@@ -456,20 +456,6 @@ function getPackagePath(debug = false) {
 }
 
 export async function parseSecretKey(environment: string, debug = false): Promise<void> {
-    if (process.env['NANGO_SECRET_KEY_PROD'] && environment === 'prod') {
-        if (debug) {
-            printDebug(`Environment is set to prod, setting NANGO_SECRET_KEY to NANGO_SECRET_KEY_PROD.`);
-        }
-        process.env['NANGO_SECRET_KEY'] = process.env['NANGO_SECRET_KEY_PROD'];
-    }
-
-    if (process.env['NANGO_SECRET_KEY_DEV'] && environment === 'dev') {
-        if (debug) {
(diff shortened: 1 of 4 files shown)


packages/cli/lib/utils.ts
```diff
@@ -114,7 +114,7 @@ export function checkEnvVars(optionalHostport?: string) {
 
 let pkgVersion: string | undefined = undefined;
 export function getPkgVersion(debug = false) {
-    if (pkgVersion) {
+    if (pkgVersion && !debug) {
```
khaliqgant, line 117: Something that is very important to my workflow is to know where the CLI is running from. Because the pkgVersion is fetched in the `getUserAgent()` method on startup and cached the path wasn't being outputted. So with this the cache is skipped so the full debug information can be outputted if `debug` is set.

packages/webapp/src/pages/EnvironmentSettings.tsx
```diff
@@ -372,7 +372,7 @@ export const EnvironmentSettings: React.FC = () => {
     };
 
     const connectSlack = async () => {
-        const connectionId = `account-${accountUUID}`;
```
bodinsamuel, line 375: is this retroactive somehow?

khaliqgant: No. Going to manually update the prod connections. Only 20 or so

bodinsamuel: maybe self hosted / enterprise are using this feature? if so could be done with a transaction

khaliqgant: Enterprise definitely isn’t using the feature and since this is only for syncs / actions can almost guarantee that self hosted isn’t either

bodinsamuel (approved):

TBonnin (approved):

PR #2171, opened 2024-05-17
chore: making the scheduler db not rely on env vars

TBonnin:
## Describe your changes

configuring the scheduler, which is a independant library, with env vars make it harder to understand where things are being configured and how. We ran into issues with the records package regarding default env vars value and behavior if/when not configured. With this commit the db url and schema are passed into the scheduler package and we don't have a global db object within the package. The `Tasks` function are now all taking `db` as their first parameter, this has the nice side effect to make it possible to run them in a transaction if needed.


packages/scheduler/lib/monitor.ts
```diff
@@ -1,11 +1,17 @@
...
 
 const logger = getLogger('Scheduler.monitor');
 
 if (!isMainThread && parentPort) {
-    new MonitorChild(parentPort);
+    const { url, schema } = workerData;
+    if (!url || !schema) {
+        throw new Error('Missing required database url and schema for monitor worker');
+    }
+    const dbClient = new DatabaseClient({ url, schema, poolMax: 10 });
```
TBonnin, line 13: the worker thread doesn't share anything with the main thread and only serializable object can be messaged between them so we are creating another db client for the worker thread. Right now the worker is only expiring tasks but when we add support for recurring tasks the worker will be responsible for all state transitions not trigger by an event

khaliqgant (approved):

bodinsamuel (approved):

PR #2172, opened 2024-05-17
feat(scheduler): all terminated tasks can have an output

TBonnin:
## Describe your changes

I started with the idea to add an `output` for failed task (the error that caused the failure) but I realized all terminated tasks can have an output. Ex: reason for cancellation, etc...
This commit ensures it is required to set an output when terminating a non successful task

### DIFF
--- a/packages/orchestrator/lib/client.integration.test.ts
+++ b/packages/orchestrator/lib/client.integration.test.ts
@@ -33,94 +33,102 @@ describe('OrchestratorClient', async () => {
         await dbClient.clearDatabase();
     });
 
-    it('should schedule immediate task', async () => {
-        const groupKey = 'groupA';
-        const res = (
-            await client.schedule({
-                name: 'Task',
-                groupKey: groupKey,
-                retry: { count: 3, max: 5 },
-                timeoutSettingsInSecs: {
-                    createdToStarted: 10,
-                    startedToCompleted: 10,
-                    heartbeat: 10
-                },
-                args: {
-                    name: 'Action',
-                    connection: {
-                        id: 1234,
-                        provider_config_key: 'P',
-                        environment_id: 5678
+    describe('schedule', () => {
+        it('should schedule immediate task', async () => {
+            const groupKey = 'groupA';
+            const res = (
+                await client.schedule({
+                    name: 'Task',
+                    groupKey: groupKey,
+                    retry: { count: 3, max: 5 },
+                    timeoutSettingsInSecs: {
+                        createdToStarted: 10,
+                        startedToCompleted: 10,
+                        heartbeat: 10
                     },
-                    activityLogId: 9876,
-                    input: { foo: 'bar' }
-                }
-            })
-        ).unwrap();
-        const tasks = (await scheduler.list({ groupKey })).unwrap();
-        expect(tasks.length).toBe(1);
-        expect(tasks[0]?.id).toBe(res.taskId);
-        expect(tasks[0]?.state).toBe('CREATED');
+                    args: {
+                        name: 'Action',
+                        connection: {
+                            id: 1234,
+                            provider_config_key: 'P',
+                            environment_id: 5678
+                        },
+                        activityLogId: 9876,
+                        input: { foo: 'bar' }
+                    }
+                })
+            ).unwrap();
+            const tasks = (await scheduler.list({ groupKey })).unwrap();
+            expect(tasks.length).toBe(1);
+            expect(tasks[0]?.id).toBe(res.taskId);
+            expect(tasks[0]?.state).toBe('CREATED');
+        });
     });
 
-    it('should execute', async () => {
-        const groupKey = 'groupB';
-        const output = { count: 9 };
+    describe('execute', () => {
+        it('should be successful when task succeed', async () => {
+            const groupKey = 'groupB';
+            const output = { count: 9 };
 
-        const processor = new MockProcessor({
-            groupKey,
-            process: async (task) => {
-                await scheduler.succeed({ taskId: task.id, output });
-            }
-        });
-        try {
-            const res = await client.execute({
-                name: 'Task',
-                groupKey: groupKey,
-                args: {
-                    name: 'Action',
-                    connection: {
-                        id: 1234,
-                        provider_config_key: 'P',
-                        environment_id: 5678
-                    },
-                    activityLogId: 9876,
-                    input: { foo: 'bar' }
+            const processor = new MockProcessor({
+                groupKey,
+                process: async (task) => {
+                    await scheduler.succeed({ taskId: task.id, output });
                 }
             });
-            expect(res.unwrap()).toEqual(output);
-        } finally {
-            processor.stop();
-        }
-    });
-    it('should return an error if execute fails', async () => {
-        const groupKey = 'groupC';
-
-        const processor = new MockProcessor({
-            groupKey,
-            process: async (task) => {
-                await scheduler.fail({ taskId: task.id });
+            try {
+                const res = await client.execute({
+                    name: 'Task',
+                    groupKey: groupKey,
+                    args: {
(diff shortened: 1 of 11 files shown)


packages/orchestrator/lib/client.integration.test.ts
```diff
@@ -31,94 +31,102 @@ describe('OrchestratorClient', async () => {
...
+                            provider_config_key: 'P',
+                            environment_id: 5678
+                        },
+                        activityLogId: 9876,
+                        input: { foo: 'bar' }
+                    }
+                });
+                expect(res.isOk()).toBe(false);
+                if (res.isErr()) {
+                    expect(res.error.payload).toBe(res.error.payload);
```
TBonnin, line 126: this `if` is actually the only addition in the file. The rest is just grouping tests

packages/orchestrator/lib/client.integration.test.ts
```diff
@@ -31,94 +31,102 @@ describe('OrchestratorClient', async () => {
...
-                    startedToCompleted: 10,
-                    heartbeat: 10
-                },
-                args: {
-                    name: 'Action',
-                    connection: {
-                        id: 1234,
-                        provider_config_key: 'P',
-                        environment_id: 5678
+    describe('schedule', () => {
```
TBonnin, line 36: You can scroll to the end of this file where the only changes is. The rest is just grouping tests

packages/scheduler/lib/models/tasks.integration.test.ts
```diff
@@ -52,26 +52,23 @@ describe('Task', () => {
         const updated = (await tasks.heartbeat(t.id)).unwrap();
         expect(updated.lastHeartbeatAt.getTime()).toBeGreaterThan(t.lastHeartbeatAt.getTime());
     });
-    it('should fail to transition to SUCCEEDED without an output', async () => {
-        const t = await startTask();
-        const updated = await tasks.transitionState({ taskId: t.id, newState: 'SUCCEEDED' });
-        expect(updated.isErr()).toBe(true);
-    });
```
TBonnin, line 61: no need anymore. It is now enforced at compile time.

packages/scheduler/lib/scheduler.ts
```diff
@@ -163,12 +163,13 @@ export class Scheduler {
...
      * @param taskId - Task ID
+     * @param error - Json object representing the error
      * @returns Task
      * @example
-     * const failed = await scheduler.fail({ taskId: '00000000-0000-0000-0000-000000000000' });
+     * const failed = await scheduler.fail({ taskId: '00000000-0000-0000-0000-000000000000', error: {message: 'error'});
      */
-    public async fail({ taskId }: { taskId: string }): Promise<Result<Task>> {
-        const failed = await tasks.transitionState({ taskId, newState: 'FAILED' });
+    public async fail({ taskId, error }: { taskId: string; error: JsonValue }): Promise<Result<Task>> {
```
TBonnin, line 174: `error: JsonValue`: error format is not enforced by the scheduler. The business logic (aka server and the processor) can define their own format. It is less prescriptive but allow for more flexibility.

packages/scheduler/lib/scheduler.ts
```diff
@@ -203,11 +204,11 @@ export class Scheduler {
      * @example
      * const cancelled = await scheduler.cancel({ taskId: '00000000-0000-0000-0000-000000000000' });
      */
-    public async cancel(cancelBy: { taskId: string } | { scheduleId: string }): Promise<Result<Task>> {
+    public async cancel(cancelBy: { taskId: string; reason: string } | { scheduleId: string; reason: string }): Promise<Result<Task>> {
```
TBonnin, line 210: we could also make the second param a `JsonValue` and allow any kind of object. Let me know what you think.

packages/orchestrator/lib/routes/v1/task/taskId/output.ts
```diff
@@ -33,13 +33,10 @@ const getHandler = (scheduler: Scheduler) => {
...
                 return res.status(204).send(); // No content yet
             case 'SUCCEEDED':
-                return res.status(200).json({ output: task.value.output });
             case 'FAILED':
-                return res.status(404).json({ error: { code: 'task_failed', message: `failed` } });
             case 'EXPIRED':
-                return res.status(404).json({ error: { code: 'task_expired', message: `expired` } });
             case 'CANCELLED':
-                return res.status(404).json({ error: { code: 'task_cancelled', message: `cancelled` } });
+                return res.status(200).json({ state: task.value.state, output: task.value.output });
```
khaliqgant, line 39: `FAILED` and `EXPIRED` respond with a 200?

packages/scheduler/lib/models/tasks.ts
```diff
@@ -226,7 +235,15 @@ export async function expiresIfTimeout(): Promise<Result<Task[]>> {
             .update({
                 state: 'EXPIRED',
                 last_state_transition_at: new Date(),
-                terminated: true
+                terminated: true,
+                output: db.raw(`
+                    CASE
```
khaliqgant, line 241: logic seems a bit verbose to be in a raw query. Unsure if there is a better way -- my only concern is maintenance

packages/scheduler/lib/scheduler.ts
khaliqgant, line 174: Not directly related to this PR but I feel if you destructured `task` it would be more readable in the fail function

```
    public async fail({ taskId, error }: { taskId: string; error: JsonValue }): Promise<Result<Task>> {
        const failed = await tasks.transitionState({ taskId, newState: 'FAILED', output: error });
        if (failed.isOk()) {
            const task = failed.value;
            const { state, retryMax, retryCount, name, payload, groupKey, createdToStartedTimeoutSecs, startedToCompletedTimeoutSecs, heartbeatTimeoutSecs } =
                task;
            this.onCallbacks[state](task);
            // Create a new task if the task is retryable
            if (retryMax > retryCount) {
                const schedulingProps: SchedulingProps = {
                    scheduling: 'immediate',
                    taskProps: {
                        name,
                        payload,
                        groupKey,
                        retryMax,
                        retryCount: retryCount + 1,
                        createdToStartedTimeoutSecs,
                        startedToCompletedTimeoutSecs,
                        heartbeatTimeoutSecs
                    }
                };
                const res = await this.schedule(schedulingProps);
                if (res.isErr()) {
                    logger.error(`Error retrying task '${taskId}': ${stringifyError(res.error)}`);
                }
            }
        }
        return failed;
    }
```

khaliqgant (approved):
Non blocking feedback and q's

packages/scheduler/lib/models/tasks.integration.test.ts
```diff
@@ -113,20 +110,23 @@ describe('Task', () => {
         await new Promise((resolve) => void setTimeout(resolve, timeout * 1100));
         const expired = (await tasks.expiresIfTimeout()).unwrap();
         expect(expired).toHaveLength(1);
+        expect(expired[0]?.output).toMatchObject({ reason: `Timeout 'createdToStartedTimeoutSecs' exceeded` });
```
bodinsamuel: just a comment not a suggestion, usually I prefer storing error code (along with  details if necessary) instead of full sentence it's more portable and easy to filter

bodinsamuel (approved):

packages/orchestrator/lib/routes/v1/task/taskId/output.ts
TBonnin: yes, since this is the `/output` endpoint and there is potentially an output it is now a 200

packages/scheduler/lib/models/tasks.ts
TBonnin: I agree that it is verbose. Do you see a better way to achieve the same thing?

PR #2173, opened 2024-05-20
feat(server): [nan-869] email verification on signup

khaliqgant:
## Describe your changes
Verify a user's email on signup and route them through an email verification flow. 

NAN-869


packages/server/lib/clients/workos.client.ts
```diff
@@ -0,0 +1,19 @@
...
+import { NangoError } from '@nangohq/shared';
+
+let workOs: WorkOS | null = null;
+
+export function getWorkOSClient(): WorkOS {
+    if (!workOs) {
+        const apiKey = process.env['WORKOS_API_KEY'];
+        const clientId = process.env['WORKOS_CLIENT_ID'];
+
+        if (apiKey && clientId) {
```
TBonnin, line 11: `clientId` is not being used. Is it on purpose?

packages/server/lib/controllers/v1/account/resendVerificationEmail.ts
```diff
@@ -0,0 +1,47 @@
...
+        });
+        return;
+    }
+
+    const { uuid } = val.data;
+
+    const user = await userService.getUserByUuid(uuid);
+
+    if (!user) {
+        res.status(404).send({ error: { code: 'user_not_found', message: 'User was found in our system.' } });
```
TBonnin:
```suggestion
        res.status(404).send({ error: { code: 'user_not_found', message: 'User was not found in our system.' } });
```

packages/server/lib/helpers/email.ts
```diff
@@ -0,0 +1,22 @@
...
+    const emailClient = EmailClient.getInstance();
+    emailClient.send(
+        email,
+        `Verify your email address`,
+        `
+<p>Hi ${name},</p>
+
+<p>We need to verify your email address ${email}</p>
+
+<p>Please verify this account by clicking <a href="${basePublicUrl}/signup/verification/${token}">here</a></p>
```
TBonnin: nitpick: I would show the full url instead of `here` just to make it clearer it is going to our domain and it is not some sort of phishing attempt

packages/server/lib/routes.ts
```diff
@@ -128,12 +129,16 @@ setupAuth(web);
...
+    web.route('/api/v1/account/signup').post(rateLimiterMiddleware, signup);
+    web.route('/api/v1/account/signup/token').post(rateLimiterMiddleware, signupWithToken);
+    web.route('/api/v1/account/signup/invite').get(rateLimiterMiddleware, authController.invitation.bind(authController));
+    web.route('/api/v1/account/logout').post(rateLimiterMiddleware, authController.logout.bind(authController));
+    web.route('/api/v1/account/signin').post(rateLimiterMiddleware, passport.authenticate('local'), signin);
     web.route('/api/v1/forgot-password').put(rateLimiterMiddleware, authController.forgotPassword.bind(authController));
     web.route('/api/v1/reset-password').put(rateLimiterMiddleware, authController.resetPassword.bind(authController));
+    web.route('/api/v1/account/resend-verification-email').post(rateLimiterMiddleware, resendVerificationEmail);
+    web.route('/api/v1/account/email/:uuid').get(rateLimiterMiddleware, getEmailByUuid);
+    web.route('/api/v1/account/verify/code').post(rateLimiterMiddleware, validateEmailAndLogin);
```
TBonnin, line 153: I would add those endpoints to https://github.com/NangoHQ/nango/blob/960c0e8eb7b6e6926dccf56054e9cc4fae6220f3/packages/server/lib/middleware/ratelimit.middleware.ts#L68 to limit potential abuse

packages/shared/lib/services/user.service.ts
```diff
@@ -18,6 +18,30 @@ class UserService {
...
     }
 
+    async getUserByUuid(uuid: string): Promise<User | null> {
+        const result = await db.knex.select('*').from<User>(`_nango_users`).where({ uuid });
+
+        if (result == null || result.length == 0 || result[0] == null) {
+            return null;
+        }
+
+        return result[0];
```
TBonnin: I would use `.first()`
```suggestion
        const result = await db.knex.select('*').from<User>(`_nango_users`).where({ uuid }).first();
        return result || null;
```

packages/shared/lib/services/user.service.ts
```diff
@@ -18,6 +18,30 @@ class UserService {
...
+            .select('*', '_nango_accounts.id as account_id', '_nango_users.id as user_id')
+            .from<User>(`_nango_users`)
+            .join('_nango_accounts', '_nango_accounts.id', '_nango_users.account_id')
+            .where({ email_verification_token: token });
+
+        if (result == null || result.length == 0 || result[0] == null) {
+            return null;
+        }
+
+        return result[0];
```
TBonnin: same I would use `.first()`

packages/shared/lib/services/user.service.ts
```diff
@@ -66,10 +90,19 @@ class UserService {
...
+            {
+                email,
+                name,
+                hashed_password: hashedPassword,
+                salt: salt,
+                account_id: accountId,
+                email_verified,
+                email_verification_token: email_verified ? null : uuid.v4()
+            },
+            ['id']
```
TBonnin: nitpick: I personally prefer using an explicit `returning()` that using the unnamed second parameter of `insert`.

packages/shared/lib/services/user.service.ts
```diff
@@ -66,10 +90,19 @@ class UserService {
         return result[0];
     }
 
-    async createUser(email: string, name: string, hashedPassword: string, salt: string, accountId: number): Promise<User | null> {
-        const result: Pick<User, 'id'> = await db.knex
-            .from<User>(`_nango_users`)
-            .insert({ email: email, name: name, hashed_password: hashedPassword, salt: salt, account_id: accountId }, ['id']);
+    async createUser(email: string, name: string, hashedPassword: string, salt: string, accountId: number, email_verified = true): Promise<User | null> {
+        const result: Pick<User, 'id'> = await db.knex.from<User>(`_nango_users`).insert(
```
TBonnin: what about 
```
const [userId] = await.db(....)
return userId ? this.getUserById(userId) : null
```

packages/server/lib/controllers/v1/account/validateEmailAndLogin.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    const userAndAccount = await userService.getUserAndAccountByToken(token);
+
+    if (!userAndAccount) {
+        logger.error('Error validating user');
+        res.status(500).send({ error: { code: 'error_validating_user', message: 'There was a problem validating the user. Please reach out to support.' } });
+        return;
+    }
+
+    await userService.verifyUserEmail(userAndAccount.user_id);
```
TBonnin, line 58: what happens if this fails?

packages/server/lib/clients/workos.client.ts
khaliqgant: Yes, it is. It is used in the auth.controller https://github.com/NangoHQ/nango/blob/master/packages/server/lib/controllers/auth.controller.ts#L333

packages/server/lib/routes.ts
khaliqgant: Thanks, updated. `web.route('/api/v1/account/email/:uuid')` won't work right because of the dynamic path?

packages/server/lib/controllers/v1/account/validateEmailAndLogin.ts
khaliqgant: This is just an update in the database, in the unlikely event it fails it'll hit the catch as provided by the `asyncWrapper`

packages/shared/lib/services/user.service.ts
khaliqgant: Updated

packages/shared/lib/services/user.service.ts
khaliqgant: Didn't do this but updated to be a bit clearer

packages/server/lib/routes.ts
TBonnin: good point. let's change the condition to check the path prefix. Something like 
```    
if(['/api/v1/account', '/api/v1/login', ...].some((path) => req.path.startsWith(path))) {

```

TBonnin (approved):

packages/server/lib/controllers/v1/account/getEmailByUuid.ts
```diff
@@ -0,0 +1,39 @@
+import { z } from 'zod';
+import { asyncWrapper } from '../../../utils/asyncWrapper.js';
+import { userService } from '@nangohq/shared';
+import { requireEmptyQuery, zodErrorToHTTP } from '@nangohq/utils';
+import type { GetEmailByUuid } from '@nangohq/types';
+
+const validation = z
+    .object({
+        uuid: z.string().regex(/^[a-f0-9]{8}-[a-f0-9]{4}-4[a-f0-9]{3}-[89ab][a-f0-9]{3}-[a-f0-9]{12}$/)
```
bodinsamuel: I think there is a `uuid()`

packages/server/lib/controllers/v1/account/signin.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    if (getUser.isErr()) {
+        res.status(401).send({ error: { code: 'unauthorized', message: getUser.error.message } });
+        return;
+    }
+
+    const user = getUser.value;
+
+    if (!user.email_verified) {
+        sendVerificationEmail(user.email, user.name, user.email_verification_token as string);
```
bodinsamuel: I feel like it's a huge side effect imo

packages/server/lib/controllers/v1/account/signup.ts
```diff
@@ -0,0 +1,70 @@
...
+
+    if ((await userService.getUserByEmail(email)) !== null) {
+        res.status(400).send({ error: { code: 'user_already_exists', message: 'User with this email already exists' } });
+        return;
+    }
+
+    const account = await accountService.createAccount(`${name}'s Organization`);
+
+    if (!account) {
+        logger.error('Error creating account');
```
bodinsamuel, line 62: Since the model is not try/catching, there is good chance it's impossible to reach this condition

packages/server/lib/controllers/v1/account/signup.ts
```diff
@@ -0,0 +1,70 @@
...
+        res.status(500).send({ error: { code: 'error_creating_account', message: 'There was a problem creating the account. Please reach out to support.' } });
+        return;
+    }
+
+    const salt = crypto.randomBytes(16).toString('base64');
+    const hashedPassword = (await util.promisify(crypto.pbkdf2)(password, salt, 310000, 32, 'sha256')).toString('base64');
+    const user = await userService.createUser(email, name, hashedPassword, salt, account.id, false);
+
+    if (!user) {
```
bodinsamuel, line 71: Same here

packages/server/lib/controllers/v1/account/signup.ts
```diff
@@ -0,0 +1,70 @@
...
+        logger.error('Error creating user');
+        res.status(500).send({ error: { code: 'error_creating_user', message: 'There was a problem creating the user. Please reach out to support.' } });
+        return;
+    }
+
+    if (!user.email_verification_token) {
```
bodinsamuel, line 77: How can this happen during signup?

packages/server/lib/middleware/ratelimit.middleware.ts
```diff
@@ -65,8 +65,20 @@ function getKey(req: Request, res: Response): string {
...
+        '/api/v1/account/signin',
+        '/api/v1/account/signup',
+        '/api/v1/forgot-password',
+        '/api/v1/reset-password',
+        '/api/v1/account/signup/token',
+        '/api/v1/account/signup/invite',
+        '/api/v1/account/resend-verification-email',
+        '/api/v1/account/email',
+        '/api/v1/account/verify/code'
+    ];
```
bodinsamuel: why not `[ 'api/v1/account' ]` if we now go with prefix?

packages/shared/lib/db/migrations/20240517171724_verified_flag.cjs
```diff
@@ -0,0 +1,15 @@
+exports.up = async function (knex, _) {
+    return knex.schema.alterTable('_nango_users', function (table) {
+        table.boolean('email_verified').defaultTo(true);
+        table.string('email_verification_token').nullable();
```
bodinsamuel, line 4: Usually it's best practice to let token expire to avoid account take over when email get hacked

packages/shared/lib/services/user.service.ts
```diff
@@ -66,13 +83,29 @@ class UserService {
...
+            .insert({
+                email,
+                name,
+                hashed_password,
+                salt,
+                account_id,
+                email_verified,
+                email_verification_token: email_verified ? null : uuid.v4()
+            })
+            .returning('id');
```
bodinsamuel, line 134: why not returning('*') and removing this.getUserById() ?

packages/webapp/src/pages/EmailVerified.tsx
```diff
@@ -0,0 +1,54 @@
...
+                })
+            });
+
+            const response = await res.json();
+
+            if (res?.status !== 200) {
+                toast.error(response.error.message || 'Issue verifying email. Please try again.', { position: toast.POSITION.BOTTOM_CENTER });
+                return;
+            } else {
+                const user: User = response['user'];
```
bodinsamuel: Since you have done a massive work it's too bad to desync types there
```suggestion
                const user: ValidateEmailAndLogin = response['user'];
```

packages/webapp/src/pages/EmailVerified.tsx
```diff
@@ -0,0 +1,54 @@
...
+                analyticsTrack('web:account_signup', {
```
bodinsamuel: not sure
```suggestion
                analyticsTrack('web:account_verify', {
```

packages/webapp/src/pages/EmailVerified.tsx
```diff
@@ -0,0 +1,54 @@
...
+                    accountId: user.accountId
+                });
+
+                signin(user);
+                toast.success('Email verified successfully!', { position: toast.POSITION.BOTTOM_CENTER });
+                navigate(`/${env}/integrations`);
+            }
+        };
+
+        verifyEmail();
```
bodinsamuel: it should be debounced with a loaded or something otherwise unrelated re-render will retrigger validation

packages/webapp/src/pages/InviteSignup.tsx
```diff
@@ -52,15 +51,27 @@ export default function InviteSignup() {
...
+                email: target.email.value,
+                password: target.password.value,
+                accountId: invitedAccountID,
+                token
+            })
+        });
 
         if (res?.status === 200) {
             const data = await res.json();
             const user: User = data['user'];
```
bodinsamuel: Same comment about type desync

packages/webapp/src/pages/Signup.tsx
```diff
@@ -29,19 +23,12 @@ export default function Signup() {
...
-                email: user.email,
-                name: user.name,
-                accountId: user.accountId
-            });
-            signin(user);
-            navigate('/');
-        } else if (res != null) {
-            const errorMessage = (await res.json()).error;
-            setServerErrorMessage(errorMessage);
+            const { uuid } = await res.json();
```
bodinsamuel: Same comment

packages/webapp/src/pages/VerifyEmail.tsx
```diff
@@ -0,0 +1,94 @@
...
+    useEffect(() => {
+        if (!uuid) {
+            navigate('/');
+        }
+    }, [uuid, navigate]);
+
+    useEffect(() => {
+        const getEmail = async () => {
+            const res = await fetch(`/api/v1/account/email/${uuid}`);
+            const response = await res.json();
```
bodinsamuel: Same comment

packages/server/lib/controllers/v1/account/signupWithToken.ts
```diff
@@ -0,0 +1,84 @@
...
+        res.status(400).send({ error: { code: 'invalid_invite_token', message: 'The token used was found to be invalid.' } });
+        return;
+    }
+    const account = await accountService.getAccountById(accountId);
+    if (!account) {
+        res.status(400).send({ error: { code: 'invalid_account_id', message: 'The account ID provided is invalid.' } });
+        return;
+    }
+    const salt = crypto.randomBytes(16).toString('base64');
+    const hashedPassword = (await util.promisify(crypto.pbkdf2)(password, salt, 310000, 32, 'sha256')).toString('base64');
```
bodinsamuel, line 57: could be worth factorizing this since it's used multiples times

packages/server/lib/controllers/v1/account/signin.ts
```diff
@@ -0,0 +1,54 @@
...
+import type { WebUser, Signin } from '@nangohq/types';
+
+const validation = z
+    .object({
+        email: z.string().email(),
+        password: z.string().min(8)
+    })
+    .strict();
+
+export const signin = asyncWrapper<Signin>(async (req, res) => {
```
bodinsamuel, line 14: Because of how the auth work, you are logged at this point if the email/password is correct, so any error should logout or we should refactor the login or move this logic in the auth middleware

packages/webapp/src/pages/VerifyEmail.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+            if (res?.status === 200) {
+                const { email, verified } = response;
+
+                if (verified) {
+                    setServerErrorMessage('Email already verified. Please navigate to the login page');
```
bodinsamuel: this could just redirect imo

packages/webapp/src/pages/VerifyEmail.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+        if (res?.status === 200) {
+            toast.success('Verification email sent again!', { position: toast.POSITION.BOTTOM_CENTER });
+        } else {
+            const response = await res.json();
+            setServerErrorMessage(response.error.message || 'Unkown error...');
+        }
+    };
+
+    if (!loaded) {
```
bodinsamuel: this is seems more related to the button that the page, it flickers when clicking

bodinsamuel (commented):
Massive work, especially on the API it's so much clearer like this 🚀

packages/server/lib/controllers/v1/account/signup.ts
khaliqgant: It probably wouldn't but this is more so to please typescript since the `email_verification_token` can be null

packages/server/lib/controllers/v1/account/signup.ts
khaliqgant: Ok with this for now as this was legacy code more or less.

packages/server/lib/controllers/v1/account/signup.ts
khaliqgant: Ok with this for now as this was legacy code more or less.

packages/shared/lib/services/user.service.ts
khaliqgant: To protect against suspended users and not duplicate logic.

```
    async getUserById(id: number): Promise<User | null> {
        const result = await db.knex.select('*').from<User>(`_nango_users`).where({ id });

        if (result == null || result.length == 0 || result[0] == null) {
            return null;
        }

        if (result[0].suspended) {
            return null;
        }

        return result[0];
    }

```

packages/server/lib/routes.ts
khaliqgant: Updated.

packages/server/lib/controllers/v1/account/getEmailByUuid.ts
khaliqgant: Yup, thanks, updated in https://github.com/NangoHQ/nango/pull/2173/commits/b4654576c1c63665bdd96435a276a41067347ebe

packages/server/lib/controllers/v1/account/signin.ts
khaliqgant: Yeah, wasn't sure what to do here, but refactored it so that a button appears to re-trigger the verification email. 

https://github.com/NangoHQ/nango/assets/1724137/a97d6151-8dbb-43a2-a723-2510f8be24db

packages/webapp/src/pages/EmailVerified.tsx
khaliqgant: Spoke with Bastien on this and OK to keep the event the same

packages/webapp/src/pages/InviteSignup.tsx
khaliqgant: Updated with https://github.com/NangoHQ/nango/pull/2173/commits/b4654576c1c63665bdd96435a276a41067347ebe

packages/webapp/src/pages/VerifyEmail.tsx
khaliqgant: Updated with https://github.com/NangoHQ/nango/pull/2173/commits/b4654576c1c63665bdd96435a276a41067347ebe

packages/server/lib/controllers/v1/account/signin.ts
khaliqgant: Not totally sure what you mean here? Can you add more detail / explain the scenario or issues with this signin flow.

packages/webapp/src/pages/VerifyEmail.tsx
khaliqgant: Yeah, updated with https://github.com/NangoHQ/nango/pull/2173/commits/b4654576c1c63665bdd96435a276a41067347ebe

packages/webapp/src/pages/VerifyEmail.tsx
khaliqgant: Couldn't reproduce. Can you show an example of this?

packages/server/lib/controllers/v1/account/signin.ts
bodinsamuel: amazing 😍

packages/server/lib/controllers/v1/account/signin.ts
bodinsamuel: when you reach this function, you have a session wether your email is verified or not. So even with an error, you will still receive a valid cookie and can continue to navigate the dashboard

bodinsamuel: https://github.com/NangoHQ/nango/assets/1637651/934d2da8-4521-453e-9fc6-547059220559

packages/webapp/src/pages/VerifyEmail.tsx
bodinsamuel: https://github.com/NangoHQ/nango/assets/1637651/f7fc5775-3087-4e14-a2b1-878ee5fc86ec

packages/server/lib/controllers/v1/account/signin.ts
khaliqgant: Nice catch! Updated with https://github.com/NangoHQ/nango/pull/2173/commits/83cce999b36b7207d54369629745d193d657e055

bodinsamuel (approved):
Looks good 💯 
A few open comments but nothing blocking

packages/shared/lib/db/migrations/20240517171724_verified_flag.cjs
khaliqgant: Added an expiration column with supporting logic

PR #2174, opened 2024-05-20
fix(server): [nan-990] hosted logic for sync tab

khaliqgant:
## Describe your changes
With https://github.com/NangoHQ/nango/commit/79545622929aa4c3dca94ffb2aa64468e00460ad not having a temporal address throws and on the connection page that causes issues for self hosted. Don't route to that query if running self hosted. Also load the Auth tab if on self hosted since there will be no syncs

NAN-990

### DIFF
--- a/packages/server/lib/controllers/sync.controller.ts
+++ b/packages/server/lib/controllers/sync.controller.ts
@@ -45,6 +45,7 @@ import {
 import type { LogContext } from '@nangohq/logs';
 import { logContextGetter } from '@nangohq/logs';
 import type { LastAction } from '@nangohq/records';
+import { isHosted } from '@nangohq/utils';
 import { records as recordsService } from '@nangohq/records';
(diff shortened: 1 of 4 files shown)


packages/server/lib/controllers/sync.controller.ts
```diff
@@ -206,6 +207,11 @@ class SyncController {
                 return;
             }
 
+            if (isHosted) {
+                res.send([]);
+                return;
+            }
+
```
TBonnin, line 214: would it make more sense to short-circuit the `getSyncs` function below or even `syncService`. Otherwise any other usage of those functions requiring a temporal client instance would lead to the same problem again.

khaliqgant: We need better handling on this in general . This was a temp fix to unblock the community issue

PR #2176, opened 2024-05-20
feat(orchestrator): dry run execute action/webhook

TBonnin:
## Describe your changes
With this commit, tasks will be scheduled in the background for each action/webhook. There is no task processor yet so the tasks will end up being expired. The result of the execution is not used, only logged.
The goal is to verify orchestrator is correctly setup

I have moved `triggerAction` and `triggerWebhook` into their own `orchestrator` file. Eventually `SyncClient` and `OrchestratorService` will go away since that's where the logic to deal with temporal is.
I initially though triggering sync could be moved to the server entirely but `triggerAction` is actually used in `run.service` to send Slack notification. I therefore had to inject the orchestratorClient and pass yet another interface in `SyncRunConfig`. We are gonna need to at some point (soon?) rethink `run.service`. I know it is in shared because it is also used by the CLI for dry runs but it has a lot of dependencies and most of them aren't used when running a dry-run (no records writing, no slack notification, no reporting, ...)

### DIFF
--- a/packages/orchestrator/lib/app.ts
+++ b/packages/orchestrator/lib/app.ts
@@ -20,17 +20,17 @@ try {
     const scheduler = new Scheduler({
         dbClient,
         on: {
-            CREATED: (task) => console.log(`Task ${task.id} created`),
-            STARTED: (task) => console.log(`Task ${task.id} started`),
-            SUCCEEDED: (task) => console.log(`Task ${task.id} succeeded`),
-            FAILED: (task) => console.log(`Task ${task.id} failed`),
-            EXPIRED: (task) => console.log(`Task ${task.id} expired`),
-            CANCELLED: (task) => console.log(`Task ${task.id} cancelled`)
+            CREATED: (task) => console.log(`Task created: ${JSON.stringify(task)}`),
+            STARTED: (task) => console.log(`Task started: ${JSON.stringify(task)}`),
+            SUCCEEDED: (task) => console.log(`Task succeeded: ${JSON.stringify(task)}`),
+            FAILED: (task) => console.log(`Task failed: ${JSON.stringify(task)}`),
+            EXPIRED: (task) => console.log(`Task expired: ${JSON.stringify(task)}`),
+            CANCELLED: (task) => console.log(`Task cancelled: ${JSON.stringify(task)}`)
         }
     });
 
-    const port = envs.NANGO_ORCHESTRATOR_PORT;
     const server = getServer({ scheduler });
+    const port = envs.NANGO_ORCHESTRATOR_PORT;
     server.listen(port, () => {
         logger.info(`🚀 Orchestrator API ready at http://localhost:${port}`);
     });
--- a/packages/server/lib/controllers/sync.controller.ts
+++ b/packages/server/lib/controllers/sync.controller.ts
@@ -48,6 +48,7 @@ import type { LastAction } from '@nangohq/records';
 import { isHosted } from '@nangohq/utils';
 import { records as recordsService } from '@nangohq/records';
 import type { RequestLocals } from '../utils/express.js';
+import { getOrchestrator } from '../utils/utils.js';
 
 class SyncController {
     public async deploySync(req: Request, res: Response<any, Required<RequestLocals>>, next: NextFunction) {
@@ -414,7 +415,7 @@ class SyncController {
                 throw new NangoError('failed_to_get_sync_client');
             }
 
-            const actionResponse = await syncClient.triggerAction({
+            const actionResponse = await getOrchestrator().triggerAction({
                 connection,
                 actionName: action_name,
                 input,
--- a/packages/shared/lib/clients/orchestrator.ts
+++ b/packages/shared/lib/clients/orchestrator.ts
@@ -1,0 +1,406 @@
+import type { LogContext, LogContextGetter } from '@nangohq/logs';
+import { Err, Ok, stringifyError, metrics, getLogger } from '@nangohq/utils';
+import type { Result } from '@nangohq/utils';
+import { NangoError } from '../utils/error.js';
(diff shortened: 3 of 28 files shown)


packages/shared/lib/services/notification/slack.service.ts
```diff
@@ -303,7 +310,7 @@ class SlackService {
             return;
         }
 
-        const actionResponse = await syncClient.triggerAction<SlackActionResponse>({
+        const actionResponse = await this.orchestrator.triggerAction<SlackActionResponse>({
```
TBonnin, line 313: I guess using an action to post to Slack is dog fooding but is there any other reason why sending Slack notification is done like this.

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -0,0 +1,406 @@
...
+}
+
+export class Orchestrator {
+    private client: OrchestratorClientInterface;
+
+    public constructor(client: OrchestratorClientInterface) {
+        this.client = client;
+    }
+
+    async triggerAction<T = any>({
```
TBonnin, line 46: the 2 functions `triggerAction` and `triggerWebhook` have been moved, adding a call to dryExecute. The rest of the functions remains unchanged

packages/shared/lib/services/sync/run.service.ts
```diff
@@ -55,6 +56,7 @@ export interface SyncRunConfig {
     integrationService: IntegrationServiceInterface;
     recordsService: RecordsServiceInterface;
     dryRunService?: NangoProps['dryRunService'];
+    orchestratorClient: OrchestratorClientInterface;
```
TBonnin, line 59: at this point I am not sure I am not making things worse 😢 
Having all those "delegates" is a clear sign the level of abstraction of `run.service` is now very wrong

packages/orchestrator/lib/app.ts
```diff
@@ -20,17 +20,17 @@ try {
...
+            CREATED: (task) => console.log(`Task created: ${JSON.stringify(task)}`),
+            STARTED: (task) => console.log(`Task started: ${JSON.stringify(task)}`),
+            SUCCEEDED: (task) => console.log(`Task succeeded: ${JSON.stringify(task)}`),
+            FAILED: (task) => console.log(`Task failed: ${JSON.stringify(task)}`),
+            EXPIRED: (task) => console.log(`Task expired: ${JSON.stringify(task)}`),
+            CANCELLED: (task) => console.log(`Task cancelled: ${JSON.stringify(task)}`)
         }
     });
 
-    const port = envs.NANGO_ORCHESTRATOR_PORT;
```
bodinsamuel, line 32: why it has been removed in favor of hardcoding?

packages/server/lib/controllers/sync.controller.ts
```diff
@@ -414,7 +417,8 @@ class SyncController {
                 throw new NangoError('failed_to_get_sync_client');
             }
 
-            const actionResponse = await syncClient.triggerAction({
+            const orchestrator = new Orchestrator(new OrchestratorClient({ baseUrl: getOrchestratorUrl() }));
```
bodinsamuel: A bit "painful" to import all that stuff that the consumer does not really care about. Having a util like `getOrchestrator()` would be clearer imo

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -0,0 +1,406 @@
...
+                    },
+                    timestamp: Date.now()
+                });
+                await logCtx.info(`Starting action workflow ${workflowId} in the task queue: ${SYNC_TASK_QUEUE}`, { input: JSON.stringify(input, null, 2) });
+            }
+
+            // Execute dry-mode: no await for now
+            const groupKey: string = 'action';
+            this.dryExecute({
+                executionId: `${groupKey}:account:${connection.account_id}:environment:${connection.environment_id}:connection:${connection.id}:action:${actionName}:at:${new Date().toISOString()}:${uuid()}`,
```
bodinsamuel: connection does not contain account_id, triggerAction actually receive Connection not NangoConnection

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -0,0 +1,406 @@
...
+                args: {
+                    name: actionName,
+                    connection: {
+                        id: connection.id!,
+                        provider_config_key: connection.provider_config_key,
+                        environment_id: connection.environment_id
+                    },
+                    activityLogId,
+                    input: input
+                }
```
bodinsamuel, line 94: we both touched this, first one to merge win 😬

bodinsamuel (approved):
Tested locally, works fine 💪🏻

packages/shared/lib/services/notification/slack.service.ts
bodinsamuel: I feel like it's handy for oauth/token reason

packages/orchestrator/lib/app.ts
TBonnin: not to add yet another config env var that we never use. I will bring it back though if you prefer.

packages/shared/lib/services/notification/slack.service.ts
TBonnin: would a `proxy` call achieve the same result without having to run a script?

packages/orchestrator/lib/app.ts
TBonnin: I brought it back

packages/shared/lib/clients/orchestrator.ts
bodinsamuel: nb: I have added a `@deprecated` mention on my branch to avoid further issue since I got bite by this too

bodinsamuel (approved):

PR #2177, opened 2024-05-21
feat(integrations): add mailgun API key

jwd-dev:
## Describe your changes
Adds a integration for mailgun, can't proxy requests because the base URL changes depending on your region. https://documentation.mailgun.com/docs/mailgun/api-reference/intro/ 

A boolean connection_metadata that could impact the base URL could work.


packages/shared/providers.yaml
```diff
@@ -1318,6 +1318,11 @@ linkhut:
     proxy:
         base_url: https://api.ln.ht
     docs: https://docs.nango.dev/integrations/all/linkhut
+mailgun:
+    categories:
+        - marketing
+    auth_mode: API_KEY
```
bodinsamuel: I feel like it's a bit more complicated than just this. API_KEY means `Authorization Bearer ${apiKey}` but they are actually using a fake Basic auth with an api key, meaning `Authorization ${base64(user:apikey)}`

https://documentation.mailgun.com/docs/mailgun/api-reference/intro/

https://github.com/mailgun/mailgun.js/blob/308717a7f70b31617afd3a427b0ee3859b8aafbe/lib/Classes/common/Request.ts#L121

Not sure if we handle that? and if not how? @khaliqgant @hassan254-prog

bodinsamuel (commented):
Thanks for your contribution, I modified a small things in the doc and added the missing logo. There is a pending question around the api key, I'll let the team catchup 👍🏻 

> can't proxy requests because the base URL changes depending on your region.

I think we support that via connectionConfig, i.e:
```yaml
    proxy:
        base_url: https://${connectionConfig.region}.evaluagent.com
```

packages/shared/providers.yaml
hassan254-prog: imo, the Basic auth_mode should work here. According to the [docs](https://documentation.mailgun.com/docs/mailgun/api-reference/intro/#authentication), the curl command for authentication is `curl --user 'api:YOUR_API_KEY'` which will add `Authorization: Basic ${base64(api:YOUR_API_KEY)}` to the header where `api` is the username and `YOUR_API_KEY` the password when creating a new connection in Nango. I guess the best way to know if it will work is by actually testing the implementation. 
@jwd-dev  could you test the following and provide us with any feedback you get. 

```
mailgun:
    categories:
        - marketing
    auth_mode: BASIC
     proxy:
        base_url: https://${connectionConfig.region}.mailgun.net
        verification:
            method: GET
            endpoint: /v4/domains
    docs: https://docs.nango.dev/integrations/all/mailgun
```


When adding a new connection, `api` is the region for US-based customers, and `api.eu` is the region for EU-based customers.

packages/shared/providers.yaml
```diff
@@ -1318,6 +1318,11 @@ linkhut:
...
+    docs: https://docs.nango.dev/integrations/all/mailgun
```
bastienbeurier, line 1330: Ideally we should verify the credentials (cf. `proxy.verification` field).

packages/shared/providers.yaml
jwd-dev: I handle the fake basic auth on my side.

jwd-dev: Tested this and it works, just pushed my changes.

packages/shared/providers.yaml
jwd-dev: Not sure exactly how to do that.

jwd-dev:
@bodinsamuel Able to take a look here?

khaliqgant (approved):

bodinsamuel (approved):

bodinsamuel:
Sorry about the delay, last week was packed!
Thanks for your contribution 🙏🏻

PR #2180, opened 2024-05-21
fix(client): build esm/cjs compatible client

bodinsamuel:
## Describe your changes

Fixes NAN-994

- Automatically inject import.meta in cjs build

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,3 +4,4 @@ build/
 coverage/
 bin/
 packages/server/nango-integrations/
+packages/node-client/tests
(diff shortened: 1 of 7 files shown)


khaliqgant (approved):
(Non blocking comment) It would be nice to have a test to verify that the node client doesn't break before we publish it kind of how we have the [cli verify test](https://github.com/NangoHQ/nango/blob/master/.github/workflows/cli-verification.yaml) to make sure it doesn't break before we publish

bodinsamuel:
Added a workflow to test cjs and esm (we only build cjs but we can import from both which created the issue)

PR #2181, opened 2024-05-21
feat: allow upsert existing basic connection

descampsk:
## Describe your changes

Following our conversation with Thomas, we agree that we can allow upsert existing basic connection.

It is already allowed for Oauth2 and ApiKey connection but not for basic credentials.

This PR remove the check which disallow the import of an existing connection with basic credentials.

### DIFF
--- a/packages/shared/lib/services/connection.service.ts
+++ b/packages/shared/lib/services/connection.service.ts
@@ -251,12 +251,6 @@ class ConnectionService {
         credentials: BasicApiCredentials | ApiKeyCredentials,
         logContextGetter: LogContextGetter
     ) {
-        const connection = await this.checkIfConnectionExists(connection_id, provider_config_key, environmentId);
-
-        if (connection) {
-            throw new NangoError('connection_already_exists');
-        }
-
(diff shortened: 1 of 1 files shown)


packages/shared/lib/services/connection.service.ts
```diff
@@ -251,12 +251,6 @@ class ConnectionService {
         credentials: BasicApiCredentials | ApiKeyCredentials,
         logContextGetter: LogContextGetter
     ) {
-        const connection = await this.checkIfConnectionExists(connection_id, provider_config_key, environmentId);
-
-        if (connection) {
-            throw new NangoError('connection_already_exists');
-        }
-
```
TBonnin, line 259: I wonder if we should remove all the other usage of the `checkIfConnectionExists` function. 
https://github.com/search?q=repo%3ANangoHQ%2Fnango+checkIfConnectionExists&type=code

@khaliqgant do you see any issue with that?

khaliqgant: This is used quite a bit in the `connection.service` so I see no reason to remove that function

khaliqgant (approved):

bodinsamuel (approved):

PR #2183, opened 2024-05-21
fix(webpapp): [nan-968] helper for formatting the frequency

khaliqgant:
## Describe your changes
Format frequency to a short time unit to prevent line breaks:
image

NAN-968

### DIFF
--- a/packages/webapp/src/utils/utils.tsx
+++ b/packages/webapp/src/utils/utils.tsx
@@ -373,22 +373,6 @@ export function cn(...inputs: ClassValue[]) {
     return twMerge(clsx(inputs));
 }
 
-export function _formatFrequency(frequency: string): string {
-    if (frequency.includes('minutes') || frequency.includes('minute')) {
-        return frequency.replace('minutes', 'm').replace('minute', 'm').replace(/\s/g, '');
-    }
-
-    if (frequency.includes('hours') || frequency.includes('hour')) {
-        return frequency.replace('hours', 'h').replace('hour', 'h').replace(/\s/g, '');
-    }
-
-    if (frequency.includes('days') || frequency.includes('day')) {
-        return frequency.replace('days', 'd').replace('day', 'd').replace(/\s/g, '');
-    }
-
-    return frequency;
-}
-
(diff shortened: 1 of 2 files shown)


packages/webapp/src/utils/utils.tsx
```diff
@@ -372,3 +372,38 @@ export function parseEndpoint(endpoint: string | FlowEndpoint): string {
...
+    if (frequency.includes('hours') || frequency.includes('hour')) {
+        return frequency.replace('hours', 'h').replace('hour', 'h').replace(/\s/g, '');
+    }
+
+    if (frequency.includes('days') || frequency.includes('day')) {
+        return frequency.replace('days', 'd').replace('day', 'd').replace(/\s/g, '');
+    }
+
+    return frequency;
+}
```
TBonnin: leftover? ;)

packages/webapp/src/utils/utils.tsx
```diff
@@ -372,3 +372,38 @@ export function parseEndpoint(endpoint: string | FlowEndpoint): string {
...
+        minute: 'm',
+        hours: 'h',
+        hour: 'h',
+        days: 'd',
+        day: 'd'
+    };
+
+    for (const [unit, abbreviation] of Object.entries(unitMap)) {
+        if (frequency.includes(unit)) {
+            frequency = frequency.replace(new RegExp(unit, 'g'), abbreviation).replace(/\s+/g, '');
```
TBonnin: why the `new  Regex(...)` and not `return frequency.replace(unit, abbreviation).replace(/\s/g, '');

TBonnin (approved):

packages/webapp/src/utils/utils.tsx
khaliqgant: Updated https://github.com/NangoHQ/nango/pull/2183/commits/df10e1da979e41b8097ab4506ddd989ed09fe66c

packages/webapp/src/utils/utils.tsx
khaliqgant: 🤫 -- removed

PR #2184, opened 2024-05-21
feat(webapp): [nan-983] show slack banner if conditions are met

khaliqgant:
## Describe your changes
* Show Slack banner on connection page for both tabs
image
* Show the connected slack name on the Environment Settings page
image

* Refactor the environment endpoint to have a better type

NAN-983 & NAN-982


packages/shared/lib/services/connection.service.ts
```diff
@@ -423,13 +423,15 @@ class ConnectionService {
...
-    public async getConnectionConfig(connection: Connection): Promise<ConnectionConfig> {
+    public async getConnectionConfig(connection: Pick<Connection, 'connection_id' | 'provider_config_key' | 'environment_id'>): Promise<ConnectionConfig> {
+        console.log(connection);
         const result = await db.knex.from<StoredConnection>(`_nango_connections`).select('connection_config').where({
             connection_id: connection.connection_id,
             provider_config_key: connection.provider_config_key,
             environment_id: connection.environment_id,
             deleted: false
         });
+        console.log(result);
```
TBonnin: 🚫

TBonnin (approved):

packages/shared/lib/services/connection.service.ts
khaliqgant: 😅 removed.

packages/webapp/src/pages/Connection/Show.tsx
```diff
@@ -225,6 +249,18 @@ We could not retrieve and/or refresh your access token due to the following erro
                         Authorization
                     </li>
                 </ul>
+
+                {!slackIsConnected && (
+                    <Info size={8} color="blue" showIcon={false} padding="mt-4 p-1">
+                        <div className="flex text-sm items-center">
+                            <IntegrationLogo provider="slack" height={6} width={6} classNames="flex mr-2" />
+                            Receive instant monitoring alerts on Slack.{' '}
+                            <span onClick={createSlackConnection} className="ml-1 cursor-pointer underline">
```
bodinsamuel: Should be a button. 
Nitpicking: buttons with async actions are better disabled when loading to avoid double send

packages/server/lib/controllers/environment.controller.ts
```diff
@@ -77,10 +89,36 @@ class EnvironmentController {
...
+                        slack_notifications_channel = connectionConfig['incoming_webhook.channel'];
+                    }
+                }
+            }
+
             const environmentVariables = await environmentService.getEnvironmentVariables(environment.id);
 
             res.status(200).send({
-                environment: { ...environment, env_variables: environmentVariables, host: baseUrl, uuid: account.uuid, email: user.email }
+                environmentAndAccount: {
```
bodinsamuel, line 114: seems like a lot of change while you could have added `slack_notifications_channel` outside of this object imo

packages/webapp/src/pages/EnvironmentSettings.tsx
```diff
@@ -616,9 +596,9 @@ export const EnvironmentSettings: React.FC = () => {
                                 </label>
                             </div>
                             <div className="">
-                                <Button className="items-center" variant="primary" onClick={slackIsConnected ? disconnectSlack : connectSlack}>
-                                    <IntegrationLogo provider="slack" height={6} width={6} classNames="" />
-                                    {slackIsConnected ? 'Disconnect' : 'Connect'}
+                                <Button className="items-center" variant="primary" onClick={slackIsConnected ? disconnectSlack : createSlackConnection}>
+                                    <IntegrationLogo provider="slack" height={5} width={6} classNames="" />
+                                    {slackIsConnected ? `Disconnect ${slackConnectedChannel}` : 'Connect'}
                                 </Button>
```
bodinsamuel: Right now this button fails silently if you don't have a Slack provider setup. OnFailure should have at least displayed a toast (which didn't) but a message or a disabled button would be even better

bodinsamuel (commented):
Seems correct, I'll test again tomorrow with a slack app setup

packages/server/lib/controllers/environment.controller.ts
khaliqgant: The big change here was really to make the environment object clearer as discussed in https://github.com/NangoHQ/nango/pull/2135#discussion_r1598342146

packages/webapp/src/pages/EnvironmentSettings.tsx
khaliqgant: Updated with https://github.com/NangoHQ/nango/pull/2184/commits/67ac5a6bf7b8f9fbfc99626c38bc06b25a30a1b0

packages/webapp/src/pages/Connection/Show.tsx
khaliqgant: Yeah, following the design here, but added in logic to disable the link with https://github.com/NangoHQ/nango/pull/2184/commits/67ac5a6bf7b8f9fbfc99626c38bc06b25a30a1b0. The element is now a button even if it doesn't appear to be (thus keeping to the design)

bodinsamuel:
While testing I noticed "prod" is hardcoded in `getAdminAuthInfo`. Can't really go further since slack is not allowing http redirect

khaliqgant:
> While testing I noticed "prod" is hardcoded in `getAdminAuthInfo`. Can't really go further since slack is not allowing http redirect

NANGO_CALLBACK_URL=https://redirectmeto.com/http://localhost:3003/oauth/callback

in the .env should resolve that

bodinsamuel:
> NANGO_CALLBACK_URL=https://redirectmeto.com/http://localhost:3003/oauth/callback

this env var doesn't exists but I see in the codebase the redirectmeto is added automatically if `NANGO_SERVER_URL` is not specified (which was my case). Don't know which scopes to add though

> While testing I noticed "prod" is hardcoded in getAdminAuthInfo.

Still an issue, unless on purpose

bodinsamuel (approved):

PR #2185, opened 2024-05-21
chore: moving integrations/webhook folder out of shared

TBonnin:
It is only used by server package so moving it there, reducing shared surface


packages/server/lib/controllers/webhook.controller.ts
```diff
@@ -1,9 +1,10 @@
 import type { Request, Response, NextFunction } from 'express';
 import tracer from 'dd-trace';
 import type { Span } from 'dd-trace';
-import { routeWebhook, featureFlags, environmentService } from '@nangohq/shared';
+import { featureFlags, environmentService } from '@nangohq/shared';
 import { metrics } from '@nangohq/utils';
 import { logContextGetter } from '@nangohq/logs';
+import routeWebhook from '../webhook/webhook.manager';
```
bodinsamuel: not sure why linter did not caught that
```suggestion
import routeWebhook from '../webhook/webhook.manager.js';
```

bodinsamuel (approved):

PR #2186, opened 2024-05-21
chore: add orchestrator to deploy Github Actions

TBonnin:
Adding orchestrator to the deploy github actions (using the unified container image)

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -1,26 +1,27 @@
-name: "[Release] Deploy service"
+name: '[Release] Deploy service'
 
 on:
     workflow_dispatch:
         inputs:
             stage:
                 type: choice
-                description: "stage to deploy to, defaults to staging"
+                description: 'stage to deploy to, defaults to staging'
                 required: true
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yaml
```diff
@@ -1,26 +1,27 @@
-name: "[Release] Deploy service"
+name: '[Release] Deploy service'
```
TBonnin, line 1: Last block is the only modification of this file. The rest is just formatting

scripts/build_docker.sh
```diff
@@ -38,7 +38,7 @@ fi
 # Move to here no matter where the file was executed
 cd "$(dirname "$0")"
 
-tags="-t nangohq/nango:${ENV}-${GIT_HASH}"
```
bodinsamuel: there is a good reason I was forced to build env specific image, the webapp inside needs different .env.file depending of the env. I wish we could remove that (we just need to sideload sentry and posthog key) but until now it should not be removed

bodinsamuel (approved):
Nice 🚀

PR #2187, opened 2024-05-21
fix(logs): consolidate metadata

bodinsamuel:
## Describe your changes

Contributes to NAN-908

- Add missing account, environment, config, syncConfig, connection
It's 95-99% done, there are still spots where it's partially missing, mostly on places where we have to fetch data gradually (oauth, controllers, etc.). To get the correct data I had to add more db calls, even with all the refacto I tried to do, it's still not ideal to carry partial objects everywhere, but it's getting there. 

- Renamed `sync` field to `syncConfig` 
I initially used sync (as in `_nango_syncs`) but I realized in this PR that actions do not have a row there which is confusing. So I had to change the IDs used to fill this field, since we are not live yet it's fine. (mapping is append compatible, the old fields are still there but not conflicting)

- Added `providerName` field to be able to display an icon near provider unique name

- Typed Temporal arguments correctly (needs to be handled by the new system @TBonnin)

- Changed a few things in the UI to accommodate added/missing payload


packages/jobs/lib/integration.service.ts
```diff
@@ -97,17 +97,6 @@ class IntegrationService implements IntegrationServiceInterface {
...
-
-            if (!script && activityLogId && writeToDb) {
-                await createActivityLogMessage({
-                    level: 'error',
-                    environment_id: environmentId,
-                    activity_log_id: activityLogId,
-                    content: `Unable to find integration file for ${syncName}`,
-                    timestamp: Date.now()
-                });
-                await logCtx?.error(`Unable to find integration file for ${syncName}`);
```
bodinsamuel, line 110: I think there was a merge error, we were logging twice the same things (see above)

packages/server/lib/controllers/proxy.controller.ts
```diff
@@ -135,32 +132,42 @@ class ProxyController {
...
+                        environment_id,
+                        activity_log_id: activityLogId,
+                        timestamp: Date.now(),
+                        content: 'Provider configuration not found'
+                    });
+                    await logCtx.error('Provider configuration not found');
+                    await logCtx.failed();
+                }
 
                 throw new NangoError('unknown_provider_config');
```
bodinsamuel, line 154: there was an issue, without an activityLogId it would not throw

packages/shared/lib/clients/sync.client.ts
```diff
@@ -569,25 +584,25 @@ class SyncClient {
                     },
                     timestamp: Date.now()
                 });
-                await logCtx.info(`Starting action workflow ${workflowId} in the task queue: ${SYNC_TASK_QUEUE}`, { input: JSON.stringify(input, null, 2) });
+                await logCtx.info(`Starting action workflow ${workflowId} in the task queue: ${SYNC_TASK_QUEUE}`);
             }
 
+            const args: ActionArgs = {
```
bodinsamuel: While trying to pass down data I noticed args were not typed and were not synced

packages/shared/lib/services/sync/config/config.service.ts
```diff
@@ -932,3 +932,23 @@ export async function getSyncConfigsByConfigIdForWebhook(environment_id: number,
 
     return result;
 }
+
+export async function getSyncConfigRaw(opts: { environmentId: number; config_id: number; name: string; isAction: boolean }): Promise<SyncConfig | null> {
```
bodinsamuel, line 936: there is technically `getSyncConfigByParams` available but it does query provider again for no reason, and report errors

packages/server/lib/controllers/sync.controller.ts
```diff
@@ -373,6 +374,11 @@ class SyncController {
             }
 
             const provider = await configService.getProviderConfig(providerConfigKey, environmentId);
+            if (!provider) {
+                res.status(400).json({ error: { code: 'not_found' } });
```
khaliqgant: Shouldn't this be a 404 and elsewhere below where `not_found` is used

packages/server/lib/controllers/sync.controller.ts
```diff
@@ -373,6 +374,11 @@ class SyncController {
...
+                return;
+            }
+            const syncConfig = await getSyncConfigRaw({ environmentId, config_id: provider.id!, name: action_name, isAction: true });
```
khaliqgant, line 383: What if this isn't found or comes back as null?

khaliqgant (approved):
Non blocking comments

packages/server/lib/controllers/sync.controller.ts
bodinsamuel: good call indeed

packages/server/lib/controllers/sync.controller.ts
bodinsamuel: indeed, forgot about this one 👌🏻

packages/jobs/lib/crons/autoIdleDemo.integration.test.ts
```diff
@@ -32,13 +33,38 @@ describe('Auto Idle Demo', async () => {
...
+                track_deletes: false,
+                type: SyncConfigType.SYNC,
+                auto_start: false,
+                attributes: {},
+                metadata: {},
+                pre_built: true,
+                is_public: false,
+                enabled: true
+            })
+            .returning('id');
```
TBonnin, line 67: can we not use `sync.service.createSync` function?

packages/logs/lib/models/helpers.ts
```diff
@@ -5,15 +5,19 @@ import { z } from 'zod';
...
 export interface FormatMessageData {
-    account?: { id: number; name?: string };
+    account?: { id: number; name: string };
     user?: { id: number } | undefined;
-    environment?: { id: number; name?: string } | undefined;
+    environment?: { id: number; name: string } | undefined;
     connection?: { id: number; name: string } | undefined;
-    config?: { id: number; name: string } | undefined;
-    sync?: { id: string; name: string } | undefined;
+    config?: { id: number; name: string; provider: string } | undefined;
```
TBonnin: should we find a better name for config which depending on the context might mean very different things?

packages/shared/lib/clients/sync.client.ts
TBonnin: Args are free form data attached to a workflow/activity. From temporal point of view they are just a blob of data passed along. To solve this problem in the orchestrator we would need to change the level of abstraction and not execute anything but explicitely execute an action/webhook/sync. I was already thinking about it but it is not implemented yet

packages/shared/lib/services/environment.service.ts
```diff
@@ -143,10 +148,12 @@ class EnvironmentService {
             q.where('secret_key_hashed', hash);
         } else if ('publicKey' in opts) {
             q.where('_nango_environments.public_key', opts.publicKey);
-        } else if ('accountId' in opts) {
-            q.where('_nango_environments.account_id', opts.accountId).where('_nango_environments.name', opts.envName);
         } else if ('accountUuid' in opts) {
             q.where('_nango_accounts.uuid', opts.accountUuid).where('_nango_environments.name', opts.envName);
+        } else if ('envName' in opts) {
```
TBonnin: can we use `accountId` which would make it immune to ordering change?

packages/shared/lib/services/sync/config/config.service.ts
TBonnin: is it not the same as `getSyncConfigsByConfigId`?

packages/types/lib/logs/messages.ts
```diff
@@ -31,7 +31,7 @@ export type MessageState = 'waiting' | 'running' | 'success' | 'failed' | 'timeo
  */
 export interface MessageOpSync {
     type: 'sync';
-    action: 'pause' | 'unpause' | 'run' | 'run_full' | 'cancel' | 'init';
```
TBonnin, line 34: why renaming `run_full` but keep `run`

packages/jobs/lib/crons/autoIdleDemo.integration.test.ts
bodinsamuel: I need a sync_config not a sync there is no function for that in sync/config.service.ts

packages/shared/lib/services/environment.service.ts
bodinsamuel: yes definitely

packages/types/lib/logs/messages.ts
bodinsamuel: `run` is the actual execution wether it's full or incremental

packages/logs/lib/models/helpers.ts
bodinsamuel: agreed, I just tried to limit confusion by following the DB/libs. There is something very unfortunate with bad table name and I'm not sure we did a good job of abstracting this rn. Happy to change to integration

packages/shared/lib/services/sync/config/config.service.ts
bodinsamuel: More or less, it doesn't take name as mandatory params. I'm never too sure when using ORM should we have one single abstract function or N ones for each context

packages/shared/lib/clients/sync.client.ts
bodinsamuel: having types should be enough imo, but strong runtime check is definitely interesting

packages/logs/lib/models/helpers.ts
TBonnin: we overuse the word 'config' I think so anything other than 'config' makes more sense to me. Up to you though

bodinsamuel: changed everything to `integration`

packages/shared/lib/services/environment.service.ts
bodinsamuel: changed

bodinsamuel:
@TBonnin addressed some of your comments, let me know which one you feel are still blocking ☺️

TBonnin (approved):

TBonnin:
> @TBonnin addressed some of your comments, let me know which one you feel are still blocking ☺️

nothing is blocking. Sorry I thought I add approved

PR #2188, opened 2024-05-21
chore: move hooks related logic to server

TBonnin:
More refactor, moving hooks logic to server, out of shared

hooks are almost exclusively used in server. The only usage outside of server is in `connection.service.js`. I didn't want to push the refactoring too far but I believe `connection.service.js` could be splitted in 2: 
- `packages/shared/model/connections` that provides logic to fetch/update/etc... connections from/to the database.
- `packages/server/service/connections` with all the business logic related to connection


packages/server/lib/controllers/connection.controller.ts
```diff
@@ -625,14 +624,30 @@ class ConnectionController {
...
+
                 const [imported] = await connectionService.importOAuthConnection(
                     connection_id,
                     provider_config_key,
                     provider,
                     environment.id,
                     account.id,
                     oAuthCredentials,
-                    logContextGetter
+                    connCreatedHook
```
TBonnin, line 650: The split of `connection.service` mentioned in the PR description would avoid passing the hook function as argument but I didn't want this refactoring to get out of control, so for now a hook function is passed as a parameter to the importConnection methods

bodinsamuel (approved):
Seems correct 👌🏻 
A lot of have moved so hard to tell

PR #2191, opened 2024-05-22
fix(yaml): handle optional connectionConfig parameters in provider URLs

oytuntez:
## Describe your changes
Providers can be configured with connectionConfig parameters, such as:
`authorization_url: https://${connectionConfig.subdomain}.api.accelo.com/oauth2/v0/authorize`

However, these connectionConfig parameters are seamingly required and not optional. This PR aims to make the string replacement of these parameters optional and `dot` / `.` aware for subdomain cases.

The UI still considers these parameters required, but at least you can use the params optionally in your authorization URL constructions via SDK or backend.

Also see tests. Here are some examples:

Given URL: `https://${subdomain}.service.com/oauth/authorize`
Given params: `{ subdomain: 'myaccount' }`
Resulting URL: `https://myaccount.service.com/oauth/authorize`

Given params: `{ subdomain: null }`
Resulting URL: `https://service.com/oauth/authorize`

- [x ] I added tests, otherwise the reason is: 
- [?] I added observability, otherwise the reason is:
- [?] I added analytics, otherwise the reason is:

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -20,6 +20,7 @@
 # NANGO_DB_POOL_MIN=<PICK-INT-OR-SKIP>
 # NANGO_DB_POOL_MAX=<PICK-INT-OR-SKIP>
 #
+# RECORDS_DATABASE_URL=${RECORDS_DATABASE_URL:-postgresql://nango:nango@nango-db:5432/nango}
 #
 # - Configure server and worker port (current value is the default for running Nango locally).
 #
--- a/packages/server/lib/utils/utils.unit.test.ts
+++ b/packages/server/lib/utils/utils.unit.test.ts
@@ -1,7 +1,7 @@
 import { expect, describe, it } from 'vitest';
 import { getConnectionMetadataFromTokenResponse, parseConnectionConfigParamsFromTemplate, getAdditionalAuthorizationParams } from './utils.js';
 import type { Template as ProviderTemplate } from '@nangohq/shared';
-import { AuthModes } from '@nangohq/shared';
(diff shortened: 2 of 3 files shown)


TBonnin:
Thank you for your contribution @oytuntez 
Can you share an example of a provider which would benefit from this change? Thank you

oytuntez:
One of the services we are connected to, [Clio](http://clio.com/), has a specific subdomain for EU region and nothing for US region, so something like: `eu.app.clio.com` VS `app.clio.com`. It doesn't have a subdomain for US region.

TBonnin:
> One of the services we are connected to, [Clio](http://clio.com/), has a specific subdomain for EU region and nothing for US region, so something like: `eu.app.clio.com` VS `app.clio.com`. It doesn't have a subdomain for US region.

I would argue that there is 2 subdomains: `app` and `eu.app`. 😃 Can you have a required `connectionConfiguration` param where you pass either of those?

oytuntez:
Haha, I like that idea, let me sleep on it. It's definitely hacky and I'd need to move "subdomain" info to other external systems (rather than "region"), but worth thinking about it.

We already have a fork due to a previously rejected PR, so I'll keep this code on ours for now.

TBonnin:
Thank you @oytuntez I am gonna close this PR for now. Feel free to ping us if you encounter an issue that Nango doesn't address yet. 
I would also encourage you to ping us on Nango Community Slack https://nango.dev/slack before to implement any changes so we can discuss about the best way to address any problem.

PR #2193, opened 2024-05-22
feat(logs): UI filtering

bodinsamuel:
## Describe your changes

Fixes NAN-1011
Fixes NAN-848
Contributes to NAN-1004

- **Update schema** to accommodate filtering and aggregation
We can't aggregate on text field so we need to use index time nested fields to have to both keyword and text. Because this schema is incompatible and I don't want to deal with migration I just bumped the index name. (everything will be trashed with os -> es migration anyway)

- `POST /logs/filters`
Get filters list for UI. Works more or less like the v1, except it looks inside opensearch. That means if there is no logs you won't find the relevant data, at first I was not convinced by this but the slow response from the DB and the "lack" of full text search made a compelling case to use aggregation directly from logs.

- **UI Operation filtering**
You can now filter by status, type, integration, connection and script. 

- Renamed `token` operation to `auth` 
It was hard to understand for end user and is tied to auth anyway.

## How to test?

- http://localhost:3000/dev/logs
- Trigger some action, sync, auth

## Results

https://github.com/NangoHQ/nango/assets/1637651/8cabef39-eb88-4043-a53c-cd2ee0a38020


packages/webapp/src/components/ui/Command.tsx
```diff
@@ -0,0 +1,124 @@
+import * as React from 'react';
```
bodinsamuel, line 1: Imported from Shadcn Command

packages/webapp/src/components/ui/Dialog.tsx
```diff
@@ -0,0 +1,72 @@
+import * as React from 'react';
```
bodinsamuel, line 1: Imported from Shadcn Dialog

packages/webapp/src/components/ui/Popover.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
```
bodinsamuel, line 1: Imported from Shadcn Popover

packages/webapp/src/hooks/useLogs.tsx
```diff
@@ -88,5 +101,55 @@ export function useSearchMessages(env: string, body: SearchMessages['Body']) {
...
+
+export function useSearchFilters(enabled: boolean, env: string, body: SearchFilters['Body']) {
+    const [loading, setLoading] = useState<boolean>(false);
+    const [data, setData] = useState<SearchFilters['Success']>();
+    const [error, setError] = useState<SearchFilters['Errors']>();
+
+    async function fetchData() {
+        setLoading(true);
+        try {
+            const res = await fetch(`/api/v1/logs/filters?env=${env}`, {
```
bodinsamuel, line 121: useSWR not supporting POST with cache is really annoying. Even regular POST calls are barely supported. 
I'll need to look for a solution because it's spamming the backend for no reason, especially when playing with filters

packages/logs/lib/models/messages.ts
```diff
@@ -179,9 +241,60 @@ export async function listMessages(opts: {
...
+            }
+        }
+    };
+
+    if (opts.search) {
+        (body.query.bool!.must! as any).push({ match_phrase_prefix: { [aggField]: { query: opts.search } } });
+    }
+    const res = await client.search<{ aggregations: { byName: { buckets: any[] } } }>({
+        index: indexMessages.index,
+        size: 0,
```
TBonnin, line 290: does that mean no limit?

packages/webapp/src/pages/Logs/Search.tsx
```diff
@@ -87,13 +161,19 @@ export const LogsSearch: React.FC = () => {
...
+                <div className="w-full">{/* <Input before={<MagnifyingGlassIcon className="w-5 h-5" />} placeholder="Search operations..." /> */}</div>
                 <MultiSelect label="Status" options={statusOptions} selected={states} defaultSelect={statusDefaultOptions} onChange={setStates} all />
+                <TypesSelect selected={types} onChange={setTypes} />
+                <SearchableMultiSelect label="Integration" selected={integrations} category={'config'} onChange={setIntegrations} />
+                <SearchableMultiSelect label="Connection" selected={connections} category={'connection'} onChange={setConnections} />
+                <SearchableMultiSelect label="Script" selected={syncs} category={'syncConfig'} onChange={setSyncs} />
+                <Button variant="zombieGray" size={'xs'}>
+                    <LightningBoltIcon />
+                    Live
+                </Button>
```
TBonnin, line 173: This button is doing nothing for now. Is that on purpose?

packages/webapp/src/pages/Logs/components/TypesSelect.tsx
```diff
@@ -0,0 +1,118 @@
...
+                            className="bg-pure-black text-white flex gap-1 items-center px-1.5 rounded-xl"
+                            onPointerDown={reset}
+                            onKeyDown={(e) => {
+                                if (['Enter', ' '].includes(e.key)) {
+                                    reset(e);
+                                }
+                            }}
+                        >
+                            <CrossCircledIcon />
+                            {selected.length}
```
TBonnin, line 64: ux: The selected count makes the previous buttons jump and shift left and right when selecting. That's a bit strange I think

packages/webapp/src/pages/Logs/components/TypesSelect.tsx
```diff
@@ -0,0 +1,118 @@
...
+                        <button
+                            className="bg-pure-black text-white flex gap-1 items-center px-1.5 rounded-xl"
+                            onPointerDown={reset}
+                            onKeyDown={(e) => {
+                                if (['Enter', ' '].includes(e.key)) {
+                                    reset(e);
+                                }
+                            }}
+                        >
+                            <CrossCircledIcon />
```
TBonnin, line 63: ux: It is not super clear that the cross circle is actually a button within a button that is clickable

packages/webapp/src/pages/Logs/components/TypesSelect.tsx
bodinsamuel: yep agreed too, it's something I need to discuss with gabrielle

packages/webapp/src/pages/Logs/components/TypesSelect.tsx
bodinsamuel: noted too 👌🏻

packages/webapp/src/pages/Logs/Search.tsx
bodinsamuel: yep, there will be a calendar there

packages/logs/lib/models/messages.ts
bodinsamuel: size: 0 means "do not return records", preferable when requesting only aggregation

bodinsamuel:
Last push fixed an issue with Sentry, in production mode when using useSearchParams it was infinite refreshing. I had to upgrade and modify a few things. Also fixed an issue with log_level

packages/logs/lib/es/schema.ts
```diff
@@ -68,7 +95,7 @@ const props: Record<keyof MessageRow, opensearchtypes.MappingProperty> = {
 };
 
 export const indexMessages: opensearchtypes.IndicesCreateRequest = {
-    index: envs.NANGO_LOGS_OS_INDEX ?? 'messages',
+    index: `20240522_${envs.NANGO_LOGS_OS_INDEX ?? 'messages'}`,
```
khaliqgant, line 98: What is this `20240522_` prefix?

packages/server/lib/controllers/apiAuth.controller.ts
```diff
@@ -292,7 +292,7 @@ class ApiAuthController {
 
         try {
             logCtx = await logContextGetter.create(
-                { id: String(activityLogId), operation: { type: 'auth' }, message: 'Authorization Basic' },
+                { id: String(activityLogId), operation: { type: 'auth', action: 'connect' }, message: 'Authorization Basic' },
```
khaliqgant: `connect` is a weird action to me. Maybe something a bit clearer, `createConnection` or something?

packages/server/lib/controllers/connection.controller.ts
```diff
@@ -77,7 +77,7 @@ class ConnectionController {
                     content: 'Unknown connection'
                 });
                 const logCtx = await logContextGetter.create(
-                    { id: String(activityLogId), operation: { type: 'token' }, message: 'Get connection web' },
```
khaliqgant, line 80: I'm not sure if I completely agree with getting rid of `token` and replacing it with `auth`. There is a difference from doing an `authorization` (auth) call vs. fetching a token or a token refresh failure. 

Also feel like this is partially a product decision to what the end user sees.

packages/server/lib/controllers/v1/logs/searchFilters.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    const val = validation.safeParse(req.body);
+    if (!val.success) {
+        res.status(400).send({ error: { code: 'invalid_body', errors: zodErrorToHTTP(val.error) } });
+        return;
+    }
+
+    const env = res.locals['environment'];
+    const body: SearchFilters['Body'] = val.data;
+    const rawOps = await model.listFilters({ accountId: env.account_id, environmentId: env.id, category: body.category, limit: 20, search: body.search });
```
khaliqgant, line 34: is the limit not configurable?

packages/server/lib/hooks/connection/post-connection.ts
```diff
@@ -101,7 +101,7 @@ async function execute(createdConnection: RecentlyCreatedConnection, provider: s
                     content: `Post connection script failed with the error: ${errorString}`
                 });
                 const logCtx = await logContextGetter.create(
-                    { id: String(activityLogId), operation: { type: 'token' }, message: 'Authentication' },
+                    { id: String(activityLogId), operation: { type: 'auth', action: 'post_connection' }, message: 'Authentication' },
```
khaliqgant, line 104: Agree this should be of type `auth` here instead of token

packages/webapp/src/hooks/useLogs.tsx
khaliqgant: Ugh

packages/logs/lib/es/schema.ts
bodinsamuel: just versionning to avoid the migration of the data in staging

packages/server/lib/controllers/v1/logs/searchFilters.ts
bodinsamuel: not right now, it's pure UI stuff

packages/server/lib/controllers/connection.controller.ts
bodinsamuel: it's mostly about categorization and avoid one category just for token refresh. Token on it's own is not really something that speaks to the customers imo, whereas auth -> token makes more sense.

Regarding the product this is aligned with what we designed (cf figma)
Screenshot 2024-05-24 at 14 13 13

packages/server/lib/controllers/apiAuth.controller.ts
bodinsamuel: happy to put create_connection

bodinsamuel: Changed

khaliqgant (approved):

TBonnin (approved):

PR #2195, opened 2024-05-23
fix(jobs): [nan-1010] use void

khaliqgant:
## Describe your changes
As explained by Thomas:
image

Use void since we're not explicitly handling the promise resolution of the webhook sending

Tested this against returning a 504 (what the customer is returning from their webhook) from a webhook with Temporal and the task execution completes even as the webhook continues to retry. With `async` usage, the temporal activity is still in running state while the webhook retries.

### DIFF
--- a/packages/shared/lib/services/notification/webhook.service.ts
+++ b/packages/shared/lib/services/notification/webhook.service.ts
@@ -18,7 +18,7 @@ import { stringifyError } from '@nangohq/utils';
 
 dayjs.extend(utc);
 
-const RETRY_ATTEMPTS = 7;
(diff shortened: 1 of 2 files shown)


bodinsamuel (approved):

TBonnin (approved):
I would also maybe lower the retry attempts to 7 or 8 instead of 10. The 10th retry almost wait for a minute

khaliqgant:
> I would also maybe lower the retry attempts to 7 or 8 instead of 10. The 10th retry almost wait for a minute

Dropped to 7

PR #2196, opened 2024-05-23
feat(logs): opensearch -> elasticsearch

bodinsamuel:
## Describe your changes

Fixes NAN-1000

Due do lack of tooling, outdated sdk and minimal cloud support, we have decided to use elasticsearch instead of opensearch. 

- Use elasticsearch 
I haven't changed the env var to avoid creating more trouble but not sure if it's a good idea. Let me know

- Disabled elasticsearch in self-hosted docker compose
Decision was made to have console log and empty state in the UI if this feature is disabled. So that's it's not required in the critical path.


khaliqgant (approved):
> I haven't changed the env var to avoid creating more trouble but not sure if it's a good idea. Let me know

Definitely think its worth changing the env var

TBonnin:
> > I haven't changed the env var to avoid creating more trouble but not sure if it's a good idea. Let me know
> 
> Definitely think its worth changing the env var

I second that

docker-compose.yaml
```diff
@@ -58,31 +59,30 @@ services:
...
+    #     volumes:
+    #         - elasticsearch-data1:/usr/share/elasticsearch/data
+    #     ports:
+    #         - 9500:9200
+    #         - 9600:9300
+    #     environment:
+    #         - discovery.type=single-node
+    #         - xpack.security.enabled=false
+    #     networks:
+    #         - nango
```
TBonnin, line 81: commented out?

TBonnin (approved):

docker-compose.yaml
bodinsamuel: The main issue we had in slack community is people not willing/understanding why it was loaded by default. Especially in oauth mode it's not that useful, so we decided to have a screen in the UI that tell how to enable this (aka uncommenting)

bodinsamuel:
> Definitely think its worth changing the env var 

Renamed ☺️

PR #2202, opened 2024-05-24
feat(orchestrator): add long polling option for /output endpoint

TBonnin:
## Describe your changes

Right now the orchestrator client, when executing a task, loops indefinitely until the task completes, fetching the output repeatedly, which leads to a ton of http requests and db queries. 
This commit adds a "longpolling" option for the `output` endpoint where the server keeps the connection opens and waits for the completion event to be emitted (or timeout to be reached).
Code is more complicated since we now rely on on event being emitted by an eventEmitter but the benefit is a less naive and less request/query hungry logic

### DIFF
--- a/packages/orchestrator/lib/routes/v1/task/taskId/output.ts
+++ b/packages/orchestrator/lib/routes/v1/task/taskId/output.ts
@@ -1,9 +1,11 @@
 import { z } from 'zod';
 import type { JsonValue } from 'type-fest';
-import type { Scheduler, TaskTerminalState } from '@nangohq/scheduler';
+import type { Scheduler, Task, TaskState } from '@nangohq/scheduler';
 import type { ApiError, Endpoint } from '@nangohq/types';
 import type { EndpointRequest, EndpointResponse, RouteHandler, Route } from '@nangohq/utils';
 import { validateRequest } from '@nangohq/utils';
+import type { EventEmitter } from 'node:events';
+import { getEventId } from '../../../../events.js';
 
 type Output = Endpoint<{
     Method: typeof method;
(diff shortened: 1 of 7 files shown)


khaliqgant (approved):

packages/orchestrator/lib/routes/v1/task/taskId/output.ts
```diff
@@ -1,52 +1,86 @@
...
+            }
+        };
+        const onCompletion = (completedTask: Task) => {
+            cleanupAndRespond((res) => res.status(200).json({ state: completedTask.state, output: completedTask.output }));
+        };
+        const timeout = setTimeout(() => {
+            cleanupAndRespond((res) => res.status(204).send());
+        }, waitForCompletionTimeoutMs);
+
+        eventEmitter.once(eventId, onCompletion);
```
bodinsamuel, line 62: In case of deploy (or else) the caller never get an answer, I guess this is acceptable risk but should be known

packages/orchestrator/lib/routes/v1/task/taskId/output.ts
```diff
@@ -1,52 +1,86 @@
...
-            case 'FAILED':
-            case 'EXPIRED':
-            case 'CANCELLED':
-                return res.status(200).json({ state: task.value.state, output: task.value.output });
+        if (req.query.waitForCompletion && (task.value.state === 'CREATED' || task.value.state === 'STARTED')) {
+            await new Promise((resolve) => resolve(timeout));
+        } else {
+            cleanupAndRespond((res) => res.status(200).json({ state: task.value.state, output: task.value.output }));
         }
+        return;
```
bodinsamuel, line 74: no answer on timeout?

bodinsamuel (commented):
👍🏻 Looks good!
Note that the previous scenario was causing strain on the DB, this one will likely cause strain on CPU/memory because of all open connections (and also limited by max open sockets). I'm not entirely sure the tradeoff is in your favor but we'll see

bodinsamuel (approved):
(wanted to approve)

packages/orchestrator/lib/routes/v1/task/taskId/output.ts
TBonnin: unless we have graceful shutdown of servers, this could happen with any requests. I agree it is more likely for long connection though

packages/orchestrator/lib/routes/v1/task/taskId/output.ts
TBonnin: yep 204. the timeout function is defined on L58

TBonnin:
> 👍🏻 Looks good! Note that the previous scenario was causing strain on the DB, this one will likely cause strain on CPU/memory because of all open connections (and also limited by max open sockets). I'm not entirely sure the tradeoff is in your favor but we'll see

it is indeed a tradeoff but scaling servers is in my experience easier than scaling dbs 😄

PR #2204, opened 2024-05-24
feat(server): [nan-981] prep for implementing error reporting UI on connection authorization sub

khaliqgant:
## Describe your changes
* In order to log connection failures we need to pass along the full `account` and `environment` to correctly log. Adds the object to prep for explicit logging and for insertion into the coming soon notifications table. Also explicitly logs a connection failure so we have something to link to for the notification.
* Also cleans up webhook logic to send twice if there is a secondary webhook url

NAN-981


packages/server/lib/controllers/connection.controller.ts
```diff
@@ -61,29 +62,29 @@ class ConnectionController {
...
+                if (credentialResponse.isOk()) {
+                    connection = credentialResponse.value;
+                } else {
+                    error = credentialResponse.error;
+                }
+            } else {
+                ({ success, error, response: connection } = await connectionService.getConnection(connectionId, providerConfigKey, environment.id));
             }
 
-            if (!connection) {
```
khaliqgant, line 72: These logs were never hit anyway.

packages/shared/lib/services/connection.service.ts
```diff
@@ -360,8 +368,6 @@ class ConnectionService {
             }
         }
 
-        await this.updateLastFetched(connection?.id as number);
```
khaliqgant, line 363: When `getConnection` is called it actually isn't fetching credentials. The correct place is `getConnectionDetails`

packages/shared/lib/services/connection.service.ts
```diff
@@ -618,21 +602,59 @@ class ConnectionService {
...
+                        environment,
+                        integration: config ? { id: config.id, name: config.unique_key, provider: config.provider } : undefined,
+                        connection: { id: connection.id, name: connection.connection_id }
+                    }
+                );
+
+                await logCtx.error('Failed to refresh credentials', error);
+                await logCtx.failed();
+
+                // TODO now insert into notifications to recall this error and link to it
```
khaliqgant, line 648: This entire PR to just get to do this here 😅

packages/server/lib/controllers/connection.controller.ts
```diff
@@ -61,29 +62,30 @@ class ConnectionController {
...
+                success = credentialResponse.isOk();
 
-                return;
+                if (credentialResponse.isOk()) {
+                    connection = credentialResponse.value;
+                } else {
+                    error = credentialResponse.error;
+                }
+            } else {
+                ({ success, error, response: connection } = await connectionService.getConnection(connectionId, providerConfigKey, environment.id));
```
TBonnin: you refactored `getConnectionCredentials` to return a `Result`. Might be worthwhile to also refactor `getConnection`. I think that would facilitate the error handling here since you wouldn't need a success or error temp variable

packages/server/lib/utils/web-socket-error.ts
```diff
@@ -31,6 +31,13 @@ export function InvalidCallbackOAuth2(): WSErr {
     };
 }
 
+export function EnvironmentOrAccountNotFound(): WSErr {
+    return {
+        type: 'account_or_environment_retrieval_err',
+        message: `The account or environment could not be retrieved.`
+    };
+}
+
```
TBonnin, line 40: is this used somewhere?

packages/shared/lib/services/connection.service.ts
```diff
@@ -55,6 +56,13 @@ const logger = getLogger('Connection');
 
 type KeyValuePairs = Record<string, string | boolean>;
 
+interface IntegrationValuesWithAccountAndEnvironment {
+    connection_id: string;
+    provider_config_key: string;
+    account: Account;
+    environment: Environment;
+}
+
```
TBonnin: not sure declaring an interface is worth it here. Duplicating the struct definition twice is fine imho. You might even even been able to please the typescript compiler with something like this below:
```
public async getOldConnections({
        days,
        limit
    }: {
        days: number;
        limit: number;
    }): Promise<{ connection_id: string; provider_config_key: string; account: Account; environment: Environment }[]> {

        type T = Awaited<ReturnType<ConnectionService['getOldConnections']>>;
        const result = await db
            .knex<StoredConnection>(`_nango_connections`)
...
            .select<T>(
...
```

TBonnin (approved):

packages/server/lib/controllers/connection.controller.ts
```diff
@@ -61,29 +62,30 @@ class ConnectionController {
...
+            let error: NangoError | null = null;
+            let connection: Connection | null = null;
+            if (instantRefresh) {
+                const credentialResponse = await connectionService.getConnectionCredentials({
+                    account,
+                    environment,
+                    connectionId,
+                    providerConfigKey,
+                    logContextGetter,
+                    instantRefresh: false
```
bodinsamuel: you set `false` here in the `instantRefresh` condition, am I missreading?

packages/server/lib/controllers/connection.controller.ts
```diff
@@ -116,10 +118,12 @@ class ConnectionController {
...
-                connection.credentials.type === ProviderAuthModes.OAuth2 ||
-                connection.credentials.type === ProviderAuthModes.App ||
-                connection.credentials.type === ProviderAuthModes.OAuth2CC
+                !instantRefresh &&
+                (connection.credentials.type === ProviderAuthModes.OAuth2 ||
+                    connection.credentials.type === ProviderAuthModes.App ||
+                    connection.credentials.type === ProviderAuthModes.OAuth2CC)
             ) {
                 const {
                     success,
```
bodinsamuel: It's quiet confusing to have a refresh in getConnectionCredentials and one here, and I feel like this call is already handled in connection.service.ts. This is the exact same condition but maybe I'm missing something

packages/shared/lib/services/connection.service.ts
```diff
@@ -618,21 +602,59 @@ class ConnectionService {
...
+                    {
+                        account,
+                        environment,
+                        integration: config ? { id: config.id, name: config.unique_key, provider: config.provider } : undefined,
+                        connection: { id: connection.id, name: connection.connection_id }
+                    }
+                );
+
+                await logCtx.error('Failed to refresh credentials', error);
+                await logCtx.failed();
```
bodinsamuel, line 646: I feel like we should log the success here because `getConnectionCredentials` is called in 5 differents places and only the getConnectionWeb is logging the success

bodinsamuel (commented):
You will have some conflict regarding the logs otherwise just a few comments 👌🏻

packages/server/lib/utils/web-socket-error.ts
khaliqgant: https://github.com/NangoHQ/nango/blob/khaliq/nan-981-implement-error-reporting-ui-on-connection-authorization-sub/packages/server/lib/controllers/oauth.controller.ts#L1077

packages/server/lib/controllers/connection.controller.ts
khaliqgant: I originally had this because `getConnectionDetails` does a refresh if needed but that disrupts the intention of the instantRefresh option since that is a force credential refresh. Thanks, updated.

packages/shared/lib/services/connection.service.ts
khaliqgant: Updated

packages/server/lib/controllers/connection.controller.ts
khaliqgant: Yup, you're absolutely right. Simplified with https://github.com/NangoHQ/nango/pull/2204/commits/00f150214423c66a746f099f2c167c0edc97d2a9

packages/shared/lib/services/connection.service.ts
khaliqgant: Only `getConnectionWeb` can have `instantRefresh` as true which is why it is the only place that logs a success.

packages/server/lib/controllers/connection.controller.ts
khaliqgant: I didn't want to touch too much and now adjusted the connection controller so the logic is less convoluted.

packages/shared/lib/services/connection.service.ts
bodinsamuel: okay so if I understand well we don't log all refresh, only the one that were forced, right? 👍🏻

bodinsamuel (approved):

packages/shared/lib/services/connection.service.ts
khaliqgant: That is correct

PR #2205, opened 2024-05-24
fix(orchestrator): harden type safety of orchestrator client execute function

TBonnin:
This commit is adding a `type: 'webhook' | 'action' | 'sync'` to the task args so the processors knows what kind of processing must happens.
It also make the orchestrator aware of Nango concepts like webhook, actions, etc... making the interface types more specific (instead of having `args: JsonValue`, partially addressing the issue with Temporal where everything input/output are `any`). This also keeps the translation between nango concepts and nango-agnostic scheduler contained within the orchestrator. When dequeing tasks the orchestrator will also translate task into `TaskAction`, `TaskWebhook`, etc...

### DIFF
--- a/packages/orchestrator/lib/client.ts
+++ b/packages/orchestrator/lib/client.ts
@@ -17,7 +17,33 @@ interface SchedulingProps {
         startedToCompleted: number;
         heartbeat: number;
     };
-    args: JsonValue;
+    args: JsonValue & { type: 'action' | 'webhook' | 'sync' };
+}
+
+interface TExecuteActionArgs {
+    args: {
+        name: string;
+        connection: {
+            id: number;
+            provider_config_key: string;
+            environment_id: number;
+        };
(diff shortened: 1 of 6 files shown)


packages/orchestrator/lib/client.ts
```diff
@@ -111,4 +139,28 @@ export class OrchestratorClient {
...
+        const { args, ...rest } = props;
+        const schedulingProps = {
+            ...rest,
+            args: {
+                ...args,
+                type: 'webhook' as const
+            }
+        };
+        return this.execute(schedulingProps);
+    }
```
TBonnin, line 165: I went with a function for each type (an `executeSync` will be added soon) instead of a single function with a Union type parameter since types are already quite complex

khaliqgant (approved):

bodinsamuel (approved):
⌨️

PR #2208, opened 2024-05-26
fix(version): store version in a file instead of reading package.json

bodinsamuel:
## Describe your changes

Fixes NAN-1020

I have done it I don't know how many times, and everytime I think it's going to be different. Reading from package.json is clunky and not portable as soon as you bundle, use an exotic install, use a serveless host, etc. We had two reports last week, after many tries I could not reproduce so instead of wasting time I implemented this regular solution. You can find this way of doing things in many libs (i.e: [stripe](https://github.com/stripe/stripe-node/blob/162ba379ee32801762a3f0f74ee4cd18f99566b5/src/stripe.core.ts#L52), [algolia](https://github.com/algolia/algoliasearch-client-javascript/blob/5828389d96c7d8b416ee4bd3017b980d3927bee4/packages/client-common/src/version.ts), etc.)

- Store version in file to secure fast and stable loading across env
Once again I had to do the work two times, using shared right now would be a mistake but as soon as we clean it up I hope we can use it.

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -34,6 +34,9 @@ if [[ ! "$VERSION" =~ ^([0-9]+\.[0-9]+\.[0-9]+|0\.0\.1-[0-9a-fA-F]{40})$ ]]; the
     exit 1
 fi
 
+sed -E -i '' "s/NANGO_VERSION = '[0-9.]+/NANGO_VERSION = '$VERSION/" "$GIT_ROOT_DIR/packages/shared/lib/version.ts"
+sed -E -i '' "s/NANGO_VERSION = '[0-9.]+/NANGO_VERSION = '$VERSION/" "$GIT_ROOT_DIR/packages/node-client/lib/version.ts"
+
 npm ci
 npm run ts-build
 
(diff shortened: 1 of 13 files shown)


khaliqgant (approved):
😅

scripts/publish.sh
```diff
@@ -34,6 +34,16 @@ if [[ ! "$VERSION" =~ ^([0-9]+\.[0-9]+\.[0-9]+|0\.0\.1-[0-9a-fA-F]{40})$ ]]; the
...
 fi
 
+# increment stored version
+# NB: macos and linux have different "sed" that don't edit in place the same way
+pushd "$GIT_ROOT_DIR/packages"
+sed -E "s/NANGO_VERSION = '[0-9a-fA-F.-]+/NANGO_VERSION = '$VERSION/" ./shared/lib/version.ts >tmp
+mv tmp ./shared/lib/version.ts
+sed -E "s/NANGO_VERSION = '[0-9a-fA-F.-]+/NANGO_VERSION = '$VERSION/" ./node-client/lib/version.ts >tmp
+mv tmp ./node-client/lib/version.ts
+popd
```
TBonnin, line 44: This is only run by CI so it doesn't have to be "portable". Otherwisse `awk -i inplace` might be work the same on all OSes

bodinsamuel: there is always a time when we'll need to test this script locally. Not familiar with awk, I'll look

PR #2212, opened 2024-05-27
fix(jobs): local runner for enterprise

khaliqgant:
## Describe your changes
Issues with the runner failing health checks for enterprise. Using a a remote runner requires the API call from jobs to the runner stay open for the duration of the sync. This can be problematic as the connection can drop or be terminated for a number of reasons. This piece of [code](https://github.com/NangoHQ/nango/blob/master/packages/jobs/lib/integration.service.ts#L124-L133] should be refactored soon so that this logic is more portable across different cloud providers besides just our own.

### DIFF
--- a/packages/jobs/lib/runner/runner.ts
+++ b/packages/jobs/lib/runner/runner.ts
@@ -29,7 +29,7 @@ export function getRunnerId(suffix: string): string {
 
 export async function getOrStartRunner(runnerId: string): Promise<Runner> {
     const waitForRunner = async function (runner: Runner): Promise<void> {
-        const timeoutMs = 5000;
+        const timeoutMs = isEnterprise ? 60000 : 5000;
         let healthCheck = false;
         const startTime = Date.now();
         while (!healthCheck && Date.now() - startTime < timeoutMs) {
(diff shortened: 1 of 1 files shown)


TBonnin (approved):

packages/jobs/lib/runner/runner.ts
```diff
@@ -29,7 +29,7 @@ export function getRunnerId(suffix: string): string {
 
 export async function getOrStartRunner(runnerId: string): Promise<Runner> {
     const waitForRunner = async function (runner: Runner): Promise<void> {
-        const timeoutMs = 5000;
+        const timeoutMs = isEnterprise ? 60000 : 5000;
```
TBonnin, line 32: not sure this is necessary since for remote runner we can assume it is always on (ie: it is not managed by nango)

bodinsamuel (commented):
Missing some context here, why would we drop the RemoteRunner?
Is it a temporary fix?

khaliqgant:
> Missing some context here, why would we drop the RemoteRunner? Is it a temporary fix?

https://nangohq.slack.com/archives/C067NRBH1LL/p1716911965222909

bodinsamuel (approved):
I see 👍🏻  
context could have been added here for posterity imo

khaliqgant:
> Missing some context here, why would we drop the RemoteRunner? Is it a temporary fix?

Fair, description updated.

PR #2213, opened 2024-05-27
feat(logs): pagination and infinite scroll

bodinsamuel:
## Context

Fixes NAN-1004
Fixes NAN-997
Contributes to NAN-908

Took me a day just to handle this :|
It was much more painful than anticipated for many reasons: 
- plenty of filters to combine
- live refresh that prepend and also updates the current rows
- infinite scroll that append but only if nothing else is changing in the UI  

When filtering the cursor changes, but when something change too fast: race condition can happen, so I had to use ref and tricks. I also had to aggregate rows into an array from which I could: update, append and prepend; I had to handle writing in this array from multiples places which in React is not easy. 
The result is honestly quite flaky and very hard to grasp, I don't give it 2 weeks until something is broken.
Ideally we should remove at least the "update", only query before or after the cursor, and update the states of the rows with websockets.

## Changes 

- **Add cursor** to searchOperations

- **Handle infinite scroll and live refresh for operations**

- Fix layout scroll box
The DashboardLayout component was a mix of fixed and flex box which made the scroll area not consistent. For example a page without content would still have a scrollbar. I also added an option to allow a content to stretch the full width.

## Test

**(I added some fake lag to help me spot loading)**

https://github.com/NangoHQ/nango/assets/1637651/3b4256b6-31ea-47e1-9004-07ee85b9c384


packages/webapp/src/components/ui/ScrollArea.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from 'react';
```
bodinsamuel, line 1: I didn't used it right now but I know the scroll is messed up on windows, it will be used when polishing

packages/logs/lib/models/messages.ts
```diff
@@ -128,20 +132,24 @@ export async function listOperations(opts: {
         });
     }
 
+    const cursor = opts.cursor ? parseCursor(opts.cursor) : undefined;
```
TBonnin, line 135: can `parseCursor` throw an exception?

packages/logs/lib/models/messages.ts
```diff
@@ -128,20 +132,24 @@ export async function listOperations(opts: {
...
         size: opts.limit,
-        sort: [{ createdAt: 'desc' }, '_score'],
+        sort: [{ createdAt: 'desc' }, 'id'],
         track_total_hits: true,
+        search_after: cursor,
         query
     });
     const hits = res.hits;
 
+    const total = typeof hits.total === 'object' ? hits.total.value : hits.hits.length;
```
TBonnin, line 146: can we use the null coalescing operator here? `hits.total?.value ?? hits.hits.length;`

khaliqgant:
FYI i took a look at this last night but would like to due a more thorough review and pull it locally to see it in action. Let me know if this is blocking and I can expedite the review @bodinsamuel

packages/server/lib/controllers/v1/logs/searchOperations.ts
```diff
@@ -31,12 +32,14 @@ const validation = z
...
                 ])
             )
+            .max(20)
             .optional()
             .default(['all']),
-        integrations: z.array(z.string()).optional().default(['all']),
-        connections: z.array(z.string()).optional().default(['all']),
-        syncs: z.array(z.string()).optional().default(['all']),
-        period: z.object({ from: z.string().datetime(), to: z.string().datetime() }).optional()
+        integrations: z.array(z.string()).max(10).optional().default(['all']),
```
TBonnin, line 38: curious about the max 10. How did you choose the value?

TBonnin (approved):

packages/server/lib/controllers/v1/logs/searchOperations.ts
bodinsamuel: it's a conservative value right now, totally arbitrary to avoid slow search

packages/logs/lib/models/messages.ts
bodinsamuel: yes, fine to me it will bubble up

packages/logs/lib/models/messages.ts
bodinsamuel: Can't because this object has multiple shapes unfortunately `: number | estypes.SearchTotalHits | undefined`

khaliqgant:
Some unrelated feedback for logs:
- when creating an unauthenticated connection the connection name doesn't show in the logs 
image
image
Happens for other connection creation as well it seems: 
image

- Why is this a cursor pointer? 
image
- If there is no payload should this box show at all? 
image
- The back arrow seems to not be centered / mis aligned 
image
- This is off center 
image

The live refresh jumps to a new pane if a new log comes in while you have another pane open which is a bit jarring

packages/webapp/src/pages/Logs/Search.tsx
```diff
@@ -213,15 +303,32 @@ export const LogsSearch: React.FC = () => {
...
+                                );
+                            })}
+                        </Table.Row>
                     )}
                 </Table.Body>
             </Table.Table>
+            {data && data.pagination.total > 0 && data.data.length > 0 && cursor.current && readyToDisplay && (
+                <div ref={bottomScrollRef}>
+                    <Button disabled={loading} variant="active" className="w-full justify-center" onClick={() => loadMore()}>
+                        {loading && <Spinner size={1} />} Load More
```
khaliqgant: Shouldn't it say "Loading more..." or just "Loading"

packages/logs/lib/models/messages.ts
```diff
@@ -128,20 +132,24 @@ export async function listOperations(opts: {
...
     return {
-        count: typeof hits.total === 'object' ? hits.total.value : hits.hits.length,
+        count: total,
         items: hits.hits.map((hit) => {
             return hit._source!;
-        })
+        }),
+        cursor: hits.hits.length > 0 && total > hits.hits.length && hits.hits.length >= opts.limit ? createCursor(hits.hits[hits.hits.length - 1]!) : null
```
khaliqgant: Written this way this logic is a bit hard to read

packages/webapp/src/pages/Logs/Search.tsx
bodinsamuel: good point, I'll put loading...

bodinsamuel:
> Why is this a cursor pointer?

It's a placeholder for the moment, they insist on going with icons but it's hard to understand the meaning behind each icons.

> If there is no payload should this box show at all?

For UI consistency, we decided to always show it. We might revisit that later.

> The back arrow seems to not be centered / mis aligned

✅ Fixing, thanks

> This is off center

✅ Actually it's not needed, thanks

> the live refresh jumps to a new pane

✅ Ah yes I have fixed that in a subsequent PR, but let me backport it here.

> when creating an unauthenticated connection the connection name doesn't show in the logs

Yes I have noted that in my todo, I need to fetch the connection when receiving the callback (for all type of integration)

packages/logs/lib/models/messages.ts
bodinsamuel: I have tried something, let me know

khaliqgant (approved):

PR #2214, opened 2024-05-28
feat(logs): messages infinite scroll and live refresh

bodinsamuel:
## Context

Fixes NAN-998
Contributes to NAN-908

More or less the same than https://github.com/NangoHQ/nango/pull/2213 The main diff here is that we don't need to update the current page state, so we use two cursor `cursorBefore` and `cursorAfter` the former to live refresh (prepend) and the later to do infinite scroll (append).

## Changes

- **Add `cursorBefore` and `cursorAfter`** to searchMessages
Note: there is no concept of search_before in Elasticsearch, you just reverse the sorting order and re-reverse the results to achieve it.

- Implement live refresh and infinite scroll in the UI

## Test

**(note the scrollbar changing position because we prepend, it's very hard to manage unless we use our own scroll bar

https://github.com/NangoHQ/nango/assets/1637651/4c802379-7eb7-4234-b5c8-b2ca6dff8526


khaliqgant:
Some unrelated feedback for logs:
- when creating an unauthenticated connection the connection name doesn't show in the logs 
image
image
Happens for other connection creation as well it seems: 
image

- Why is this a cursor pointer? 
image
- If there is no payload should this box show at all? 
image
- The back arrow seems to not be centered / mis aligned 
image
- This is off center 
image

The live refresh jumps to a new pane if a new log comes in while you have another pane open which is a bit jarring

packages/webapp/src/pages/Logs/ShowMessage.tsx
```diff
@@ -66,7 +66,7 @@ export const ShowMessage: React.FC<{ message: MessageRow }> = ({ message }) => {
                                 return { code: { padding: '0', whiteSpace: 'pre-wrap' } };
                             }}
                         >
-                            {JSON.stringify({ error: message.error || undefined, output: message.meta || undefined }, null, 2)}
+                            {JSON.stringify({ error: message.error?.message || undefined, output: message.meta || undefined }, null, 2)}
```
TBonnin, line 69: is the `|| undefined` necessary?

packages/logs/lib/models/messages.ts
```diff
@@ -249,20 +253,46 @@ export async function listMessages(opts: {
...
     });
     const hits = res.hits;
 
+    const total = typeof hits.total === 'object' ? hits.total.value : hits.hits.length;
+    const totalPage = hits.hits.length;
+    const items = hits.hits.map((hit) => {
+        return hit._source!;
+    });
+    if (opts.cursorBefore) {
+        // In case we set before we have to reverse the message since we inverted the sort
```
TBonnin, line 285: Can we add a link to https://github.com/elastic/elasticsearch/issues/29449 Might be handy if we ever run into the limitation mentioned in the issue

TBonnin (approved):

packages/webapp/src/pages/Logs/ShowMessage.tsx
bodinsamuel: yes because messge.error can be `null` and null is not automatically removed from json stringify, but `undefined` will be stripped.

PR #2216, opened 2024-05-28
feat(logs): daily index, policy, retention

bodinsamuel:
## Context

Fixes NAN-1033

I tried to keep the ops related to Elasticsearch simple but it's obvious that I was just procrastinating something I knew was going to be annoying. Right now we have one index for everything which would be fine if it was small, but it's going to be big; and we know we need to delete old logs too.

This introduce daily index, with automatic index deletion after 15days, and an alias for easy search. 
The main drawbacks, when we update an Operation (state and enrich), we need to know which index is in. And it's more difficult than anticipated: the operation can have started in the last 7 days, we can't write with index alias, we can't getById with index alias either. So we need to search for the operation using a regular search and use the `createdAt` to build the index name.

## Changes 

> [!Note]
> All interesting changes are in `packages/logs/`

- **Introduce index template**
Index template allows us to set the mappings, settings once per "pattern" 

- **Introduce daily index**
Daily index are automatically created by the ingest pipeline. This pipeline takes the `createdAt`, format it, pick the index name and create it if not existing using the index template.

- **Introduce retention policy**
This policy is simple, it's either a hot data tier or should be deleted. We can, in the future, make it more granular with warm and cold tier that helps reduce cost but for now it covers the basics.

## Related changes

- **Update LogContext** instantiation when required
Because we can't update an operation without having the operation, it made things a bit more complicated but I think I managed to have minimal side effect by splitting the **LogContext** class into two: **stateless and normal**. 

1. **Stateless** can only insert logs, handy **for persist** so it doesn't impact log performance. 
1. **Normal** LogContext now requires an await when instantiating it by operation id, this happen in **oauth callback** response and few others places


TBonnin (approved):

TBonnin (commented):
Love the PR description. Thank you

khaliqgant (approved):

PR #2219, opened 2024-05-29
fix(connection): upsert return full connection, logs

bodinsamuel:
## Describe your changes

Fixes NAN-1041

- Upserting connection now returns full row 
Some part of the code still query connection again after that, ideally we would reuse the object but I tried to limit the number of changes

- Add missing context for logs where needed

### DIFF
--- a/packages/shared/lib/services/connection.service.ts
+++ b/packages/shared/lib/services/connection.service.ts
@@ -89,29 +89,35 @@ class ConnectionService {
 
             encryptedConnection.updated_at = new Date();
 
-            await db.knex.from<StoredConnection>(`_nango_connections`).where({ id: storedConnection.id, deleted: false }).update(encryptedConnection);
+            const connection = await db.knex
+                .from<StoredConnection>(`_nango_connections`)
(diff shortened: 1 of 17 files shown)


packages/shared/lib/services/connection.service.ts
```diff
@@ -124,7 +130,7 @@ class ConnectionService {
         accountId: number
     ): Promise<ConnectionUpsertResponse[]> {
         const storedConnection = await this.checkIfConnectionExists(connectionId, providerConfigKey, environment_id);
-        const config_id = await configService.getIdByProviderConfigKey(environment_id, providerConfigKey);
+        const config_id = await configService.getIdByProviderConfigKey(environment_id, providerConfigKey); // TODO remove that
```
TBonnin, line 133: why does it need to be removed?

TBonnin (approved):

packages/shared/lib/services/connection.service.ts
bodinsamuel: we are already fetching the config earlier in most cases

khaliqgant (approved):

PR #2220, opened 2024-05-29
feat(logs): timeout old operations

bodinsamuel:
## Describe your changes

Fixes NAN-1036

Since we have only 15days retention and syncs are unlimited it's mostly for cosmetic and some smaller better time scoped operations (action, oauth, webhook).

- Add expiration date to all operations
- Cron to set `timeout` to all expired operations

## How to test?

- Trigger an oauth but don't complete the process and wait 5-10minutes
or
- Add an operation, change the expiresAt manually, wait for the cron

(The query in curl)
```sh
#!/usr/bin/env bash

curl -X POST "http://localhost:55132/20240528_messages.2024-05-30/_search" -H 'Content-Type: application/json' -d'
{
  "query": {
            "bool": {
                "must": [{ "range": { "expiresAt": { "lt": "now" } } }],
                "must_not": { "exists": { "field": "parentId" } },
                "should": [{ "term": { "state": "waiting" } }, { "term": { "state": "running" } }]
            }
        }
}
'
```


packages/logs/lib/models/helpers.ts
```diff
@@ -57,10 +58,11 @@ export function getFormattedMessage(
...
         meta: meta || data.meta || null,
 
-        createdAt: data.createdAt || new Date().toISOString(),
-        updatedAt: data.updatedAt || new Date().toISOString(),
+        createdAt: data.createdAt || now.toISOString(),
+        updatedAt: data.updatedAt || now.toISOString(),
         startedAt: data.startedAt || null,
-        endedAt: data.endedAt || null
+        endedAt: data.endedAt || null,
+        expiresAt: data.operation ? data.expiresAt || new Date(now.getTime() + 7 * 86400 * 1000).toISOString() : null
```
TBonnin: why 7 days by default. Can any operation last longer than 24hours?

packages/server/lib/controllers/onboarding.controller.ts
```diff
@@ -432,7 +432,12 @@ class OnboardingController {
             }
 
             logCtx = await logContextGetter.create(
-                { id: String(activityLogId), operation: { type: 'action' }, message: 'Start action' },
+                {
+                    id: String(activityLogId),
+                    operation: { type: 'action' },
+                    message: 'Start action',
+                    expiresAt: new Date(Date.now() + 86400 * 1000).toISOString()
+                },
```
TBonnin, line 440: 1 day for an action that's pretty generous

TBonnin (approved):

packages/logs/lib/models/helpers.ts
bodinsamuel: technically yes I believe, syncs can run as long as they need if the heartbeat works no?

packages/server/lib/controllers/onboarding.controller.ts
bodinsamuel: it's just for the heartbeat, unfortunately I have no way to have something more precise. But if everything works well everything should be handled by temporal/v2 :D

packages/logs/lib/models/helpers.ts
khaliqgant: > syncs can run as long as they need if the heartbeat works no?

A sync can't run longer than 24 hours https://github.com/NangoHQ/nango/blob/master/packages/jobs/lib/workflows.ts#L6

packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -86,7 +86,12 @@ class OAuthController {
 
         try {
             logCtx = await logContextGetter.create(
-                { id: String(activityLogId), operation: { type: 'auth', action: 'create_connection' }, message: 'Authorization OAuth' },
+                {
+                    id: String(activityLogId),
+                    operation: { type: 'auth', action: 'create_connection' },
+                    message: 'Authorization OAuth',
+                    expiresAt: new Date(Date.now() + 300 * 1000).toISOString()
```
khaliqgant: Feel like this expiresAt should be a constant somewhere since it is repeated in so many places

```
new Date(Date.now() + 300 * 1000).toISOString()
```

packages/server/lib/controllers/onboarding.controller.ts
khaliqgant: Would be nice to abstract out these times to one place in case we need to adjust

khaliqgant (changes requested):
Abstraction request

packages/server/lib/controllers/onboarding.controller.ts
bodinsamuel: yeah good point, actually I wasn't aware of the constant you shared on previous message so maybe I can expose them globally somehow

bodinsamuel:
Abstracted the values, I had to expose them through utils let me know.
(could also be helpful for you thomas)

packages/utils/lib/workflows.ts
```diff
@@ -0,0 +1,13 @@
...
+export const SYNC_MAX_ATTEMPTS = 3;
+export const MAX_SYNC_DURATION = ms(SYNC_TIMEOUT);
+
+export const ACTION_TIMEOUT = '15m';
+export const ACTION_MAX_ATTEMPTS = 1; // no retry
+export const MAX_ACTION_DURATION = ms(ACTION_TIMEOUT);
+
+export const WEBHOOK_TIMEOUT = '15m';
+export const WEBHOOK_MAX_ATTEMPTS = 3;
+export const MAX_WEBHOOK_DURATION = ms(WEBHOOK_TIMEOUT);
```
bodinsamuel, line 13: Note I didn't removed them from jobs/workflows.ts because temporal can't compile utils 🤦🏻

bodinsamuel:
Okay this time should be fine, I could not replace webhooks timing because I can't import logs in shared

packages/utils/lib/workflows.ts
TBonnin: I think it is fine. I just have to remember to remove it when getting rid of temporal

TBonnin (approved):

khaliqgant (approved):
Thanks

PR #2221, opened 2024-05-29
feat(orchestrator): implement task processor

TBonnin:
## Describe your changes
This PR is implementing the orchestrator processor which, for now, is being run by jobs. 
The processor is behind a flag `flag:orchestrator:dryrun:process:global`. It is still in dryrun mode since no real processing is happening, tasks are either immediately succeeded or failed. It is the last step before actually processing webhook/actions

### Notes
- I have added orchestrator endpoints to support dequeueing, set task state, heartbeat and their respective functions in the orchestrator client
- A processor dequeues tasks in a infinite loop and process the task via the `process` functions passed as argument. 
- A processor also check if pending tasks have been completed (ie: they might have expired or be cancelled) and send abort signal.
- In jobs, 2 processor worker_thread are being started. One for action and one for webhook to reproduce what we have right now with webhooks being in a separate temporal queue

Tested in staging


packages/jobs/lib/app.ts
```diff
@@ -1,25 +1,41 @@
...
+    const processorFlagTimer = setInterval(async () => {
+        const isProcessorEnabled = await getFlag();
+        if (isProcessorEnabled && processor.isStopped()) {
+            processor.start();
+        }
+        if (!isProcessorEnabled && !processor.isStopped()) {
+            processor.stop();
+        }
+    }, 1000);
+
```
TBonnin, line 38: using `setInterval` because of the flag so it can be enabled/disabled without restarting jobs. Once stable it will be a simple `processor.start()`

packages/jobs/lib/processor/process.ts
```diff
@@ -0,0 +1,34 @@
...
+    // TODO: Implement action processing
+    // Returning a successful result for now
+    return Ok({ taskId: task.id, dryrun: true });
+}
+
+async function webhook(task: TaskWebhook): Promise<Result<JsonValue>> {
+    // TODO: Implement action processing
+    // Returning an error for now
+    return Err(`Not implemented: ${JSON.stringify({ taskId: task.id })}`);
+}
```
TBonnin: dry-run: not doing real processing yet, just marking action as successful and webhook as failed for testing purpse

packages/orchestrator/lib/clients/client.ts
```diff
@@ -0,0 +1,264 @@
...
+import { putRoute as putTaskRoute } from '../routes/v1/tasks/taskId.js';
+import { postRoute as postHeartbeatRoute } from '../routes/v1/tasks/taskId/heartbeat.js';
+import type { Result, Route } from '@nangohq/utils';
+import { Ok, Err, routeFetch, stringifyError } from '@nangohq/utils';
+import type { Endpoint } from '@nangohq/types';
+import type { ClientError, SchedulingProps, ExecuteActionProps, ExecuteProps, ExecuteReturn, ExecuteWebhookProps, TaskAction, TaskWebhook } from './types.js';
+import { validateTask } from './validate.js';
+import type { JsonValue } from 'type-fest';
+
+export class OrchestratorClient {
```
TBonnin, line 14: everything until `dequeue` was already there. I have moved the file in a `clients` subdir

packages/orchestrator/lib/routes/health.ts
```diff
@@ -10,12 +10,9 @@ type Health = Endpoint<{
 const path = '/health';
 const method = 'GET';
 
-export const handler: RouteHandler<Health> = {
+export const getHandler: RouteHandler<Health> = {
```
TBonnin: Since each route file corresponds to a path it is possible to have multiple Endpoint/handler per file, one for each HTTP method. So I renamed the  Endpoint and handlers to make it clearer

packages/orchestrator/lib/clients/client.ts
```diff
@@ -0,0 +1,264 @@
...
+            params: { taskId }
+        });
+        if ('error' in res) {
+            return Err({
+                name: res.error.code,
+                message: res.error.message || `Error heartbeating task '${taskId}'`,
+                payload: { taskId }
+            });
+        } else {
+            return Ok(undefined);
```
khaliqgant, line 200: should this be `null`?

khaliqgant (approved):

packages/orchestrator/lib/clients/client.ts
TBonnin: `null` is not assignable to type `void`. Only `undefined` is. 🤷

packages/orchestrator/lib/routes/v1/search.ts
```diff
@@ -0,0 +1,56 @@
+import { z } from 'zod';
+import type { Scheduler, Task } from '@nangohq/scheduler';
+import type { ApiError, Endpoint } from '@nangohq/types';
+import type { EndpointRequest, EndpointResponse, RouteHandler, Route } from '@nangohq/utils';
+import { validateRequest } from '@nangohq/utils';
+
```
khaliqgant: What is the purpose of search in the context of the processor?

TBonnin: It is the search task endpoint. The processor must be able to check if pending tasks have been cancelled so it need to be able to search for specific tasks. 
This endpoint could also be used in internal tooling to display list of tasks per queue/state/etc...

bodinsamuel:
I'm late in my reviews, checking this in a couple of hours 👍🏻

packages/orchestrator/lib/clients/processor.ts
```diff
@@ -0,0 +1,143 @@
...
+        this.stopped = true;
+        this.process = process;
+        this.groupKey = opts.groupKey;
+        this.orchestratorClient = opts.orchestratorClient;
+        this.queue = new Queue(opts.maxConcurrency);
+        this.abortControllers = new Map();
+        this.checkForTerminatedInterval = opts.checkForTerminatedInterval || 1000;
+    }
+
+    public async start() {
```
bodinsamuel:
```suggestion
    public start() {
```

packages/orchestrator/lib/clients/processor.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    public stop() {
+        this.stopped = true;
+        if (this.terminatedTimer) {
+            clearInterval(this.terminatedTimer);
+        }
+    }
+
+    private async checkForTerminatedTasks() {
+        if (!this.stopped && this.abortControllers.size > 0) {
```
bodinsamuel: nitpicking: early return in general makes this kind of code more readable

packages/orchestrator/lib/clients/processor.ts
```diff
@@ -0,0 +1,143 @@
...
+                        this.abortControllers.delete(task.id);
+                    }
+                }
+            }
+        }
+        return;
+    }
+
+    private async processingLoop() {
+        while (!this.stopped) {
```
bodinsamuel, line 73: there is no cpu nice here? don't you fear it will just eat the whole cpu?

packages/orchestrator/lib/clients/validate.ts
```diff
@@ -0,0 +1,53 @@
...
+const actionSchema = z.object({
+    ...commonSchemaFields,
+    payload: actionArgsSchema
+});
+const webhookSchema = z.object({
+    ...commonSchemaFields,
+    payload: webhookArgsSchema
+});
+
+export function validateTask(task: Task): Result<TaskAction | TaskWebhook> {
```
bodinsamuel, line 24: I feel like `Task{ type: string }` would have been more scalable, because we'll need to add syncs and then maybe more stuff along the way

packages/orchestrator/lib/routes/health.ts
bodinsamuel: I tried to keep one per file by prefixing the filename, i.e: `getHealth.ts` would make sense to me and keep files small

packages/orchestrator/lib/clients/processor.ts
```diff
@@ -0,0 +1,143 @@
...
+
+    constructor(maxConcurrency: number) {
+        this.maxConcurrency = maxConcurrency;
+        this.queue = [];
+    }
+
+    private async acquire(): Promise<void> {
+        if (this.queue.length < this.maxConcurrency) {
+            return;
+        }
```
bodinsamuel, line 119: are you sure about that? you are dequeueing 2 times the max task just to discard most of them

packages/jobs/lib/processor/process.ts
```diff
@@ -0,0 +1,34 @@
+import type { OrchestratorTask, TaskWebhook, TaskAction } from '@nangohq/nango-orchestrator';
+import type { JsonValue } from 'type-fest';
+import { Err, Ok } from '@nangohq/utils';
+import type { Result } from '@nangohq/utils';
+
+export async function process(task: OrchestratorTask): Promise<Result<JsonValue>> {
```
bodinsamuel: I would avoid this reserved keyword, (`process.env` etc)

packages/orchestrator/lib/clients/processor.ts
```diff
@@ -0,0 +1,143 @@
...
+                }
+                for (const task of tasks.value) {
+                    await this.processTask(task);
+                }
+            }
+        }
+        return;
+    }
+
+    private async processTask(task: OrchestratorTask): Promise<void> {
```
bodinsamuel, line 89:
```suggestion
    private processTask(task: OrchestratorTask): Promise<void> {
```

packages/jobs/lib/processor/processor.ts
```diff
@@ -0,0 +1,50 @@
+import { getLogger } from '@nangohq/utils';
```
bodinsamuel, line 1: if you are looking for a name that is not a variant of process, here it looks a cluster of workers not really a processor

packages/jobs/lib/processor/processor.worker.boot.ts
```diff
@@ -0,0 +1,11 @@
+import { isMainThread, parentPort, workerData } from 'node:worker_threads';
+import { getLogger } from '@nangohq/utils';
+import { ProcessorChild } from './processor.worker.js';
+
+const logger = getLogger('Scheduler.monitor');
```
bodinsamuel:
```suggestion
const logger = getLogger('something.else');
```

bodinsamuel: (also not sure what is the overhead of creating one child logger per file but we'll see)

packages/orchestrator/lib/routes/v1/tasks/taskId/heartbeat.ts
```diff
@@ -0,0 +1,45 @@
...
+    };
+    Error: ApiError<'post_heartbeat_failed'>;
+    Success: never;
+}>;
+
+const validate = validateRequest<PostHeartbeat>({
+    parseParams: (data) => z.object({ taskId: z.string().uuid() }).parse(data)
+});
+
+const putHandler = (scheduler: Scheduler) => {
```
bodinsamuel: why put and post at the same time?

packages/scheduler/lib/models/tasks.ts
```diff
@@ -133,16 +132,19 @@ export async function get(db: knex.Knex, taskId: string): Promise<Result<Task>>
...
+    }
     if (params?.groupKey) {
         query.where('group_key', params.groupKey);
     }
     if (params?.state) {
         query.where('state', params.state);
     }
     const limit = params?.limit || 100;
-    const tasks = await query.limit(limit);
+    const tasks = await query.limit(limit).orderBy('id');
```
bodinsamuel, line 147: because we just talked about it, I feel like it's probably missing
- groupKey + state 
- groupKey (where state=created)
- groupKey (where state=started)

packages/orchestrator/lib/server.ts
```diff
@@ -33,9 +37,13 @@ export const getServer = (scheduler: Scheduler, eventEmmiter: EventEmitter): Exp
...
-    createRoute(server, healthHandler);
-    createRoute(server, scheduleHandler(scheduler));
-    createRoute(server, outputHandler(scheduler, eventEmmiter));
+    createRoute(server, getHealthHandler);
+    createRoute(server, postScheduleHandler(scheduler));
+    createRoute(server, postSearchHandler(scheduler));
+    createRoute(server, putTaskHandler(scheduler));
+    createRoute(server, getOutputHandler(scheduler, eventEmmiter));
+    createRoute(server, postHeartbeatHandler(scheduler));
+    createRoute(server, getDequeueHandler(scheduler, eventEmmiter));
```
bodinsamuel: I would be careful about making it a GET, because it's not idempotent

packages/orchestrator/lib/routes/v1/search.ts
```diff
@@ -0,0 +1,56 @@
...
+        z
+            .object({
+                groupKey: z.string().min(1).optional(),
+                limit: z.coerce.number().positive().optional(),
+                ids: z.array(z.string().uuid()).optional()
+            })
+            .parse(data)
+});
+
+const getHandler = (scheduler: Scheduler) => {
```
bodinsamuel: postHandler maybe no?

packages/orchestrator/lib/routes/v1/dequeue.ts
```diff
@@ -0,0 +1,90 @@
...
+                cleanupAndRespond((res) => res.status(500).json({ error: { code: 'dequeue_failed', message: getTasks.error.message } }));
+            } else {
+                cleanupAndRespond((res) => res.status(200).json(getTasks.value));
+            }
+        };
+        const timeout = setTimeout(() => {
+            cleanupAndRespond((res) => res.status(200).send([]));
+        }, waitForCompletionTimeoutMs);
+
+        eventEmitter.once(eventId, onTaskStarted);
```
bodinsamuel: not sure I get the wait for completion in that case, because if don't receive an answer you'll just wait 60s for nothing

packages/orchestrator/lib/clients/processor.ts
```diff
@@ -0,0 +1,143 @@
...
+        }
+        return;
+    }
+
+    private async processingLoop() {
+        while (!this.stopped) {
+            if (this.queue.available() > 0) {
+                const tasks = await this.orchestratorClient.dequeue({ groupKey: this.groupKey, limit: this.queue.available() * 2 }); // fetch more than available to keep the queue full
+                if (tasks.isErr()) {
+                    return Err(tasks.error);
```
bodinsamuel: this kills the loop forever and the error is unchecked, is it on purpose?

packages/jobs/lib/app.ts
```diff
@@ -1,25 +1,41 @@
...
-    server.listen(JOBS_PORT);
-    logger.info(`🚀 service ready at http://localhost:${JOBS_PORT}`);
-    const temporalNs = process.env['TEMPORAL_NAMESPACE'] || 'default';
+    const port = envs.NANGO_JOBS_PORT;
+    const temporalNs = envs.TEMPORAL_NAMESPACE;
+    const orchestratorServiceUrl = envs.ORCHESTRATOR_SERVICE_URL;
+    server.listen(port);
+    logger.info(`🚀 service ready at http://localhost:${port}`);
     const temporal = new Temporal(temporalNs);
+    const processor = new Processor(orchestratorServiceUrl);
```
bodinsamuel, line 22: nodemon should restart job when orchestrator is modified

packages/jobs/lib/env.ts
```diff
@@ -0,0 +1,3 @@
+import { ENVS, parseEnvs } from '@nangohq/utils';
+
+export const envs = parseEnvs(ENVS.required({ ORCHESTRATOR_SERVICE_URL: true }));
```
bodinsamuel, line 3: should be in .env.example (at least commented) otherwise new dev install won't work

bodinsamuel (commented):
Very nice progress! That's a lot of code and complexity to review, hopefully we'll get used to it quickly 💯

packages/orchestrator/lib/clients/processor.ts
bodinsamuel: nevermind it's long polling. 
it's not clear when reading the code

packages/orchestrator/lib/routes/health.ts
TBonnin: sensible. I will do the same then

packages/orchestrator/lib/clients/processor.ts
TBonnin: let me remove the default for waitForCompletion param. That will make thing slightly more clear

packages/orchestrator/lib/clients/validate.ts
TBonnin: they don't have the same fields though. Having some type guard functionality is nice though even though I agree that the definitions are more complex

packages/orchestrator/lib/clients/processor.ts
TBonnin: they are not discarded they are added to the queue

packages/jobs/lib/processor/process.ts
TBonnin: good point. will change

packages/jobs/lib/processor/processor.ts
TBonnin: I was trying to avoid `worker` 😀 which also has a lot of meanings

packages/jobs/lib/processor/processor.worker.boot.ts
TBonnin: I have been doing it by lazyness but I agree that is probably not optimal to have one child logger per file

packages/scheduler/lib/models/tasks.ts
TBonnin: yes. I will add missing index in next PR

packages/orchestrator/lib/server.ts
TBonnin: good point I will make it a POST endpoint

packages/orchestrator/lib/routes/v1/dequeue.ts
TBonnin: you are waiting for a task to be created/dequeuable, so you don't have to keep making request. Trying hard not to have a push based system that would requires handling states but in the scenario of a processor it could be an acceptable solution.

packages/orchestrator/lib/clients/processor.ts
TBonnin: good catch. I didn't intended to get out of the loop but just log the error.

packages/jobs/lib/processor/process.ts
TBonnin: renamed it to handler. which is also quite generic but at least it is not a reserved keywork. Any other suggestion is welcome 😁

packages/orchestrator/lib/clients/processor.ts
bodinsamuel: I think waitForMessage might be more clear (but I get that you named it consistently with the others so not pushing)

packages/jobs/lib/processor/process.ts
bodinsamuel: handle is fine 👍🏻

packages/orchestrator/lib/clients/processor.ts
TBonnin: you are right the name is not great. what about simply `longPolling=true` or `longPoll=true` and keeping it consistent between endpoints

packages/orchestrator/lib/routes/v1/dequeue.ts
bodinsamuel: okay took me a while to understand the executeAction -> execute -> schedule -> immediate -> event 
that works 👍🏻

packages/orchestrator/lib/server.ts
bodinsamuel: it's still named getDequeue 🤭

packages/orchestrator/lib/clients/validate.ts
bodinsamuel: I mean you can still validate the payload but it's just a waste of time to not use a discriminator 👍🏻

packages/orchestrator/lib/clients/processor.ts
bodinsamuel: Ah yeah I miss read the condition my bad 👍🏻

packages/orchestrator/lib/clients/processor.ts
```diff
@@ -0,0 +1,145 @@
...
+        }
+        return;
+    }
+
+    private async processingLoop() {
+        while (!this.stopped) {
+            if (this.queue.available() > 0) {
+                const tasks = await this.orchestratorClient.dequeue({ groupKey: this.groupKey, limit: this.queue.available() * 2, waitForCompletion: true }); // fetch more than available to keep the queue full
+                if (tasks.isErr()) {
+                    logger.error(`failed to dequeue tasks: ${stringifyError(tasks.error)}`);
```
bodinsamuel, line 77: just a heads up, happened to me once when the API stop responding all fetch are failing fast ultimately using the whole cpu. Might be an edge case but fyi

bodinsamuel (approved):
🛫

packages/orchestrator/lib/clients/processor.ts
TBonnin: good point. introducing a little delay in case of error

packages/orchestrator/lib/server.ts
TBonnin: 🤦 (I am using this emoji a lot today)

packages/orchestrator/lib/clients/processor.ts
TBonnin: I have merged so I am not blocked but I will change it to whatever we decide

PR #2222, opened 2024-05-29
feat(server): [nan-981] implement error reporting UI on connection authorization sub notification

khaliqgant:
## Describe your changes
- Adds error notifications to the List connections page with the ability to filter the connections that have an error or not
- Adds an icon on the Connections sidebar to indicate that there is any connection with an error
- Adds a banner for a sync if it has a recent error 
- Adds a banner for the authorization tab if it has a connection refresh error
- removes some of the needless server side formatting logic so that the connection object is consistent across server and client 

Note that I added types to `nangohq/types` so that the new route could be used. I'll swap out types in another PR to not make this one too bloated.

NAN-981


packages/shared/lib/db/database.ts
```diff
@@ -2,32 +2,7 @@ import knex from 'knex';
 import type { Knex } from 'knex';
 import { metrics, retry } from '@nangohq/utils';
 import type { Pool } from 'tarn';
-
```
khaliqgant: This was moved out because currently trying to create a migration file fails using the cli 

```
Invalid or unexpected token
/Users/khaliqgant/Projects/nango/short-term-feature/packages/utils/dist/environment/detection.js:7
export const isCloud = process.env['NANGO_CLOUD']?.toLowerCase() === 'true';
                                                  ^

SyntaxError: Invalid or unexpected token
    at require.extensions..js (/Users/khaliqgant/Projects/nango/short-term-feature/node_modules/knex/bin/cli.js:465:18)
```

due to the usage of `utils` in the database file which gets called by the knexfile.cjs file

packages/utils/lib/result.ts
```diff
@@ -2,31 +2,32 @@
 By convention Left represents a failed computation
 And Right represents a successful one
 */
-export interface Left<T, E extends Error> {
+export interface Left<T, E extends Error, P = unknown> {
```
TBonnin: are you trying to add a payload for the Error case? 
If yes, it is already supported since `E extends Error`. 

Just define your ErrorType that extends Error. Have a look at the orchestrator `ClientError` for instance https://github.com/NangoHQ/nango/blob/5121560fcf4b9b71d524e3715520bd7b961f8f82/packages/orchestrator/lib/client.ts#L49-L52

Then you can have a `Result<SuccessType, YourErrorTypeWithPayload>`, without having to modify `Result`

packages/shared/lib/db/migrations/20240529180658_add_notifications_table_index.cjs
```diff
@@ -0,0 +1,9 @@
+exports.config = { transaction: false };
+
+exports.up = async function (knex, _) {
+    await knex.schema.raw('CREATE INDEX CONCURRENTLY "idx_sync_id_active_true" ON "_nango_ui_notifications" USING BTREE ("sync_id") WHERE active = true');
```
khaliqgant: Didn’t want to over optimize but this is added with the query in the sync.service in mind

packages/shared/lib/models/Connection.ts
```diff
@@ -86,4 +86,5 @@ export interface ConnectionList {
     provider: string;
     created: string;
     metadata?: Metadata | null;
+    error_log_id?: number | string | null;
```
khaliqgant: Number or string so that it can work with the old kids and new logs while we are in transition

packages/shared/lib/models/Sync.ts
```diff
@@ -66,6 +66,7 @@ export interface ReportedSyncJobStatus {
     frequency: string;
     finishedAt: Date;
     nextScheduledSyncAt: Date | null;
+    latestExecutionStatus: SyncStatus;
```
khaliqgant, line 69: Client requested update

TBonnin (commented):
I haven't managed to finish reviewing this PR. I will continue first thing tomorrow morning. Sorry about that.

packages/utils/lib/result.ts
khaliqgant: Thanks, updated with https://github.com/NangoHQ/nango/pull/2222/commits/759d73d0588f7c56dfa615d82d1f4b5e9eccdc70 but feels a little less clean

packages/shared/lib/services/sync/sync.service.ts
```diff
@@ -249,12 +240,14 @@ export const getSyncs = async (nangoConnection: Connection): Promise<(Sync & { s
...
                 `
-            ),
-            syncJobTimestampsSubQuery
+            )
         )
         .join(SYNC_SCHEDULE_TABLE, function () {
             this.on(`${SYNC_SCHEDULE_TABLE}.sync_id`, `${TABLE}.id`).andOn(`${SYNC_SCHEDULE_TABLE}.deleted`, '=', db.knex.raw('FALSE'));
         })
+        .leftJoin('_nango_ui_notifications', function () {
+            this.on(`_nango_ui_notifications.sync_id`, `${TABLE}.id`).andOn(`_nango_ui_notifications.active`, '=', db.knex.raw('TRUE'));
```
bodinsamuel: I recently discovered `andOnVal` that solves the issue of passing a value in the column arg

packages/types/lib/account/db.ts
```diff
@@ -0,0 +1,10 @@
+export interface Account {
```
bodinsamuel, line 1: not suggesting anything (at least not for this PR) 
I tried to prefix database type with DB to avoid any confusion, not sure if it's the best but I feel like it working to disambiguate, wdyt?

packages/shared/lib/services/sync/sync.service.ts
```diff
@@ -249,12 +240,14 @@ export const getSyncs = async (nangoConnection: Connection): Promise<(Sync & { s
...
                 ) as latest_sync
                 `
-            ),
-            syncJobTimestampsSubQuery
+            )
         )
         .join(SYNC_SCHEDULE_TABLE, function () {
             this.on(`${SYNC_SCHEDULE_TABLE}.sync_id`, `${TABLE}.id`).andOn(`${SYNC_SCHEDULE_TABLE}.deleted`, '=', db.knex.raw('FALSE'));
         })
+        .leftJoin('_nango_ui_notifications', function () {
```
bodinsamuel: if you have multiples active notifications it will mess up this query. I would do it in an other query just for simplicity sake

packages/shared/lib/services/sync/run.service.ts
```diff
@@ -735,7 +743,7 @@ export default class SyncRun {
 
         if (!this.isWebhook) {
             try {
-                await this.slackNotificationService.reportFailure(
+                void this.slackNotificationService.reportFailure(
```
bodinsamuel: if you don't await the try/catch will not work and crash the service (unless I'm mistaken)

packages/server/lib/routes.ts
```diff
@@ -44,8 +45,8 @@ import {
     getEmailByExpiredToken
 } from './controllers/v1/account/index.js';
 import { searchMessages } from './controllers/v1/logs/searchMessages.js';
-import { setMetadata } from './controllers/v1/connection/setMetadata.js';
-import { updateMetadata } from './controllers/v1/connection/updateMetadata.js';
+import { setMetadata } from './controllers/v1/connection/api/setMetadata.js';
```
bodinsamuel: I didn't realized when I reviewed it the other day, but the path of the file should (ideally) reflect the path on the API. However setMetadata is not in `/v1/connection` but in `/connection`. 
Not suggesting to change but might clarify the expectation and remove the need to use `/web`

packages/shared/lib/services/notification/error.service.ts
```diff
@@ -0,0 +1,80 @@
...
+            if (created?.[0]) {
+                return Ok(created[0]);
+            } else {
+                return Err('Failed to create notification');
+            }
+        },
+        get: async (id: number): Promise<UINotification | null> => {
+            return await db.knex.from<UINotification>(DB_TABLE).where({ type: 'auth', connection_id: id, active: true }).first();
+        },
+        invalidate: async ({ connection_id }: Required<Pick<AuthNotification, 'connection_id'>>): Promise<void> => {
```
bodinsamuel: I feel like clear/delete (or similar would be clearer) I thought it was invalidating the auth at first

packages/shared/lib/services/notification/error.service.ts
```diff
@@ -0,0 +1,80 @@
...
+            await db.knex.from<UINotification>(DB_TABLE).where({ type: 'auth', connection_id }).update({ active: false });
```
bodinsamuel: any reason to keep those rows instead of deleting?

packages/shared/lib/services/notification/error.service.ts
```diff
@@ -0,0 +1,80 @@
...
+
+interface SyncNotification extends AuthNotification {
+    sync_id: string;
+}
+
+export const errorNotificationService = {
+    auth: {
+        create: async ({ type, action, connection_id, activity_log_id, log_id, active }: AuthNotification): Promise<Result<UINotification>> => {
+            await errorNotificationService.auth.invalidate({ connection_id });
+            const created = await db.knex
```
bodinsamuel: should be done in transaction to avoid race condition (an other process doing the same thing and you end up with two active rows)

packages/shared/lib/db/migrations/20240527094039_add_notifications_table.cjs
```diff
@@ -0,0 +1,22 @@
+const DB_TABLE = '_nango_ui_notifications';
```
bodinsamuel: opinion: not sure I prefer consistency over gradually removing prefix, wdyt?

Also I would be careful about naming that, it's not really notifications but error log and it might not be tied to the UI for long

packages/server/lib/controllers/v1/connection/web/get.ts
```diff
@@ -0,0 +1,174 @@
...
+                    });
+                    break;
+            }
+        }
+        return;
+    }
+
+    const { value: connection } = credentialResponse;
+
+    const config: IntegrationConfig | null = await configService.getProviderConfig(connection.provider_config_key, environment.id);
```
bodinsamuel: not for this PR: this call is already done in getCredentials we should probably note where calls are duplicated to improve when we have the time

packages/server/lib/controllers/connection.controller.ts
```diff
@@ -31,193 +28,9 @@ import type { RequestLocals } from '../utils/express.js';
...
-                    credentials,
-                    rawCredentials
-                }
-            });
-        } catch (err) {
-            next(err);
-        }
-    }
-
-    async getConnectionsWeb(_: Request, res: Response<any, Required<RequestLocals>>, next: NextFunction) {
```
bodinsamuel, line 182: it wasn't used?

packages/webapp/src/components/ui/Info.tsx
```diff
@@ -13,10 +13,10 @@ interface InfoProps {
```
bodinsamuel, line 13: looking at the figma the horizontal padding seems incorrect

bodinsamuel: Screenshot 2024-05-30 at 12 58 58

packages/webapp/src/pages/Connection/Authorization.tsx
```diff
@@ -4,14 +4,14 @@ import { Loading } from '@geist-ui/core';
```
bodinsamuel, line 4: When I got in this tab with an error I get `Cannot read properties of undefined (reading 'type')` 
edit: might be due to a 500 actually (I just broke the credentials to create an error :D)

packages/webapp/src/pages/Connection/List.tsx
```diff
@@ -133,45 +153,64 @@ export default function ConnectionList() {
...
+                                label="Integrations"
+                                options={providers.map((integration: string) => {
+                                    return { name: integration, value: integration };
+                                })}
+                                selected={selectedIntegration}
+                                defaultSelect={defaultFilter}
+                                onChange={handleIntegrationChange}
+                                all
+                            />
+                            <MultiSelect
```
bodinsamuel, line 185: Nice that you could reuse that without change 👌🏻

packages/webapp/src/pages/Connection/Show.tsx
bodinsamuel, line 1: Note:
Because `GET /connection/:id` has this side effects, it spams the logs quite a lot. Especially in dev I get 3 calls per page (1 normal, 1 extra because of react strict mode, 1 I don't get) so it quickly adds up. 
I understand the appeal of this side effects but I don't think it's a good idea

bodinsamuel (commented):
Works well, I have some open comments 👏🏻

packages/shared/lib/services/sync/sync.service.ts
khaliqgant: Ah yes, I saw that and meant to use that here but forgot, thanks

packages/types/lib/account/db.ts
khaliqgant: Yeah, that's a good idea.

packages/shared/lib/services/connection.service.ts
```diff
@@ -658,16 +673,31 @@ class ConnectionService {
...
+                if (activityLogId) {
+                    await errorNotificationService.auth.create({
+                        type: 'auth',
+                        action: 'token_refresh',
+                        connection_id: connection.id,
+                        activity_log_id: activityLogId,
+                        log_id: logCtx.id,
+                        active: true
+                    });
+                }
```
TBonnin, line 705: It looks like we are now duplicating some activity logs so they can be used in a different way. 
I understand we cannot used activity logs directly because of the need for a longer retention period but instead of inventing a new model/logic would it make sense to implement a mechanism to persist some of `logCtx` data to the db?
So instead of duplicating the object being saved we call something like `persistLongTerm(logCtx)`. This single function would then transform the data to whatever format we want and save it to the db. That could  greatly improve the reusability and prevent the need to create new function every time we want to add another error type. (I would also not be surprised if in a near future we also use the same mechanism to surface warning or important info)

Feel free to ping me if you want to discuss about this topic.

TBonnin (commented):
Happy to discuss about the comment I left. I think it is worth spending some time about how this feature is going to be extended in the future

packages/shared/lib/services/connection.service.ts
bodinsamuel: We discussed this with Thomas, I agree we need some kind of way to limit the number of things doing similar stuff (same problem with error reporting in general). Ideally I think we would have an event stream where we subscribe to what's interesting to each system, in the meantime I'm fine with the current state. 

Note: I didn't realized during my review but the proposed system (storing the activityLogId) is not playing well with the retention of 15days (old and new logs system), so it might indeed need some change unless we are fine with it.

packages/shared/lib/db/migrations/20240527094039_add_notifications_table.cjs
khaliqgant: Yeah, I also prefer to be consistent and see no reason to remove the prefix

packages/webapp/src/pages/Connection/Show.tsx
khaliqgant: Thanks for raising. Spoke to Bastien about this and we'll pare it back if needed later

packages/webapp/src/pages/Connection/List.tsx
khaliqgant: Ideally I wanted to use the multi select with search but that was very much tied to the ES implementation so left it alone

packages/webapp/src/components/ui/Info.tsx
khaliqgant: Thanks, updated.

packages/server/lib/controllers/connection.controller.ts
khaliqgant: Moved to the new route structure

packages/shared/lib/services/notification/error.service.ts
khaliqgant: Updated to be `clear`

packages/shared/lib/services/notification/error.service.ts
khaliqgant: For historical purposes, but updated it to `delete` now

packages/server/lib/routes.ts
khaliqgant: Updated

packages/shared/lib/services/sync/run.service.ts
khaliqgant: Removed the try catch

packages/shared/lib/services/sync/sync.service.ts
khaliqgant: Updated to json_build_object or null if empty

packages/shared/lib/services/connection.service.ts
khaliqgant: So we're not duplicating logs but rather linking to them. So a reference to the log is stored but the error isn't duplicated anywhere. The 15 day retention shouldn't be an issue for auth refreshes since stale tokens refresh every day so there should always be a fresh log.

If a sync doesn't run for 15 days it is dormant anyway so don't see an issue there really. Also, if a log is deleted with cascading the entry in the active log table is also deleted.

packages/webapp/src/pages/Connection/List.tsx
bodinsamuel: yeah search is rarely portable, unless you do everything in the browser or you had a function parameter but then the component is much more complicated to use

packages/shared/lib/db/migrations/20240527094039_add_notifications_table.cjs
```diff
@@ -0,0 +1,22 @@
...
+
+exports.up = async function (knex, _) {
+    return knex.schema.createTable(DB_TABLE, function (table) {
+        table.increments('id').primary();
+        table.string('type', 'varchar(255)').notNullable();
+        table.string('action', 'varchar(255)').notNullable();
+        table.integer('connection_id').unsigned().notNullable();
+        table.foreign('connection_id').references('id').inTable('_nango_connections').onDelete('CASCADE');
+        table.integer('activity_log_id').unsigned().notNullable();
+        table.foreign('activity_log_id').references('id').inTable('_nango_activity_logs').onDelete('CASCADE');
```
TBonnin: would that require a change of the schema when we stop adding activity logs to the db and we cannot have a foreign key anymore linking to the activity_logs table?

packages/shared/lib/services/connection.service.ts
```diff
@@ -523,14 +525,47 @@ class ConnectionService {
...
+                '_nango_connections.metadata',
+                db.knex.raw(`
+                    CASE
+                        WHEN COUNT(${ACTIVE_LOG_TABLE}.activity_log_id) = 0 THEN NULL
+                        ELSE json_build_object(
+                            'activity_log_id', ${ACTIVE_LOG_TABLE}.activity_log_id,
+                            'log_id', ${ACTIVE_LOG_TABLE}.log_id
+                        )
+                    END as active_logs
+                `)
```
TBonnin, line 550: are we returning the active logs data as part of any customer  facing API endpoints?

packages/shared/lib/services/notification/error.service.ts
```diff
@@ -0,0 +1,91 @@
...
+import type { Result } from '@nangohq/utils';
+
+import db from '../../db/database.js';
+
+const DB_TABLE = '_nango_active_logs';
+
+type ErrorNotification = Required<Pick<ActiveLog, 'type' | 'action' | 'connection_id' | 'activity_log_id' | 'log_id' | 'active'>>;
+type SyncErrorNotification = ErrorNotification & Required<Pick<ActiveLog, 'sync_id'>>;
+
+export const errorNotificationService = {
```
TBonnin, line 13: should all the variables/classes in this file be named after `activeLogs` because this is basically a model interacting with the db to save/retrieve/delete `activeLogs`

TBonnin (approved):
I am not gonna block the PR but  think it would have been worthwhile coming up with a generic way to store any activity logs into an `active_logs` in postgres to avoid having to extend the `error.service` in the future as well as having a more future proof data schema that doesn't require alteration if we need to add extra fields.

I would also like to note that we went to a enormous level of effort to get the logs out of postgres to now bringing back a subset of them into postgres

packages/shared/lib/services/connection.service.ts
khaliqgant: No, it is only internal

packages/shared/lib/services/notification/error.service.ts
khaliqgant: Yeah, I had the same thought, but I concluded with is that `activeLogs` are actually pretty flexible and this service is one kind of usage of them. In theory we could have another class that uses them differently.

packages/shared/lib/db/migrations/20240527094039_add_notifications_table.cjs
khaliqgant: Good point updated to be nullable with https://github.com/NangoHQ/nango/pull/2222/commits/3f44cbdd64d302a2ac9228c775119946ae93ed48

khaliqgant:
> I am not gonna block the PR but think it would have been worthwhile coming up with a generic way to store any activity logs into an `active_logs` in postgres to avoid having to extend the `error.service` in the future as well as having a more future proof data schema that doesn't require alteration if we need to add extra fields.
> 
> I would also like to note that we went to a enormous level of effort to get the logs out of postgres to now bringing back a subset of them into postgres

Valid feedback and noted. Thanks.

packages/webapp/src/pages/Connection/Authorization.tsx
khaliqgant: Added fallback for type if it is missing which in `unauthenticated` provider it is

packages/shared/lib/services/notification/error.service.ts
```diff
@@ -0,0 +1,91 @@
...
+        get: async (id: number): Promise<ActiveLog | null> => {
+            return await db.knex.from<ActiveLog>(DB_TABLE).where({ type: 'auth', connection_id: id, active: true }).first();
+        },
+        clear: async ({ connection_id, trx }: { connection_id: ActiveLog['connection_id']; trx?: Knex.Transaction }): Promise<void> => {
+            if (trx) {
+                await trx.from<ActiveLog>(DB_TABLE).where({ type: 'auth', connection_id }).delete();
+            } else {
+                await db.knex.from<ActiveLog>(DB_TABLE).where({ type: 'auth', connection_id }).delete();
+            }
+        }
```
bodinsamuel: note that you can swap them transparently, with just a little bit of type help 
```suggestion
        clear: async ({ connection_id, trx = db.knex }: { connection_id: ActiveLog['connection_id']; trx?: Knex.Transaction | Knex }): Promise<void> => {
            await trx.from<ActiveLog>(DB_TABLE).where({ type: 'auth', connection_id }).delete();
        }
```

bodinsamuel (approved):
🤺

packages/shared/lib/services/notification/error.service.ts
khaliqgant: Thanks, updated with https://github.com/NangoHQ/nango/pull/2222/commits/e92debc5f97efd7a3dc7b4e85259f9da3bbc9052

PR #2224, opened 2024-05-30
fix(logs): first prod feedback

bodinsamuel:
## Changes

Directly related to implem:
- UI: bad loading state when opening the operation drawer
- UI: search in operation's logs was not working after latest PR 
- ES: if we don't `refresh=true` index after inserting operation it can sometimes be missing from the index when we reach jobs making it impossible to update state, unfortunately that will slow down insert of operation not sure by how much

Other related changes:
- `syncRunService` clarify what we expect in terms of logging options (with types) when dryRun or not
We either `writeToDB` and needs everything related to logs, or nothing. 

- `syncRunService` now takes a mandatory logCtx (if writeToDb=true)
In `activities.ts` we create an operation for `sync run`, but when we entered this service (on the same machine) we had to get the operation back with the activityLogId which was definitely unnecessary and slowing down everything. 

- `triggerAction()` never takes an optional activityLogId and `writeLogs` at all
It helps simplify the code and ensure we have the data in activities.ts. 
I could not find a place where we don't pass an activityLogId and use this writeLogs=false. I saw it was introduced in an old [PR](https://github.com/NangoHQ/nango/pull/1202) let me know @khaliqgant


packages/jobs/lib/activities.ts
```diff
@@ -505,41 +508,17 @@ export async function reportFailure(
...
-            debug: Boolean(debug)
-        });
-
         if ('syncJobId' in workflowArguments) {
             await updateSyncJobStatus(workflowArguments.syncJobId, SyncStatus.STOPPED);
         } else {
             await updateLatestJobSyncStatus(workflowArguments.syncId, SyncStatus.STOPPED);
         }
 
-        await syncRun.cancel();
```
bodinsamuel, line 542: I removed the instantiation of syncRunService here because we were not using any of the required parameters, seems like unnecessary overhead. 
Let met know if that's alright

packages/jobs/lib/activities.ts
```diff
@@ -98,6 +98,7 @@ export async function runAction(args: ActionArgs): Promise<ServiceResponse> {
         orchestratorClient,
         logContextGetter,
         writeToDb: true,
+        logCtx: await logContextGetter.get({ id: String(activityLogId) }),
```
TBonnin, line 105: do you still need the logContextGetter above then?

packages/jobs/lib/activities.ts
TBonnin: looks correct. That's the only place where run.cancel() is being called if I am not mistaken

TBonnin (approved):
a small comment about the `logContextGetter` in run.services but lgtm overall

packages/jobs/lib/activities.ts
bodinsamuel: yes because we still create other operation for webhook and slack if I remember well, definitely not related but until better it's there :p

TBonnin: indeed the logContextGetter is passed to the `slackNotificationService` function. Since the `slackNotificationService` is also passed as a param can we give the `logContextGetter` to the `slackNotificationService` constructor and avoid having to pass it down to each of its methods. That I think would remove the need to pass it to the run service.

bodinsamuel: SlackService is not passed as a param to runService. But I realised orchestrator was also passed just for slack, so I tried a bigger change, let me know: 1d41fd4e136f18b484a86c36f6ffb70c51a8a491

TBonnin: I love it. I should have done that instead of injecting the orchestratorClient

khaliqgant (approved):

PR #2226, opened 2024-05-30
chore(persist): make NANGO_ENCRYPTION_KEY required in persist service

TBonnin:
Instead of failing when persisting records, persist will fail to start if the encryption key is not setup (with a error message mentioning the doc url)
Also adding to the docs how to generate a encryption key
Screenshot 2024-05-30 at 10 23 52

https://linear.app/nango/issue/NAN-922/improve-error-when-no-encryption-key

### DIFF
--- a/docs-v2/host/self-host/self-hosting-instructions.mdx
+++ b/docs-v2/host/self-host/self-hosting-instructions.mdx
@@ -74,11 +74,16 @@ NANGO_DASHBOARD_PASSWORD=<PICK-A-PASSWORD>
 ### Encrypt sensitive data[](#encrypt-sensitive-data 'Direct link to Encrypt sensitive data')
 
 You can enforce encryption of sensitive data (tokens, secret key, app secret)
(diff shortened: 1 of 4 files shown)


bodinsamuel (approved):
Ah yes I was meaning to document that missing step too.
Can you add the same openssl comment in the .env.example?
Ideally we could also uncomment it even if not strictly mandatory, what do you think?

TBonnin:
> Ah yes I was meaning to document that missing step too. Can you add the same openssl comment in the .env.example? Ideally we could also uncomment it even if not strictly mandatory, what do you think?

added the command in `.env.example`. Keeping `NANGO_ENCRYPTION_KEY` commented out though since technically it is not required to run the entire stack. Happy to revisit though

bodinsamuel:
> added the command in .env.example.

Forgot to push 🫡

TBonnin:
> > added the command in .env.example.
> 
> Forgot to push 🫡

🤦 
https://github.com/NangoHQ/nango/pull/2227

PR #2228, opened 2024-05-30
fix(server.proxy): [nan-1049] if the response not chunked manually piece the response together

khaliqgant:
## Describe your changes
Issue reported in the community. The proxy assumed that providers send back the data chunked but that is not the case. When it is not chunked we need to manually piece together the chunks and parse the string data.

NAN-1049

### DIFF
--- a/packages/server/lib/controllers/proxy.controller.ts
+++ b/packages/server/lib/controllers/proxy.controller.ts
@@ -33,12 +33,14 @@ import {
     connectionService,
     configService
 } from '@nangohq/shared';
-import { metrics } from '@nangohq/utils';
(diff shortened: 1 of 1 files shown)


TBonnin (approved):
a bit annoying that we have to deal with provider malfunctioning APIs, sending back chunked response without setting the `transfer-encoding: chunked` header

khaliqgant:
> a bit annoying that we have to deal with provider malfunctioning APIs, sending back chunked response without setting the `transfer-encoding: chunked` header

Yeah and it is only for certain api endpoints as well 😩

AnandChowdhary:
Sorry you had to deal with this weird inconsistency and thank you so much for your hard work, much appreciated!

PR #2229, opened 2024-05-30
fix(server.proxy): [nan-1051] allow DELETE to have a body

khaliqgant:
## Describe your changes
Allow a delete call to have a body

NAN-1051


TBonnin (commented):
I am curious, which API requires a DELETE with a body?

TBonnin (approved):

khaliqgant:
> I am curious, which API requires a DELETE with a body?

https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-webhooks/#api-rest-api-3-webhook-delete

PR #2232, opened 2024-05-31
fix(docker): missing orchestrator env

bodinsamuel:
## Describe your changes

fix the ci


khaliqgant (approved):
I think at some point we should actually remove this as I don't think anyone runs the app via the cli anymore

TBonnin (approved):

PR #2235, opened 2024-05-31
feat(pkgs): create kvstore 

bodinsamuel:
## Describe your changes

Fixes NAN-1071

- Create a reusable package for KVStore
Duplicated for now, can't remove it entirely from `shared` because FeatureFlags and Locking, but it's a first step. I stopped exporting it so it's no longer used by external package and I'll do a second pass to remove it entirely.
Note: I did that because logs can't depend on shared

- Use it in Jobs

- Use it in Logs
As discussed orally, the main id is to be able to retrieve the indexName quickly without doing a full search on the cluster. Maybe it's the fix maybe not.

### DIFF
--- a/packages/kvstore/lib/RedisStore.ts
+++ b/packages/kvstore/lib/RedisStore.ts
@@ -1,0 +1,47 @@
+import type { KVStore } from './KVStore.js';
+import { createClient } from 'redis';
+import type { RedisClientType } from 'redis';
+
+export class RedisKVStore implements KVStore {
+    private client: RedisClientType;
+
+    constructor(url: string) {
+        this.client = createClient({ url: url });
+
+        this.client.on('error', (err) => {
+            console.error(`Redis (kvstore) error: ${err}`);
+        });
+    }
+
+    public async connect(): Promise<void> {
+        return this.client.connect().then(() => {});
+    }
+
+    public async get(key: string): Promise<string | null> {
+        return this.client.get(key);
+    }
+
+    public async set(key: string, value: string, opts?: { canOverride?: boolean; ttlInMs?: number }): Promise<void> {
+        const options: any = {};
+        if (opts) {
+            if (opts.ttlInMs && opts.ttlInMs > 0) {
+                options['PX'] = opts.ttlInMs;
+            }
+            if (opts.canOverride === false) {
+                options['NX'] = true;
+            }
+        }
+        const res = await this.client.set(key, value, options);
+        if (res !== 'OK') {
+            throw new Error(`Failed to set key: ${key}, value: ${value}, ${JSON.stringify(options)}`);
+        }
+    }
+
+    public async exists(key: string): Promise<boolean> {
+        return (await this.client.exists(key)) > 0;
+    }
+
+    public async delete(key: string): Promise<void> {
+        await this.client.del(key);
+    }
+}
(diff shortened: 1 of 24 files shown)


packages/kvstore/lib/RedisStore.ts
```diff
@@ -0,0 +1,47 @@
...
+    public async set(key: string, value: string, opts?: { canOverride?: boolean; ttlInMs?: number }): Promise<void> {
+        const options: any = {};
+        if (opts) {
+            if (opts.ttlInMs && opts.ttlInMs > 0) {
+                options['PX'] = opts.ttlInMs;
+            }
+            if (opts.canOverride === false) {
+                options['NX'] = true;
+            }
+        }
```
bodinsamuel, line 33: Everything is copy pasted except the change in signature here (and the usage of the variable)

packages/logs/lib/models/messages.ts
```diff
@@ -161,10 +162,16 @@ export async function listOperations(opts: {
...
+export async function getOperation(opts: { id: MessageRow['id']; indexName?: string | null }): Promise<MessageRow> {
+    if (opts.indexName) {
+        const res = await client.get<OperationRow>({ index: opts.indexName, id: opts.id });
+        return res._source!;
+    }
+
     // Can't perform a getById because we don't know in which index the operation is in
     const res = await client.search<OperationRow>({
         index: indexMessages.index,
+        size: 1,
```
bodinsamuel, line 174: could help too

packages/utils/tsconfig.json
```diff
@@ -4,5 +4,10 @@
         "rootDir": "lib",
         "outDir": "dist"
     },
-    "include": ["lib/**/*"]
+    "include": ["lib/**/*"],
+    "references": [
+        {
+            "path": "../types"
+        }
```
bodinsamuel, line 11: sometimes I don't understand typescript, not sure how it has worked until now

khaliqgant (approved):

packages/logs/lib/models/logContextGetter.ts
```diff
@@ -54,7 +56,10 @@ export const logContextGetter = {
     async get({ id }: { id: MessageRow['id'] }, options?: Options): Promise<LogContext> {
         try {
             if (envs.NANGO_LOGS_ENABLED) {
-                const operation = await getOperation({ id });
+                const store = await getKVStore();
+                const indexName = await store.get(`es:operation:${id}:indexName`);
+
+                const operation = await getOperation({ id, indexName });
                 return new LogContext({ parentId: id, operation }, options);
```
TBonnin, line 63: do you know how much load is it going to add on redis?

TBonnin (approved):

packages/logs/lib/models/logContextGetter.ts
bodinsamuel: not really, the keys are short lived and with the current rate of actions it's going to be less than 200 keys per minutes

bodinsamuel: No impact on Redis (surprising actually)
Screenshot 2024-06-03 at 10 55 00


But almost halved CPU for ES 🙌🏻 
Screenshot 2024-06-03 at 10 54 41

khaliqgant: nice!

PR #2236, opened 2024-05-31
fix(database): [nan-1063] migrate to database package

khaliqgant:
## Describe your changes
As part of another ticket related to webhooks, I want to make a specific webhook package. I don't want the webhook package to depend on shared but shared contained the database as of now. This PR moves the database connection logic to its own package freeing it from shared and making it easier to use freely in other packages now. Note that the seeders have to stay due to the usage of services.

Contributes to NAN-1063


khaliqgant:
Integration tests are failing: need to update this https://github.com/NangoHQ/nango/blob/master/tests/setup.ts#L51

bodinsamuel (commented):
Nice 🚀 
I have a feeling it shouldn't work with the CLI, the sync run service is importing multiple files that requires this package.
edit: yeah the CI is failing :/ 

Side note (just saying not requesting a change): since we have started bundling services with their Database model, it would make sense this package to be inside `server`.

khaliqgant:
> Nice 🚀 I have a feeling it shouldn't work with the CLI, the sync run service is importing multiple files that requires this package. edit: yeah the CI is failing :/
> 
> Side note (just saying not requesting a change): since we have started bundling services with their Database model, it would make sense this package to be inside `server`.

Yeah. Updated and all checks passing. Sub bundling might make sense but don’t want to do that as part of this change

TBonnin (approved):
looks good. The seeders could be refactored so the required dependencies is injected. Like that they could be moved within the database package

packages/database/.gitignore
```diff
@@ -0,0 +1,3 @@
+tsconfig.tsbuildinfo
```
bodinsamuel, line 1: this .gitignore shouldn't be necessary

scripts/publish.sh
```diff
@@ -47,17 +47,37 @@ popd
...
 popd
 
+pushd "$GIT_ROOT_DIR/packages/database"
+jq '.bundleDependencies = true' package.json >temp.json && mv temp.json package.json
+npm install "@nangohq/utils@file:vendor/nangohq-utils-1.0.0.tgz" --workspaces=false
+npm install --workspaces=false
+npm pack --pack-destination "$GIT_ROOT_DIR/packages/shared/vendor"
+popd
+pushd "$GIT_ROOT_DIR/packages/shared"
+npm install "@nangohq/database@file:vendor/nangohq-database-1.0.0.tgz"
```
bodinsamuel, line 70: I feel like this double bundling will make the cli install even worst

bodinsamuel (approved):

scripts/publish.sh
khaliqgant: about 5 seconds slower

```
npm install @nangohq/cli@0.0.1-bf727d78a619ba3472a30392fa307cb8a424b214  11.30s user 4.37s system 110% cpu 14.121 total
```

```
npm install nango  6.55s user 3.20s system 139% cpu 6.982 total
```

PR #2237, opened 2024-05-31
chore: actions and webhooks can be executed by orchestrator

TBonnin:
The feature is behind 2 flags:
- orchestrator:immediate:ENVIRONMENT_ID to ramp up by environment
- orchestrator:immediate:global to enable it to everybody

In addition to add calls to the run.service in jobs.processor.handler this commit:
- adds the missing connection_id to the tasks payload (I had missed it)
- adds some tracing to be able to compare the execution via temporal/orchestrator
- deal with some postgres weirdness wrt json

### DIFF
--- a/packages/jobs/lib/processor/handler.ts
+++ b/packages/jobs/lib/processor/handler.ts
@@ -1,7 +1,13 @@
 import type { OrchestratorTask, TaskWebhook, TaskAction, TaskPostConnection } from '@nangohq/nango-orchestrator';
+import { jsonSchema } from '@nangohq/nango-orchestrator';
 import type { JsonValue } from 'type-fest';
 import { Err, Ok } from '@nangohq/utils';
 import type { Result } from '@nangohq/utils';
+import { configService, createSyncJob, getSyncByIdAndName, syncRunService, SyncStatus, SyncType } from '@nangohq/shared';
+import { logContextGetter } from '@nangohq/logs';
+import { records as recordsService } from '@nangohq/records';
+import integrationService from '../integration.service.js';
+import { bigQueryClient, slackService } from '../clients.js';
 
 export async function handler(task: OrchestratorTask): Promise<Result<JsonValue>> {
     task.abortController.signal.onabort = () => {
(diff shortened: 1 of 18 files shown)


packages/jobs/lib/processor/handler.ts
```diff
@@ -22,13 +28,78 @@ async function abort(_task: OrchestratorTask): Promise<Result<void>> {
 }
 
 async function action(task: TaskAction): Promise<Result<JsonValue>> {
```
TBonnin, line 33: when executing via Temporal the `reportFailure` function is also called in case of an exception but for actions/webhooks the only relevant logic is this telemetry logs https://github.com/NangoHQ/nango/blob/4c6e7e0b5ca2d84358d62c4a11d2c385f3877528/packages/jobs/lib/activities.ts#L476 which doesn't seem to be used (in favor of those ones https://github.com/NangoHQ/nango/blob/master/packages/shared/lib/clients/orchestrator.ts#L184). So I have decided not to called `reportFailure` then. Let me know if I am mistaken

packages/scheduler/lib/models/tasks.integration.test.ts
```diff
@@ -151,6 +151,14 @@ describe('Task', () => {
         expect(l5.length).toBe(2);
         expect(l5.map((t) => t.id)).toStrictEqual([t1.id, t2.id]);
     });
+    it('should be successfully saving json output', async () => {
+        const outputs = [1, 'one', true, null, ['a', 'b'], { a: 1, b: 2, s: 'two', arr: ['a', 'b'] }, [{ id: 'a' }, { id: 'b' }]];
+        for (const output of outputs) {
+            const task = await createTaskWithState(db, 'STARTED');
+            console.log('output', output, typeof output);
```
bodinsamuel:
```suggestion
```

packages/scheduler/lib/models/tasks.ts
```diff
@@ -184,19 +184,37 @@ export async function transitionState(
     }
 
     const output = 'output' in props ? props.output : null;
+    const asPostgresJson = (val: JsonValue) => {
+        if (val === null) {
+            return null;
+        }
+        if (Array.isArray(val)) {
+            // https://github.com/brianc/node-postgres/issues/442
```
bodinsamuel, line 192: ahah I have this exact comment in a 8 years old codebase 😬

packages/jobs/lib/processor/handler.ts
bodinsamuel: I think it highlights once again we need to split this in: SyncRun ActionRun WebhookRun and DryRun to clarify everything

packages/jobs/lib/processor/handler.ts
```diff
@@ -22,13 +28,78 @@ async function abort(_task: OrchestratorTask): Promise<Result<void>> {
...
 async function webhook(task: TaskWebhook): Promise<Result<JsonValue>> {
-    // TODO: Implement action processing
-    // Returning an error for now
-    return Err(`Not implemented: ${JSON.stringify({ taskId: task.id })}`);
+    const providerConfig = await configService.getProviderConfig(task.connection.provider_config_key, task.connection.environment_id);
+    if (providerConfig === null) {
+        return Err(`Provider config not found for connection: ${task.connection}`);
+    }
+
+    const sync = await getSyncByIdAndName(task.connection.id, task.parentSyncName);
```
bodinsamuel, line 73: ranting: descriptive names are great but this one is confusing since it's not using sync_id :D

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -77,86 +76,113 @@ export class Orchestrator {
...
-                            },
-                            (error) => {
-                                logger.error(`Error: Action '${executionId}' failed: ${stringifyError(error)}`);
-                            }
-                        );
+                    const actionResult = await this.client.executeAction({
+                        name: executionId,
+                        groupKey,
+                        args,
+                        timeoutSettingsInSecs: {
```
bodinsamuel: since there is a method dedicated, shouldn't this object be just the default value?

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -306,71 +331,101 @@ export class Orchestrator {
...
-                ]
-            });
-
-            const { success, error, response } = webhookHandler;
+                    if (res.isErr()) {
+                        span.setTag('error', res.error);
+                    }
+                } catch (e: unknown) {
+                    span.setTag('error', e);
+                    throw e;
```
bodinsamuel, line 413: it didn't used to throw, is it okay?

bodinsamuel (commented):
Just a quick pass, haven't had time to test this but looks good

packages/scheduler/lib/models/tasks.ts
TBonnin: 😬 and 😢

packages/shared/lib/clients/orchestrator.ts
khaliqgant: Also curious about this throw

khaliqgant (approved):

packages/shared/lib/clients/orchestrator.ts
TBonnin: you are right. I had custom values for the dry run but this is not necessary anymore. Thanks

packages/shared/lib/clients/orchestrator.ts
TBonnin: it didn't have a local try-catch before. This is just re-throwing for it to be handled by the main try-catch

bodinsamuel (approved):
Tested, works great 🙌🏻 

(not sure I already told you but when you push force github can't show what has changed, which makes review a bit more complicated)

TBonnin:
> Tested, works great 🙌🏻
> 
> (not sure I already told you but when you push force github can't show what has changed, which makes review a bit more complicated)

good point. Sorry that's an old habit of mine. I will add commits from now on instead of amending

PR #2238, opened 2024-05-31
chore: rename orchestrator waitForCompletion param to longPolling

TBonnin:
waitForCompletion isn't a very good name for the dequeue endpoint but I wanted to keep consistency with the /output endpoint. Renaming it to longPolling so it is consistent but also works for different endpoints

Note: it would break at deploy time since either the client will use the new name but the orchestrator service isn't yet (or vice versa depending on which service get started first). I will deploy while we are still in dry run mode then.


bodinsamuel (approved):
Much clearer ✨

khaliqgant (approved):
conflict :/ think from my update

Issue #2233, opened 2024-05-31
Please add config to pass HTTP_PROXY and HTTPS_PROXY so that requests to the third party integrations go via this proxy


mithlesh135:
const axios = require('axios');
const HttpProxyAgent = require('http-proxy-agent');
const HttpsProxyAgent = require('https-proxy-agent');

// Get the proxy URL from the environment variables
const proxyHttpUrl = process.env.HTTP_PROXY || process.env.http_proxy;
const proxyHttpsUrl = process.env.HTTPS_PROXY || process.env.https_proxy;

// Create the appropriate proxy agent
const httpAgent = proxyHttpUrl ? new HttpProxyAgent(proxyHttpUrl) : undefined;
const httpsAgent = proxyHttpsUrl ? new HttpsProxyAgent(proxyHttpsUrl) : undefined;

khaliqgant:
Will let you know once this is deployed, thanks for your contribution.

PR #2243, opened 2024-06-02
feat(proxy): support sending requests through HTTPS_PROXY

mithlesh135:
## Describe your changes
For external providers generally access is via a proxy, currently that is missing, this PR add the capability to set http_proxy or https_proxy in env and app will take care of proxying the request via this tunnel

#2233


khaliqgant (approved):

khaliqgant (approved):

bodinsamuel:
Is HTTP (no "s") a requirement for you? 
I don't think we should support that because it's a liability for credentials and potential security breach.

TBonnin (changes requested):
I think https://github.com/NangoHQ/nango/blob/6581d8a3114e7df3ed383907ac9db542eeae5f41/packages/shared/lib/services/proxy.service.ts should also be modified to take into account the proxy env var

mithlesh135:
> Is HTTP (no "s") a requirement for you? I don't think we should support that because it's a liability for credentials and potential security breach.

So the proxy will be a forward proxy which will sit on our side, so application can talk to this proxy on http, this proxy will be talking on https with external client

khaliqgant:
> I think https://github.com/NangoHQ/nango/blob/6581d8a3114e7df3ed383907ac9db542eeae5f41/packages/shared/lib/services/proxy.service.ts should also be modified to take into account the proxy env var

If they're running syncs and actions, right? Which I don't think they would be since this is self hosted.

mithlesh135:
> I think https://github.com/NangoHQ/nango/blob/6581d8a3114e7df3ed383907ac9db542eeae5f41/packages/shared/lib/services/proxy.service.ts should also be modified to take into account the proxy env var

Done

mithlesh135:
> > I think https://github.com/NangoHQ/nango/blob/6581d8a3114e7df3ed383907ac9db542eeae5f41/packages/shared/lib/services/proxy.service.ts should also be modified to take into account the proxy env var
> 
> If they're running syncs and actions, right? Which I don't think they would be since this is self hosted.

I have added it to this file as well

TBonnin:
> > > I think https://github.com/NangoHQ/nango/blob/6581d8a3114e7df3ed383907ac9db542eeae5f41/packages/shared/lib/services/proxy.service.ts should also be modified to take into account the proxy env var
> > 
> > 
> > If they're running syncs and actions, right? Which I don't think they would be since this is self hosted.
> 
> I have added it to this file as well

Thank you. Even though you might not use it if you are not running any integration scripts, it is good to be cover all use cases so people running enterprise releases can also use those env vars

TBonnin (approved):

bodinsamuel:
> So the proxy will be a forward proxy which will sit on our side, so application can talk to this proxy on http, this proxy will be talking on https with external client

Not sure I understand your answer. Also I'm not entirely getting the scope, should all calls to the external API goes through your Proxy or just the proxy one? If the former then it's missing some spots.


Here is a proposal for this PR https://github.com/NangoHQ/nango/compare/proposal-http-agent?expand=1

Differences:
- Renamed to utils/axios.ts
- No support for HTTP proxy (unless I'm missing something, we should never reach anything without encryption)
- Support for regular agent (
- Missing connection file (@khaliqgant can you double check)

khaliqgant:
> * Missing connection file (@khaliqgant can you double check)

Yeah, good call on the connection.service file. For that matter if it is all outgoing requests then the `webhook.service.ts` file should be updated as well.

mithlesh135:
> > So the proxy will be a forward proxy which will sit on our side, so application can talk to this proxy on http, this proxy will be talking on https with external client
> 
> Not sure I understand your answer. Also I'm not entirely getting the scope, should all calls to the external API goes through your Proxy or just the proxy one? If the former then it's missing some spots.
> 
> Here is a proposal for this PR https://github.com/NangoHQ/nango/compare/proposal-http-agent?expand=1
> 
> Differences:
> 
> * Renamed to utils/axios.ts
> * No support for HTTP proxy (unless I'm missing something, we should never reach anything without encryption)
> * Support for regular agent (
> * Missing connection file (@khaliqgant can you double check)

Taken care for most of proposals, let me know if I should rename as well

mithlesh135:
> > * Missing connection file (@khaliqgant can you double check)
> 
> Yeah, good call on the connection.service file. For that matter if it is all outgoing requests then the `webhook.service.ts` file should be updated as well.

added both the files

mithlesh135:
> https://github.com/NangoHQ/nango/compare/proposal-http-agent?expand=1

@bodinsamuel   Support for http-proxy will be needed, most of these kind of proxies connect to own apps via http only as app and this proxy will sit in same network so its fine to communicate over http, now this proxy will be making https calls to external provider


Screenshot 2024-06-03 at 6 58 44 PM

packages/server/lib/controllers/proxy.controller.ts
```diff
@@ -7,7 +7,7 @@ import type { UrlWithParsedQuery } from 'url';
 import url from 'url';
 import querystring from 'querystring';
 import type { AxiosError, AxiosRequestConfig, AxiosResponse } from 'axios';
-import axios from 'axios';
+import { axiosInstance as axios } from '@nangohq/shared/lib/axiosInstance/index.js';
```
bodinsamuel: import is wrong, it should be export from shared index.ts

packages/shared/lib/axiosInstance/index.ts
bodinsamuel, line 1: would still put it in utils, to avoid having one folder per util

bodinsamuel (commented):
I now see the point with the graph thanks 👌🏻 🙏🏻 

Still believe it's a footgun but we'll see. Added a few comments, otherwise looks good. I'll apply my change regarding the agent in an other PR 👍🏻

packages/shared/lib/axiosInstance/index.ts
mithlesh135: ok, let me move it to utils then

mithlesh135: not getting one folder per util ?

packages/server/lib/controllers/proxy.controller.ts
mithlesh135: fixed

packages/shared/lib/axiosInstance/index.ts
mithlesh135: @bodinsamuel Done (moved to utils)

packages/shared/lib/clients/oauth2.client.ts
```diff
@@ -93,10 +111,14 @@ export async function getFreshOAuth2Credentials(
         }
 
         if (Boom.isBoom(e)) {
-            const payload = { external_message: e.message, external_request_details: e.output, dataMessage: errorPayload };
+            const payload = {
+                external_message: e.message,
+                external_request_details: e.output,
+                dataMessage: errorPayload instanceof Buffer ? errorPayload.toString() : errorPayload
```
bodinsamuel, line 117: This would output raw buffer

packages/shared/lib/clients/oauth2.client.ts
```diff
@@ -93,10 +111,14 @@ export async function getFreshOAuth2Credentials(
...
+            };
             nangoErr = new NangoError(`refresh_token_external_error`, payload);
         } else {
-            nangoErr = new NangoError(`refresh_token_external_error`, { message: errorPayload });
+            nangoErr = new NangoError(`refresh_token_external_error`, { message: e.message });
```
bodinsamuel, line 121: This would output raw buffer

packages/shared/lib/services/connection.service.ts
```diff
@@ -702,6 +702,7 @@ class ConnectionService {
                     });
                 }
 
+                // TODO: this leak credentials to the logs
                 const errorWithPayload = new NangoError(error.type, connection);
 
                 return Err(errorWithPayload);
```
bodinsamuel, line 708: This return is not nice, it's leaking credentials and also force the API to return a 400 which is not true

bodinsamuel (commented):
@TBonnin @khaliqgant 
Can you test locally please? I can't manage to refresh on my side I keep getting `incomplete_raw_credentials` for some reason.

khaliqgant (approved):
Tested this and works as expected for normal operations with Nango (proxy, auth call etc)

khaliqgant:
> @TBonnin @khaliqgant Can you test locally please? I can't manage to refresh on my side I keep getting `incomplete_raw_credentials` for some reason.

Noting here that if you were testing with Github and forcing a refresh in the code it is going to give the `incomplete_raw_credentials` response because Github doesn't issue a refresh token

bodinsamuel:
> Noting here that if you were testing with Github and forcing a refresh in the code it is going to give the incomplete_raw_credentials response because Github doesn't issue a refresh token

Alright, thanks a lot for re-testing.

PR #2244, opened 2024-06-03
fix: allow folder hosting

bodinsamuel:


bodinsamuel:
Just keeping that open to not forget about few fixes inside that are relevant

bodinsamuel:
Closed in favor of more recent changes

PR #2246, opened 2024-06-03
fix(ui): use common fetch method

bodinsamuel:
## Describe your changes

Contributes to NAN-1079

One of the many PR to make our current setup more flexible to hostname/pathname changes. 

- Use one single method to call the API
There is no way to enforce that right now and until we migrate to one lib/way to do so please be careful about not introducing fetch again ☺️ 
This will help change the hostname/pathname in a single place, and also introduce the typed API


packages/webapp/src/pages/InteractiveDemo/index.tsx
```diff
@@ -28,12 +29,6 @@ export const InteractiveDemo: React.FC = () => {
     const env = useStore((state) => state.env);
     const { environmentAndAccount } = useEnvironment(env);
 
-    useEffect(() => {
-        if (env !== 'dev') {
-            window.location.href = `/${env}/integrations`;
-        }
-    }, [env]);
```
bodinsamuel, line 35: could not reach this state

khaliqgant (approved):

PR #2250, opened 2024-06-03
chore(processor): add tracing/logging to processTask function

TBonnin:
This commit is adding a trace to better monitor and debug task processing.

### DIFF
--- a/packages/jobs/lib/processor/processor.worker.ts
+++ b/packages/jobs/lib/processor/processor.worker.ts
@@ -4,6 +4,7 @@ import { Worker, isMainThread } from 'node:worker_threads';
 import { getLogger, stringifyError } from '@nangohq/utils';
 import { OrchestratorClient, OrchestratorProcessor } from '@nangohq/nango-orchestrator';
 import { handler } from './handler.js';
+import tracer from 'dd-trace';
 
 const logger = getLogger('jobs.processor.worker');
 
@@ -84,7 +85,7 @@ export class ProcessorChild {
 
     async start(): Promise<void> {
         logger.info(`Starting Processor: ${JSON.stringify(this.opts)}`);
-        this.processor.start();
+        this.processor.start({ tracer });
     }
 
     stop(): void {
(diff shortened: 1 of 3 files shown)


packages/orchestrator/lib/clients/processor.ts
```diff
@@ -94,23 +95,44 @@ export class OrchestratorProcessor {
...
                 if (task.abortController.signal.aborted) {
                     // task was aborted while waiting in the queue
+                    logger.info(`task ${task.id} was aborted before processing started`);
                     return;
                 }
                 const res = await this.handler(task);
                 if (res.isErr()) {
-                    await this.orchestratorClient.failed({ taskId: task.id, error: res.error });
+                    const setFailed = await this.orchestratorClient.failed({ taskId: task.id, error: res.error });
+                    if (setFailed.isErr()) {
```
TBonnin, line 119: before if the call to `orchestratorClient.failed` failed we had no info about it (and the task will eventually expire). At least now we log something

khaliqgant (approved):

TBonnin:
deployed this branch to staging but I can see any traces in DD. Not sure why. I will check tomorrow

packages/jobs/lib/processor/processor.worker.ts
```diff
@@ -84,7 +85,7 @@ export class ProcessorChild {
 
     async start(): Promise<void> {
         logger.info(`Starting Processor: ${JSON.stringify(this.opts)}`);
-        this.processor.start();
+        this.processor.start({ tracer });
```
bodinsamuel, line 88: you sure you need to pass the tracer? I don't think it's necessary

bodinsamuel (approved):

packages/jobs/lib/processor/processor.worker.ts
TBonnin: It is not entirely necessary but since it is used in the orchestrator library I wanted to make it clear where the Tracer is coming from

PR #2252, opened 2024-06-04
fix(orchestratorClient): race condition in dequeue

TBonnin:
A race condition would happen in dequeue long polling if tasks are being created while tasks are being dequeued. Those newly created tasks would be dequeued as well but never returned to the processor, hence ending up being expired.
The solution to avoid the race condition is simple: only listen to the task creation event if no tasks have been dequeued

### DIFF
--- a/packages/orchestrator/lib/clients/processor.integration.test.ts
+++ b/packages/orchestrator/lib/clients/processor.integration.test.ts
@@ -103,10 +103,10 @@ async function processN(handler: (task: TaskAction | TaskWebhook | TaskPostConne
         handler,
         opts: { orchestratorClient, groupKey, maxConcurrency: n, checkForTerminatedInterval: 100 }
     });
+    processor.start();
     for (let i = 0; i < n; i++) {
         await scheduleTask({ groupKey });
     }
-    processor.start();
     // Wait so the processor can process all tasks
     await new Promise((resolve) => setTimeout(resolve, 1000));
     return processor;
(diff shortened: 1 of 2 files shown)


packages/orchestrator/lib/clients/processor.integration.test.ts
```diff
@@ -103,10 +103,10 @@ async function processN(handler: (task: TaskAction | TaskWebhook | TaskPostConne
         handler,
         opts: { orchestratorClient, groupKey, maxConcurrency: n, checkForTerminatedInterval: 100 }
     });
+    processor.start();
```
TBonnin, line 106: The race condition would have been triggered if the tasks had been created after the processor was started

packages/orchestrator/lib/routes/v1/postDequeue.ts
```diff
@@ -69,14 +69,13 @@ const handler = (scheduler: Scheduler, eventEmitter: EventEmitter) => {
...
 
-        eventEmitter.once(eventId, onTaskStarted);
-
         const getTasks = await scheduler.dequeue({ groupKey, limit });
         if (getTasks.isErr()) {
             cleanupAndRespond((res) => res.status(500).json({ error: { code: 'dequeue_failed', message: getTasks.error.message } }));
             return;
         }
         if (longPolling && getTasks.value.length === 0) {
+            eventEmitter.once(eventId, onTaskStarted);
```
TBonnin, line 78: Tasks could be dequeued at L72 but also at the same time in the event listener function on L61. But only the first set of dequeued tasks will be returned, leaving the other tasks as started. They will eventually expires. 
The fix is simple: only listen to task creation event if long polling and initial dequeuing returned no tasks

bodinsamuel (approved):

khaliqgant (approved):

PR #2254, opened 2024-06-04
feat(webhooks): [nan-1063] webhook on refresh error and slack notification

khaliqgant:
## Describe your changes
Add in webhook and slack delivery for connection refresh error and for slack refresh resolution.

## Additional Work
- Add in a dedicated `webhooks` package which required a lot of type movement which bloated this PR a bit. Add in additional refresh success and error hooks to be able to consolidate logic in one place. 

NAN-1063


packages/webhooks/lib/sync.ts
khaliqgant, line 1: Will flesh out these empty files in a future PR.

packages/webhooks/lib/types.ts
```diff
@@ -0,0 +1,36 @@
+import type { SyncType, FailedConnectionError, AuthOperationType, AuthModeType, SyncResult } from '@nangohq/types';
+import type { WebhookType } from './enums.js';
+
```
khaliqgant, line 3: Maybe these should go in the types package? In case users want to export these types. Would have to remove the enums. Thoughts?

scripts/publish.sh
```diff
@@ -84,6 +84,7 @@ bump_other_pkg "shared" "types"
 bump_other_pkg "server" "types"
 bump_other_pkg "webapp" "types"
 bump_other_pkg "cli" "types"
+bump_other_pkg "webhooks" "types"
```
bodinsamuel, line 87: actually I'm not even sure why I bumped anything that is not published (server and webapp)

packages/webhooks/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "dependencies": {
+        "@nangohq/logs": "file:../logs",
+        "@nangohq/utils": "file:../utils",
+        "axios": "^1.3.4",
+        "dayjs": "^1.11.7",
+        "dayjs-plugin-utc": "^0.1.2",
+        "exponential-backoff": "^3.1.1"
+    },
+    "devDependencies": {
+        "@nangohq/types": "^0.39.32",
```
bodinsamuel, line 26: could be `file:`

packages/shared/lib/services/notification/webhook.service.unit.test.ts
```diff
@@ -25,164 +18,16 @@ const integration: Config = {
...
-    connection_id: '1',
-    provider_config_key: 'providerkey',
-    connection_config: {},
-    credentials: {},
-    environment_id: 1
-};
 
 const getLogCtx = () => new LogContext({ parentId: '1', operation: {} as any }, { dryRun: true, logToConsole: false });
 
+const spy = vi.spyOn(axiosInstance, 'post');
```
bodinsamuel, line 24: indeed :D

packages/shared/lib/services/connection.service.ts
```diff
@@ -594,14 +583,26 @@ class ConnectionService {
...
+        connectionRefreshFailedHook
     }: {
         account: Account;
         environment: Environment;
         connectionId: string;
         providerConfigKey: string;
         logContextGetter: LogContextGetter;
         instantRefresh: boolean;
+        connectionRefreshSuccessHook: (args: { connection: Connection; environment: Environment; config: ProviderConfig }) => Promise<void>;
+        connectionRefreshFailedHook: (args: {
```
TBonnin: I'll call it `onRefreshFailed` and `onRefreshSuccess` or something like this

packages/shared/lib/services/connection.service.ts
```diff
@@ -594,14 +583,26 @@ class ConnectionService {
...
+            connection: Connection;
+            activityLogId: number;
+            logCtx: LogContext;
+            authError: { type: string; description: string };
+            environment: Environment;
+            template: ProviderTemplate;
+            config: ProviderConfig;
+        }) => Promise<void>;
```
TBonnin, line 605: we are soon gonna need to an event based mechanism. Passing callbacks like this won't be manageable if we add a few more

packages/webhooks/lib/enums.ts
```diff
@@ -0,0 +1,5 @@
+export enum WebhookType {
```
TBonnin, line 1: why its own file and not in ./types?

bodinsamuel:
Sorry I started reviewing and got blocked on something else, will finish tomorrow morning!

packages/webhooks/lib/utils.ts
```diff
@@ -0,0 +1,136 @@
...
+        return true;
+    }
+
+    const authNotSelected = options?.auth && !environment.send_auth_webhook;
+
+    if (!hasAnyWebhook || authNotSelected) {
+        return false;
+    }
+
+    return true;
```
TBonnin, line 85: cannot we simplify it like this: 
```
  if (!hasAnyWebhook) {
    return false;
  }
  
  if (options?.forward) {
    return true;
  }
  
  if (options?.auth && !environment.send_auth_webhook) {
    return false;
  }
  
  return true;
```

TBonnin (approved):

packages/shared/lib/services/connection.service.ts
khaliqgant: Yeah, agreed.

packages/webhooks/lib/enums.ts
khaliqgant: In case we move it to types later I wanted it separate and since enums aren't a strict type so the import can be `import type...`

packages/webhooks/lib/utils.ts
khaliqgant: This whole condition is going to change with the additional notifications so will simplify this later.

packages/shared/lib/services/notification/slack.service.ts
```diff
@@ -210,26 +202,19 @@ export class SlackService {
...
      */
-    async reportFailure(
-        nangoConnection: NangoConnection,
-        syncName: string,
-        syncType: SyncType,
-        originalActivityLogId: number,
-        environment_id: number,
-        provider: string
-    ) {
+    async reportFailure(nangoConnection: NangoConnection, name: string, type: string, originalActivityLogId: number, environment_id: number, provider: string) {
```
bodinsamuel, line 205: I would just question the rename from syncName to name. I'm all for short name when in context but here it's not explicit when used and not explicit when reading this function code.

What motivated this?

packages/webhooks/lib/types.ts
bodinsamuel: that would be great if types does not export any code to avoid potential mistakes, if you both want to I'm not going to against it but if we can avoid it I'm happy :D

bodinsamuel (approved):
Works great 🥇 The amount of conflict I will have is off the roof :D 

Just a side note, we are trying to split our PRs a bit (not perfect at all but trying) that would also help you getting faster reviews.

packages/webhooks/lib/types.ts
khaliqgant: I agree that `types` should not export code and I spent a lot of time with this PR changing enums to interfaces to avoid just that.

bodinsamuel: yeah and imo I doesn't make the codebase any less good (or better tbf) it's just one less thing to import

packages/shared/lib/services/connection.service.ts
khaliqgant: Updated with https://github.com/NangoHQ/nango/pull/2254/commits/94814ab1aca3d0db51076c07d45b98e3a3d0526f

packages/shared/lib/services/notification/slack.service.ts
khaliqgant: I changed the name because it isn't always a `syncName` anymore but can be a connection name or a sync name with the addition of `auth` slack notifications

PR #2256, opened 2024-06-04
chore: add sync type and validation for orchestrator TaskSync

TBonnin:
Just the types and validation. No sync are yet scheduled or processed

### DIFF
--- a/packages/orchestrator/lib/clients/validate.ts
+++ b/packages/orchestrator/lib/clients/validate.ts
@@ -45,7 +45,6 @@ export const postConnectionArgsSchema = z.object({
     postConnectionName: z.string().min(1),
     fileLocation: z.string().min(1),
     activityLogId: z.number().positive(),
-    input: jsonSchema,
     ...commonSchemaArgsFields
 });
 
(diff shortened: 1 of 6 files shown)


packages/orchestrator/lib/clients/validate.ts
```diff
@@ -1,17 +1,64 @@
...
+    ...commonSchemaArgsFields
+});
+export const postConnectionArgsSchema = z.object({
+    type: z.literal('post-connection-script'),
+    postConnectionName: z.string().min(1),
+    fileLocation: z.string().min(1),
+    activityLogId: z.number().positive(),
+    input: jsonSchema,
+    ...commonSchemaArgsFields
+});
```
TBonnin, line 49: adding syncArgsSchema and moving all the zod schema in `validate.ts` where they belong

packages/orchestrator/lib/clients/validate.ts
```diff
@@ -1,17 +1,64 @@
...
+export const postConnectionArgsSchema = z.object({
+    type: z.literal('post-connection-script'),
+    postConnectionName: z.string().min(1),
+    fileLocation: z.string().min(1),
+    activityLogId: z.number().positive(),
+    input: jsonSchema,
```
bodinsamuel: Can you remove input then? cc #2259

bodinsamuel (approved):
lgtm

packages/orchestrator/lib/clients/validate.ts
TBonnin: yep

PR #2257, opened 2024-06-04
fix(logs): handle buffered activities

bodinsamuel:
## Describe your changes

Fixes NAN-944

- Allow buffered activities in proxy to send their timestamp to Persist
- Fix: finally managed to find a type for `log()` and its weird signature 
- Pass PersistAPI to Nango() as dependency for mocking purpose

### DIFF
--- a/packages/persist/lib/server.ts
+++ b/packages/persist/lib/server.ts
@@ -41,11 +41,14 @@ server.post(
         params: z.object({
             environmentId: z.string().transform(Number).pipe(z.number().int().positive()) as unknown as z.ZodNumber
         }),
-        body: z.object({
-            activityLogId: z.number(),
-            level: z.enum(logLevelValues),
-            msg: z.string()
-        })
+        body: z
+            .object({
+                activityLogId: z.number(),
+                level: z.enum(logLevelValues),
+                msg: z.string(),
+                timestamp: z.number().optional() // Optional until fully deployed
+            })
+            .strict()
     }),
     persistController.saveActivityLog.bind(persistController)
 );
--- a/packages/shared/lib/sdk/sync.ts
+++ b/packages/shared/lib/sdk/sync.ts
@@ -574,8 +574,8 @@ export class NangoAction {
      * http = green
      * silly = light green
      */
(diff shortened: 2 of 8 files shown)


packages/persist/lib/server.ts
```diff
@@ -41,11 +41,14 @@ server.post(
...
-            msg: z.string()
-        })
+        body: z
+            .object({
+                activityLogId: z.number(),
+                level: z.enum(logLevelValues),
+                msg: z.string(),
+                timestamp: z.number().optional() // Optional until fully deployed
+            })
+            .strict()
```
bodinsamuel, line 51: Since we don't deploy everything at once I took a safe approach, will change it to required later

packages/shared/lib/services/proxy.service.ts
```diff
@@ -536,13 +535,7 @@ class ProxyService {
...
     }
 
-    private async handleResponse(
-        response: AxiosResponse,
-        config: ApplicationConstructedProxyConfiguration,
-        activityLogId: number,
-        environment_id: number,
-        url: string
-    ): Promise<RouteResponse & Activities> {
+    private handleResponse(config: ApplicationConstructedProxyConfiguration, activityLogId: number, environment_id: number, url: string): Activities {
```
bodinsamuel, line 538: changed to promise to sync, since it didn't had any await

packages/shared/lib/sdk/sync.ts
```diff
@@ -565,7 +574,9 @@ export class NangoAction {
      * http = green
      * silly = light green
      */
-    public async log(...args: any[]): Promise<void> {
+    public async log(s: any, a?: { level: LogLevel }): Promise<void>;
+    public async log(s: string, ...args: [any, { level: LogLevel }]): Promise<void>;
```
TBonnin: can we call `s` message and `a` options or something similar to make it slightly easier to read. Also shouldn't `level` be `level?`?

TBonnin (approved):

packages/shared/lib/sdk/sync.ts
bodinsamuel: definitely, got carried away when I found the working type :D

PR #2225, opened 2024-05-30
feat(server): [nan-906] run external client post connection scripts

khaliqgant:
## Describe your changes
Allow a customer to run a post installation script provided by them. Note that this task has a client deadline so has a bit more urgency. 

## Issue ticket number and link
NAN-906

## Checklist before requesting a review (skip if just adding/editing APIs & templates)
- [ ] I added tests, otherwise the reason is: 
- [ ] I added observability, otherwise the reason is:
- [ ] I added analytics, otherwise the reason is:

PR #2259, opened 2024-06-04
fix: post script connection error + handle in logs

bodinsamuel:
## Describe your changes

Fixes NAN-1093

I haven't paid enough attention when reviewing, there was a few things not working in post connection and in logs

PSC:
- pass missing `connection_id`, ~and missing `input` (@TBonnin I didn't wanted to change the default for input and I'm sure you already touching it in your own PR)~

Logs:
- Handle post connection as it was meant to (in operation `auth`)
- Create a log also for internal post connection for consistency
- Properly append error to the log (do not json stringify)
- UI: Show message correctly on operation hover

## Test

Screenshot 2024-06-04 at 18 17 34

- @TBonnin those scripts does not work on my setup, maybe it's something on my machine but they just timeout
Screenshot 2024-06-04 at 18 19 11


packages/server/lib/hooks/connection/post-connection.ts
```diff
@@ -87,34 +100,7 @@ async function execute(createdConnection: RecentlyCreatedConnection, provider: s
...
-                    provider_config_key: upsertedConnection.provider_config_key,
-                    environment_id: environment.id
-                };
-
-                const activityLogId = await createActivityLogAndLogMessage(log, {
-                    level: 'error',
-                    environment_id: environment.id,
-                    timestamp: Date.now(),
-                    content: `Post connection script failed with the error: ${errorString}`
-                });
```
bodinsamuel, line 108: took the liberty to delete v1 logs, because I'm stripping everything down soon anyway

packages/server/lib/hooks/connection/external-post-connection.ts
```diff
@@ -41,7 +41,7 @@ export async function externalPostConnection(
     const activityLogId = await createActivityLog(log);
 
     const logCtx = await logContextGetter.create(
-        { id: String(activityLogId), operation: { type: 'post-connection-script' }, message: 'Start action' },
+        { id: String(activityLogId), operation: { type: 'auth', action: 'post_connection' }, message: 'Start external post connection' },
```
khaliqgant: I don't know if "Start external post connection" makes sense. Would rather it say "Start external post connection script".

packages/server/lib/hooks/connection/post-connection.ts
```diff
@@ -74,8 +75,20 @@ async function execute(createdConnection: RecentlyCreatedConnection, provider: s
         const handler = handlers[`${provider.replace(/-/g, '')}PostConnection`];
 
         if (handler) {
+            logCtx = await logContextGetter.create(
+                { operation: { type: 'auth', action: 'post_connection' }, message: 'Start internal post connection' },
```
khaliqgant: See note above about adding "script.

"external" vs "internal" might be confusing to the user. Maybe for both it is just "Start post connection script". I think originally we didn't log this (which was intentional) and maybe since it is internal we should log it and on failure we just keep an internal `logger` log so we can know about it. A user might be confused to see this log that they wouldn't necessarily know anything about.

khaliqgant (changes requested):
Wording and process changes. Let me know what you think.

TBonnin:
>     * @TBonnin those scripts does not work on my setup, maybe it's something on my machine but they just timeout
> 
> 
> Screenshot 2024-06-04 at 18 19 11

Post connection script are currently not handled. https://github.com/NangoHQ/nango/pull/2225/files#diff-b5bb1e099cf24eec3ed6f52521ac05474b29cbef782cbf0b0db8dae7a2dee3ecR39-R41
I think @khaliqgant is planning to add post connection script support in the processor. In the meantime they should be routed to Temporal

packages/server/lib/hooks/connection/external-post-connection.ts
bodinsamuel: oops yes you are right

packages/server/lib/hooks/connection/post-connection.ts
bodinsamuel: My own feeling about this is that it's part of the system that interact with the third parties so it should be logged. And it's not even enough imo since we don't handle the buffered logs in the proxy call.

> A user might be confused to see this log that they wouldn't necessarily know anything about.

I feel like it's even worse if we only show errors to the users, they could think it never works

Also, no logs can mean "nothing was executed" or "something was a success "or "something crashed", which is not super nice for observability. And stdout don't scale indefinitely either. 

---

Happy to involve @bastienbeurier on this. The context, we currently log: 
- external post connection script: failure and success  
- internal post connection: failure

**We don't we log successful internal post connection, should we do it or keep it as is?**
(side note: we never log the absence of post connection)

Alternative, should this log be part of the auth:create_connection log instead?

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -527,11 +528,13 @@ export class Orchestrator {
...
                         connection: {
                             id: connection.id!,
+                            connection_id: connection.connection_id,
                             provider_config_key: connection.provider_config_key,
                             environment_id: connection.environment_id
                         },
                         activityLogId,
-                        fileLocation
+                        fileLocation,
+                        input: null
```
TBonnin: should removed since Post Connection Script don't take any input

packages/server/lib/hooks/connection/post-connection.ts
bastienbeurier: IMO, we should err on the side of transparency, logging everything but being very explicit about what is executed by the platform script vs. what is executed by a user script.

khaliqgant (approved):

TBonnin (approved):

PR #2260, opened 2024-06-04
feat(orchestrator): add support for schedules

TBonnin:
Based on https://github.com/NangoHQ/nango/pull/2256. Review this one first 🙏 

This PR is adding model and logic to create schedules in the orchestrator. In a following PR I will add a scheduler worker_thread that will create tasks based on exisiting schedules when necessary

### DIFF
--- a/packages/orchestrator/lib/clients/client.integration.test.ts
+++ b/packages/orchestrator/lib/clients/client.integration.test.ts
@@ -34,9 +34,34 @@ describe('OrchestratorClient', async () => {
         scheduler.stop();
         await dbClient.clearDatabase();
     });
+
+    describe('recurring schedule', () => {
+        it('should be created', async () => {
+            const res = await client.recurring({
+                name: 'Task',
+                startsAt: new Date(),
+                frequencyMs: 300_000,
+                args: {
+                    type: 'sync',
+                    syncId: 'sync-a',
+                    syncName: rndStr(),
+                    syncJobId: 5678,
+                    connection: {
+                        id: 123,
+                        connection_id: 'C',
+                        provider_config_key: 'P',
+                        environment_id: 456
+                    },
+                    debug: false
+                }
+            });
+            expect(res.isOk()).toBe(true);
+        });
+    });
+
     describe('heartbeat', () => {
         it('should be successful', async () => {
-            const scheduledTask = await client.schedule({
+            const scheduledTask = await client.immediate({
                 name: 'Task',
                 groupKey: rndStr(),
                 retry: { count: 0, max: 0 },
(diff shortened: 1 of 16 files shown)


packages/orchestrator/lib/clients/client.integration.test.ts
```diff
@@ -34,9 +34,34 @@ describe('OrchestratorClient', async () => {
...
+                }
+            });
+            expect(res.isOk()).toBe(true);
+        });
+    });
+
     describe('heartbeat', () => {
         it('should be successful', async () => {
-            const scheduledTask = await client.schedule({
+            const scheduledTask = await client.immediate({
```
TBonnin, line 64: I was originally planning to have a single function to create immediate tasks and recurring schedules but creating a schedule doesn't necessary mean scheduling a task so the return type is different. 
There is now two methods: `immediate` to schedule a task immediately and `recurring` that is creating a schedule

packages/scheduler/lib/scheduler.ts
```diff
@@ -82,7 +83,7 @@ export class Scheduler {
     }
 
     /**
-     * Schedule a task
+     * Schedule a task immediatly
```
bodinsamuel:
```suggestion
     * Schedule a task immediately
```

packages/scheduler/lib/db/migrations/20240604135056_schedules_model.ts
```diff
@@ -0,0 +1,39 @@
...
+                updated_at timestamp with time zone NOT NULL,
+                deleted_at timestamp with time zone NULL
+            );
+        `);
+        // add foreign key schedule_id to tasks_table, cascade delete and nullable
+        await trx.raw(`
+            ALTER TABLE ${TASKS_TABLE}
+            ADD COLUMN IF NOT EXISTS schedule_id uuid REFERENCES ${SCHEDULES_TABLE}(id) ON DELETE CASCADE;
+        `);
+        // TODO: add indexes
```
bodinsamuel, line 32: I guess at least `where deleted_at IS NULL` would be great.

packages/scheduler/lib/db/migrations/20240604135056_schedules_model.ts
```diff
@@ -0,0 +1,39 @@
...
+                'DELETED'
+            );
+        `);
+        await trx.raw(`
+            CREATE TABLE IF NOT EXISTS ${SCHEDULES_TABLE} (
+                id uuid PRIMARY KEY,
+                name varchar(255) NOT NULL,
+                state schedule_states NOT NULL,
+                starts_at timestamp with time zone NOT NULL,
+                frequency interval NOT NULL,
```
bodinsamuel, line 20: funny never used or saw that type before

bodinsamuel (approved):
Looks good 👏🏻

packages/scheduler/lib/db/migrations/20240604135056_schedules_model.ts
TBonnin: Yes I will add the indexes in next PR when I add the logic to schedule tasks based on schedules

khaliqgant (approved):
Looks good. As discussed the API around scheduling (get next runs), updating only a schedule frequency via API call should be supported.

PR #2263, opened 2024-06-04
fix(server): [nan-979] Remove dupes if two or more syncs have an error for the connections list page

khaliqgant:
## Describe your changes
If a connection has two or more syncs that have an error it would duplicate that connection on the list page. Use [join lateral](https://www.postgresql.org/docs/9.4/queries-table-expressions.html) > 9.3 postgres version to limit the join.

NAN-979


packages/shared/lib/services/connection.service.ts
```diff
@@ -528,9 +528,9 @@ class ConnectionService {
     ): Promise<{ id: number; connection_id: string; provider: string; created: string; metadata: Metadata; active_logs: ActiveLogIds }[]> {
         const queryBuilder = db.knex
             .from<Connection>(`_nango_connections`)
-            .leftJoin(ACTIVE_LOG_TABLE, function () {
-                this.on('_nango_connections.id', '=', `${ACTIVE_LOG_TABLE}.connection_id`).andOnVal(`${ACTIVE_LOG_TABLE}.active`, true);
-            })
+            .joinRaw(
+                `LEFT JOIN LATERAL (SELECT * FROM ${ACTIVE_LOG_TABLE} WHERE _nango_connections.id = ${ACTIVE_LOG_TABLE}.connection_id AND ${ACTIVE_LOG_TABLE}.active = true LIMIT 1) ${ACTIVE_LOG_TABLE} ON true`
```
TBonnin: not sure the active logs belong into this query but can you add a comment to explain why we are doing a lateral join subquery.

TBonnin (approved):

packages/shared/lib/services/connection.service.ts
khaliqgant: Added comment with https://github.com/NangoHQ/nango/pull/2263/commits/16291a900206a59a5f0e312338eb711a0ae1d0a6

packages/shared/lib/services/connection.service.ts
```diff
@@ -528,9 +528,14 @@ class ConnectionService {
...
         const queryBuilder = db.knex
             .from<Connection>(`_nango_connections`)
-            .leftJoin(ACTIVE_LOG_TABLE, function () {
-                this.on('_nango_connections.id', '=', `${ACTIVE_LOG_TABLE}.connection_id`).andOnVal(`${ACTIVE_LOG_TABLE}.active`, true);
-            })
+            // A connection can have multiple different syncs that are failing
+            // which means it could have different entries in the active logs table.
+            // We only need to know that the connection has any active logs
+            // and we don't want the connection to show multiple times.
+            // We use a lateral join to get only one active log.
```
bodinsamuel: my take: either a second query or sub select would make all of this clearer

bodinsamuel (approved):

packages/shared/lib/services/connection.service.ts
khaliqgant: Point taken, updated to a sub query

PR #2265, opened 2024-06-04
feat(integrations): add support for instantly api key

jwd-dev:
## Describe your changes

- Adds support for instantly api key.
- Adds a simple set campaign name action script.


packages/shared/providers.yaml
```diff
@@ -1161,6 +1161,15 @@ insightly:
             method: GET
             endpoint: /v3.1/Contacts
     docs: https://docs.nango.dev/integrations/all/insightly
+instantly:
+    categories:
+        - marketing
+    auth_mode: API_KEY
+    proxy:
+        base_url: https://api.instantly.ai/
+        headers:
```
bastienbeurier: Could you test that it works to insert the API key in the headers? 

In the docs, they use a [query param for GET](https://developer.instantly.ai/campaign-1/get-campaign-accounts), and a [body param for POST](https://developer.instantly.ai/campaign-1/add-sending-account-to-campaign).

jwd-dev: We use it locally in headers, will confirm for you though.

jwd-dev: Actually you're correct, I was confused with another one of our integrations.

hassan254-prog:
@jwd-dev , I have made a small change to the pull request, when you have a moment, please check it out  and run the sample action script on your end then provide me with your feedback. Thank you!

bodinsamuel (approved):

integration-templates/instantly/actions/set-campaign-name.ts
```diff
@@ -0,0 +1,45 @@
...
+        const connection = await nango.getConnection();
+
+        let api_key: string;
+        if ('apiKey' in connection.credentials) {
+            api_key = connection.credentials.apiKey;
+        } else {
+            throw new nango.ActionError({
+                message: `API key credentials is incomplete`
+            });
+        }
```
TBonnin, line 24: nitpicking: we can probably use the null coaleasce operator here:
```
const api_key: string = connection.credentials.apiKey ?? 
  throw new nango.ActionError({
    message: 'API key credentials is incomplete'
  });
```

TBonnin (approved):

integration-templates/instantly/actions/set-campaign-name.ts
hassan254-prog: @TBonnin TypeScript's strict type checking wouldn't be happy here. An error is raised since ts can't guarantee that `credentials` is of type `ApiKeyCredentials`

PR #2267, opened 2024-06-04
fix: actions/webhooks without input should not failed

TBonnin:
fixing an issue where undefined input will fail the action/webhook instead of considering it as no input


khaliqgant (approved):
Thinking about this edge case, locally I have a lot of actions that don't have input so ideally I would have been using the orchestrator and this would have been caught.

I think as we develop big updates like this we should encourage devs to actively use the feature in their normal day to day development once it was ready. We should encourage this and it will benefit us in the long run to caught issues earlier.

bodinsamuel (approved):

PR #2268, opened 2024-06-05
feat: expose environment name in scripts

TBonnin:
Customer wants to have access to the environment name from within their custom scripts. See linear ticket for more info.

https://linear.app/nango/issue/NAN-967/surface-env-name-in-scripts


bodinsamuel (approved):
⛱️

PR #2270, opened 2024-06-05
fix(orchestrator): accept bigger task output

TBonnin:
I was a bit conservative with my 100k payload limit for the orchestrator server. Upping to 10mb because some customers are hitting the limit. This commit also fixes setting the default timeout and we don't log the task payload and ouptput anymore

### DIFF
--- a/packages/orchestrator/lib/clients/client.ts
+++ b/packages/orchestrator/lib/clients/client.ts
@@ -110,14 +110,14 @@ export class OrchestratorClient {
         const { args, ...rest } = props;
         const schedulingProps = {
             ...rest,
+            timeoutSettingsInSecs: {
+                createdToStarted: 30,
+                startedToCompleted: 15 * 60,
+                heartbeat: 999999 // actions don't need to heartbeat
+            },
             args: {
                 ...args,
-                type: 'action' as const,
-                timeoutSettingsInSecs: {
-                    createdToStarted: 30,
-                    startedToCompleted: 30,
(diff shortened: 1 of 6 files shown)


packages/orchestrator/lib/clients/client.ts
```diff
@@ -110,14 +110,14 @@ export class OrchestratorClient {
         const { args, ...rest } = props;
         const schedulingProps = {
             ...rest,
+            timeoutSettingsInSecs: {
+                createdToStarted: 30,
+                startedToCompleted: 15 * 60,
```
TBonnin, line 115: in addition to moving the timeout out of args (bugfix), setting the startedToCompleted timeout to 15 minutes which is the same as temporal https://github.com/NangoHQ/nango/blob/e943ec3a1aadab30eae14e4e5d398877c7a0548d/packages/jobs/lib/workflows.ts#L8

khaliqgant (approved):

packages/orchestrator/lib/server.ts
```diff
@@ -17,7 +17,7 @@ const logger = getLogger('Orchestrator.server');
 export const getServer = (scheduler: Scheduler, eventEmmiter: EventEmitter): Express => {
     const server = express();
 
-    server.use(express.json({ limit: '100kb' }));
+    server.use(express.json({ limit: '10mb' }));
```
bodinsamuel, line 20: what are they sending this big? 😭

bodinsamuel (approved):

PR #2273, opened 2024-06-05
feat(cli): [nan-1106] import relative files in syncs/actions

khaliqgant:
## Describe your changes
Use tsup to import relative files in syncs. The error reporting with tsup isn't as nice as ts-node so the tsup transpiles only and ts-node is still used to type check and report nice errors. The result is actually faster than previous. tsup doesn't transpile if there are any compilation errors.

Note that an additional `vite.cli.config` file was added which doesn't use threads because it was interfering with `tsup`

### Edge Cases Handled
- [x] If the main file imports another file which in turn imports another file then all imported files should be typed checked and checked to make sure [nango calls are used correctly](https://github.com/NangoHQ/nango/blob/master/packages/cli/lib/services/parser.service.ts#L10)
- [x] If a library is imported like `crypto` we shouldn't try and type check that file
- [x] If a tsconfig.json file exists in their repo it isn't used and the one we want to use in the nango cli directory is instead used
- [x] Limit possible import files to be in the `nango-integrations` directory
- [ ] Update to download the entire nango-integrations directory instead of just the particular file (v2)
- [ ] Verify that changing the transpiled code doesn't produce different results the ts-node transpilation

NAN-1106


packages/cli/lib/tests/helpers.ts
```diff
@@ -0,0 +1,19 @@
+import path from 'path';
+import * as fs from 'fs';
```
bodinsamuel:
```suggestion
import fs from 'fs/promises';
```

vite.cli.config.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export default defineConfig({
+    test: {
+        include: ['**/*.unit.cli-{test,spec}.?(c|m)[jt]s?(x)'],
+        env: {
+            NANGO_ENCRYPTION_KEY: 'RzV4ZGo5RlFKMm0wYWlXdDhxTFhwb3ZrUG5KNGg3TmU=',
+            NANGO_LOGS_ES_URL: 'http://fake.com',
+            NANGO_LOGS_ES_USER: '',
+            NANGO_LOGS_ES_PWD: '',
+            NANGO_LOGS_ENABLED: 'false'
```
bodinsamuel: Unnecessary I think
```suggestion
```

bodinsamuel (commented):
I have this error when running `node ../nango/packages/cli/dist/index.js compile`

```sh
Error compiling "/Users/samuelbodin/code/nango-integrations/unauthenticated/syncs/trackDeletes.ts":
SyntaxError: Unexpected token '', "��z��" is not valid JSON
    at JSON.parse (<anonymous>)
    at new AnyMap (/Users/samuelbodin/code/nango/node_modules/@jridgewell/src/any-map.ts:20:37)
    at mapSourcePosition (/Users/samuelbodin/code/nango/node_modules/@cspotcode/source-map-support/source-map-support.js:387:14)
    at wrapCallSite (/Users/samuelbodin/code/nango/node_modules/@cspotcode/source-map-support/source-map-support.js:592:20)
    at Function.prepareStackTrace (/Users/samuelbodin/code/nango/node_modules/@cspotcode/source-map-support/source-map-support.js:671:41)
    at prepareStackTraceCallback (node:internal/errors:145:29)
    at getStackString (node:internal/util/inspect:1240:16)
    at formatError (node:internal/util/inspect:1369:15)
    at formatRaw (node:internal/util/inspect:985:14)
    at formatValue (node:internal/util/inspect:840:10)
```

packages/cli/lib/tests/helpers.ts
khaliqgant: Updated

vite.cli.config.ts
khaliqgant: Updated

khaliqgant:
> I have this error when running `node ../nango/packages/cli/dist/index.js compile`
> 
> ```shell
> Error compiling "/Users/samuelbodin/code/nango-integrations/unauthenticated/syncs/trackDeletes.ts":
> SyntaxError: Unexpected token '', "��z��" is not valid JSON
>     at JSON.parse (<anonymous>)
>     at new AnyMap (/Users/samuelbodin/code/nango/node_modules/@jridgewell/src/any-map.ts:20:37)
>     at mapSourcePosition (/Users/samuelbodin/code/nango/node_modules/@cspotcode/source-map-support/source-map-support.js:387:14)
>     at wrapCallSite (/Users/samuelbodin/code/nango/node_modules/@cspotcode/source-map-support/source-map-support.js:592:20)
>     at Function.prepareStackTrace (/Users/samuelbodin/code/nango/node_modules/@cspotcode/source-map-support/source-map-support.js:671:41)
>     at prepareStackTraceCallback (node:internal/errors:145:29)
>     at getStackString (node:internal/util/inspect:1240:16)
>     at formatError (node:internal/util/inspect:1369:15)
>     at formatRaw (node:internal/util/inspect:985:14)
>     at formatValue (node:internal/util/inspect:840:10)
> ```

Can you try this again? Should be resolved.

packages/cli/lib/services/parser.service.ts
```diff
@@ -7,14 +7,31 @@ import parser from '@babel/parser';
 import { SyncConfigType } from '@nangohq/shared';
 
 class ParserService {
-    public callsAreUsedCorrectly(filePath: string, type = SyncConfigType.SYNC, modelNames: string[]): boolean {
+    public getImportedFiles(filePath: string): string[] {
         const code = fs.readFileSync(filePath, 'utf-8');
+        const ast = parser.parse(code, { sourceType: 'module', plugins: ['typescript'] });
+        const importedFiles: string[] = [];
+        const traverseFn = (traverse as any).default || traverse;
```
TBonnin, line 14: I know it is unrelated to this PR but I've always wondered why does this `.default` thing is doing and why we are not using traverse directly

packages/cli/lib/services/compile.service.ts
```diff
@@ -86,43 +87,83 @@ export async function compileAllFiles({
...
+        const importedFilePathWithExtension = importedFilePath + '.ts';
+
+        /// if it is a library import then we can skip it
+        if (!fs.existsSync(importedFilePathWithExtension)) {
+            continue;
+        }
+
+        const cwd = process.cwd().split('/').pop();
+        const fullPath = path.resolve(importedFilePathWithExtension);
+        if (cwd === 'nango-integrations' && fullPath.includes(process.cwd())) {
```
TBonnin: shouldn't it be the negative? ie: if fullPath doesn't includes nango-integration then show error?

TBonnin (commented):
I think there is an inverted condition with the nango-integrations directory detection but other than that it seems to work as expected. I haven't tested very involved or big import though

Question about your PR notes: 
> Update to download the entire nango-integrations directory instead of just the particular file (v2)
why will we need to do that?

packages/cli/lib/services/compile.service.ts
khaliqgant: Thanks, this was the last thing I added last night without testing thoroughly. Updated and added a test case to support it + improved the error message

packages/cli/lib/services/parser.service.ts
khaliqgant: I think during tests traverse wasn't available for some reason

khaliqgant:
> I think there is an inverted condition with the nango-integrations directory detection but other than that it seems to work as expected. I haven't tested very involved or big import though
> 
> Question about your PR notes:
> 
> > Update to download the entire nango-integrations directory instead of just the particular file (v2)
> > why will we need to do that?

So a user could download all the code including helper files.

bodinsamuel (approved):
Works great 👏🏻 
I have a question about unintended side-effects, it's now possible to load any npm package inside the build, I don't think it's okay even if it's handy?

khaliqgant:
> Works great 👏🏻 I have a question about unintended side-effects, it's now possible to load any npm package inside the build, I don't think it's okay even if it's handy?

Thanks for catching this! Should be resolved by https://github.com/NangoHQ/nango/pull/2273/commits/19803210c0c6db6c5426b64c6d5c12ae208cefa5

packages/cli/lib/services/compile.service.ts
```diff
@@ -12,6 +13,8 @@ import { TYPES_FILE_NAME } from '../constants.js';
 import modelService from './model.service.js';
 import parserService from './parser.service.js';
 
+const ALLOWED_IMPORTS = ['url', 'crypto', 'zod'];
```
bodinsamuel:
```suggestion
const ALLOWED_IMPORTS = ['url', 'crypto', 'zod', 'node:url', 'node:crypto'];
```

bodinsamuel (approved):
Nice, it now catches the most obvious incorrect import, and execute appropriately.
It's still possible to import node_modules by using relative path (i.e: `import {attemptifyAsync} from '../../node_modules/stubborn-fs/src/attemptify';`). It's a bit harder to catch, especially that stuff could live outside `node_modules`. 

Side note:
- You can't import `.js` file, I'm sure it's not really an issue but it displays the wrong error message in that case. (and it reduces the possibility to import packages)
- Because they can import stuff, they'll need a package.json, so we need to check wether the zod/node version is matching what we are using otherwise the types will be wrong and/or expected execution will be slightly different

packages/cli/lib/services/compile.service.ts
khaliqgant: updated

TBonnin (approved):

khaliqgant:
> Nice, it now catches the most obvious incorrect import, and execute appropriately. It's still possible to import node_modules by using relative path (i.e: `import {attemptifyAsync} from '../../node_modules/stubborn-fs/src/attemptify';`). It's a bit harder to catch, especially that stuff could live outside `node_modules`.

Yeah, hard to catch these edge cases.

> 
> Side note:
> 
> * You can't import `.js` file, I'm sure it's not really an issue but it displays the wrong error message in that case. (and it reduces the possibility to import packages)

Something to revisit in future iterations perhaps.

> * Because they can import stuff, they'll need a package.json, so we need to check wether the zod/node version is matching what we are using otherwise the types will be wrong and/or expected execution will be slightly different

Yeah, we should publish in our documentation what versions of things we're using to make sure it aligns

bburns:
> > Nice, it now catches the most obvious incorrect import, and execute appropriately. It's still possible to import node_modules by using relative path (i.e: `import {attemptifyAsync} from '../../node_modules/stubborn-fs/src/attemptify';`). It's a bit harder to catch, especially that stuff could live outside `node_modules`.
> 
> Yeah, hard to catch these edge cases.

Some example code for avoiding path traversal attacks -

```
const root = '/var/www/'
exports.validatePath = user_input => {
  if (user_input.indexOf('\0') !== -1) {
    return 'Access denied'
  }
  if (!/^[a-z0-9]+$/.test(user_input)) {
    return 'Access denied'
  }
  const path = require('path')
  const safe_input = path.normalize(user_input).replace(/^(\.\.(\/|\\|$))+/, '')
  const path_string = path.join(root, safe_input)
  if (path_string.indexOf(root) !== 0) {
    return 'Access denied'
  }
  return path_string
}
```

From https://www.stackhawk.com/blog/node-js-path-traversal-guide-examples-and-prevention/

PR #2274, opened 2024-06-05
feat(orchestrator): scheduling tasks based on schedules

TBonnin:
This PR:
- adds retry and timeout fields to schedule so they can be passed to the tasks
- adds a scheduling worker that is repeatedly scan the database to find schedules due to run
- adds indexes to tasks and schedules table to speed up the scheduling query

### DIFF
--- a/packages/scheduler/lib/scheduler.ts
+++ b/packages/scheduler/lib/scheduler.ts
@@ -4,16 +4,15 @@ import type { Task, TaskState, Schedule, ScheduleProps, ImmediateProps } from '.
 import * as tasks from './models/tasks.js';
 import * as schedules from './models/schedules.js';
 import type { Result } from '@nangohq/utils';
-import { stringifyError, getLogger } from '@nangohq/utils';
-import { MonitorWorker } from './monitor.worker.js';
+import { stringifyError } from '@nangohq/utils';
+import { MonitorWorker } from './workers/monitor/monitor.worker.js';
+import { SchedulingWorker } from './workers/scheduling/scheduling.worker.js';
 import type { DatabaseClient } from './db/client.js';
-
-const logger = getLogger('Scheduler');
+import { logger } from './utils/logger.js';
 
 export class Scheduler {
-    // TODO: scheduling recurring tasks
-
     private monitor: MonitorWorker | null = null;
+    private scheduling: SchedulingWorker | null = null;
     private onCallbacks: Record<TaskState, (task: Task) => void>;
     private dbClient: DatabaseClient;
 
@@ -50,6 +49,9 @@ export class Scheduler {
                 }
             });
             this.monitor.start();
+            this.scheduling = new SchedulingWorker({ databaseUrl: dbClient.url, databaseSchema: dbClient.schema });
+            // TODO: ensure there is only one instance of the scheduler
+            this.scheduling.start();
         } else {
             throw new Error('Scheduler must be instantiated in the main thread');
         }
@@ -57,6 +59,7 @@ export class Scheduler {
 
     stop(): void {
         this.monitor?.stop();
+        this.scheduling?.stop();
     }
 
     /**
--- a/packages/scheduler/lib/workers/scheduling/scheduling.integration.test.ts
+++ b/packages/scheduler/lib/workers/scheduling/scheduling.integration.test.ts
@@ -1,0 +1,160 @@
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { dueSchedules } from './scheduling.js';
+import { getTestDbClient } from '../../db/helpers.test.js';
+import { DbSchedule, SCHEDULES_TABLE } from '../../models/schedules.js';
+import { uuidv7 } from 'uuidv7';
+import type knex from 'knex';
+import type { Schedule, ScheduleState, Task, TaskState } from '../../types.js';
+import { TASKS_TABLE, DbTask } from '../../models/tasks.js';
+
+describe('dueSchedules', () => {
+    const dbClient = getTestDbClient();
+    const db = dbClient.db;
+
+    beforeEach(async () => {
+        await dbClient.migrate();
+    });
+
+    afterEach(async () => {
+        await dbClient.clearDatabase();
+    });
+
+    it('should not return schedule that is deleted', async () => {
+        await addSchedule(db, { state: 'DELETED', frequency: '3 minutes' });
+        const due = await dueSchedules(db);
+        expect(due.isOk()).toBe(true);
+        console.log(due.unwrap());
+        expect(due.unwrap().length).toBe(0);
+    });
+    it('should not return schedule that is paused', async () => {
+        await addSchedule(db, { state: 'PAUSED', frequency: '3 minutes' });
+        const due = await dueSchedules(db);
+        expect(due.isOk()).toBe(true);
+        console.log(due.unwrap());
+        expect(due.unwrap().length).toBe(0);
+    });
+    it('should not return schedule that is set to start in the future', async () => {
+        await addSchedule(db, { state: 'STARTED', frequency: '3 minutes', startsAt: Seconds.after(1 * 60) });
+        const due = await dueSchedules(db);
+        expect(due.isOk()).toBe(true);
+        console.log(due.unwrap());
+        expect(due.unwrap().length).toBe(0);
+    });
+    it('should not return schedule that have a recent task completed', async () => {
+        const startsAt = Seconds.ago(-5 * 60); // 5 minutes ago
+        const schedule = await addSchedule(db, { startsAt, frequency: '10 minutes' });
+        await addTask(db, {
+            scheduleId: schedule.id,
+            state: 'SUCCEEDED',
+            startsAfter: startsAt,
+            lastStateTransitionAt: Seconds.after(20, startsAt)
+        });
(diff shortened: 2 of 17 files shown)


packages/scheduler/lib/scheduler.ts
```diff
@@ -50,6 +49,9 @@ export class Scheduler {
                 }
             });
             this.monitor.start();
+            this.scheduling = new SchedulingWorker({ databaseUrl: dbClient.url, databaseSchema: dbClient.schema });
+            // TODO: ensure there is only one instance of the scheduler
```
TBonnin, line 53: It is on my list to have a simple distributed lease mechanism to ensure there is always one single scheduling worker. For now it is enforced by the fact that orchestrator service is set to 1 instance and doesn't autoscale.

packages/scheduler/lib/workers/scheduling/scheduling.ts
```diff
@@ -0,0 +1,49 @@
...
+                    .from({ t: TASKS_TABLE })
+                    .whereRaw('t.schedule_id = s.id')
+                    .where(function () {
+                        this.where({ state: 'CREATED' }).orWhere({ state: 'STARTED' });
+                    })
+            )
+            // filter out schedules that have tasks started after the last run time
+            .whereNotExists(
+                db.select('id').from({ t: TASKS_TABLE }).whereRaw('t.schedule_id = s.id').andWhere('t.starts_after', '>=', db.raw('lrt.last_run_time'))
+            );
```
TBonnin: This last part is critical because the number of tasks for a given schedule will grow overtime. We probably are gonna need to archive completed tasks at some point

packages/scheduler/lib/workers/scheduling/scheduling.worker.ts
```diff
@@ -0,0 +1,113 @@
...
+                    scheduleId: schedule.id,
+                    startsAfter: new Date(),
+                    name: `${schedule.name}:${new Date().toISOString()}`,
+                    payload: schedule.payload,
+                    groupKey: schedule.groupKey,
+                    retryCount: schedule.retryCount,
+                    retryMax: schedule.retryMax,
+                    createdToStartedTimeoutSecs: schedule.createdToStartedTimeoutSecs,
+                    startedToCompletedTimeoutSecs: schedule.startedToCompletedTimeoutSecs,
+                    heartbeatTimeoutSecs: schedule.heartbeatTimeoutSecs
```
TBonnin, line 104: all the new fields were added to be used here when creating the task

khaliqgant (approved):
Looks good. I find the query a big hard to read with aliases like `s` and `lrt` but not blocking feedback. `console.log` is left over in `packages/scheduler/lib/workers/scheduling/scheduling.ts`

Would like to test scheduling locally and interact with it. Easiest way to get setup?

packages/scheduler/lib/workers/scheduling/scheduling.integration.test.ts
```diff
@@ -0,0 +1,160 @@
...
+
+    afterEach(async () => {
+        await dbClient.clearDatabase();
+    });
+
+    it('should not return schedule that is deleted', async () => {
+        await addSchedule(db, { state: 'DELETED', frequency: '3 minutes' });
+        const due = await dueSchedules(db);
+        expect(due.isOk()).toBe(true);
+        console.log(due.unwrap());
```
bodinsamuel:
```suggestion
```

bodinsamuel: and all after

packages/scheduler/lib/workers/scheduling/scheduling.ts
```diff
@@ -0,0 +1,48 @@
+import type knex from 'knex';
+import type { Schedule } from '../../types.js';
+import { DbSchedule, SCHEDULES_TABLE } from '../../models/schedules.js';
+import type { Result } from '@nangohq/utils';
+import { Err, Ok, stringifyError } from '@nangohq/utils';
+import { TASKS_TABLE } from '../../models/tasks.js';
+
+export async function dueSchedules(db: knex.Knex): Promise<Result<Schedule[]>> {
+    try {
+        const query = db
```
bodinsamuel, line 10: A bit hard to read, at this point not sure knex is helping especially that you don't have params. 
And the table's aliases are too short :D

packages/scheduler/lib/workers/scheduling/scheduling.ts
```diff
@@ -0,0 +1,48 @@
...
+            .whereNotExists(
+                db
+                    .select('id')
+                    .from({ t: TASKS_TABLE })
+                    .whereRaw('t.schedule_id = s.id')
+                    .where(function () {
+                        this.where({ state: 'CREATED' }).orWhere({ state: 'STARTED' });
+                    })
+            )
+            // filter out schedules that have tasks started after the last run time
```
bodinsamuel: how is it possible?
edit: okay I think it's because last_run_time is computed not real, right?

packages/scheduler/lib/workers/scheduling/scheduling.ts
```diff
@@ -0,0 +1,48 @@
...
+                        `)
+                    )
+                    .from({ s: SCHEDULES_TABLE })
+                    .where({ state: 'STARTED' })
+                    .whereRaw('s.starts_at <= NOW()')
+            )
+            .select('*')
+            .from<DbSchedule>({ s: SCHEDULES_TABLE })
+            .joinRaw('JOIN last_run_times lrt ON s.id = lrt.id')
+            // filter out schedules that have a running task
```
bodinsamuel, line 28: Might be easier to store the state either on the same table or on a dedicated table instead of joining a large table to compute a state

packages/scheduler/lib/workers/scheduling/scheduling.ts
```diff
@@ -0,0 +1,48 @@
...
+    try {
+        const query = db
+            .with(
+                'last_run_times',
+                // calculate the last run time for each schedule that is started/not deleted
+                db
+                    .select(
+                        's.id',
+                        db.raw(`
+                            s.starts_at + (FLOOR(EXTRACT(EPOCH FROM (NOW() - s.starts_at)) / EXTRACT(EPOCH FROM s.frequency)) * s.frequency) AS last_run_time
```
bodinsamuel: Would it be easier to just store this when you run a task? at least it would reflect actual usage not idealistic past

bodinsamuel (commented):
Looks okay, I would challenge the way we compute all of those states, as much for ease of use than reporting.

packages/scheduler/lib/workers/scheduling/scheduling.ts
TBonnin: To be honest I found knex syntax to get in the way when queries are not straightforward. Maybe it is just me but I found SQL quite nice to read in general. I am happy to rewrite this one in rawSql if you prefer

packages/scheduler/lib/workers/scheduling/scheduling.ts
TBonnin: 1. it is supposed to be the expected due time, not the actual one. Let me rename the CTE and field to make it clearer
2. There is a tradeoff between runtime vs pre-computing: runtime is more expensive but with pre-computing you need to ensure the pre-computation always happen successfully and potential mutation error must be handled. I would prefer to stick with runtime calculation until we hit scaling issue.

packages/scheduler/lib/workers/scheduling/scheduling.ts
TBonnin: See my comment above. Let's see how fast it it. We'll add another state (and all the complexity of maintaining it) when we hit performance issue

packages/scheduler/lib/workers/scheduling/scheduling.ts
TBonnin: I think my naming choice was bad. By last_run_time I meant last_due_time. I'll rename it

TBonnin: renamed the CTE to `due_dates` and the field to `dueAt`. https://github.com/NangoHQ/nango/pull/2274/commits/71869458c000ce867f78a5f71b7be7f22d873414 
Let me know if you have any other comments

packages/scheduler/lib/workers/scheduling/scheduling.ts
bodinsamuel: I don't prefer, I actually think it's a topic that just "depends"😁 
Query builders are very nice when you have external parameters and optional composability, but not that nice with complex/nested queries. I always prefer consistency and if possible "one rule", but regarding sql I think maintainability should always be the main driver

TBonnin: I am gonna keep it as is for now. It is actually not horrible with knex (SQL would probably be a bit more verbose and be without syntax highlighting) and at least it is consistent

bodinsamuel (approved):
Looks good, still nitpicking about the table's alias in the query otherwise 👌🏻

TBonnin:
> Looks good, still nitpicking about the table's alias in the query otherwise 👌🏻

We are gonna need to agree to disagree here. I personally find long name cumbersome, not knowing if they refer to a column/table/etc... 😀

PR #2277, opened 2024-06-06
fix: return error message when action fails

TBonnin:
we were not returning the error message just the payload. The error response can be further improved but for now I am trying to figure out what the actual error is

### DIFF
--- a/packages/shared/lib/clients/orchestrator.ts
+++ b/packages/shared/lib/clients/orchestrator.ts
@@ -114,7 +114,8 @@ export class Orchestrator {
                         groupKey,
                         args
                     });
-                    res = actionResult.mapError((e) => new NangoError('action_failure', e.payload ?? { error: e.message }));
+
+                    res = actionResult.mapError((e) => new NangoError('action_failure', { error: e.message, ...(e.payload ? { payload: e.payload } : {}) }));
                     if (res.isErr()) {
                         span.setTag('error', res.error);
                     }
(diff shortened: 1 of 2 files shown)


packages/shared/lib/utils/error.ts
```diff
@@ -254,7 +254,7 @@ export class NangoError extends Error {
                 this.status = 400;
                 this.message = `Failed to perform the action. Please try again.`;
                 if (this.payload) {
-                    this.message += ` Please make sure this value exists in the Nango dashboard ${JSON.stringify(this.payload, null, 2)}`;
```
TBonnin, line 257: not sure what this was coming from or useful for

bodinsamuel: :D

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -114,7 +114,8 @@ export class Orchestrator {
                         groupKey,
                         args
                     });
-                    res = actionResult.mapError((e) => new NangoError('action_failure', e.payload ?? { error: e.message }));
+
+                    res = actionResult.mapError((e) => new NangoError('action_failure', { error: e.message, ...(e.payload ? { payload: e.payload } : {}) }));
```
bodinsamuel, line 118: Can't wait to refacto NangoError

bodinsamuel (approved):

PR #2278, opened 2024-06-06
fix: actions/webhooks should always report failure when failing

TBonnin:
There were 2 error handling path in actions/webhooks execution logic. One of this path was not setting logs and telemetry correctly. This commit removes the early return and throws the error so it can be handled by the try catch

### DIFF
--- a/packages/shared/lib/clients/orchestrator.ts
+++ b/packages/shared/lib/clients/orchestrator.ts
@@ -173,23 +173,7 @@ export class Orchestrator {
             }
 
             if (res.isErr()) {
-                await createActivityLogMessageAndEnd({
-                    level: 'error',
-                    environment_id,
-                    activity_log_id: activityLogId,
-                    timestamp: Date.now(),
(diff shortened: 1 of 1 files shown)


packages/shared/lib/clients/orchestrator.ts
```diff
@@ -173,23 +173,7 @@ export class Orchestrator {
...
-                await createActivityLogMessageAndEnd({
-                    level: 'error',
-                    environment_id,
-                    activity_log_id: activityLogId,
-                    timestamp: Date.now(),
-                    content: `The action workflow ${workflowId} did not complete successfully`
-                });
-                await logCtx.error(`The action workflow ${workflowId} did not complete successfully`);
-                return res;
+                throw res.error;
```
TBonnin, line 176: logging would be taken care by the try catch https://github.com/NangoHQ/nango/pull/2278/files#diff-485470c374a0419d6a217c25b8d8dcee97ad5959e068630d075b77a1d7bc423fL222-L234

bodinsamuel (approved):

PR #2280, opened 2024-06-06
fix(dockerfile): persist use unified Dockerfile

bodinsamuel:
## Describe your changes

Fixes NAN-607

Finally have time to make progress on this. This one is pretty much straightforward, the only trick is that render do not allow to change the image name with the API so when this merge, I'll need to manually change it in production once.
Staging is actually Running the new image if you want to test.

- Persist: use unified Dockerfile

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -91,7 +91,11 @@ jobs:
             - name: Deploy persist
               run: |
                   SERVICE_ID=${{ fromJson('{ "production": "srv-cmt150ol6cac73apstq0", "staging": "srv-cmsfiqqcn0vc73bhcod0" }')[inputs.stage] }}
-                  curl -sS -f --request POST "https://api.render.com/v1/services/$SERVICE_ID/deploys" --header "authorization: Bearer ${{ secrets.RENDER_API_KEY }}" --data "{ \"imageUrl\": \"docker.io/nangohq/nango:$IMAGE_TAG_PREFIX-${{ github.sha }}\" }"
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yaml
```diff
@@ -88,13 +88,16 @@ jobs:
...
               run: |
-                  SERVICE_ID=${{ fromJson('{ production: "srv-cmt150ol6cac73apstq0", staging: "srv-cmsfiqqcn0vc73bhcod0" }')[inputs.stage] }}
-                  curl --request POST "https://api.render.com/v1/services/$SERVICE_ID/deploys" --header "authorization: Bearer ${{ secrets.RENDER_API_KEY }}"
+                  SERVICE_ID=${{ fromJson('{ "production": "srv-cmt150ol6cac73apstq0", "staging": "srv-cmsfiqqcn0vc73bhcod0" }')[inputs.stage] }}
+                  IMAGE_TAG_PREFIX=${{ fromJson('{ "production": "prod", "staging": "staging" }')[inputs.stage] }}
+                  IMAGE="docker.io/nangohq/nango:${IMAGE_TAG_PREFIX}-${{ github.sha }}"
+
+                  echo "Deploying $IMAGE to $SERVICE_ID"
+
+                  curl -sS --fail-with-body --request POST "https://api.render.com/v1/services/$SERVICE_ID/deploys" --header "authorization: Bearer ${{ secrets.RENDER_API_KEY }}" --data "{ \"imageUrl\": \"$IMAGE\" }"
```
TBonnin, line 99: TIL `--fail-with-body` 👍

TBonnin (approved):

.github/workflows/deploy.yaml
bodinsamuel: yes I also discovered that today :D

PR #2281, opened 2024-06-06
feat(webhooks): [nan 1064] webhook on sync error

khaliqgant:
## Describe your changes
Add in logic to send webhooks on sync failure

NAN-1064


packages/webhooks/lib/sync.unit.test.ts
```diff
@@ -0,0 +1,153 @@
+/* eslint-disable @typescript-eslint/unbound-method */
```
khaliqgant, line 1: Copied over from shared

packages/webhooks/lib/sync.ts
```diff
@@ -0,0 +1,101 @@
...
+import dayjs from 'dayjs';
+import utc from 'dayjs/plugin/utc.js';
+import type { LogContext } from '@nangohq/logs';
+import type { NangoSyncWebhookBody } from './types.js';
+import { deliver, shouldSend } from './utils.js';
+import { WebhookType } from './enums.js';
+
+dayjs.extend(utc);
+
+export const sendSync = async ({
```
khaliqgant, line 11: Logic essentially copied over from shared

packages/shared/lib/services/sync/run.service.ts
```diff
@@ -103,6 +106,7 @@ export default class SyncRun {
     recordsService: RecordsServiceInterface;
     dryRunService?: NangoProps['dryRunService'];
     slackNotificationService?: SlackService;
+    sendSyncWebhook?: (params: SendSyncParams) => Promise<void>;
```
TBonnin, line 109: I think I am gonna propose to Bastien a `refactor run.service` project because it has become a mammoth class that does everything. Its `run` function is 500+ lines I think

packages/shared/lib/services/notification/webhook.service.ts
```diff
@@ -4,12 +4,10 @@ import dayjs from 'dayjs';
 import utc from 'dayjs/plugin/utc.js';
```
TBonnin, line 4: what still live in this file? Is everything moved over to the webhooks package?

packages/types/lib/api.ts
```diff
@@ -76,3 +76,8 @@ export interface Endpoint<T extends EndpointDefinition> {
      */
     Reply: ResDefaultErrors | (T['Error'] extends ApiError<any> ? T['Error'] | T['Success'] : T['Success']);
 }
+
+export interface ErrorPayload {
+    type: string;
+    description: string;
+}
```
TBonnin, line 83: Is this an api related type?

packages/shared/lib/services/notification/webhook.service.ts
khaliqgant: webhook forwarding logic which I'll bring over in another PR.

packages/types/lib/api.ts
khaliqgant: internal API I would say

packages/shared/lib/services/sync/run.service.ts
khaliqgant: Yeah, it is a beast. Definitely needed.

packages/types/lib/api.ts
bodinsamuel: if we can keep this file for http api, maybe rename the file if clearer 👌🏻

packages/shared/lib/services/sync/run.service.ts
```diff
@@ -420,11 +461,10 @@ export default class SyncRun {
             }
 
             const startTime = Date.now();
+            const syncStartDate = new Date();
```
bodinsamuel, line 464: Could be moved at the very beginning of the method or could be a property of the class directly

bodinsamuel (approved):
🦾 Minor comments

TBonnin (approved):

packages/shared/lib/services/sync/run.service.ts
khaliqgant: Will make a note to follow up on this later

PR #2283, opened 2024-06-07
chore(orchestrator): add endpoint to api/client to run a schedule

TBonnin:
Adding the endpoint in orchestrator api/client to create a schedule run, that will be used when manually triggering a sync

This commit also contains some preliminary work to add sync support to the orchestrator (attempt attribute, unique name...) + some TODOs. 
Sorry for squeezing several changes in a single PR/commit. I started to add support for sync in the orchestrator but it is quite a big change so I paused and submitting this partial PR

### DIFF
--- a/packages/jobs/lib/processor/handler.ts
+++ b/packages/jobs/lib/processor/handler.ts
@@ -67,10 +67,10 @@ async function action(task: TaskAction): Promise<Result<JsonValue>> {
         return Err(error);
     }
     const res = jsonSchema.safeParse(response);
-    if (res.success) {
-        return Ok(res.data);
+    if (!res.success) {
+        return Err(`Invalid action response format: ${response}. Action: ${JSON.stringify(task)}`);
     }
-    return Err(`Invalid action response format: ${response}. Action: ${JSON.stringify(task)}`);
+    return Ok(res.data);
 }
 
 async function webhook(task: TaskWebhook): Promise<Result<JsonValue>> {
@@ -111,10 +111,10 @@ async function webhook(task: TaskWebhook): Promise<Result<JsonValue>> {
         return Err(error);
     }
     const res = jsonSchema.safeParse(response);
-    if (res.success) {
-        return Ok(res.data);
+    if (!res.success) {
+        return Err(`Invalid webhook response format: ${response}. Webhook: ${JSON.stringify(task)}`);
     }
-    return Err(`Invalid webhook response format: ${response}. Webhook: ${JSON.stringify(task)}`);
+    return Ok(res.data);
 }
 
 async function postConnection(task: TaskPostConnection): Promise<Result<JsonValue>> {
--- a/packages/jobs/lib/processor/handler.ts
+++ b/packages/jobs/lib/processor/handler.ts
@@ -68,7 +68,7 @@ async function action(task: TaskAction): Promise<Result<JsonValue>> {
     }
     const res = jsonSchema.safeParse(response);
     if (!res.success) {
-        return Err(`Invalid action response format: ${response}. Action: ${JSON.stringify(task)}`);
+        return Err(`Invalid action response format: ${response}. TaskId: ${task.id}`);
     }
     return Ok(res.data);
 }
@@ -112,7 +112,7 @@ async function webhook(task: TaskWebhook): Promise<Result<JsonValue>> {
     }
     const res = jsonSchema.safeParse(response);
     if (!res.success) {
-        return Err(`Invalid webhook response format: ${response}. Webhook: ${JSON.stringify(task)}`);
+        return Err(`Invalid webhook response format: ${response}. TaskId: ${task.id}`);
     }
     return Ok(res.data);
 }
--- a/packages/orchestrator/lib/clients/client.ts
+++ b/packages/orchestrator/lib/clients/client.ts
@@ -1,5 +1,6 @@
 import { route as postImmediateRoute } from '../routes/v1/postImmediate.js';
 import { route as postRecurringRoute } from '../routes/v1/postRecurring.js';
+import { route as postRecurringRunRoute } from '../routes/v1/recurring/postRecurringRun.js';
 import { route as postDequeueRoute } from '../routes/v1/postDequeue.js';
 import { route as postSearchRoute } from '../routes/v1/postSearch.js';
 import { route as getOutputRoute } from '../routes/v1/tasks/taskId/getOutput.js';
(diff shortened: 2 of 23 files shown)


packages/shared/lib/clients/orchestrator.ts
```diff
@@ -597,25 +611,7 @@ export class Orchestrator {
...
-                    environment_id: connection.environment_id,
-                    activity_log_id: activityLogId,
-                    timestamp: Date.now(),
-                    content: `The post connection script workflow ${workflowId} did not complete successfully`
-                });
-                await logCtx.error(`The post connection script workflow ${workflowId} did not complete successfully`);
-                await logCtx.failed();
-
-                return res;
+                throw res.error;
```
TBonnin, line 588: forgot this one in my fix this morning https://github.com/NangoHQ/nango/pull/2278

packages/jobs/lib/processor/handler.ts
```diff
@@ -65,10 +65,10 @@ async function action(task: TaskAction): Promise<Result<JsonValue>> {
         return Err(error);
     }
     const res = jsonSchema.safeParse(response);
-    if (res.success) {
-        return Ok(res.data);
+    if (!res.success) {
+        return Err(`Invalid action response format: ${response}. Action: ${JSON.stringify(task)}`);
```
bodinsamuel: is it safe to stringify the task/reponse in term of credentials and such?

packages/orchestrator/lib/clients/client.ts
```diff
@@ -79,6 +85,50 @@ export class OrchestratorClient {
...
+        const res = await this.routeFetch(postRecurringRunRoute)({
+            body: {
+                scheduleName: props.scheduleName
+            }
+        });
+        if ('error' in res) {
+            return Err({
+                name: res.error.code,
+                message: res.error.message || `Error creating recurring schedule`,
+                payload: JSON.stringify(props)
```
bodinsamuel: any reason to always stringify payload? feels like it's what give us the \\\\\\" in the recent errors we had

packages/scheduler/lib/db/migrations/20240606211412_unique_name.ts
```diff
@@ -0,0 +1,25 @@
...
+import { TASKS_TABLE } from '../../models/tasks.js';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.raw(`
+            ALTER TABLE ${SCHEDULES_TABLE}
+            ADD CONSTRAINT tasks_unique_name UNIQUE (name);
+        `);
+    await knex.raw(`
+            ALTER TABLE ${TASKS_TABLE}
+            ADD CONSTRAINT schedules_unique_name UNIQUE (name);
```
bodinsamuel: schedules on task table?

packages/scheduler/lib/db/helpers.test.ts
```diff
@@ -5,3 +5,12 @@ export const getTestDbClient = () =>
...
     });
+
+export const rndStr = (length: number = 5) => {
+    const chars = '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ';
+    let result = '';
+    for (let i = 0; i < length; i++) {
+        result += chars.charAt(Math.floor(Math.random() * chars.length));
+    }
+    return result;
+};
```
bodinsamuel: nanoid is available in utils

packages/orchestrator/lib/routes/v1/recurring/postRecurringRun.ts
```diff
@@ -0,0 +1,46 @@
...
+    Body: {
+        scheduleName: string;
+    };
+    Error: ApiError<'recurring_run_failed'>;
+    Success: { scheduleId: string };
+}>;
+
+const validate = validateRequest<PostRecurringRun>({
+    parseBody: (data: any) => {
+        return z.object({ scheduleName: z.string().min(1) }).parse(data);
```
bodinsamuel: I just noticed you are not using `strict()` in your zod objects validation. Would be great to add it, extra params can have insidious consequences when you spread object or else

packages/scheduler/lib/scheduler.ts
```diff
@@ -103,16 +100,50 @@ export class Scheduler {
...
      * };
      * const scheduled = await scheduler.immediate(schedulingProps);
      */
-    public async immediate(props: ImmediateProps): Promise<Result<Task>> {
-        const taskProps = {
-            ...props,
-            startsAfter: new Date(),
-            scheduleId: null
-        };
+    public async immediate(props: ImmediateProps | { scheduleName: string }): Promise<Result<Task>> {
```
bodinsamuel, line 106: this function should run in a transaction

bodinsamuel (commented):
Looks good 🪡
Minor comments

packages/jobs/lib/processor/handler.ts
TBonnin: You are right. It contains the action input which contains customers' data. I will change to return the id only. That's the minimum for debugging

packages/orchestrator/lib/clients/client.ts
TBonnin: You are right. no need to stringify

packages/orchestrator/lib/routes/v1/recurring/postRecurringRun.ts
TBonnin: I will add strict

packages/scheduler/lib/db/migrations/20240606211412_unique_name.ts
TBonnin: oups I inverted the names

bodinsamuel: you fixed the down but not the up

TBonnin (commented):
@bodinsamuel I think I've addressed your comments. Thank you very much for the review

packages/scheduler/lib/scheduler.ts
TBonnin: not sure why. there is only a single query that is inserting data.

bodinsamuel: You get search the schedule, then search for a task then insert the task.
A concurrent call could invalidate the schedule or that there is no pending task

TBonnin: I don't think a transaction would prevent those scenarios.
Those are valid concerns though, especially if a task is scheduled concurrently. I will need to address it somehow

bodinsamuel: the transaction should lock the rows and execute two concurrent calls to immediate in a row instead of together, but maybe only if you specify FOR UPDATE. Might be more easily solved with a shared lock indeed. 
Also just thinking out loud, a schedule deletion should cancel the running tasks if possible

bodinsamuel (approved):

packages/scheduler/lib/scheduler.ts
TBonnin: wrote down a note to address concurrent task scheduling. 
For the schedule deletion, yes I will do that when implementing schedule deletion

PR #2284, opened 2024-06-07
fix(webapp): [nan-1147] remove refresh option on refresh token

khaliqgant:
## Describe your changes
Remove refresh button on the refresh token field because you can't get a new refresh token by clicking the button so it is misleading. Only the refresh should be on the access token because you can retrieve a new access token.

NAN-1147


bodinsamuel (approved):
Looks okay but it's missing the why in the context 🤔

khaliqgant:
> Looks okay but it's missing the why in the context 🤔

Updated

PR #2288, opened 2024-06-07
fix(logs): missing env, sort query params

bodinsamuel:
## Describe your changes

Fixes NAN-1152

- Missing env in some URL
I always use dev so totally missed that
- Sort the query params to avoid unnecessary URL change
- Compare query params to only set what's is not the default
- Fix query param `connections` not being used properly

### DIFF
--- a/packages/webapp/src/pages/Logs/Search.tsx
+++ b/packages/webapp/src/pages/Logs/Search.tsx
@@ -29,7 +29,7 @@ import type {
 import Spinner from '../../components/ui/Spinner';
 // import { Input } from '../../components/ui/input/Input';
 // import { MagnifyingGlassIcon } from '@radix-ui/react-icons';
-import { formatQuantity } from '../../utils/utils';
+import { formatQuantity, stringArrayEqual } from '../../utils/utils';
 import { Link, useSearchParams } from 'react-router-dom';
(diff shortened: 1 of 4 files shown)


packages/webapp/src/pages/Logs/Search.tsx
```diff
@@ -181,27 +189,44 @@ export const LogsSearch: React.FC = () => {
...
+            if (integrations.length > 0 && JSON.stringify(integrations) !== JSON.stringify(integrationsDefaultOptions)) {
+                tmp.set('integrations', integrations as any);
+            }
+            if (connections.length > 0 && JSON.stringify(connections) !== JSON.stringify(connectionsDefaultOptions)) {
+                tmp.set('connections', connections as any);
+            }
+            if (syncs.length > 0 && JSON.stringify(syncs) !== JSON.stringify(syncsDefaultOptions)) {
+                tmp.set('syncs', syncs as any);
+            }
+            if (types.length > 0 && JSON.stringify(types) !== JSON.stringify(typesDefaultOptions)) {
```
TBonnin: does the string comparison only work if the param arrays are sorted the same way?

bodinsamuel: good point, will change that

khaliqgant (approved):

PR #2289, opened 2024-06-07
feat(orchestrator): pause/unpause/delete sync

TBonnin:
Adding way to pause/unpause/delete a schedule in the orchestrator
Also allowing to create a schedule in paused state. This is going to be required to support sync auto_start

### DIFF
--- a/packages/orchestrator/lib/clients/client.ts
+++ b/packages/orchestrator/lib/clients/client.ts
@@ -1,5 +1,6 @@
 import { route as postImmediateRoute } from '../routes/v1/postImmediate.js';
 import { route as postRecurringRoute } from '../routes/v1/postRecurring.js';
+import { route as putRecurringRoute } from '../routes/v1/putRecurring.js';
 import { route as postRecurringRunRoute } from '../routes/v1/recurring/postRecurringRun.js';
 import { route as postDequeueRoute } from '../routes/v1/postDequeue.js';
 import { route as postSearchRoute } from '../routes/v1/postSearch.js';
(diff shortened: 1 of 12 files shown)


packages/scheduler/lib/models/schedules.ts
```diff
@@ -15,7 +15,8 @@ interface ScheduleStateTransition {
 export const validScheduleStateTransitions = [
     { from: 'STARTED', to: 'PAUSED' },
     { from: 'STARTED', to: 'DELETED' },
-    { from: 'PAUSED', to: 'STARTED' }
+    { from: 'PAUSED', to: 'STARTED' },
+    { from: 'PAUSED', to: 'DELETED' }
```
TBonnin, line 19: Missed state transition. Deleting a paused schedule should be allowed

packages/orchestrator/lib/clients/client.ts
```diff
@@ -344,7 +370,6 @@ export class OrchestratorClient {
         }
     }
 
-    //TODO: rename to cancelTask?
```
TBonnin, line 347: I don't think we are gonna need a cancelSync.

khaliqgant: Can you explain what you mean by this exactly?

khaliqgant (approved):
Question inline

packages/orchestrator/lib/clients/client.ts
TBonnin: The `cancel` function is cancelling a task (aka: a sync execution). Yesterday I though we might need to have a `cancelSchedule` so I added a TODO to rename the existing function to `cancelTask`. But it happens that we don't cancel schedule (which are only paused/unpaused or deleted), only tasks/jobs. So keeping the name `cancel` for the function is ok for now

packages/orchestrator/lib/routes/v1/putRecurring.ts
```diff
@@ -0,0 +1,52 @@
...
+});
+
+const handler = (scheduler: Scheduler) => {
+    return async (req: EndpointRequest<PostRecurring>, res: EndpointResponse<PostRecurring>) => {
+        const { scheduleName, state } = req.body;
+        const schedule = await scheduler.setScheduleState({ scheduleName, state });
+        if (schedule.isErr()) {
+            return res.status(500).json({ error: { code: 'recurring_failed', message: schedule.error.message } });
+        }
+        return res.status(201).json({ scheduleId: schedule.value.id });
```
bodinsamuel: nitpick: 200 might be more accurate

bodinsamuel (commented):
🎷

PR #2294, opened 2024-06-10
fix(logs): match preset from date range

bodinsamuel:
## Describe your changes

Fixes NAN-1132

- Automatically match preset from date range when accessing the UI with a URL
We always store the period and compute diff to see if one diff match a preset

- fix: correctly set a default range at first load


packages/webapp/src/utils/logs.tsx
```diff
@@ -31,13 +32,72 @@ export function getLogsUrl(
...
+    return { from, to };
+}
+
+// Define presets
+export const presets = [
+    { name: 'last5m', label: 'Last 5 minutes' },
+    { name: 'last1h', label: 'Last hour' },
+    { name: 'last24h', label: 'Last 24 hours' },
+    { name: 'last3', label: 'Last 3 days' },
+    { name: 'last7', label: 'Last 7 days' },
```
khaliqgant: Would prefer everything has a unit

khaliqgant (approved):
Small nit

packages/webapp/src/utils/logs.tsx
bodinsamuel: good point

PR #2296, opened 2024-06-10
fix: prevent concurrent scheduling of tasks

TBonnin:
Addressing you comment from last week @bodinsamuel about concurrent scheduling of tasks

Tasks based on schedules can be scheduled in 3 different ways:
- the scheduling worker
- manual triggering of a sync
- pausing/deleting a sync 
- 
- We are now locking the schedule rows if necessary to ensure no concurrent scheduling logic happens that could lead to race conditions and tasks being scheduled concurrently by 2 different code paths

### DIFF
--- a/packages/scheduler/lib/scheduler.ts
+++ b/packages/scheduler/lib/scheduler.ts
@@ -104,52 +104,55 @@ export class Scheduler {
      * const scheduled = await scheduler.immediate(schedulingProps);
      */
     public async immediate(props: ImmediateProps | { scheduleName: string }): Promise<Result<Task>> {
-        let taskProps: tasks.TaskProps;
-        if ('scheduleName' in props) {
-            const getSchedules = await schedules.search(this.dbClient.db, { name: props.scheduleName, limit: 1 });
-            if (getSchedules.isErr()) {
-                return Err(getSchedules.error);
+        return this.dbClient.db.transaction(async (trx) => {
+            let taskProps: tasks.TaskProps;
+            if ('scheduleName' in props) {
+                // forUpdate = true so that the schedule is locked to prevent any concurrent update or concurrent scheduling of tasks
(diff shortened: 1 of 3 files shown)


packages/scheduler/lib/scheduler.ts
```diff
@@ -104,52 +104,55 @@ export class Scheduler {
      * const scheduled = await scheduler.immediate(schedulingProps);
      */
     public async immediate(props: ImmediateProps | { scheduleName: string }): Promise<Result<Task>> {
-        let taskProps: tasks.TaskProps;
```
TBonnin, line 107: diff is really confusing here. The change is just wrapping everything in a transaction and using `forUpdate` when calling `schedules.search`

packages/scheduler/lib/scheduler.ts
```diff
@@ -286,7 +289,8 @@ export class Scheduler {
      */
     public async setScheduleState({ scheduleName, state }: { scheduleName: string; state: ScheduleState }): Promise<Result<Schedule>> {
         return this.dbClient.db.transaction(async (trx) => {
-            const schedule = await schedules.search(trx, { name: scheduleName, limit: 1 });
+            // forUpdate = true so that the schedule is locked to prevent any concurrent update or concurrent scheduling of tasks
+            const schedule = await schedules.search(trx, { name: scheduleName, limit: 1, forUpdate: true });
```
TBonnin, line 293: not using `skipLocked` here and above because it would fail the request. Therefore we just wait for the schedule to be unlocked. The drawback is that it might makes those calls slower

packages/scheduler/lib/scheduler.ts
bodinsamuel: the trick is to hideWhitespaces changes, not sure why it's not the default
Screenshot 2024-06-10 at 17 33 22

bodinsamuel (commented):
Thanks 🚜

bodinsamuel (approved):

PR #2297, opened 2024-06-10
fix(cli): use absolute path

bodinsamuel:
## Describe your changes

Fixes NAN-1174

- Use absolute path everywhere 
Using relative path is making the testing quite cumbersome and unreliable. Passing `fullPath` makes things a bit more idem potent and less surprising. Basically we call `process.cwd()` only in `index.ts` and the rest of the functions are taking this param as base path.

- Updated the unit test with temporary folders
They were all compiling to the same folder making debugging slightly annoying and adding potential conflict.

## Test

I have tested locally but if you can try on your setup that would be really helpful since it's touching everything.
```sh
node packages/cli/dist/index.js dev
node packages/cli/dist/index.js compile
node packages/cli/dist/index.js dryrun
node packages/cli/dist/index.js deploy dev
```


packages/cli/lib/services/compile.service.ts
```diff
@@ -206,16 +231,23 @@ async function compile({
...
+
     await build({
         entryPoints: [file.inputPath],
-        tsconfig: getNangoRootPath() + '/tsconfig.dev.json',
+        tsconfig: path.join(getNangoRootPath()!, 'tsconfig.dev.json'),
         skipNodeModulesBundle: true,
         silent: !debug,
-        outDir: dirname,
+        outDir: path.join(fullPath, 'dist'),
+        outExtension: () => ({ js: '.js' }),
```
bodinsamuel, line 244: I realized when testing more deeply that if you had a package.json with type:module it would compile to cjs instead so I'm forcing the extension

TBonnin (approved):
lgtm. Tested locally also and everything worked

khaliqgant:
I’ll try and test this today

khaliqgant (approved):
Tested and works as expected.

PR #2298, opened 2024-06-10
feat(orchestrator): add update frequency

TBonnin:


packages/orchestrator/lib/routes/v1/putRecurring.ts
```diff
@@ -7,43 +7,59 @@ import { validateRequest } from '@nangohq/utils';
...
 const method = 'PUT';
 
-export type PostRecurring = Endpoint<{
+export type PutRecurring = Endpoint<{
     Method: typeof method;
     Path: typeof path;
     Body: {
-        scheduleName: string;
-        state: 'STARTED' | 'PAUSED' | 'DELETED';
+        schedule: { name: string; state: 'STARTED' | 'PAUSED' | 'DELETED' } | { name: string; frequencyMs: number };
```
TBonnin, line 14: We currently don't update frequency and state at the same time so making the endpoint a disjoint union type because it is simpler to implement. We will refactor if updating a schedule requires to change multiple attributes at the same time

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -669,6 +677,53 @@ export class Orchestrator {
...
+        syncId: string;
+        interval: string;
+        syncName: string;
+        environmentId: number;
+        activityLogId?: number;
+        logCtx?: LogContext;
+    }): Promise<Result<void>> {
+        const isGloballyEnabled = await featureFlags.isEnabled('orchestrator:schedule', 'global', false);
+        const isEnvEnabled = await featureFlags.isEnabled('orchestrator:schedule', `${environmentId}`, false);
+        const isOrchestrator = isGloballyEnabled || isEnvEnabled;
```
TBonnin, line 697: we do update in both temporal and orchestrator and return the result based on the flag. The idea is to have the schedules mirrored in both systems and keep their states up-to-date. The same flag will be used in `jobs` to determine which execution between the temporal one and the orchestrator one will actually run.

packages/orchestrator/lib/clients/client.integration.test.ts
```diff
@@ -62,6 +62,34 @@ describe('OrchestratorClient', async () => {
...
+                        id: 123,
+                        connection_id: 'C',
+                        provider_config_key: 'P',
+                        environment_id: 456
+                    },
+                    debug: false
+                }
+            });
+
+            const res = await client.updateSyncFrequency({ scheduleName: name, frequencyMs: 600_000 });
```
bodinsamuel, line 90: would be great to check that it's been actually updated

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -45,13 +45,32 @@ const nameOfType = 'sync/action';
 
 type FlowWithVersion = Omit<IncomingFlowConfig, 'fileBody'>;
 
+interface OrchestratorInterface {
```
bodinsamuel: you could export this interface from the orchestrator and install as devDevp

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -669,6 +677,53 @@ export class Orchestrator {
...
+                    operation: LogActionEnum.SYNC_CLIENT,
+                    environmentId,
+                    metadata: {
+                        syncName,
+                        scheduleName,
+                        frequencyMs
+                    }
+                });
+            } else {
+                await logCtx?.info(`Updated sync "${syncId}" frequency to ${frequencyMs}ms.`);
```
bodinsamuel: syncId is already in the metadata, no entirely needed for customer facing message

bodinsamuel (approved):
Looks good 🪗
(nb: not tested)

khaliqgant (approved):

packages/shared/lib/services/sync/config/deploy.service.ts
TBonnin: good idea. I like this

packages/orchestrator/lib/clients/client.integration.test.ts
TBonnin: true but there is an unit test in scheduler that is testing the internal function is correctly updating. This test is more focus on confirming that the communication between the client and the scheduler service is successful, assuming that what' internal to the scheduler is working as expected. Soon I am gonna need to add a get schedule functionality to the client. I will update this test to check the value has been updated

packages/shared/lib/services/sync/config/deploy.service.ts
TBonnin: I don't even need to do that. There is already a OrchestratorClient interface (I even forget what I have done last week) and I can use that to instantiate a mock orchestrator with a noop OrchestratorClient in the test. [2862bca](https://github.com/NangoHQ/nango/pull/2298/commits/2862bca38b2df1618de4432fa57847e42f131ddb)

PR #2300, opened 2024-06-10
fix(webhooks): [nan-1173] fix webhooks

khaliqgant:
## Describe your changes
Extra parameter `rawBody` pushed the arguments so that an error was being thrown on any webhook ```2024-06-10T17:48:43.711Z [ERROR] [Webhook.Manager] error processing incoming webhook for hubspot -  logContextGetter.create is not a function```

This happened when the `rawBody` argument was added with this [change](https://github.com/NangoHQ/nango/commit/5501f45294630ee614108737f3f093d52a63843b#diff-63266d085481df27d38eb6629f683f2885fec64be0139f5a403a76b0b3cbaf21R13) and then later an additional argument was added but wasn't updated for the other webhooks. When the last argument `logContextGetter` was added with this [change](https://github.com/NangoHQ/nango/commit/dfd74ff2e0d790b0bfbcab65cc42b0bf59b6dafe#diff-63266d085481df27d38eb6629f683f2885fec64be0139f5a403a76b0b3cbaf21R23) the arguments became misaligned.

NAN-1173

### DIFF
--- a/packages/server/lib/controllers/oauth.controller.ts
+++ b/packages/server/lib/controllers/oauth.controller.ts
@@ -245,7 +245,7 @@ class OAuthController {
             }
 
             // certain providers need the credentials to be specified in the config
-            if (overrideCredentials) {
+            if (overrideCredentials && (overrideCredentials['oauth_client_id_override'] || overrideCredentials['oauth_client_secret_override'])) {
                 if (overrideCredentials['oauth_client_id_override']) {
                     config.oauth_client_id = overrideCredentials['oauth_client_id_override'];
 
@@ -1545,6 +1545,9 @@ class OAuthController {
                 );
             } else {
                 await updateSuccessActivityLog(activityLogId, template.auth_mode === 'CUSTOM' ? null : true);
+                if (template.auth_mode === 'CUSTOM') {
+                    await logCtx.success();
+                }
             }
 
             await telemetry.log(LogTypes.AUTH_TOKEN_REQUEST_SUCCESS, 'OAuth2 token request succeeded', LogActionEnum.AUTH, {
@@ -1555,7 +1558,6 @@ class OAuthController {
                 authMode: String(template.auth_mode)
             });
 
-            await logCtx.success();
             return publisher.notifySuccess(res, channel, providerConfigKey, connectionId, pending);
         } catch (err) {
             const prettyError = stringifyError(err, { pretty: true });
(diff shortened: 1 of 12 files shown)


packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -245,7 +245,7 @@ class OAuthController {
             }
 
             // certain providers need the credentials to be specified in the config
-            if (overrideCredentials) {
+            if (overrideCredentials && (overrideCredentials['oauth_client_id_override'] || overrideCredentials['oauth_client_secret_override'])) {
```
TBonnin, line 248: not sure I follow why we are adding this condition

packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -1545,6 +1545,9 @@ class OAuthController {
                 );
             } else {
                 await updateSuccessActivityLog(activityLogId, template.auth_mode === 'CUSTOM' ? null : true);
+                if (template.auth_mode === 'CUSTOM') {
```
TBonnin: why only when `CUSTOM`

TBonnin (commented):
named parameters FTW. Maybe we should have a rule like if a function has more than 2 parameters we must use an object.

packages/server/lib/controllers/oauth.controller.ts
bodinsamuel: this seems incorrect indeed unless I missed something

bodinsamuel (commented):
same comment as Thomas

packages/server/lib/controllers/oauth.controller.ts
khaliqgant: With a github oauth app credentials are actually passed in so in this scenario override was a false positive.

packages/server/lib/controllers/oauth.controller.ts
khaliqgant: Yup, thanks, removed

packages/shared/lib/services/connection.service.ts
```diff
@@ -985,6 +985,8 @@ class ConnectionService {
         });
         await logCtx.info('App connection was approved and credentials were saved');
 
+        await logCtx.success();
```
bodinsamuel: Can you remove this one too, success is applied at a higher level

bodinsamuel (approved):

packages/shared/lib/services/connection.service.ts
khaliqgant: Done with https://github.com/NangoHQ/nango/pull/2300/commits/6ecbe5c2badf1952121fa4c5a2a28620de0dc281

PR #2301, opened 2024-06-10
chore: delete sync schedule

TBonnin:
We delete sync schedule in both temporal and orchestrator. If schedule doesn't exist in orchestrator we just fails silently

### DIFF
--- a/packages/server/lib/controllers/flow.controller.ts
+++ b/packages/server/lib/controllers/flow.controller.ts
@@ -246,7 +246,7 @@ class FlowController {
                 const syncs = await getSyncsByConnectionIdsAndEnvironmentIdAndSyncName(connections, environmentId, syncName);
 
                 for (const sync of syncs) {
-                    await syncOrchestrator.softDeleteSync(sync.id, environmentId);
+                    await syncOrchestrator.softDeleteSync(sync.id, environmentId, orchestrator);
                 }
             }
 
--- a/packages/shared/lib/services/config.service.ts
+++ b/packages/shared/lib/services/config.service.ts
@@ -12,6 +12,7 @@ import encryptionManager from '../utils/encryption.manager.js';
 import syncOrchestrator from './sync/orchestrator.service.js';
 import { deleteSyncFilesForConfig, deleteByConfigId as deleteSyncConfigByConfigId } from '../services/sync/config/config.service.js';
 import environmentService from '../services/environment.service.js';
+import type { Orchestrator } from '../clients/orchestrator.js';
 
 class ConfigService {
     templates: Record<string, ProviderTemplate> | null;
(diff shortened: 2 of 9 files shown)


packages/shared/lib/services/config.service.ts
```diff
@@ -167,7 +168,7 @@ class ConfigService {
         return { id: id[0]?.id, unique_key: config.unique_key } as Pick<ProviderConfig, 'id' | 'unique_key'>;
     }
 
-    async deleteProviderConfig(providerConfigKey: string, environment_id: number): Promise<number> {
+    async deleteProviderConfig(providerConfigKey: string, environment_id: number, orchestrator: Orchestrator): Promise<number> {
```
TBonnin, line 171: here we have a good example of a dependency being called deep down into the codebase and requires the dependency to be passed along a chain of function calls. This is not pretty and not really sustainable on the long term but I am not gonna refactor now all the services that depends on temporal/orchestrator

packages/shared/lib/services/sync/orchestrator.service.ts
```diff
@@ -161,33 +161,35 @@ export class OrchestratorService {
         await deleteSyncConfig(syncConfigId);
     }
 
-    public async softDeleteSync(syncId: string, environmentId: number) {
-        await deleteScheduleForSync(syncId, environmentId);
+    public async softDeleteSync(syncId: string, environmentId: number, orchestrator: Orchestrator) {
+        await deleteScheduleForSync(syncId, environmentId); // TODO: legacy, to remove once temporal is removed
+
+        await orchestrator.deleteSync({ syncId, environmentId });
```
TBonnin, line 167: like with update, schedules are deleted in both temporal and orchestrator

packages/server/lib/controllers/flow.controller.ts
```diff
@@ -243,7 +246,7 @@ class FlowController {
                 const syncs = await getSyncsByConnectionIdsAndEnvironmentIdAndSyncName(connections, environmentId, syncName);
 
                 for (const sync of syncs) {
-                    await syncOrchestrator.softDeleteSync(sync.id, environmentId);
+                    await syncOrchestrator.softDeleteSync(sync.id, environmentId, orchestrator);
```
bodinsamuel, line 249: 🤭 I remember you talked about enforcing named parameters after two ahah

packages/shared/lib/services/config.service.ts
bodinsamuel: We know it's mostly because of the CLI (and because we forced ourselves to not publish any packages until we fix things). Any refactor will be stuck by the fact that CLI still depends on shared

bodinsamuel (approved):

packages/server/lib/controllers/flow.controller.ts
TBonnin: I am happy to strictly enforce this rule if we all agree. 😁

packages/shared/lib/services/config.service.ts
TBonnin: It is not just that. There are a few basic programming principles that we don't follow much in the codebase. For example too many functions are doing more than one thing. This one is a good example, it updates the db, call deleteSyncFiles, also call deleteSyncs.
Couple that with too many globally accessible components and you end up with highly couple subsystems that are hard to modify/refactor :(

PR #2302, opened 2024-06-11
feat(webhooks): [nan-1142] migrate to the external webhooks table

khaliqgant:
## Describe your changes
Move settings to an external table for webhook management. Script can be run more than once to not force an immediate release.

NAN-1142

### DIFF
--- a/packages/database/lib/migrations/20240611074119_update_external_webhooks_nullable.cjs
+++ b/packages/database/lib/migrations/20240611074119_update_external_webhooks_nullable.cjs
@@ -1,0 +1,15 @@
+const TABLE = '_nango_external_webhooks';
+
+exports.up = function (knex) {
+    return knex.schema.alterTable(TABLE, function (t) {
+        t.string('primary_url').nullable().alter();
+        t.string('secondary_url').nullable().alter();
+    });
+};
+
+exports.down = function (knex) {
+    return knex.schema.alterTable(TABLE, function (t) {
+        t.string('primary_url').notNullable().alter();
+        t.string('secondary_url').notNullable().alter();
+    });
+};
--- a/packages/database/lib/migrations/20240611074119_update_external_webhooks_nullable.cjs
+++ b/packages/database/lib/migrations/20240611074119_update_external_webhooks_nullable.cjs
@@ -4,6 +4,7 @@ exports.up = function (knex) {
     return knex.schema.alterTable(TABLE, function (t) {
         t.string('primary_url').nullable().alter();
         t.string('secondary_url').nullable().alter();
+        t.renameColumn('on_auth_refesh_error', 'on_auth_refresh_error');
         t.unique(['environment_id']);
     });
 };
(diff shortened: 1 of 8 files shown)


scripts/one-off/webhook-settings-migration/migrate.ts
```diff
@@ -0,0 +1,48 @@
+import './loadEnv.js';
```
bodinsamuel, line 1: better doing that in a DB migration (without transaction), otherwise self-hosted and entreprise won't be migrated

scripts/one-off/webhook-settings-migration/migrate.ts
```diff
@@ -0,0 +1,48 @@
...
+import { database } from '@nangohq/database';
+
+async function migrate() {
+    console.log('Starting webhook settings migration...');
+
+    let id = 0;
+
+    // eslint-disable-next-line no-constant-condition
+    while (true) {
+        const environments = await database.knex.select('*').from('_nango_environments').where('id', '>', id).limit(1000);
```
bodinsamuel: missing sortBY

packages/database/lib/migrations/20240611074119_update_external_webhooks_nullable.cjs
```diff
@@ -0,0 +1,16 @@
...
+        t.string('secondary_url').nullable().alter();
+        t.unique(['environment_id']);
+    });
+};
+
+exports.down = function (knex) {
+    return knex.schema.alterTable(TABLE, function (t) {
+        t.string('primary_url').notNullable().alter();
+        t.string('secondary_url').notNullable().alter();
+    });
```
bodinsamuel, line 18: I don't believe in Down, but you need to remove unique index

scripts/one-off/webhook-settings-migration/migrate.ts
khaliqgant: Indeed, good call. Updated with https://github.com/NangoHQ/nango/pull/2302/commits/139ecad30edb7e906df644551557aef12ae448f2

scripts/one-off/webhook-settings-migration/migrate.ts
khaliqgant: Updated.

packages/database/lib/migrations/20240611074119_update_external_webhooks_nullable.cjs
khaliqgant: Updated

TBonnin (commented):
why is there both a migration and a one-off script?

scripts/one-off/webhook-settings-migration/migrate.ts
bodinsamuel: do you want to delete this script?

bodinsamuel (approved):

khaliqgant:
> why is there both a migration and a one-off script?

> do you want to delete this script?

No, not yet. This will be merged and run before the rest will be ready so it is possible that some settings have changed so I'll run the one off script one last time before moving the UI to point to the new fields.

PR #2303, opened 2024-06-11
feat(nango.yaml): dedicated package, stricter parsing, wider types support

bodinsamuel:
Fixes NAN-1166
Fixes https://linear.app/nango/issue/NAN-1096/add-comment-at-the-top-of-modelsts-file-to-explain-that-its-auto

## 🤓 Context

First **I want to apologize for the length of this PR**, it's way too big but it was "all or nothing" situation unfortunately.

When I started playing with types/json schema validation **I wanted better output** for errors at compile time:
- realized the CLI had **2 different ways to parse** the nango.yaml
- realized **nango-config.service.ts** was not covering the same case that the one in cli
- realized neither were **not handling many types/edge cases** (i.e: arrays, number, boolean, union, exotic char in model name, etc.)
- then after modifying everything (took me a day), I realized it was **also parsing the table sync_configs** (for a reason I still don't get) so it was just breaking everything.

So I had no choice, to rewrite everything just because I wanted to cover more cases to be compatible with json schema 😵‍💫

The **Good news**:
- we now have a dedicated package to parse nango-yaml and it's battle tested (hopefully) 
- we now have one less thing to remove from shared (nb: the old code is still there and it's still used by sync_configs)
  There is only **13 mentions to shared left** in CLI
 
## 🚜 Changes

- New published package `@nangohq/nango-yaml` 
It will be published, as soon as you validate this PR otherwise the CLI won't pass.

- New output when validating nango.yaml
A lot of new edge cases, errors and warnings are now discovered. It might be breaking but at least it's explicit.

- **[breaking]** `endpoint` format is enforced (verb /uri)
- **[breaking]** `runs` is explicitly required
- **[breaking]** format of `NangoFlows` inside model.ts
- **[breaking]** Input and Output specifying a js value (instead of a model) now use a `type` interface with a unique generated name
- **[new]** Support for unions, arrays, type array, cyclic ref, literal string, number, true, false, null, undefined, multiple extends
- **[fix]** Disallow exotic char in model and property name
- **[fix]** Always compile model.ts after loading nango.yaml 
- ...not exhaustive changelist

> [!WARNING]
> UI won't show the appropriate model (fixing that later) and this is a breaking change for `GET /flows` `GET /config/:key`

## 😭 How do I review this huge PR?

- Starts with: `packages/nango-yaml/parser.v2.ts` 
- You can read: `packages/nango-yaml/modelsParser.ts` which handle the yaml -> js types 
- Then `packages/cli/services/model.service.ts` 
- Optional `packages/types/nangoYaml/index.ts` (describe yaml then parsed yaml)
- And voila, the rest is mostly side effects of having good types and slightly different objects

## 🧪 How do I test?

Please run the cli on your machine, I have extensively tested but it's so critical and super user dependent.
```sh
node ../nango/packages/cli/dist/index.js init 
node ../nango/packages/cli/dist/index.js generate // try to delete a script and make it re-generate it 
node ../nango/packages/cli/dist/index.js compile // check your model.ts
node ../nango/packages/cli/dist/index.js deploy dev
```
```yaml
models:
  Test1:
    id: number
    str: "literal"
    issue: Unknown
    arr:
      - hello
      - boy

  Test2:
      id: integer
      title: string
      url: string
  Parent:
      id: integer
  Test3:
      __extends: Parent
      oui: true
      non: false
      OUI: True
      NON: False
      num: 1
      kd?: Pick<dfdk>
      nulle: null
      union: True | False
      nes:
        ted: number
   Test4:
     __string: number
```

## Example

Screenshot 2024-06-13 at 11 40 10


packages/nango-yaml/lib/modelsParser.ts
```diff
@@ -0,0 +1,200 @@
...
+            );
+            return false;
+        }
+
+        // At this point we are sure it's a Model
+        this.parseOne({ name, fields: this.raw[name]! });
+        return true;
+    }
+
+    parseFields({ fields, parent }: { fields: NangoYamlModelFields; parent: string }): NangoModelField[] {
```
bodinsamuel, line 73: This is the main function to parse models and handle all the edge cases regarding types

packages/nango-yaml/lib/helpers.unit.test.ts
```diff
@@ -0,0 +1,154 @@
+import { expect, describe, it } from 'vitest';
```
bodinsamuel, line 1: Moved from shared

packages/nango-yaml/lib/helpers.ts
```diff
@@ -0,0 +1,192 @@
...
+export function isDisallowedType(value: string) {
+    return typesWithGenerics.some((type) => value.startsWith(`${type}<`));
+}
+
+const regQuote = /^[a-zA-Z0-9_]+$/;
+export function shouldQuote(name: string) {
+    return !regQuote.test(name);
+}
+
+export function getProviderConfigurationFromPath({ filePath, parsed }: { filePath: string; parsed: NangoYamlParsed }): NangoYamlParsedIntegration | null {
```
bodinsamuel, line 180: moved from shared

packages/nango-yaml/lib/helpers.ts
```diff
@@ -0,0 +1,192 @@
...
+    primitives: ['string', 'number', 'boolean', 'bigint', 'symbol'],
+    builtInObjects: ['Object', 'Array', 'Function', 'Date', 'RegExp', 'Map', 'Set', 'WeakMap', 'WeakSet', 'Promise', 'Symbol', 'Error'],
+    utilityTypes: ['Record', 'Partial', 'Readonly', 'Pick', 'Omit', 'Awaited', 'Required', 'Exclude', 'Extract', 'Uppercase', 'Lowercase']
+};
+const typesLowercase = Object.values(JAVASCRIPT_AND_TYPESCRIPT_TYPES)
+    .flat()
+    .map((v) => v.toLocaleLowerCase());
+const typesWithGenerics = [...JAVASCRIPT_AND_TYPESCRIPT_TYPES.builtInObjects, ...JAVASCRIPT_AND_TYPESCRIPT_TYPES.utilityTypes];
+// Only used externally
+export function isJsOrTsType(type?: string): boolean {
```
bodinsamuel, line 115: I moved it before realizing it wasn't what I expected, it will be deleted as soon as I migrate the backend to use the new schema

packages/nango-yaml/lib/helpers.ts
```diff
@@ -0,0 +1,192 @@
...
+
+/**
+ * Get Interval
+ * @desc get the interval based on the runs property in the yaml. The offset
+ * should be the amount of time that the interval should be offset by.
+ * If the time is 1536 and the interval is 30m then the next time the sync should run is 1606
+ * and then 1636 etc. The offset should be based on the interval and should never be
+ * greater than the interval
+ */
+export function getInterval(runs: string, date: Date): IntervalResponse | Error {
```
bodinsamuel, line 53: Moved from shared, with the diff that ServiceResponse does not exists and not available in CLI so I just return an Error if any

packages/nango-yaml/lib/helpers.ts
```diff
@@ -0,0 +1,192 @@
...
+import ms from 'ms';
+import type { StringValue } from 'ms';
+import type { NangoYaml, NangoYamlParsed, NangoYamlParsedIntegration } from '@nangohq/types';
+
+interface IntervalResponse {
+    interval: StringValue;
+    offset: number;
+}
+
+export function determineVersion(configData: NangoYaml): 'v1' | 'v2' {
```
bodinsamuel, line 11: moved from shared

packages/cli/package.json
```diff
@@ -17,15 +17,14 @@
     },
     "license": "SEE LICENSE IN LICENSE FILE IN GIT REPOSITORY",
     "scripts": {
-        "postbuild": "npm run copy:types",
         "copy:files": "copyfiles -u 1 lib/templates/* dist",
-        "copy:types": "cp `node -p \"require.resolve('@nangohq/shared/dist/sdk/sync.d.ts')\"` './dist/nango-sync.d.ts'",
```
bodinsamuel, line 22: this one is a big bet, I'm trying to load the file from user's global node_modules, should work but let's see :D

packages/cli/lib/sync.unit.cli-test.ts
```diff
@@ -374,164 +383,62 @@ describe('generate function tests', () => {
...
-        const dir = getTestDirectory('validation-array-model');
-        await init({ absolutePath: dir });
-
-        await fs.promises.copyFile(`${fixturesPath}/nango-yaml/v2/model-array-types/nango.yaml`, `${dir}/nango.yaml`);
-        await expect(generate({ debug: false, fullPath: dir })).resolves.toBeUndefined();
-        const modelsFile = await fs.promises.readFile(`${dir}/models.ts`, 'utf8');
-        expect(modelsFile).not.toContain(`'Other[]' | null | undefined;`);
-        expect(modelsFile).toContain(`Other[] | null | undefined;`);
-    });
-
```
bodinsamuel, line 531: All of this has been moved either to model.service.ts or removed because no longer relevant

TBonnin:
haven't had time to review yet. I will do it on Friday. Sorry about that

khaliqgant (changes requested):
Tested this against a customer nango-integration to get a sense of it. I couldn't really do much because there were a ton of errors and not all helpful in some cases without a line number reference

```
integrations > salesforce > syncs > activities > endpoint
  error endpoint must be a URI (or an array of URIs) with an HTTP verb, i.e: "GET /tickets/ticket"
```

Shouldn't be so strictly enforced and the intention when this was built was the syncs fall back to `GET` and actions to `POST`

Also have this opaque error at the end 

```[object Object]```

Will do a code review next week

File: packages/cli/lib/services/model.service.ts [hunk 9]
```diff
@@ -1,211 +1,167 @@
...
+
+    output.push('// ------ SDK');
+    output.push(generateSDKTypes());
+    output.push('// ------ /SDK');
+    output.push('');
+
+    output.push(`// ------ Flows
+export const NangoFlows = ${JSON.stringify(parsed.integrations, null, 2)} as const;
+// ------ /Flows
+`);
```
TBonnin: can you directly interpolate the output string to remove the noise?
```suggestion
return `
// ---------------------------
// This file was generated by Nango (v${NANGO_VERSION})
// It's recommended to version this file
// https://nango.dev
// ---------------------------

// ------ Models
${generateInterfaces({ parsed }).join('\n\n')}
// ------ /Models

// ------ SDK
${generateSDKTypes()}
// ------ /SDK

// ------ Flows
export const NangoFlows = ${JSON.stringify(parsed.integrations, null, 2)} as const;
// ------ /Flows
`;
```

packages/nango-yaml/lib/parser.v2.ts
```diff
@@ -0,0 +1,261 @@
...
+    ParserErrorInvalidRuns,
+    ParserErrorModelIsLiteral
+} from './errors.js';
+import { getInterval } from './helpers.js';
+
+export class NangoYamlParserV2 extends NangoYamlParser {
+    parse(): boolean {
+        const yaml = this.raw as unknown as NangoYamlV2;
+        const output: NangoYamlParsedIntegration[] = [];
+        this.modelsParser.parseAll();
```
TBonnin, line 21: I am gonna say something stupid but since we are basically validating json. Can we not use zod to validate the schema, plus a little bit of post-processing to validate references/duplication?

packages/nango-yaml/lib/parser.v2.ts
```diff
@@ -0,0 +1,261 @@
...
+            if (!action) {
+                continue;
+            }
+
+            const modelNames = new Set<string>();
+
+            const modelOutput = this.getModelForOutput({ rawOutput: action.output, usedModels, name: actionName, type: 'action', integrationName });
+            if (modelOutput && modelOutput.length > 0) {
+                modelOutput.forEach((m) => modelNames.add(m.name));
+            }
```
TBonnin:
```suggestion
            const modelOutput = this.getModelForOutput({ rawOutput: action.output, usedModels, name: actionName, type: 'action', integrationName });
            const modelNames = new Set<string>((this.getModelForOutput(modelOutput) || []).map((m) => m.name)
```

packages/nango-yaml/lib/parser.v2.ts
```diff
@@ -0,0 +1,261 @@
...
+
+            const modelInput = this.getModelForInput({ usedModels, rawInput: action.input, name: actionName, type: 'action', integrationName });
+            if (modelInput) {
+                modelNames.add(modelInput.name);
+            }
+
+            const endpoint: NangoSyncEndpoint = {};
+            if (action.endpoint) {
+                const split = action.endpoint.split(' ') as [HTTP_VERB, string];
```
TBonnin: does that mean we don't validate if the verb is actually an accepted HTTP_VERB?

packages/nango-yaml/lib/parser.v2.ts
```diff
@@ -0,0 +1,261 @@
...
+
+                this.endpoints.add(action.endpoint);
+                endpoint[split[0]] = split[1];
+            }
+
+            const parsedAction: ParsedNangoAction = {
+                name: actionName,
+                type: 'action',
+                description: action.description || '',
+                scopes: Array.isArray(action.scopes) ? action.scopes : action.scopes ? action.scopes.split(',') : [],
```
TBonnin: I found double ternary super confusing. That might just be me though

packages/nango-yaml/lib/parser.v2.ts
```diff
@@ -0,0 +1,261 @@
...
+        }
+
+        for (const integrationName in yaml.integrations) {
+            const integration = yaml.integrations[integrationName];
+            if (!integration) {
+                continue;
+            }
+
+            // check that models are used only once per integration
+            const usedModels = new Set<string>();
```
TBonnin: I don't really understand the use of this mutable set `usedModels` passed around `parseSyncs` and `parseActions` can you not get the models from the parsed result and then check for errors (duplication, etc...)?

packages/nango-yaml/lib/parser.ts
```diff
@@ -0,0 +1,75 @@
...
+import { ModelsParser } from './modelsParser.js';
+import { ParserErrorDuplicateModel, ParserErrorMissingId, ParserErrorModelIsLiteral } from './errors.js';
+import type { ParserError } from './errors.js';
+
+export abstract class NangoYamlParser {
+    raw: NangoYaml;
+    parsed: NangoYamlParsed | undefined;
+    modelsParser: ModelsParser;
+    // check that every endpoint is unique across syncs and actions
+    endpoints = new Set<string>();
```
TBonnin: why a variable for that while the endpoints can be retrieve looking at the `parsed` content?

packages/nango-yaml/lib/parser.v2.ts
```diff
@@ -0,0 +1,261 @@
...
+                if (tmp.length !== modelOutput.length) {
+                    this.errors.push(new ParserErrorEndpointsMismatch({ syncName: syncName, path: [integrationName, 'syncs', syncName] }));
+                    continue;
+                }
+
+                let failed = false;
+                for (const endpoint of tmp) {
+                    const split = endpoint.split(' ') as [HTTP_VERB, string];
+
+                    if (this.endpoints.has(endpoint)) {
```
TBonnin: linked to my question above about the `endpoints` class field: this function relies on the internal state of the parser instance, which might be different if the parsing functions are reordered. Would it be simpler to first deal with the pure parsing and then in a second pass, take care of the validations like duplicate endpoints, etc... it would make those parsing function smaller, without side effects and easier to compose/reorder

TBonnin (commented):
thank you for abstracting this logic into its own package (and for the type friendly approach)
I am always a bit scared when I encounter complex parsing being done by hand. I am happy to discuss my comments on Monday.
I am also curious about the breaking changes and release plan

packages/cli/lib/services/model.service.ts
bodinsamuel: yeah true, it didn't was that simple at the beginning makes sense now 👍🏻

packages/nango-yaml/lib/parser.v2.ts
bodinsamuel: it's done in the json schema validation

packages/nango-yaml/lib/parser.v2.ts
bodinsamuel: I'm not sure I get this. Json schema is validating the schema (like zod would do), but all the parsing has almost nothing to do with validation, it's handling typescript syntax inside the yaml

packages/nango-yaml/lib/parser.v2.ts
bodinsamuel: true, changing

packages/nango-yaml/lib/parser.v2.ts
bodinsamuel: fair, I'll move the endpoint for clarification and the rest of the validation stuff, feels like it's less clear but functions are smaller

packages/nango-yaml/lib/parser.v2.ts
bodinsamuel: resolved below

packages/nango-yaml/lib/parser.v2.ts
bodinsamuel: It's less readable imo, I prefer explicit longer code for maintainability

bodinsamuel:
Rep to @khaliqgant 

> Shouldn't be so strictly enforced and the intention when this was built was the syncs fall back to GET and actions to POST

Alright, I don't agree it's good DX but reverted to have a fallback.

>  some cases without a line number reference

I don't provide any line number ref, only the path of the yaml if possible.

> Also have this opaque error at the end [object Object]

Could not reproduce, happy to dig if you send me the customer name in DM

bodinsamuel:
> I am always a bit scared when I encounter complex parsing being done by hand.

Not much choice here, the yaml parsing is still done by a state machine (I suppose) but the model syntax half string half typescript stuff is harder to parser in any other way.

> I am also curious about the breaking changes and release plan

Except the few validation breaking changes, the release should be not different than any other release

khaliqgant:
> Could not reproduce, happy to dig if you send me the customer name in DM

This seems to happen when there is a `type` used in the sync instead of a `sync_type`

```
syncs:
          salesforce-contacts:
              type: full
              runs: every day
              auto_start: false
              output: Contact
              input: FieldSchema
              endpoint: /salesforce/contact
              scopes: offline_access,api
              description: |
                Syncs contacts based on a metedata defined object. Requires metadata
                Auto start false.
              webhook-subscriptions:
                - contact.update
```

with this configuration gives:

```
➜  nango-integrations git:(main) ✗ npx nango scc
nango.yaml validation failed

integrations > salesforce > syncs > salesforce-contacts
  error must NOT have additional properties [additionalProperties]
  additionalProperty: type

[object Object]
null
```

If a model is reference but doesn't actually exist 

It seems the `__string` isn't handled correctly:

```
  Contact:
    id: string
    userId: string
    __string: string
```
produces:
```
export interface Contact {
  [key: string]: 'string';
  id: string;
  userId: string;
};
```

where previously it would correctly output:
```
export interface Contact {
  id: string;
  userId: string;
 [key: string]: string;
}
```

Given this:

```
        actions:
          salesforce-fetch-nested-fields:
              description: Fetch available fields for a nested field
              input: string
              endpoint: /salesforce/fetch-nested-fields
              output: NestedFieldSchema
              scopes: offline_access,api
```

The outputted model is 
```
"input": "Anonymous_salesforce_action_salesforcefetchnestedfields_input"
```

which is unexpected. I see why you did it to encourage people to use models but IMO an input of a string is perfectly valid and shouldn't require a model. But maybe this is more of a product question and good for Bastien to weigh in. 

Also if a model is declared but doesn't actually exist the yaml parsing continues but feel like we should catch this?

khaliqgant:
When running `nango scc` I definitely miss getting the fields outputted -- not sure if this is coming in the piepline of your work. It used to output this:

```
{
        "name": "salesforce-fetch-fields",
        "type": "action",
        "models": [
          {
            "name": "FieldSchema",
            "fields": [
              {
                "name": "fields",
                "type": "Field[]"
              },
              {
                "name": "childRelationships",
                "type": "ChildField[]"
              },
              {
                "name": "userIdField",
                "type": "string | undefined | null"
              }
            ]
          },
```

now just gives
```
        {
          "name": "salesforce-fetch-fields",
          "type": "action",
          "description": "Fetch available contact fields",
          "scopes": [
            "offline_access",
            "api"
          ],
          "input": null,
          "output": [
            "FieldSchema"
          ],
          "usedModels": [
            "FieldSchema"
          ],
          "endpoint": {
            "POST": "/salesforce/fetch-fields"
          }
        }
      ]
```

khaliqgant:
> [new] Support for unions, arrays, type array, cyclic ref, literal string, number, true, false, null, undefined, multiple extends

Can you give examples of each of these?

`multiple extends` works nicely!

```
  SelectedField:
    name: string
    label: string
    type: string
    foo:
      __extends: Entity
```

bodinsamuel:
> This seems to happen when there is a type used in the sync instead of a sync_type

`type` was never officially supported (cf the json schema) so to me I just fixed the validation.

> It seems the __string isn't handled correctly:

✅ Thanks, missed that in my tests

> [object Object]

✅ Okay found the issue thanks

> I definitely miss getting the fields outputted 

Added the models output, but yeah I'm not keeping the raw format in memory

> input of a string is perfectly valid and shouldn't require a model

It's still perfectly valid, but I need one model per input/output so that I can generate json schema (and more) in a single way. With just one way there is no need to check everywhere in the backend if a string is jsType or a model.

> Also if a model is declared but doesn't actually exist the yaml parsing continues but feel like we should catch this?

That's the problem of supporting js types, there is no way to differentiate literal string and model name. So I added a warning about literal string but I don't see a better way rn

> [new] Support for unions, arrays, type array, cyclic ref, literal string, number, true, false, null, undefined, multiple extends

You have in the test and PR desc.

```yaml
models:
  Ref:
    id: string
  Ref2:
    name: string
  Base:
    unions: string | null | Ref | true[]
    arrUnion:
      - Ref
      - true
    literal: hello
    num: 1
    literalNull: null[]
    obj:
      nested: string
    optional?: string
    __extends: Ref,Ref2
```

TBonnin (approved):

packages/nango-yaml/lib/modelsParser.ts
```diff
@@ -0,0 +1,201 @@
+import type { NangoModel, NangoModelField, NangoYamlModel, NangoYamlModelFields } from '@nangohq/types';
+import { getNativeDataType, getPotentialTypeAlias, isDisallowedType, shouldQuote } from './helpers.js';
+import { ParserError, ParserErrorCycle, ParserErrorDataSyntax, ParserErrorExtendsNotFound, ParserErrorInvalidModelName } from './errors.js';
+
+export class ModelsParser {
```
khaliqgant, line 5: Small nit but I would keep this singular, `ModelParser`

khaliqgant (approved):
Small nits, but works well. I was using this branch as I was writing customer scripts today 💯

This is a bit weird:

```
        {
          "name": "selectedFields",
          "value": [
            {
              "name": "0",
              "value": "SelectedField[]",
              "model": true,
              "optional": false
            },
            {
              "name": "1",
              "tsType": true,
              "array": false,
              "optional": false
            },
            {
              "name": "2",
              "value": null,
              "tsType": true,
              "array": false,
              "optional": false
            }
          ],
          "union": true,
          "optional": false
        }

```

This is when the type is like this

```
selectedFields: SelectedField[] | undefined | null
```

weird in the sense that the name is the array index, but get that this might be technically challenging.

packages/nango-yaml/lib/modelsParser.ts
bodinsamuel: I see what you mean, it's like that because it's parsing the `models` list and contains all the models maybe ModelsPropParser would have been clearer

bodinsamuel:
> SelectedField[] | undefined | null

Interesting, actually it should be supported. Adding this edge case, thanks a lot for the thorough testing 🙏🏻

PR #2307, opened 2024-06-11
fix(logs): cosmetic fix

bodinsamuel:
## Describe your changes

- Consistent date format across browser
Turns out `Intl.DateTimeFormat` is up to the browser, so each one have a different way to present what is supposed to be standard :| 
- Add `:action` only if the operation has one
- Fix HTTP 500 error handling in Logs
- Change HTTP 500 error signature to follow new pattern


TBonnin:
>       Turns out `Intl.DateTimeFormat` is up to the browser, so each one have a different way to present what is supposed to be standard :|

😢

TBonnin (approved):

PR #2308, opened 2024-06-11
feat(integrations): add Jira Data Center support

zdhickman:
## Describe your changes

First pass at adding Jira Data Center support. Locally, I was able to set up a new connection and fetch some Jira projects. ~Unsure about how refresh tokens might function~ refresh tokens work OOTB

### DIFF
--- a/docs-v2/integrations/all/jira-data-center.mdx
+++ b/docs-v2/integrations/all/jira-data-center.mdx
@@ -1,0 +1,29 @@
+---
+title: Jira Data Center
+sidebarTitle: Jira Data Center
+---
+
+API configuration: [`jira-data-center`](https://nango.dev/providers.yaml)
+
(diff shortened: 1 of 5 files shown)


bastienbeurier:
@zdhickman it's possible that the refresh logic already works with your implementation. Do you mind trying to refresh the token from the UI to see if it works? You have to go to the Connections tab, click on the connection, go to the Authorization subtab, press the refresh button next to the access token. Thank you!

zdhickman:
@bastienbeurier just tested this out and was able to refresh the access token!

bastienbeurier (approved):

bodinsamuel (approved):
🎯

bodinsamuel:
Thanks for your contribution @zdhickman 🥇

PR #2309, opened 2024-06-11
feat: schedule syncs via the orchestrator

TBonnin:
New syncs are going to be created in both the orchestrator and temporal but they are not yet processed via the orchestrator.
This will allow us to check schedule are created correctly and tasks are scheduled properly

Next PR will modify `handler.ts` to actually process the sync task.

### DIFF
--- a/packages/jobs/lib/processor/handler.ts
+++ b/packages/jobs/lib/processor/handler.ts
@@ -35,7 +35,7 @@ async function abort(_task: OrchestratorTask): Promise<Result<void>> {
 }
 
 async function sync(task: TaskSync): Promise<Result<JsonValue>> {
-    return Err(`Not implemented: ${JSON.stringify({ taskId: task.id })}`);
+    return Ok({ id: task.id });
 }
 
 async function action(task: TaskAction): Promise<Result<JsonValue>> {
--- a/packages/jobs/lib/processor/processor.ts
+++ b/packages/jobs/lib/processor/processor.ts
@@ -21,10 +21,17 @@ export class Processor {
     start() {
         logger.info('Starting task processors');
         try {
+            const syncWorker = new ProcessorWorker({
+                orchestratorUrl: this.orchestratorServiceUrl,
+                groupKey: 'sync',
+                maxConcurrency: 200
+            });
+            syncWorker.start();
+
             const actionWorker = new ProcessorWorker({
                 orchestratorUrl: this.orchestratorServiceUrl,
                 groupKey: 'action',
-                maxConcurrency: 500
+                maxConcurrency: 200
             });
             actionWorker.start();
 
@@ -34,7 +41,7 @@ export class Processor {
                 maxConcurrency: 50
             });
             webhookWorker.start();
-            this.workers = [actionWorker, webhookWorker];
+            this.workers = [syncWorker, actionWorker, webhookWorker];
             this.stopped = false;
         } catch (e) {
             logger.error(e);
--- a/packages/orchestrator/lib/clients/types.ts
+++ b/packages/orchestrator/lib/clients/types.ts
@@ -18,7 +18,6 @@ interface SyncArgs {
         environment_id: number;
         connection_id: string;
     };
-    syncJobId: number;
 }
 interface ActionArgs {
     actionName: string;
@@ -89,7 +88,6 @@ export function TaskSync(props: TaskCommonFields & SyncArgs): TaskSync {
         attempt: props.attempt,
         syncId: props.syncId,
         syncName: props.syncName,
-        syncJobId: props.syncJobId,
         debug: props.debug,
         connection: props.connection,
         abortController: new AbortController(),
--- a/packages/orchestrator/lib/routes/v1/postRecurring.ts
+++ b/packages/orchestrator/lib/routes/v1/postRecurring.ts
@@ -5,6 +5,7 @@ import type { ApiError, Endpoint } from '@nangohq/types';
 import type { EndpointRequest, EndpointResponse, RouteHandler, Route } from '@nangohq/utils';
 import { validateRequest } from '@nangohq/utils';
 import { syncArgsSchema } from '../../clients/validate.js';
+import type { TaskType } from '../../types.js';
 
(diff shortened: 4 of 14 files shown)


packages/jobs/lib/processor/handler.ts
```diff
@@ -35,7 +35,7 @@ async function abort(_task: OrchestratorTask): Promise<Result<void>> {
 }
 
 async function sync(task: TaskSync): Promise<Result<JsonValue>> {
-    return Err(`Not implemented: ${JSON.stringify({ taskId: task.id })}`);
+    return Ok({ id: task.id });
```
TBonnin, line 38: Marking all syncs are successful

packages/jobs/lib/processor/processor.ts
```diff
@@ -21,10 +21,17 @@ export class Processor {
...
+                groupKey: 'sync',
+                maxConcurrency: 200
+            });
+            syncWorker.start();
+
             const actionWorker = new ProcessorWorker({
                 orchestratorUrl: this.orchestratorServiceUrl,
                 groupKey: 'action',
-                maxConcurrency: 500
+                maxConcurrency: 200
```
TBonnin, line 34: in temporal syncs and actions shares the same queue with a max concurrency at 500.
Using different queues in the orchestrator with an arbitrary max concurrency at 200. (we usually never have more than 2 digits running workflows in temporal)

packages/orchestrator/lib/clients/types.ts
```diff
@@ -18,7 +18,6 @@ interface SyncArgs {
         environment_id: number;
         connection_id: string;
     };
-    syncJobId: number;
```
TBonnin, line 21: `syncJobId` is actually not necessary. SyncJob will be created/updated at processing time.

packages/orchestrator/lib/routes/v1/postRecurring.ts
```diff
@@ -27,7 +28,7 @@ export type PostRecurring = Endpoint<{
             startedToCompleted: number;
             heartbeat: number;
         };
-        args: JsonValue;
+        args: JsonValue & { type: TaskType };
```
TBonnin, line 31: bug fix. I had forgotten about it

packages/server/lib/hooks/hooks.ts
```diff
@@ -78,8 +79,7 @@ export const connectionCreated = async (
     const { connection, environment, auth_mode } = createdConnectionPayload;
 
     if (options.initiateSync === true && !isHosted) {
-        const syncClient = await SyncClient.getInstance();
-        await syncClient?.initiate(connection.id as number, logContextGetter);
+        await orchestratorService.createSyncForConnection(connection.id as number, logContextGetter, orchestrator);
```
TBonnin, line 82: move the `initiate` function out of `syncClient` since it will be necessary to move/refactor it anyway when getting rid of temporal

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -822,27 +826,46 @@ export class Orchestrator {
                 recordsService: props.recordsService,
                 initiator: props.initiator
             });
-        }
+        };
```
TBonnin, line 829: I would have like the commands to both run via temporal and via the orchestrator so that the state of a schedule is kept in-sync  but unfortunately the RUN_FULL command has a bunch of side effects (deleted records, resetting lastSyncDate, ...) so I am only running both when not a RUN_FULL.

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -857,4 +880,247 @@ export class Orchestrator {
...
+            });
+            if (logCtx) {
+                await logCtx.error('Failed to init sync', { error: err });
+                await logCtx.failed();
+            }
+            return Err(`Failed to schedule sync: ${err}`);
+        }
+    }
+    // TODO: remove once temporal is removed
+    async scheduleSyncHelper(
```
TBonnin, line 1041: same as with `runSyncCommand`. This function responsibility is to manage the flag related logic

packages/shared/lib/services/sync/orchestrator.service.ts
```diff
@@ -61,14 +62,58 @@ interface CreateSyncArgs {
     syncName: string;
 }
 
+const logger = getLogger('orchestrator.service');
+
 export class OrchestratorService {
-    public async create(
+    public async createSyncForConnection(nangoConnectionId: number, logContextGetter: LogContextGetter, orchestrator: Orchestrator): Promise<void> {
```
TBonnin, line 68: this is the `initiate` function moved from `syncClient` and renamed `createSyncForConnection`.
I wanted to merge it with the `create` function (renamed `createSyncForConnections`) but the input types are too different and I would have need to refactor the connectionCreatedHook too much. So until I find the courage to do it those 2 functions will remain. At least now they live next to each other

packages/shared/lib/services/sync/orchestrator.service.ts
```diff
@@ -438,17 +499,19 @@ export class OrchestratorService {
...
                 providerConfigKey,
                 environmentId,
                 flow as unknown as IncomingFlowConfig,
                 logContextGetter,
+                orchestrator,
                 false
             );
         }
     }
+    // TODO:
```
TBonnin, line 514: I'll take care of returning the correct sync data depending on the flag in next PR

packages/shared/lib/clients/orchestrator.ts
bodinsamuel: sorry about that, not that that those side effects could be done in Jobs if it helps

bodinsamuel (approved):

packages/shared/lib/clients/orchestrator.ts
TBonnin: that's a good point. we would need to pass the sync type as part of the task args but that's feasible

TBonnin: I am not gonna do it as past of this PR though

PR #2310, opened 2024-06-12
fix(scheduler): retrying when task fails can interfere with scheduling

TBonnin:
when a task fails a retry one is potentially created. We are now locking the schedule to ensure failing a task and retrying doesn't interfere with the scheduling worker or any other update of the schedule.

Forgot this scenario when I implemented schedule locking in https://github.com/NangoHQ/nango/pull/2296

### DIFF
--- a/packages/scheduler/lib/scheduler.ts
+++ b/packages/scheduler/lib/scheduler.ts
@@ -232,29 +232,42 @@ export class Scheduler {
      * const failed = await scheduler.fail({ taskId: '00000000-0000-0000-0000-000000000000', error: {message: 'error'});
      */
     public async fail({ taskId, error }: { taskId: string; error: JsonValue }): Promise<Result<Task>> {
-        const failed = await tasks.transitionState(this.dbClient.db, { taskId, newState: 'FAILED', output: error });
-        if (failed.isOk()) {
-            const task = failed.value;
-            this.onCallbacks[task.state](task);
-            // Create a new task if the task is retryable
-            if (task.retryMax > task.retryCount) {
-                const taskProps: ImmediateProps = {
-                    name: `${task.name}:${task.retryCount + 1}`, // Append retry count to make it unique
-                    payload: task.payload,
-                    groupKey: task.groupKey,
-                    retryMax: task.retryMax,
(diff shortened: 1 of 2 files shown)


packages/scheduler/lib/scheduler.ts
```diff
@@ -232,29 +232,42 @@ export class Scheduler {
...
+            const task = await tasks.get(trx, taskId);
+            if (task.isErr()) {
+                return Err(`fail: Error fetching task '${taskId}': ${stringifyError(task.error)}`);
+            }
+            // if task is from a schedule,
+            // lock the schedule to prevent concurrent update or scheduling of tasks
+            // while we are potentially creating a new retry task
+            if (task.value.scheduleId) {
+                await schedules.search(trx, { id: task.value.scheduleId, limit: 1, forUpdate: true });
+            }
```
TBonnin, line 245: This does add one or two extra db queries but that's the price to pay to avoid race condition.

packages/scheduler/lib/scheduler.ts
```diff
@@ -232,29 +232,42 @@ export class Scheduler {
...
-                    retryCount: task.retryCount + 1,
-                    createdToStartedTimeoutSecs: task.createdToStartedTimeoutSecs,
-                    startedToCompletedTimeoutSecs: task.startedToCompletedTimeoutSecs,
-                    heartbeatTimeoutSecs: task.heartbeatTimeoutSecs
-                };
-                const res = await this.immediate(taskProps);
-                if (res.isErr()) {
-                    logger.error(`Error retrying task '${taskId}': ${stringifyError(res.error)}`);
+        return await this.dbClient.db.transaction(async (trx) => {
+            const task = await tasks.get(trx, taskId);
```
bodinsamuel, line 236: should this be get for update instead of a regular get? otherwise it won't prevent other process to read the task

bodinsamuel (approved):

packages/scheduler/lib/scheduler.ts
TBonnin: I don't think this is absolutely necessary. Let me know if you disagree

PR #2313, opened 2024-06-12
fix(ui): Make the connection creation date into a <time> element with a title

eabruzzese:
## Describe your changes

This PR updates the "Created" column's value to be a proper `<time>` element with a `title` attribute, which allows the user to hover and see the exact datetime in a tooltip.

Screenshot 2024-06-12 at 11 16 38 AM

- [ ] I added tests, otherwise the reason is: very minor cosmetic change
- [ ] I added observability, otherwise the reason is: very minor cosmetic change
- [ ] I added analytics, otherwise the reason is: very minor cosmetic change


bastienbeurier:
Thank you for contributing this! We'll get back to you within 24h.

bodinsamuel (approved):
Thanks for your contribution, we should probably add that on the whole platform

PR #2315, opened 2024-06-12
feat: process syncs via the orchestrator

TBonnin:
This PR:
- adds processing of sync in the orchestrator processor (behind a flag)
- adds support for sync cancellation in the orchestrator processor
- add support for sync heartbeat in the orchestrator processor

### DIFF
--- a/packages/jobs/lib/activities.ts
+++ b/packages/jobs/lib/activities.ts
@@ -32,7 +32,8 @@ import {
     isInitialSyncStillRunning,
     getSyncByIdAndName,
     getLastSyncDate,
-    getSyncConfigRaw
+    getSyncConfigRaw,
+    featureFlags
 } from '@nangohq/shared';
 import { records as recordsService } from '@nangohq/records';
 import { getLogger, stringifyError, errorToObject } from '@nangohq/utils';
@@ -47,6 +48,13 @@ const logger = getLogger('Jobs');
 export async function routeSync(args: InitialSyncArgs): Promise<boolean | object | null> {
     const { syncId, syncJobId, syncName, nangoConnection, debug } = args;
     let environmentId = nangoConnection?.environment_id;
+
+    const isGloballyEnabled = await featureFlags.isEnabled('orchestrator:schedule', 'global', false);
+    const isEnvEnabled = await featureFlags.isEnabled('orchestrator:schedule', `${environmentId}`, false);
+    const isOrchestrator = isGloballyEnabled || isEnvEnabled;
+    if (isOrchestrator) {
+        return false;
+    }
 
     // https://typescript.temporal.io/api/classes/activity.Context
     const context: Context = Context.current();
@@ -112,6 +120,13 @@ export async function scheduleAndRouteSync(args: ContinuousSyncArgs): Promise<bo
     const { syncId, syncName, nangoConnection, debug } = args;
     let environmentId = nangoConnection?.environment_id;
     let syncJobId;
+
+    const isGloballyEnabled = await featureFlags.isEnabled('orchestrator:schedule', 'global', false);
+    const isEnvEnabled = await featureFlags.isEnabled('orchestrator:schedule', `${environmentId}`, false);
+    const isOrchestrator = isGloballyEnabled || isEnvEnabled;
+    if (isOrchestrator) {
+        return false;
+    }
 
     const initialSyncStillRunning = await isInitialSyncStillRunning(syncId);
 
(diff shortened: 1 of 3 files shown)


packages/jobs/lib/activities.ts
```diff
@@ -113,6 +121,13 @@ export async function scheduleAndRouteSync(args: ContinuousSyncArgs): Promise<bo
     let environmentId = nangoConnection?.environment_id;
     let syncJobId;
 
+    const isGloballyEnabled = await featureFlags.isEnabled('orchestrator:schedule', 'global', false);
+    const isEnvEnabled = await featureFlags.isEnabled('orchestrator:schedule', `${environmentId}`, false);
+    const isOrchestrator = isGloballyEnabled || isEnvEnabled;
+    if (isOrchestrator) {
+        return false;
+    }
+
```
TBonnin, line 130: @khaliqgant Can you confirm those are the only two entry points for sync processing via Temporal? thank you

packages/jobs/lib/processor/handler.ts
```diff
@@ -29,13 +49,174 @@ export async function handler(task: OrchestratorTask): Promise<Result<JsonValue>
...
 
 async function sync(task: TaskSync): Promise<Result<JsonValue>> {
-    return Ok({ id: task.id });
+    const isGloballyEnabled = await featureFlags.isEnabled('orchestrator:schedule', 'global', false);
+    const isEnvEnabled = await featureFlags.isEnabled('orchestrator:schedule', `${task.connection.environment_id}`, false);
+    const isOrchestrator = isGloballyEnabled || isEnvEnabled;
+    if (!isOrchestrator) {
+        return Ok({ dryrun: true });
+    }
+    let logCtx: LogContext | undefined;
```
TBonnin, line 71: code below is mostly copied from activities.ts. I know @bodinsamuel that I am copying over legacy logs stuff. Tell me if I can avoid it right now. Thank you

packages/jobs/lib/activities.ts
TBonnin: Also is that ok to return `false` when sync is not processed via Temporal?

packages/jobs/lib/processor/handler.ts
bodinsamuel: I'll clean up, I have so many conflict anyway

packages/orchestrator/lib/clients/processor.ts
```diff
@@ -141,8 +143,18 @@ export class OrchestratorProcessor {
...
         });
     }
+
+    private heartbeat(task: OrchestratorTask): NodeJS.Timeout {
+        return setInterval(async () => {
+            const res = await this.orchestratorClient.heartbeat({ taskId: task.id });
+            if (res.isErr()) {
+                logger.error(`failed to send heartbeat for task ${task.id}: ${stringifyError(res.error)}`);
+            }
+        }, 300_000);
```
bodinsamuel, line 158: that's a long delay between heartbeat no? I would say 60s is already "huge"

bodinsamuel (approved):

packages/jobs/lib/activities.ts
khaliqgant: Correct that every sync that temporal performs has to go through this function. 

> Also is that ok to return false when sync is not processed via Temporal?

The return has no real consequence to the application. It just affects what is shown in temporal 
image

So it actually might be better to return `true`. It shouldn't show as a fail just because it didn't process the sync IMO.

packages/orchestrator/lib/clients/processor.ts
khaliqgant: This is the current setting. I think we should maintain the current setting to keep the number of moving parts to a minimum and change this if we think it is necessary later.

If there is suddenly an issue with not using temporal anymore this is one less thing we have to check on that was changed that could be causing issues.

khaliqgant (approved):

packages/jobs/lib/activities.ts
TBonnin: 👍 changing it to true. thank you [02ac105](https://github.com/NangoHQ/nango/pull/2315/commits/02ac10542cf1e5bbc411b244cb44cc51f159fab7)

packages/orchestrator/lib/clients/processor.ts
TBonnin: it is a long delay between heartbeat indeed but like Khaliq said I didn't want to change too many things. I am happy to lower it once we temporal is removed

PR #2316, opened 2024-06-12
chore: getting sync data from orchestrator

TBonnin:
Sync data like next run times and status should come from Orchestrator when flag is on.
Added an endpoint to search for schedules in order to get info about state of a schedule and its next due date


packages/shared/lib/services/sync/orchestrator.service.ts
```diff
@@ -559,21 +590,10 @@ export class OrchestratorService {
...
-    }
-
-    private reportedStatus(
-        sync: Sync,
-        latestJob: SyncJob | null,
-        schedule: SyncSchedule | null,
-        status: SyncStatus,
-        nextScheduledSyncAt?: string | Date | null,
-        includeJobStatus = false
-    ) {
```
TBonnin, line 576: this function was always used alongside the fetchSyncData. I merged them together into a syncStatus one

packages/shared/lib/services/sync/run.service.ts
```diff
@@ -905,7 +905,7 @@ export default class SyncRun {
         }
 
         await updateSuccessActivityLog(this.activityLogId, false);
-        await updateSyncJobStatus(this.syncJobId, SyncStatus.STOPPED);
+        await updateSyncJobStatus(this.syncJobId, SyncStatus.ERROR);
```
TBonnin: we are in `reportFailureForResults` function. let's set the status to error.

packages/shared/lib/services/sync/orchestrator.service.ts
```diff
@@ -583,7 +603,7 @@ export class OrchestratorService {
             status,
             frequency: schedule?.frequency,
             latestResult: latestJob?.result,
-            latestExecutionStatus: latestJob?.status === SyncStatus.STOPPED ? SyncStatus.ERROR : latestJob?.status
+            latestExecutionStatus: latestJob?.status
```
TBonnin, line 614: see my comment and change below

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -79,6 +82,20 @@ export class Orchestrator {
...
     }
 
+    async searchSchedules(props: { syncId: string; environmentId: number }[]): Promise<Result<Map<string, OrchestratorSchedule>>> {
+        const scheduleNames = props.map(({ syncId, environmentId }) => this.getScheduleName({ environmentId, syncId }));
+        const schedules = await this.client.searchSchedules({ scheduleNames, limit: scheduleNames.length });
+        if (schedules.isErr()) {
+            return Err(`Failed to get schedules: ${stringifyError(schedules.error)}`);
+        }
+        const scheduleMap = schedules.value.reduce((map, schedule) => {
+            const syncId = schedule.name.split(':').pop()!;
```
TBonnin: a bit convoluted but I don't have a better solution.

packages/shared/lib/services/sync/sync.service.ts
```diff
@@ -284,75 +288,99 @@ export const getSyncs = async (nangoConnection: Connection): Promise<(Sync & { s
...
-            } else if (!syncSchedule.schedule?.state?.paused && sync.schedule_status === SyncStatus.PAUSED) {
-                sync = {
+    const isGloballyEnabled = await featureFlags.isEnabled('orchestrator:schedule', 'global', false);
+    const isEnvEnabled = await featureFlags.isEnabled('orchestrator:schedule', `${nangoConnection.environment_id}`, false);
+    const isOrchestrator = isGloballyEnabled || isEnvEnabled;
+    if (isOrchestrator) {
+        const searchSchedulesProps = result.map((sync) => {
+            return { syncId: sync.id, environmentId: nangoConnection.environment_id };
+        });
+        const schedules = await orchestrator.searchSchedules(searchSchedulesProps);
```
TBonnin, line 300: Scheduler will be called by the UI (with auto-refresh). That might put a lot of pressure on the orchestrator service if a lot of connections pages stay open. We'll see if this is a problem. Ideally this call should be paginated but at least now we are doing a single request to fetch all schedules instead of a request per sync

packages/orchestrator/lib/clients/validate.ts
```diff
@@ -135,3 +135,48 @@ export function validateTask(task: Task): Result<OrchestratorTask> {
...
+        const schedule: OrchestratorSchedule = {
+            id: validation.data.id,
+            name: validation.data.name,
+            state: validation.data.state,
+            frequencyMs: validation.data.frequencyMs,
+            nextDueDate: getNextDueDate(validation.data.startsAt, validation.data.frequencyMs)
+        };
+        return Ok(schedule);
+    }
+    return Err(`Cannot validate task ${JSON.stringify(schedule)}: ${validation.error}`);
```
bodinsamuel, line 180: side note: that would be great if Err could take a payload because this is a lot of things that become unreadable when stringified

packages/orchestrator/lib/routes/v1/schedules/postSearch.ts
```diff
@@ -0,0 +1,54 @@
...
+    return async (req: EndpointRequest<PostSearch>, res: EndpointResponse<PostSearch>) => {
+        const { names, limit } = req.body;
+        const getSchedules = await scheduler.searchSchedules({
+            limit,
+            ...(names ? { names } : {})
+        });
+        if (getSchedules.isErr()) {
+            return res.status(500).json({ error: { code: 'search_failed', message: getSchedules.error.message } });
+        }
+        return res.status(201).json(getSchedules.value);
```
bodinsamuel: Seems unintentional 
```suggestion
        return res.status(200).json(getSchedules.value);
```

packages/orchestrator/lib/routes/v1/tasks/postSearch.ts
```diff
@@ -34,7 +34,7 @@ const validate = validateRequest<PostSearch>({
 const handler = (scheduler: Scheduler) => {
     return async (req: EndpointRequest<PostSearch>, res: EndpointResponse<PostSearch>) => {
         const { ids, groupKey, limit } = req.body;
-        const getTasks = await scheduler.search({
+        const getTasks = await scheduler.searchTasks({
             ...(ids ? { ids } : {}),
             ...(groupKey ? { groupKey } : {}),
             ...(limit ? { limit } : {})
```
bodinsamuel, line 40: can't comment the status code here, but same comment

packages/shared/lib/clients/orchestrator.ts
bodinsamuel: your name seems to follow `property:value` pattern, maybe a unserializer that split and output an object would clarify `environment:2:connection:9` => `{environment: 2, connection: 9}`

packages/shared/lib/services/sync/orchestrator.service.ts
```diff
@@ -511,13 +517,42 @@ export class OrchestratorService {
...
 
+    private async syncStatus(sync: Sync, environmentId: number, includeJobStatus: boolean, orchestrator: Orchestrator): Promise<ReportedSyncJobStatus> {
+        const isGloballyEnabled = await featureFlags.isEnabled('orchestrator:schedule', 'global', false);
+        const isEnvEnabled = await featureFlags.isEnabled('orchestrator:schedule', `${environmentId}`, false);
+        const isOrchestrator = isGloballyEnabled || isEnvEnabled;
+        if (isOrchestrator) {
+            const latestJob = await getLatestSyncJob(sync.id);
+            const schedules = await orchestrator.searchSchedules([{ syncId: sync.id, environmentId }]);
+            if (schedules.isErr()) {
+                throw new Error(`Failed to get schedule for sync ${sync.id} in environment ${environmentId}: ${schedules.error}`);
```
bodinsamuel: is error an Error object? if yes, needs a stringification

packages/shared/lib/services/sync/sync.service.ts
```diff
@@ -284,75 +288,99 @@ export const getSyncs = async (nangoConnection: Connection): Promise<(Sync & { s
...
+        if (schedules.isErr()) {
+            throw new Error(`Failed to get schedules for environment ${nangoConnection.environment_id}: ${schedules.error}`);
```
bodinsamuel: same comment about error stringifcation

bodinsamuel (commented):
Looks good, few minor comments

bodinsamuel:
Sorry I pushed in your PR by accident :D

packages/orchestrator/lib/clients/validate.ts
khaliqgant: Extended error orchestrator specific object?

packages/shared/lib/services/sync/orchestrator.service.ts
```diff
@@ -23,15 +23,7 @@ import SyncClient from '../../clients/sync.client.js';
 import configService from '../config.service.js';
 import type { LogLevel } from '../../models/Activity.js';
 import type { Connection, NangoConnection } from '../../models/Connection.js';
```
khaliqgant, line 25: This file should probably be renamed in an effort to reduce confusion with the orchestrator package.

khaliqgant (approved):
Failing test.

Non blocking feedback. Would like to do a thorough review once schedules can be processed by the orchestrator. Can you let me know once this is possble?

packages/shared/lib/services/sync/orchestrator.service.ts
TBonnin: you are right. Its functions are createSyncs, deleteSync, getSyncStatus, etc... it is basically a sync manager. I am not a huge fan of the work manager but I don't have a better idea so I am gonna use that for now. Feel free to recommend anything else

packages/orchestrator/lib/routes/v1/schedules/postSearch.ts
TBonnin: af3a5875dca2dc6060f05dea3212f17870c623e2 alongside fixes for other orchestrator responses

packages/shared/lib/services/sync/orchestrator.service.ts
TBonnin: b28e95a150d4e15f74441871a2d03070ef061627

packages/shared/lib/services/sync/sync.service.ts
TBonnin: b28e95a150d4e15f74441871a2d03070ef061627

packages/shared/lib/services/sync/orchestrator.service.ts
TBonnin: 0476c68391f9d30f38e9285ce3b0eef3466d945c

packages/shared/lib/clients/orchestrator.ts
TBonnin: fe9c4ff51b79d9bd69ae17cc23bca41303f42260

packages/orchestrator/lib/clients/validate.ts
bodinsamuel: actually with our afternoon meeting, seems like Err could stay like this and the only param could be an `extends Error` with or without a payload

TBonnin: Err can take anything that extends Error

bodinsamuel: yep I was saying that we don't need to change Err if go this path :p

bodinsamuel (approved):

PR #2317, opened 2024-06-13
fix: event should be emitted when task is created by scheduling worker

TBonnin:
Created task events are useful for long polling to be instantly notified of a new created tasks.

### DIFF
--- a/packages/scheduler/lib/workers/monitor/monitor.worker.ts
+++ b/packages/scheduler/lib/workers/monitor/monitor.worker.ts
@@ -7,7 +7,7 @@ import { setTimeout } from 'node:timers/promises';
 import type knex from 'knex';
 import { logger } from '../../utils/logger.js';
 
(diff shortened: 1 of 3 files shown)


packages/scheduler/lib/workers/monitor/monitor.worker.ts
```diff
@@ -7,7 +7,7 @@ import { setTimeout } from 'node:timers/promises';
 import type knex from 'knex';
 import { logger } from '../../utils/logger.js';
 
-interface MessageOut {
+interface ExpiredTasksMessage {
```
TBonnin, line 10: unrelated but wanted to be consistent with the naming of the same notifying mechanism in the monitor and scheduling worker

bodinsamuel (approved):

khaliqgant (approved):

PR #2319, opened 2024-06-13
feat(webhooks): [nan-1142] update options on environment settings to allow more use settings

khaliqgant:
## Describe your changes
- Updates the UI to use the external webhooks table which was already successfully migrated
- Updates types in the node-client, specifically the webhook success and error objects. 
- Updates routes to use the url and settings controller files for better validation.

Note that the environment settings for webhook columns will be removed in a future PR.

NAN-1142


TBonnin:
I'll review on Friday. Sorry about the delay

packages/node-client/lib/types.ts
```diff
@@ -1,4 +1,14 @@
 import type { ParamsSerializerOptions } from 'axios';
+import type { NangoSyncWebhookBodySuccess, NangoSyncWebhookBodyError, NangoAuthWebhookBodySuccess, NangoAuthWebhookBodyError } from '@nangohq/types';
```
bodinsamuel, line 2: :feels-good:

packages/node-client/lib/types.ts
```diff
@@ -1,4 +1,14 @@
...
+
+export type {
+    NangoSyncWebhookBodySuccess,
+    NangoSyncWebhookBodySuccess as NangoSyncWebhookBody,
```
bodinsamuel, line 7: You can put (if you want) /** @deprecated use NangoSyncWebhookBodySuccess */

packages/shared/lib/services/external-webhook.service.ts
```diff
@@ -0,0 +1,25 @@
...
+export async function update(id: number, data: WebhookSettings): Promise<void> {
+    await db.knex.from<ExternalWebhook>('_nango_external_webhooks').where({ environment_id: id }).update({
+        on_sync_completion_always: data.alwaysSendWebhook,
+        on_auth_creation: data.sendAuthWebhook,
+        on_auth_refresh_error: data.sendRefreshFailedWebhook,
+        on_sync_error: data.sendSyncFailedWebhook
+    });
+}
+
+export async function updatePrimaryUrl(id: number, primaryUrl: string): Promise<void> {
```
bodinsamuel: `id` is ambiguous here, its environmentId not webhookId (comment for all methods here)

packages/server/lib/controllers/v1/environment/webhook/settings.ts
```diff
@@ -0,0 +1,39 @@
+import { z } from 'zod';
```
bodinsamuel: updateSettings.ts or patchSettings.ts would clarify (the goal is to have one endpoint per file)

packages/server/lib/routes.ts
```diff
@@ -168,17 +170,16 @@ web.route('/api/v1/account/admin/switch').post(webAuth, accountController.switch
 
 web.route('/api/v1/environment').get(webAuth, environmentController.getEnvironment.bind(environmentController));
 web.route('/api/v1/environment/callback').post(webAuth, environmentController.updateCallback.bind(environmentController));
-web.route('/api/v1/environment/webhook').post(webAuth, environmentController.updateWebhookURL.bind(environmentController));
-web.route('/api/v1/environment/webhook-secondary').post(webAuth, environmentController.updateSecondaryWebhookURL.bind(environmentController));
+web.route('/api/v1/environment/webhook/url/primary').patch(webAuth, updatePrimaryUrl);
+web.route('/api/v1/environment/webhook/url/secondary').patch(webAuth, updateSecondaryUrl);
```
bodinsamuel: imo: still seems unnecessary vs `/webhook/url/:name`

packages/types/lib/environment/api/webhook.ts
```diff
@@ -0,0 +1,46 @@
...
+    Querystring: {
+        env: string;
+    };
+    Path: '/api/v1/environment/webhook/url/primary';
+    Body: {
+        url: string;
+    };
+    Success: {
+        url: string;
+    };
```
bodinsamuel, line 33: Regarding api design I would be very careful of just outputing data in the main object, if any fields is named `error` then typescript will be confused. Also it will return inconsistent data type (if you data is an array then suddenly the object is now an array).And it's easier and more consistent to always answer `{ error: Error } | { data: any }` (or something like that)

packages/types/lib/webhooks/api.ts
```diff
@@ -0,0 +1,59 @@
+import type { AuthOperationType, AuthModeType } from '../auth/api.js';
```
bodinsamuel, line 1: I guess everything can be called API but it's confusing with other api.ts (maybe we should rename other files to http.api.ts

packages/server/lib/controllers/v1/environment/webhook/url.ts
```diff
@@ -0,0 +1,61 @@
+import { z } from 'zod';
+import { asyncWrapper } from '../../../../utils/asyncWrapper.js';
+import type { UpdatePrimaryUrl, UpdateSecondaryUrl } from '@nangohq/types';
+import { requireEmptyQuery, zodErrorToHTTP } from '@nangohq/utils';
+import { externalWebhookService } from '@nangohq/shared';
+
+const validation = z
+    .object({
+        url: z.string()
```
bodinsamuel: z.string().url()

packages/shared/lib/services/external-webhook.service.ts
```diff
@@ -0,0 +1,25 @@
...
+import type { ExternalWebhook, WebhookSettings } from '@nangohq/types';
+
+export async function get(id: number): Promise<ExternalWebhook | null> {
+    const result = await db.knex.select('*').from<ExternalWebhook>('_nango_external_webhooks').where({ environment_id: id }).first();
+
+    return result || null;
+}
+
+export async function update(id: number, data: WebhookSettings): Promise<void> {
+    await db.knex.from<ExternalWebhook>('_nango_external_webhooks').where({ environment_id: id }).update({
```
bodinsamuel: At no point we seems create a row for `_nango_external_webhooks`, it will break for new user or I missed something

packages/webhooks/lib/utils.ts
```diff
@@ -115,17 +146,23 @@ export const deliver = async ({
...
+                    }
                 } else {
-                    await logCtx?.error(
+                    await logCtx.error(
                         `${webhookType} sent webhook successfully to the ${type} ${url} but received a ${response.status} response code${endingMessage ? ` ${endingMessage}` : ''}. Please send a 2xx on successful receipt.`,
                         body as Record<string, unknown>
                     );
+                    if (webhookType === 'forward') {
+                        await logCtx.failed();
+                    }
```
bodinsamuel: I would advocate for never calling failed/success deep inside methods, it's hard to follow the context everywhere. This function could return a boolean or a res, so you could at least do it closer to the creation

packages/webhooks/lib/forward.ts
```diff
@@ -0,0 +1,83 @@
...
+                ...body,
+                connectionId
+            },
+            webhookType: 'forward',
+            activityLogId: null,
+            environment,
+            logCtx,
+            incomingHeaders: webhookOriginalHeaders
+        });
+    }
```
bodinsamuel, line 90: logCtx is never success/failed

bodinsamuel (commented):
Works locally 📤
Minor comments

packages/node-client/lib/types.ts
khaliqgant: Updated

packages/shared/lib/services/external-webhook.service.ts
khaliqgant: Updated.

packages/server/lib/controllers/v1/environment/webhook/settings.ts
khaliqgant: Uodated

packages/server/lib/routes.ts
khaliqgant: Fair, but kept it the same.

packages/types/lib/environment/api/webhook.ts
khaliqgant: Yeah, good point. Updated

packages/server/lib/controllers/v1/environment/webhook/url.ts
khaliqgant: Thanks, updated

packages/shared/lib/services/external-webhook.service.ts
khaliqgant: Great catch! Updated.

packages/webhooks/lib/utils.ts
khaliqgant: Updated

packages/webhooks/lib/forward.ts
khaliqgant: Updated with the above https://github.com/NangoHQ/nango/pull/2319#discussion_r1639592905

bodinsamuel (approved):

packages/node-client/package.json
```diff
@@ -36,6 +36,7 @@
         "README.md"
     ],
     "devDependencies": {
+        "@nangohq/types": "^0.40.0",
```
TBonnin, line 39: should we use `../types` for devDependencies?

packages/webhooks/lib/sync.ts
```diff
@@ -1,101 +1,114 @@
...
-        body.startedAt = dayjs(now).toDate().toISOString();
-        body.failedAt = new Date().toISOString();
-    }
+        successBody = {
+            ...body,
+            success: true,
+            responseResults: {
+                added: responseResults.added,
+                updated: responseResults.updated,
+                deleted: 0
```
TBonnin, line 84: why `deleted: 0` and not `deleted: responseResults.deleted || 0`

TBonnin (approved):

packages/node-client/package.json
khaliqgant: Since node-client is published we have to use the published version and not the local one

TBonnin: does `npm install` also install the devDependencies?

khaliqgant: Yes, unless you pass a `--production` flag or if `NODE_ENV` environment variable is set to `production`

packages/webhooks/lib/sync.ts
khaliqgant: fair, but i'll leave as is

PR #2321, opened 2024-06-13
chore(logger): handle object, better local output

bodinsamuel:
## Describe your changes

Got tired of getting caught by logger.info() that doesn't accept a second param.

- You can now use `logger.log('message', { data })` 
- Setup color locally, let me know what you think. I feel like it improves readability
- Set `concurrently` names (only 3 letters to keep stable and short) to identify which process is logging
- In production removes timestamp because it's already set by the provider
- `getLogger` now returns a proper child. 
  I don't know if there was a performance hit by creating N logger (most probably there but small) but at least it's now gone

Screenshot 2024-06-13 at 16 29 56


TBonnin (approved):
looks nice 🎨. love the colors

khaliqgant (approved):

PR #2322, opened 2024-06-13
chore: disable sync retry

TBonnin:
the goal is to avoid the noise of retrying tasks while migrating to orchestrator.


bodinsamuel (approved):
Good for me, follows what we have currently with temporal

PR #2323, opened 2024-06-13
fix(slack): only notify success when refreshing

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1200/slack-message-are-always-resolved-even-if-incorrect

- `onRefreshSuccess` was called at every getConnectionCredentials, even when no refresh were happening
That means a proxy call would set success but then a refresh would set failed, back and forth until the end of time :D
Let me know if that makes sense @khaliqgant 

- Added more logs just so we can debug this quicker in prod

### DIFF
--- a/packages/shared/lib/services/connection.service.ts
+++ b/packages/shared/lib/services/connection.service.ts
@@ -708,18 +708,18 @@ class ConnectionService {
                 const errorWithPayload = new NangoError(error.type, connection);
 
                 return Err(errorWithPayload);
+            } else {
+                await onRefreshSuccess({
+                    connection,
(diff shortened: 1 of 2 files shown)


khaliqgant (approved):
Thanks!

So refreshes are being incorrectly marked as resolved which means there is still an underlying issue with the connection refresh not working for that connection, correct?

bodinsamuel:
>So refreshes are being incorrectly marked as resolved which means there is still an underlying issue with the connection refresh not working for that connection, correct?

Yes, looking at the log their refresh_token is just expired

PR #2324, opened 2024-06-13
fix(webhooks): [nan-1204] make queryTimestamp always present

khaliqgant:
## Describe your changes
Make queryTimestamp always present

NAN-1204

### DIFF
--- a/packages/webhooks/lib/sync.ts
+++ b/packages/webhooks/lib/sync.ts
@@ -58,6 +58,7 @@ export const sendSync = async ({
     };
 
     if (success) {
+        body.queryTimeStamp = null;
         body.modifiedAfter = dayjs(now).toDate().toISOString();
         if (syncType !== 'INITIAL') {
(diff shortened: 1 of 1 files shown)


packages/webhooks/lib/sync.ts
```diff
@@ -58,6 +58,7 @@ export const sendSync = async ({
     };
 
     if (success) {
+        body.queryTimeStamp = null;
```
TBonnin, line 61:
```suggestion
        body.queryTimeStamp = syncType !== 'INITIAL' ? now as unknown as string : null;
```
and remove the if below

TBonnin (approved):

bodinsamuel (approved):

packages/webhooks/lib/sync.ts
khaliqgant: Updated

PR #2325, opened 2024-06-13
fix(logs): missing success for syncs

bodinsamuel:
## Describe your changes

- `finishFlow` is called by all scripts but this part was only relevant for syncs
- Misread the reportResults thinking it was only for action, anyway the success was at the wrong place so it's still better now :D

### DIFF
--- a/packages/shared/lib/services/sync/run.service.ts
+++ b/packages/shared/lib/services/sync/run.service.ts
@@ -611,27 +611,31 @@ export default class SyncRun {
 
     async finishFlow(models: string[], syncStartDate: Date, version: string, totalRunTime: number, trackDeletes?: boolean): Promise<void> {
         let i = 0;
-        for (const model of models) {
-            let deletedKeys: string[] = [];
-            if (!this.isWebhook && trackDeletes) {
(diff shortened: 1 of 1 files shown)


packages/shared/lib/services/sync/run.service.ts
```diff
@@ -611,27 +611,31 @@ export default class SyncRun {
 
     async finishFlow(models: string[], syncStartDate: Date, version: string, totalRunTime: number, trackDeletes?: boolean): Promise<void> {
         let i = 0;
-        for (const model of models) {
-            let deletedKeys: string[] = [];
-            if (!this.isWebhook && trackDeletes) {
-                deletedKeys = await this.recordsService.markNonCurrentGenerationRecordsAsDeleted({
-                    connectionId: this.nangoConnection.id as number,
+
+        if (!this.isAction && !this.isWebhook && !this.isPostConnectionScript) {
```
TBonnin, line 615: no `isSync`?

TBonnin (approved):

packages/shared/lib/services/sync/run.service.ts
bodinsamuel: does not exist, I didn't wanted to start a refacto now :D

PR #2326, opened 2024-06-13
chore: one-off script to migrate undeleted schedules to orchestrator db

TBonnin:
tested in staging

### DIFF
--- a/scripts/one-off/schedules-migration/migrate.ts
+++ b/scripts/one-off/schedules-migration/migrate.ts
@@ -1,0 +1,159 @@
+import Knex from 'knex';
+import ms from 'ms';
+import { uuidv7 } from 'uuidv7';
+
+const SOURCE_DB_URL = process.env['SOURCE_DB_URL'] || '';
+const sourceKnex = Knex({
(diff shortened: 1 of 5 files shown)


scripts/one-off/schedules-migration/migrate.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    const schedules = await readSchedules();
+    for (const schedule of schedules) {
+        try {
+            const res = await writeSchedule(schedule);
+            console.log('Schedule migrated:', res);
+        } catch (err) {
+            console.error('Error migrating schedule:', err);
+        }
+        await new Promise((resolve) => setTimeout(resolve, 1000));
```
TBonnin, line 143: Not absolutely necessary but waiting so the scheduling worker doesn't have to handle all new schedules at the same time.

khaliqgant:
We'll have to document this so that evertrue can run this in their infra as well. A more robust README.md would be great

bodinsamuel (commented):
🕶️

bodinsamuel (approved):

PR #2331, opened 2024-06-14
fix: Orchestrator bugs

TBonnin:
a couple of small bugfixes for the orchestrator. Each commit should be self-explanatory

### DIFF
--- a/packages/orchestrator/lib/clients/validate.ts
+++ b/packages/orchestrator/lib/clients/validate.ts
@@ -3,7 +3,7 @@ import type { Schedule, Task } from '@nangohq/scheduler';
 import type { OrchestratorSchedule, OrchestratorTask } from './types.js';
 import { TaskAction, TaskWebhook, TaskPostConnection, TaskSync } from './types.js';
 import { z } from 'zod';
-import { Err, Ok } from '@nangohq/utils';
+import { Err, Ok, stringifyError } from '@nangohq/utils';
 import type { Result } from '@nangohq/utils';
 import { jsonSchema } from '../utils/validation.js';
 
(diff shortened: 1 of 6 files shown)


packages/scheduler/lib/scheduler.ts
```diff
@@ -330,18 +330,24 @@ export class Scheduler {
...
-            if (!schedule.value[0]) {
+            if (!found.value[0]) {
                 return Err(`Schedule '${scheduleName}' not found`);
             }
+            const schedule = found.value[0];
+
+            if (schedule.state === state) {
+                // No-op if the schedule is already in the desired state
+                return Ok(schedule);
+            }
```
TBonnin, line 347: making `setScheduleState` a no-op if newState == currentState

packages/scheduler/lib/workers/scheduling/scheduling.worker.ts
```diff
@@ -114,9 +114,9 @@ export class SchedulingChild {
                 });
                 if (task.isErr()) {
                     logger.error(`Failed to create task for schedule: ${schedule.id}`);
-                    return;
```
TBonnin, line 117: continuing  on task creation failure

packages/orchestrator/lib/clients/validate.ts
```diff
@@ -133,7 +133,7 @@ export function validateTask(task: Task): Result<OrchestratorTask> {
             })
         );
     }
-    return Err(`Cannot validate task ${JSON.stringify(task)}: ${action.error || webhook.error || postConnection.error}`);
+    return Err(`Cannot validate task ${JSON.stringify(task)}: ${sync.error || action.error || webhook.error || postConnection.error}`);
```
bodinsamuel: none of there are stringified? :D

packages/orchestrator/lib/events.ts
```diff
@@ -23,10 +23,11 @@ export class EventsHandler extends EventEmitter {
         this.onCallbacks = {
             CREATED: (task: Task) => {
                 on.CREATED(task);
-                this.emit(`task:started:${task.groupKey}`, task);
+                this.emit(`task:created:${task.groupKey}`, task);
```
bodinsamuel, line 26: oops, after 15 years I still can't understand how we can read the same lines over and over, and still miss stuff like this ahah Brain is goofy

bodinsamuel (approved):

PR #2258, opened 2024-06-04
feat(logs): remove v1

bodinsamuel:
## Describe your changes

Fixes NAN-585

PR #2332, opened 2024-06-14
chore(logs.v1): remove from persist

bodinsamuel:
## Describe your changes

Split of #2258

Since it's not possible to merge a full removal I'll split. This one is small but should remove 80% of the traffic to the table.

- Remove v1 from persist
- Allows mix of string | number during the transition
We still send the old activityLogId so it doesn't change anything, but during the transition some will be up to date and not others.


khaliqgant (approved):

packages/persist/lib/server.integration.test.ts
```diff
@@ -3,8 +3,8 @@ import { server } from './server.js';
 import fetch from 'node-fetch';
 import type { AuthCredentials, Connection, Sync, Job as SyncJob, Environment, Account } from '@nangohq/shared';
 import db, { multipleMigrations } from '@nangohq/database';
-import { createActivityLog, environmentService, connectionService, createSync, createSyncJob, SyncType, SyncStatus, accountService } from '@nangohq/shared';
-import { logContextGetter } from '@nangohq/logs';
+import { environmentService, connectionService, createSync, createSyncJob, SyncType, SyncStatus, accountService } from '@nangohq/shared';
+import { logContextGetter, migrateMapping } from '@nangohq/logs';
```
TBonnin: what about`migrateMapping as migrateLogsMapping` so when you read the code without the import it is clearer what mappings we are talking about?

TBonnin (approved):

packages/persist/lib/server.integration.test.ts
bodinsamuel: true

PR #2333, opened 2024-06-14
fix(slack): only notify success if it has refreshed (again)

bodinsamuel:
## Describe your changes

Fix of the previous fix.  A little omission when I read the code,`refreshCredentialsIfNeeded` always return credentials wether it was refreshed or not.

- Correctly notify success if it has been refreshed (hopefully good this time)

### DIFF
--- a/packages/shared/lib/services/connection.service.ts
+++ b/packages/shared/lib/services/connection.service.ts
@@ -639,11 +639,7 @@ class ConnectionService {
         const template: ProviderTemplate = configService.getTemplate(config?.provider);
 
         if (connection?.credentials?.type === 'OAUTH2' || connection?.credentials?.type === 'APP' || connection?.credentials?.type === 'OAUTH2_CC') {
-            const {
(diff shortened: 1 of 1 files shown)


packages/shared/lib/services/connection.service.ts
```diff
@@ -815,7 +811,7 @@ class ConnectionService {
         template: ProviderTemplateOAuth2;
         environment_id: number;
         instantRefresh?: boolean;
-    }): Promise<ServiceResponse<OAuth2Credentials | AppCredentials | AppStoreCredentials | OAuth2ClientCredentials>> {
+    }): Promise<ServiceResponse<{ refreshed: boolean; credentials: OAuth2Credentials | AppCredentials | AppStoreCredentials | OAuth2ClientCredentials }>> {
```
khaliqgant, line 814: This method should be marked as private

khaliqgant (approved):

PR #2334, opened 2024-06-14
feat(integrations): [nan-1187] add asana templates

khaliqgant:
## Describe your changes
Add asana templates with a unified `Task` and `User` model

Since the compile file is what gets copied over to a specific users integrations, even if it uses relative imports it still should _just work_. Needs confirmation.

NAN-1187

### DIFF
--- a/integration-templates/asana/syncs/tasks.ts
+++ b/integration-templates/asana/syncs/tasks.ts
@@ -1,65 +1,53 @@
-import type { AsanaTask, NangoSync } from '../../models';
+import type { NangoSync, BaseAsanaModel, AsanaTask, Task } from '../../models';
+// eslint-disable-next-line import/extensions
+import { toUser } from '../mappers/to-user';
+// eslint-disable-next-line import/extensions
+import { toTask } from '../mappers/to-task';
 
-export default async function fetchData(nango: NangoSync) {
-    // Get the user's workspaces & projects
-    // For testing we just get the first project of the first workspace
-    const workspaces = await paginate(nango, '/api/1.0/workspaces');
-    const workspace = workspaces[0];
+export default async function fetchData(nango: NangoSync): Promise<void> {
+    const lastSyncDate = nango.lastSyncDate;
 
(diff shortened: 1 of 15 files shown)


packages/shared/lib/services/flow.service.unit.test.ts
```diff
@@ -157,28 +157,6 @@ describe('Flow service tests', () => {
...
-`,
-                    endpoint: '/asana/tasks',
-                    output: 'AsanaTask',
-                    runs: 'every 30min',
-                    scopes: ['default'],
-                    sync_type: 'full'
-                }
-            }
-        });
+        expect(flows.integrations).toHaveProperty('github');
```
bodinsamuel, line 160: I found this test useful, helped me found some bugs in the parser

integration-templates/asana/syncs/tasks.ts
```diff
@@ -1,65 +1,53 @@
...
-            callParams['offset'] = `${page}`;
-        }
-
-        const resp = await nango.get({
-            endpoint: endpoint,
-            params: {
-                limit: `${MAX_PAGE}`,
-                ...callParams
+import type { NangoSync, BaseAsanaModel, AsanaTask, Task } from '../../models';
+// eslint-disable-next-line import/extensions
```
bodinsamuel: why? .js should work everywhere no?

bodinsamuel (approved):

packages/shared/lib/services/flow.service.unit.test.ts
khaliqgant: Ok cool, I'll add something back similar

khaliqgant: Added in a test for algolia since the Asana object is much more verbose now

integration-templates/asana/syncs/tasks.ts
khaliqgant: Yeah, updated. Had to tweak the logic some to accommodate

bodinsamuel (approved):
lgtm

PR #2335, opened 2024-06-14
fix: remove orchestrator event emmitter max listeners limit

TBonnin:
EventEmmitter max listeners default limit is 10.
In our case each processor dequeuing for a given group_key adds a listener so we need the limit to be higher.

I know each listener is adding performance and memory overheard but for now orchestrator runs on half a gig memory machine and using less than 30% of memory so we have some margin


bodinsamuel (approved):
I'm not too worried about the memory impact

PR #2242, opened 2024-06-01
feat(google-mail): action script to send an email using google-mail integration

Pyloris:
## Describe your changes
- Modified nango.yaml file of `/integration-templates/google-mail/nango.yaml` and added 2 models.
- Added an Action script `/integration-templates/google-mail/actions/send-email.ts` for sending an email.
- 
## Issue ticket number and link
#2241
ISSUE TICKET: [NAN-1075](https://linear.app/nango/issue/NAN-1075/google-mail-integration-does-not-support-sending-email)

PR #2337, opened 2024-06-14
fix(integration-templates): add workflow and edit existing

khaliqgant:
## Describe your changes
With relative imports the logic needs to be adjusted some. Also adds a workflow to upload all templates as a safeguard as when a template is made by an external contributor it doesn't get uploaded due to lack of permissions

### DIFF
--- a/.github/workflows/integration-template-upload.yaml
+++ b/.github/workflows/integration-template-upload.yaml
@@ -39,11 +39,20 @@ jobs:
               if: steps.changed-files-specific.outputs.any_changed == 'true'
               run: |
                   wd=$(pwd)
+
+                  echo "Changed files:"
+                  echo "${{ steps.changed-files-specific.outputs.all_changed_files }}"
+
                   for file in ${{ steps.changed-files-specific.outputs.all_changed_files }}; do
                     dir=$(dirname $file)
 
-                    dir=${dir//\/syncs/}
-                    dir=${dir//\/actions/}
+                    parts=(${dir//\// })
+                    if [ ${#parts[@]} -gt 1 ]; then
+                      unset parts[-2]
+                    fi
+
+                    dir=$(IFS=/; echo "${parts[*]}")
+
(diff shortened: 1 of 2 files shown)


.github/workflows/integration-template-upload.yaml
```diff
@@ -39,11 +39,20 @@ jobs:
...
                     dir=$(dirname $file)
 
-                    dir=${dir//\/syncs/}
-                    dir=${dir//\/actions/}
+                    parts=(${dir//\// })
+                    if [ ${#parts[@]} -gt 1 ]; then
+                      unset parts[-2]
+                    fi
+
+                    dir=$(IFS=/; echo "${parts[*]}")
```
TBonnin, line 54: if you want more bash magic I think you can do this with a one liner:
```
dir=${dir%/*/*}/${dir##*/}
```
Removing the last 2 elements of the path with the first curly bracket and re-adding the last element in the second

.github/workflows/sync-integration-templates.yaml
```diff
@@ -0,0 +1,39 @@
+name: 'Sync Integration Templates'
```
TBonnin, line 1: I am not sure I follow why we need this extra action. Aren't the merge triggering the templates upload action above done by us, even on external contributions? if yes what's the permissions issue?

khaliqgant: It is more of a fail safe than anything. It’s a force sync of all templates since the regular one looks just for changed files. 

For some reason this action contribution https://github.com/NangoHQ/nango/pull/2242 didnt get uploaded to s3. In that instance we need a way to sync integration templates. I need to look into exactly why it failed but I assumed it was a permissions issue but not 100% sure.

bodinsamuel: suggestion for future: this could be done at the start of the server process instead

bodinsamuel (approved):

PR #2341, opened 2024-06-14
fix(orchestrator): accept every in sync interval

TBonnin:
I was not using the cleanInterval function in the scenario of a deploy

### DIFF
--- a/packages/shared/lib/clients/orchestrator.ts
+++ b/packages/shared/lib/clients/orchestrator.ts
@@ -728,8 +728,26 @@ export class Orchestrator {
 
         // Orchestrator
         const scheduleName = ScheduleName.get({ environmentId, syncId });
-        const frequencyMs = ms(interval as StringValue);
-        const res = await this.client.updateSyncFrequency({ scheduleName, frequencyMs });
+
+        const cleanInterval = this.cleanInterval(interval);
+        if (isOrchestrator && cleanInterval.isErr()) {
+            errorManager.report(cleanInterval.error, {
+                source: ErrorSourceEnum.CUSTOMER,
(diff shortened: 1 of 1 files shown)


packages/shared/lib/clients/orchestrator.ts
```diff
@@ -728,8 +728,26 @@ export class Orchestrator {
...
+                metadata: {
+                    syncName,
+                    scheduleName,
+                    interval
+                }
+            });
+            return Err(cleanInterval.error);
+        }
+        let res: Result<void> = Ok(undefined);
+        if (cleanInterval.isOk()) {
```
TBonnin, line 747: It is a bit convoluted but this `if` is only necessary because of the flag. Will go away with temporal

khaliqgant: I don't follow completely but since it blocking for a customer -- approving.

khaliqgant (approved):

packages/shared/lib/clients/orchestrator.ts
khaliqgant: Looked at this again. cleanInterval is a weird name to me. At first I read it as clearInterval. But the function is checking the interview and making sure it is valid right? 

validateInterval?
validateFrequency

khaliqgant: formatInterval

TBonnin: My initial though was it is a function that takes Nango interval and "clean" it for `ms` but I am happy to rename it. I will do that when I remove the flag

khaliqgant: Got it. Thanks

PR #2345, opened 2024-06-17
fix(oauth2_cc): add necessary parameters when creating connection

hassan254-prog:
## Describe your changes
Add necessary parameters for generating and refreshing an access_token in `oauth2_cc`.

### DIFF
--- a/packages/shared/lib/services/connection.service.ts
+++ b/packages/shared/lib/services/connection.service.ts
@@ -1049,7 +1049,6 @@ class ConnectionService {
         const headers: Record<string, string> = {};
         const params = new URLSearchParams();
 
-        console.log('hassan');
         const bodyFormat = template.body_format || 'form';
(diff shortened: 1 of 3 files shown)


packages/shared/lib/services/connection.service.ts
```diff
@@ -788,7 +788,7 @@ class ConnectionService {
 
                 const oauth2Creds: OAuth2ClientCredentials = {
                     type: 'OAUTH2_CC',
-                    token: rawCreds['token'],
+                    token: rawCreds['access_token'] || rawCreds['data']['token'],
```
hassan254-prog, line 791: To ensure that `personio` does not break for other users.

khaliqgant:
Will try and test this tomorrow to make sure nothing breaks with these changes

khaliqgant (approved):

PR #2347, opened 2024-06-17
fix(webhooks): [nan-1142] allow empty string for url

khaliqgant:
## Describe your changes
Allow a webhook to be deleted

NAN-1142


packages/server/lib/controllers/v1/environment/webhook/updatePrimaryUrl.ts
```diff
@@ -6,7 +6,7 @@ import { externalWebhookService } from '@nangohq/shared';
 
 const validation = z
     .object({
-        url: z.string().url()
+        url: z.string().url().or(z.literal(''))
```
bodinsamuel, line 9: could have been optional() maybe but that works too

bodinsamuel (approved):

PR #2349, opened 2024-06-17
chore(logs.v1): remove API, proxy

bodinsamuel:
## Describe your changes

- Remove old API and associated models
- Remove activities from proxy
- Rewrite logs buffer to take advantage of the new format
- Fix logging when there was proxy configuration error


packages/server/lib/controllers/proxy.controller.ts
```diff
@@ -138,70 +97,41 @@ class ProxyController {
...
-                res,
-                method: method as HTTP_VERB,
-                configBody: proxyConfig,
-                activityLogId,
-                environment_id: environment.id,
-                isSync,
-                isDryRun,
-                logCtx
-            });
+            await this.sendToHttpMethod({ res, method: method as HTTP_VERB, configBody: proxyConfig, logCtx });
```
TBonnin, line 134: I like how it simplifies the function signature

packages/shared/lib/sdk/sync.ts
```diff
@@ -427,26 +427,21 @@ export class NangoAction {
...
-                for (const log of activityLogs) {
+            // We batch save, since we have buffered the createdAt it shouldn't impact order
+            await Promise.all(
+                logs.map(async (log) => {
                     if (log.level === 'debug') {
-                        continue;
+                        return;
                     }
-
-                    if (!this.dryRun) {
```
TBonnin, line 443: Where is dryRun being handled now?

TBonnin (approved):

packages/shared/lib/sdk/sync.ts
bodinsamuel: in the logCtx creation, the `dryRun` parameter. That's mostly why I embedded it so it's transparent and one less param to carry around

PR #2352, opened 2024-06-17
chore: prevents two scheduling workers to run at the same time

TBonnin:
I thought I will be able to get away with implementing a mechanism to prevent two scheduling workers to run at the same time because we for now have a single instance of the orchestrator but in fact during deployment there is a short period where there are more than one instance running.

Having more than one scheduling workers at the same time might have unexpected consequences and race conditions.
This commit implements a simple distributed leader election backed by postgres to ensure there is never more than one scheduling worker running at once. 
A worker can acquire a lease by inserting a row in the db. Any other worker will see its lease acquisition fail. Then the mechanism relies on one assumption: the leader always renew its lease before it expires.
While the lease is valid only the current leader can renew it. Other workers can try to be elected leader, it will only be successful if the lease has expired.

first part of https://linear.app/nango/issue/NAN-1230/orchestrator-support-multiple-instances

### DIFF
--- a/packages/orchestrator/lib/clients/client.integration.test.ts
+++ b/packages/orchestrator/lib/clients/client.integration.test.ts
@@ -6,6 +6,7 @@ import { OrchestratorClient } from './client.js';
 import getPort from 'get-port';
 import { EventsHandler } from '../events.js';
 import { nanoid } from '@nangohq/utils';
+import { setTimeout } from 'timers/promises';
 
 const dbClient = getTestDbClient();
 const eventsHandler = new EventsHandler({
@@ -33,6 +34,7 @@ describe('OrchestratorClient', async () => {
 
     afterAll(async () => {
         scheduler.stop();
+        await setTimeout(1000); // waiting for the scheduler to stop
         await dbClient.clearDatabase();
     });
 
--- a/packages/scheduler/lib/workers/scheduling/scheduling.worker.ts
+++ b/packages/scheduler/lib/workers/scheduling/scheduling.worker.ts
@@ -7,6 +7,8 @@ import type knex from 'knex';
 import { logger } from '../../utils/logger.js';
 import { dueSchedules } from './scheduling.js';
 import * as tasks from '../../models/tasks.js';
+import { LeaderElection } from '../leader.election.js';
+import { uuidv7 } from 'uuidv7';
 
 interface CreatedTasksMessage {
     ids: string[];
@@ -58,12 +60,19 @@ export class SchedulingChild {
     private parent: MessagePort;
     private cancelled: boolean = false;
     private tickIntervalMs = 100;
+    private electionLeader: LeaderElection;
+    private nodeId = uuidv7();
 
     constructor(parent: MessagePort, db: knex.Knex) {
         if (isMainThread) {
             throw new Error('Scheduling should not be instantiated in the main thread');
         }
         this.db = db;
+        this.electionLeader = new LeaderElection({
+            db,
+            leaseTimeoutMs: 30_000,
+            leaderKey: 'scheduling'
+        });
         this.parent = parent;
         this.parent.on('message', async (msg: 'start' | 'stop') => {
             switch (msg) {
@@ -71,7 +80,7 @@ export class SchedulingChild {
                     await this.start();
                     break;
                 case 'stop':
-                    this.stop();
+                    await this.stop();
                     break;
             }
         });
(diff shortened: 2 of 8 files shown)


packages/scheduler/lib/workers/scheduling/scheduling.worker.ts
```diff
@@ -58,20 +60,27 @@ export class SchedulingChild {
...
+    private nodeId = uuidv7();
 
     constructor(parent: MessagePort, db: knex.Knex) {
         if (isMainThread) {
             throw new Error('Scheduling should not be instantiated in the main thread');
         }
         this.db = db;
+        this.electionLeader = new LeaderElection({
+            db,
+            leaseTimeoutMs: 30_000,
```
TBonnin, line 73: picked 30s because worse case scenario it would take less than a minute for a new leader to be elected when the previous one is terminated/crashed, which I think is acceptable for our use case.
That also means that when running locally if stopping/starting the apps quickly the previous lease might not have expired but the scheduling worker will eventually be elected leader on the second attempt

packages/orchestrator/lib/clients/client.integration.test.ts
```diff
@@ -33,6 +34,7 @@ describe('OrchestratorClient', async () => {
 
     afterAll(async () => {
         scheduler.stop();
+        await setTimeout(1000); // waiting for the scheduler to stop
```
TBonnin, line 37: I wish I didn't need this but the scheduler internals rely on messages passing to the worker_thread which is asynchronous

packages/scheduler/lib/workers/scheduling/scheduling.worker.ts
```diff
@@ -81,14 +90,20 @@ export class SchedulingChild {
...
             await this.schedule();
             await setTimeout(this.tickIntervalMs);
         }
     }
 
-    stop(): void {
+    async stop(): Promise<void> {
         logger.info('Stopping scheduling...');
         this.cancelled = true;
+        await this.electionLeader.release(this.nodeId);
```
bodinsamuel, line 106: Can you add a log when it stops?

packages/scheduler/lib/workers/scheduling/scheduling.worker.ts
```diff
@@ -81,14 +90,20 @@ export class SchedulingChild {
         logger.info('Starting scheduling...');
         // eslint-disable-next-line no-constant-condition
         while (!this.cancelled) {
+            const leader = await this.electionLeader.elect(this.nodeId);
+            if (leader.isErr()) {
+                await setTimeout(this.electionLeader.leaseTimeoutMs);
+                continue;
+            }
             await this.schedule();
             await setTimeout(this.tickIntervalMs);
```
bodinsamuel, line 99: Can you add a log when it's elected? it helps for debugging

bodinsamuel (commented):
Works locally 🤸🏼
Two comments:
- I would have gone with Redis for the sake of simplicity and not maintaining a feature that do already exists + the automatic ttl + the fact that you can keep a connection open
- 30s lock is too long. When you deploy and blue becomes green, there will be a mandatory 30s drop of activity.

TBonnin:
> Works locally 🤸🏼 Two comments:
> 
>     * I would have gone with Redis for the sake of simplicity and not maintaining a feature that do already exists + the automatic ttl + the fact that you can keep a connection open

Redis would mean adding a dependency. The benefit is calling SET is simpler than the postgres SQL query but given it is only a few lines and how often dependencies comes back to hit us I would prefer to stick with a single store in the orchestrator unless we cannot do otherwise

>     * 30s lock is too long. When you deploy and blue becomes green, there will be a mandatory 30s drop of activity.

What about 15s. It is still potentially a 15s gap but between a  deterministic gap of 15 or less when we deploy and unpredictible race conditions that would be a nightware to debug I would take the 15s or less gap

bodinsamuel:
> What about 15s. 

Sure, I would hope 1s would work, I don't see why not unless the process is unresponsive for 1s which is a stretch. 

> Redis would mean adding a dependency.

Redis is already a dep, at some point we should allow ourselves to use what's best. Alternatively, pg_advisory_lock with a infinite running transaction could maybe work. There would be no race condition and be instantaneous. 

---

Not blocking this PR, just suggesting. Feel free to merge 👍🏻

bodinsamuel (approved):
Few open comments about logging, otherwise gtm

khaliqgant (commented):
Am OK with using postgres as a lock mechanism.

TBonnin:
closing in favor of https://github.com/NangoHQ/nango/pull/2359/files?diff=unified&w=1

What we actually need is ensuring the scheduling is not happening concurrently. We don't need to ensure only one worker is running but that only one scheduling transaction at a time

PR #2356, opened 2024-06-18
fix(integration-templates): [nan-1239] fix integration-template upload

khaliqgant:
## Describe your changes
Correctly create the integration directory before uploading

NAN-1239


package-lock.json
bodinsamuel, line 1: you are probably not up to date

integration-templates/asana/nango.yaml
```diff
@@ -145,6 +145,8 @@ models:
         workspace: string | undefined
     AsanaWorkspace:
         __extends: BaseAsanaModel
+        id: string
```
bodinsamuel: same

bodinsamuel (approved):

package-lock.json
khaliqgant: Thanks, pulled the latest

PR #2357, opened 2024-06-18
fix(deps): upgrade vitest

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1240/deps-upgrade-vitest

- Upgrade vitest major before they launch v2
It was easier than expected. I wish we could reup parallelism on integration tests but we currently assume they run sequentially. It's mostly around persist and records where we drop the schema entirely or truncate the table. Ideally we won't do that, or if needed we use a unique `schema_name` to avoid dropping concurrent tests.


package.json
```diff
@@ -88,7 +88,7 @@
         "testcontainers": "^9.12.0",
         "tsx": "^4.6.2",
         "typescript": "5.3.3",
-        "vitest": "0.33.0",
+        "vitest": "1.6.0",
```
TBonnin, line 91: big jump :D

TBonnin (approved):

khaliqgant (approved):
thanks. good minus the conflicts

PR #2358, opened 2024-06-18
fix(server): deploy+confirmation split and check

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1243/api-types-and-check-deploy-endpoint

- Move endpoint `/sync/deploy` to `postDeploy.ts`
- Move endpoint `/sync/deploy/confirmation` to `postConfirmation.ts`
- Add small integration tests 
Don't want to overblown this PR especially that any good deploy will trigger temporal and such
- Type CLI
- Handle errors from API correctly

NB: it should be retro compatible but I'll let you test with `nango deploy dev` (use the actual CLI to test)


packages/nango-yaml/lib/parser.ts
```diff
@@ -168,6 +172,10 @@ export abstract class NangoYamlParser {
                 }
                 if (action.endpoint) {
                     for (const [verb, endpoint] of Object.entries(action.endpoint)) {
+                        if (!endpoint) {
+                            continue; // TS pleasing
+                        }
```
TBonnin, line 178: why do we need to allow undefined value in 
```
export type NangoSyncEndpoint = {
    [key in HTTP_VERB]?: string;
};
```
An endpoint without a path should not be valid and representable

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -834,7 +826,7 @@ async function compileDeployInfo({
         ...flow,
         name: syncName,
         version
-    });
+    } as SyncDeploymentResult); // TODO: remove this as
```
TBonnin: you are casting type and sync_type above why not here as well. 

Anyway I think the weirdness comes from this line: https://github.com/NangoHQ/nango/blob/master/packages/shared/lib/services/sync/config/deploy.service.ts#L646 
I am not sure why the assignation is there

packages/shared/lib/services/sync/manager.service.ts
```diff
@@ -131,12 +132,14 @@ export class SyncManagerService {
...
-                orchestrator.scheduleSyncHelper(
+
+                // Is this not awaited on purpose?
+                void orchestrator.scheduleSyncHelper(
                     connection,
                     createdSync as Sync,
                     syncConfig as ProviderConfig,
                     syncName,
-                    { ...sync, returns: sync.models, input: '' },
+                    { ...sync, returns: sync.models, input: '' } as NangoIntegrationData,
```
TBonnin, line 140: let's find a solution to this sync_type incompatibility instead of casting. I am happy to pair if you want to

packages/types/lib/nangoYaml/index.ts
```diff
@@ -123,5 +123,5 @@ export interface NangoModelField {
 }
 
 export type NangoSyncEndpoint = {
-    [key in HTTP_VERB]?: string;
+    [key in HTTP_VERB]?: string | undefined;
```
TBonnin, line 127: see my question above about undefined

TBonnin (commented):
I would prefer to find a solution to the sync_type type incompatibility before we merge. Let me know if this is or isn't a blocker though.

packages/nango-yaml/lib/parser.ts
bodinsamuel: because zod optional() is setting `VERB?: string | undefined`

packages/types/lib/nangoYaml/index.ts
bodinsamuel: replied above

bodinsamuel:
> I would prefer to find a solution to the sync_type type incompatibility before we merge. Let me know if this is or isn't a blocker though.

Well I didn't want to remove both Enum entirely to keep the PR shorter. Fair enough, I have removed SyncConfigType but if you don't mind I'll fix SyncType on a other PR, the change is slightly more complex

TBonnin:
> > I would prefer to find a solution to the sync_type type incompatibility before we merge. Let me know if this is or isn't a blocker though.
> 
> Well I didn't want to remove both Enum entirely to keep the PR shorter. Fair enough, I have removed SyncConfigType but if you don't mind I'll fix SyncType on a other PR, the change is slightly more complex

as long as it gets fixed ;)

packages/nango-yaml/lib/parser.ts
TBonnin: I copy pasted the wrong line. I don't mean the undefined key I mean the undefined value
```
    [key in HTTP_VERB]?: string | undefined;
                                  ^^^^^^^^^^ this
```

khaliqgant:
Will test and review this tomorrow. Sorry for the delay

packages/nango-yaml/lib/parser.ts
bodinsamuel: yep, I set that because otherwise the type is not compatible with zod type. There might be a way to improve the zod type I just didn't found how

TBonnin: what about something like 
```
            endpoints: z
                .array(
                    z.record(
                       z.union([z.literal('GET'), z.literal('PUT'), z.literal('POST'), ...]), 
                       z.string()
                    )
                )
                .optional(),

```

khaliqgant (approved):
Code looks good, but unfortunately I didn't get a chance to test this thoroughly, but don't want to block

packages/nango-yaml/lib/parser.ts
bodinsamuel: indeed, it works I find it way less clear to understand tho and doesn't enforce that any is set.
I think the only viable alternative is to have 
`z.array(z.union([z.object({ GET: z.string }), z.object({ POST: z.string() }) ...]});`

TBonnin: harder to read schema/types vs not being able to represent state that are impossible: for me the choice would be easy ;)

PR #2360, opened 2024-06-18
fix(logs): insert message without an ID

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1244/insert-messages-without-an-id

Now that we have Elasticsearch in prod I can focus a bit more time on performance quick wins.

- Insert messages (not operation) without an ID
I knew this from the beginning but I just wanted to simplify the code. They recommend not using an ID because if you pass one, all the instances need to search in their whole catalog to see if this ID already exists. Whereas if you don't send it, they will autogenerate an ordered id without checking. This should help but no clue how much.

- Clarify types between OperationRow and MessageRow which were different already but a bit mixed up
It's not perfect yet since most fields are optional 
- Fix some type errors

### DIFF
--- a/packages/logs/lib/models/helpers.ts
+++ b/packages/logs/lib/models/helpers.ts
@@ -1,5 +1,5 @@
 import { nanoid } from '@nangohq/utils';
-import type { MessageRow } from '@nangohq/types';
+import type { MessageRow, OperationRow, OperationRowInsert } from '@nangohq/types';
 import { z } from 'zod';
 import type { estypes } from '@elastic/elasticsearch';
 import { defaultOperationExpiration } from '../env.js';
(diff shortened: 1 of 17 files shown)


TBonnin (approved):
curious to see how much it helps

packages/logs/lib/models/helpers.ts
```diff
@@ -50,8 +42,45 @@ export function getFormattedMessage(
         syncConfigName: syncConfig?.name || data.syncConfigName || null,
 
         jobId: data.jobId || null,
+        meta: meta || data.meta || null,
 
         userId: user?.id || data.userId || null,
+        parentId: null
+    };
+}
+export function getFormattedMessage(data: Partial<MessageRow>, { meta }: FormatMessageData = {}): MessageRow {
```
TBonnin, line 51: should this method return a `Pick<MessageRow, ...>` to make it clear it is only setting a subset of the fields?

bodinsamuel: it's still returning everything, just more null than before :D

TBonnin: forget what I said. I thought the function was only used in getFormattedOperation but it is called from a lot of places

PR #2362, opened 2024-06-18
feat(cli): exports json schema

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1154/produce-json-schema-locally
Fixes https://linear.app/nango/issue/NAN-1155/upload-json-schema-when-deploying

- Output `schema.ts` and `schema.json` in `.nango/` folder
We decided to not output additional stuff at the root level to avoid cluttering the main folder. Customers can choose what they to commit. And we will add more output in the coming weeks/months (most likely zod, then others languages maybe)

- Rename `load` to `loadValidateParse`
For clarification because it wasn't super obvious it could fail with validation error.

- Touch a bit deploy.service to prepare new files to be sent but the logic is the same as before (not yet sending json)


packages/cli/lib/services/deploy.service.ts
```diff
@@ -308,20 +308,6 @@ class DeployService {
                     continue;
                 }
 
-                const { path: integrationFilePath, result: integrationFileResult } = localFileService.checkForIntegrationDistFile(
-                    action.name,
-                    providerConfigKey,
-                    './'
-                );
```
bodinsamuel, line 315: check now done by loadScriptFiles

packages/cli/lib/index.ts
```diff
@@ -19,7 +19,7 @@ import { DryRunService } from './services/dryrun.service.js';
 import { v1toV2Migration, directoryMigration } from './services/migration.service.js';
 import { getNangoRootPath, upgradeAction, NANGO_INTEGRATIONS_LOCATION, printDebug } from './utils.js';
 import type { ENV, DeployOptions } from './types.js';
-import { load } from './services/config.service.js';
+import { loadValidateParse } from './services/config.service.js';
```
TBonnin: I would simply call it parse. For me loading and validation is included in parsing

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -168,14 +168,11 @@ export class DryRunService {
...
         if (syncInfo?.type === 'action') {
-            type = 'action';
-        }
-
-        if (isPostConnectionScript) {
-            type = 'post-connection-script';
+            type = 'actions';
+        } else if (isPostConnectionScript) {
+            type = 'post-connection-scripts';
         }
```
TBonnin, line 176: type is being pluralized. Does that have any impact on already existing `nango-intergrations` folders?

bodinsamuel: we used to add an `s` dynamically to this variable in another function so it's actually simplifying the code

packages/cli/lib/index.ts
bodinsamuel: fair, changed in [0094971](https://github.com/NangoHQ/nango/pull/2362/commits/009497156a9e7929fcf441c74e72f392983015a9)

bodinsamuel:
@khaliqgant happy to get your review on this change

packages/cli/lib/services/deploy.service.ts
```diff
@@ -155,7 +158,6 @@ class DeployService {
                     singleDeployMode
                 };
                 const response = await http.post(confirmationUrl, bodyConfirmation, { headers: enrichHeaders() });
-                console.log(JSON.stringify(response.data, null, 2));
```
khaliqgant, line 158: Any particular reason this removed?

khaliqgant (changes requested):
Issue with removed line in the deploy process.

Looks great otherwise!

Notice that 

```
    date_created: date
    date_last_modified: date
```

produces 

```
        "date_created": {
          "type": "string",
          "format": "date-time"
        },
        "date_last_modified": {
          "type": "string",
          "format": "date-time"
        }
```

which is most accurate but I have been putting 

```
    date_created: string
    date_last_modified: string
```

because they are in fact strings and I find it confusing otherwise, but curious on your thoughts in line of this json schema output?

packages/cli/lib/services/deploy.service.ts
bodinsamuel: ah yes removed that during debug, thanks

bodinsamuel:
> Notice that [...] date

I'm fine with this. I personally think it was a not a good idea to support date, because it is not json portable and JS Date default output is not ISO, but we decided to support it so the json schema is only reflecting that. 
Note that the JSON schema is also expecting string but with a specific format so it won't break anything regarding your setup (or any setup I think)

khaliqgant (approved):

PR #2363, opened 2024-06-18
fix(logs.v1): remove auth, action, hooks

bodinsamuel:
## Describe your changes

- Remove activities from Account, Auth, Action and Hooks
Initially wanted less but turns out everything is super tied (surprising). 

- Temporarily allowed activityLogId to be string in the relevant places


packages/server/lib/hooks/hooks.ts
```diff
@@ -167,14 +158,13 @@ export const connectionRefreshFailed = async ({
...
     authError: { type: string; description: string };
-    activityLogId: number;
     logCtx: LogContext;
 }): Promise<void> => {
     await errorNotificationService.auth.create({
         type: 'auth',
         action: 'token_refresh',
         connection_id: connection.id!,
-        activity_log_id: activityLogId,
+        activity_log_id: -1,
```
bodinsamuel, line 167: This will be removed soon, just want to avoid a db migration along with code change

packages/shared/lib/services/sync/run.service.ts
```diff
@@ -169,7 +175,7 @@ export default class SyncRun {
         this.writeToDb = config.writeToDb;
         if (config.writeToDb) {
             this.slackNotificationService = config.slackService;
-            this.activityLogId = config.activityLogId;
+            this.activityLogId = config.activityLogId as unknown as number;
```
khaliqgant: This cast mixed with the one below that casts it to a string is confusing

khaliqgant (approved):
Casting confusion question inline

packages/shared/lib/services/sync/run.service.ts
bodinsamuel: true, it's only temporary during the transition but let me change that

PR #2366, opened 2024-06-19
fix(nango-yaml): fine tune parsing

bodinsamuel:
## Describe your changes

Fix more edge cases and fine tune messaging

- Handle union type in input/output correctly
- Fix: Action template without an `input` 
- Be more explicit about unsupported typescript syntax
- Do not complain about anonymous model in case of union, it's a legitimate use case of anonymous model
- Do not complain about action input model reuse, it's too noisy to be useful


packages/nango-yaml/lib/errors.ts
```diff
@@ -110,11 +110,11 @@ export class ParserErrorExtendsNotFound extends ParserError {
...
 }
 
-export class ParserErrorDataSyntax extends ParserError {
+export class ParserErrorTypeSyntax extends ParserError {
     constructor(options: { value: string; path: string[] }) {
         super({
-            code: 'data_syntax_warning',
-            message: `Data type "${options.value}" contains some unsupported typescript syntax`,
+            code: 'type_syntax_error',
+            message: `Type "${options.value}" contains some unsupported typescript syntax, please use Yaml compatible syntax.`,
```
khaliqgant: What is `Yaml compatible syntax`? I don't think this is clear enough for the user. I see a lot of support questions stemming from this type of error message.

khaliqgant (changes requested):
Adjustment for clarity in the error response requested

packages/nango-yaml/lib/errors.ts
bodinsamuel: I can't detail all the alternatives, there is Record, Pick, Omit, Partial, Required, Parameters, Readonly, etc. 
Happy to have a better message if you think of something 👌🏻

khaliqgant: I actually think this is a good time to link to the documentation with these types documented and the yaml alternative. Perhaps an entry under the advanced section? https://docs.nango.dev/customize/guides/advanced

bodinsamuel: There is already a section in the API references, I can point that https://docs.nango.dev/reference/integration-configuration#model-types

khaliqgant: Ah yes, perfect.

khaliqgant (approved):

bodinsamuel:
Thanks @khaliqgant 👌🏻

PR #2369, opened 2024-06-19
fix(db): add models_json_schema to sync_configs

bodinsamuel:
## Describe your changes

Contributes to https://linear.app/nango/issue/NAN-1155/upload-json-schema-when-deploying

- Add `models_json_schema` to sync_configs table
After I upload the json_schema when deploying, each sync_configs will hold its json schema to be able to validate input/output at runtime


TBonnin:
I imagine the exisiting `model_schema` column is not enough.
Aren't those 2 columns redundant though?

bodinsamuel:
> I imagine the exisiting model_schema column is not enough.
> Aren't those 2 columns redundant though?

They kinda are but not really, model_schema is used for the UI. I don't want to break everything (it's already kinda broken tho). After I upload json_schema I might be able to generate code snippets directly with the json_schema and drop the column but I don't want to do everything at once

bodinsamuel:
closed by mistake*

TBonnin (approved):

PR #2370, opened 2024-06-19
feat(integrations): [nan-1247] add checkr post connection script

khaliqgant:
## Describe your changes
In integration scripts it is useful to know if the checkr account has account hierarchy enabled or not. Add a post connection script to automatically inject the boolean.

NAN-1247

### DIFF
--- a/packages/server/lib/hooks/connection/checkr-post-connection.ts
+++ b/packages/server/lib/hooks/connection/checkr-post-connection.ts
@@ -1,0 +1,32 @@
+import type { InternalNango as Nango } from './post-connection.js';
+import { isAxiosError } from 'axios';
+
+export default async function execute(nango: Nango) {
+    const connection = await nango.getConnection();
+    let access_token: string;
+    if ('access_token' in connection.credentials) {
+        access_token = connection.credentials.access_token;
+    } else {
+        return;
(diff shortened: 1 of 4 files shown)


packages/server/lib/hooks/connection/checkr-post-connection.ts
```diff
@@ -0,0 +1,32 @@
...
+import { isAxiosError } from 'axios';
+
+export default async function execute(nango: Nango) {
+    const connection = await nango.getConnection();
+    let access_token: string;
+    if ('access_token' in connection.credentials) {
+        access_token = connection.credentials.access_token;
+    } else {
+        return;
+    }
```
TBonnin, line 11: you can slightly simplify like this I think: 
```
const access_token: string | undefined = connection.credentials.access_token;
if (!access_token) {
  return;
}
```

TBonnin (approved):

packages/server/lib/hooks/connection/checkr-post-connection.ts
khaliqgant: Thanks. TS complains about this because access_token isn't always guaranteed to exist on credentials.

PR #2373, opened 2024-06-19
perf(scheduler): create tasks in parallel

TBonnin:
## Describe your changes

Tasks are currently created by scheduler one by one. We can parallelize the tasks creation to speed up the scheduling transaction

### DIFF
--- a/packages/scheduler/lib/workers/scheduling/scheduling.worker.ts
+++ b/packages/scheduler/lib/workers/scheduling/scheduling.worker.ts
@@ -104,8 +104,8 @@ export class SchedulingChild {
                 if (schedules.isErr()) {
                     return Err(`Failed to get due schedules: ${stringifyError(schedules.error)}`);
                 } else {
-                    for (const schedule of schedules.value) {
-                        const task = await tasks.create(trx, {
+                    const createTasks = schedules.value.map((schedule) =>
+                        tasks.create(trx, {
                             scheduleId: schedule.id,
                             startsAfter: new Date(),
                             name: `${schedule.name}:${new Date().toISOString()}`,
@@ -116,11 +116,16 @@ export class SchedulingChild {
                             createdToStartedTimeoutSecs: schedule.createdToStartedTimeoutSecs,
                             startedToCompletedTimeoutSecs: schedule.startedToCompletedTimeoutSecs,
                             heartbeatTimeoutSecs: schedule.heartbeatTimeoutSecs
-                        });
-                        if (task.isErr()) {
-                            logger.error(`Failed to create task for schedule: ${schedule.id}`);
+                        })
+                    );
+                    const res = await Promise.allSettled(createTasks);
+                    for (const taskRes of res) {
+                        if (taskRes.status === 'rejected') {
+                            logger.error(`Failed to schedule task: ${taskRes.reason}`);
+                        } else if (taskRes.value.isErr()) {
+                            logger.error(`Failed to schedule task: ${stringifyError(taskRes.value.error)}`);
                         } else {
-                            taskIds.push(task.value.id);
+                            taskIds.push(taskRes.value.value.id);
                         }
                     }
                 }


packages/scheduler/lib/workers/scheduling/scheduling.worker.ts
```diff
@@ -116,11 +116,16 @@ export class SchedulingChild {
...
+                    );
+                    const res = await Promise.allSettled(createTasks);
+                    for (const taskRes of res) {
+                        if (taskRes.status === 'rejected') {
+                            logger.error(`Failed to schedule task: ${taskRes.reason}`);
+                        } else if (taskRes.value.isErr()) {
+                            logger.error(`Failed to schedule task: ${stringifyError(taskRes.value.error)}`);
                         } else {
-                            taskIds.push(task.value.id);
+                            taskIds.push(taskRes.value.value.id);
```
TBonnin, line 128: a bit weird but the PromiseAllSettled result type is `PromiseSettleResult<Result<Task>>[]`

packages/scheduler/lib/workers/scheduling/scheduling.worker.ts
```diff
@@ -116,11 +116,16 @@ export class SchedulingChild {
...
-                        if (task.isErr()) {
-                            logger.error(`Failed to create task for schedule: ${schedule.id}`);
+                        })
+                    );
+                    const res = await Promise.allSettled(createTasks);
+                    for (const taskRes of res) {
+                        if (taskRes.status === 'rejected') {
+                            logger.error(`Failed to schedule task: ${taskRes.reason}`);
+                        } else if (taskRes.value.isErr()) {
+                            logger.error(`Failed to schedule task: ${stringifyError(taskRes.value.error)}`);
```
bodinsamuel, line 126: you could have done all of that inside the .map (if that makes more sense)

bodinsamuel (approved):
Indeed. Looks good, just an open comment about the transaction

packages/scheduler/lib/workers/scheduling/scheduling.worker.ts
TBonnin: you mean with a `then`? The goal is not to await for the result in the map

packages/scheduler/lib/workers/scheduling/scheduling.worker.ts
```diff
@@ -116,11 +116,16 @@ export class SchedulingChild {
                             createdToStartedTimeoutSecs: schedule.createdToStartedTimeoutSecs,
                             startedToCompletedTimeoutSecs: schedule.startedToCompletedTimeoutSecs,
                             heartbeatTimeoutSecs: schedule.heartbeatTimeoutSecs
-                        });
-                        if (task.isErr()) {
-                            logger.error(`Failed to create task for schedule: ${schedule.id}`);
+                        })
+                    );
+                    const res = await Promise.allSettled(createTasks);
```
bodinsamuel, line 121: Wait my comment was not sent 🤔
I was saying (before and after this pr) you are not killing the transaction if any tasks fails, is it on purpose?

packages/scheduler/lib/workers/scheduling/scheduling.worker.ts
bodinsamuel: I meant keeping the code exactly the same but inside the map, since you don't kill the transaction
```ts
schedules.value.map(async (schedule) => {
      const task = await tasks.create(trx, {
          scheduleId: schedule.id,
          startsAfter: new Date(),
          name: `${schedule.name}:${new Date().toISOString()}`,
          createdToStartedTimeoutSecs: schedule.createdToStartedTimeoutSecs,
          startedToCompletedTimeoutSecs: schedule.startedToCompletedTimeoutSecs,
          heartbeatTimeoutSecs: schedule.heartbeatTimeoutSecs
      });
      if (task.isErr()) 
      [...]
  });

packages/scheduler/lib/workers/scheduling/scheduling.worker.ts
TBonnin: yes it is on purpose. Failing to create a task should not prevent other tasks to be created. 
The same scheduling logic will try to create the task again in the next iteration if needed

packages/scheduler/lib/workers/scheduling/scheduling.worker.ts
TBonnin: I see. But you are awaiting while iterating through the list. no?

bodinsamuel: should not no, this is a common pattern `await Promise.all(array.map(async))`

PR #2375, opened 2024-06-20
fix(run): use SyncConfig instead of NangoConfig

bodinsamuel:
## Describe your changes

Contributes to https://linear.app/nango/issue/NAN-1261/pass-json-schema-to-runner

I need to pass json_schema that will live inside `_sync_configs`, instead of adding a new param I decided to group the params that were belonging to SyncConfig object (and also remove the need for loading NangoConfig locally). nb: this SyncConfig was already fetched from the inside the method which was thing less flexible for CLI and testing.

- Add required parameter `syncConfig: SyncConfig`
The side effects are minimal but I managed to reduce/remove a lot of code inside `run()` so the main change are in run.service.ts

- Remove default export and use SyncRunService class name


packages/jobs/lib/activities.ts
```diff
@@ -466,14 +479,21 @@ export async function runPostConnectionScript(args: PostConnectionScriptArgs): P
...
         nangoConnection,
-        syncName: name,
+        syncConfig: {
+            sync_name: name,
+            file_location,
+            models: [],
+            track_deletes: false,
+            type: SyncConfigType.SYNC,
+            version: '0' // TODO: pass the correct version
+        },
```
bodinsamuel, line 495: Post Script Connection is the only weird kid regarding run service. My take with the future refacto, we should have one base class and N class per script type so they only require what's relevant for them

packages/shared/lib/models/Sync.ts
```diff
@@ -259,11 +259,12 @@ export interface SyncConfigWithProvider {
 }
 
 export interface RunScriptOptions {
+    syncConfig?: SyncConfig;
     syncName: string;
     syncId: string;
     activityLogId: number | undefined;
     nangoProps: NangoProps;
-    integrationData: NangoIntegrationData;
+    integrationData: Pick<NangoIntegrationData, 'fileLocation'>;
```
bodinsamuel, line 198: This will be removed in next PR, but since we do multiple steps deployment it's safer not to modify runner signature

packages/shared/lib/services/sync/run.service.integration.test.ts
```diff
@@ -200,45 +200,6 @@ describe('Running sync', () => {
     });
 });
 
-describe('SyncRun', () => {
-    it('should initialize correctly', () => {
```
bodinsamuel, line 204: Already tested in unit

packages/shared/lib/services/sync/run.service.ts
```diff
@@ -219,411 +212,274 @@ export default class SyncRun {
...
-                    content
-                });
-                await this.logCtx?.debug(content);
-            } else {
-                logger.info(content);
-            }
-        }
-        let nangoConfig = this.loadLocation
-            ? await loadLocalNangoConfig(this.loadLocation)
-            : await getSyncConfig(this.nangoConnection, this.syncName, this.isAction);
```
bodinsamuel, line 239: This is the main change, it is now done outside this method and simply ask syncConfig with or without loadLocation because we just need a few params not the full nango.yaml

packages/shared/lib/services/sync/run.service.ts
```diff
@@ -219,411 +212,274 @@ export default class SyncRun {
...
             }
 
-            if (!isCloud && !integrationFilesAreRemote && !isPublic) {
-                const { path: integrationFilePath, result: integrationFileResult } = localFileService.checkForIntegrationDistFile(
-                    this.syncName,
-                    providerConfigKey,
-                    this.loadLocation
-                );
-                if (!integrationFileResult) {
-                    const message = `Integration was attempted to run for ${this.syncName} but no integration file was found at ${integrationFilePath}.`;
```
bodinsamuel, line 376: Removed this check for local integration, this check is already done by the CLI and by local-integration, it's one less thing to refactor/carry in this class but let me know if you agree

packages/shared/lib/services/sync/run.service.ts
```diff
@@ -219,411 +212,274 @@ export default class SyncRun {
...
-                            }
-                        });
-
-                        return { success: false, error: new NangoError('action_script_failure', message, 500), response: false };
-                    }
-                } else {
-                    if (configModels[configInput as unknown as string]) {
-                        // TODO use joi or zod to validate the input dynamically
-                    }
-                }
```
bodinsamuel, line 427: I also removed all of this for the moment, it's just barely working. And in next PR I will introduce input/output validation but I'm not yet sure this will happen like this

khaliqgant: Yeah, good to remove this. This was really done as a proof of concept

packages/jobs/lib/processor/handler.ts
```diff
@@ -265,18 +275,28 @@ async function webhook(task: TaskWebhook): Promise<Result<JsonValue>> {
...
     }
 
+    const syncConfig = await getSyncConfigRaw({
+        environmentId: providerConfig.environment_id,
+        config_id: providerConfig.id!,
+        name: task.parentSyncName,
+        isAction: false
+    });
+    if (!syncConfig) {
+        return Err(`Action config not found: ${JSON.stringify(task)}`);
```
TBonnin: other error contains only the task id. Can we do the same for consistency sake?

packages/shared/lib/services/sync/run.service.ts
```diff
@@ -219,411 +212,274 @@ export default class SyncRun {
...
                     }
                 });
-            } else {
-                logger.error(message);
+                return { success: false, error: new NangoError(errorType, message, 404), response: false };
             }
-
-            const errorType = this.determineErrorType();
-            return { success: false, error: new NangoError(errorType, message, 404), response: false };
+            ({ environment, account } = environmentAndAccountLookup);
```
TBonnin, line 254: why the parenthesis?

TBonnin (approved):

packages/shared/lib/services/sync/run.service.ts
bodinsamuel: because it reassigns variables declared before (not my code)

packages/jobs/lib/processor/handler.ts
bodinsamuel: not exactly the case, but changed all of them https://github.com/NangoHQ/nango/pull/2375/commits/6dff6f992dc7dabbb68e7583d64436921ae81d55 
stringifying anything in message should be prohibited :p

bodinsamuel:
Can I have you review on this @khaliqgant? thanks ☺️

packages/jobs/lib/activities.ts
```diff
@@ -466,14 +478,21 @@ export async function runPostConnectionScript(args: PostConnectionScriptArgs): P
...
         writeToDb: true,
         nangoConnection,
-        syncName: name,
+        syncConfig: {
+            sync_name: name,
+            file_location,
+            models: [],
+            track_deletes: false,
+            type: 'sync',
+            version: '0' // TODO: pass the correct version
```
khaliqgant: Why not just pass the version from the `external-post-connection` script, since we already have it from the query https://github.com/NangoHQ/nango/blob/master/packages/server/lib/hooks/connection/external-post-connection.ts#L21

khaliqgant (approved):
Non blocking feedback

packages/jobs/lib/activities.ts
bodinsamuel: changed 👌🏻

packages/jobs/lib/processor/handler.ts
TBonnin: 🙈 thank you. Yes you are right I am gonna stop stringifying things in error message.

bodinsamuel: dw we'll improve that with the next iteration of Error

PR #2377, opened 2024-06-20
fix(nango-yaml): support __string advanced syntax

bodinsamuel:
## Describe your changes

- Fix advanced syntax for `__string`
The parsing was correct for union but not the output due to a leftover, added multiple e2e test.
The parsing was coming too late for object and array.

### DIFF
--- a/packages/cli/fixtures/nango-yaml/v2/advanced-syntax/nango.yaml
+++ b/packages/cli/fixtures/nango-yaml/v2/advanced-syntax/nango.yaml
@@ -39,3 +39,13 @@ models:
 
     Record:
         __string: any
+
+    RecordUnion:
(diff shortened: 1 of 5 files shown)


packages/cli/fixtures/nango-yaml/v2/advanced-syntax/nango.yaml
```diff
@@ -39,3 +39,13 @@ models:
...
     Record:
         __string: any
+
+    RecordUnion:
+        __string: any | string | boolean
+    RecordArrayUnion:
+        __string:
+            - any
+            - string
+    RecordObject:
```
khaliqgant, line 49: nice

khaliqgant (approved):

PR #2378, opened 2024-06-20
feat(integration-templates): [nan-1192] add checkr syncs and actions

khaliqgant:
## Describe your changes
Add checkr syncs and actions:

- `GET /background-check/service-list`: returns a possible list of services that Checkr can perform for a background check. Takes into account if the connection has account hierarchy enabled or not and returns the node the background check is associated with if so
- `GET /background-check/service-parameters`: obtains the parameters needed to trigger a background check for a particular background check 
- `POST /background-check/trigger`: trigger a background check for a candidate
- `POST /candidates`: create a candidate to be able to trigger background checks for

Adds a single sync:
- `GET /background-checks`: retrieves all background checks for all candidates

NAN-1192


integration-templates/checkr-partner/actions/background-check-services.ts
```diff
@@ -0,0 +1,52 @@
...
+
+                    return nodePackages;
+                });
+                services = nodePackages.flat();
+            }
+        }
+    }
+
+    return { services };
+}
```
TBonnin, line 52: I feel it would be more readable if broken up into smaller functions: 

```
export default async function runAction(nango: NangoAction): Promise<CheckrServicesResponse> {
    const { config, connection_config } = await constructRequestWithConnectionConfig(nango, '/v1/packages');
    const accountHierarchyEnabled = connection_config['accountHierarchyEnabled'] || false;

    let services: CheckrService[] = fetch({nango, config, endpoint: '/v1/packages'});

    if (accountHierarchyEnabled) {
        const nodes = await fetch({nango, config, endpoint: '/v1/nodes?include=packages'});
        services = process(nodes, services);
    }

    return { services };
}

async function fetch({nango, config, endpoint} : {nango: NangoAction, config: ProxyConfiguration, endpoint: string}) {
    const res = [];
    for await (const checkrServices of nango.paginate({ ...config, endpoint })) {
        res.push(...checkrServices);
    }
    return res;
}

function process(nodes: any[], services: CheckrService[]): CheckrService[] {
    return nodes.flatMap(node => {
        if(node.packages.length === 0) {
           return [];
        } else {
             return node.packages.map((pkg: string) => ({
                ...services.find(service => service.slug === pkg),
                node: node.custom_id
            }))
        }
    });
}
```
Haven't tested so details might be wrong

TBonnin (approved):
don't know enough about the provider to judge the business logic but code looks ok. Don't we want to systematically validate actions input with zod?

integration-templates/checkr-partner/actions/background-check-services.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @desc this handles 3 scenarios:
+ * 1. If accountHierarchyEnabled is true, then we need to fetch the services from /v1/nodes?include=packages and return the packages under their respective nodes
+ * with the associated packages
+ * 2. If accountHierarchyEnabled is false, then we need to fetch the services from /v1/packages
+ * 3. If accountHierarchyEnabled is true and there are no services under the node, then we need to fetch the services from /v1/packages
+ *
+ * This means we first fetch all the available packages. If accountHierarchyEnabled is true
+ * and there are nodes with packages we need to grab them from the packages to
+ * attach them to the node
+ */
```
bodinsamuel, line 15: imo comments are easier to understand at the exact position, I need to do a lot of back and forth to understand why it's done this way

integration-templates/checkr-partner/helpers/construct-request.ts
```diff
@@ -0,0 +1,54 @@
...
+): Promise<{ config: ProxyConfiguration; connection_config: Record<string, any> }> {
+    const connection = await nango.getConnection();
+    let access_token: string;
+    if ('access_token' in connection.credentials) {
+        access_token = connection.credentials.access_token;
+    } else {
+        throw new nango.ActionError({
+            message: `access_token is missing`
+        });
+    }
```
bodinsamuel, line 39: is there any reason a connection would not have an acces_token?

integration-templates/checkr-partner/mappers/to-background-check.ts
```diff
@@ -0,0 +1,13 @@
+import type { CheckrTriggeredBackgroundCheck, BackgroundCheck } from '../../models';
```
bodinsamuel, line 1: english: not sure about the term mappers but it's not my main language, I would call them formatters maybe

bodinsamuel (approved):
Looks good minor comments

integration-templates/checkr-partner/mappers/to-background-check.ts
khaliqgant: Thanks, IMO `mappers` is clear

integration-templates/checkr-partner/helpers/construct-request.ts
khaliqgant: More of a check to make TS happy

integration-templates/checkr-partner/actions/background-check-services.ts
khaliqgant: Thanks, kept as is h/e

PR #2380, opened 2024-06-20
fix: don't update schedule if frequency is alread set to the same value

TBonnin:
We were always updating the frequency even when we didn't need to.

This doesn't address the log saying frequency has been updated. I will do that later

### DIFF
--- a/packages/scheduler/lib/scheduler.ts
+++ b/packages/scheduler/lib/scheduler.ts
@@ -383,12 +383,16 @@ export class Scheduler {
      */
     public async setScheduleFrequency({ scheduleName, frequencyMs }: { scheduleName: string; frequencyMs: number }): Promise<Result<Schedule>> {
         return this.dbClient.db.transaction(async (trx) => {
-            const schedule = await schedules.search(trx, { names: [scheduleName], limit: 1, forUpdate: true });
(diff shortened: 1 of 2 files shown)


packages/shared/lib/services/sync/manager.service.ts
```diff
@@ -131,7 +131,7 @@ export class SyncManagerService {
                 }
 
                 const createdSync = await createSync(connection.id as number, syncName);
-                orchestrator.scheduleSyncHelper(
+                await orchestrator.scheduleSyncHelper(
```
TBonnin, line 134: unrelated but I noticed a missing await

bodinsamuel (approved):

packages/scheduler/lib/scheduler.ts
```diff
@@ -383,13 +383,17 @@ export class Scheduler {
      */
     public async setScheduleFrequency({ scheduleName, frequencyMs }: { scheduleName: string; frequencyMs: number }): Promise<Result<Schedule>> {
         return this.dbClient.db.transaction(async (trx) => {
-            const schedule = await schedules.search(trx, { names: [scheduleName], limit: 1, forUpdate: true });
```
TBonnin, line 386: we are doing a single update below so no need to lock the schedule. One could argue that we don't need the transaction anymore

PR #2381, opened 2024-06-20
fix(eslint): enable @typescript-eslint/require-await

bodinsamuel:
## Describe your changes

- enable `@typescript-eslint/require-await` and fix all related errors


packages/jobs/lib/runner/local.runner.ts
```diff
@@ -71,7 +71,7 @@ export class LocalRunner implements Runner {
                 });
             }
 
-            return new LocalRunner(runnerId, `http://localhost:${port}`, childProcess);
+            return Promise.resolve(new LocalRunner(runnerId, `http://localhost:${port}`, childProcess));
```
TBonnin, line 74: not sure I understand those `Promise.resolve` changes. Isn't that the point of the async keyword, to wrap the return value into a promise?

bodinsamuel: yes, usually the fix is to remove the async keyword but in those case where I added Promise.resolve it's because we are inhering or expecting a Promise sometimes. 
Linter doesn't say it won't work, it just says it's incorrect

TBonnin: ok I see the problem is not the the async keyword but the linter who really wants to see an await/Promise inside the function

TBonnin (approved):

PR #2383, opened 2024-06-20
perf: use last_scheduled_task_id for dueSchedules query

TBonnin:
scanning tasks table to determine if a schedule has a running task or had a recently run task doesn't scale
we are now using the last_scheduled_task_id column in schedules to avoid heavy subqueries

### DIFF
--- a/packages/scheduler/lib/workers/scheduling/scheduling.ts
+++ b/packages/scheduler/lib/workers/scheduling/scheduling.ts
@@ -7,40 +7,26 @@ import { TASKS_TABLE } from '../../models/tasks.js';
 
 export async function dueSchedules(db: knex.Knex): Promise<Result<Schedule[]>> {
     try {
-        const query = db
-            .with(
-                'due_dates',
-                // calculate the most recent due date for each schedule that is started/not deleted
-                db
-                    .select(
-                        's.id',
-                        db.raw(`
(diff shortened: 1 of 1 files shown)


packages/scheduler/lib/workers/scheduling/scheduling.ts
```diff
@@ -7,40 +7,27 @@ import { TASKS_TABLE } from '../../models/tasks.js';
...
+                    // schedule with last task not running and was started before the last due time
+                    .orWhere(function () {
+                        this.whereNotIn('t.state', ['CREATED', 'STARTED']).andWhere(
+                            't.starts_after',
+                            '<',
+                            db.raw(`s.starts_at + (floor(extract(EPOCH FROM (now() - s.starts_at)) / extract(EPOCH FROM s.frequency)) * s.frequency)`)
+                        );
+                    });
+            })
+            .debug(true)
```
khaliqgant: guessing this debug is temporary?

khaliqgant (approved):
Non blocking question

packages/scheduler/lib/workers/scheduling/scheduling.ts
TBonnin: no i forgot to remove it

PR #2387, opened 2024-06-21
fix(connection): refresh credentials should re-fetch connection before proceeding

bodinsamuel:
## Describe your changes

Should fix https://linear.app/nango/issue/NAN-1164/credal-box-connection-failing-to-refresh-intermittently

- When refreshing credentials if another one was running the `connection` reference was obsolete but we would reuse it to try to refresh again. The short term fix is to re-refetch and continue, but we could return the new credentials directly

### DIFF
--- a/packages/shared/lib/services/connection.service.ts
+++ b/packages/shared/lib/services/connection.service.ts
@@ -647,7 +647,7 @@ class ConnectionService {
 
         if (connection?.credentials?.type === 'OAUTH2' || connection?.credentials?.type === 'APP' || connection?.credentials?.type === 'OAUTH2_CC') {
             const { success, error, response } = await this.refreshCredentialsIfNeeded({
-                connection,
+                oldConnection: connection,
                 providerConfig: config,
                 template: template as ProviderTemplateOAuth2,
                 environment_id: environment.id,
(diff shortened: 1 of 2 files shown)


packages/shared/lib/services/connection.service.ts
```diff
@@ -811,23 +811,23 @@ class ConnectionService {
     }
 
     private async refreshCredentialsIfNeeded({
-        connection,
+        oldConnection,
```
bodinsamuel, line 814: changed this to avoid confusion inside the function and clarify that it should be treated as obsolete

khaliqgant (approved):
Thanks. This fix looks a bit messy and recursive but if you verified it locally then I'm ok with it

bodinsamuel:
> This fix looks a bit messy and recursive

what else would you envision? the refresh_token is most likely outdated so I don't see any other way (except returning directly the new credentials)

khaliqgant:
> > This fix looks a bit messy and recursive
> 
> what else would you envision? the refresh_token is most likely outdated so I don't see any other way (except returning directly the new credentials)

Didn't have anything in mind w/o looking deeper at it, so am OK with this

PR #2388, opened 2024-06-21
fix(logs): handle maximum filters gracefully

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1267/logs-displays-a-message-when-maximum-filter-is-reached-instead-of

- Handle maximum filters gracefully 
- Increase maximum filters that can be selected

Screenshot 2024-06-21 at 16 06 37


khaliqgant (approved):
20 is an ES limitation?

bodinsamuel:
> 20 is an ES limitation?

nope, just an arbitrary one to not crush the performances.  if I enable all filters I have close to 100 already which could be use against us

PR #2396, opened 2024-06-24
fix(axios): disable axios proxy when passing agent

mithlesh135:
…h kicks in with https_proxy env set

## Describe your changes
As we are using httpProxyAgent we need to disable the axios default proxy so the httpProxyAgent is considered not the proxy with kicks in by default when we have https_proxy env set

### DIFF
--- a/packages/utils/lib/axios.ts
+++ b/packages/utils/lib/axios.ts
@@ -1,5 +1,6 @@
 import { HttpProxyAgent } from 'http-proxy-agent';
 import { HttpsProxyAgent } from 'https-proxy-agent';
+import type { AxiosRequestConfig } from 'axios';
 import axios from 'axios';
 import http from 'node:http';
 import type { Agent as HttpAgent } from 'node:http';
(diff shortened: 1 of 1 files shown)


packages/utils/lib/axios.ts
```diff
@@ -20,6 +20,7 @@ if (hasHttpsProxy) {
 }
 
 export const axiosInstance = axios.create({
+    proxy: false,
```
mithlesh135: this will disable the default axios proxy, and work with set httpProxyAgent, currently default axios proxy takes precedence and causes issue as it finds a set https_proxy env

TBonnin: shouldn't it be `proxy: !(hasHttpProxy || hasHttpProxy)`

mithlesh135: done

bodinsamuel (approved):

bodinsamuel:
@khaliqgant @TBonnin can you take a look too please? 👍🏻

TBonnin (approved):

PR #2398, opened 2024-06-24
fix(metrics): add sync, action, webhook, post script, proxy

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1279/track-new-metric-and-log-details-in-datadog

We only have logs to compute those metrics, and we want to have more than 2 weeks history. It's a bit messy because the obvious place is to put that in `processor.ts` or `hander.ts` but then it would not failure that happen before exec (i.e: at orchestration). But then sync does not get orchestrated the same way, so a bit annoying. And then early return makes stuff a bit more repetitive.

- Add metric for sync execution
- Add metric for action execution
- Add metric for post script connection execution
- Add metric for incoming webhook
- Add metric for proxy
It's the more messy since we have dozen of early return conditions
- Fix typo in Reconcile


packages/jobs/lib/processor/handler.ts
```diff
@@ -35,7 +34,14 @@ export async function handler(task: OrchestratorTask): Promise<Result<JsonValue>
...
     if (task.isSync()) {
-        return sync(task);
+        const start = Date.now();
+        const res = await sync(task);
+        if (res.isErr()) {
+            metrics.increment(metrics.Types.SYNC_FAILURE);
+        } else {
+            metrics.increment(metrics.Types.SYNC_SUCCESS);
+            metrics.duration(metrics.Types.SYNC_TRACK_RUNTIME, Date.now() - start);
+        }
```
bodinsamuel: It was the best place I found because they can be triggered or run on schedule, please advice if there is a better place. (`processTask` would be better but it's generic and I would have to add some ifs)

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -171,6 +172,8 @@ export class Orchestrator {
                     const errorMsg = `Execute: Failed to parse input '${JSON.stringify(input)}': ${stringifyError(e)}`;
                     const error = new NangoError('action_failure', { error: errorMsg });
                     span.setTag('error', e);
+
+                    metrics.increment(metrics.Types.ACTION_FAILURE);
```
TBonnin: we already have the `execute.action` trace above (and the associated metrics) not sure we need to add another metric. Same for webhook.

packages/jobs/lib/processor/handler.ts
TBonnin: that's the main entry point for syncs so I think it is ok.

I would prefer to wrap `const res = await sync(task)` in a `execute.sync` span, and automatically get the hits/duration/errors metrics instead of managing metrics individually. 
It also has the main benefit to allow us to add tags that can be helpful for debugging

packages/shared/lib/clients/orchestrator.ts
bodinsamuel: Traces' metric does not give the break down per dimensions, it would only give us a volume of actions which is great but not enough in that case
see: https://us3.datadoghq.com/metric/summary?filter=trace.e&metric=trace.execute.action

packages/jobs/lib/processor/handler.ts
bodinsamuel: same answer than below, it would give us the volume/duration but not the breakdown per status

packages/shared/lib/clients/orchestrator.ts
TBonnin: I am not sure I understand what you mean by dimensions. We seems to only be interested in success vs failure which is what `execute.action.hits` and `execute.action.errors` will give you. Am I missing something?

bodinsamuel: hits gives the full volume and errors only gives the error if we tagged the span appropriately (if relevant)
there is no way to get only success (hits - errors)
cf https://us3.datadoghq.com/metric/explorer?fromUser=false&start=1719174058630&end=1719260458630&paused=false#N4Ig7glgJg5gpgFxALlAGwIYE8D2BXJVEADxQEYAaELcqyKBAC1pEbghkcLIF8qo4AMwgA7CAgg4RKUAiwAHOChASAtnADOcAE4RNIKtrgBHPJoQaUAbVBGN8qVoD6gnNtUZCKiOq279VKY6epbINiAiGOrKQdpYZAYgUJ4YThr42gDGSsgg6gi6mZaBZnHKGniqyAUY2QB0cMRwmQRwdbUSUnWM4hrAAFQ87RpOmfgiCAAUAJQgfKCR0bmxWABMickIqel4WTl5iIXFICvlldXatW2Nza3tmZ0iDdrabn2Dw6PjU7M8ALpUVzuPCYULhIGqEEYGKleJzCigCFQmE6NZzAEgDRyNA5UDyDA4hAIfZQHAwUagjQQTKJNCiOBOOSKco4OlQWn0pz0JjKERuDxoeFJCD2TBYRkKfYAIww2ks-x4fEx8jpCAAwlJhDAUCIQWgeEA

TBonnin: a simple formula in datadog (hits - errors) will give your the success https://us3.datadoghq.com/metric/explorer?fromUser=false&graph_layout=multi&start=1719174606022&end=1719261006022&paused=false#N4Ig7glgJg5gpgFxALlAGwIYE8D2BXJVEADxQEYAaELcqyKBAC1pEbghkcLIF8qo4AMwgA7CAgg4RKUAiwAHOChASAtnADOcAE4RNIKtrgBHPJoQaUAbVBGN8qVoD6gnNtUZCKiOq279VKY6epbINiAiGOrKQdpYZAYgUJ4YThr42gDGSsgg6gi6mZaBZnHKGniqyAUY2QB0cMRwmQRwdbUSUnWM4hrAAFQ87RpOmfgiCAAUAJQgfKCR0bmxWABMickIqel4WTl5iIXFICvlldXatW2Nza3tmZ0iDdrabn2Dw6PjU7M8ALpUVzuPCYULhDRyNA5UDyDBQhAIfYwIxwaRUWHwxFOUQCUjIAAsfBAQNUIIwMVK8TmFFAEKwUJkIAxiERyjAGHciWZCLg2JEuJQhMBblJmApOjW1NAJLJ4riZAABABaBUrdb-Kh0hlEbms3JQHAwUagjQQTKJNCiXlyRTlHCWqAWq1OehMZQiEVw6lJCD2TBYJw2-YAIw5ln+PCJfvEAGEpMIYCgRCC0DwgA

bodinsamuel: Not very convenient tbh
If you don't mind and to not block this PR I added traces for each entry points in handler + the original metrics, let's compare and see if there is no drawbacks using one or the other 73ad6b95587e5cc25c8c774f7361f07c13589daf

bodinsamuel:
@TBonnin let me know if there was a reason to return instead of throw https://github.com/NangoHQ/nango/pull/2398/commits/85a175e0f36ecd07a11aef2d310b9d5744e80897

TBonnin:
Throwing is fine. Note that this code would be greatly simplified when removing temporal

TBonnin (approved):

PR #2399, opened 2024-06-24
feat(integration-templates): [nan 1235] unanet integration

khaliqgant:
## Describe your changes
Adds several actions to unanet

NAN-1235


packages/shared/flows.yaml
```diff
@@ -2877,6 +2877,101 @@ integrations:
                         links:
```
bodinsamuel, line 2877: This file will soon become unmanageable :D

integration-templates/unanet/nango.yaml
```diff
@@ -0,0 +1,96 @@
+integrations:
+    unanet:
+        actions:
+            create-lead:
+                endpoint: POST /lead
```
bodinsamuel, line 5: out of curiosity, do we recommend having action endpoints not prefixed with the integration name?

bodinsamuel (approved):
Looks good

integration-templates/unanet/nango.yaml
khaliqgant: Don't think we have anything official, but whenever you call an action you must pass in the provider config key so prefixing with the integration seems redundant

packages/shared/flows.yaml
khaliqgant: Yeah, I've never touched this by hand, `flows.js` just updates this programmatically.

PR #2401, opened 2024-06-24
fix(orchestrator): update schedule last task when running sync manually and add runSyncCommand error loggging

TBonnin:
2 commits, 2 changes:
- update the schedule last task when sync is run manually, otherwise we can run into a scenario where a sync is run manually right before the schedule due date. Then the scheduling worker would create a new task, not knowing about the manually run one and 2 syncs would be running at the same time.
- we were not logging errors in `runSyncCommand`

### DIFF
--- a/packages/scheduler/lib/scheduler.ts
+++ b/packages/scheduler/lib/scheduler.ts
@@ -178,6 +178,9 @@ export class Scheduler {
             const created = await tasks.create(trx, taskProps);
             if (created.isOk()) {
                 const task = created.value;
+                if (task.scheduleId) {
+                    await schedules.update(trx, { id: task.scheduleId, lastScheduledTaskId: task.id });
+                }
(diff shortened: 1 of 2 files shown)


packages/scheduler/lib/scheduler.ts
```diff
@@ -178,6 +178,9 @@ export class Scheduler {
             const created = await tasks.create(trx, taskProps);
             if (created.isOk()) {
                 const task = created.value;
+                if (task.scheduleId) {
+                    await schedules.update(trx, { id: task.scheduleId, lastScheduledTaskId: task.id });
+                }
```
TBonnin, line 183: doing it in `scheduler.immediate` so it can be done in the transaction.

khaliqgant (approved):

bodinsamuel (approved):

PR #2402, opened 2024-06-24
fix: /sync/status should return sync frequency

TBonnin:
sync.frequency is null unless the value is overwritten for a given connection sync 
This fix ensures that we fetch the sync config value if no overwrite exists

https://linear.app/nango/issue/NAN-1269/frequency-not-returned-from-sync-status-endpoint

### DIFF
--- a/packages/shared/lib/services/sync/manager.service.ts
+++ b/packages/shared/lib/services/sync/manager.service.ts
@@ -1,4 +1,4 @@
-import { deleteSyncConfig, deleteSyncFilesForConfig, getSyncConfig } from './config/config.service.js';
+import { deleteSyncConfig, deleteSyncFilesForConfig, getSyncConfig, getSyncConfigByParams } from './config/config.service.js';
 import connectionService from '../connection.service.js';
 import { deleteScheduleForSync, getSchedule, updateScheduleStatus } from './schedule.service.js';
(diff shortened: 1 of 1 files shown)


packages/shared/lib/services/sync/manager.service.ts
```diff
@@ -493,13 +493,30 @@ export class SyncManagerService {
...
         const isGloballyEnabled = await featureFlags.isEnabled('orchestrator:schedule', 'global', false);
         const isEnvEnabled = await featureFlags.isEnabled('orchestrator:schedule', `${environmentId}`, false);
         const isOrchestrator = isGloballyEnabled || isEnvEnabled;
         if (isOrchestrator) {
             const latestJob = await getLatestSyncJob(sync.id);
             const schedules = await orchestrator.searchSchedules([{ syncId: sync.id, environmentId }]);
+            let frequency = sync.frequency;
+            if (!frequency) {
+                const syncConfig = await getSyncConfigByParams(environmentId, sync.name, providerConfigKey);
+                frequency = syncConfig?.runs || '';
```
TBonnin: what's best to return here: an empty string or null?

khaliqgant: null

khaliqgant (approved):

bodinsamuel (approved):

PR #2403, opened 2024-06-24
chore: remove temporal

TBonnin:
Also remove any usage of the schedules table (table and data will be remove later)

https://linear.app/nango/issue/NAN-1222/remove-temporal-from-codebase
https://linear.app/nango/issue/NAN-1091/update-documentation-about-temporal


packages/jobs/lib/crons/autoIdleDemo.ts
```diff
@@ -37,15 +36,26 @@ export function cronAutoIdleDemo(): void {
     });
 }
 
-export async function exec(): Promise<void> {
+export async function exec({ orchestrator }: { orchestrator: Orchestrator }): Promise<void> {
     logger.info('[autoidle] starting');
 
-    const syncs = await findPausableDemoSyncs();
+    const syncs = await findDemoSyncs();
```
TBonnin, line 30: It is bit more complicated since schedules are not in the same database anymore and we cannot simply join syncs with schedules tables to find out which ones are running. We could limit to only search the syncs updated in the last X days so the size doesn't grow indefinitely over time, assuming this cron runs successfully at least once during this period. Can you think of a better idea? @khaliqgant @bodinsamuel

packages/jobs/lib/crons/deleteSyncsData.ts
```diff
@@ -55,12 +58,10 @@ export async function exec(): Promise<void> {
...
-            let countSchedules = 0;
-            do {
-                countSchedules = await softDeleteSchedules({ syncId: sync.id, limit: limitSchedules });
-                logger.info(`[deleteSyncs] soft deleted ${countSchedules} schedules`);
-                metrics.increment(metrics.Types.JOBS_DELETE_SYNCS_DATA_SCHEDULES, countSchedules);
-            } while (countSchedules >= limitSchedules);
+            const resSchedule = await orchestrator.deleteSync({ syncId: sync.id, environmentId: sync.environmentId });
+            const deletedScheduleCount = resSchedule.isErr() ? 1 : 0;
+            logger.info(`[deleteSyncs] soft deleted ${deletedScheduleCount} schedules`);
+            metrics.increment(metrics.Types.JOBS_DELETE_SYNCS_DATA_SCHEDULES, deletedScheduleCount);
```
TBonnin, line 65: forgot this cron job. I will cleanup by hand the schedules that we are currently not deleting

dev/docker-compose.dev.yaml
```diff
@@ -14,34 +14,6 @@ services:
...
-        depends_on:
-            - temporal
-        environment:
-            - TEMPORAL_ADDRESS=temporal:7233
-            - TEMPORAL_CORS_ORIGINS=http://localhost:3000
-        ports:
-            - 8081:8080
-        networks:
-            - nango
-
```
bodinsamuel, line 44: Feels good

packages/jobs/lib/crons/autoIdleDemo.ts
bodinsamuel: yes make sense, not ideal but will scale for a long time

bodinsamuel (commented):
I'll let you finish but looks good, and sorry about all the conflicts :|

packages/jobs/lib/crons/autoIdleDemo.ts
TBonnin: A long term solution would be to add a `pause_after` feature into the orchestrator. For now I will keep the cron job

packages/jobs/lib/integration.service.ts
```diff
@@ -32,14 +27,13 @@ class IntegrationService implements IntegrationServiceInterface {
...
+        this.runningScripts.set(syncId, { ...scriptObject, cancelled: true });
 
         const res = await runner.client.cancel.mutate({
             syncId
         });
 
-        if (res.isOk()) {
-            this.runningScripts.set(syncId, { ...scriptObject, cancelled: true });
-        } else {
+        if (res.isErr()) {
```
TBonnin, line 36: fixing a race condition where the runner script execution could return before the `runningScripts` was updated, causing the `runScript` to return a non cancellation error

packages/orchestrator/lib/clients/processor.ts
```diff
@@ -117,6 +117,10 @@ export class OrchestratorProcessor {
                 heartbeat = this.heartbeat(task);
                 const res = await this.handler(task);
                 if (res.isErr()) {
+                    if (task.abortController.signal.aborted) {
+                        // task was aborted. No need to set it as failed
+                        return;
+                    }
```
TBonnin, line 123: We would attempt to set the already cancelled task to failed, which would fail. No need to do that

bodinsamuel (approved):
🗿Congrats!
Just a little details in the UI, the frequency column is less readable imo
Screenshot 2024-06-28 at 10 19 21

TBonnin:
> 🗿Congrats! Just a little details in the UI, the frequency column is less readable imo Screenshot 2024-06-28 at 10 19 21

Is it something you see in prod? This PR isn't intended to make any functional changes

bodinsamuel:
I'm seeing that locally, it's just a display thing not really critical

TBonnin:
> I'm seeing that locally, it's just a display thing not really critical

what's in your nango.yaml? The UI should use the same value

bodinsamuel:
> what's in your nango.yaml? The UI should use the same value

for example the one displayed as "every1h" is `runs: every 1 hour` in nango.yaml

PR #2404, opened 2024-06-25
fix(syncConfigs): handle new model_schema + display in the UI

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1270/missing-data-in-api-ref-generation
Fixes https://linear.app/nango/issue/NAN-1292/show-types-in-ui

- Correctly show new model in the UI 
You can test using your own scripts and try with a new provider checking the available templates

- Update `config.service.ts` to return a StandardNangoConfig without going through unnecessary parsing
We were fetching from the DB and then passing through the parsing again for no reason except outputting the same type. It's confusing and prevent me to get rid of the nango-config.service.ts, so I slightly rewrote the formatter to make it output the right format directly. I also grouped the 3 functions that were querying the same data with different params.


bodinsamuel:
~Missed action input, fixing that~
Fixed, good for review

khaliqgant (approved):
unfortunately don't have too much bandwidth to do a thorough review but don't want to block. Looks good high level

TBonnin (approved):

PR #2406, opened 2024-06-25
fix(deploy): pass jsonSchema

bodinsamuel:
## Describe your changes

Contributes to https://linear.app/nango/issue/NAN-1261/pass-json-schema-to-runner

- Pass new property JsonSchema when deploying
It's not used or stored yet.
Added one test that assert the `package()` is correct and one integration test that checks deploy works.

- Small rework of `deploy` function to remove params passed as reference

- Revert the storage of raw `undefined` in model schema
It's not compatible with JSON so it's now stored as a string instead (but still displayed correctly)

## Tests

- Try the UI and sync without deploying anything else
- Try uploading with previous version `nango deploy dev` it shouldn't change anything
- Try uploading with new version `node packages/cli/dist/index.js deploy dev` it shouldn't break anything :D


khaliqgant (approved):

packages/cli/lib/services/deploy.service.ts
```diff
@@ -296,8 +290,8 @@ class DeployService {
                     attributes: {},
                     metadata: metadata,
                     input: sync.input || undefined,
-                    // sync_type: sync.sync_type as SyncType,
-                    type: sync.type as any,
+                    sync_type: sync.sync_type,
+                    type: sync.type,
```
TBonnin, line 294: what's the difference between sync_type and type and why it is uncommented here and not below?

packages/cli/lib/services/deploy.service.ts
```diff
@@ -365,7 +358,7 @@ class DeployService {
             return null;
         }
 
-        return { flowConfigs: postData, postConnectionScriptsByProvider, jsonSchema: jsonSchema };
+        return { flowConfigs: postData, postConnectionScriptsByProvider, jsonSchema: JSON.parse(jsonSchema) as JSONSchema7 };
```
TBonnin: parsing can fail. are we ok throwing here?

TBonnin (approved):

packages/cli/lib/services/deploy.service.ts
bodinsamuel: good point

packages/cli/lib/services/deploy.service.ts
bodinsamuel: - type is action or sync
- sync_type is full or incremental 

it was commented by mistake, but it's not really used in the backend. Those two names are confusing but yeah legacy stuff

PR #2407, opened 2024-06-25
fix(logs.v1): remove remaining

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-585/stop-writing-from-v1

- Remove remaining v1 calls
Data will be dropped after this


TBonnin (approved):
🚢

PR #2413, opened 2024-06-26
fix(tracing): various

bodinsamuel:
## Describe your changes

- Fix wrong c/c in span name
- Fix throwing instead of returning when triggering webhook
- Do not log `/health` in datadog
- Add same request logging to orchestrator and persist
- Add route to tracing when using `createRoute`


khaliqgant (approved):

packages/orchestrator/lib/server.ts
```diff
@@ -23,21 +23,10 @@ export const getServer = (scheduler: Scheduler, eventEmmiter: EventEmitter): Exp
...
-        };
-        next();
-        if (res.statusCode < 400) {
-            logger.info(`${req.method} ${req.path} -> ${res.statusCode}`);
-        }
-    });
+    // Log all requests
+    if (process.env['ENABLE_REQUEST_LOG'] !== 'false') {
+        server.use(requestLoggerMiddleware({ logger }));
+    }
```
TBonnin, line 29: `requestLoggerMiddleware` doesn't log the body in case of error like the previous code was doing. Are we ok losing this functionality?

packages/orchestrator/lib/server.ts
```diff
@@ -23,21 +23,10 @@ export const getServer = (scheduler: Scheduler, eventEmmiter: EventEmitter): Exp
...
-            originalSend.call(this, body) as any;
-            return this;
-        };
-        next();
-        if (res.statusCode < 400) {
-            logger.info(`${req.method} ${req.path} -> ${res.statusCode}`);
-        }
-    });
+    // Log all requests
+    if (process.env['ENABLE_REQUEST_LOG'] !== 'false') {
```
TBonnin, line 27: is this env already enabled in prod/staging?

TBonnin (approved):
couple of questions but looks good otherwise

packages/orchestrator/lib/server.ts
bodinsamuel: I haven't used this feature, happy to put it back if you feel it's important.

packages/orchestrator/lib/server.ts
bodinsamuel: it's on by default, I'm just giving the possibility to disable it

packages/orchestrator/lib/server.ts
TBonnin: not super important. we can re-add it if needed

PR #2415, opened 2024-06-26
feat(tba-integration): [nan-1259] support netsuite tba authorization

khaliqgant:
## Describe your changes
Adds support for the TBA flow employed by netsuite

NAN-1259


packages/frontend/lib/index.ts
```diff
@@ -128,7 +130,11 @@ export default class Nango {
             return this.customAuth(providerConfigKey, connectionId, this.convertCredentialsToConfig(credentials), connectionConfig);
         }
 
-        const url = this.hostBaseUrl + `/oauth/connect/${providerConfigKey}${this.toQueryString(connectionId, options as ConnectionConfig)}`;
+        let url = this.hostBaseUrl + `/oauth/connect/${providerConfigKey}${this.toQueryString(connectionId, options as ConnectionConfig)}`;
+
+        if (options && 'credentials' in options && 'token_id' in options.credentials && 'token_secret' in options.credentials) {
+            url = this.hostBaseUrl + `/auth/tba/${providerConfigKey}${this.toQueryString(connectionId, options as ConnectionConfig)}`;
+        }
```
TBonnin, line 137: I feel the logic and condition above should leave in credentials, for example a`isCustomAuth()` and `urlPath()` functions would significantly simplify the code in `nango.auth()`

packages/server/lib/controllers/auth/tba.ts
```diff
@@ -0,0 +1,234 @@
...
+        connectionId,
+        params: {
+            oauth_token: oauth_token as string,
+            state: session.id.replace(/-/g, '')
+        },
+        connectionConfig
+    });
+
+    res.redirect(redirectUrl.toString());
+});
```
TBonnin, line 235: a lot of the code above can fail and throw. what kind of response message is sent back to the user in this case?

packages/server/lib/helpers/tba.ts
```diff
@@ -0,0 +1,100 @@
...
+import type { OAuthSession } from '@nangohq/shared';
+import querystring from 'querystring';
+import * as crypto from 'node:crypto';
+import type { Template as ProviderTemplate, IntegrationConfig as ProviderConfig } from '@nangohq/types';
+import { interpolateStringFromObject } from '@nangohq/shared';
+
+export function percentEncode(str: string) {
+    return encodeURIComponent(str)
+        .replace(/[!'()*]/g, (c) => '%' + c.charCodeAt(0).toString(16).toUpperCase())
+        .replace(/%20/g, '%20');
```
TBonnin: are we replacing anything here?

packages/server/lib/helpers/tba.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    encodedParams.sort((a, b) => {
+        if (a[0] === b[0]) {
+            return a[1] < b[1] ? -1 : 1;
+        }
+        return a[0] < b[0] ? -1 : 1;
+    });
+
+    return encodedParams.map((pair) => pair.join('=')).join('&');
```
TBonnin: you can chain everything I think 
```
export function encodeParameters(allParams: Record<string, string>): string {
    return Object.entries(allParams)
        .map(([key, value]) => [percentEncode(key), percentEncode(value)] as [string, string])
        .sort(([a, aVal], [b, bVal]) => {
            if (a === b) {
                return aVal < bVal ? -1 : 1;
            }
            return a < b ? -1 : 1;
        })
        .map((pair) => pair.join('='))
        .join('&');
}
```

packages/frontend/lib/index.ts
khaliqgant: Yeah, agreed. Let me get this out and refactor this some tomorrow.

packages/server/lib/controllers/auth/tba.ts
khaliqgant: The generic error which wouldn't be helpful. I'll add in more error reporting and catching but for now this is going out for one customer with one use case so will make it more robust tomorrow.

khaliqgant:
As mentioned in the comments this is an initial implementation with a need to refactor, but it is a bit difficult to test locally w/o verifying with the client first. So will ship this now and revisit tomorrow.

PR #2416, opened 2024-06-26
fix(sdk): handle internal issue when reaching persist

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1304/persist-calls-sometimes-fails

- Handle ECONNREFUSED and ECONNRESET when reaching Persist
Right now a call that failed like this would break the script with an ugly error, it just makes it more readable and easily identifiable by us. I have noted that a better solution would be to retry on top of that. This can happen when we deploy or when the service scale down.
Also I have decided to make logs less critical by failing "silently" unfortunately there is no way right now to warn without stopping the script and we can't log because it just failed :D Let me know if you think it's okay


khaliqgant (commented):
Feel like we should retry and that should be part of this initial fix. We already have retry logic in a few different places - don’t think it would be a big lift to add it here.

packages/shared/lib/sdk/sync.ts
```diff
@@ -691,19 +691,26 @@ export class NangoAction {
...
+                    activityLogId: this.activityLogId,
+                    level: options.level ?? 'info',
+                    timestamp: options.timestamp,
+                    msg: content
+                }
+            });
+        } catch (err) {
+            logger.error('Failed to log to persist, due to an internal error', err);
+            // TODO: find a way to report that
+            return;
```
TBonnin, line 728: Though one. Failing silently allows the script to continue when logging fails but there is no way to report to the user since the logging failed. I think it is fine to fail silently in order not to stop the script, which is the most important thing. Maybe, in addition to the TODO add a comment to explain our reasoning for now

TBonnin (approved):
good idea mapping the error from `Persist`. agreed that a quick retry mechanism would be nice.

bodinsamuel:
Added the exponential backoff. Put it in utils so we can reuse it (but didn't touch proxy)

packages/shared/lib/sdk/sync.ts
bodinsamuel: added a comment

PR #2419, opened 2024-06-27
fix(deploy): reconcile flag was turned off

bodinsamuel:
## Describe your changes

- Re up reconcile flag in CLI 
I put it to false by mistake, not entirely sure why this flag exists tbh (especially if always true)

- Added an integration to check deletion is working

### DIFF
--- a/packages/cli/lib/services/deploy.service.ts
+++ b/packages/cli/lib/services/deploy.service.ts
@@ -145,7 +145,7 @@ class DeployService {
         const nangoYamlBody = response.yaml;
 
         const url = process.env['NANGO_HOSTPORT'] + `/sync/deploy`;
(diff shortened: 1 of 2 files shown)


khaliqgant (approved):
> I put it to false by mistake, not entirely sure why this flag exists tbh (especially if always true)

Thanks, reconcile isn't always set to true. It is set to false if not in auto confirm mode

bodinsamuel:
> Thanks, reconcile isn't always set to true. It is set to false if not in auto confirm mode

Mmh then I messed up something, let me double check

bodinsamuel:
Okay so we agreed the reconcile flag is not useful because there is a confirmation endpoint, most probably legacy stuff to remove

PR #2420, opened 2024-06-27
fix: schedule nextDueDate should be null when schedule is not running

TBonnin:
this would cause the github demo to fail at the fetching data step because it relies on checking for nextDueDate to trigger a sync or not.

https://linear.app/nango/issue/NAN-1293/interactive-demo-fetch-new-data-step-doesnt-work-anymore


bodinsamuel (approved):
Thanks for looking into it

PR #2425, opened 2024-06-27
fix(tba-integration): [nan 1259] support netsuite tba authorization

khaliqgant:
## Describe your changes
Machine to machine flow for tba which simplifies the flow.

NAN-1259


packages/frontend/lib/index.ts
```diff
@@ -119,9 +119,9 @@ export default class Nango {
         if (
             options &&
             'credentials' in options &&
-            (!('oauth_client_id_override' in options.credentials) || !('oauth_client_secret_override' in options.credentials)) &&
-            !('token_id' in options.credentials) &&
-            !('token_secret' in options.credentials) &&
+            (('token_id' in options.credentials && 'token_secret' in options.credentials) ||
```
khaliqgant, line 122: As mentioned previously this should be abstracted to be made more readable. Will tackle in a later update.

packages/frontend/lib/index.ts
```diff
@@ -119,9 +119,9 @@ export default class Nango {
...
+                !('oauth_client_id_override' in options.credentials) ||
+                !('oauth_client_secret_override' in options.credentials)) &&
```
TBonnin, line 124: do we still need the following block below if we are now getting into `customAuth`:

```
        if (options && 'credentials' in options && 'token_id' in options.credentials && 'token_secret' in options.credentials) {
            url = this.hostBaseUrl + `/auth/tba/${providerConfigKey}${this.toQueryString(connectionId, options as ConnectionConfig)}`;
        }
```

packages/server/lib/controllers/auth/tba.ts
```diff
@@ -1,30 +1,30 @@
...
+import { zodErrorToHTTP } from '@nangohq/utils';
+import { analytics, configService, AnalyticsTypes, getConnectionConfig, connectionService } from '@nangohq/shared';
+import type { TbaAuthorization, TbaCredentials } from '@nangohq/types';
 import { defaultOperationExpiration, logContextGetter } from '@nangohq/logs';
 import { hmacCheck } from '../../utils/hmac.js';
+import { connectionCreated as connectionCreatedHook, connectionTest as connectionTestHook } from '../../hooks/hooks.js';
+
+const bodyValidation = z
+    .object({
+        token_id: z.string().nonempty(),
```
TBonnin: string nonempty is deprecated I think. `.min(1)` should be used

packages/shared/lib/services/connection.service.ts
```diff
@@ -113,6 +114,68 @@ class ConnectionService {
...
+                    connection_id: connectionId,
+                    provider_config_key: providerConfigKey,
+                    config_id: config.id as number,
+                    credentials,
+                    metadata,
+                    connection_config: connectionConfig,
+                    environment_id: environment.id
+                })
+            )
+            .returning('*');
```
TBonnin, line 172: no explicit error handling. What is returned to the end user if any of the db queries fails

packages/shared/lib/services/connection.service.ts
```diff
@@ -113,6 +114,68 @@ class ConnectionService {
...
+                credentials,
+                connection_config: connectionConfig,
+                environment_id: environment.id
+            });
+            encryptedConnection.updated_at = new Date();
+            const connection = await db.knex
+                .from<StoredConnection>(`_nango_connections`)
+                .where({ id: storedConnection.id, deleted: false })
+                .update(encryptedConnection)
+                .returning('*');
```
TBonnin, line 153: can we have an `insert/merge` instead of 2 distinct queries

packages/frontend/lib/index.ts
khaliqgant: Thanks, removed

packages/server/lib/controllers/auth/tba.ts
khaliqgant: Updated

packages/shared/lib/services/connection.service.ts
khaliqgant: This is the same logic (for better or worse) of the other upsert connections, so error handling should be added to all if anything, but given a time crunch will leave as is for now.

packages/shared/lib/services/connection.service.ts
khaliqgant: Thanks, will leave as is.

PR #2426, opened 2024-06-28
fix(syncs): use JSON schema and validate action input

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1261/pass-json-schema-to-runner
Fixes https://linear.app/nango/issue/NAN-1157/use-json-schema-to-validate-action-input

### Main changes

- **Store JSON Schema** in `sync_configs.models_json_schema`
When we deploy we use the name of the models to only store the relevant json schema. 

- Pass full **SyncConfig to NangoProps** down to the Runner
I needed more stuff, and it didn't make sense to Pick 90% of the props. 

- Validate Action Input in the Runner

### Other changes

- Fix **errors being json stringified 4 times** from runner -> jobs -> orch -> server
To achieve that I needed to pass the constructor name and payload across multiple stages. This should help with more than just my own stuff but it's not fully fixed everywhere. Error serializing/deserializing is a massive issue and a big reason why our current way of handling error is not working entirely.

- UI: show error payload in the UI

- Fix `$ref` being appended one nango.yaml only contained one model
- Fix issue with model name being wrongly accumulated across syncs

## Test

- Call `POST /action/trigger` with/without an input 
- Deploy `node packages/cli/dist/index.js deploy dev`
- Call `POST /action/trigger` with a wrong input (you should receive an error)
- 
Screenshot 2024-07-01 at 10 49 52
Screenshot 2024-07-01 at 10 49 43


packages/shared/lib/clients/orchestrator.ts
```diff
@@ -142,7 +142,13 @@ export class Orchestrator {
                 args
             });
 
-            const res = actionResult.mapError((e) => new NangoError('action_failure', { error: e.message, ...(e.payload ? { payload: e.payload } : {}) }));
+            const res = actionResult.mapError((err) => {
+                if ('payload' in err && typeof err.payload === 'object' && err.payload && 'type' in err.payload) {
+                    const pl = err.payload as unknown as NangoError;
+                    return new NangoError(pl['type'], pl.payload, pl.status);
```
bodinsamuel: This means when the runner returns a known NangoError it will no longer log `action_failure`, hope that's alright

khaliqgant (approved):

packages/logs/lib/client.ts
```diff
@@ -67,7 +67,14 @@ export class LogContextStateless {
             type: 'log',
             level: 'error',
             message,
-            error: err ? { name: err.name, message: err.message } : null,
+            error: err
+                ? {
+                      name: error instanceof Error ? error.constructor.name : err.name,
+                      message: err.message,
+                      type: 'type' in err ? (err.type as string) : null,
+                      payload: 'payload' in err ? err.payload : null
```
TBonnin, line 75: can we use `??`:
```
      type: err.type as string ?? null,
      payload: err.payload ?? null
```

packages/orchestrator/lib/clients/client.ts
```diff
@@ -370,7 +370,12 @@ export class OrchestratorClient {
     }
 
     public async failed({ taskId, error }: { taskId: string; error: Error }): Promise<Result<OrchestratorTask, ClientError>> {
-        const output = { name: error.name, message: error.message };
+        const output = {
+            name: error.name,
+            type: 'type' in error ? (error.type as string) : null,
+            message: error.message,
+            payload: 'payload' in error ? (error.payload as any) : null
+        };
```
TBonnin, line 378: same question as above wrt using `??`

packages/runner/lib/exec.ts
```diff
@@ -87,6 +87,19 @@ export async function exec(
...
                     }
+
+                    // Validate action input against json schema
+                    if (inputParams) {
+                        const val = validateInput({
+                            input: inputParams,
+                            modelName: nangoProps.syncConfig.input,
+                            jsonSchema: nangoProps.syncConfig.models_json_schema
+                        });
+                        if (val !== true) {
```
TBonnin: `if(!val)` no?

packages/shared/lib/clients/orchestrator.ts
TBonnin: I feel like the logic should live with the NangoError definition. A function like `toNangoError(error: error): NangoError || null`. 
Then here it becomes:
```
const nangoError = toNangoError(err)
if(nangoError) {
    return nangoError
}
return new NangoError(...)
```

packages/shared/lib/sdk/dataValidation.ts
```diff
@@ -0,0 +1,34 @@
...
+    const ajv = new Ajv({ allErrors: true, discriminator: true });
+    const validate = ajv.compile({ ...(jsonSchema as any), ...(jsonSchema['definitions']![modelName] as any) });
+    const val = validate(input);
+    if (val) {
+        return true;
+    }
+
+    const bag: ErrorObject[] = [];
+    for (const error of validate.errors!) {
+        if (!error.message || error.message === ' ') {
```
TBonnin: space?

TBonnin (approved):

packages/logs/lib/client.ts
bodinsamuel: type and payload do not exists in regular Error put I could put Error | NangoError in the inital type to help

packages/shared/lib/sdk/dataValidation.ts
bodinsamuel: Ah yes leftover from the CLI thanks

packages/runner/lib/exec.ts
bodinsamuel: it's `true` when it's valid and an array of errors when not valid.
But `!val` is not the opposite of `not true`, I could put Array.isArray to be clearer (and I can't use ok/err because it's in utils and I need that function in the CLI)

packages/logs/lib/client.ts
bodinsamuel: edit: nevermind it returns an object so its the only "typesafe" way with unkowns properties

packages/orchestrator/lib/clients/client.ts
bodinsamuel: Same answer

packages/shared/lib/clients/orchestrator.ts
bodinsamuel: good point, added the helper 👌🏻

PR #2437, opened 2024-07-01
fix: missing syncs in connections page

TBonnin:
we partitioned syncs by jobs.sync_id but for syncs that have not run yet this value is null. All syncs that have not run will then be in the same bucket and we will only return the first one with `WHERE job_row_number = 1` Fixing the bug by also partitioning on sync.id to properly bucket all syncs independently even the ones that have not run yet

### DIFF
--- a/packages/shared/lib/services/sync/sync.service.ts
+++ b/packages/shared/lib/services/sync/sync.service.ts
@@ -186,7 +186,7 @@ export const getSyncs = async (
                             )
                         END as latest_sync `),
                     db.knex.raw(
-                        `ROW_NUMBER() OVER (PARTITION BY ${SYNC_JOB_TABLE}.sync_id ORDER BY ${SYNC_JOB_TABLE}.deleted ASC, ${SYNC_JOB_TABLE}.updated_at DESC) as job_row_number`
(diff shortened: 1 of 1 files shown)


packages/shared/lib/services/sync/sync.service.ts
```diff
@@ -186,7 +186,7 @@ export const getSyncs = async (
                             )
                         END as latest_sync `),
                     db.knex.raw(
-                        `ROW_NUMBER() OVER (PARTITION BY ${SYNC_JOB_TABLE}.sync_id ORDER BY ${SYNC_JOB_TABLE}.deleted ASC, ${SYNC_JOB_TABLE}.updated_at DESC) as job_row_number`
+                        `ROW_NUMBER() OVER (PARTITION BY ${TABLE}.id, ${SYNC_JOB_TABLE}.sync_id ORDER BY ${SYNC_JOB_TABLE}.updated_at DESC) as job_row_number`
```
TBonnin, line 189: also deleting the `ORDER BY ${SYNC_JOB_TABLE}.deleted ASC` because we always filter on `${SYNC_JOB_TABLE}.deleted = FALSE` [below](https://github.com/NangoHQ/nango/pull/2437/files#diff-7be9b1d2fee1ed263243cfe652dbb029abdc165631a55b0677119bf8fb3a743cR205)

khaliqgant (approved):

PR #2439, opened 2024-07-01
fix: pinning CLI dependencies

TBonnin:
Customers might pin the CLI version but nango dependencies are not pinned which leads to issue if those packages are updated, creating a mismatch between CLI version and dependencies version. Until we come up with a real versioning strategy, pinning the nango dependencies to avoid any version mismatch


scripts/publish.sh
```diff
@@ -88,6 +88,7 @@ bump_other_pkg "nango-yaml" "types"
 # NangoYaml
 bump_and_npm_publish "@nangohq/nango-yaml" "$VERSION"
 bump_other_pkg "cli" "nango-yaml"
+bump_other_pkg "shared" "nango-yaml"
```
TBonnin, line 91: fixing the missing upgrade of nango-yaml as shared dep

khaliqgant (approved):

scripts/publish.sh
bodinsamuel: thanks 🙏🏻  probably got lost in a merge conflict

bodinsamuel (approved):

PR #2440, opened 2024-07-01
chore(CLI): force confirmation when deletedSyncs

TBonnin:
This is the first step to address https://linear.app/nango/project/data-loss-prevention-255ab3eaad30/overview
More specifically addressing sync deletion/renaming.

Forcing confirmation when deploying via the CLI and sync(s) with connection(s) is(are) being deleted
Ignoring `--allow-confirm` flag since it is used for more than one scenario.
Behavior can be by-passed with explicit --allow-destrictive flag at the risk of deploying a change that could cause records to be deleted

To deploy syncs deletion users must pass the very explicit `--allow-destructive` flag to their CI or deploy manually with this flag.

### DIFF
--- a/packages/cli/lib/services/deploy.service.ts
+++ b/packages/cli/lib/services/deploy.service.ts
@@ -99,7 +99,7 @@ class DeployService {
     }
 
     public async prep({ fullPath, options, environment, debug = false }: { fullPath: string; options: DeployOptions; environment: string; debug?: boolean }) {
-        const { env, version, sync: optionalSyncName, action: optionalActionName, autoConfirm } = options;
+        const { env, version, sync: optionalSyncName, action: optionalActionName, autoConfirm, allowDestructive } = options;
         await verificationService.necessaryFilesExist({ fullPath, autoConfirm, checkDist: false });
 
         await parseSecretKey(environment, debug);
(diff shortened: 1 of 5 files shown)


khaliqgant (commented):
Was testing this but faced an error that is unrelated but blocking

```
Error deploying the scripts with the following error: {
  "error": {
    "message": "An unhandled error of type 'error_updating_sync_schedule_frequency' with payload '{\"syncId\":\"b05e3ed1-1ab3-4fbd-8246-ac8bec9d0ec5\",\"environmentId\":2,\"interval\":\"every hour\"}' has occurred",
    "code": "unhandled_error_updating_sync_schedule_frequency",
    "payload": {
      "syncId": "b05e3ed1-1ab3-4fbd-8246-ac8bec9d0ec5",
      "environmentId": 2,
      "interval": "every hour"
    }
  }
}
```

It might be because this schedule was never created in the orchestrator but was previously in temporal and was never migrated – so this might be an edge case that we don't need to handle. Unsure completely though.

packages/cli/lib/services/deploy.service.ts
```diff
@@ -147,57 +147,77 @@ class DeployService {
...
+                    )
+                );
+                deletedSyncsConnectionsCount += sync.connections;
+            }
 
-                const confirmation = await promptly.confirm('Do you want to continue y/n?');
+            // force confirmation :
+            // - if auto-confirm flag is not set
+            // - OR if there are deleted syncs with connections (and allow-destructive flag is not set)
+            const shouldConfirmDestructive = deletedSyncs.length > 0 && deletedSyncsConnectionsCount > 0 && !allowDestructive;
```
bodinsamuel: Shouldn't it be
```suggestion
            const shouldConfirmDestructive = (deletedSyncs.length > 0 || deletedSyncsConnectionsCount > 0) && !allowDestructive;
```

bodinsamuel (approved):

khaliqgant (approved):

packages/cli/lib/services/deploy.service.ts
TBonnin: I was not planning to ask for confirmation when deleting a sync which has no connection since it is not deleting any records or anything that cannot be recovered.
True that `deletedSync.length > 0` is not necessary though since if this is the case `deletedSyncsConnectionsCount` would also be `0`. Will remove it

TBonnin:
> Was testing this but faced an error that is unrelated but blocking
> 
> ```
> Error deploying the scripts with the following error: {
>   "error": {
>     "message": "An unhandled error of type 'error_updating_sync_schedule_frequency' with payload '{\"syncId\":\"b05e3ed1-1ab3-4fbd-8246-ac8bec9d0ec5\",\"environmentId\":2,\"interval\":\"every hour\"}' has occurred",
>     "code": "unhandled_error_updating_sync_schedule_frequency",
>     "payload": {
>       "syncId": "b05e3ed1-1ab3-4fbd-8246-ac8bec9d0ec5",
>       "environmentId": 2,
>       "interval": "every hour"
>     }
>   }
> }
> ```
> 
> It might be because this schedule was never created in the orchestrator but was previously in temporal and was never migrated – so this might be an edge case that we don't need to handle. Unsure completely though.

did you by any chance manage to check the orchestrator logs? if your theory is correct you should see a `PUT /schedule/... -> 404`

TBonnin:
@bastienbeurier I am ready to merge this change. Do we need to let customers know before I merge and publish a new version of the CLI? I am thinking of the customers using nango in CI for instance

bastienbeurier:
> @bastienbeurier I am ready to merge this change. Do we need to let customers know before I merge and publish a new version of the CLI? I am thinking of the customers using nango in CI for instance

I would do a developer update on the #from-nango channel of the [community slack](https://nango.dev/slack) with the same message layout as this [developer update](https://nango-community.slack.com/archives/C03QBHSMPUM/p1719391407116769).

TBonnin:
@bodinsamuel @khaliqgant can I have another review. Deploy is now failing when running in CI and requires confirmation (which cannot be granted when running in CI) https://github.com/NangoHQ/nango/pull/2440/commits/53e15f79a7b07d27714fbb58a38d9d28b667242b

Note that customers would need to update their github action to something like this:
```
name: Deploy Nango Integrations
'on':
    push:
        branches:
            - main
        workflow_dispatch:
            inputs:
                stage:
                    type: choice
                    description: 'stage to deploy to, defaults to dev'
                    required: true
                    default: 'dev'
                    options:
                        - dev
                        - prod
                allowDestructive:
                    type: choice
                    description: 'deploy destructive changes without confirmation'
                    required: true
                    default: false
                    options:
                        - true
                        - false
jobs:
    deploy:
        runs-on: ubuntu-latest
        steps:
            - name: Check if NANGO_DEV_SECRET_KEY is set
              run: |
                  if [ -z "${{ secrets.NANGO_DEV_SECRET_KEY }}" ]; then
                    echo "Error: NANGO_DEV_SECRET_KEY secret is not set"
                    exit 1
                  fi
            - name: Checkout code
              uses: actions/checkout@v4
            - name: Install Nango CLI
              run: npm install nango -g
            - name: run Nango deploy
              run: |
                  cd nango-integrations
                  if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
                      stage="${{ inputs.stage }}"
                  elif [ "${{ github.ref }}" == "refs/heads/main" ]; then
                      stage="prod"
                  else
                      stage="dev"
                  fi
                  nango deploy "$stage" --auto-confirm ${{ inputs.allowDestructive == 'true' && '--allow-destructive' || '' }}
```

bodinsamuel (approved):

khaliqgant:
> @bodinsamuel @khaliqgant can I have another review. Deploy is now failing when running in CI and requires confirmation (which cannot be granted when running in CI) [53e15f7](https://github.com/NangoHQ/nango/commit/53e15f79a7b07d27714fbb58a38d9d28b667242b)
> 
> Note that customers would need to update their github action to something like this:
> 
> ```
> name: Deploy Nango Integrations
> 'on':
>     push:
>         branches:
>             - main
>         workflow_dispatch:
>             inputs:
>                 stage:
>                     type: choice
>                     description: 'stage to deploy to, defaults to dev'
>                     required: true
>                     default: 'dev'
>                     options:
>                         - dev
>                         - prod
>                 allowDestructive:
>                     type: choice
>                     description: 'deploy destructive changes without confirmation'
>                     required: true
>                     default: false
>                     options:
>                         - true
>                         - false
> jobs:
>     deploy:
>         runs-on: ubuntu-latest
>         steps:
>             - name: Check if NANGO_DEV_SECRET_KEY is set
>               run: |
>                   if [ -z "${{ secrets.NANGO_DEV_SECRET_KEY }}" ]; then
>                     echo "Error: NANGO_DEV_SECRET_KEY secret is not set"
>                     exit 1
>                   fi
>             - name: Checkout code
>               uses: actions/checkout@v4
>             - name: Install Nango CLI
>               run: npm install nango -g
>             - name: run Nango deploy
>               run: |
>                   cd nango-integrations
>                   if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
>                       stage="${{ inputs.stage }}"
>                   elif [ "${{ github.ref }}" == "refs/heads/main" ]; then
>                       stage="prod"
>                   else
>                       stage="dev"
>                   fi
>                   nango deploy "$stage" --auto-confirm ${{ inputs.allowDestructive == 'true' && '--allow-destructive' || '' }}
> ```

We should make an issue for this to make sure we go through customer repos and update this.

TBonnin:
> We should make an issue for this to make sure we go through customer repos and update this.

Do you know if we have a list of customers that are using nango deploy github action?

bastienbeurier:
> > We should make an issue for this to make sure we go through customer repos and update this.
> 
> Do you know if we have a list of customers that are using nango deploy github action?

Sharing on Slack!

bastienbeurier:
> > We should make an issue for this to make sure we go through customer repos and update this.
> 
> Do you know if we have a list of customers that are using nango deploy github action?

Sharing on Slack!

TBonnin:
FYI @khaliqgant @bodinsamuel It was agreed with @bastienbeurier to be reactive regarding the update of customers github action and not try to update them all right away

PR #2442, opened 2024-07-02
fix(node): correctly export Auth and Webhooks type

bodinsamuel:
## Describe your changes

I tried upgrading the sample-app but we changed the Webhooks type and realised we don't export the Auth types and realised the auth types were desync between shared/node/types etc..

- Node: Export Webhook type with an aggregated type `NangoWebhookBody`
- Node: Export auth types from source of truth
- Node: Export HTTP from source of truth
- Shared: Remove `AuthModes` enum
- SDK: add missing Auth types

NB: ideally we don't re-export anything from `types` in `node-client` but it's easier for retro-compat


packages/node-client/lib/index.ts
```diff
@@ -236,13 +242,24 @@ export class Nango {
...
+        | OAuth1Token
+        | BasicApiCredentials
+        | ApiKeyCredentials
+        | AppCredentials
+        | OAuth2ClientCredentials
+        | AppStoreCredentials
+        | UnauthCredentials
+        | CustomCredentials
+        | TbaCredentials
+    > {
```
bodinsamuel, line 256: Not entirely sure it was on purpose, I feel like it was incorrect unless only some credentials are returned from the api

TBonnin: I don't know if it was done on purpose @khaliqgant might know

TBonnin (approved):
looks good

packages/node-client/lib/index.ts
khaliqgant: Nope, was just missing some of the recently added types but this method should actually be updated to actually return just the access token but that's for another task/issue.

PR #2443, opened 2024-07-02
fix: frequency when no quantity

TBonnin:
if frequency is set to `every hour` the connection page shows `h`.
This commit ensures such frequencies shows as `1h` (or `1d`, ...) 

Also adding abbreviation for month and year even though we don't have a lot of such syncs, it is just to be consistent

Screenshot 2024-07-02 at 10 38 11

### DIFF
--- a/packages/webapp/src/utils/utils.tsx
+++ b/packages/webapp/src/utils/utils.tsx
@@ -219,8 +219,8 @@ export function formatFrequency(frequency: string): string {
         hour: 'h',
         days: 'd',
         day: 'd',
-        months: 'mon',
-        month: 'mon',
+        months: 'mos',
+        month: 'mo',
         years: 'y',
         year: 'y'
     };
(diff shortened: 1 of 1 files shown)


packages/webapp/src/utils/utils.tsx
```diff
@@ -218,10 +218,20 @@ export function formatFrequency(frequency: string): string {
         hours: 'h',
         hour: 'h',
         days: 'd',
-        day: 'd'
+        day: 'd',
+        months: 'mon',
+        month: 'mon',
```
bodinsamuel: why not `M` less clear but consistent in term of size?

bodinsamuel (approved):

packages/webapp/src/utils/utils.tsx
TBonnin: I thought about that. As you said it is consistent but less clear. I don't mind either way. @khaliqgant any preference?

bodinsamuel: not pushing just saying 👍🏻

khaliqgant: month abbreviation should actually be `mo` if singular and `mos` if plural

khaliqgant (approved):

PR #2444, opened 2024-07-02
fix(validation): use feature flag

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1330/use-feature-flags

- Use feature flags to easily enable/disable this check at runtime
We expect most people to not deploy in the coming days but it's just to be ultra safe. Also once the flag is in the runner, it's no longer possible to check it again so if the sync is long running it's not going to help.

- Add log when feature is disabled
- Add metric to log volume
- Fix: add missing tag error on span


packages/shared/lib/services/sync/run.utils.ts
```diff
@@ -0,0 +1,22 @@
...
+    validateSyncRecords: boolean;
+    validateSyncMetadata: boolean;
+}
+
+export async function getRunnerFlags(): Promise<RunnerFlags> {
+    const validateActionInput = await featureFlags.isEnabled('runner.validateActionInput', 'global', true);
+    const validateActionOutput = await featureFlags.isEnabled('runner.validateActionOutput', 'global', true);
+    const validateSyncRecords = await featureFlags.isEnabled('runner.validateSyncRecords', 'global', true);
+    const validateSyncMetadata = await featureFlags.isEnabled('runner.validateSyncMetadata', 'global', true);
+
```
TBonnin, line 17: nitpicking: you could fetch all the flags concurrently

TBonnin (approved):

packages/shared/lib/services/sync/run.utils.ts
bodinsamuel: good point

PR #2450, opened 2024-07-03
fix(webapp): fix spinner button

khaliqgant:
## Describe your changes
The enabled button was displaced

Before:
image

After:
image

NAN-1333

### DIFF
--- a/packages/webapp/src/pages/Integration/components/EnableDisableSync.tsx
+++ b/packages/webapp/src/pages/Integration/components/EnableDisableSync.tsx
@@ -266,12 +266,14 @@ export default function EnableDisableSync({
                 modalOkLink={modalOkButtonLink}
                 modalCancelLink={modalCancelButtonLink}
             />
-            {showSpinner && (!('version' in flow) || flow.version === null) && modalShowSpinner && (
-                <span className="mr-2">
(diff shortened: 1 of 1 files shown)


packages/webapp/src/pages/Integration/components/EnableDisableSync.tsx
```diff
@@ -266,12 +266,14 @@ export default function EnableDisableSync({
...
-            )}
-            <ToggleButton enabled={enabled} onChange={() => toggleSync(flow)} />
+            <div className="flex">
+                {showSpinner && (!('version' in flow) || flow.version === null) && modalShowSpinner && (
+                    <span className="mr-1">
+                        <Spinner size={1} />
+                    </span>
+                )}
+                <ToggleButton enabled={enabled} onChange={() => toggleSync(flow)} />
+            </div>
```
bodinsamuel, line 276: you can use `gap-N` instead of having to put margin

bodinsamuel (approved):

packages/webapp/src/pages/Integration/components/EnableDisableSync.tsx
khaliqgant: Thanks.

PR #2454, opened 2024-07-03
feat: validate action output, support dryun

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1160/make-sure-it-works-in-dryrun-too
Fixes https://linear.app/nango/issue/NAN-1161/use-json-schema-to-validate-action-output

### Main

- CLI: validate input and output
There is no way to disable this locally.
- Runner: validate output
- Runner: disable logging user generated error
- Sync Run: fix output in dry run

### Other
- CLI: change output of dryrun to clarify what's coming from the SDK or not
- SDK/node-client: Fix invalid jsdoc comments

## Tests

- Try running dry mode or execute an action 

Screenshot 2024-07-03 at 14 46 19


packages/cli/lib/services/dryrun.service.ts
```diff
@@ -255,14 +264,26 @@ export class DryRunService {
         });
 
         try {
+            console.log('---');
             const secretKey = process.env['NANGO_SECRET_KEY'];
             const results = await syncRun.run(lastSyncDate, true, secretKey, process.env['NANGO_HOSTPORT']);
+            console.log('---');
```
TBonnin, line 270: console.log

TBonnin (approved):

packages/cli/lib/services/dryrun.service.ts
bodinsamuel: it's what we use across the cli (the logger is in utils which is not a published dep)

khaliqgant (approved):
Looks good and works well.

This is going to break a lot of scripts. There is an option to disable this during execution in the runner, correct?

khaliqgant:
> Looks good and works well.

This is going to break a lot of scripts. There is an option to disable this during execution in the runner, correct?

bodinsamuel:
> This is going to break a lot of scripts. There is an option to disable this during execution in the runner, correct?

Yeah we just talked about that with Bastien, we'll add a way to disable validation in dryrun mode and it will stay as soft validation in production for a few weeks/month at least

PR #2463, opened 2024-07-04
feat(integration-templates): [nan-1328] update xero more after test

khaliqgant:
## Describe your changes
Test xero and make type adjustments and cleanup

NAN-1168


integration-templates/xero/types.ts
```diff
@@ -0,0 +1,111 @@
...
+
+export interface Payment {
+    PaymentID: string;
+    Date: string; // Date in ISO string format
+    BankAmount: number;
+    Amount: number;
+    CurrencyRate: number;
+    PaymentType: 'ACCPAYPAYMENT' | 'ACCRECPAYMENT';
+    Status: 'AUTHORISED' | 'PENDING' | 'VOIDED';
+    UpdatedDateUTC: string; // Date in ISO string format
```
bodinsamuel, line 102: date is not working for you?

bodinsamuel (approved):

integration-templates/xero/types.ts
khaliqgant: Well, it is but TS is complaining and the data coming from the API is actually a string.

PR #2464, opened 2024-07-04
feat(cli-reference): [nan-1168] allow a user to reference a file to use for the input or metadata

khaliqgant:
## Describe your changes
Allow a user to reference a file:
```
npx nango dryrun create-invoice x -i @fixtures/xero/invoice.json
```

NAN-1168


packages/cli/lib/services/dryrun.service.ts
```diff
@@ -183,18 +183,40 @@ export class DryRunService {
...
-        }
+        let normalizedInput;
 
-        try {
-            stubbedMetadata = JSON.parse(rawStubbedMetadata as unknown as string);
-        } catch {
-            stubbedMetadata = rawStubbedMetadata;
+        // the user can reference a file for input
+        // if it includes both '@' and 'json'
+        if (actionInput && actionInput.toString().includes('@') && actionInput.toString().includes('.json')) {
```
bodinsamuel:
```suggestion
        if (actionInput && actionInput.toString().includes('@') && actionInput.toString().endsWith('.json')) {
```

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -183,18 +183,40 @@ export class DryRunService {
...
+            // fetch the file contents
+            const fileContents = localFileService.readFile(actionInput.toString());
+            if (!fileContents) {
+                console.log(chalk.red('The file could not be read. Please make sure it exists.'));
+                return;
+            }
+            normalizedInput = JSON.parse(fileContents);
```
bodinsamuel: JSON.parse can throw an error, note that the error message is helpful but it could be better presented to the user

bodinsamuel (approved):
🎰
Works well, should be documented in the CLI option!

khaliqgant:
Thanks @bodinsamuel, feedback has been addressed

PR #2465, opened 2024-07-04
fix(cli): [nan-1344] allow type imports

khaliqgant:
## Describe your changes
Skip type imports for compilation

NAN-1344


bodinsamuel (approved):
Works with `import type { AnyMxRecord } from 'dns';`
but doesn't with `import { type AnyMxRecord } from 'dns';`

khaliqgant:
> import { type AnyMxRecord } from 'dns';

Updated with https://github.com/NangoHQ/nango/pull/2465/commits/dbf9f0d6dea5591b256c538f62b3930d3ac8a125

PR #2469, opened 2024-07-04
feat(integrations): add support for zoominfo

hassan254-prog:
## Describe your changes

- Add support for zoominfo

[EXT-98](https://linear.app/nango/issue/EXT-98/add-support-for-zoominfo-api)

- Authentication in ZoomInfo is a little bit different, For us to receive a valid JWT we will need to send the following request

```
curl --location 'https://api.zoominfo.com/authenticate' \
--data '{
    "username": "username",
    "password": "password"
}'
```
This will return :
```
{
"success": true,
"jwt": "token"
}
```

- The token is valid for 60 mins but from the provider's [docs](https://api-docs.zoominfo.com/#477888fc-8308-4645-81ca-ca7a6d7ba3d1) it is recommended to refresh it every 55 minutes.

- Once generated, we can use it as follows: `Authorization: Bearer YOUR_ACCESS_TOKEN`

- This authentication is almost similar to `OAUTH2_CC` so imo there is no need to reinvent the wheel


packages/shared/lib/services/connection.service.ts
```diff
@@ -821,11 +821,13 @@ class ConnectionService {
                     expiresAt = parseTokenExpirationDate(rawCreds['expires_at']);
                 } else if (rawCreds['expires_in']) {
                     expiresAt = new Date(Date.now() + Number.parseInt(rawCreds['expires_in'], 10) * 1000);
+                } else {
+                    expiresAt = new Date(Date.now() + 55 * 60 * 1000);
```
bodinsamuel: can you make a constant at the file level, and a little explanation

packages/shared/lib/services/connection.service.ts
```diff
@@ -811,7 +811,7 @@ class ConnectionService {
             }
 
             case 'OAUTH2_CC': {
-                if (!rawCreds['access_token'] && !rawCreds['data']['token']) {
+                if (!rawCreds['access_token'] && !(rawCreds['data'] && rawCreds['data']['token']) && !rawCreds['jwt']) {
```
bodinsamuel, line 815: all oauth2_cc have a jwt in their raw creds?

bodinsamuel (commented):
👌🏻 
Not entirely sure about some of the changes, I don't have enough knowledge regarding those

packages/shared/lib/services/connection.service.ts
hassan254-prog: @bodinsamuel, just updated the PR. I have defined it within the function itself so that `DEFAULT_EXPIRES_AT` can reflect the current time dynamically each time it's accessed.

packages/shared/lib/services/connection.service.ts
hassan254-prog: Not necessarily, since `JWT` is not universally present in all `OAuth2_CC` instances. The current structure ensures flexibility to handle different scenarios in `rawCreds`

packages/shared/lib/services/connection.service.ts
bodinsamuel: what I mean is that you can put this expiration as a constant without changing the code
`expiresAt = new Date(Date.now() + DEFAULT_EXPIRES_AT_MS);`

packages/shared/lib/services/connection.service.ts
bodinsamuel: ah yes I see, all good 👍🏻

bodinsamuel (approved):

packages/shared/lib/services/connection.service.ts
hassan254-prog: ah yes 💯, updated in [here](https://github.com/NangoHQ/nango/pull/2469/commits/0c7c385da0e99787a7a880e7bc9957bf829ef937)

TBonnin (approved):

PR #2472, opened 2024-07-05
fix(ui): toggle was not in sync with actual state

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1357/ui-sync-toggle-bug

- Sync toggle was calling setEnabled to soon
- Change a bit the function name to clarify

### DIFF
--- a/packages/webapp/src/pages/Integration/components/EnableDisableSync.tsx
+++ b/packages/webapp/src/pages/Integration/components/EnableDisableSync.tsx
@@ -75,7 +75,7 @@ export default function EnableDisableSync({
         setModalTitleColor('text-white');
     };
 
-    const enableSync = (flow: Flow) => {
+    const showEnableSyncModal = (flow: Flow) => {
         resetModal();
         setModalTitle(`Enable ${flow.type}?`);
         setModalTitleColor('text-white');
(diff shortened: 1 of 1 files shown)


TBonnin (approved):

packages/webapp/src/pages/Integration/components/EnableDisableSync.tsx
```diff
@@ -196,14 +196,15 @@ export default function EnableDisableSync({
             success = await createNewFlow(flowPayload);
         }
 
-        if (success && flow?.type === 'sync') {
+        if (success) {
```
khaliqgant, line 199: have a question on this removal

bodinsamuel: please write it :D 
From my understanding there is no reason to have two branch here, both action and sync goes through this function and both need to be enabled in the UI. Let me know if that's not the case

khaliqgant: Tested locally and works as expected.

khaliqgant (approved):

PR #2473, opened 2024-07-05
feat(internal-integration-templates): [nan-1343] allow integration templates to be run from the repo

khaliqgant:
## Describe your changes
Makes for an easier workflow to be able to run and test integration templates directly from the repo instead of making another directory and then copying and pasting the files back and forth.

So a command would look like this:

```
bash scripts/run-integration-template.bash asana dryrun ${sync|action} ${connection_id}
```

NAN-1343


scripts/run-integration-template.bash
```diff
@@ -0,0 +1,15 @@
...
+INTEGRATION=$1
+shift
+
+NANGO_SECRET_KEY_DEV=$(grep NANGO_SECRET_KEY_DEV .env | cut -d '=' -f2)
+
+cd integration-templates/$INTEGRATION
+
+echo $NANGO_SECRET_KEY_DEV
+
+NANGO_SECRET_KEY_DEV=$NANGO_SECRET_KEY_DEV NANGO_HOSTPORT=http://localhost:3003 _NANGO_IN_REPO=true npx nango compile
```
TBonnin: will the CLI not complain it is not being run in a `nango-integrations` folder?

scripts/run-integration-template.bash
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+INTEGRATION=$1
+shift
+
+NANGO_SECRET_KEY_DEV=$(grep NANGO_SECRET_KEY_DEV .env | cut -d '=' -f2)
+
+cd integration-templates/$INTEGRATION
+
+echo $NANGO_SECRET_KEY_DEV
```
TBonnin: I will not echo the secret key so it doesn't end up in our bash history

packages/cli/lib/services/compile.service.ts
```diff
@@ -200,6 +200,10 @@ async function compile({
     const syncConfig = [...providerConfiguration.syncs, ...providerConfiguration.actions].find((sync) => sync.name === file.baseName);
     const type = syncConfig?.type || 'sync';
 
+    if (process.env['_NANGO_IN_REPO']) {
+        file.inputPath = file.inputPath.split('/').slice(0, -1).join('/') + `/${type}s/` + file.baseName + '.ts';
+    }
+
```
TBonnin: not a huge fan of shipping code to customers that would only be used for dev work but I understand the pain of having to move the built-in integrations during dev work. Can we have the bash script below create a temporary folder structure expected by nango (creating folder and copying file) and do the cleanup. Like this we don't have to have any code changes and the hack is self-contained in the bash script?

scripts/run-integration-template.bash
khaliqgant: Whoops. That was debug left over. Removing

packages/cli/lib/services/compile.service.ts
khaliqgant: Yeah, I was trying to avoid that initially but I agree in a sense it feels dirty to mess with the logic just for our internal usage.

Updated to create a temp directory like you suggested.

scripts/run-integration-template.bash
```diff
@@ -0,0 +1,27 @@
...
+popd () {
+    command popd "$@" > /dev/null
+}
+
+INTEGRATION=$1
+shift
+
+NANGO_SECRET_KEY_DEV=$(grep NANGO_SECRET_KEY_DEV .env | cut -d '=' -f2)
+
+rm -rf temp
```
TBonnin: nitpick: I would use a less generic name for the temp directory, something like `tmp/run-integration-template`. Or even putting it in the OS `/tmp` dir. Up to you though.

TBonnin (approved):

scripts/run-integration-template.bash
khaliqgant: Updated

PR #2474, opened 2024-07-05
fix: handle big payload across the stack

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1363/big-payload-is-crashing-the-action-execution

A big output from an action would break logging, orchestrator and the UI (all for different reasons).
The error management being still not good it's messy to handle and propagate the correct error but I think I managed something not too ugly.

- Runner: Handle big payload validation 
- Runner: Better handle JSON.stringification when logging and avoid error by truncating before sending
- Orch: stop stringifying errors
- Orch: return json response in the payload as json
- Orch: handle big payload (> 10mb) preventing task to be marked as completed
- UI: handle big payload in the message panel

## Test

```ts

export default async function runAction(nango: NangoSync): Promise<string> {
  return '1'.repeat(1_999_999_999);
}
```

Screenshot 2024-07-05 at 18 37 41


packages/orchestrator/lib/clients/processor.ts
```diff
@@ -121,26 +122,29 @@ export class OrchestratorProcessor {
                         // task was aborted. No need to set it as failed
                         return;
                     }
+
                     const setFailed = await this.orchestratorClient.failed({ taskId: task.id, error: res.error });
                     if (setFailed.isErr()) {
-                        logger.error(`failed to set task ${task.id} as failed: ${stringifyError(setFailed.error)}`);
+                        throwIfPayloadTooBig(setFailed.error);
```
TBonnin, line 128: not a huge fan of now having 2 error handling paths. can we instead check the error and call `orchestratorClient.failed()` again with a `payload_too_big` error.

bodinsamuel: that's basically what it does, throwing will call `failed()` with the error name, and I don't have to copy paste 2 more times the call + isErr()
let me know if you still fell like it's worth the added cost

TBonnin: I see your point. I wonder if this is not the right level to handle this error, exposing too much of the internals of the orchestrator. What if `.failed()` `.succeed()` and all functions passing a payload were dealing with the payload to big error internally, calling `failed()` themselves. Otherwise everywhere those functions are being called must handle this scenario explicitely

bodinsamuel: Definitely a weird place to handle it, but also I can't really see better place because the only function around processTask is processingLoop and it's generic too
nb: it's the only place where it matters because we still want to mark the task as failed no matter the input size.

TBonnin: the other place would be in the [orchestrator](https://github.com/NangoHQ/nango/blob/525e34011f3612acca6813023e40d92c816bed05/packages/orchestrator/lib/routes/v1/tasks/putTaskId.ts#L54), no exposing the payload_too_big error. I understand if you don't want to take care of this in this PR, if not add a TODO and a ticket and assign it to me

TBonnin (approved):

packages/orchestrator/lib/clients/processor.ts
bodinsamuel: the thing is this error pops before entering the actual endpoint (because it's a middleware), so it's tricky

bodinsamuel: I'll create a linear issue

TBonnin: in the orchestrator client then: https://github.com/NangoHQ/nango/blob/525e34011f3612acca6813023e40d92c816bed05/packages/orchestrator/lib/clients/client.ts#L383

Still a level higher and "users" of the client won't have to deal with this internal error. Would need to be handled in `failed()`, `succeed()`, etc...

PR #2475, opened 2024-07-05
chore(CLI): Protect against sync model renaming destructive consequences

TBonnin:
Force confirmation when a model is being deleted or renamed.
The consequences of renaming a model are highlighted in our docs at https://docs.nango.dev/reference/integration-configuration#changing-a-model-or-script, with this PR we are making sure users confirm model renaming/deletion at deploy time

https://linear.app/nango/issue/NAN-li/protect-against-sync-model-renaming-destructive-consequences

### DIFF
--- a/docs-v2/reference/cli.mdx
+++ b/docs-v2/reference/cli.mdx
@@ -70,7 +70,7 @@ Global command flags:
 
 ```bash
 # Command flag to auto-confirm all prompts (useful for CI).
-# Note: Destructive changes (like removing a sync) requires confirmation, even when --auto-confirm is set. To bypass this restriction, the --allow-destructive flag can be passed to nango deploy.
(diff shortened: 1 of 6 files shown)


bodinsamuel (approved):
Works well 🚨
I think we could improve the output and the display in general, it's easy to confuse the json output with the compilation, and big red paragraph is not easy to read

TBonnin:
> Works well 🚨 I think we could improve the output and the display in general, it's easy to confuse the json output with the compilation, and big red paragraph is not easy to read

@bastienbeurier any suggestions for the error message. I tried to be succinct but I am sure it can be improved. Here is how it looks now:
Screenshot 2024-07-08 at 09 51 04

TBonnin:
Merging for now. Let me know @bastienbeurier if you think we should tweak the error message.

bastienbeurier:
> Merging for now. Let me know @bastienbeurier if you think we should tweak the error message.

Sorry for the delay on this. The message looks good to me, thanks!

PR #2449, opened 2024-07-03
feat(integrations): add support for Coros

henrymgarrett:
## Describe your changes
Adding support for the Coros API Provider.

Coros provides a test environment, so I followed the pattern for sandbox integrations, and created a `coros-sandbox` provider as well.

Screenshot 2024-07-02 at 8 49 08 PM
Screenshot 2024-07-02 at 8 49 17 PM

PR #2476, opened 2024-07-05
feat(oauth2): add coros provider to list of clientProviders

hassan254-prog:
## Describe your changes

- Add `coros` and `coros-sandbox` to the list of `shouldUseProviderClient()`. The refresh of tokens in `coros` is a little bit different; when the token is refreshed, the only response received is:
```
{
"result": "0000",
"message": "OK"
}
```
We need to let Nango know when the token will expire.
[EXT-101](https://linear.app/nango/issue/EXT-101/add-support-for-coros-api)

This will build on top of [2449](https://github.com/NangoHQ/nango/pull/2449)


packages/shared/lib/clients/provider.client.ts
```diff
@@ -373,6 +388,66 @@ class ProviderClient {
...
+            grant_type: 'authorization_code'
+        };
+        try {
+            const response = await axios.post(tokenUrl, body, { headers: headers });
+            if (response.status === 200 && response.data !== null) {
+                return {
+                    access_token: response.data['access_token'],
+                    refresh_token: response.data['refresh_token'],
+                    expires_in: response.data['expires_in'],
+                    openId: response.data['openId']
```
TBonnin, line 409: should we make sure those attributes actually exists? maybe using zod validation? I know we are not doing it for the other ones but it is never too late to start. :D

packages/shared/lib/clients/provider.client.ts
```diff
@@ -373,6 +388,66 @@ class ProviderClient {
...
+
+        const body = {
+            client_id: clientId,
+            refresh_token: refreshToken,
+            client_secret: clientSecret,
+            grant_type: 'refresh_token'
+        };
+        try {
+            const response = await axios.post(refreshTokenUrl, body, { headers: headers });
+            if (response.status === 200 && response.data !== null && response.data.result === '0000') {
```
TBonnin, line 442: can we add a comment to describe what's coros api response in case of refresh success?

TBonnin (approved):

packages/shared/lib/clients/provider.client.ts
bodinsamuel: This file needs a full a refacto tbh not sure it's worth starting an initiative

packages/shared/lib/clients/provider.client.ts
```diff
@@ -373,6 +388,66 @@ class ProviderClient {
...
+                return {
+                    access_token: accessToken,
+                    refresh_token: refreshToken,
+                    expires_in: corosExpiresIn
+                };
+            }
+            throw new NangoError('coros_token_refresh_request_error');
```
bodinsamuel: we only have one error for refresh `refresh_token_external_error` + payload (should be json)

bodinsamuel (commented):
Minor comments otherwise lgtm

packages/shared/lib/clients/provider.client.ts
hassan254-prog: Thank you for pointing this out 😊!

packages/shared/lib/clients/provider.client.ts
hassan254-prog: @TBonnin , I agree 💯, just [updated](https://github.com/NangoHQ/nango/pull/2476/commits/afbe4fd40e82d562f16f8c13f77188b51a75df1a) my PR. Thanks!

packages/shared/lib/clients/provider.client.ts
```diff
@@ -373,6 +388,72 @@ class ProviderClient {
...
+                    access_token: response.data['access_token'],
+                    refresh_token: response.data['refresh_token'],
+                    expires_in: response.data['expires_in'],
+                    openId: response.data['openId']
+                };
+            }
+
+            throw new NangoError('coros_token_request_error');
+        } catch (e: any) {
+            throw new NangoError('coros_token_request_error', e.message);
```
bodinsamuel: same message about refresh_token_external_error

hassan254-prog: @bodinsamuel, wouldn't it be better if we added `request_token_external_error` to the list of NangoError `types`?

bodinsamuel: ah yes I didn't saw it wasn't a refresh 👌🏻

hassan254-prog: Thanks for clarifying this 🙏. Updated in [here](https://github.com/NangoHQ/nango/pull/2476/commits/c37778ebe37e04d4938eba0465885f4e4c47a4be).

bodinsamuel (approved):

PR #2477, opened 2024-07-06
feat(connections): Add optional display name, customer domain, and customer email context to connections

Daspy11:
## Describe your changes
When adding a new connection using `nango.create` or `nango.auth`, you can now optionally specify a display name, customer domain, and customer email. These are stored in the connection config field in the connection. The dashboard has been updated correspondingly to display this information if it is present.

Loom explaining the changes in more detail coming shortly. Will start work on bonus items from the brief tomorrow.

N/A

- [ ] I added tests, otherwise the reason is: I may still do, need to examine what's in place already more closely.
- [ ] I added observability, otherwise the reason is: I may still do, pending discussion
- [ ] I added analytics, otherwise the reason is: I may still do, pending discussion

### DIFF
--- a/docs-v2/integrate/guides/authorize-an-api.mdx
+++ b/docs-v2/integrate/guides/authorize-an-api.mdx
@@ -83,7 +83,11 @@ For OAuth, the `nango.auth()` method will trigger the OAuth flow in a popup, to 
 
 ```js
 nango
-    .auth('<INTEGRATION-ID>', '<CONNECTION-ID>')
+    .auth('<INTEGRATION-ID>', '<CONNECTION-ID>', {
+        display_name: '<OPTIONAL-DISPLAY-NAME>',
+        customer_email: '<OPTIONAL-DISPLAY-NAME>'
+        customer_domain: '<OPTIONAL-CUSTOMER-DOMAIN>',
+    })
     .then((result) => {
         // Show success UI.
     })
--- a/docs-v2/reference/sdks/frontend.mdx
+++ b/docs-v2/reference/sdks/frontend.mdx
@@ -24,7 +24,7 @@ let nango = new Nango({ publicKey: '<PUBLIC-KEY>' });
       </ResponseField>
 
       <ResponseField name="host" type="string">
-        Omitting the host points to Nango Cloud. For local development, use `http://localhost:3003`. Use your instance URL if self-hosting. 
+        Omitting the host points to Nango Cloud. For local development, use `http://localhost:3003`. Use your instance URL if self-hosting.
       </ResponseField>
 
       <ResponseField name="websocketsPath" type="string">
@@ -106,6 +106,15 @@ const result = nango.auth('<INTEGRATION-ID>', '<CONNECTION-ID>', {
 
   <ResponseField name="options" type="object">
     <Expandable title="options" defaultOpen>
+      <ResponseField name="display_name" type="string">
+            A human-readable display name for the connection.
+      </ResponseField>
+      <ResponseField name="customer_email" type="string">
+            The email address of the end-user. This field is just used for ease of reference and does not alter the behaviour of the connection.
+      </ResponseField>
+      <ResponseField name="customer_domain" type="string">
+            The end-user's domain. This field is just used for ease of reference and does not alter the behaviour of the connection.
+      </ResponseField>
       <ResponseField name="params" type="object">
         Specify additional [connection configuration](/integrate/guides/authorize-an-api#apis-requiring-connection-specific-configuration-for-authorization) necessary to perform the authorization request.
       </ResponseField>
@@ -119,11 +128,11 @@ const result = nango.auth('<INTEGRATION-ID>', '<CONNECTION-ID>', {
       </ResponseField>
 
       <ResponseField name="authorization_params" type="object">
-        For OAuth, specify the query parameters of the authorization URL. 
+        For OAuth, specify the query parameters of the authorization URL.
       </ResponseField>
 
       <ResponseField name="user_scope" type="string[]">
-        For Slack OAuth, specify user-specific scopes. 
+        For Slack OAuth, specify user-specific scopes.
       </ResponseField>
 
       <ResponseField name="credentials" type="object">
--- a/packages/frontend/lib/index.ts
+++ b/packages/frontend/lib/index.ts
@@ -420,18 +420,6 @@ export default class Nango {
                 query.push(`user_scope=${encodeURIComponent(connectionConfig.user_scope.join(','))}`);
             }
 
-            if (connectionConfig.display_name) {
-                query.push(`params[display_name]=${encodeURIComponent(connectionConfig.display_name)}`);
-            }
-
-            if (connectionConfig.customer_email) {
-                query.push(`params[customer_email]=${encodeURIComponent(connectionConfig.customer_email)}`);
-            }
-
-            if (connectionConfig.customer_domain) {
-                query.push(`params[customer_domain]=${encodeURIComponent(connectionConfig.customer_domain)}`);
-            }
-
             if (connectionConfig.credentials) {
                 const credentials = connectionConfig.credentials;
                 if ('oauth_client_id_override' in credentials) {
@@ -470,9 +458,6 @@ interface ConnectionConfig {
     user_scope?: string[];
     authorization_params?: Record<string, string | undefined>;
     credentials?: OAuthCredentialsOverride | BasicApiCredentials | ApiKeyCredentials | AppStoreCredentials | TBACredentials;
-    display_name?: string;
-    customer_email?: string;
-    customer_domain?: string;
 }
 
 interface OAuthCredentialsOverride {
--- a/packages/frontend/lib/index.ts
+++ b/packages/frontend/lib/index.ts
@@ -413,11 +413,23 @@ export default class Nango {
             }
 
             if (connectionConfig.hmac) {
-                query.push(`hmac=${connectionConfig.hmac}`);
+                query.push(`hmac=${encodeURIComponent(connectionConfig.hmac)}`);
             }
 
             if (connectionConfig.user_scope) {
-                query.push(`user_scope=${connectionConfig.user_scope.join(',')}`);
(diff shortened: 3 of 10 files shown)


packages/webapp/src/pages/Connection/List.tsx
```diff
@@ -20,59 +20,55 @@ import { useStore } from '../../store';
...
-    const [selectedIntegration, setSelectedIntegration] = useState<string[]>(defaultFilter);
+    const connections: Connection[] = data?.connections || [];
+
+    const [selectedIntegrations, setSelectedIntegrations] = useState<string[]>(defaultFilter);
     const [connectionSearch, setConnectionSearch] = useState<string>('');
-    const [states, setStates] = useState<string[]>(defaultFilter);
+    const [selectedStatuses, setSelectedStatuses] = useState<string[]>(defaultFilter);
+    const [selectedCustomerDomains, setSelectedCustomerDomains] = useState<string[]>(defaultFilter);
 
-    useEffect(() => {
```
Daspy11, line 35: This useEffect hook was entirely duplicative of the second one from what I could tell

packages/webapp/src/pages/Connection/List.tsx
```diff
@@ -20,59 +20,55 @@ import { useStore } from '../../store';
...
 
+        return (data.connections || []).filter((connection) => connection.active_logs).length;
+    }, [data]);
+
+    // eslint-disable-next-line react-hooks/exhaustive-deps
     const debouncedSearch = useCallback(
         debounce((value: string) => {
-            if (!value.trim()) {
-                setConnectionSearch('');
-                setFilteredConnections(data?.connections || []);
```
Daspy11, line 72: This line was causing a bug where if you searched with an empty/only whitespaced search box, the list of connections displayed would ignore your filters.

packages/webapp/src/pages/Connection/List.tsx
```diff
@@ -175,66 +179,112 @@ export default function ConnectionList() {
...
-                                    >
-                                        <div className="flex items-center w-2/3 gap-2 py-2 truncate">
-                                            <span className="break-words break-all truncate">{connectionId}</span>
-                                            {active_logs && <ErrorCircle />}
-                                            <CopyButton dark text={connectionId} />
-                                        </div>
-                                        <div className="flex items-center w-1/3 gap-3">
-                                            <div className="w-7">
-                                                <IntegrationLogo provider={provider} height={7} width={7} />
+                            <div className="rounded-md overflow-hidden">
```
Daspy11, line 238: Makes the row hover background rounded to match the header

docs-v2/integrate/guides/authorize-an-api.mdx
```diff
@@ -83,7 +83,11 @@ For OAuth, the `nango.auth()` method will trigger the OAuth flow in a popup, to
 
 ```js
 nango
-    .auth('<INTEGRATION-ID>', '<CONNECTION-ID>')
+    .auth('<INTEGRATION-ID>', '<CONNECTION-ID>', {
+        display_name: '<OPTIONAL-DISPLAY-NAME>',
+        customer_email: '<OPTIONAL-DISPLAY-NAME>'
```
bodinsamuel, line 88:
```suggestion
        customer_email: '<OPTIONAL-CUSTOMER-EMAIL>'
```

docs-v2/reference/sdks/frontend.mdx
```diff
@@ -106,6 +106,15 @@ const result = nango.auth('<INTEGRATION-ID>', '<CONNECTION-ID>', {
...
   <ResponseField name="options" type="object">
     <Expandable title="options" defaultOpen>
+      <ResponseField name="display_name" type="string">
+            A human-readable display name for the connection.
+      </ResponseField>
+      <ResponseField name="customer_email" type="string">
+            The email address of the end-user. This field is just used for ease of reference and does not alter the behaviour of the connection.
+      </ResponseField>
+      <ResponseField name="customer_domain" type="string">
+            The end-user's domain. This field is just used for ease of reference and does not alter the behaviour of the connection.
```
bodinsamuel, line 116: I'm not sure I understand what domain means without more context

packages/frontend/lib/index.ts
```diff
@@ -413,11 +413,11 @@ export default class Nango {
             }
 
             if (connectionConfig.hmac) {
-                query.push(`hmac=${connectionConfig.hmac}`);
+                query.push(`hmac=${encodeURIComponent(connectionConfig.hmac)}`);
```
bodinsamuel, line 416: can the HMAC contains exotic char?

packages/shared/lib/services/connection.service.ts
```diff
@@ -579,7 +581,17 @@ class ConnectionService {
...
+        {
+            id: number;
+            connection_id: string;
+            provider: string;
+            created: string;
+            metadata: Metadata;
+            active_logs: ActiveLogIds;
+            connection_config: ConnectionConfig;
+        }[]
+    > {
```
bodinsamuel, line 594: Should probably be `ConnectionList[]` no? (the type name is confusing but can be changed later)

edit: nevermind the type is only in the frontend 😞

packages/webapp/src/components/MultiSelect.tsx
```diff
@@ -12,9 +12,10 @@ export interface MultiSelectArgs<T> {
     defaultSelect: T[];
     all?: boolean;
     onChange: (selected: T[]) => void;
+    emptyLabel?: string;
 }
 
-export const MultiSelect: React.FC<MultiSelectArgs<any>> = ({ label, options, selected, defaultSelect, all, onChange }) => {
+export const MultiSelect: React.FC<MultiSelectArgs<any>> = ({ label, options, selected, defaultSelect, all, onChange, emptyLabel = 'No framework found.' }) => {
```
bodinsamuel, line 18: thanks 👌🏻 I would go even further and remove the default (I made the component and it was later moved to be reusable but this details slipped)

packages/webapp/src/pages/Connection/List.tsx
```diff
@@ -20,59 +20,55 @@ import { useStore } from '../../store';
 
 const defaultFilter = ['all'];
 
+function truncateWithEllipsis(text: string) {
+    if (text.length <= 6) {
+        return text;
+    }
+    return text.slice(0, 3) + '...' + text.slice(-3);
+}
```
bodinsamuel, line 28: I could reuse that actually, do you mind putting that in utils (eventually in a new file utils/string.ts to avoid cluttering the main file)

packages/webapp/src/pages/Connection/List.tsx
```diff
@@ -175,66 +179,112 @@ export default function ConnectionList() {
...
+                                                </span>
+                                            </div>
+                                            <div className="flex items-center w-1/4 gap-3">
+                                                <div className="w-7">
+                                                    <IntegrationLogo provider={provider} height={7} width={7} />
+                                                </div>
+                                                <p className="break-words break-all">{providerConfigKey}</p>
+                                            </div>
+                                            <div className="flex w-1/4 items-center">
+                                                <span className="break-words break-all truncate text-gray-400 mr-2">{truncateWithEllipsis(connectionId)}</span>
```
bodinsamuel, line 276: Can you add a tooltip? (using Tooltip.tsx not geist)
also there is a problem with arbitrary truncate, you truncate even if it could fit in the column width. You can try to make it more flexible, or use `truncate` in tailwind it's less good but more automated

packages/webapp/src/pages/Connection/List.tsx
```diff
@@ -175,66 +179,112 @@ export default function ConnectionList() {
...
+                                            className={`flex gap-4 ${
+                                                id !== connections.at(-1)?.id ? 'border-b border-border-gray' : ''
+                                            } min-h-[4em] px-2 justify-between items-center hover:bg-hover-gray cursor-pointer`}
+                                            onClick={() => {
+                                                navigate(`/${env}/connections/${encodeURIComponent(providerConfigKey)}/${encodeURIComponent(connectionId)}`);
+                                            }}
+                                        >
+                                            <div className="flex items-center w-1/4 gap-2 py-2 truncate">
+                                                <span className="break-words break-all truncate">
+                                                    {connection_config?.display_name ? (connection_config.display_name as string | undefined) : '-'}
```
bodinsamuel, line 260: we probably need to fallback to connectionId because right now it's not usable for legacy connection 
Screenshot 2024-07-08 at 11 46 55

packages/webapp/src/pages/Connection/List.tsx
```diff
@@ -175,66 +179,112 @@ export default function ConnectionList() {
...
+                                    options={connections.reduce((acc: { name: string; value: string }[], connection: Connection) => {
+                                        const sitename = connection?.connection_config ? connection.connection_config['customer_domain'] : undefined;
+                                        if (typeof sitename === 'string') {
+                                            acc.push({
+                                                name: sitename,
+                                                value: sitename
+                                            });
+                                        }
+                                        return acc;
+                                    }, [])}
```
bodinsamuel, line 206: Can you put it in a memo? no reason to recompile that at every render

bodinsamuel (commented):
Thanks @Daspy11 looks good overall 💪🏻 🚀 
minor comments here and there, mostly around types and details

packages/webapp/src/pages/Connection/Authorization.tsx
```diff
@@ -46,6 +49,25 @@ export default function Authorization(props: AuthorizationProps) {
                 )}
             </div>
             <div className="flex">
+                <div className="flex flex-col w-1/2">
```
khaliqgant, line 52: Do we want to show these fields here if not present? Most connections won't have it and it takes up a lot of space with no values.

packages/webapp/src/pages/Connection/List.tsx
khaliqgant: Feel like if the user isn't using these fields then we shouldn't show these empty columns at all. To make it more discoverable it would be nice to add in optional fields in the Connection/Create screen in addition to being on the edit connection screen as added in #2478

bodinsamuel:
Closed in favor of recent changes 👍🏻

PR #2478, opened 2024-07-08
feat(connections): Allow editing of connection display name, customer email, and customer domain via detail view on web app

Daspy11:
## Describe your changes
Adds functionality to edit connections to the web app. Extends off of [this PR](https://github.com/NangoHQ/nango/pull/2477).

image

This is a tentative draft PR so we can discuss the approach in more depth. We don't have a dedicated endpoint to update a connectionConfig, so I've jammed the parameters into the update metadata endpoint which equally doesn't feel right. The obvious approach would be to add a dedicated updateConnectionConfig endpoint, but doing that without causing footguns means we'd need to keep a separate list of potential connectionConfig parameters and whether or not they can be edited.

All this is making me lean towards the idea that perhaps these fields are neither metadata nor connection configuration, and should instead be considered _frontmatter_ - human intelligible metadata to be used in interfaces. 

N/A

### DIFF
--- a/packages/server/lib/controllers/connection/updateMetadata.ts
+++ b/packages/server/lib/controllers/connection/updateMetadata.ts
@@ -1,25 +1,24 @@
 import { z } from 'zod';
 import { asyncWrapper } from '../../utils/asyncWrapper.js';
-import { requireEmptyQuery, zodErrorToHTTP } from '@nangohq/utils';
-import type { ApiError, UpdateMetadata, MetadataBody } from '@nangohq/types';
+import { zodErrorToHTTP } from '@nangohq/utils';
+import type { ApiError, UpdateMetadata, MetadataBody, ConnectionConfig } from '@nangohq/types';
 import { connectionService } from '@nangohq/shared';
 import type { Connection } from '@nangohq/shared';
+import db from '@nangohq/database';
 
 const validation = z
     .object({
         connection_id: z.union([z.string().min(1), z.array(z.string().min(1))]),
         provider_config_key: z.string().min(1),
-        metadata: z.record(z.unknown())
+        metadata: z.record(z.unknown()).optional(),
+        display_name: z.string().optional(),
+        customer_domain: z.string().optional(),
+        customer_email: z.string().optional(),
+        env: z.record(z.unknown()).optional()
     })
     .strict();
 
 export const updateMetadata = asyncWrapper<UpdateMetadata>(async (req, res) => {
-    const emptyQuery = requireEmptyQuery(req);
-    if (emptyQuery) {
-        res.status(400).send({ error: { code: 'invalid_query_params', errors: zodErrorToHTTP(emptyQuery.error) } });
-        return;
-    }
-
     const val = validation.safeParse(req.body);
     if (!val.success) {
         res.status(400).send({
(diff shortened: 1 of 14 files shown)


packages/server/lib/controllers/connection/updateMetadata.ts
```diff
@@ -1,25 +1,24 @@
...
 const validation = z
     .object({
         connection_id: z.union([z.string().min(1), z.array(z.string().min(1))]),
         provider_config_key: z.string().min(1),
-        metadata: z.record(z.unknown())
+        metadata: z.record(z.unknown()).optional(),
+        display_name: z.string().optional(),
+        customer_domain: z.string().optional(),
+        customer_email: z.string().optional(),
+        env: z.record(z.unknown()).optional()
```
Daspy11, line 17: Is this the right place to put this?

packages/shared/lib/services/connection.service.ts
```diff
@@ -522,26 +524,42 @@ class ConnectionService {
...
+        if (trx) {
+            await query.transacting(trx);
+        } else {
+            await query;
+        }
     }
 
-    public async updateMetadata(connections: Connection[], metadata: Metadata): Promise<void> {
-        await db.knex.transaction(async (trx) => {
+    public async updateMetadata(connections: Connection[], metadata: Metadata, trx?: Knex.Transaction): Promise<void> {
```
Daspy11, line 540: The functionality here stays the same, but needed to add an optional transaction parameter so this can be batched with updating the connection config.

packages/server/lib/routes.ts
```diff
@@ -199,6 +199,7 @@ web.route('/api/v1/integration/:providerConfigKey/connections').get(webAuth, con
 web.route('/api/v1/provider').get(configController.listProvidersFromYaml.bind(configController));
 
 web.route('/api/v1/connection').get(webAuth, connectionController.listConnections.bind(connectionController));
+web.route('/api/v1/connection/metadata').patch(webAuth, updateMetadata);
```
bodinsamuel, line 202: I definitely think you can create a dedicated endpoint, i.e: `PUT api/v1/connection/config`

packages/shared/lib/services/connection.service.ts
```diff
@@ -522,26 +524,42 @@ class ConnectionService {
         await trx.from<StoredConnection>(`_nango_connections`).whereIn('id', ids).andWhere({ deleted: false }).update({ metadata });
     }
 
-    public async replaceConnectionConfig(connection: Connection, config: ConnectionConfig) {
-        await db.knex
+    public async replaceConnectionConfig(connection: Connection, config: ConnectionConfig, trx?: Knex.Transaction) {
+        const query = db.knex
```
bodinsamuel, line 528: usually you can swap the object; it's annoying that knex is not exactly the same type as trx but it works almost transparently 
```ts
    public async replaceConnectionConfig(connection: Connection, config: ConnectionConfig, trx?: Knex.Transaction) {
        const query = (trx || db.knex)
```

packages/webapp/src/pages/Connection/Authorization.tsx
```diff
@@ -45,7 +81,24 @@ export default function Authorization(props: AuthorizationProps) {
...
+                    <div className="flex items-center gap-2">
+                        <span className="text-white break-all">{connection.connection_id}</span>
+                        <CopyButton text={connection.connection_id} dark />
+                    </div>
+                </div>
+            </div>
+            <div className="flex space-x-8">
+                <div className="flex flex-col w-1/2">
+                    <span className="text-gray-400 text-xs uppercase mb-2">Customer Email</span>
+                    <EditableField value={customerEmail} onChange={setCustomerEmail} onSave={handleSave} dark />
```
bodinsamuel, line 100: should enforce email type

bodinsamuel (commented):
Took a quick look, works fine 👌🏻

packages/webapp/src/components/ui/input/EditableField.tsx
```diff
@@ -0,0 +1,56 @@
...
+        <div className={`flex items-center gap-2 ${className}`}>
+            {isEditing ? (
+                <input
+                    type="text"
+                    value={editValue}
+                    onChange={handleChange}
+                    onBlur={handleBlur}
+                    onKeyDown={(e) => e.key === 'Enter' && handleBlur()}
+                    autoFocus
+                    className="border-border-gray bg-active-gray text-text-light-gray focus:border-white focus:ring-white block w-full appearance-none rounded-md border px-3 py-1 text-sm placeholder-gray-400 shadow-sm focus:outline-none"
```
khaliqgant, line 42: Seems to be missing a save / confirmation button. For instance when changing the frequency, you can hit enter or also click the area seen here:

image

bodinsamuel:
Closed in favor of recent changes

PR #2483, opened 2024-07-09
feat(account): check password strength

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1028/password-strength-check

### Major changes
- UI: Check password strength
There was no design associated with this feature so improvised something, not sure it's good tbh 😅 
 
- API: dedicated file for forgot password and reset password

- UI: use Input and Button where needed

### Minor changes

- Fix bug in reset password where token was not set back to null after successful attempt
- Add loader to some button in auth pages
- Add more Shadcn/Radix component
- Add input size
- Add e2e test for signup

--- 

Screenshot 2024-07-09 at 15 20 21


packages/server/lib/controllers/v1/account/putResetPassword.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    try {
+        jwt.verify(token, resetPasswordSecret());
+
+        const hashedPassword = (await pbkdf2(password, user.salt, 310000, 32, 'sha256')).toString('base64');
+
+        user.hashed_password = hashedPassword;
+        user.reset_password_token = null;
+        console.log('on edit', user);
```
TBonnin: 🎮 🪵

packages/server/lib/controllers/v1/account/signin.ts
```diff
@@ -7,7 +7,7 @@ import type { WebUser, Signin } from '@nangohq/types';
 const validation = z
     .object({
         email: z.string().email(),
-        password: z.string().min(8)
+        password: z.string().min(8).max(64)
```
TBonnin, line 10: why a upper limit?

packages/server/lib/controllers/v1/account/signup.ts
```diff
@@ -9,10 +9,17 @@ import type { Signup } from '@nangohq/types';
 
 const logger = getLogger('Server.Signup');
 
+export const passwordSchema = z
+    .string()
+    .min(8)
+    .max(64)
+    .refine((value) => {
+        return value.match(/[a-z]+/) && value.match(/[A-Z]+/) && value.match(/[0-9]/) && value.match(/[^a-zA-Z0-9]/);
+    }, 'Password should be least 8 characters with lowercase, uppercase, a number and a special character');
```
TBonnin, line 18: not a huge fan of restricting what users can do or not do with their password. In general length is more important than complexity. Not a big deal but I am curious to know why we decided to go with those requirements.

Also `/[^a-zA-Z0-9]/` includes all non-alphanumeric character, including whitespaces, etc... I wonder if this is what you wanted

packages/webapp/src/pages/Account/components/Password.tsx
```diff
@@ -0,0 +1,66 @@
...
+        }
+        if (local.match(/[A-Z]/)) {
+            strength += 25;
+        }
+        if (local.match(/[0-9]/)) {
+            strength += 25;
+        }
+        if (local.match(/[^a-zA-Z0-9]/)) {
+            strength += 25;
+        }
```
TBonnin: I found this confusing when I tested it. 
Even though the bar is supposed to represent the password strength, it doesn't behave like one. For instance a 8 lowercase characters will have the same bar as a password of 32 chars. That doesn't make sense.
I think we either keep the bar but make sure it represents the real strength of the password or we have a check list of requirements that we automatically cross

TBonnin (commented):
The password strength bar is confusing I think but the rest looks good.

packages/webapp/src/pages/Account/components/Password.tsx
```diff
@@ -0,0 +1,66 @@
...
+    setPassword,
+    ...props
+}) => {
+    const [local, setLocal] = useState('');
+    const { strength, color } = useMemo(() => {
+        let strength = 0;
+        if (local.length >= 1) {
+            strength += 12.5;
+        }
+        if (local.length > 8) {
```
TBonnin: `at least 8` should be `>= 8`

packages/server/lib/controllers/v1/account/signup.ts
bodinsamuel: yep that's what I wanted. Not a fan either but we have a minimum of 8 right now and can't really break this (unless we put more effort into a slow migration process), plus certification often requires those stuff (i.e: SOC2 has a checklist on password if I recall correctly)

packages/server/lib/controllers/v1/account/signin.ts
bodinsamuel: DDoS mitigation

packages/webapp/src/pages/Account/components/Password.tsx
bodinsamuel: There is not a lot of diff between 8 not random lowercase char and 32 not random lowercase char, since my check is rather naive it's a indication not a scientific check.
Also there is a big diff between 8 random char with numbers and special chars, vs 32 lowercase char (and even bigger vs 32 repeated char for example). 
cf: see here 58minutes for "a"{8} and 4hours for "a"{32}, not really safer while `aAè8-kjn` takes centuries
https://lowe.github.io/tryzxcvbn/


Strength is an indication not something sure, but happy to remove the bar and just show the checks on the right

packages/server/lib/routes.ts
```diff
@@ -152,8 +154,8 @@ if (AUTH_ENABLED) {
     web.route('/api/v1/account/signup/invite').get(rateLimiterMiddleware, authController.invitation.bind(authController));
     web.route('/api/v1/account/logout').post(rateLimiterMiddleware, authController.logout.bind(authController));
     web.route('/api/v1/account/signin').post(rateLimiterMiddleware, passport.authenticate('local'), signin);
-    web.route('/api/v1/account/forgot-password').put(rateLimiterMiddleware, authController.forgotPassword.bind(authController));
-    web.route('/api/v1/account/reset-password').put(rateLimiterMiddleware, authController.resetPassword.bind(authController));
+    web.route('/api/v1/account/forgot-password').post(rateLimiterMiddleware, postForgotPassword);
+    web.route('/api/v1/account/reset-password').put(rateLimiterMiddleware, putResetPassword);
```
khaliqgant, line 158: why the HTTP method at the beginning of the method? It reads a bit oddly and we don't do this currently anywhere else.

khaliqgant (approved):
Question about method name -- non blocking.

packages/server/lib/routes.ts
bodinsamuel: we do it (or at least some), in orchestrator it's all like that and I used this pattern at the end of logs dev (so not all are named like this). It's handy to communicate expectation, also allows to dedup (i.e: putRecurring and postRecurring)

bodinsamuel:
Changed to a list of requirements following Thomas suggestion, happy to iterate later if we feel like it's not enough 👌🏻 
Screenshot 2024-07-10 at 11 16 12

packages/server/lib/routes.ts
khaliqgant: Got it, thanks for clarifying.

PR #2484, opened 2024-07-09
fix(auth): do not display refresh button if not possible

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1389/fix-ui-do-not-refresh-if-not-possible

- Do not display refresh token button if not possible

### DIFF
--- a/packages/webapp/src/pages/Connection/Authorization.tsx
+++ b/packages/webapp/src/pages/Connection/Authorization.tsx
@@ -137,7 +137,12 @@ export default function Authorization(props: AuthorizationProps) {
             {(connection.credentials.type === 'OAUTH2' || connection.credentials.type === 'APP') && connection.credentials.access_token && (
                 <div className="flex flex-col">
                     <span className="text-gray-400 text-xs uppercase mb-1">Access Token</span>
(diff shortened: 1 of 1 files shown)


packages/webapp/src/pages/Connection/Authorization.tsx
```diff
@@ -137,7 +137,12 @@ export default function Authorization(props: AuthorizationProps) {
             {(connection.credentials.type === 'OAUTH2' || connection.credentials.type === 'APP') && connection.credentials.access_token && (
                 <div className="flex flex-col">
                     <span className="text-gray-400 text-xs uppercase mb-1">Access Token</span>
-                    <SecretInput disabled value={refreshing ? 'Refreshing...' : connection.credentials.access_token} copy={true} refresh={handleForceRefresh} />
+                    <SecretInput
+                        disabled
+                        value={refreshing ? 'Refreshing...' : connection.credentials.access_token}
+                        copy={true}
```
khaliqgant, line 143: Is it true that no other types besides oauth2 have a refresh token? I don’t recall exactly

bodinsamuel: this input is only displayed for Oauth2 and App, and it doesn't seems to contain refresh_token

khaliqgant (approved):

PR #2485, opened 2024-07-09
chore(scheduler): cleanup tasks and schedules

TBonnin:
Hard deleting deleted schedules and terminated tasks older than 30 days in order to prevent those tables to grow indefinitely.
I picked 30 days arbitrarily but it gives us plenty of times for debugging if we need.

Cleaning up is following the same design as scheduling and monitoring. It is a worker_thread with an infinite loop deleting tasks and schedules

https://linear.app/nango/issue/NAN-1223/delete-tasks-after-30-days-and-deleted-schedules-after-90-days

tested locally

### DIFF
--- a/packages/scheduler/lib/models/schedules.ts
+++ b/packages/scheduler/lib/models/schedules.ts
@@ -232,3 +232,26 @@ export async function search(
         return Err(new Error(`Error searching schedules: ${stringifyError(err)}`));
     }
 }
+
+export async function hardDeleteOlderThanNDays(db: knex.Knex, days: number): Promise<Result<DbSchedule[]>> {
+    try {
+        const deleted = await db
+            .from<DbSchedule>(SCHEDULES_TABLE)
+            .where(
+                'id',
+                '=',
+                db.raw(`ANY(ARRAY(
+                    SELECT "id"
+                    FROM ${SCHEDULES_TABLE}
+                    WHERE "deleted_at" < NOW() - INTERVAL '${days} days'
+                    ORDER BY "id" ASC
+                    LIMIT 1000
+                  ))`)
+            )
+            .del()
+            .returning('*');
+        return Ok(deleted);
+    } catch (err: unknown) {
+        return Err(new Error(`Error hard deleting schedules older than ${days} days: ${stringifyError(err)}`));
+    }
+}
--- a/packages/scheduler/lib/models/schedules.ts
+++ b/packages/scheduler/lib/models/schedules.ts
@@ -235,6 +235,7 @@ export async function search(
 
 export async function hardDeleteOlderThanNDays(db: knex.Knex, days: number): Promise<Result<DbSchedule[]>> {
     try {
+        // NOTE: only deleting one schedule at a time to avoid massive cascading deletes of tasks
         const deleted = await db
             .from<DbSchedule>(SCHEDULES_TABLE)
             .where(
@@ -245,7 +246,7 @@ export async function hardDeleteOlderThanNDays(db: knex.Knex, days: number): Pro
                     FROM ${SCHEDULES_TABLE}
                     WHERE "deleted_at" < NOW() - INTERVAL '${days} days'
                     ORDER BY "id" ASC
-                    LIMIT 1000
+                    LIMIT 1
                   ))`)
             )
             .del()
--- a/packages/scheduler/lib/models/tasks.ts
+++ b/packages/scheduler/lib/models/tasks.ts
@@ -5,6 +5,7 @@ import { Ok, Err, stringifyError } from '@nangohq/utils';
 import { taskStates } from '../types.js';
 import type { TaskState, Task, TaskTerminalState, TaskNonTerminalState } from '../types.js';
 import { uuidv7 } from 'uuidv7';
+import { SCHEDULES_TABLE } from './schedules.js';
 
 export const TASKS_TABLE = 'tasks';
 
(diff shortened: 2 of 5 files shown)


packages/scheduler/lib/models/schedules.ts
```diff
@@ -232,3 +232,26 @@ export async function search(
...
 }
+
+export async function hardDeleteOlderThanNDays(db: knex.Knex, days: number): Promise<Result<DbSchedule[]>> {
+    try {
+        const deleted = await db
+            .from<DbSchedule>(SCHEDULES_TABLE)
+            .where(
+                'id',
+                '=',
+                db.raw(`ANY(ARRAY(
```
TBonnin, line 244: using `any(array())` instead of `in` because the query plan says it is faster

packages/scheduler/lib/models/tasks.ts
```diff
@@ -305,3 +306,36 @@ export async function expiresIfTimeout(db: knex.Knex): Promise<Result<Task[]>> {
...
+    try {
+        // Delete terminated tasks where lastStateTransitionAt is older than N days
+        // unless it is the most recent task for an given schedule
+        const tasks = await db
+            .from<DbTask>(TASKS_TABLE)
+            .where(
+                'id',
+                '=',
+                // schedules.id is null if the task is not associated with a schedule (ie: actions)
+                // or if the task is not the last task for its associated schedule
```
TBonnin, line 320: subquery below is a bit more complicated than the schedules one because we need to keep the latest tasks for a given schedule even if it is older than 30 days. This is achieved by joining with `schedules.last_scheduled_task_id`. Any row with an empty left join is either a task without schedule (ie: actions) or a task that is not the latest one for any schedule

packages/scheduler/lib/models/tasks.ts
```diff
@@ -305,3 +306,36 @@ export async function expiresIfTimeout(db: knex.Knex): Promise<Result<Task[]>> {
...
+                db.raw(`ANY(ARRAY(
+                    SELECT ${TASKS_TABLE}.id
+                    FROM ${TASKS_TABLE}
+                    LEFT JOIN ${SCHEDULES_TABLE} ON ${SCHEDULES_TABLE}.last_scheduled_task_id = ${TASKS_TABLE}.id
+                    WHERE ${TASKS_TABLE}.terminated = true
+                    AND ${TASKS_TABLE}.starts_after < NOW() - INTERVAL '${days} days'
+                    AND ${SCHEDULES_TABLE}.id IS NULL
+                    ORDER BY ${TASKS_TABLE}.id ASC
+                    LIMIT 1000
```
TBonnin, line 329: 1000 is arbitrary. We peak at 600+ new tasks per minutes today for instance. 
This query will running every 10s and delete 1000 entries which means deleting ~6000 entries a minutes so we have some margin

packages/scheduler/lib/models/tasks.ts
```diff
@@ -305,3 +306,36 @@ export async function expiresIfTimeout(db: knex.Knex): Promise<Result<Task[]>> {
...
+                'id',
+                '=',
+                // schedules.id is null if the task is not associated with a schedule (ie: actions)
+                // or if the task is not the last task for its associated schedule
+                db.raw(`ANY(ARRAY(
+                    SELECT ${TASKS_TABLE}.id
+                    FROM ${TASKS_TABLE}
+                    LEFT JOIN ${SCHEDULES_TABLE} ON ${SCHEDULES_TABLE}.last_scheduled_task_id = ${TASKS_TABLE}.id
+                    WHERE ${TASKS_TABLE}.terminated = true
+                    AND ${TASKS_TABLE}.starts_after < NOW() - INTERVAL '${days} days'
```
TBonnin, line 326: using `starts_after` because there is already an existing index for it. Didn't want to add another index and since `terminated = true` it doesn't matter which date we are using

packages/scheduler/lib/models/tasks.ts
bodinsamuel: Do we really need to keep the latest tho? We use the latest to be able to dedup tasks but ultimately if it's 60 days old it's not running for sure

packages/scheduler/lib/workers/cleanup/cleanup.worker.ts
```diff
@@ -0,0 +1,112 @@
+import * as fs from 'fs';
```
bodinsamuel, line 1: Don't know enough about worker but I feel like it's not going to help.
It creates a whole new nodejs instance just to run a setInterval, locally I can definitely the dev is slower than before I just to be mindful about memory and cpu consumption

packages/scheduler/lib/models/schedules.ts
```diff
@@ -232,3 +232,26 @@ export async function search(
...
+                    SELECT "id"
+                    FROM ${SCHEDULES_TABLE}
+                    WHERE "deleted_at" < NOW() - INTERVAL '${days} days'
+                    ORDER BY "id" ASC
+                    LIMIT 1000
```
bodinsamuel: This logic will definitely be too "simple" because of foreign key. If a schedule executes every 5minutes, it has ~20K tasks in the last 60days, if you have 1000 schedules that's 20M rows. 
Even deleting 20K can be too slow for PG depending on the lock patterns and db size. I would at least lower the limit drastically

bodinsamuel (commented):
Looks good overall but I have some comments regarding memory and deletion

packages/scheduler/lib/models/tasks.ts
TBonnin: In theory syncs can have a frequency greater than 30 days, which would break scheduling/reporting if we delete the last one. 
In practice, current max frequency is actually 30 days but I prefer to cover this case so we don't run into issues if customers set frequency > 30days or if we decide in the future to lower this 30 days arbitrary threshold

packages/scheduler/lib/models/schedules.ts
TBonnin: discussed online but writing here for visibility:
Cascading deletion of tasks could indeed be problematic if we are deleting a lot of schedules at the same time. 
In practice schedules are marked as deleted very infrequently so very few would be hard deleted at the same time. However to minimize any potential issue we said we'll set the limit to `1`. For a schedule with frequency 5mins that less than 9k task rows

packages/scheduler/lib/workers/cleanup/cleanup.worker.ts
TBonnin: Also discussed online. It is true that worker_threads add overheard but in prod we are very far from maxing out the orchestrator instance resources even though we are running on very small vms. It might also have an impact for local setup but I will prefer to go for this solution and only revisit if it becomes unbearably slow for local development

packages/scheduler/lib/models/schedules.ts
TBonnin: limit lowered to 1 https://github.com/NangoHQ/nango/pull/2485/commits/ae0a6c78668e2092933275dd09162ab887d81a75 @bodinsamuel

packages/scheduler/lib/workers/cleanup/cleanup.worker.ts
khaliqgant: Got it. Would be sure to test and verify this on staging to make sure no spikes occur with a lot of schedules. Happy to help generate more staging schedules if needed.

khaliqgant (approved):
See note about testing.

packages/scheduler/lib/workers/cleanup/cleanup.worker.ts
TBonnin: as mentioned below, we are gonna delete one schedule at a time to minimize as much as we can the impact of schedule deletions

bodinsamuel (approved):

PR #2486, opened 2024-07-09
fix(types): correct Timestamps, move env and account

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1395/types-fix-timestamps-its-never-undefined-or-null
Fixes https://linear.app/nango/issue/NAN-1396/types-rename-account-to-team
Fixes https://linear.app/nango/issue/NAN-1397/types-rename-env-to-dbenv

I started by just moving `/api/v1/account` to `/api/v1/team`, to realise some types were wrong or mixed, Timestamps was also incorrect and since there was a desync+copy/paste across the codebase I had to fix a lot of stuff plus remove the legacy in shared.
In the current Team v2 project, we are more and more using the word Team instead of Account, and we are also confusing User and Account in the codebase so I took the opportunity to start renaming the type.

- Types: Timestamps can't be undefined or null, it's always Date
- Types: rename Environment to DBEnvironment
- Types: rename EnvironmentVariables to DBEnvironmentVariables
- Types: rename Account to DBTeam 
I have just renamed the type, I will probably rename the property `account` everywhere on a second PR.

- API: Dedicated file for endpoint `/api/v1/team`

NB: some types are still wrong, in the node-client and sdk because we are using Connection instead of a dedicated type that would not contains some `Date` (that are actually strings)


packages/types/lib/account/db.ts
```diff
@@ -1,10 +0,0 @@
-export interface Account {
-    id: number;
-    name: string;
-    secret_key: string;
-    host?: string | null;
-    websockets_path?: string;
-    uuid: string;
-    is_admin?: boolean;
-    is_capped?: boolean;
-}
```
bodinsamuel, line 10: The type was the most incorrect, non-existing fields and defined inside types that are user related

packages/shared/lib/utils/encryption.manager.ts
```diff
@@ -82,12 +81,12 @@ export class EncryptionManager extends Encryption {
         return storedConnection;
     }
 
-    public encryptConnection(connection: Connection): StoredConnection {
+    public encryptConnection(connection: Omit<Connection, 'created_at' | 'updated_at'>): Omit<StoredConnection, 'created_at' | 'updated_at'> {
```
bodinsamuel, line 84: this is bit sad, we are not passing complete Connection so I had to deal the way I can

packages/shared/lib/services/environment.service.ts
```diff
@@ -171,11 +177,15 @@ class EnvironmentService {
             // Store only successful attempt to not pollute the memory
             hashLocalCache.set(opts.secretKey, hash);
         }
-        return { account: res.account, environment: encryptionManager.decryptEnvironment(res.environment)! };
+        return {
+            // Getting things with row_to_json breaks to automatic string -> date parser
+            account: { ...res.account, created_at: new Date(res.account.created_at), updated_at: new Date(res.account.updated_at) },
```
bodinsamuel, line 182: just fyi

packages/types/lib/environment/db.ts
```diff
@@ -9,7 +9,7 @@ export interface EnvironmentVariable extends Timestamps {
     value_tag?: string | null;
 }
 
-export interface Environment extends Timestamps {
+export interface DBEnvironment extends Timestamps {
```
TBonnin, line 12: It might be too difficult to do right now but I would prefer to see `DB...` types used internally in model where data is written/read from the db and have it mapped/returned as non `DB` types for the rest of the codebase to use. 
Like that we can take care of any necessary type or case transformation. That's one of the reason we have camelCase and snake_case mixed all over the codebase.

TBonnin (commented):
Team instead of Account makes a lot of sense. Should the table be renamed also?

Made a comment about the `DB` types. Happy to discuss in person

packages/types/lib/environment/db.ts
bodinsamuel: What would you suggest in term of naming? 
Something like `DBEnvironment` for the model, `Environment` for internal use and `APIEnvironment` for external use? Could work, not entirely sure that's it's worth it. 
Just to clarify my view, for me DB means internal but shorter, that allows us to differentiate two non-compatible interface, but that's about it. 
For the camelCase and snake_case issue that's for sure annoying, but it's a lot of effort for not a lot of benefits (except making sure we only select full model every time). Plus it's totally possible to use camelCase in pg (just a bit annoying to quote fields) so the difference might not even exists for new table if we wanted to.

Happy to hear more possible things that could be solved by having more types, I might forgot something.

TBonnin: discussed online,
I am in favor of several specific types adapted to where/what they are used for. Your example with `APIEnvironment`, `DBEnvironment` and `Environment` is a very good example. You can even push it further because, like you mentioned, you might even need different types for DB creations and updates for instance.
Basically I prefer everything to be explicit, even if it means more types and mapping between them. Otherwise we end up with a monster type that must handle all use cases, with a lot of optional fields to fit all scenarios and you never really know what you can reliably depend on or not.

For now I am ok with the `DBsomething` naming. At least is clear that models are coming directly from the db

TBonnin (approved):

PR #2491, opened 2024-07-10
feat(integration-templates): [nan-1350] internal template versioning process

khaliqgant:
## Describe your changes
Allow a bump in the nango.yaml version to signify that a template can be upgraded in the UI

image

- Also fixes the versions not showing when deploying integrations via the CLI

NAN-1350


packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -91,12 +97,19 @@ export async function deploy({
...
+        const deployResult: SyncDeploymentResult = {
             name: flow.syncName,
             models: flow.models,
+            version: response.syncConfig.version as string,
             providerConfigKey: flow.providerConfigKey,
             type: flow.type
-        });
+        };
+
+        if (response.syncConfig.version) {
```
khaliqgant, line 114: Without this the deployment result was showing as

```
Successfully deployed the scripts: contacts@vundefined, create-lead@vundefined, update-lead@vundefined, get-leads@vundefined, get-schema@vundefined, get-company@vundefined, create-company@vundefined, create-opportunity@vundefined, create-contact@vundefined, list-stages@vundefined, event-types@vundefined, event-invitees@vundefined, sub-app@vundefined, sub@vundefined, support@vundefined, revoked-sync@vundefined, revoked-action@vundefined, get-accounts@vundefined, webh@vundefined, fake-data@vundefined, by-action@vundefined, bad-response@vundefined, four-oh-one@vundefined, triggering-action@vundefined, blocked@vundefined, super-long-action-that-should-reach-some-kind-of-limit-for-being-too-long@vundefined!
```

packages/server/lib/controllers/config.controller.ts
```diff
@@ -59,6 +59,22 @@ const separateFlows = (flows: NangoSyncConfig[]): FlowConfigs => {
     );
 };
 
+const findPotentialUpgrades = (enabledFlows: NangoSyncConfig[], publicFlows: NangoSyncConfig[]) => {
+    return enabledFlows.map((flow) => {
+        const publicFlow = publicFlows.find((publicFlow) => publicFlow.name === flow.name);
+        if (!publicFlow) {
+            return flow;
+        }
```
TBonnin, line 67: Can `publicFlow` be undefined or null in a `find`?

packages/server/lib/controllers/config.controller.ts
```diff
@@ -59,6 +59,22 @@ const separateFlows = (flows: NangoSyncConfig[]): FlowConfigs => {
...
+        if (publicFlow && publicFlow.version !== flow.version) {
```
TBonnin, line 68: what if the version is smaller?

packages/server/lib/controllers/v1/flow/upgrade/preBuilt.ts
```diff
@@ -0,0 +1,49 @@
...
+
+const validation = flowConfig.extend({
+    id: z.string(),
+    upgrade_version: z.string(),
+    last_deployed: z.string(),
+    is_public: z.boolean(),
+    pre_built: z.boolean()
+});
+
+export const upgradePreBuilt = asyncWrapper<UpgradePreBuiltFlow>(async (req, res) => {
```
TBonnin: we should agree on a definitive naming convention and filename for endpoints.
I really like  when the filepath matches the endpoint path with the file name including the method.  For instance the endpoint here could be `PUT /api/v1/flow/prebuilt` and the file `.../controllers/v1/flow/putPreBuild.ts`.
I like it because when looking for the endpoint code you can directly look for the filepath without having to first look into `routes` and all the code relative to a subpath is neatly organized in the same subdir

packages/shared/lib/services/file/local.service.ts
```diff
@@ -48,12 +48,10 @@ class LocalFileService {
     public putIntegrationFile(syncName: string, fileContents: string, distPrefix: boolean) {
         try {
             const realPath = fs.realpathSync(process.env['NANGO_INTEGRATIONS_FULL_PATH'] as string);
-            if (!fs.existsSync(`${realPath}${distPrefix ? '/dist' : ''}/${syncName}`)) {
```
TBonnin, line 51: why is this being deleted?

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -91,12 +97,19 @@ export async function deploy({
...
         syncConfigs.push(response.syncConfig);
-        deployResults.push({
+        const deployResult: SyncDeploymentResult = {
             name: flow.syncName,
             models: flow.models,
+            version: response.syncConfig.version as string,
             providerConfigKey: flow.providerConfigKey,
             type: flow.type
-        });
+        };
```
TBonnin, line 112: you can have a one liner:
```
const deployResult: SyncDeploymentResult = {
...

...(response.syncConfig.version ? { version: response.syncConfig.version} : {})
}
```

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -204,6 +217,162 @@ export async function deploy({
...
+        throw new NangoError('file_upload_error');
+    }
+
+    await remoteFileService.copy(
+        provider,
+        `${type}s/${name}.ts`,
+        `${env}/account/${account.id}/environment/${environment.id}/config/${syncConfig.nango_config_id}/${name}.ts`,
+        environment.id,
+        `${name}.ts`
+    );
```
TBonnin, line 284: we don't check if this copy was successful or not?

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -204,6 +217,162 @@ export async function deploy({
...
+        sync_name: name,
+        sync_type: flowConfig.sync_type,
+        runs: flowConfig.runs,
+        models: flowConfig.models,
+        metadata: flowConfig.metadata || {},
+        nango_config_id: syncConfig.nango_config_id,
+        file_location,
+        version,
+        active: true,
+        model_schema: JSON.stringify(model_schema) as unknown as SyncModelSchema[],
```
TBonnin, line 303: Is it useful to parse it above to then stringify it here?

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -204,6 +217,162 @@ export async function deploy({
...
+        environment_id: environment.id,
+        deleted: false,
+        type,
+        auto_start: flowConfig.auto_start === false ? false : true,
+        pre_built: true,
+        is_public,
+        enabled: true,
+        webhook_subscriptions: flowConfig.webhookSubscriptions || []
+    } as SyncConfig;
```
TBonnin: is this typecasting necessary?

packages/types/lib/flow/http.api.ts
```diff
@@ -0,0 +1,12 @@
+import type { ApiError, Endpoint } from '../api';
+import type { IncomingFlowConfigUpgrade } from '../deploy/incomingFlow';
+
+export type UpgradePreBuiltFlow = Endpoint<{
+    Method: 'POST';
+    Path: '/api/v1/flow/upgrade/pre-built';
```
TBonnin, line 6: see my comment above. I think `PUT /api/v1/flow/prebuilt` would be a better path.

packages/webapp/src/pages/Integration/FlowPage.tsx
```diff
@@ -62,6 +64,73 @@ export default function FlowPage(props: FlowPageProps) {
...
+                    id: sync.id,
+                    enabled: sync.enabled,
+                    version: sync.version,
+                    upgrade_version: sync.upgrade_version,
+                    last_deployed: sync.last_deployed
+                });
+            }
+        } else {
+            const action = actions.find((action: Flow) => action.name === flow?.name);
+            if (action) {
```
TBonnin, line 93: same question. Can action be ever null/undefined?

packages/server/lib/controllers/config.controller.ts
khaliqgant: That shouldn't really happen. I see no scenario where we downgrade a public template.

packages/shared/lib/services/file/local.service.ts
khaliqgant: Because in an upgrade scenario we want to overwrite the file so the check for the local file existence is no needed. Also in an install scenario it would be OK to overwrite the file as well.

packages/server/lib/controllers/config.controller.ts
khaliqgant: Very possible. If it is a custom sync there will be no corresponding `publicFlow`

packages/webapp/src/pages/Integration/FlowPage.tsx
khaliqgant: Answered https://github.com/NangoHQ/nango/pull/2491#discussion_r1672773562

packages/shared/lib/services/sync/config/deploy.service.ts
khaliqgant: I personally find that less readable

packages/shared/lib/services/sync/config/deploy.service.ts
khaliqgant: In this case it is just copying the typescript file which isn't essential to the operation. If it fails it isn't really a big deal and shouldn't block the rest of the operation.

packages/shared/lib/services/sync/config/deploy.service.ts
khaliqgant: Updated

packages/types/lib/flow/http.api.ts
khaliqgant: Updated to a `PUT`

packages/shared/lib/services/sync/config/deploy.service.ts
khaliqgant: Yeah, necessary to insert it correctly into the jsonb column

packages/server/lib/controllers/v1/flow/upgrade/preBuilt.ts
khaliqgant: I hear ya, updated

packages/cli/lib/nango.yaml.schema.v2.json
```diff
@@ -50,6 +50,12 @@
                                             "type": "auto_start must be a boolean."
                                         }
                                     },
+                                    "version": {
+                                        "type": "string",
+                                        "errorMessage": {
+                                            "type": "version must be a string."
+                                        }
```
bodinsamuel, line 71: can you add pattern regex so it's consistently enforced?

packages/webapp/src/components/ui/button/Button.tsx
```diff
@@ -4,6 +4,8 @@ import { cva } from 'class-variance-authority';
 import { forwardRef } from 'react';
 import { cn } from '../../../utils/utils';
 
+export type ButtonVariants = 'primary' | 'secondary' | 'success' | 'danger' | 'zombie' | 'zombieGray' | 'yellow' | 'black' | 'active' | 'hover' | 'zinc';
```
bodinsamuel: to avoid desync you can maybe do something like `export type ButtonVariants = VariantProps<typeof buttonStyles>['variant'];`

packages/server/lib/controllers/v1/flow/upgrade/putPreBuilt.ts
```diff
@@ -0,0 +1,49 @@
...
+    const val = validation.safeParse(req.body);
+
+    if (!val.success) {
+        res.status(400).send({
+            error: { code: 'invalid_body', errors: zodErrorToHTTP(val.error) }
+        });
+        return;
+    }
+
+    const flowConfig: IncomingFlowConfigUpgrade = val.data as IncomingFlowConfigUpgrade;
```
bodinsamuel: Since zod and types is not synced we tried to assing the body to the type however if you enforce type cast it's defeating the purpose. Here for example `is_public` is not compatible 
You can fix it by using `is_public: z.literal(true),`
```suggestion
    const flowConfig: IncomingFlowConfigUpgrade = val.data;
```

packages/shared/lib/services/sync/config/deploy.service.ts
bodinsamuel: also agree on readability

bodinsamuel (approved):
Looks good 👍🏻 
Minor comments

packages/server/lib/controllers/v1/flow/upgrade/putPreBuilt.ts
khaliqgant: Good point, updated

packages/webapp/src/components/ui/button/Button.tsx
khaliqgant: Thanks, updated

packages/cli/lib/nango.yaml.schema.v2.json
khaliqgant: Added

bodinsamuel: there is a small typo interger

khaliqgant: Thanks, will fix.

packages/shared/lib/services/sync/config/deploy.service.ts
TBonnin: It is true the line is less readable but it allows a self-contained struct definition in a single block instead of having to mutate it in a second time. Not a massive deal since here it happens right below but as you know codebase evolves

khaliqgant: Yeah, understood your point

PR #2493, opened 2024-07-10
feat: refactor Team Settings

bodinsamuel:
### Major changes
- Team settings revamp
Fixes https://linear.app/nango/issue/NAN-1409/team-settings-revamp
Redesign the page following Figma design. It's more or less the same thing as before just cleaner.

- New endpoints for team edit, and invite

- Remove member release the email
Fixes https://linear.app/nango/issue/NAN-1383/remove-member-to-release-email
When we remove a member we now remove it from the team and create this person a new team (we can't have 0 team for the moment)

### Minor changes
- Bulk invitation
Fixes https://linear.app/nango/issue/NAN-1380/bulk-invite

- Possibility to revoke invitation
Fixes https://linear.app/nango/issue/NAN-1385/revoke-invitations

- Re inviting should invalidate previous invitation
Fixes https://linear.app/nango/issue/NAN-1382/re-invite-should-invalid-previous-invite

## Test

https://github.com/user-attachments/assets/397f99a6-7f1e-491a-acd4-801d81f29bfa


packages/server/lib/controllers/v1/team/invite/postInvite.ts
```diff
@@ -0,0 +1,59 @@
+import { asyncWrapper } from '../../../../utils/asyncWrapper.js';
+import { requireEmptyQuery, zodErrorToHTTP } from '@nangohq/utils';
+import type { PostInvite } from '@nangohq/types';
```
TBonnin: if the route is `/api/v1/team/invite`, should this file path be `.../controllers/v1/team/postInvite.ts` (without the `invite` subfolder).

packages/server/lib/controllers/v1/team/users/deleteTeamUser.ts
```diff
@@ -0,0 +1,59 @@
...
+
+    // Account ID is not nullable until we change the way we deal with default account so we create a temp one
+    const newTeam = await accountService.createAccount(`${user.name}'s Organization`);
+    if (!newTeam) {
+        res.status(500).send({ error: { code: 'server_error' } });
+        return;
+    }
+
+    const removed = await userService.update({ id: user.id, account_id: newTeam.id });
+    if (removed <= 0) {
```
TBonnin: `removed`?

packages/server/lib/controllers/v1/team/invite/postInvite.ts
```diff
@@ -0,0 +1,59 @@
...
+    for (const email of body.emails) {
+        const existingUser = await userService.getUserByEmail(email);
+        if (existingUser && existingUser.account_id === account.id) {
+            continue;
+        }
+
+        const invitation = await db.knex.transaction(async (trx) => {
+            await expirePreviousInvitations({ email, accountId: account.id, trx });
+
+            return await inviteEmail({ email, name: email, accountId: account.id, invitedByUserId: user.id, trx });
```
TBonnin: nitpick: how do you feel about `return await` syntax. For me it is just unnecessarily heavy and unless necessary for type inference reason I don't really see the point. Curious about what you think?

packages/shared/lib/services/account.service.ts
```diff
@@ -88,12 +80,12 @@ class AccountService {
...
      * Create Account
-     * @desc create a new account and assign to the default environmenets
+     * @desc create a new account and assign to the default environments
      */
     async createAccount(name: string): Promise<DBTeam | null> {
-        const result: void | Pick<DBTeam, 'id'> = await db.knex.from<DBTeam>(`_nango_accounts`).insert({ name: name }, ['id']);
+        const result = await db.knex.from<DBTeam>(`_nango_accounts`).insert({ name }).returning('*');
 
-        if (Array.isArray(result) && result.length === 1 && result[0] != null && 'id' in result[0]) {
+        if (Array.isArray(result) && result.length === 1 && result[0]) {
```
TBonnin: can the condition be simplified to `if (result?.[0])`
Same in invitations.ts

packages/shared/lib/services/invitations.ts
```diff
@@ -0,0 +1,101 @@
...
+        .where({ account_id: accountId, accepted: false })
+        .whereRaw('expires_at > ?', date);
+
+    return result || [];
+}
+
+export async function acceptInvitation(token: string) {
+    const result = await db.knex.from<DBInvitation>(`_nango_invited_users`).where({ token }).update({ accepted: true, expires_at: new Date() });
+
+    return result;
```
TBonnin: no need for result here. no?

packages/shared/lib/services/invitations.ts
```diff
@@ -0,0 +1,101 @@
...
+            id: 1,
+            email: '',
+            name: '',
+            account_id: 0,
+            invited_by: 0,
+            token: '',
+            expires_at: new Date(),
+            accepted: true,
+            created_at: new Date(),
+            updated_at: new Date()
```
TBonnin: should `date` be used here?

TBonnin (approved):

packages/server/lib/controllers/v1/team/users/deleteTeamUser.ts
bodinsamuel: oops yes copy/pasta 👍🏻

packages/shared/lib/services/account.service.ts
bodinsamuel: yeah indeed, I'm not trusting 100% knex type but should work

packages/shared/lib/services/invitations.ts
bodinsamuel: what do you mean? (note: I just copied this code)

packages/server/lib/controllers/v1/team/invite/postInvite.ts
bodinsamuel: I personally prefer explicit over implicit also it's more consistent if you try/catch. So if we most of the time await but then at some place we return directly I feel like it's more surprising than always awaiting.
That also means we can't enforce it by an eslint rules and can be a footgun if you wrongfully do 
```ts
try {
  return promise();
} catch {}
```

There is --I think-- a benefits from returning instead of awaiting for the stack optimization but not entirely sure

packages/server/lib/controllers/v1/team/invite/postInvite.ts
bodinsamuel: I think what we did is: `path.split('/)`, so if `/api/v1/team` is in folder team, I feel like it's consistent to put invite in its own folder too. 
Now that I think about it, I could have put invite in `/api/v1/invite` maybe that could clarify, wdyt?

packages/shared/lib/services/invitations.ts
TBonnin: I am talking about using `const date = new Date();` above instead of creating new dates

bodinsamuel: ah yes definitely, good point

packages/server/lib/controllers/v1/team/invite/postInvite.ts
bodinsamuel: moved those to `/api/v1/invite` no reason for them to depend on team since most endpoints are also bound to a team

PR #2494, opened 2024-07-10
fix: token refresh locking logic

TBonnin:
One customer is experiencing connection refresh failures with the following error:
Acquiring lock for key: lock:refresh:XXX timed out after 12000ms They are making concurrent calls to `/proxy` while the token is expired. Each request trying to acquire the lock is waiting and then attempts again to refresh the token. When a lot of concurrent requests, waiting for all the other requests in the queue to refetch the connection token can exceed the lock acquisition timeout.
This commit is introducing 2 fixes:
- after waiting for acquiring the lock, first check if the connection has been refreshed, if yes returns the credentials directly.
- if lock acquisition times out, also check if the connection has been refreshed and only fails if it is not the case.

https://linear.app/nango/issue/NAN-1164/salesforcebox-connections-failing-to-refresh-intermittently

### DIFF
--- a/packages/shared/lib/services/connection.service.ts
+++ b/packages/shared/lib/services/connection.service.ts
@@ -863,7 +863,11 @@ class ConnectionService {
     }): Promise<ServiceResponse<{ refreshed: boolean; credentials: OAuth2Credentials | AppCredentials | AppStoreCredentials | OAuth2ClientCredentials }>> {
         const providerConfigKey = providerConfig.unique_key;
 
-        const getConnection = async (): Promise<{ connection: Connection; shouldRefresh: boolean }> => {
+        // fetch connection and return credentials if they are fresh
+        const getConnectionAndFreshCredentials = async (): Promise<{
+            connection: Connection;
+            freshCredentials: OAuth2Credentials | AppCredentials | AppStoreCredentials | OAuth2ClientCredentials | null;
+        }> => {
             const { success, error, response: connection } = await this.getConnection(connectionId, providerConfigKey, environmentId);
 
             if (!success || !connection) {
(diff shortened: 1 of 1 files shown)


packages/shared/lib/services/connection.service.ts
```diff
@@ -881,18 +882,70 @@ class ConnectionService {
...
+                                success: true,
+                                error: null,
+                                response: {
+                                    refreshed: true,
+                                    credentials: connection.credentials as OAuth2Credentials | AppCredentials | AppStoreCredentials | OAuth2ClientCredentials
+                                }
+                            };
+                        }
+                    }
+                    throw err;
```
TBonnin: there is some code duplication but it is not trivial to consolidate the logci into one function. Let me know if you think I should spend more time trying to be more DRY

khaliqgant: I find this logic very confusing and hard to follow. This method has always been hard to follow IMO but if possible it would be great to make it more readable / easier to follow. Definitely push back if you think it would be too much of a time sink though

khaliqgant (approved):
Fix makes sense, but the logic is hard to follow. See my feedback inline

packages/shared/lib/services/connection.service.ts
TBonnin: The locking logic is indeed complicated. My quick attempt at trying to abstract the complexity was not very successful though, I was basically just moving the complexity around. I am happy to pair if you have some idea about how to improve the locking logic

TBonnin: Made an attempt at removing the duplication [7ad0229](https://github.com/NangoHQ/nango/pull/2494/commits/7ad022989373db94678aa34db4de5b8b616c8a8e)
The concurrency/locking logic is still the same though

packages/shared/lib/services/connection.service.ts
```diff
@@ -846,150 +847,151 @@ class ConnectionService {
...
+        if (!shouldRefresh) {
+            return {
+                success: true,
+                error: null,
+                response: {
+                    refreshed: false,
+                    credentials: connection.credentials as OAuth2Credentials | AppCredentials | AppStoreCredentials | OAuth2ClientCredentials
+                }
+            };
+        }
```
bodinsamuel: this early call/return is incorrect, if a refresh is in progress the credentials can be invalid just after the return

bodinsamuel (commented):
Agreed with Khaliq it's hard to read, maybe there is an alternative with a recursive call (not a good idea) or a while loop.

Also I think my comment about the early return could explain some bugs, we should always wait for lock and re-fetch the connection  .

```ts
# pseudo code
const lockKey = `lock:refresh:${environment_id}:${providerConfigKey}:${connectionId}`;
const ttlInMs = 10000;
const acquisitionTimeoutMs = ttlInMs * 1.2; 
let maxTries = 0;
try {
    do {
        try {
            await this.locking.tryAcquire(lockKey, ttlInMs, acquisitionTimeoutMs);
        } catch (err) {
            report(err);
            maxTries += 1;
            continue;
        }

        const { connection, shouldRefresh } = await getConnection();
        if (!shouldRefresh) {
            return { success: true, error: null, response: { refreshed: false, credentials: connection.credentials as any } };
        }

        const newCreds = await this.getNewCredentials(connection, providerConfig, template);
        return { success: true, error: null, response: { refreshed: true, credentials: newCreds } };
    } while (maxTries <= 10);
} finally {
    await this.locking.release(lockKey);
}
```

packages/shared/lib/services/connection.service.ts
TBonnin: good point. that's another problem that I haven't address, even though in practice I am not sure it can happen.

TBonnin:
> Agreed with Khaliq it's hard to read, maybe there is an alternative with a recursive call (not a good idea) or a while loop.
> 
> Also I think my comment about the early return could explain some bugs, we should always wait for lock and re-fetch the connection .

I agree the early return might explain some issues. I have modified my PR to always acquire the lock and re-fetch the connection. 
I haven't added the loop because I feel it would add to much latency to retry several times. I also kept the attempt at checking the freshness of the credentials when the lock times out. It is still duplication but only twice :D [fce7cbe](https://github.com/NangoHQ/nango/pull/2494/commits/fce7cbe9cfe993cc9fc3f899f429da1b59e6067c)
Let me know what you think @bodinsamuel

bodinsamuel:
Yes agreed for now, we know it can be improved like this but it has some drawbacks 👌🏻

bodinsamuel (approved):

PR #2495, opened 2024-07-10
feat(integration-templates): improve sharepoint templates

hassan254-prog:
## Describe your changes

- Added mapper functions to both templates.
- Removed the logic of saving `deltaTokens` for incremental sync.
- Since we no longer need the paginate helper function, I adopted the built-in `nango.paginate` function to replace it.

[EXT-95](https://linear.app/nango/issue/EXT-95/add-sharepoint-template)


integration-templates/sharepoint-online/actions/list-sites.ts
```diff
@@ -1,40 +1,35 @@
...
-    const config = {
+/**
+ * Retrieves SharePoint sites using NangoAction, maps them to Site objects,
+ * updates the SharePoint metadata with the sites to sync, and returns the mapped sites.
+ *
+ * @param nango An instance of NangoAction for handling listing of sites.
+ * @param _input Optional input parameter (not typically used for this specific action).
+ * @returns An array of Site objects representing SharePoint sites
+ */
+export default async function runAction(nango: NangoAction, _input?: void): Promise<Site[]> {
```
TBonnin: I would remove the input all together if it is not being used

integration-templates/sharepoint-online/actions/list-sites.ts
```diff
@@ -1,40 +1,35 @@
...
+    const response = await nango.get<{ value: SharePointSite[] }>({
         endpoint: 'v1.0/sites',
-        paginate: {
-            type: 'link',
-            limit_name_in_request: '$top',
-            response_path: 'value',
-            link_path_in_response_body: '@odata.nextLink',
-            limit: 100
-        },
```
TBonnin, line 15: why is paginate not used anymore?

integration-templates/sharepoint-online/mappers/to-file.ts
```diff
@@ -0,0 +1,26 @@
...
+ * @returns FileMetadata object representing file metadata.
+ */
+export function toFile(file: DriveItem): FileMetadata {
+    const fileMetadata: FileMetadata = {
+        id: file.id,
+        etag: file.eTag,
+        cTag: file.cTag,
+        name: file.name,
+        is_folder: 'folder' in file,
+        mime_type: 'folder' in file ? null : file.file?.mimeType ?? null,
```
TBonnin: can it simply be `file.file?.mimeType ?? null`?

integration-templates/sharepoint-online/mappers/to-file.ts
```diff
@@ -0,0 +1,26 @@
...
+        path: file.parentReference.path,
+        raw_source: file,
+        updated_at: new Date(file.lastModifiedDateTime).toISOString(),
+        download_url: 'folder' in file ? null : file['@microsoft.graph.downloadUrl'],
```
TBonnin: same question here.

integration-templates/sharepoint-online/mappers/to-site.ts
```diff
@@ -0,0 +1,17 @@
...
+ * @returns Site object representing SharePoint site information.
+ */
+export function toSite(site: SharePointSite): Site {
+    return {
+        id: site.id,
+        name: site.name,
+        createdDateTime: site.createdDateTime,
+        webUrl: site.webUrl
+    };
+}
```
TBonnin, line 17: looks like `Site` and `SharePointSite` have the same shape. Is this mapping necessary then?

integration-templates/sharepoint-online/nango.yaml
```diff
@@ -4,35 +4,42 @@ integrations:
             list-sites:
                 description: |
                     This action will be used to display a list of sites to the end-user, who will pick the ones he wants to sync.
-                endpoint: GET /sharepoint/list-sites
+                endpoint: POST /sharepoint/list-sites
```
TBonnin: looks like the action is simply fetching data. Why moving to a POST method?

integration-templates/sharepoint-online/syncs/file-sync.ts
```diff
@@ -1,121 +1,108 @@
...
+                response_path: 'value',
+                link_path_in_response_body: '@odata.nextLink',
+                limit: 100
+            },
+            retries: 10
+        };
+        // Paginate through lists and filter documentlibraries
+        for await (const lists of nango.paginate(config)) {
+            const documentLibraries = lists.filter((list: any) => list.list.template === 'documentLibrary');
+            siteIdToLists[siteId] = documentLibraries.map((l: any) => l.id);
```
TBonnin: you don't have to create an intermediary variable by chaining filter and map. ex:
```
siteIdToLists[siteId] = lists
    .filter((list: any) => list.list.template === 'documentLibrary')
    .map((l: any) => l.id)
```
`flatMap` is also an option

integration-templates/sharepoint-online/syncs/file-sync.ts
```diff
@@ -1,121 +1,108 @@
...
 
-    await collectPaginatedData<FileMetadata>(nango, params, async (page: PaginationResponse<FileMetadata>) => {
-        for (const item of page.value) {
+    // Paginate through list items and sync each file metadata
+    for await (const listItems of nango.paginate(config)) {
+        for (const item of listItems) {
             const file = await fetchDriveItemDetails(nango, siteId, listId, item.id);
-            const metadata = fileMetadata(file);
-            await nango.batchSave<FileMetadata>([metadata], 'FileMetadata');
+            await nango.batchSave<FileMetadata>([file], 'FileMetadata');
```
TBonnin: saving one file at a time might not be the most effective. can we do something like:
```
for await (const listItems of nango.paginate(config)) {
        const allMetadata = Promise.all(listItems.map((item) => fetchDriveItemDetails(nango, siteId, listId, item.id)));
        await nango.batchSave<FileMetadata>([allMetadata], 'FileMetadata');
}
```

packages/shared/flows.yaml
```diff
@@ -2670,39 +2670,46 @@ integrations:
                 description: >
                     This action will be used to display a list of sites to the end-user,
                     who will pick the ones he wants to sync.
-                endpoint: GET /sharepoint/list-sites
+                endpoint: POST /sharepoint/list-sites
```
TBonnin: same question about POST vs GET

integration-templates/sharepoint-online/actions/list-sites.ts
hassan254-prog: Updated this in the [commit](https://github.com/NangoHQ/nango/pull/2495/commits/6385422e1a163b9a87bc615308f11fe8910880b4) 👍

integration-templates/sharepoint-online/mappers/to-file.ts
hassan254-prog: Updated this in the [commit](https://github.com/NangoHQ/nango/pull/2495/commits/6385422e1a163b9a87bc615308f11fe8910880b4) 👍

integration-templates/sharepoint-online/mappers/to-file.ts
hassan254-prog: same in here 👍

integration-templates/sharepoint-online/mappers/to-site.ts
hassan254-prog: The `SharepointSite`  interface includes  all the fields returned by sharepoint endpoint that we need to map out to a slimmer `Site` object with only essential ones. Please let me know if this mapping is still necessary.

integration-templates/sharepoint-online/nango.yaml
hassan254-prog: Updated this as well 👍

integration-templates/sharepoint-online/syncs/file-sync.ts
hassan254-prog: Thanks for the pointing this out🙏! Updated.

integration-templates/sharepoint-online/nango.yaml
```diff
@@ -8,31 +8,38 @@ integrations:
...
                 sync_type: incremental
-                track_deletes: true
                 runs: every 1 hour
                 scopes:
                     - Sites.ReadWrite.All
                     - offline_access
-                auto_start: true
+                input: SharePointMetadata
+                auto_start: false
                 output: FileMetadata
```
khaliqgant, line 25:
```suggestion
                output: FileMetadata
                version: 1.0.0
```

khaliqgant: We can talk about this versioning piece during our next sync

integration-templates/sharepoint-online/syncs/file-sync.ts
hassan254-prog: This implementation will introduce parallel requests, which can potentially defeat the request retry policy. Additionally, it may cause problems due to Sharepoint’s API throttling. Please let me know if we should proceed with this approach?

integration-templates/sharepoint-online/actions/list-sites.ts
hassan254-prog: Since we expect the number of sites to be very small, there was no need to paginate here.

integration-templates/sharepoint-online/nango.yaml
hassan254-prog: @khaliqgant, sure, let's discuss the versioning piece during our next sync. I think it might be related to the `version@undefined` error I was encountering earlier during `nango deploy`.

bodinsamuel (approved):

integration-templates/sharepoint-online/syncs/file-sync.ts
TBonnin: good point about concurrency. we can still fetch the details sequentially and make a single call to batchSave instead of one per item. What do you think?
```
for await (const listItems of nango.paginate(config)) {
        const allMetadata = listItems.map((item) => await fetchDriveItemDetails(nango, siteId, listId, item.id));
        await nango.batchSave<FileMetadata>(allMetadata, 'FileMetadata');
}
```

hassan254-prog: That is a better alternative. However, my initial focus was also on optimizing memory efficiency. Wouldn't this alternative approach potentially impact memory efficiency, especially when we hold data for too long, such as during retries while fetching drive item? What do you think?

TBonnin: is memory a problem? also how big are pages? If they are a few hundred items I think it is fine to hold them in memory before to batchSave them. The overhead to batchSave (which is a round-trip to persist over the network) would make saving much slower.

bodinsamuel:
Is this ready to merge?

integration-templates/sharepoint-online/syncs/file-sync.ts
hassan254-prog: Yes, in my opinion, we need to consider memory efficiency when writing syncs. We should batch save as soon as possible. However, in this case, we can hold in memory since we're dealing with a slimmer file object and not saving the actual file itself. The response is paginated, so we will always have fewer records to batch save. I've also made a small tweak to the code: we cannot use `await` directly inside a synchronous callback function like `map`. Please take a look at the [changes](https://github.com/NangoHQ/nango/pull/2495/commits/3a0c1ce17b46a86e78475e0c3b2a0279c16e73f8) and provide me with your feedback. Thank you!

hassan254-prog:
@bodinsamuel, waiting for approval from @TBonnin.

TBonnin (approved):

hassan254-prog:
@bodinsamuel, this looks good to merge.

Issue #2312, opened 2024-06-12
Error compiling a sync .ts file -  SyntaxError: Unexpected token '', "��z��" is not valid JSON


bburns:
I'm getting this same error - am on Windows - 

```
> nango dev --debug

Debug mode enabled
.env file detected and loaded
Is run via npx: false. Is locally installed: false
Version 0.40.0 of nango is installed.
Latest version of nango is 0.40.0.
Current full working directory is read as: C:\Users\bburns\Workspace\neomem\nango\nango-integrations
Current stripped directory is read as: nango-integrations
Found nango.yaml file successfully.
Watching ./nango.yaml
Watching ./**/*.ts, ./nango.yaml

Error compiling google-mail\syncs\emails.ts:
SyntaxError: Unexpected token '', "��z��" is not valid JSON
    at JSON.parse (<anonymous>)
    at new AnyMap (C:\Users\bburns\AppData\Roaming\npm\node_modules\nango\node_modules\@cspotcode\source-map-support\node_modules\@jridgewell\src\any-map.ts:20:37)
    at mapSourcePosition (C:\Users\bburns\AppData\Roaming\npm\node_modules\nango\node_modules\@cspotcode\source-map-support\source-map-support.js:387:14)
    at wrapCallSite (C:\Users\bburns\AppData\Roaming\npm\node_modules\nango\node_modules\@cspotcode\source-map-support\source-map-support.js:592:20)
    at Function.prepareStackTrace (C:\Users\bburns\AppData\Roaming\npm\node_modules\nango\node_modules\@cspotcode\source-map-support\source-map-support.js:671:41)
    at maybeOverridePrepareStackTrace (node:internal/errors:140:29)
    at prepareStackTrace (node:internal/errors:114:5)
    at getStackString (node:internal/util/inspect:1240:16)
    at formatError (node:internal/util/inspect:1369:15)
    at formatRaw (node:internal/util/inspect:985:14)
```

emails.ts:

```
import type { NangoSync, GmailEmail } from '../../models'

export default async function fetchData(nango: NangoSync) {
  const backfillPeriod = new Date(Date.now() - 24 * 60 * 60 * 1000) // 24 hours ago.
  const { lastSyncDate } = nango
  const syncDate = lastSyncDate || backfillPeriod

  const pageSize = 100
  let nextPageToken: string | undefined = ''

  do {
    const response: any = await nango.proxy({
      method: 'GET',
      endpoint: '/gmail/v1/users/me/messages',
[...]

PR #2496, opened 2024-07-10
fix(cli): handle Windows paths (#2312)

bburns:
This commit uses forward slashes on Windows for the calls to glob.sync and tsup.build - those are where `nango compile` and `nango dev` were failing.

I copied the slash function from https://github.com/sindresorhus/slash, as it just converts \ to /, and didn't want to deal with package-lock.json changes.

It also uses path.join or path.resolve instead of `${foo}/pok/bar.ts`, os.tmpdir() instead of '/tmp', etc.

getNangoRootPath() now always returns a string, since getPackagePath() will throw an error if it can't find the package.

And I expanded the regex in packages/node-client/lib/utils.unit.test.ts to handle "nango-node-client/0.41.0 (linux/5.15.153.1-microsoft-standard-WSL2; node.js/20.12.2)".

I ran the tests on Windows with git bash and Command Prompt, and with WSL2 Ubuntu -

    npm i
    npm run ts-build
    npm run test:cli
    npm run test:unit

and in a test nango project, these worked -

    $ node /c/users/bburns/Workspace/forks/nango/packages/cli/dist/index.js generate --debug
    $ node /c/users/bburns/Workspace/forks/nango/packages/cli/dist/index.js compile --debug
    $ node /c/users/bburns/Workspace/forks/nango/packages/cli/dist/index.js dev --debug

Also added a workflow to run test:cli and test:unit on Windows - .github/workflows/tests-windows.yaml.

#2312


bodinsamuel:
Thanks for your contribution, it was on our todo list so definitely helps. Reviewing that rn 👍🏻

packages/cli/lib/tests/helpers.ts
```diff
@@ -1,7 +1,8 @@
 import path from 'node:path';
 import fs from 'node:fs/promises';
+import os from 'node:os';
 
-export const fixturesPath = path.join(__dirname, '..', '..', 'fixtures');
+export const fixturesPath = './packages/cli/fixtures';
```
bodinsamuel, line 5: I remember having issues with this, but seems to work fine 🤔 I should have added a comment

packages/cli/lib/utils.ts
```diff
@@ -271,23 +271,18 @@ export function getUserAgent(): string {
...
 
-export function getNangoRootPath(debug = false) {
+export function getNangoRootPath(debug = false): string {
     const packagePath = getPackagePath(debug);
-    if (!packagePath) {
-        if (debug) {
-            printDebug('Could not find nango cli root path locally');
-        }
-        return null;
-    }
```
bodinsamuel, line 281: Indeed legacy stuff, thanks

.github/workflows/tests-windows.yaml
```diff
@@ -0,0 +1,35 @@
...
+concurrency:
+    group: tests-${{ github.event.pull_request.number || github.ref }}
+    cancel-in-progress: true
+
+jobs:
+    tests:
+        runs-on: windows-latest
+        strategy:
+            matrix:
+                node-version: [18.x, 20.x]
```
bodinsamuel: I think it's possible to use the existing `tests.yaml` and add the os in the matrix 

example https://github.com/vitest-dev/vitest/blob/main/.github/workflows/ci.yml

bodinsamuel (commented):
Massive work 💪🏻 
Minor comments, but looks good

bburns:
Okay, I wasn't sure if test:integration should run on Windows also, so
separated the others out.

bburns:
I realized I needed to set the tests-windows.yaml concurrency group to something unique, so fixed that, rebased to the latest upstream, and tested locally. 

It looks okay, except for the deployment actions - maybe could add some guards for if the PR is coming from a fork? I'm not too familiar with github ci/cd yet...

packages/cli/lib/tests/helpers.ts
bburns: yeah __dirname seemed to be missing in that file...

bodinsamuel (approved):
Looks good! 💪🏻 
I just need to discuss one point with the team otherwise it's good to merge

bodinsamuel:
Fixes https://linear.app/nango/issue/NAN-1324/nango-cli-doesnt-compile-on-windows

bodinsamuel:
Alright, I just modified the windows tests to only run CLI, because there are other issues to compile to full codebase in Windows.

bodinsamuel (approved):
Thanks a lot for this deeply needed change 🙏🏻

bburns:
No problem - love this project!

bodinsamuel:
It has been released in `0.42.0` 🙌🏻

PR #2497, opened 2024-07-11
fix(webapp): [nan-1295] stable ordering of syncs and actions

khaliqgant:
## Describe your changes
Sort the sync and actions by name to prevent reordering when enabling

NAN-1295


bodinsamuel (approved):
👌🏻 
Made sense to have things grouped by enabled / disabled. Usually in this kind of UI the grouping is done once at load time, so when you click it stays at the same place but when you refresh the grouping kicks again

PR #2500, opened 2024-07-11
fix(integration-templates): [nan-1366] add .nango dir and fix path

khaliqgant:
## Describe your changes
- Add the `.nango` directory on template upload
- Fix the path of the sync template to be the integration instead of the dir

NAN-1366

### DIFF
--- a/.github/workflows/sync-integration-templates.yaml
+++ b/.github/workflows/sync-integration-templates.yaml
@@ -30,10 +30,11 @@ jobs:
                       popd
 
                       cp -r /tmp/nango-temp/nango-integrations/dist ./$integration/dist
+                      cp -r /tmp/nango-temp/nango-integrations/.nango ./$integration/.nango
 
                       rm -rf /tmp/nango-temp
 
-                      aws s3 sync $dir s3://${{ secrets.AWS_BUCKET_NAME }}/$integration/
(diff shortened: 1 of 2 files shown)


.github/workflows/sync-integration-templates.yaml
```diff
@@ -30,10 +30,11 @@ jobs:
                       popd
 
                       cp -r /tmp/nango-temp/nango-integrations/dist ./$integration/dist
+                      cp -r /tmp/nango-temp/nango-integrations/.nango ./$integration/.nango
 
                       rm -rf /tmp/nango-temp
 
-                      aws s3 sync $dir s3://${{ secrets.AWS_BUCKET_NAME }}/$integration/
+                      aws s3 sync $integration s3://${{ secrets.AWS_BUCKET_NAME }}/$integration/
```
bodinsamuel, line 37: does the upload was broken before? it feels like an incompatible change no?

bodinsamuel (approved):

.github/workflows/sync-integration-templates.yaml
khaliqgant: Yeah. Was broken before.

TBonnin (approved):

PR #2501, opened 2024-07-11
fix(nango-yaml): disallow empty array

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1417/nangoyaml-should-throw-on-empty-array

- Do not allow this confusing syntax


khaliqgant (approved):
Docs could probably use a quick update of what we don’t allow and a quick note on why

PR #2502, opened 2024-07-12
fix(validation): cache ajv compilation

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1418/cache-ajv-data-validation

- Cache AJV compiled schema 
We saw an increase in batchSave latency, initially we thought there was an issue with the request but it's the records validation that was to blame. We compiled the schema at every record which was not a good idea. 
In this PR we now cache schema per `model-version` which allows stable cache and automatic invalidation on deploy. There is no automated cache clear strategy as everything is happening in the runner/cli, but it's definitely something to keep in mind if we use it somewhere else.
This drastically improve performance (or get it back to what it was before):
for 10K large records `42s -> 4s`

### DIFF
--- a/packages/shared/lib/sdk/dataValidation.ts
+++ b/packages/shared/lib/sdk/dataValidation.ts
@@ -1,38 +1,55 @@
-import type { ErrorObject } from 'ajv';
+import type { ErrorObject, ValidateFunction } from 'ajv';
 import { Ajv } from 'ajv';
 import addFormats from 'ajv-formats';
 import type { JSONSchema7 } from 'json-schema';
 
 export interface ValidateProps {
+    version: string;
     input: unknown;
     modelName: string | undefined;
     jsonSchema: JSONSchema7 | undefined | null;
 }
 
-export function validateData({ input, modelName, jsonSchema }: ValidateProps): true | (ErrorObject | Error)[] {
+let ajv: Ajv;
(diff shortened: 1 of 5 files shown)


khaliqgant (approved):

packages/shared/lib/sdk/dataValidation.ts
```diff
@@ -1,38 +1,55 @@
...
 export interface ValidateProps {
+    version: string;
     input: unknown;
     modelName: string | undefined;
     jsonSchema: JSONSchema7 | undefined | null;
 }
 
-export function validateData({ input, modelName, jsonSchema }: ValidateProps): true | (ErrorObject | Error)[] {
+let ajv: Ajv;
+const cache = new Map<string, ValidateFunction<any>>();
```
TBonnin, line 14: we don't limit the size of the cache. I guess it is not so much of a problem at the moment since it is one per customer/runner and also limited by the number of model/version. Might become a problem if we start sharing runners in the furure

TBonnin (approved):

packages/shared/lib/sdk/dataValidation.ts
bodinsamuel: yep definitely not future proof, if we share the runners it's also missing the accountId or something more unique like that

PR #2507, opened 2024-07-12
chore: dryrun should not depend on run.service

TBonnin:
First step of the refactoring of run.service. 
The end goal is to be able to split `run` into two components: one to start a script execution and the other one to complete/report on it results.

In order to be able to do that I am first removing the usage of run.service by cli dryrun. 
Very little logic from the run.service is actually necessary for dryrun and it is actually ok for dryrun execution to be synchronous since they are executed locally via the CLI 
I have not touched the runScript interface, other than some input fields tweaking. Even though I could push the refactoring further and avoid nangProps attributes that are only required for dryrun (ex: logMessages for instance I believe) I don't want to dig into a rabbit hole too much for now. Let me know what you think.

I have tested dry running sync/action on my machine but I would greatly appreciate if you could confirm with your own syncs/actions. Thank you

https://linear.app/nango/issue/NAN-1420/extract-dryrun-logic-from-runservice


packages/cli/lib/services/dryrun.service.ts
```diff
@@ -394,4 +388,133 @@ export class DryRunService {
             process.exit(1);
         }
     }
+
```
TBonnin, line 391: moving runScript into `dryrun.service`. I don't really see the point of having it in a different file. I have not change anything with the function, it is an exact copy of what it was before.

packages/jobs/lib/integration.service.ts
```diff
@@ -65,9 +63,11 @@ class IntegrationService implements IntegrationServiceInterface {
 
         const logCtx = activityLogId ? logContextGetter.getStateLess({ id: String(activityLogId) }) : null;
         try {
+            const integrationData = { fileLocation: nangoProps.syncConfig.file_location };
+            const environmentId = nangoProps.environmentId;
```
TBonnin, line 67: no point of having parameters that can be retrieved from another param

packages/shared/lib/models/Sync.ts
```diff
@@ -158,17 +157,20 @@ export interface RunScriptOptions {
...
-    integrationData: Pick<NangoIntegrationData, 'fileLocation'>;
-    environmentId: number;
     writeToDb: boolean;
     isInvokedImmediately: boolean;
     isWebhook: boolean;
     optionalLoadLocation?: string | undefined;
     input?: object | undefined;
 }
-export interface IntegrationServiceInterface {
+
```
TBonnin, line 168: this below is temporary. Types will change when I refactor and move `run.service`.

Also shared types will be moved to types package

packages/cli/lib/services/dryrun.service.ts
bodinsamuel: note: 
idealistically we should be able to have the script runner as a dedicated package and use it in the backend + in the cli

bodinsamuel (approved):
Tested locally, works well 👌🏻

packages/cli/lib/services/dryrun.service.ts
TBonnin: agreed. At the end that is probably the only thing that is common between the two paths. Not sure it will stay this way when we implement sandboxing though

bodinsamuel: cloudflare wrangled is doing that I think https://github.com/cloudflare/workers-sdk

PR #2513, opened 2024-07-15
feat: refactor User Settings

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1430/refacto-profile-settings

### Major changes
- New user settings page
- Add related endpoints

### Minor changes
- Reduce session duration
I just happen to notice that long session period. This is debatable but all security recommendations are stricter than that, and we store information there that can be outdated + we show credentials, it seems like it's a good idea to go lower (ideally max 24h but it's also very annoying)

- Remove unused types
- Renamed WebUser to ApiUser for consistency


packages/shared/lib/services/user.service.ts
```diff
@@ -9,17 +9,9 @@ const VERIFICATION_EMAIL_EXPIRATION = 3 * 24 * 60 * 60 * 1000;
...
-        if (result == null || result.length == 0 || result[0] == null) {
-            return null;
-        }
-
-        if (result[0].suspended) {
-            return null;
-        }
-
-        return result[0];
+        return result && !result.suspended ? result : null;
```
TBonnin: could the `suspended` be part of the query?

packages/shared/lib/services/user.service.ts
```diff
@@ -91,23 +83,15 @@ class UserService {
     }
 
     async getUserByEmail(email: string): Promise<User | null> {
-        const result = await db.knex.select('*').from<User>(`_nango_users`).where({ email: email });
-
-        if (result == null || result.length == 0 || result[0] == null) {
-            return null;
-        }
+        const result = await db.knex.select('*').from<User>(`_nango_users`).where({ email: email }).first();
```
TBonnin, line 82: we don't check `suspended` in this function and the one below? Is it intended?

packages/shared/lib/services/user.service.ts
```diff
@@ -169,11 +149,13 @@ class UserService {
...
-    async update({ id, ...data }: { id: number } & Omit<Partial<DBUser>, 'id'>): Promise<number> {
-        return await db.knex
+    async update({ id, ...data }: { id: number } & Omit<Partial<DBUser>, 'id'>): Promise<DBUser | null> {
+        const up = await db.knex
             .from<DBUser>(`_nango_users`)
             .update({ ...data, updated_at: new Date() })
-            .where({ id });
+            .where({ id })
+            .returning('*');
+        return up[0] || null;
```
TBonnin: you can also do 
```
const [up] = await ...
return up || null
````

File: packages/webapp/src/components/PrivateRoute.tsx [hunk 4]
```diff
@@ -71,7 +71,7 @@ export const PrivateRoute: React.FC = () => {
         }
     }, [user, identify]);
 
-    if (loading || !ready) {
+    if (loading || !ready || loadingUser) {
```
TBonnin: what's the difference between loading and loadingUser

packages/server/lib/clients/auth.client.ts
```diff
@@ -32,7 +33,7 @@ export function setupAuth(app: express.Router) {
             store: sessionStore,
             name: 'nango_session',
             unset: 'destroy',
-            cookie: { maxAge: 30 * 24 * 60 * 60 * 1000, secure: false },
+            cookie: { maxAge: 7 * 24 * 60 * 60 * 1000, secure: false },
```
TBonnin, line 36: at some point we are gonna need to review our cookie policy. In the meantime 7 days is probably a good compromise

TBonnin (approved):

packages/webapp/src/components/PrivateRoute.tsx
bodinsamuel: let me rename the other loading 👌🏻

packages/shared/lib/services/user.service.ts
bodinsamuel: true

packages/shared/lib/services/user.service.ts
bodinsamuel: good point

packages/server/lib/clients/auth.client.ts
bodinsamuel: yes, long sessions are a liability

bodinsamuel:
Thanks for the review @TBonnin it forced me to retest and realized I messed up a condition just before pushing 👌🏻

PR #2516, opened 2024-07-16
feat(webhooks): [nan-1432] add checkr webhook forwarding possibility

khaliqgant:
## Describe your changes
Support checkr webhooks. TBD on how to verify the webhook signature. [Docs](https://docs.checkr.com/#section/Webhooks/Securing-webhooks) 

NAN-1432


packages/server/lib/webhook/checkr-webhook-routing.ts
```diff
@@ -0,0 +1,48 @@
...
+
+    logger.info('received', { configId: integration.id });
+
+    if (!validate(integration, signature, rawBody)) {
+        logger.error('invalid signature', { configId: integration.id });
+        // TODO the verification should use the API key
+        //return;
+    }
+
+    const parsedBody = body as CheckrBody;
```
TBonnin, line 40: There is no parsing. Is that intended?

packages/server/lib/webhook/webhook.manager.ts
```diff
@@ -34,9 +34,17 @@ async function execute(
...
+
     const { environment, account } = environmentAndAccountLookup;
 
-    const handler = handlers[`${integration.provider.replace(/-/g, '')}Webhook`];
+    if (!template || !template['webhook_routing_script']) {
+        return;
+    }
+
+    const webhookRoutingScript = template['webhook_routing_script'];
+    const handler = handlers[webhookRoutingScript.replace(/-/g, '')];
```
TBonnin: I am personally not super fan of this kind of routing based on filename/string. It makes it very hard to navigate/grep in an IDE. Could the keys in handlers be set explicitly to at least match the providers.yaml value?

packages/server/lib/webhook/checkr-webhook-routing.ts
khaliqgant: Yeah, right now checkr is just used for forwarding

packages/server/lib/webhook/checkr-webhook-routing.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const route: WebhookHandler = async (nango, integration, headers, body, rawBody, logContextGetter: LogContextGetter) => {
+    const signature = headers['x-checkr-signature'];
+
+    logger.info('received', { configId: integration.id });
+
+    if (!validate(integration, signature, rawBody)) {
+        logger.error('invalid signature', { configId: integration.id });
+        // TODO the verification should use the API key
+        //return;
```
bodinsamuel, line 37: I'm not I get what is preventing you to return here?

bodinsamuel (approved):

packages/server/lib/webhook/checkr-webhook-routing.ts
khaliqgant: In testing there is no API key so we actually have no way to validate the webhook signature right now. Checkr will be forward only for now and that fits to the customers spec

packages/server/lib/webhook/webhook.manager.ts
khaliqgant: Fair, updated with https://github.com/NangoHQ/nango/pull/2516/commits/77155c4bb1a08d2832a6397844765aa132579c46

TBonnin (approved):

PR #2518, opened 2024-07-16
fix: update sync frequency

TBonnin:
Updating frequency from the integration UI is failing. 
The endpoint retrieves all the syncs for the given integration and attempt to update the schedule frequency in the orchestrator. 
However the query to retrieve the syncs was wrongly returning sync that didn't belong to the current user because the join between syncs and sync_configs was done only on the name which is not unique. It could have had pretty bad consequence. Fortunately the orchestrator endpoint relies on a matching environment/sync and failed to retrieve the schedule so no data was corrupted. Just the frequency would not be updated. No other code was using the faulty query.

This commit fixes the query by also joining sync and connections to only select rows where connection's environment_id matches. 
Note: We should eventually add a foreign key to join syncs and sync_configs table

https://linear.app/nango/issue/NAN-1441/update-frequency-from-the-integration-ui-doesnt-work


khaliqgant (approved):
> We should eventually add a foreign key to join syncs and sync_configs table

Sooner than later I'd say

PR #2519, opened 2024-07-16
fix(xero-integration-template): [nan-1438] xero updates to match the API better

khaliqgant:
## Describe your changes
Update the xero create-invoice to better match the API. Mappings are as follows:
```
# fees
# Xero => Nango
Quantity => units
UnitAmount => precise_unit_amount
LineAmount => amount_cents
TaxAmount => taxes_amount_cents
DiscountAmount => discount_amount_cents
DiscountRate => discount_rate
```

### DIFF
--- a/integration-templates/xero/mappers/to-invoice.ts
+++ b/integration-templates/xero/mappers/to-invoice.ts
@@ -32,8 +32,8 @@ function toInvoiceItem(xeroInvoiceItem: XeroLineItem): InvoiceFee {
         precise_unit_amount: Number(xeroInvoiceItem.UnitAmount),
         account_code: xeroInvoiceItem.AccountCode,
         account_external_id: xeroInvoiceItem.AccountId,
-        amount_cents: parseFloat(xeroInvoiceItem.LineAmount) * 100, // Amounts in xero are not in cents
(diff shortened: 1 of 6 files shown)


integration-templates/xero/mappers/to-invoice.ts
```diff
@@ -100,6 +116,14 @@ export function toXeroInvoice(invoice: CreateInvoice | Invoice) {
             xeroItem['TaxAmount'] = item.taxes_amount_cents / 100;
         }
 
+        if (item.discount_amount_cents) {
+            xeroItem['DiscountAmount'] = item.discount_amount_cents / 100;
```
TBonnin: manipulating money with arithmetic operations can lead to rounding issue. Could it be the case here?

TBonnin (approved):

integration-templates/xero/mappers/to-invoice.ts
khaliqgant: Made some adjustments to be a bit more precise https://github.com/NangoHQ/nango/pull/2519/commits/287742b4efc37af3b0b6e02abe20c8af2bd53e69. Worth noting that this logic was taken from an existing client you're familiar with that is in production

PR #2520, opened 2024-07-17
feat: invitation UI

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1384/invitation-page-ui

### Major changes
- New page to accept an invitation when already logged
It will automatically transfer the user to the team. If declined the invitation will no longer work

- Reuse Signup form and Endpoint when receiving invitation while being unlogged
The two ways to signup were already desync in term of logic, so I'm now just passing an optional invite token and adapted the logic accordingly.

### Minor changes
- Fix desync api formatting
- Fix swrfetcher error type
- Fix toast click zone preventing us to click on the user bloc

## Test

- Enabled AUTH in `detection.ts`
- Invite an email
- Get the token directly in db and go to `/signup/__TOKEN__`
- You should be able to decline or accept

*Already logged in*
Screenshot 2024-07-17 at 11 03 34

*New account*
Screenshot 2024-07-17 at 11 02 52


packages/webapp/src/pages/Account/components/SignupForm.tsx
```diff
@@ -0,0 +1,157 @@
+import { useState } from 'react';
```
bodinsamuel, line 1: Moved from Signup

packages/webapp/src/hooks/useTeam.tsx
```diff
@@ -34,38 +35,6 @@ export async function apiPutTeam(env: string, body: PutTeam['Body']) {
     }
 }
 
-export async function apiPostInvite(env: string, body: PostInvite['Body']) {
```
bodinsamuel, line 37: Forgot to move that into it's own file

packages/webapp/src/pages/Account/InviteSignup.tsx
```diff
@@ -1,182 +1,119 @@
...
-                                    <input
-                                        id="email"
-                                        name="email"
-                                        type="email"
-                                        autoComplete="email"
-                                        defaultValue={invitedEmail}
-                                        placeholder="Email"
-                                        required
-                                        readOnly={!isEnterprise()}
-                                        className={`${isEnterprise() ? '' : 'cursor-not-allowed outline-none border-transparent focus:border-transparent focus:ring-0 border-none '}bg-bg-black text-text-light-gray block h-11 focus:outline-none w-full appearance-none rounded-md px-3 py-2 text-[14px] shadow-sm`}
```
khaliqgant, line 116: How is enterprise handled now?

khaliqgant (approved):
Question inline around enterprise

packages/webapp/src/pages/Account/InviteSignup.tsx
bodinsamuel: Mmh I didn't noticed that, the only diff is that they can change the email?

packages/server/lib/controllers/v1/account/signup.ts
```diff
@@ -63,30 +61,64 @@ export const signup = asyncWrapper<Signup>(async (req, res) => {
...
     if (!user) {
-        logger.error('Error creating user');
         res.status(500).send({ error: { code: 'error_creating_user', message: 'There was a problem creating the user. Please reach out to support.' } });
         return;
     }
 
-    if (!user.email_verification_token) {
-        res.status(400).send({ error: { code: 'email_already_verified', message: 'Email address was already verified, please login.' } });
+    // Ask for email validation if not coming from an invitation
+    if (!token) {
```
TBonnin, line 102: I  would have inverted to avoid the negation but I am nitpicking ;)
```
// log directly if we are coming from an invite
if(token) {
   ...
   return
}

// Ask for email validation if not coming from an invitation
...

TBonnin (commented):
code looks good but when I am testing locally I cannot decline or accept the invite. When I click on the buttons I am getting an error. Looking at the request the `POST /api/v1/invite/TOKEN` fails with error `Invitation does not exists or is expired`. Maybe I am doing something wrong

bodinsamuel:
> Looking at the request the POST /api/v1/invite/TOKEN fails with error Invitation does not exists or is expired. Maybe I am doing something wrong

if it's not your email, it's not allowed maybe it's that

packages/webapp/src/pages/Account/InviteSignup.tsx
bodinsamuel: changed to match the condition, but I would love to know why it's like that

packages/server/lib/controllers/v1/account/signup.ts
bodinsamuel: changed 👌🏻

TBonnin:
> > Looking at the request the POST /api/v1/invite/TOKEN fails with error Invitation does not exists or is expired. Maybe I am doing something wrong
> 
> if it's not your email, it's not allowed maybe it's that

not sure what do you mean by not my email. I am running locally so I don't have an email which is probably the cause of the issue

Anyway feel free to merge. The code looks good and I trust that you have tested the functionalities thoroughly

TBonnin (approved):

PR #2524, opened 2024-07-17
chore: add start endpoint to runner

TBonnin:
## Describe your changes

First setup of refactoring jobs/runner to remove the long request when executing scripts, replacing it by a 2-way communication between job and runners
In order to avoid breakage when deploying, I am first modifying the runner. Adding a new `start` endpoint and keeping the old `run` endpoint.
For now `start` is not being called.

https://linear.app/nango/issue/NAN-1421/refactor-runservice

### DIFF
--- a/packages/runner/lib/client.unit.test.ts
+++ b/packages/runner/lib/client.unit.test.ts
@@ -7,6 +7,28 @@ describe('Runner client', () => {
     const port = 3095;
     const serverUrl = `http://localhost:${port}`;
     let client: ReturnType<typeof getRunnerClient>;
+    const nangoProps: NangoProps = {
+        host: 'http://localhost:3003',
+        connectionId: 'connection-id',
+        environmentId: 1,
+        providerConfigKey: 'provider-config-key',
+        activityLogId: 1,
+        secretKey: 'secret-key',
+        nangoConnectionId: 1,
+        syncId: 'sync-id',
+        syncJobId: 1,
+        lastSyncDate: new Date(),
+        dryRun: true,
+        attributes: {},
+        track_deletes: false,
+        logMessages: {
+            counts: { updated: 0, added: 0, deleted: 0 },
+            messages: []
+        },
+        syncConfig: {} as SyncConfig,
+        runnerFlags: {} as any,
+        stubbedMetadata: {}
+    };
 
     beforeAll(() => {
         client = getRunnerClient(serverUrl);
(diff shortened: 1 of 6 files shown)


packages/runner/lib/client.unit.test.ts
```diff
@@ -18,42 +40,19 @@ describe('Runner client', () => {
...
-        };
-        const jsCode = `
-        f = async (nango) => {
-            const s = nango.lastSyncDate.toISOString();
-            const b = Buffer.from("hello world");
-            const t = await Promise.resolve(setTimeout(() => {}, 5));
-            return [1, 2, 3]
-        };
-        exports.default = f
-        `;
```
TBonnin, line 52: move the code execution verification into exec test file

packages/runner/lib/exec.ts
```diff
@@ -13,20 +13,17 @@ import { logger } from './utils.js';
 
 export async function exec(
     nangoProps: NangoProps,
-    isInvokedImmediately: boolean,
-    isWebhook: boolean,
+    scriptType: 'sync' | 'action' | 'webhook' | 'post-connection-script',
```
TBonnin, line 16: currently in run.service we have some `isAction`, `isWebhook` booleans that are transformed into `isInvokedImmediately` and again in `isAction` in this file.
In order to be consistent and make it easy to follow, I am gonna change those booleans into a union type that is passed along the chain from jobs to runner and back to jobs

packages/runner/lib/exec.unit.test.ts
```diff
@@ -0,0 +1,41 @@
...
+            const b = Buffer.from("hello world");
+            const t = await Promise.resolve(setTimeout(() => {}, 5));
+            return [1, 2, 3]
+        };
+        exports.default = f
+        `;
+        const res = exec(nangoProps, 'sync', jsCode);
+        await expect(res).resolves.toEqual([1, 2, 3]);
+    });
+});
```
TBonnin, line 41: moved from the other test file.

packages/runner/lib/server.ts
```diff
@@ -58,14 +67,51 @@ function runProcedure() {
...
+            setImmediate(async () => {
+                try {
+                    const { error, response } = await exec(nangoProps, scriptType, code, codeParams);
+                    await httpFetch({
+                        method: 'PUT',
+                        url: `${jobsServiceUrl}/task/${taskId}`,
+                        data: superjson.stringify({
+                            scriptType,
+                            error,
+                            output: response
```
TBonnin, line 103: I am not 100% of the response schema yet. I might modify it when I implement the PUT `/task/ID` endpoint on `jobs` side

packages/runner/lib/server.ts
```diff
@@ -58,14 +67,51 @@ function runProcedure() {
                     input: codeParams
                 });
                 usage.track(nangoProps);
-                return await exec(nangoProps, input.isInvokedImmediately, input.isWebhook, code, codeParams);
+                const scriptType: 'sync' | 'action' | 'webhook' = input.isWebhook ? 'webhook' : input.isInvokedImmediately ? 'action' : 'sync';
+                return await exec(nangoProps, scriptType, code, codeParams);
             } finally {
```
khaliqgant, line 72: this type specification is a bit messy but is this temporary since it is combining the two?

khaliqgant (approved):
Non blocking question inline

packages/runner/lib/server.ts
TBonnin: The `runProcedure` and therefore the awkward scriptType transformation will go away once jobs is refactored

packages/runner/lib/exec.ts
bodinsamuel: I guess it would be nice to have this related to available tasks in orchestrator so it's synced

bodinsamuel (approved):

PR #2526, opened 2024-07-18
feat(tableau-integrations): add support for tableau api

hassan254-prog:
## Describe your changes

- Add support for tableau api
- Tableau follows a different approach from what we currently support to generate tokens, which are later used to authenticate their REST API.
- To generate these tokens you will have to send a POST request with the following body 
```
{
    "credentials": {
        "personalAccessTokenName": "token_name",
        "personalAccessTokenSecret": "token_secret",
        "site": {
            "contentUrl": "content_url"
        }
    }
}
```
Also for proxy headers, you will have to append the following field for authentication
```
X-tableau-Auth: ${token}
```
[EXT-100](https://linear.app/nango/issue/EXT-100/add-support-for-tableau)


docs-v2/integrations/all/tableau.mdx
```diff
@@ -0,0 +1,33 @@
...
+
+-   [How to generate your Personal Access Token](https://help.tableau.com/current/pro/desktop/en-us/useracct.htm#create-and-revoke-personal-access-tokens)
+-   [Auth related docs](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_concepts_auth.htm#make-a-sign-in-request-with-a-personal-access-token)
+-   [API endpoints](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api.htm)
+
+<Tip>Need help getting started? Get help in the [community](https://nango.dev/slack).</Tip>
+
+## API gotchas
+
+- When creating a new Nango connection, please supply your Token name as your PAT NAME, Token secret as your PAT SECRET, and the content URL, which is the value that appears after `/site/` in the browser address bar for Tableau Cloud. This field is left blank if you are using Tableau Server.
```
bodinsamuel: Can you provide an example?

packages/shared/lib/services/proxy.service.ts
```diff
@@ -416,6 +422,14 @@ class ProxyService {
                     };
                 }
                 break;
+            case 'TABLEAU':
+                {
+                    const token = config.token as TableauCredentials;
+                    headers = {
+                        'X-tableau-Auth': token
```
bodinsamuel, line 429: they are annoying :\

packages/frontend/lib/index.ts
```diff
@@ -488,6 +522,12 @@ interface TBACredentials {
     oauth_client_secret_override?: string;
 }
 
+interface TableauCredentials {
+    pat_name: string;
+    pat_secret: string;
```
bodinsamuel, line 527: There is no way to make that less specific to Tableau? feels like it's a regular PAT credentials with a dedicated header

bodinsamuel (commented):
Looks good, I just wonder how generic can it be

docs-v2/integrations/all/tableau.mdx
hassan254-prog: Updated this in the latest [commit](https://github.com/NangoHQ/nango/pull/2526/commits/9e893fa6ae4ee2cb36f97b5c892d8b76cb5444b4) 👍

packages/shared/lib/services/proxy.service.ts
hassan254-prog: Yes very annoying 😅

packages/frontend/lib/index.ts
hassan254-prog: I was indeed considering that angle, where we might get another provider with a similar authentication method, so as to make it less specific to Tableau. However, it appears that this is a custom `auth_mode` specific to Tableau.

khaliqgant:
@TBonnin can you review when you have a moment?

bodinsamuel (approved):

packages/shared/lib/utils/utils.ts
```diff
@@ -94,6 +94,22 @@ export function parseTokenExpirationDate(expirationDate: any): Date {
...
+    const days = Number(daysStr);
+    const hours = Number(hoursStr);
+    const minutes = Number(minutesStr);
+
+    if (isNaN(days) || isNaN(hours) || isNaN(minutes)) {
+        return;
+    }
+
+    const millisecondsToAdd = (days * 24 * 60 * 60 + hours * 60 * 60 + minutes * 60) * 1000;
+    return new Date(Date.now() + millisecondsToAdd);
```
TBonnin: can be simplified slightly I think:
```
  const [days, hours, minutes] = timeStr.split(':').map(Number);
  
  if (days && hours && minutes) {
    const milliseconds = ((days * 24 + hours) * 60 + minutes) * 60 * 1000;
    return new Date(Date.now() + milliseconds);
  }
  
  return undefined;
```

TBonnin (approved):

packages/shared/lib/utils/utils.ts
hassan254-prog: Ah, yes, that simplifies this function, updated in this latest [commit](https://github.com/NangoHQ/nango/pull/2526/commits/364d3113a7a21b0770c381f3a2702b7729d1fd86). Thank you! 🙏

PR #2528, opened 2024-07-18
fix(cli): [nan-1457] auto confirm for dry run

khaliqgant:
## Describe your changes
Honor the `--auto-confirm` option for dry run to automatically cycle through logs

NAN-1457

### DIFF
--- a/.github/workflows/sync-integration-templates.yaml
+++ b/.github/workflows/sync-integration-templates.yaml
@@ -39,3 +39,4 @@ jobs:
               env:
                   AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
                   AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+                  AWS_DEFAULT_REGION: ${{ secrets.AWS_REGION }}
--- a/packages/cli/lib/services/dryrun.service.ts
+++ b/packages/cli/lib/services/dryrun.service.ts
@@ -368,9 +368,9 @@ export class DryRunService implements ScriptExecutorInterface {
 
                 while (index < logMessages.messages.length) {
                     const remaining = logMessages.messages.length - index;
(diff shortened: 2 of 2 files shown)


.github/workflows/sync-integration-templates.yaml
```diff
@@ -39,3 +39,4 @@ jobs:
               env:
                   AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
                   AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+                  AWS_DEFAULT_REGION: ${{ secrets.AWS_REGION }}
```
TBonnin, line 42: is that part of this PR?

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -368,9 +368,9 @@ export class DryRunService implements ScriptExecutorInterface {
 
                 while (index < logMessages.messages.length) {
                     const remaining = logMessages.messages.length - index;
-                    const confirmation = await promptly.confirm(
-                        `There are ${remaining} logs messages remaining. Would you like to see the next 10 log messages? (y/n)`
-                    );
+                    const confirmation = options.autoConfirm
+                        ? true
+                        : await promptly.confirm(`There are ${remaining} logs messages remaining. Would you like to see the next 10 log messages? (y/n)`);
```
TBonnin, line 373: if there is a lot of records can it overload their terminal?

TBonnin (approved):

.github/workflows/sync-integration-templates.yaml
khaliqgant: It probably should be a part of another PR but tried to squeeze it in 😅

packages/cli/lib/services/dryrun.service.ts
khaliqgant: Maybe? Feel like that is an edge case that we can live with for now.

PR #2530, opened 2024-07-19
fix(ui): feedback on profile/team/invitation

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/SAM-11/integration-feedback

- Change wording
- Fix: Admin shown to everybody
- Change button shape
- Remove input for email
- Remove some copy button

### DIFF
--- a/packages/webapp/src/pages/Team/components/Users.tsx
+++ b/packages/webapp/src/pages/Team/components/Users.tsx
@@ -13,6 +13,10 @@ export const columns: ColumnDef<ApiUser | ApiInvitation>[] = [
         header: 'Name',
         size: 180,
         cell: ({ row }) => {
+            if ('created_at' in row.original) {
+                return '-';
+            }
+
             return <div className="truncate text-sm">{row.original.name || '-'}</div>;
         }
     },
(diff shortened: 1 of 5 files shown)


khaliqgant (approved):

packages/webapp/src/pages/Team/components/Users.tsx
```diff
@@ -13,6 +13,10 @@ export const columns: ColumnDef<ApiUser | ApiInvitation>[] = [
         header: 'Name',
         size: 180,
         cell: ({ row }) => {
+            if ('created_at' in row.original) {
+                return '-';
+            }
+
```
TBonnin, line 19: what is this for?

TBonnin (approved):

packages/webapp/src/pages/Team/components/Users.tsx
bodinsamuel: when we invite there is no name available, I was putting the email but we decided it was clearer like this 
Screenshot 2024-07-19 at 16 04 12

PR #2532, opened 2024-07-22
fix(server): add CSP, cors

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1339/[pen-test]-fix-security-misconfiguration-exposed-in-pen-test
Fixes https://linear.app/nango/issue/NAN-1340/[pen-test]-content-security-policy-csp-header-not-implemented
Fixes https://linear.app/nango/issue/NAN-1454/remediate-medium-severity-vulnerabilities-from-pen-test

- Add helmet.js to fix all notice related to SOC2 
  - CSP
  - No Sniff
  - No embed in iframe
  - HSTS
  
- Setup stricter cors
Allows all origin for public api and expose headers, restrict private api


packages/server/lib/middleware/security.ts
```diff
@@ -0,0 +1,51 @@
...
+    app.use(helmet.ieNoOpen());
+    app.use(helmet.frameguard({ action: 'sameorigin' }));
+    app.use(helmet.dnsPrefetchControl());
+    app.use(
+        helmet.hsts({
+            maxAge: 5184000
+        })
+    );
+
+    app.use(
```
khaliqgant: This intentional?

bodinsamuel: what is?

khaliqgant: nvm, misread due to whitespace, ignore!

khaliqgant (approved):
Looks good. We'll have to remember anytime we add a public facing dependency to add it to the security.ts file. Is there any way we can get notified or automate tests so we know when we're missing something?

bodinsamuel:
> Looks good. We'll have to remember anytime we add a public facing dependency to add it to the security.ts file. Is there any way we can get notified or automate tests so we know when we're missing something?

I just put it there because the code would double the length of the main file, but I don't have strong feeling about this tbh

packages/server/lib/middleware/security.ts
```diff
@@ -0,0 +1,51 @@
...
+        // == "Content-Security-Policy"
+        helmet.contentSecurityPolicy({
+            reportOnly: reportOnly === 'true' || typeof reportOnly === 'undefined',
```
TBonnin: is `!(reportOnly === 'false')` not clearer?

packages/server/lib/middleware/security.ts
```diff
@@ -0,0 +1,51 @@
+import { basePublicUrl, baseUrl } from '@nangohq/utils';
+import type { Router } from 'express';
+import helmet from 'helmet';
+
+export function securityMiddleware(app: Router): void {
```
TBonnin: can we have this function return an array of middlewares than can then be used in the router. ex:
```
app.use(securityMiddlewares)
```

TBonnin (approved):

packages/server/lib/middleware/security.ts
bodinsamuel: or `reportOnly !== 'false'` :D

packages/server/lib/middleware/security.ts
bodinsamuel: indeed

packages/server/lib/middleware/security.ts
TBonnin: :face_palm:

PR #2536, opened 2024-07-22
fix(integration-templates-xero): validate update-invoices correctly

khaliqgant:
## Describe your changes
- Input expects and array and should let the end user know that
- allow an update to not have so many required properties


integration-templates/xero/mappers/to-invoice.ts
```diff
@@ -86,45 +86,47 @@ export function toXeroInvoice(invoice: CreateInvoice | Invoice) {
         xeroInvoice['Url'] = invoice.url;
     }
 
-    for (const item of invoice.fees) {
-        const xeroItem: Record<string, any> = {
-            Description: item.description,
-            AccountCode: item.account_code
-        };
+    if (invoice.fees) {
```
TBonnin, line 89: is the `if` necessary?

TBonnin (approved):

integration-templates/xero/mappers/to-invoice.ts
khaliqgant: Compilation complains if the conditional isn't there
```
xero/mappers/to-invoice.ts:89:28 - error TS18048: 'invoice.fees' is possibly 'undefined'.

89         for (const item of invoice?.fees) {
```

TBonnin: what about ` for (const item of invoice?.fees || []) {`

PR #2540, opened 2024-07-23
feat: expose env.js from backend to frontend

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1487/expose-envjs-to-pass-env-variables-from-back-to-front

### Major changes

- Expose a fake `env.js` from backend to frontend
The goal is to be able to pass feature flags and env var dynamically before react loads. The ultimate goal is to remove the need for specific compilation to enable/disable some features and also adapt the UI based on the flags while being in sync with the backend. 
I haven't done a full pass yet but it showcase the possibilities. Once I remove the posthog and sentry keys we can stop compiling production/staging/enteprise docker images.

### Minor changes
- Fix cookie locally
I added flag to secure cookies but it's not possible to secure them in `http` or `localhost`.

- Removed some unused images
- Enable static assets caching
Not sure why it was disabled, let me know.

### 🧪 How to tests?

> NB: It's currently in staging

- Locally: set `FLAG_AUTH_ENABLED=false|true` `FLAG_MANAGED_AUTH_ENABLED=false|true` to enable/disable stuff in the UI


packages/server/lib/clients/auth.client.ts
```diff
@@ -32,7 +32,9 @@ export function setupAuth(app: express.Router) {
             store: sessionStore,
             name: 'nango_session',
             unset: 'destroy',
-            cookie: { maxAge: 7 * 24 * 60 * 60 * 1000, secure: true, path: '/', sameSite: 'none', httpOnly: true },
+            cookie: baseUrl.startsWith('https')
+                ? { maxAge: 7 * 24 * 60 * 60 * 1000, secure: true, path: '/', sameSite: 'none', httpOnly: true }
+                : { maxAge: 7 * 24 * 60 * 60 * 1000, secure: false, path: '/', httpOnly: true },
```
TBonnin, line 37: what happened locally before when secure was set to true?

packages/server/lib/controllers/v1/getEnvJs.ts
```diff
@@ -0,0 +1,25 @@
...
+        features: {
+            logs: envs.NANGO_LOGS_ENABLED,
+            scripts: flagHasScripts,
+            auth: flagHasAuth,
+            managedAuth: flagHasManagedAuth,
+            interactiveDemo: isCloud || isLocal
+        }
+    };
+    res.setHeader('content-type', 'text/javascript');
+    res.set('Cache-Control', 'public, max-age=3600');
```
TBonnin, line 23: could browser cache `env.js` for too long and use content that is stale compared to what frontend expect?

packages/server/lib/middleware/access.middleware.ts
```diff
@@ -129,20 +129,26 @@ export class AccessMiddleware {
...
                 if (err) {
-                    return next(err);
+                    res.status(500).send({ error: { code: 'server_error', message: 'failed to no-auth' } });
+                    return;
                 }
 
-                next();
+                await fillLocalsFromSession(req, res, next);
             });
             return;
```
TBonnin, line 149: can we remove this early return and the `await fillLocalsFromSession` above and let the execution of the function continue to the `fillLocalsFromSession` below?

packages/server/lib/routes.ts
```diff
@@ -107,6 +108,7 @@ const upload = multer({ storage: multer.memoryStorage() });
 router.get('/health', (_, res) => {
     res.status(200).send({ result: 'ok' });
 });
+router.get('/env.js', getEnvJs);
```
TBonnin, line 111: should we have the extension in the route path or not? I believe this is the only endpoint we are doing so

packages/server/lib/routes.ts
```diff
@@ -299,7 +301,7 @@ router.use(web);
 const webappBuildPath = '../../../webapp/build';
 const staticSite = express.Router();
 staticSite.use('/assets', express.static(path.join(dirname(), webappBuildPath), { immutable: true, maxAge: '1y' }));
-staticSite.use(express.static(path.join(dirname(), webappBuildPath), { setHeaders: () => ({ 'Cache-Control': 'no-cache, private' }) }));
+staticSite.use(express.static(path.join(dirname(), webappBuildPath), { cacheControl: true, maxAge: '1h' }));
```
TBonnin, line 305: If we don't hash static files for each deploy or prune the cache, there is a chance the browser would use stale content. Might not be an issue though

TBonnin (commented):
question about caching but the rest lgtm

> Once I remove the posthog and sentry keys we can stop compiling production/staging/enteprise docker images.

❤️

packages/server/lib/clients/auth.client.ts
bodinsamuel: before it was set to false actually

packages/server/lib/controllers/v1/getEnvJs.ts
bodinsamuel: yes that's why I append the compiled hash so actually it could be cached for longer but I don't want to take too much risk with cache

packages/server/lib/routes.ts
bodinsamuel: it's returning a javascript file not a json, so I wanted to be sure it's was highlighted, wdyt

packages/server/lib/middleware/access.middleware.ts
bodinsamuel: it's in a callback so I can't control the possible race-condition if I do that

packages/server/lib/routes.ts
bodinsamuel: yeah that's why I looked at the files, the only ones left are logo I'll probably integrate them in the code directly later. Worst case it's cached for one hour

packages/server/lib/routes.ts
TBonnin: I don't feel strongly either way, just wanted to make sure it was intended.

TBonnin (approved):

bodinsamuel:
Im going to go ahead and merge that

PR #2541, opened 2024-07-23
feat: in product metrics

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1466/empty-state-when-elasticsearch-is-not-present
Fixes https://linear.app/nango/issue/NAN-1467/empty-state-when-there-is-no-data
Fixes https://linear.app/nango/issue/NAN-1465/endpointquery-to-get-metrics
Fixes https://linear.app/nango/issue/NAN-1464/code-the-ui

- New endpoint `POST /api/v1/logs/insights` 
Retrieve insights by operations type, depending on the performance I might add some Redis cache.

- Dashboard homepage
The UI now displays the homepage by default, if you have finished your interactive demo.

## Test

> Currently deployed in staging 

Screenshot 2024-07-25 at 10 25 11


packages/webapp/src/components/TopNavBar.tsx
```diff
@@ -1,15 +1,15 @@
...
 import Info from './ui/Info';
+import { useUser } from '../hooks/useUser';
 
 export default function NavBar() {
     const signout = useSignout();
-    const email = useStore((state) => state.email);
+    const { user: me } = useUser();
     const isHNDemo = useMemo(() => {
-        return Boolean(email.match(/demo-[a-z0-9]+@example.com/));
-    }, [email]);
```
bodinsamuel, line 12: Missed this use of store email in a prev PR

packages/logs/lib/models/operations.ts
```diff
@@ -0,0 +1,50 @@
+import type { estypes } from '@elastic/elasticsearch';
```
TBonnin: why not calling this file `insights.ts` instead of the very generic name `operations`

packages/logs/lib/models/operations.ts
```diff
@@ -0,0 +1,50 @@
...
+    });
+
+    const agg = res.aggregations!['histogram'];
+
+    const computed: InsightsHistogramEntry[] = [];
+    for (const item of agg.buckets as any[]) {
+        const success = (item.state_agg.buckets as { key: string; doc_count: number }[]).find((i) => i.key === 'success');
+        const failure = (item.state_agg.buckets as { key: string; doc_count: number }[]).find((i) => i.key === 'failed');
+        computed.push({ key: item.key_as_string, total: item.doc_count, success: success?.doc_count || 0, failure: failure?.doc_count || 0 });
+    }
```
TBonnin: there are lots of type coercion here. Is there any way to compute without forcing the hand of the compiler?

TBonnin (approved):
Very nice.

I wonder how it might affect ES perf if a lot of users are loading the homepage.

From a feature point of view, like I said on Slack, I would love to be able to click on the graph to go the the logs with filter pre-set

packages/logs/lib/models/operations.ts
bodinsamuel: I wanted to split operations and messages, but I can indeed isolate this part even more 👌🏻

packages/logs/lib/models/operations.ts
bodinsamuel: yeah that's annoying, in the ES defs everything is optional and everything can be a string or an array all the time, plus in that case sub aggregations are not even typed. Happy to get some help

bodinsamuel:
> I wonder how it might affect ES perf if a lot of users are loading the homepage.

Yup, I don't want to over-engineered until it's needed but I'm sure I'll add some cache at some point

> From a feature point of view, like I said on Slack, I would love to be able to click on the graph to go the the logs with filter pre-set

Added, a global click on the card and each bar filtered on the day

packages/logs/lib/models/operations.ts
TBonnin: yeah Iooked at the ES library and type like `'number' | 'string' | 'MyInvolvedBucketType'` doesn't make things easy

TBonnin:
> Added, a global click on the card and each bar filtered on the day

Thank you. Being able to also have the status (error, success) pre-filtered would be great as well. Not sure it is possible to detect click on section of the histogram though

PR #2546, opened 2024-07-24
fix(ci): providers validation

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1497/providersyaml-is-not-correctly-validated-against

After a user reported a typo I realized this validation was letting a lot of stuff go unnoticed

- Rewrote the script in javascript
I have nothing against python but it's a bit inconvenient to have this (and python installs are notoriously painful)

- Added missing fields in the schema
- Fix few errors in providers.yaml


packages/shared/providers.yaml
```diff
@@ -413,7 +413,7 @@ canny:
     proxy:
         base_url: https://canny.io/api/v1
         query:
-            apiKey: ${apiKey}
+            api_key: ${apiKey}
```
hassan254-prog: @bodinsamuel , this might break for users using canny, as the provider requires this specific [field](https://developers.canny.io/api-reference#intro): `apiKey=YOUR_API_KEY`

bodinsamuel: ah I see it's a free form field thanks 👌🏻

packages/shared/providers.yaml
```diff
@@ -534,7 +533,6 @@ close:
         - offline_access
     proxy:
         base_url: https://api.close.com/api
-        docs: https://developer.close.com/
```
TBonnin, line 537: why removing this?

TBonnin (approved):

khaliqgant:
Shouldn’t the Python validation file be removed then?

packages/shared/providers.yaml
bodinsamuel: it's a leftover when we changed to point to our doc

bodinsamuel:
> Shouldn’t the Python validation file be removed then?

🤦🏻 done

PR #2547, opened 2024-07-24
chore: refactor runner and run.service

TBonnin:
## Describe your changes

First I would like to apologize for the size of the PR. 

The goal of this big PR is to make the communication between `jobs` and `runner` asynchronous, instead of relying on a long running request that can fail and depends on `jobs` in-memory state.

The cancellation/aborting mechanism had to be refactored also since it was also relying on an in-memory state in `jobs`

The `run.service` logic has been split into one file per script type. It is not the most DRY solution but it avoids the current situation where it is very hard to follow what part of run.service applies to which script type. They all differ slightly from each other and unless we have a good abstraction around them it is better imho to have a bit of duplication. Let me know what you think.

Notes:
- server/runner/orchestrator must be deployed first before jobs

https://linear.app/nango/issue/NAN-1421/refactor-runservice


packages/cli/lib/services/dryrun.service.ts
```diff
@@ -499,12 +504,7 @@ export class DryRunService implements ScriptExecutorInterface {
                     return { success: false, error: err, response: null };
                 }
 
-                let errorType = 'sync_script_failure';
-                if (isWebhook) {
```
bodinsamuel, line 505: Webhooks can't be dryrun?

packages/shared/lib/sdk/sync.integration.test.ts
```diff
@@ -37,10 +37,13 @@ describe('Connection service integration tests', () => {
             }
 
             const nangoProps: NangoProps = {
+                scriptType: 'sync',
                 host: 'http://localhost:3003',
-                accountId: environment.account_id,
+                teamId: environment.account_id,
+                teamName: 'team',
```
bodinsamuel: Could have been great to pass `team: Pick<id, name>` instead

packages/jobs/lib/routes/tasks/putTask.ts
```diff
@@ -0,0 +1,122 @@
...
+    Error: ApiError<'put_task_failed'>;
+    Success: never;
+}>;
+
+const jsonLiteralSchema = z.union([z.string(), z.number(), z.boolean(), z.null()]);
+export const jsonSchema: z.ZodType<JsonValue> = z.lazy(() => z.union([jsonLiteralSchema, z.array(jsonSchema), z.record(jsonSchema)]));
+
+const validate = validateRequest<PutTask>({
+    parseBody: (data) =>
+        z
```
bodinsamuel, line 88: I'm wondering what's the overhead of re-creating this zod object all the time

packages/webhooks/lib/sync.ts
```diff
@@ -80,7 +80,7 @@ export const sendSync = async ({
                 deleted: 0
             },
             modifiedAfter: dayjs(now).toDate().toISOString(),
-            queryTimeStamp: operation !== 'INITIAL' ? (now as unknown as string) : null
+            queryTimeStamp: now as unknown as string // Deprecated
```
khaliqgant, line 83: does this need to change? Can you go through why this is needed?

TBonnin: no sync is ever set as `INITIAL` anymore so this condition is always true and therefore unnecessary

packages/jobs/lib/scripts/operations/utils/getRunner.ts
```diff
@@ -0,0 +1,12 @@
...
+import type { Runner } from '../../../runner/runner.js';
+import { getOrStartRunner, getRunnerId } from '../../../runner/runner.js';
+
+export async function getRunner(teamId: number): Promise<Result<Runner>> {
+    // a runner per account in prod only
+    const runnerId = isProd ? getRunnerId(`${teamId}`) : getRunnerId('default');
+    // fallback to default runner if account runner isn't ready yet
+    const runner = await getOrStartRunner(runnerId).catch(() => getOrStartRunner(getRunnerId('default')));
+    return Ok(runner);
+}
```
TBonnin: It is a util function because it is called by both `start` and `abort`

packages/jobs/lib/routes/tasks/taskId/postHeartbeat.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const validate = validateRequest<PostHeartbeat>({
+    parseParams: (data) => z.object({ taskId: z.string().uuid() }).strict().parse(data)
+});
+
+const handler = async (req: EndpointRequest<PostHeartbeat>, res: EndpointResponse<PostHeartbeat>) => {
+    const { taskId } = req.params;
+    const heartbeat = await orchestratorClient.heartbeat({ taskId });
+    if (heartbeat.isErr()) {
+        res.status(400).json({ error: { code: 'heartbeat_failed', message: `heartbeat failed: taskId=${taskId} error='${heartbeat.error}'` } });
```
bodinsamuel: Note the json for error contains a payload/errors properties that you can use

packages/jobs/lib/routes/tasks/taskId/postHeartbeat.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
```
bodinsamuel, line 31: Missing return

packages/jobs/lib/execution/operations/output.ts
```diff
@@ -0,0 +1,83 @@
...
+        case 'webhook':
+            await handleWebhookOutput({ nangoProps });
+            break;
+        case 'post-connection-script':
+            await handlePostConnectionOutput({ nangoProps });
+            break;
+    }
+    const setSuccess = await orchestratorClient.succeed({ taskId, output: output });
+    if (setSuccess.isErr()) {
+        handlePayloadTooBigError({ taskId, error: setSuccess.error, nangoProps });
```
bodinsamuel:
```suggestion
        await handlePayloadTooBigError({ taskId, error: setSuccess.error, nangoProps });
```

packages/jobs/lib/routes/tasks/putTask.ts
```diff
@@ -0,0 +1,122 @@
...
+    if (!nangoProps) {
+        res.status(400).json({ error: { code: 'put_task_failed', message: 'missing nangoProps' } });
+        return;
+    }
+    if (error) {
+        await handleError({ taskId, nangoProps, error });
+    } else {
+        await handleOutput({ taskId, nangoProps, output: output });
+    }
+    res.status(201).send();
```
bodinsamuel: 200 seems more correct

packages/runner/lib/server.ts
```diff
@@ -92,18 +96,31 @@ function startProcedure() {
             // executing in the background and returning immediately
             // sending the result to the jobs service when done
             setImmediate(async () => {
+                const heartbeat = setInterval(async () => {
+                    await httpFetch({
+                        method: 'POST',
+                        url: `${jobsServiceUrl}/tasks/${taskId}/heartbeat`
+                    });
+                }, heartbeatIntervalMs);
```
bodinsamuel, line 104: On v1 I think it's fine, but could probably be safer on a different thread to avoid missing a beat

packages/orchestrator/lib/abort.ts
```diff
@@ -0,0 +1,39 @@
+import type { Scheduler, Task } from '@nangohq/scheduler';
+import { validateTask } from './clients/validate.js';
+import { Ok, stringifyError } from '@nangohq/utils';
+import type { Result } from '@nangohq/utils';
+import { logger } from './utils.js';
+
+export async function scheduleAbortTask({ scheduler, task }: { scheduler: Scheduler; task: Task }): Promise<Result<Task | undefined>> {
+    const aborted = validateTask(task);
+    if (aborted.isOk()) {
```
bodinsamuel: You are not rejecting if the validation doesn't pass?

packages/jobs/lib/routes/postIdle.ts
```diff
@@ -0,0 +1,45 @@
...
+            .object({ runnerId: z.string().min(1), idleTimeMs: z.number().positive() })
+            .strict()
+            .parse(data)
+});
+
+const handler = async (req: EndpointRequest<PostIdle>, res: EndpointResponse<PostIdle>) => {
+    try {
+        logger.info(`[runner ${req.body.runnerId}]: idle for ${req.body.idleTimeMs}ms. Suspending...`);
+        await suspendRunner(req.body.runnerId);
+        return res.status(201).json({ status: 'ok' });
```
bodinsamuel: Also 200?

packages/jobs/lib/execution/action.ts
```diff
@@ -0,0 +1,221 @@
...
+            activityLogId: task.activityLogId,
+            runTime: 0,
+            error,
+            environment: { id: task.connection.environment_id, name: environment?.name || 'unknown' },
+            ...(account?.id && account?.name ? { team: { id: account.id, name: account.name } } : {})
+        });
+        return Err(error);
+    }
+}
+export async function handleActionOutput({ nangoProps }: { nangoProps: NangoProps }): Promise<void> {
```
bodinsamuel: I would have named handle*Success to be consistent with handle*Error

bodinsamuel (commented):
Looks good, tested locally to sync run/cancel, action run.
- Some integration test on the API would be great
- I had multiple times `hasn't started after 5000ms` maybe the `waitForRunner` is not reactive enough

Few questions, but seems like it's mostly good

packages/cli/lib/services/dryrun.service.ts
khaliqgant: no, they can't really but might a nice addition in the future

packages/jobs/lib/routes/tasks/putTask.ts
TBonnin: good point. We don't have to recreate it every time. will change

packages/jobs/lib/routes/tasks/taskId/postHeartbeat.ts
TBonnin: sorry. what do you mean?

packages/jobs/lib/routes/tasks/putTask.ts
TBonnin: actually a 204 no content might even be better

packages/runner/lib/server.ts
TBonnin: indeed. However the heartbeat timeout is much bigger than the hearbeat interval so we can miss a lot of them without consequences

packages/orchestrator/lib/abort.ts
TBonnin: you are right. don't remember why i did that but that doesn't seem correct. will change to return an error if it doesn't validate

packages/jobs/lib/execution/action.ts
TBonnin: good idea

packages/shared/lib/sdk/sync.integration.test.ts
TBonnin: let me see. if this is feasible right now

TBonnin: done

bodinsamuel:
Ping me when you want a last review 👌🏻

bodinsamuel (approved):
🎸

PR #2548, opened 2024-07-25
fix(signup): correctly verify email when using invitation

bodinsamuel:
## Describe your changes

- Set email as verified when signing up with an invitation
- Clarify method signature

### DIFF
--- a/packages/server/lib/controllers/v1/account/managed/getCallback.ts
+++ b/packages/server/lib/controllers/v1/account/managed/getCallback.ts
@@ -101,7 +101,12 @@ export const getManagedCallback = asyncWrapper<GetManagedCallback>(async (req, r
         }
 
         // Create a user
-        user = await userService.createUser(authorizedUser.email, name, '', '', accountId);
+        user = await userService.createUser({
+            email: authorizedUser.email,
+            name,
+            account_id: accountId,
+            email_verified: true
+        });
         if (!user) {
             res.status(500).send({ error: { code: 'error_creating_user', message: 'There was a problem creating the user. Please reach out to support.' } });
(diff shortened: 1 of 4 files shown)


packages/shared/lib/services/user.service.ts
```diff
@@ -95,14 +95,21 @@ class UserService {
...
+        account_id,
+        email_verified
+    }: {
+        email: string;
+        name: string;
+        hashed_password?: string;
+        salt?: string;
+        account_id: number;
+        email_verified: boolean;
+    }): Promise<User | null> {
```
TBonnin, line 112: the _'if more than 1 parameter, we better use a struct`_ principle strikes again

packages/server/lib/controllers/v1/account/managed/getCallback.ts
```diff
@@ -101,7 +101,12 @@ export const getManagedCallback = asyncWrapper<GetManagedCallback>(async (req, r
         }
 
         // Create a user
-        user = await userService.createUser(authorizedUser.email, name, '', '', accountId);
+        user = await userService.createUser({
+            email: authorizedUser.email,
+            name,
+            account_id: accountId,
+            email_verified: true
```
TBonnin, line 108: I guess this line is the fix

TBonnin (approved):

packages/server/lib/controllers/v1/account/managed/getCallback.ts
bodinsamuel: no actually it was true by default, the actual fix is on signup as it was set as false

PR #2550, opened 2024-07-25
feat(integrations): [nan-1195] teams integration

khaliqgant:
## Describe your changes
- Adds botbuilder as a runner dep. botbuilder is used to handle the authentication and messaging for a teams bot. 
- Adds post connection for teams
- Adds ability to handle webhooks from teams

NAN-1195


TBonnin (commented):
I am personally against adding a library to the script runtime that is just for a specific  integration and/or a specific customer. 
Every lib we add will be for us to keep up to date. It is increasing the size of the CLI and runner bundle. botbuilder is 78M btw. It also increases the attack surface area.

If for commercial reasons we must do it, fine but from a technical point of view I don't think this is a good idea.

cc @bastienbeurier @bodinsamuel

khaliqgant:
> I am personally against adding a library to the script runtime that is just for a specific integration and/or a specific customer. Every lib we add will be for us to keep up to date. It is increasing the size of the CLI and runner bundle. botbuilder is 78M btw. It also increases the attack surface area.
> 
> If for commercial reasons we must do it, fine but from a technical point of view I don't think this is a good idea.
> 
> cc @bastienbeurier @bodinsamuel

Few things:

1. If we forgo allowing dry run for this action we can not add it to the floor as a dependency. I’m fairly certain the customer won’t test this using dry run but rather the deployed version 
2. There is a looks like limited REST api which might cover the needed use cases but absolutely won’t have the bandwidth to rip this out for the REST implementation but we can mark it for a later improvement 
3. Conditional runner package installations? We don’t have any kind of setup in place for this but throwing out another idea

packages/server/lib/webhook/microsoft-teams-webhook-routing.ts
```diff
@@ -0,0 +1,10 @@
+import type { WebhookHandler } from './types.js';
+import type { LogContextGetter } from '@nangohq/logs';
+
+const route: WebhookHandler = async (nango, integration, _headers, body, _rawBody, logContextGetter: LogContextGetter) => {
+    const response = await nango.executeScriptForWebhooks(integration, body, 'type', 'channelData.tenant.id', logContextGetter, 'tenantId');
```
bodinsamuel, line 5: there is no signature verification?

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -14,6 +14,7 @@ import * as vm from 'node:vm';
 import * as url from 'url';
 import * as crypto from 'crypto';
 import * as zod from 'zod';
+import * as botbuilder from 'botbuilder';
```
bodinsamuel: I might not have understood something, you are enabling this dep but not saying why we need botbuilder since it seems unused?

bodinsamuel (commented):
Agreed with Thomas but most importantly I don't know the reasons

packages/server/lib/webhook/microsoft-teams-webhook-routing.ts
khaliqgant: Not that I could find. No signature in the header and no docs that I saw

packages/cli/lib/services/dryrun.service.ts
khaliqgant: I’m saying it’s needed in the runner but probably no one will use it for dry run so it might not be needed in the cli

khaliqgant:
> Agreed with Thomas but most importantly I don't know the reasons

Added a snippet in the description. I linked to the destination customer code in slack https://nangohq.slack.com/archives/C067NRBH1LL/p1721935746329009?thread_ts=1721935156.927499&channel=C067NRBH1LL&message_ts=1721935746.329009

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -425,7 +424,7 @@ export class DryRunService implements ScriptExecutorInterface {
                             case 'zod':
                                 return zod;
                             case 'botbuilder':
-                                return botbuilder;
+                                throw new Error('Module "botbuilder" is not allowed. Please test the integration using the Nango dashboard');
```
bodinsamuel, line 427: it's probably already covered by `ALLOWED_IMPORTS`

bodinsamuel (approved):

PR #2556, opened 2024-07-29
fix(proxy): send no data if empty body

TBonnin:
Dropbox API is very strict about receiving empty body and will fail if passing an empty object {}. I am gonna blame express body parser for returning a empty object when request has no body. 
This commit ensures that the empty body at all is send if data is empty. The drawback is that it won't be possible to actually send an empty object when this PR is merged. 

https://linear.app/nango/issue/NAN-1423/proxy-do-not-default-body-to-if-no-payload-is-provided-by-user

### DIFF
--- a/packages/server/lib/controllers/proxy.controller.ts
+++ b/packages/server/lib/controllers/proxy.controller.ts
@@ -334,7 +334,8 @@ class ProxyController {
                 headers,
                 decompress
             };
-            if (data && ['POST', 'PUT', 'PATCH', 'DELETE'].includes(method)) {
+            const nonEmpty = (v: unknown) => v != null && v !== '' && (typeof v !== 'object' || Object.keys(v).length > 0);
+            if (nonEmpty(data) && ['POST', 'PUT', 'PATCH', 'DELETE'].includes(method)) {
                 requestConfig.data = data;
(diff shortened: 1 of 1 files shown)


bodinsamuel (commented):
you have access to the raw body to be able to differentiate empty vs undefined `req.rawBody`

TBonnin:
> you have access to the raw body to be able to differentiate empty vs undefined `req.rawBody`

oh yeah. Forgot about that. Thank you. Will modify the PR

TBonnin:
@bodinsamuel modifed the PR to have proxy controller forward the raw body instead of the parsed body https://github.com/NangoHQ/nango/pull/2556/commits/bb6d1ad432e692306d57bf578025c19677b6cbbe

bodinsamuel (approved):
Seems correct, I wonder if that will also help with the file upload later 👀

khaliqgant (commented):
Just making sure you tested this with different cases as this is a change that impacts a lot.

TBonnin:
> Just making sure you tested this with different cases as this is a change that impacts a lot.

Yes, let me also put this change behind a flag so it can easily be enabled/disabled.

PR #2558, opened 2024-07-29
fix(jobs): putTask error payload is not always an Record<string, unknown>

TBonnin:
Some erroring tasks are not completing because the call to jobs PUT /tasks/... is failing with a zod valiation error:
```
Error: PUT http://jobs-production/tasks/0190feed-d5c6-72c9-bce0-4bc3e2708c43: status=400, response={"error":{"code":"invalid_request","errors":[{"code":"invalid_type","message":"Expected object, received array","path":["error","payload"]}]}}
```

Ideally we would make sure the payload sent is of correct type but it is easier to be more flexible on Jobs side since one of my next project is to improve the errors logs and I will probably modify the error handling in the runner.

### DIFF
--- a/packages/jobs/lib/routes/tasks/putTask.ts
+++ b/packages/jobs/lib/routes/tasks/putTask.ts
@@ -20,7 +20,7 @@ type PutTask = Endpoint<{
         error?:
             | {
                   type: string;
-                  payload: Record<string, unknown>; //TODO: can be an array?
+                  payload: Record<string, unknown>;
                   status: number;
               }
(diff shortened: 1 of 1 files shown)


khaliqgant (approved):

bodinsamuel (approved):
Not sure I get the zod transformation, trusting you 👌🏻

TBonnin:
> Not sure I get the zod transformation, trusting you 👌🏻

basically transorming any unknown value to `{ message: theUnknownValue}`

PR #2560, opened 2024-07-29
fix: sync/status type must be INCREMENTAL or INITIAL

TBonnin:
reverting a breaking change: /sync/status type must be INCREMENTAL or INITIAL instead of INCREMENTAL or FULL

The breaking change is due to 2 recent modifications:
1. when removing Temporal, `type` was changed to be INCREMENTAL or FULL
2. Sync Config sync_type is now lowercase

The 2nd change should probably be fixed as well but I am not sure of the consequences. For now just fixing the endpoint to  unblock customers and revert the breaking change

### DIFF
--- a/packages/jobs/lib/execution/sync.ts
+++ b/packages/jobs/lib/execution/sync.ts
@@ -70,7 +70,7 @@ export async function startSync(task: TaskSync, startScriptFn = startScript): Pr
         team = accountAndEnv.account;
         environment = accountAndEnv.environment;
 
-        syncType = syncConfig.sync_type === SyncType.INCREMENTAL && lastSyncDate ? SyncType.INCREMENTAL : SyncType.FULL;
+        syncType = syncConfig.sync_type?.toLowerCase() === SyncType.INCREMENTAL.toLowerCase() && lastSyncDate ? SyncType.INCREMENTAL : SyncType.FULL;
 
         logCtx = await logContextGetter.create(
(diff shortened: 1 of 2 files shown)


khaliqgant (approved):
Thanks.

> Sync Config sync_type is now lowercase

Yeah, I noticed that as well. It was an inadvertent change I believe introduced a few weeks ago. Shouldn't have a big impact as the small bugs it introduced should have been resolved by now I believe.

packages/jobs/lib/execution/sync.ts
```diff
@@ -70,7 +70,7 @@ export async function startSync(task: TaskSync, startScriptFn = startScript): Pr
         team = accountAndEnv.account;
         environment = accountAndEnv.environment;
 
-        syncType = syncConfig.sync_type === SyncType.INCREMENTAL && lastSyncDate ? SyncType.INCREMENTAL : SyncType.FULL;
+        syncType = syncConfig.sync_type?.toLowerCase() === SyncType.INCREMENTAL.toLowerCase() && lastSyncDate ? SyncType.INCREMENTAL : SyncType.FULL;
```
TBonnin, line 73: Sync config sync_type is now lowercase https://github.com/NangoHQ/nango/blob/7c376b44ec2775c86da11bb281b941df8c49c54f/packages/types/lib/nangoYaml/index.ts#L2 We should probably fix this but I don't want to introduce any regression until I checked with @bodinsamuel For now just fixing the SyncJob.

PR #2561, opened 2024-07-29
fix: webhook syncType must be INITIAL instead of FULL

TBonnin:
following up on https://github.com/NangoHQ/nango/pull/2560 which was reverting a breaking change of /sync/status endpoint.
 This commit is fixing the same issue in webhook body


khaliqgant (approved):
Thanks for following up.

bodinsamuel (approved):

PR #2567, opened 2024-07-30
fix(runner): serialize error correctly

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1529/error-serialization-is-broken-in-runnerjobs

- Correctly serialize error from runner to jobs
I already had fixed this a few weeks ago but the recent migration broke it again. Hopefully, at some point, we'll have some e2e tests. The main idea is to keep looking for known patterns and fallback to std error.
I also ejected from `formatScriptError` since dryrun is already too different vs regular exec. At some point we'll need to find a way to use the same code.

### Test

```ts

export default async function runAction(nango: NangoSync): Promise<void> {
    throw new Error('foobar');
    throw new nango.ActionError('foobar');
}
```


khaliqgant (approved):

packages/jobs/lib/execution/operations/utils/errors.ts
```diff
@@ -0,0 +1,27 @@
+import { NangoError, deserializeNangoError } from '@nangohq/shared';
+import { stringifyError } from '@nangohq/utils';
+
+export function getFormattedScriptError({ err, defaultErrorType, scriptName }: { err: unknown; defaultErrorType: string; scriptName: string }) {
+    if (typeof err !== 'object') {
+        return new NangoError('invalid_error', err);
+    }
+    if (!err) {
+        return new NangoError('invalid_error', null as any);
+    }
```
TBonnin, line 10: I wish typescript had a `something` type representing an existing value that could not be null or undefined

packages/jobs/lib/execution/operations/utils/errors.ts
```diff
@@ -0,0 +1,27 @@
+import { NangoError, deserializeNangoError } from '@nangohq/shared';
+import { stringifyError } from '@nangohq/utils';
+
+export function getFormattedScriptError({ err, defaultErrorType, scriptName }: { err: unknown; defaultErrorType: string; scriptName: string }) {
```
TBonnin: let's specify the return type

TBonnin: also would something like `toNangoError(...)` or `asNangoErro()` be a more descriptive name?

packages/runner/lib/exec.ts
```diff
@@ -167,13 +170,38 @@ export async function exec(
                         },
                         response: null
                     };
+                } else {
+                    const tmp = errorToObject(error);
+                    return {
+                        success: false,
+                        error: {
+                            type: 'http_error',
+                            payload: { name: tmp.name || 'Error', code: tmp.code, message: tmp.message || 'no message' },
```
TBonnin: is 'no message' better than an empty string?

packages/shared/lib/utils/error.ts
```diff
@@ -622,6 +622,21 @@ export class NangoError extends Error {
                 this.message = `The sync job results could not be updated: ${this.payload}`;
                 break;
 
+            case 'invalid_error':
+                this.status = 500;
+                this.message = `An invalid error of type: ${typeof this.payload}`;
+                break;
+
+            case 'http_error':
```
TBonnin: should this error prefixed with `script_` since it only applies in the context of a script?

TBonnin (commented):
small comments but no dealbreaker. lgtm

TBonnin (approved):

PR #2568, opened 2024-07-30
fix(webhook): [nan-1528] operation should reflect the sync job run type not the type of sync it is

khaliqgant:
## Describe your changes
Right now the webhook uses the `sync_type` as specified in the nango.yaml:
```
syncs:
   some-sync:
       sync_type: full
```
but instead it should reflect the type of job run it last peformed. 

NAN-1528


bodinsamuel (commented):
Not sure I agree, lastSyncDate is always set even if you dont use it. Why would not use what was set in the nangoyaml?

khaliqgant:
> Not sure I agree, lastSyncDate is always set even if you dont use it. Why would not use what was set in the nangoyaml?

https://github.com/NangoHQ/nango/blob/master/packages/shared/lib/clients/orchestrator.ts#L554 last sync date is not set on a full refresh.

> Why would not use what was set in the nangoyaml?

This is more so re-instating legacy behavior that customers use already. That being said, an `INITIAL` sync is the first time a full sync is run, while an `INCREMENTAL` sync are the sync runs that are performed after the initial one.

In this customers case, they have a `sync_type: full` that is running and every time the sync webhook is outputting an `INITIAL` sync which isn't correct. https://nangohq.slack.com/archives/C05FUKTV1RP/p1722357464652009?thread_ts=1722273791.615089&cid=C05FUKTV1RP

TBonnin:
> In this customers case, they have a `sync_type: full` that is running and every time the sync webhook is outputting an `INITIAL` sync which isn't correct. https://nangohq.slack.com/archives/C05FUKTV1RP/p1722357464652009?thread_ts=1722273791.615089&cid=C05FUKTV1RP

Shouldn't a full sync always be `INITIAL`? why would it be INCREMENTAL? 

If this was the previous behaviour fine but it is very confusing then.

TBonnin (approved):
i am ok bringing back previous behavior even though I don't understand it.

Thank you @khaliqgant for taking care of the issue

khaliqgant:
> > In this customers case, they have a `sync_type: full` that is running and every time the sync webhook is outputting an `INITIAL` sync which isn't correct. https://nangohq.slack.com/archives/C05FUKTV1RP/p1722357464652009?thread_ts=1722273791.615089&cid=C05FUKTV1RP
> 
> Shouldn't a full sync always be `INITIAL`? why would it be INCREMENTAL?
> 
> If this was the previous behaviour fine but it is very confusing then.

A full sync should always be `FULL` but not `INITIAL`. In this case I see `INCREMENTAL` to mean "subsequent" syncs whereas `INITIAL` is simply the first time the sync is run so that would apply when the sync is first run or if the sync is reset using the syn command `FULL_SYNC`

TBonnin:
> A full sync should always be `FULL` but not `INITIAL`. In this case I see `INCREMENTAL` to mean "subsequent" syncs whereas `INITIAL` is simply the first time the sync is run so that would apply when the sync is first run or if the sync is reset using the syn command `FULL_SYNC`

I am glad you got to the issue before me because I hadn't understood the legacy behavior.
It is very confusing that `sync_type: INCREMENTAL` in nango.yaml means the sync only fetches data that has changed but `syncType: INCREMENTAL` in webhook could mean it is not the first execution of a `sync_type: FULL` 🤯

khaliqgant:
> > A full sync should always be `FULL` but not `INITIAL`. In this case I see `INCREMENTAL` to mean "subsequent" syncs whereas `INITIAL` is simply the first time the sync is run so that would apply when the sync is first run or if the sync is reset using the syn command `FULL_SYNC`
> 
> I am glad you got to the issue before me because I hadn't understood the legacy behavior. It is very confusing that `sync_type: INCREMENTAL` in nango.yaml means the sync only fetches data that has changed but `syncType: INCREMENTAL` in webhook can it is not the first execution of a `sync_type: FULL`

I agree that it might be confusing now but it is more because `sync_type` was something added on later so the confusion/complexity comes from adding stuff on top of existing behavior. In this case the customer is a bit more established so built their logic based on logic established early on in Nango.

PR #2569, opened 2024-07-30
feat(anrok-integrations): [nan-1527] add anrok actions

khaliqgant:
## Describe your changes
Add anrok integrations https://apidocs.anrok.com/#tag/Transactions

NAN-1527


integration-templates/anrok/actions/create-ephemeral-transaction.ts
```diff
@@ -0,0 +1,66 @@
...
+import type { AnrokResponse } from '../types';
+import { mapFees } from '../mappers/fees.js';
+
+export default async function runAction(nango: NangoAction, rawInput: Transaction[]): Promise<TransactionActionResponse> {
+    const response: TransactionActionResponse = {
+        succeeded: [],
+        failed: []
+    };
+
+    const input = Array.isArray(rawInput) ? rawInput : [rawInput];
```
TBonnin, line 12: would the input validation not fail if `rawInput` is actually not an array?

integration-templates/anrok/actions/create-ephemeral-transaction.ts
```diff
@@ -0,0 +1,66 @@
...
+            },
+            lineItems: transaction.fees.map((fee) => ({
+                id: fee.item_id,
+                productExternalId: fee.item_code || '',
+                amount: fee.amount_cents || 0
+            }))
+        };
+
+        if (transaction.contact.taxable && transaction.contact.tax_number) {
+            anrokTransaction.customerTaxIds = [];
```
TBonnin: why assigning an empty array and then pushing a value into it instead of assigning an array with the value?

integration-templates/anrok/actions/void-transaction.ts
```diff
@@ -0,0 +1,32 @@
...
+                headers: {
+                    'Content-Type': 'application/json'
+                },
+                data: {}
+            })
+            .then(() => {
+                const successTransaction = {
+                    ...transaction
+                };
+                response.succeeded.push(successTransaction);
```
TBonnin: can directly do `response.succeeded.push(transaction);`

TBonnin (approved):
one question about accepting non array input and validation but the rest lgtm

integration-templates/anrok/actions/create-ephemeral-transaction.ts
khaliqgant: it would when validation is active but it isn't blocking currently, just warns

integration-templates/anrok/actions/create-ephemeral-transaction.ts
khaliqgant: Unsure tbh, didn't write this part, but i'll update

PR #2574, opened 2024-07-31
fix(oauth2_cc): accommodate expires_in for fiserv

hassan254-prog:
## Describe your changes
Fiserv returns `expires_in` in milliseconds, small follow up from [2573](https://github.com/NangoHQ/nango/pull/2573)
[EXT-111](https://linear.app/nango/issue/EXT-111/add-support-for-fiserv)


packages/shared/lib/services/connection.service.ts
```diff
@@ -821,10 +821,12 @@ class ConnectionService {
 
                 let expiresAt: Date | undefined;
 
+                //fiserv returns expires_in in milliseconds
                 if (rawCreds['expires_at']) {
                     expiresAt = parseTokenExpirationDate(rawCreds['expires_at']);
                 } else if (rawCreds['expires_in']) {
-                    expiresAt = new Date(Date.now() + Number.parseInt(rawCreds['expires_in'], 10) * 1000);
+                    const expiresIn = Number.parseInt(rawCreds['expires_in'], 10);
+                    expiresAt = new Date(Date.now() + (provider === 'fiserv' ? expiresIn : expiresIn * 1000));
```
TBonnin: not a huge fan of making our codebase more complex to handle a bug from a provider. The oauth RFC clearly indicates that expires_in should be in seconds https://www.rfc-editor.org/rfc/rfc6749#section-4.2.2

hassan254-prog: I was also against this. As we see more and more custom flows, perhaps in the future we could have a separate package that handles only custom flows, thereby maintaining a clean codebase.

khaliqgant: Could this be abstracted into the yaml maybe and override the default there?

hassan254-prog: @khaliqgant , thank you for this cleaner approach. I have updated my PR with this [commit](https://github.com/NangoHQ/nango/pull/2574/commits/d06b792b3189da98ea69732cc739be811dce8871)

khaliqgant (changes requested):
Validation needs to be updated as well https://github.com/NangoHQ/nango/tree/master/scripts/validation/providers

hassan254-prog:
Thanks @khaliqgant for catching this, I have updated in this [commit](https://github.com/NangoHQ/nango/pull/2574/commits/c2cdc638af41bed78328f980f0c10336d7a8e465)

khaliqgant (approved):

bodinsamuel (approved):

PR #2575, opened 2024-08-01
fix: support special characters in db urls

TBonnin:
This commit URL encodes db username and password when used inside postgres db url
If db urls are set as env vars, users must URL encode them if they contain any special char

### DIFF
--- a/packages/orchestrator/lib/app.ts
+++ b/packages/orchestrator/lib/app.ts
@@ -12,7 +12,7 @@ const databaseSchema = envs.ORCHESTRATOR_DATABASE_SCHEMA;
 const databaseUrl =
     envs.ORCHESTRATOR_DATABASE_URL ||
     envs.NANGO_DATABASE_URL ||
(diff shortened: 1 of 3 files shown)


packages/orchestrator/lib/app.ts
```diff
@@ -12,7 +12,7 @@ const databaseSchema = envs.ORCHESTRATOR_DATABASE_SCHEMA;
 const databaseUrl =
     envs.ORCHESTRATOR_DATABASE_URL ||
     envs.NANGO_DATABASE_URL ||
-    `postgres://${envs.NANGO_DB_USER}:${envs.NANGO_DB_PASSWORD}@${envs.NANGO_DB_HOST}:${envs.NANGO_DB_PORT}/${envs.NANGO_DB_NAME}`;
```
TBonnin, line 15: assuming only db user and password can contain special chars.

bodinsamuel (approved):

PR #2576, opened 2024-08-01
chore: log error in handlePayloadTooBigError

TBonnin:
I noticed the following error this morning when a runner send the output of a task back to jobs:
Error: PUT /tasks/:taskId: status=500, response={"error":{"code":"server_error","message":"Cannot read properties of undefined (reading 'code')"}} It looks like setting the task as failed/succeed fails but the error doesn't have the shape we expect.
Adding a try/catch to log the error and let the execution continue and log the setSucceed/setFailed error


bodinsamuel (approved):
it's unlikely caused by `payload_too_big` since it's well handled now server.ts:51 
maybe a server error, but we'll see

PR #2577, opened 2024-08-01
feat(integrations): add support for Circle So

hassan254-prog:
## Describe your changes

- Add support for circle

[EXT-114](https://linear.app/nango/issue/EXT-114/add-support-for-circle)


packages/shared/providers.yaml
```diff
@@ -492,6 +492,15 @@ chorus:
     proxy:
         base_url: https://chorus.ai
     docs: https://docs.nango.dev/integrations/all/chorus
+circle:
+    categories:
+        - communication
+    auth_mode: API_KEY
+    proxy:
```
hassan254-prog, line 499: `proxy.verification` won't work here as the provider always returns `status: 200`. I had to manually test with a small sync script.

docs-v2/integrations/all/circle.mdx
```diff
@@ -0,0 +1,31 @@
+---
+title: Circle
```
bodinsamuel: can we rename that to Circle.so? I thought we were talking about circle ci

docs-v2/integrations/all/circle.mdx
```diff
@@ -0,0 +1,31 @@
...
+sidebarTitle: Circle
+---
+
+API configurations: [`circle`](https://nango.dev/providers.yaml)
```
bodinsamuel: something like this
```suggestion
API configurations: [`circle-so`](https://nango.dev/providers.yaml)
```

bodinsamuel (approved):
lgtm except naming :D

docs-v2/integrations/all/circle.mdx
hassan254-prog: @bodinsamuel ,that makes perfect sense for distinguishing between the two. I have updated this in the latest [commit](https://github.com/NangoHQ/nango/pull/2577/commits/45e36a2368a9bf464331fc72f4c41e1d59884531).

packages/shared/providers.yaml
khaliqgant: What does it return if it fails? An error in the payload? Anything in the headers?

docs-v2/integrations/all/circle.mdx
bodinsamuel: thanks 🙏🏻

bodinsamuel (approved):

TBonnin (approved):

packages/shared/providers.yaml
hassan254-prog: @khaliqgant , the response returns with a status of `200` even after sending the wrong APIkey. A good example is: `{ "status": "unauthorized", "message": "Your account could not be authenticated." }`

PR #2582, opened 2024-08-01
fix(cli): [nan-1145] only compile and deploy single sync

khaliqgant:
## Describe your changes
Deploying by passing in a single sync or action shouldn't compile all the other files and should only deploy the specified file. Right now it compiles all files and if a sync is passed in it tries to deploy all other actions. This PR fixes both issues

NAN-1145


packages/cli/lib/services/deploy.service.ts
```diff
@@ -297,7 +325,7 @@ class DeployService {
             }
 
             for (const sync of integration.syncs) {
-                if (optionalSyncName && optionalSyncName !== sync.name) {
+                if (optionalActionName || (optionalSyncName && optionalSyncName !== sync.name)) {
```
TBonnin: I would short-circuit outside of the for loop, otherwise if optionalActionName is defined we are still iterating through the syncs even if we know we don't need to

packages/cli/lib/services/deploy.service.ts
```diff
@@ -132,7 +132,35 @@ class DeployService {
...
+        let successfulCompile = false;
+
+        if (singleDeployMode) {
+            const scriptName: string = String(optionalSyncName || optionalActionName);
+            const type = optionalSyncName ? 'syncs' : 'actions';
+            const providerConfigKey = response.parsed.integrations.find((integration) => {
+                if (optionalSyncName) {
+                    return integration.syncs.find((sync) => sync.name === scriptName);
+                } else {
+                    return integration.actions.find((action) => action.name === scriptName);
```
TBonnin, line 144: you can get rid of the `if` by doing 
```
return integration[type].find((script) => script.name === scriptName)
```

TBonnin (approved):
2 small comments but otherwise lgtm

packages/cli/lib/services/compile.service.ts
```diff
@@ -83,15 +83,15 @@ export async function compileSingleFile({
...
-    tsconfig,
     debug = false
 }: {
     fullPath: string;
     file: ListedFile;
-    tsconfig: string;
     parsed: NangoYamlParsed;
     debug: boolean;
 }) {
+    const tsconfig = fs.readFileSync(path.join(getNangoRootPath(), 'tsconfig.dev.json'), 'utf8');
```
bodinsamuel: So each file will load the tsconfig now even on normal compilation?

bodinsamuel (approved):
Just a quick question on tsconfig

packages/cli/lib/services/compile.service.ts
khaliqgant: No, `compileSingleFile` is only called in two instances:
1. A `--sync ${syncName}` OR `--action ${actionName}` is passed to the deploy call
2. During `nango dev` command which watches files but I updated that with https://github.com/NangoHQ/nango/pull/2582/commits/bcc0db7f03f834861e4c64c815ee37453ab888bf to only load the tsconfig once

packages/cli/lib/services/deploy.service.ts
khaliqgant: Typescript yelled at me when I tried to make it dynamic

PR #2584, opened 2024-08-02
fix(proxy): error response with no data should report failure

TBonnin:
When the proxy response is an error without any data, the log operation was not marked as failed

before:
Screenshot 2024-08-02 at 14 21 24

after:
Screenshot 2024-08-02 at 14 21 17


bodinsamuel (approved):
indeed

PR #2585, opened 2024-08-02
fix(proxy): consolidate error message

TBonnin:
## Describe your changes
Making sure the error log is the same in proxy.controller

### DIFF
--- a/packages/server/lib/controllers/proxy.controller.ts
+++ b/packages/server/lib/controllers/proxy.controller.ts
@@ -268,10 +268,9 @@ class ProxyController {
                 status
             } = error.toJSON() as any;
 
-            const errorObject = { message, stack, code, status, url, method };
+            await this.reportError(error, url, config, message, logCtx);
 
-            await logCtx.error(`${method.toUpperCase()} request to ${url} failed`, { error });
-            await logCtx.failed();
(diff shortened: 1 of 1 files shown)


packages/server/lib/controllers/proxy.controller.ts
```diff
@@ -368,10 +367,10 @@ class ProxyController {
 
     private async reportError(error: AxiosError, url: string, config: ApplicationConstructedProxyConfiguration, errorMessage: string, logCtx: LogContext) {
         const safeHeaders = proxyService.stripSensitiveHeaders(config.headers, config);
-        await logCtx.error('The provider responded back with an error code', {
+        await logCtx.error(`${error.request?.method.toUpperCase()} ${url} failed with status '${error.response?.status}'`, {
             code: error.response?.status,
             url,
-            error: errorMessage,
+            error: new Error(errorMessage),
```
TBonnin, line 373: must be an error otherwise show as "unknown error" https://github.com/NangoHQ/nango/blob/8bcd76bab6181541c03feb77b622b7c38d1a818a/packages/logs/lib/client.ts#L65

bodinsamuel: good catch

bodinsamuel (approved):

PR #2578, opened 2024-08-01
feat(integrations): add support for tumblr

hassan254-prog:
## Describe your changes

- Add support for tumblr

## Issue ticket number and link
[EXT-115](https://linear.app/nango/issue/EXT-115/add-support-for-tumblr)

PR #2592, opened 2024-08-06
feat(integrations): add salesforce-experience-cloud provider

gorets:
- [x] I added observability, otherwise the reason is:
- [x] I added analytics, otherwise the reason is:


gorets:
Also need to add changes to the `packages/shared/lib/clients/provider.client.ts` add `case 'salesforce-experience-cloud':` to two places. I tried it, but your precommit hook changed additional rows.

packages/shared/providers.yaml
```diff
@@ -2366,6 +2366,17 @@ salesforce-sandbox:
     proxy:
         base_url: ${connectionConfig.instance_url}
     docs: https://docs.nango.dev/integrations/all/salesforce
+salesforce-experience-cloud:
+    auth_mode: OAUTH2
+    authorization_url: https://${connectionConfig.subdomain}.my.site.com/services/oauth2/authorize
```
bodinsamuel, line 2398: my.site.com? is it a real value or a placeholder?
if placeholder it should probably be `https://${connectionConfig.domain}/services

packages/shared/providers.yaml
```diff
@@ -2366,6 +2366,17 @@ salesforce-sandbox:
...
+    token_url: https://${connectionConfig.subdomain}.my.site.com/services/oauth2/token
+    default_scopes:
+        - offline_access
+    token_response_metadata:
+        - instance_url
+    proxy:
+        base_url: ${connectionConfig.instance_url}
+    docs: https://docs.nango.dev/integrations/all/salesforce
```
bodinsamuel: wrong doc link

bodinsamuel (commented):
Thanks for your contribution 🙏🏻 
It's missing a few things, documentation and logo (can be a symlink)
You can follow what has been done previously https://github.com/NangoHQ/nango/pull/2578/files

packages/shared/providers.yaml
gorets: yes, it is correct domain for `Salesforce Experience Cloud Sites`

image

https://developer.salesforce.com/docs/atlas.en-us.chatterapi.meta/chatterapi/quickstart_communities.htm

packages/shared/providers.yaml
gorets: This is part of the Salesforce product as well as salesforce-sandbox. I don't know what link to put here.

bodinsamuel:
I have added what was missing, merging that. Thanks 🙏🏻

gorets:
> Also need to add changes to the `packages/shared/lib/clients/provider.client.ts` add `case 'salesforce-experience-cloud':` to two places. I tried it, but your precommit hook changed additional rows.

Need remember to make these edits, otherwise the token update logic will not work correctly.

bodinsamuel:
> Need remember to make these edits, otherwise the token update logic will not work correctly.

Ah yes, can you share the diff please? ~Also not sure why the precommit wouldn't work :/~ If the precommit adds more stuff push it, no worries 👍🏻

gorets:
[add_salesforce-experience-cloud_provider.patch](https://github.com/user-attachments/files/16579669/add_salesforce-experience-cloud_provider.patch)

bodinsamuel (approved):

bodinsamuel:
I'll be deployed during the day or tomorrow 👌🏻

PR #2599, opened 2024-08-09
fix: add base_url for gitlab

hassan254-prog:
## Describe your changes

- Add `base_url` for gitlab
- Add missing `OAuth2ClientCredentials` to ConnectionConfig

### DIFF
--- a/packages/frontend/lib/index.ts
+++ b/packages/frontend/lib/index.ts
@@ -491,7 +491,14 @@ interface ConnectionConfig {
     hmac?: string;
     user_scope?: string[];
     authorization_params?: Record<string, string | undefined>;
-    credentials?: OAuthCredentialsOverride | BasicApiCredentials | ApiKeyCredentials | AppStoreCredentials | TBACredentials | TableauCredentials;
+    credentials?:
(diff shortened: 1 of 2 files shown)


packages/frontend/lib/index.ts
```diff
@@ -491,7 +491,14 @@ interface ConnectionConfig {
...
     authorization_params?: Record<string, string | undefined>;
-    credentials?: OAuthCredentialsOverride | BasicApiCredentials | ApiKeyCredentials | AppStoreCredentials | TBACredentials | TableauCredentials;
+    credentials?:
+        | OAuthCredentialsOverride
+        | BasicApiCredentials
+        | ApiKeyCredentials
+        | AppStoreCredentials
+        | TBACredentials
+        | TableauCredentials
+        | OAuth2ClientCredentials;
```
bodinsamuel, line 501: weird that it was missing this whole time

bodinsamuel (approved):

packages/frontend/lib/index.ts
hassan254-prog: @bodinsamuel , yes it is very weird 😟.

PR #2602, opened 2024-08-09
fix(api): upsert connection metadata/config

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1555/connection-creation-should-allow-metadata

`POST /connection` was not always using metadata and connection_config depending on the auth type

- Standardize method signature so each auth method accept all parameters
- Fix apiKey/unauth/etc. now save metadata at creation and update
- Fix oauth import disambiguate credentials and metadata


TBonnin (approved):
lgtm

PR #2604, opened 2024-08-12
fix(webhook): allow handler to save records

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1545/using-nangosync-in-webhook-handler-is-not-possible

- Allow webhook handlers to call batchSave
To do that we create a temporary job to have a syncJobId. Ultimately this is not that useful but because batchSave needs that to compute the results it was the easiest solution. 
Calling batch save in a webhook handler has many side effects and could be problematic with track_deletes, I would not recommend using it in combination of a regular sync.

- Allow `unauthenticated` to receive webhook
It's handy for debugging purposes, and very unlikely that any custom integration could fit the required model shape. However, if they can, we have no reason to block it.

## 🧪 Test

```yaml
      webhooks:
        description: proxy
        sync_type: full
        track_deletes: false
        runs: every 30 days
        auto_start: false
        output: Webhook
        endpoint: GET /unauth/webhook
        webhook-subscriptions: '*'
```

```ts
import type { NangoSync, Webhook } from "../../models";

  await nango.log("do nothing");
}

export async function onWebhookPayloadReceived(
  nango: NangoSync,
  payload: any,
): Promise<void> {
  await nango.batchSave<Webhook>([payload.data], "Webhook");
}
```

```sh
curl -i -XPOST http://localhost:3003/webhook/__ENV_UUID__/unauthenticated -H "content-type: application/json" -d '{"connectionId":"test-connection-id", "type": "*", "data": {"id": "str"}}'
```

Screenshot 2024-08-12 at 17 40 12

### DIFF
--- a/packages/jobs/lib/execution/sync.ts
+++ b/packages/jobs/lib/execution/sync.ts
@@ -1,6 +1,5 @@
 import type { Config, Job, NangoProps, SyncConfig } from '@nangohq/shared';
 import {
-    addSyncConfigToJob,
     environmentService,
     externalWebhookService,
     getApiUrl,
@@ -40,7 +39,7 @@ export async function startSync(task: TaskSync, startScriptFn = startScript): Pr
     let logCtx: LogContext | undefined;
     let team: DBTeam | undefined;
     let environment: DBEnvironment | undefined;
-    let syncJob: Pick<Job, 'id'> | null = null;
+    let syncJob: Job | null = null;
     let lastSyncDate: Date | null = null;
     let syncType: SyncType = SyncType.FULL;
     let providerConfig: Config | null = null;
@@ -83,7 +82,16 @@ export async function startSync(task: TaskSync, startScriptFn = startScript): Pr
             }
         );
 
-        syncJob = await createSyncJob(task.syncId, syncType, SyncStatus.RUNNING, task.name, task.connection, task.id, logCtx.id);
+        syncJob = await createSyncJob({
+            sync_id: task.syncId,
+            type: syncType,
+            status: SyncStatus.RUNNING,
+            job_id: task.name,
+            nangoConnection: task.connection,
+            sync_config_id: syncConfig.id!,
+            run_id: task.id,
+            log_id: logCtx.id
+        });
         if (!syncJob) {
             throw new Error(`Failed to create sync job for sync: ${task.syncId}. TaskId: ${task.id}`);
         }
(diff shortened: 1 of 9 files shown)


packages/jobs/lib/execution/sync.ts
```diff
@@ -99,8 +107,6 @@ export async function startSync(task: TaskSync, startScriptFn = startScript): Pr
             });
         }
 
-        await addSyncConfigToJob(syncJob.id, syncConfig.id!);
```
bodinsamuel, line 102: Seems completely unnecessary to do it in two steps, probably a legacy thing

packages/jobs/lib/execution/webhook.ts
```diff
@@ -43,6 +57,20 @@ export async function startWebhook(task: TaskWebhook): Promise<Result<void>> {
...
+        syncJob = await createSyncJob({
+            sync_id: sync.id,
+            type: SyncType.INCREMENTAL,
+            status: SyncStatus.RUNNING,
+            job_id: task.name,
+            nangoConnection: task.connection,
+            sync_config_id: syncConfig.id!,
+            run_id: task.id,
+            log_id: logCtx.id
+        });
```
bodinsamuel, line 69: I checked, job table is not used for deduplication so it should not block syncs to run. But happy to get a second pair of eyes

packages/jobs/lib/execution/sync.ts
TBonnin: indeed

packages/jobs/lib/execution/webhook.ts
TBonnin: job table is not used by the scheduler, which also means sync and webhook scripts could run concurrently and create all sort of undefined behavior if saving the same model. I think we should prevent this scenario to happen at compile time to avoid future headaches

TBonnin (approved):
This is unblocking customers but it creates potential concurrency issue if a script batchSaves the same model in scheduled sync and webhook.

Would you mind opening a ticket to discuss about this problem once @khaliqgant is back. I would be in favor to at least at first prevent batchSaving the same models in syncs/webhook at compile time

packages/jobs/lib/execution/webhook.ts
bodinsamuel: it could be a legitimate use case though, let's say you have a sync for your customers and get a webhook at each new customer, it's handy to be able to push the new customer directly to the record list without waiting for the next sync to trigger.

TBonnin: I know but nango is not fully designed for that at the moment

PR #2605, opened 2024-08-12
fix(anrok): handle errors correctly, fix types

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1557/anrok-actions-returns-a-500-when-calling-actions

- Handle errors correctly 
Catch err was assumed to be AxiosError but it was a Javascript error

- Fix types for `notTaxedReason` can be missing from anrok response 
That was the original error

- Fix types for `validation_errors` it has never been an array

- Changed to async/await for clarity, there was no reason to mix both.

### DIFF
--- a/integration-templates/anrok/utils.ts
+++ b/integration-templates/anrok/utils.ts
@@ -1,0 +1,9 @@
+export function errorToObject(err: unknown) {
+    if ('response' in (err as any)) {
+        return (err as any).response.data;
+    }
+    if (err instanceof Error) {
+        return JSON.parse(JSON.stringify(err, ['name', 'message']));
+    }
+    return err;
+}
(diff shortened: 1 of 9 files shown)


integration-templates/anrok/utils.ts
```diff
@@ -0,0 +1,9 @@
+export function errorToObject(err: unknown) {
+    if ('response' in (err as any)) {
+        return (err as any).response.data;
+    }
```
TBonnin, line 4: would it make sense to use `isAxiosError`? here we assume that if there is `response` field we return `data` but there is no guarantuee

TBonnin (approved):
small comment about the AxiosError handling but looks good otherwise

integration-templates/anrok/utils.ts
bodinsamuel: we don't allow import axios, maybe that's something we should change or expose in the sdk
(imo we will get rid of axios at some point so I wouldn't bother)

bodinsamuel: let me know if you prefer to have this fixed rn 👍🏻

TBonnin: don't spend too much time on that. it is fine for now

PR #2607, opened 2024-08-13
fix: POST /connection: add support for OAUTH2_CC auth mode

TBonnin:
Connection for OAUTH2_CC provider cannot be imported at the moment. Probably an oversight when OAUTH2_CC auth mode was added

https://linear.app/nango/issue/NAN-1561/cannot-import-a-oauth2-cc-connection

### DIFF
--- a/packages/server/lib/controllers/connection.controller.ts
+++ b/packages/server/lib/controllers/connection.controller.ts
@@ -1,7 +1,15 @@
 import type { Request, Response, NextFunction } from 'express';
 import type { Config as ProviderConfig, OAuth2Credentials, AuthCredentials, ConnectionList, ConnectionUpsertResponse } from '@nangohq/shared';
 import db from '@nangohq/database';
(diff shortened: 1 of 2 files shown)


packages/server/lib/controllers/connection.controller.ts
```diff
@@ -397,6 +405,69 @@ class ConnectionController {
...
+
+                if (!access_token) {
+                    errorManager.errRes(res, 'missing_access_token');
+                    return;
+                }
+
+                const { expires_at: parsedExpiresAt } = connectionService.parseRawCredentials(
+                    { access_token, expires_at },
+                    template.auth_mode
+                ) as OAuth2ClientCredentials;
```
TBonnin, line 422: I used the same pattern used with other oauth mode but using `parseRawCredentials` just to keep the expires_at seems a bit dirty to me

bodinsamuel (approved):
🙏🏻

PR #2608, opened 2024-08-13
fix(url): handle special characters in connection params

hassan254-prog:
## Describe your changes

- Updated implementation to handle various input scenarios including special characters.

[EXT-121](https://linear.app/nango/issue/EXT-121/fix-stripe-app-sandbox-handling-special-chars-in-connection-params)

### DIFF
--- a/packages/shared/lib/clients/oauth2.client.ts
+++ b/packages/shared/lib/clients/oauth2.client.ts
@@ -18,8 +18,8 @@ export function getSimpleOAuth2ClientConfig(
     connectionConfig: Record<string, string>
 ): Merge<ModuleOptions, { http: WreckHttpOptions }> {
     const templateTokenUrl = typeof template.token_url === 'string' ? template.token_url : (template.token_url!['OAUTH2'] as string);
-    const tokenUrl = new URL(interpolateString(templateTokenUrl.replace(/connectionConfig\./g, ''), encodeParameters(connectionConfig)));
-    const authorizeUrl = new URL(interpolateString(template.authorization_url!.replace(/connectionConfig\./g, ''), encodeParameters(connectionConfig)));
+    const tokenUrl = makeUrl(templateTokenUrl, connectionConfig);
+    const authorizeUrl = makeUrl(template.authorization_url!, connectionConfig);
 
     const headers = { 'User-Agent': 'Nango' };
 
@@ -155,4 +155,11 @@ export async function getFreshOAuth2Credentials(
         });
         return { success: false, error, response: null };
     }
+}
+
+function makeUrl(template: string, config: Record<string, any>): URL {
+    const cleanTemplate = template.replace(/connectionConfig\./g, '');
+    const encodedParams = encodeParameters(config);
+    const interpolatedUrl = interpolateString(cleanTemplate, encodedParams);
+    return new URL(interpolatedUrl);
 }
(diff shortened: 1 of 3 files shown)


packages/shared/lib/utils/utils.ts
```diff
@@ -239,3 +239,7 @@ export function getConnectionConfig(queryParams: any): Record<string, string> {
     const arr = Object.entries(queryParams).filter(([, v]) => typeof v === 'string'); // Filter strings
     return Object.fromEntries(arr) as Record<string, string>;
 }
+
+export function encodeParameters(params: Record<string, any>): Record<string, string> {
+    return Object.fromEntries(Object.entries(params).map(([key, value]) => [key, encodeURIComponent(String(value))]));
```
bodinsamuel, line 244: isn't this === to `new URLSearchParams(params).toString()`

hassan254-prog: @bodinsamuel  `new URLSearchParams(params).toString()` and `Object.fromEntries(Object.entries(params).map(([key, value]) => [key, encodeURIComponent(String(value))]))` are similar but handle encoding differently.

- `URLSearchParams(params).toString()`: Encodes parameters for query strings but may not handle all special characters as thoroughly.
- `encodeURIComponent`: Provides more comprehensive encoding of special characters.

hassan254-prog: Yes they are the same, but in our code, `encodeParameters` is used to prepare the `connectionConfig` object for interpolation into a URL. This is necessary because `interpolateStringFromObject` expects an object with pre-encoded values, ensuring that when these values are interpolated into a URL, they do not introduce any invalid characters or break the URL format.

packages/shared/lib/clients/oauth2.client.ts
```diff
@@ -18,10 +18,9 @@ export function getSimpleOAuth2ClientConfig(
     connectionConfig: Record<string, string>
 ): Merge<ModuleOptions, { http: WreckHttpOptions }> {
     const templateTokenUrl = typeof template.token_url === 'string' ? template.token_url : (template.token_url!['OAUTH2'] as string);
-    const strippedTokenUrl = templateTokenUrl.replace(/connectionConfig\./g, '');
-    const tokenUrl = new URL(interpolateString(strippedTokenUrl, connectionConfig));
-    const strippedAuthorizeUrl = template.authorization_url!.replace(/connectionConfig\./g, '');
-    const authorizeUrl = new URL(interpolateString(strippedAuthorizeUrl, connectionConfig));
+    const tokenUrl = new URL(interpolateString(templateTokenUrl.replace(/connectionConfig\./g, ''), encodeParameters(connectionConfig)));
+    const authorizeUrl = new URL(interpolateString(template.authorization_url!.replace(/connectionConfig\./g, ''), encodeParameters(connectionConfig)));
```
TBonnin: logic looks ok but it is quite hard to read. what about having a local function like this:
```
function makeUrl(template: string, config: Record<string, any>): URL {
  const cleanTemplate = template.replace(/connectionConfig\./g, '');
  const encodedParams = encodeParameters(config);
  const interpolatedUrl = interpolateString(cleanTemplate, encodedParams);
  return new URL(interpolatedUrl);
}

const tokenUrl = makeUrl(templateTokenUrl, connectionConfig);
const authorizeUrl = makeUrl(template.authorization_url!, connectionConfig);
```

TBonnin (approved):

packages/shared/lib/clients/oauth2.client.ts
hassan254-prog: Ah yes, making it more readable. I've updated the PR with the [latest](https://github.com/NangoHQ/nango/pull/2608/commits/bfe901e97884118a43eb4515b50788548c4f4bab) commit. Thank you🙏!

PR #2614, opened 2024-08-14
fix: runner request to jobs /idle endpoint

TBonnin:
I broke the request made to jobs /idle endpoint when refactoring jobs/runner. It prevents idle runner to be suspended.


bodinsamuel (commented):
👌🏻

Issue #2376, opened 2024-06-20
Build multi-arch images for arm64


winterrobert:
What would it take to build multi-arch images also for ARM?

We are running nango locally on Kubernetes/K3d and can't run it on Mac currently. As we are not using docker-compose we can't use the workaround of setting `platform` in the [docker-compose file like in the repo](https://github.com/NangoHQ/nango/blob/master/docker-compose.yaml#L20).

If it's not on the roadmap instead of building multi-arch from source just for us I could help out make it happen upstream depending on how complex it would be.

winterrobert:
The link says "You don't have access to this workspace" - do anyone have time to summarize the status of the issue?

bastienbeurier:
Closing due to inactivity (feel free to re-open if it's still relevant).

PR #2619, opened 2024-08-19
Arm support

haouarihk:
## Describe your changes
- Added arm64 build support using buildx
- Added optional secret `DOCKER_REPO_LINK` so that forks can test/push to other repos without changing the code.

Note: it hasn't been fully tested, reason github action timed out before finishing in my testing.

#2376 


TBonnin:
sorry for the very long delay. Supporting arm is currently not on the roadmap. Closing this PR but feel free to ping us again if you feel like this should be prioritized.

PR #2620, opened 2024-08-20
feat: Add integration provider for Holded

vrouet:


packages/shared/providers.yaml
```diff
@@ -1410,6 +1410,17 @@ highlevel-white-label:
     refresh_params:
         grant_type: refresh_token
     docs: https://docs.nango.dev/integrations/all/highlevel
+holded:
+    categories:
+        - accounting
```
hassan254-prog, line 1415: Please add `holded` to the following individual categories in the documentation.

packages/shared/providers.yaml
```diff
@@ -1410,6 +1410,17 @@ highlevel-white-label:
...
+        - crm
+        - invoicing
+    auth_mode: API_KEY
+    proxy:
```
hassan254-prog, line 1419: With `API_KEY` and `BASIC` auth modes, it is advisable to test the implementation using the `proxy.verification` field when creating a new connection on Nango. You can have a look at the following [example](https://github.com/NangoHQ/nango/blob/1381cffd77d4b024011e994d595ef7dc2083222f/packages/shared/providers.yaml#L1675).

hassan254-prog (commented):
@vrouet LGTM. I’ve left a few small comments inline 👍

packages/shared/providers.yaml
vrouet: I noticed that, but these verification endpoint normally use an endpoint to get the current user, or something similar, and there is no such endpoint on Holded's API (see https://developers.holded.com/reference/api-key). What do you recommend?

packages/shared/providers.yaml
vrouet: Done!

packages/shared/providers.yaml
hassan254-prog: For consistency, you can replace the logo with , the [following](https://codefile.io/f/e3ZHAU34ny).

vrouet: I see that some of the other providers use a random list endpoint, so I've added GET `/invoicing/v1/contacts`, let me know if it looks ok to you

hassan254-prog: @vrouet , regarding the `proxy.verification` field, when creating a new connection: Nango checks if the status is `success` based on the API response status from the verification endpoint. This allows you to determine whether your implementation is correct. After adding this to the `providers.yaml` file, does creating a new connection work successfully?

vrouet: Yes, I did succeed in creating a connection on my local Nango with that provider config

vrouet: Logo updated :+1:

bodinsamuel (approved):
Thanks @hassan254-prog for taking care of the initial review.
Looks good on my end 👌🏻

hassan254-prog (approved):

bodinsamuel:
thanks for your contribution @vrouet 🙌🏻 
It will be deployed during the week

PR #2621, opened 2024-08-20
feat(integrations): add support for bigcommerce

hassan254-prog:
## Describe your changes

- Add support for `bigcommerce`
- Ensure we correctly handle `connectionConfig` for both `token_params` and `authorization_params`

[EXT-124](https://linear.app/nango/issue/EXT-124/add-support-for-bigcommerce)


docs-v2/integrations/all/bigcommerce.mdx
```diff
@@ -0,0 +1,62 @@
...
+
+<Tip>Need help getting started? Get help in the [community](https://nango.dev/slack).</Tip>
+
+## API gotchas
+
+- BigCommerce offers various apps with different levels of visibility. For more details, see [BigCommerce types of Apps](https://developer.bigcommerce.com/docs/integrations/apps/guide/types)
+- After creating a new connection on Nango, you will need to rewrite your HTTP request headers to use the `X-Auth-Token` header to pass the API account's `access_token` instead of the Authorization header as described in [here](https://developer.bigcommerce.com/docs/start/authentication/api-accounts#how-to-migrate). Please look at the following example:
+
+```js
+        const connection = await nango.getConnection();
```
bodinsamuel: what's up with the large indentation?

packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -479,12 +481,27 @@ class OAuthController {
...
+                return publisher.notifyErr(res, channel, providerConfigKey, connectionId, error);
+            }
+
+            if (template.token_params && missesInterpolationParamInObject(template.token_params, connectionConfig)) {
+                const error = WSErrBuilder.InvalidConnectionConfig('token_params', JSON.stringify(connectionConfig));
+                await logCtx.error(error.message, { connectionConfig });
+                await logCtx.failed();
+
+                return publisher.notifyErr(res, channel, providerConfigKey, connectionId, error);
+            }
```
bodinsamuel, line 498: Seems like a waste of resources to do it multiple times, not for this PR but we should just do everything and check the full object recursively at the end

bodinsamuel (approved):
Looks good 👌🏻 
The whole interpolation code needs a big cleanup tho

packages/server/lib/utils/utils.ts
```diff
@@ -43,6 +43,15 @@ export function missesInterpolationParam(str: string, replacers: Record<string,
...
 
+/**
+ * A helper function to check if any string in an object misses interpolation params.
+ * For example:
+ * missesInterpolationParamInObject({ context: 'stores/${storeHash}', response_type: 'code' }, { storeHash: 'abc123' }) -> returns false
+ * missesInterpolationParamInObject({ context: 'stores/${storeHash}', response_type: 'code' }, {}) -> returns true
+ */
+export function missesInterpolationParamInObject(params: Record<string, string>, replacers: Record<string, any>) {
+    return Object.values(params).some((param) => missesInterpolationParam(param, replacers));
+}
```
TBonnin, line 54: not for this PR but I would prefer to see a single unit-tested function `interpolate` that support nested content and return an error when missing param instead of having 4 or 5 flavors of the same thing

TBonnin (approved):

packages/server/lib/utils/utils.ts
hassan254-prog: Yes, I agree that we should refactor this to reduce redundancy 👍.

packages/server/lib/controllers/oauth.controller.ts
hassan254-prog: I agree that a recursive check for the entire object at once would be more efficient and reduce redundancy.

docs-v2/integrations/all/bigcommerce.mdx
hassan254-prog: I have updated this with the latest [commit](https://github.com/NangoHQ/nango/pull/2621/commits/8a639c929f2f0d94d57d40da3687feb50657f26c). Thank you for noting that 🙏.

PR #2622, opened 2024-08-21
fix: update error message for provider key missing

kndwin:
## Describe your changes
I spent like an hour trying to understand where to pass this in (had to look through the source code) and thought it'll be an easy PR to push

### DIFF
--- a/packages/cli/lib/services/dryrun.service.ts
+++ b/packages/cli/lib/services/dryrun.service.ts
@@ -120,7 +120,7 @@ export class DryRunService {
             if (!providerConfigKey) {
                 console.log(
                     chalk.red(
-                        `Provider config key not found, please check that the provider exists for this sync name: ${syncName} by going to the Nango dashboard.`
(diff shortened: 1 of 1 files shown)


bastienbeurier:
Thanks for contributing this @kndwin and sorry for the confusion!

bodinsamuel (approved):
Thanks for your contribution @kndwin 
Can I ask what scenario lead to this? Because this error message is shown only when no syncs or actions matched anyway

kndwin:
Yeah it turns out my <sync_name> I passed in was misspelled anyway 

nango dryrun <sync_name> <connection_id> 

The error though made me think I didn't pass in the provider config key so went on a long journey to realised I made 2 errors (not passing in --integration-id and misspelling the <sync_name>)

After passing in the --integration-id I had a new error (The sync/action was not executed (I think it was that)) which prompted me to just console log all the args inside the node_module folder of the nango global file and then I noticed my error

(Would be nice actually to suggest the closest name but probably a feature request)

bodinsamuel:
yes indeed we should probably suggest a list of syncs if nothing matched. I'll merge this as-is and note to improve this scenario 👌🏻

PR #2623, opened 2024-08-21
fix(api): allow disabling rate limit

bodinsamuel:
## Describe your changes

- Add flag to enable/disable api rate limit
Enabled by default if not specified. Mostly for local tests but got annoyed to remove it manually multiple times


packages/utils/lib/environment/detection.ts
```diff
@@ -22,3 +22,4 @@ export const flagHasScripts = isLocal || isEnterprise || isCloud || isTest;
 export const flagHasAuth = process.env['FLAG_AUTH_ENABLED'] !== 'false';
 export const flagHasManagedAuth =
     process.env['FLAG_MANAGED_AUTH_ENABLED'] === 'true' && Boolean(process.env['WORKOS_API_KEY'] && process.env['WORKOS_CLIENT_ID']);
+export const flagHasAPIRateLimit = process.env['FLAG_API_RATE_LIMIT_ENABLED'] !== 'false';
```
TBonnin, line 25: I personally prefer disabling flags so without the right config the feature is enabled and must be disabled for non special scenario. The idea is to make it harder to misconfigure a service. (ie: more annoying from a devEx point of view but less prone to error in prod) What do you think?

bodinsamuel: that's what it is, if not specified it's enabled hence the `!== 'false'`

 I'm just opting for a positive flag name instead of a negative one to avoid a double negative when disable is disabled, e.g: 

```
# harder to understand imo
FLAG_API_RATE_LIMIT_DISABLED=true
FLAG_API_RATE_LIMIT_DISABLED=false

# vs
FLAG_API_RATE_LIMIT_ENABLED=true
FLAG_API_RATE_LIMIT_ENABLED=false
```

wdyt?

TBonnin: got it. I had misread the condition. I don't mind the `FLAG_API_RATE_LIMIT_DISABLED=true` but `FLAG_API_RATE_LIMIT_ENABLED=false` is fine. The other options you won't have them since there are equivalent of not having the flag at all

TBonnin (approved):

PR #2624, opened 2024-08-21
fix(api): DELETE integration split logic, dedicated files

bodinsamuel:
## Describe your changes

Contributes to https://linear.app/nango/issue/NAN-1537/script-v2-integrate-ui

Public API and internal one shared the same endpoint function to delete an integration. 

- Dedicated file for `DELETE /config/:providerConfigKey`
- Dedicated file for `DELETE /api/v1/integration/:providerConfigKey`

## 🧪 Test

- Go to your dashboard and delete an integration


TBonnin (approved):
not sure to fully understand why we need to split the logic but code looks good

bodinsamuel:
> not sure to fully understand why we need to split the logic but code looks good

The main action (deleting stuff) should be the same but public API is sensitive to breaking change, from experience it's not a good idea to share code with internal stuff. In that case, input and output can evolve overtime while public API should stay stable.

PR #2627, opened 2024-08-21
fix: upload templates github action

TBonnin:
Templates relying on zod are failing to be uploaded. ex: https://github.com/NangoHQ/nango/actions/runs/10489880648/job/29055453274

This commit:
- install zod before compiling
- add a way to manually trigger uploading all templates

@bodinsamuel can you think of a way to safely test this? 


.github/workflows/integration-template-upload.yaml
```diff
@@ -1,14 +1,10 @@
...
 on:
+    workflow_dispatch:
     push:
         branches:
             - master
-    pull_request:
-        branches:
-            - master
-        types:
-            - closed
```
TBonnin, line 12: I think the pull request trigger is redundant. We want to upload when a commit is pushed to master, that's it.

khaliqgant:
https://github.com/NangoHQ/nango/blob/master/.github/workflows/sync-integration-templates.yaml

TBonnin:
> https://github.com/NangoHQ/nango/blob/master/.github/workflows/sync-integration-templates.yaml

Missed this one 😄. Not sure we need 2 actions doing almost the same thing.

bodinsamuel:
> https://github.com/NangoHQ/nango/blob/master/.github/workflows/sync-integration-templates.yaml

Arg, my bad for leading you into this @TBonnin 
Might still be a good idea to have one way to push templates though

.github/workflows/integration-template-upload.yaml
```diff
@@ -56,10 +49,11 @@ jobs:
                     cp -r $dir/* /tmp/nango-temp/nango-integrations/$integration/
 
                     cd /tmp/nango-temp/nango-integrations
+                    nango install zod
```
bodinsamuel: npm install zod?

bodinsamuel (approved):
Looks good, github actions are notoriously painful to debug so I guess you can either trigger manually on your branch or test in master. Happy to review another PR if any bug arises 👍🏻

.github/workflows/integration-template-upload.yaml
bodinsamuel: now that I think about it we might also need `soap` and botbuilder

TBonnin:
> Looks good, github actions are notoriously painful to debug so I guess you can either trigger manually on your branch or test in master. Happy to review another PR if any bug arises 👍🏻

`on:workflow_dispatch` is not enabled until this is merged to master unfortunately. 

Edit: giving me an idea, let me do a simple PR that is just adding on:workflow_dispatch. Then I will be able to test the script changes

TBonnin:
Testing the manual upload to sync all templates. Merging now to ensure the merge based logic is also working

PR #2630, opened 2024-08-22
fix: wrong file path for schema.zod.js in templates

TBonnin:
## Describe your changes

Fixing `Importing libraries is not allowed. Please remove the import "../../schema.zod.js" from ...` error in templates. Also adding the missing schema.zod.ts for exact-online

Haven't fixed the github action not failing on this compile error btw.
We also probably want some compilation of templates on PRs.


bodinsamuel (approved):
Thanks

PR #2631, opened 2024-08-22
fix: better payload for proxy logs and operations

TBonnin:
Showing request and response info in operations and logs payload for proxy call.

Operation payload shows info about the request to nango api
Log payload shows info about the request to the 3rd party api

Successful operation:
Screenshot 2024-08-22 at 13 07 53
Successful log:
Screenshot 2024-08-22 at 13 08 01

Failed operation:
Screenshot 2024-08-22 at 13 07 34
Failed log:
Screenshot 2024-08-22 at 13 07 43

https://linear.app/nango/issue/NAN-1526/improve-proxy-logs

### DIFF
--- a/packages/server/lib/controllers/proxy.controller.ts
+++ b/packages/server/lib/controllers/proxy.controller.ts
@@ -345,8 +345,16 @@ class ProxyController {
                 chunks.push(data);
             });
             stringify.on('end', () => {
-                const data = chunks.length > 0 ? Buffer.concat(chunks).toString() : 'unknown error';
-                void this.reportError(error, url, config, data, logCtx);
+                const data = chunks.length > 0 ? Buffer.concat(chunks).toString() : '';
+                let errorData: string | Record<string, string> = data;
+                if (error.response?.headers?.['content-type']?.includes('application/json')) {
+                    try {
+                        errorData = JSON.parse(data);
+                    } catch {
+                        // Intentionally left blank - errorData will be a string
+                    }
+                }
+                void this.reportError(error, url, config, errorData, logCtx);
             });
         } else {
             await logCtx.error('Unknown error');
@@ -415,11 +423,17 @@ class ProxyController {
         }
     }
 
-    private async reportError(error: AxiosError, url: string, config: ApplicationConstructedProxyConfiguration, errorMessage: string, logCtx: LogContext) {
+    private async reportError(
+        error: AxiosError,
+        url: string,
+        config: ApplicationConstructedProxyConfiguration,
+        errorContent: string | Record<string, string>,
+        logCtx: LogContext
+    ) {
         const safeHeaders = proxyService.stripSensitiveHeaders(config.headers, config);
         await logCtx.http(`${error.request?.method.toUpperCase()} ${url} failed with status '${error.response?.status}'`, {
             meta: {
-                error: errorMessage
+                content: errorContent
             },
             request: {
                 method: config.method,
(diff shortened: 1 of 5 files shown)


packages/server/lib/controllers/proxy.controller.ts
```diff
@@ -213,7 +238,18 @@ class ProxyController {
         logCtx: LogContext;
     }) {
         const safeHeaders = proxyService.stripSensitiveHeaders(config.headers, config);
-        await logCtx.info(`${config.method.toUpperCase()} ${url} was successful`, { headers: safeHeaders });
+        await logCtx.http(`${config.method.toUpperCase()} ${url} was successful`, {
+            meta: null,
```
TBonnin: meta could potentially be optional. Not a blocker for now though

packages/server/lib/controllers/proxy.controller.ts
```diff
@@ -161,6 +161,31 @@ class ProxyController {
...
+            reqHeaders['authorization'] = 'REDACTED';
+            await logCtx?.enrichOperation({
+                request: {
+                    url: `${req.protocol}://${req.get('host')}${req.originalUrl}`,
+                    method: req.method,
+                    headers: reqHeaders
+                },
+                response: {
+                    code: res.statusCode,
+                    headers: getHeaders(res.getHeaders())
```
TBonnin, line 186: we are gonna log way more content with the headers. Is it gonna be ok with ES?

bodinsamuel: I need to increase the disk size anyway. It will increase the storage a bit but most logs are empty (!= operation)

packages/server/lib/controllers/proxy.controller.ts
bodinsamuel: You can definitely removed it from the required field

packages/server/lib/controllers/proxy.controller.ts
```diff
@@ -379,14 +423,27 @@ class ProxyController {
...
+            },
+            request: {
+                method: config.method,
+                url,
+                headers: safeHeaders
+            },
+            response: {
+                code: error.response?.status || 500,
+                headers: error.response?.headers as Record<string, string>
+            }
```
bodinsamuel, line 445: note we can also modify the client to build the object automatically (like the way I did with `error`) just to reuse the logic

packages/server/lib/controllers/proxy.controller.ts
```diff
@@ -161,6 +161,31 @@ class ProxyController {
...
+                    if (typeof value === 'string') {
+                        headers[key] = value;
+                    } else if (Array.isArray(value)) {
+                        headers[key] = value.join(', ');
+                    }
+                }
+                return headers;
+            };
+            const reqHeaders = getHeaders(req.headers);
+            reqHeaders['authorization'] = 'REDACTED';
```
bodinsamuel, line 177: You can't use proxy.service->stripSensitiveHeaders ?

bodinsamuel (approved):
Works great, few comments. 
However it doesn't show request/response on success and I don't know why looking at the code 🤔

packages/server/lib/controllers/proxy.controller.ts
TBonnin: it requires a proxy config but things can fail before having a valid config

TBonnin:
> Works great, few comments. However it doesn't show request/response on success and I don't know why looking at the code 🤔

I had the following error once after switching branches
```
...
Root causes:\n\t\tversion_conflict_engine_exception: [K6FDcMae6BnUbxqxn4ga]: version conflict, required seqNo [5], primary term [1]. current document has seqNo [6] and primary term [1]"}
[orc] 09:09:30.743 info
```
Did you get the same thing?

packages/server/lib/controllers/proxy.controller.ts
TBonnin: Not sure it make sense to tie logs to Axios.

bodinsamuel (approved):

PR #2634, opened 2024-08-23
fix: action and webhook logs improvements

TBonnin:
- Added request/response details to action operation payload
- Removed duplicates logs. We were indeed logging more or less the same thing at different levels (handler in jobs, trigger function and/or controller).
- Log messages for actions and webhooks are now shorter and mostly without data (ex: 'Started action A', 'The action failed')
Details like connection, provider, truncated response, error payload are now in the log payload

I want to add request/response details for proxied requests made within a script but it requires modifying the persist API which is ingesting the logs coming from the runner. I will do that in next PR. It would also be useful for the validation warning message

Note: You can ignore the `[Object Object], 100` log in the screenshot below. It is me manually logging something in the test script

Action operation:
Screenshot 2024-08-23 at 11 16 01

Action started log
Screenshot 2024-08-23 at 11 16 45

Action successful log
Screenshot 2024-08-23 at 11 17 15

Failed action operation:
Screenshot 2024-08-23 at 11 17 49

Action failed log
Screenshot 2024-08-23 at 11 18 11

Same for webhook.
Note that operation is not enriched with request/response. @samuel I'll ping you to see if we can add it
Screenshot 2024-08-23 at 11 25 26

https://linear.app/nango/issue/NAN-1525/improve-actions-error-logs

### DIFF
--- a/packages/jobs/lib/execution/action.ts
+++ b/packages/jobs/lib/execution/action.ts
@@ -105,11 +105,6 @@ export async function startAction(task: TaskAction): Promise<Result<void>> {
     }
 }
 export async function handleActionSuccess({ nangoProps }: { nangoProps: NangoProps }): Promise<void> {
-    const logCtx = await logContextGetter.get({ id: String(nangoProps.activityLogId) });
-    const content = `${nangoProps.syncConfig.sync_name} action was run successfully and results are being sent synchronously.`;
-
-    await logCtx.info(content);
-
     const connection: NangoConnection = {
         id: nangoProps.nangoConnectionId!,
         connection_id: nangoProps.connectionId,
@@ -219,5 +214,4 @@ async function onFailure({
             }
         });
     }
-    await logCtx.error(error.message, { error });
 }
--- a/packages/server/lib/controllers/sync.controller.ts
+++ b/packages/server/lib/controllers/sync.controller.ts
@@ -33,7 +33,7 @@ import {
 import type { LogContext } from '@nangohq/logs';
 import { defaultOperationExpiration, logContextGetter } from '@nangohq/logs';
 import type { LastAction } from '@nangohq/records';
-import { isHosted } from '@nangohq/utils';
+import { getHeaders, isHosted } from '@nangohq/utils';
 import { records as recordsService } from '@nangohq/records';
 import type { RequestLocals } from '../utils/express.js';
 import { getOrchestrator } from '../utils/utils.js';
@@ -324,7 +324,6 @@ class SyncController {
                 return;
             } else {
                 span.setTag('nango.error', actionResponse.error);
-                await logCtx.error('Failed to run action', { error: actionResponse.error });
                 await logCtx.failed();
 
                 errorManager.errResFromNangoErr(res, actionResponse.error);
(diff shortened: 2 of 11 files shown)


packages/jobs/lib/execution/action.ts
```diff
@@ -105,11 +105,6 @@ export async function startAction(task: TaskAction): Promise<Result<void>> {
     }
 }
 export async function handleActionSuccess({ nangoProps }: { nangoProps: NangoProps }): Promise<void> {
-    const logCtx = await logContextGetter.get({ id: String(nangoProps.activityLogId) });
```
bodinsamuel, line 108: This will be a relieve for the infra I think 👌🏻

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -110,9 +109,13 @@ export class Orchestrator {
...
         });
         const startTime = Date.now();
-        const workflowId = `${SYNC_TASK_QUEUE}.ACTION:${actionName}.${connection.connection_id}.${uuid()}`;
         try {
-            await logCtx.info(`Starting action workflow ${workflowId} in the task queue: ${SYNC_TASK_QUEUE}`, { input });
+            await logCtx.info(`Starting action '${actionName}'`, {
+                input,
+                action: actionName,
+                connection: connection.connection_id,
+                integration: connection.provider_config_key
```
bodinsamuel, line 117: those 2 things are already logged in the operation (applicable for multiple lines)

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -389,22 +425,31 @@ export class Orchestrator {
...
 
-            const content = `The post connection script workflow ${workflowId} was successfully run. A truncated response is: ${JSON.stringify(res.value, null, 2)?.slice(0, 100)}`;
+            const content = `The post connection script was successfully run.`;
 
-            await logCtx.info(content);
+            await logCtx.info(content, {
+                postConnection: name,
+                connection: connection.connection_id,
+                integration: connection.provider_config_key
+            });
```
bodinsamuel, line 446: res.value is no longer logged?

packages/server/lib/controllers/sync.controller.ts
```diff
@@ -340,6 +339,20 @@ class SyncController {
             }
 
             next(err);
+        } finally {
+            const reqHeaders = getHeaders(req.headers);
```
bodinsamuel, line 343: On your other PR you also need this + the `redacted`
Maybe you should consolidate everything with this new function

bodinsamuel (approved):
Works well, way cleaner 👌🏻 
Minor comments, also there is no message when a sync is starting, it was like that before but it feels inconsistent

packages/server/lib/controllers/sync.controller.ts
TBonnin: yes that's my goal

packages/shared/lib/clients/orchestrator.ts
TBonnin: since post connection scripts are not inline they are not supposed to have meaningful output so I removed the logging of the result. Let me know if you disagree

packages/shared/lib/clients/orchestrator.ts
TBonnin: can we easily show them in the message payload then?

TBonnin:
> also there is no message when a sync is starting, it was like that before but it feels inconsistent

I will work on sync logs improvements when I am done with actions

packages/shared/lib/clients/orchestrator.ts
bodinsamuel: they are shown on top of the Operation drawer

TBonnin: yes but this is to also show this info in the individual message payload.

bodinsamuel: ideally we don't store those information again just for display reason, you can pass the operation to the message drawer if needed

TBonnin: got it. Let me remove them. This info is not absolutely required in the message drawing for now

PR #2637, opened 2024-08-23
feat(integrations): add support for builder.io

hassan254-prog:
## Describe your changes

- Add support for Builder.io

[EXT-128](https://linear.app/nango/issue/EXT-128/add-support-for-builderio)


packages/shared/providers.yaml
```diff
@@ -407,6 +407,28 @@ brex-staging:
...
+builder-io-public:
+    categories:
+        - dev-tools
+        - design
+        - cms
+    auth_mode: API_KEY
+    proxy:
+        base_url: https://${connectionConfig.domain}
+        query:
+            apiKey: ${apiKey}
```
TBonnin, line 430: does they really have 2 API with 2 ways to pass the api key?

hassan254-prog: Yes, they provide users with both `private` and `public` keys. For `public` keys, you append them to the query parameters, while `private` keys are added to the authorization header, as follows; 
image
image

bodinsamuel: might be a way to avoid cors issue in the frontend

packages/webapp/public/images/template-logos/builder-io-private.svg
bodinsamuel, line 1: not a symlink?

bodinsamuel (approved):

TBonnin (approved):

packages/shared/providers.yaml
hassan254-prog: Yes, using the public API key appended to the URL query parameters should help avoid CORS issues on the frontend.

PR #2638, opened 2024-08-27
feat(pagination): [nan-1532] add offset_start_value

khaliqgant:
## Describe your changes
- Add `offset_start_value` optional property for APIs that don't start with offset 0
- Remove unused `offset_name_in_response` value

NAN-1532

### DIFF
--- a/packages/shared/providers.yaml
+++ b/packages/shared/providers.yaml
@@ -3269,7 +3269,6 @@ whoop:
         paginate:
             type: cursor
             offset_name_in_request: nextToken
-            offset_name_in_response: next_token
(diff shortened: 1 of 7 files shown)


packages/shared/providers.yaml
```diff
@@ -3269,7 +3269,6 @@ whoop:
         paginate:
             type: cursor
             offset_name_in_request: nextToken
-            offset_name_in_response: next_token
```
bodinsamuel, line 3272: unused?

bodinsamuel (approved):

PR #2642, opened 2024-08-27
feat(api): prebuilt endpoints

bodinsamuel:
## Describe your changes

Contributes to https://linear.app/nango/issue/NAN-1537/script-v2-integrate-ui

- Feat: New dedicated file for Template deploy
- Feat: simplify payload body for deploy and upgrade to only specify what we need and get the actual data from the backend to make sure it's the correct data
- Fix: reverse API path for template from `/deploy|upgrade/pre-built` -> `/pre-build/deploy|upgrade` just for the sake of REST logic
- Fix: getNangoConfigIdAndLocationFromId was not multi-tenant safe (any id could be passed)
- Fix: missing track_deletes on template deploy
- Fix: removed unused FlowCreate.tsx

## 🧪 Test

- Create Exact Online integration
- Deploy any template (action / sync)

## Summary by CodeRabbit

- **New Features**
  - Introduced tracking for deletions in synchronization configurations.
  - Added new endpoints for deploying and upgrading pre-built flows.
  
- **Bug Fixes**
  - Enhanced error handling across various API endpoints for better clarity and specificity.

- **Refactor**
  - Streamlined data structures and API interactions for flow management.
  - Removed deprecated interfaces and properties to simplify the codebase.

- **Documentation**
  - Updated API endpoint paths and request body structures for clarity.

- **Tests**
  - Modified test cases to reflect the addition of the `track_deletes` property.


packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -414,18 +377,14 @@ export async function deployPreBuilt({
     let nango_config_id: number;
     let provider_config_key: string;
 
-    if (nangoYamlBody) {
```
bodinsamuel, line 417: It felt unused/impossible let me know

packages/webapp/src/pages/Integration/components/EnableDisableSync.tsx
```diff
@@ -88,17 +66,17 @@ export default function EnableDisableSync({
...
         }
-        const res = await createFlow([flow]);
+        const res = await createFlow(body);
 
-        return finalizeEnableSync(res, flow.model_schema);
+        return finalizeEnableSync(res);
     };
 
-    const reEnableFlow = async (flow: ExtendedFlow): Promise<boolean> => {
+    const reEnableFlow = async (flow: any): Promise<boolean> => {
```
khaliqgant, line 79: this `any` is temporary?

packages/shared/lib/services/sync/config/deploy.service.ts
khaliqgant: Yeah, don't see any case for this.

khaliqgant (approved):
non blocking inline question about usage of `any`

packages/webapp/src/pages/Integration/components/EnableDisableSync.tsx
bodinsamuel: yes indeed, I should have added a comment.

packages/shared/lib/services/sync/config/deploy.service.ts
khaliqgant: Looked at this again and I think this PR breaks the admin deploy command https://github.com/NangoHQ/nango/blob/master/packages/cli/lib/services/deploy.service.ts#L79-L99

It is not really used anymore (internal only) so maybe we should remove that functionality completely?

PR #2655, opened 2024-08-30
feat(internal-script-development): [nan-1629] add internal deploy logic

khaliqgant:
## Describe your changes
Adds an internal only command `admin:deploy-internal ${custom-environment}` that when the allow listed account uuid is provided will deploy syncs and actions to a custom environment that is created dynamically with the integration and connections copied from the `dev` environment if they exist.

- Adds a Github action that iterates over all templates and attempts to deploy to a custom environment in the shared dev account
- Adds an additional internal admin endpoint which handles creating the custom environment, copying credentials if they exist and copying connections if they exist, and then deploying

NAN-1629

## Summary by CodeRabbit

- **New Features**
	- Introduced a new GitHub Actions workflow for automating the deployment of integration templates.
	- Added a hidden CLI command for internal deployment of Nango integrations.
	- Implemented a new API endpoint for internal deployment operations.
	- Enhanced the `ConfigService` to copy provider configuration credentials between environments.
	- Added methods to the `ConnectionService` for retrieving and duplicating connections based on environment and configuration IDs.

- **Bug Fixes**
	- Improved error handling and validation for deployment operations.

- **Documentation**
	- Updated type definitions for better clarity and error management in the API.


.github/workflows/deploy-scripts-internally.yaml
```diff
@@ -0,0 +1,43 @@
...
+                  dirs=$(find integration-templates -maxdepth 1 -type d | tail -n +2)
+                  for dir in $dirs; do
+                    integration=$(basename $dir)
+
+                    mkdir -p /tmp/nango-temp/nango-integrations/$integration
+
+                    cp $dir/nango.yaml /tmp/nango-temp/nango-integrations/
+                    cp -r $dir/* /tmp/nango-temp/nango-integrations/$integration/
+
+                    cd /tmp/nango-temp/nango-integrations
```
bodinsamuel, line 35: This is the Nth time we have done that, just because we enforce `nango-integrations` to be the folder, without too many reasons. Maybe at some point, we should just check if there is nango.yaml and that's good

packages/cli/lib/index.ts
```diff
@@ -293,4 +293,15 @@ program
...
+program
+    .command('admin:deploy-internal', { hidden: true })
+    .description('Deploy a Nango integration to the internal Nango dev account')
+    .arguments('environment')
+    .option('-nre, --nango-remote-environment [nre]', 'Optional: Set the Nango remote environment (local, staging, production).')
+    .action(async function (this: Command, environment: string) {
+        const { debug, nangoRemoteEnvironment } = this.opts();
+        const fullPath = process.cwd();
+        await deployService.internalDeploy({ fullPath, environment, debug, options: { env: nangoRemoteEnvironment || 'prod' } });
+    });
```
bodinsamuel, line 305: Feels bad to expose internal stuff in a public api :/

packages/server/lib/controllers/sync/deploy/postDeployInternal.ts
```diff
@@ -0,0 +1,133 @@
...
+        if (!environment) {
+            res.status(500).send({
+                error: { code: 'environment_creation_error', message: 'There was an error creating the environment, please try again' }
+            });
+            return;
+        }
+
+        // since we're making a new environment, we want to make sure the config creds and
+        // connections are copied from the dev environment
+        // so that the syncs and actions can be run right away
```
bodinsamuel, line 71: It feels unnecessary, we have about 60 connections in total that could have been moved manually

packages/shared/lib/services/config.service.ts
```diff
@@ -277,6 +277,31 @@ class ConfigService {
...
+        providerConfigKey: string
+    ): Promise<{ copiedToId: number; copiedFromId: number } | null> {
+        const fromConfig = await this.getProviderConfig(providerConfigKey, fromEnvironmentId);
+
+        if (!fromConfig || !fromConfig.id) {
+            return null;
+        }
+
+        const { id: foundConfigId, ...configWithoutId } = fromConfig;
+        const id = await this.createProviderConfig({
```
bodinsamuel: it's returning * not id

bodinsamuel (approved):
Looks okay. 

I feel like it would be so much better to have a way to create environment programmatically (via the API) but it's not possible unfortunately with the API key being tied to the env :/ 
Most of the code new code is super specific to us and this env, a bit annoying to expose and maintain that.

packages/shared/lib/services/config.service.ts
khaliqgant: True. Updated with https://github.com/NangoHQ/nango/pull/2655/commits/845c6c56b6969b1c75ef14783c6ef7cc5e0de5ab

PR #2660, opened 2024-08-30
chore: pass full log messages from scripts

TBonnin:
## Describe your changes

Please review https://github.com/NangoHQ/nango/pull/2659 first

Sending full message logs from scripts instead of simplified ones allowing:
- to attach request and response when proxying 
- better display of `nango.log('with payload', {a: 1})`. Payload now shows as payload in the UI and it is not appended as string to the message anymore. It is automatically fixing the display of validation output for instance

Screenshot 2024-08-30 at 14 28 51

Screenshot 2024-08-30 at 14 28 04

Screenshot 2024-08-30 at 14 29 19

https://linear.app/nango/issue/NAN-1525/improve-actions-error-logs

## Summary by CodeRabbit

- **New Features**
	- Enhanced logging capabilities with a new structured log format, including additional fields for improved detail and clarity.
- **Bug Fixes**
	- Updated logging mechanisms to improve type safety and streamline logging processes.
- **Chores**
	- Removed the `LogsBuffer` type to simplify the logging structure.
- **Documentation**
	- Updated test cases to validate new log structures, ensuring consistency and clarity in logging behavior.


packages/shared/lib/sdk/sync.ts
```diff
@@ -13,11 +13,20 @@ import type { SyncConfig } from '../models/Sync.js';
...
+export const oldLevelToNewLevel = {
+    debug: 'debug',
+    info: 'info',
+    warn: 'warn',
+    error: 'error',
+    verbose: 'debug',
+    silly: 'debug',
+    http: 'info'
+} as const;
+
```
TBonnin, line 30: duplicating this here because shared doesn't depend on `package/logs`. I will remove it from `logs` once it is not used by `persist` anymore

khaliqgant (approved):

packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
```diff
@@ -46,6 +46,15 @@ import type { AxiosInstance } from 'axios';
...
 import type { DBTeam } from '@nangohq/types';
+export declare const oldLevelToNewLevel: {
+    readonly debug: "debug";
+    readonly info: "info";
+    readonly warn: "warn";
+    readonly error: "error";
+    readonly verbose: "debug";
+    readonly silly: "debug";
+    readonly http: "info";
+};
```
bodinsamuel, line 57: We really need to find a solution for this ugly inlining

packages/shared/lib/sdk/sync.ts
```diff
@@ -676,9 +685,17 @@ export class NangoAction {
...
 
-        const content = stringifyAndTruncateLog(args, 99_000);
+        const [message, payload] = args;
 
-        await this.sendLogToPersist(content, { level, timestamp: Date.now() });
+        await this.sendLogToPersist({
+            type: 'log',
+            level: oldLevelToNewLevel[level],
+            source: 'user',
+            message: String(message),
```
bodinsamuel: message is `any`, and a lot of customers are putting records object or array directly as first argument that's why I added the stringify step; truncating was mostly there to avoid loosing a message if it was too big since customers would have no way to know a message has been skipped

bodinsamuel (commented):
I haven't tested but just a quick review

packages/shared/lib/sdk/sync.ts
bodinsamuel: I should have documented maybe

TBonnin: ok. I will bring back the stringification of message

TBonnin: stringifying message and also correctly handle when payload is an array: https://github.com/NangoHQ/nango/pull/2660/commits/0804014faec2914c605ca6b3acf0698c380fb087

bodinsamuel (approved):
Works well 👌🏻 
Now we can do it, I'm wondering if we should put an empty message instead of stringifying objects. It's shame we can benefits from the payload view
Screenshot 2024-09-05 at 10 34 31

PR #2671, opened 2024-09-03
Feature/add connection token

komyg:
## Describe your changes

- Added a migration to add the `connection_token` column to the `_nango_connections` table.
- Added a migration to add the `connection_token` column to the `_nango_oauth_sessions` table.
- Created a new `authentication` method in the frontend SDK that does not require the `connectionId` and that returns the new `connectionToken` created by the backend.
- Deprecated the old `auth` method in the frontend SDK.
- Updated the `create` component to call the new `authentication` method, that does not require a `connectionId` if no `connectionId` is set by the user.
- Updated the oauth controller to create the `connectionToken` and save it.
- Updated the oauth controller to set the `connectionToken` as the default value for the `connectionId`.
- Updated the `ConnectionService` to save the `connectionToken` on the database.
- The HMAC generation no longer requires a `connectionId`.
- Updated the docs.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `connection_token` feature in the database schema for enhanced connection management.
	- Added a new authentication method to improve the clarity and structure of authentication results.
	- Enhanced connection handling in the OAuth flow with dynamic token generation.
	- Updated connection creation process to better manage connection states and user feedback.
	- Modified API functions to accept optional parameters, improving flexibility.
	- Renamed authentication method in the SDK for consistency and clarity.

- **Bug Fixes**
	- Improved error handling in the authentication process to provide clearer user messages.

- **Documentation**
	- Updated interfaces to include optional `connectionToken` properties for better flexibility in OAuth sessions.
	- Revised documentation to reflect changes in the authentication method and parameters for improved user guidance.

### DIFF
--- a/packages/frontend/lib/index.ts
+++ b/packages/frontend/lib/index.ts
@@ -26,9 +26,25 @@ export interface AuthResult {
     isPending?: boolean;
 }
 
+export interface AuthenticationResult {
+    providerConfigKey: string;
+    connectionToken: string;
+    isPending?: boolean;
+}
+
 interface AuthOptions {
     detectClosedAuthWindow?: boolean; // If true, `nango.auth()` would fail if the login window is closed before the authorization flow is completed
 }
+
+type AuthConfigOptions = (
+    | ConnectionConfig
+    | OAuth2ClientCredentials
+    | OAuthCredentialsOverride
+    | BasicApiCredentials
+    | ApiKeyCredentials
+    | AppStoreCredentials
+) &
+    AuthOptions;
 
 export default class Nango {
     private hostBaseUrl: string;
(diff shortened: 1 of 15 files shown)


packages/frontend/lib/index.ts
```diff
@@ -106,16 +122,41 @@ export default class Nango {
     /**
      * Initiates the authorization process for a connection
      * @param providerConfigKey - The key identifying the provider configuration on Nango
+     * @param options - Optional. Additional options for authorization
+     * @returns A promise that resolves with the authorization result
+     */
+    public async authentication(providerConfigKey: string, options?: AuthConfigOptions): Promise<AuthenticationResult> {
```
khaliqgant, line 128: I think I would prefer a different name`authenticate` perhaps?

PR #2678, opened 2024-09-04
feat(integration-templates): [nan 1643] dynamic tests against integration templates

khaliqgant:
## Describe your changes
- Dynamically run integration template tests given some pre defined mock files. 
- Generate mock files using dryrun by appending a `--save-responses` flag to the dry run command.
- Add support to the dryrun command for the `saveResponses` logic. Note that some methods are missing support like `getMetadata`. I'll add more support as I generate more mock files for other integration tests.
- Add mocks for `calendly` to run tests for it
- Adjust the flows github action to run the tests

# Result
Based on any mocks that were generated via the `--save-responses` flag tests will be generated and run in the `integration-flows` github action:

```
-------------------|---------|----------|---------|---------|-------------------
File               | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s 
-------------------|---------|----------|---------|---------|-------------------
All files          |   95.08 |    81.25 |     100 |   95.08 |                   
 event-invitees.ts |   95.55 |    83.33 |     100 |   95.55 | 9-10              
 event-types.ts    |   93.54 |       75 |     100 |   93.54 | 9-10              
 events.ts         |   95.65 |    83.33 |     100 |   95.65 | 9-10              
-------------------|---------|----------|---------|---------|-------------------
Test Suites: 3 passed, 3 total
Tests:       4 passed, 4 total
Snapshots:   0 total
Time:        53.851 s
Ran all test suites.
```

NAN-1643

## Summary by CodeRabbit

- **New Features**
	- Implemented a new script to automate the generation and execution of integration tests.
	- Introduced functionality for saving API responses to JSON files, enhancing data management.

- **Improvements**
	- Enhanced GitHub Actions workflow to include testing in the Nango integration compilation process.
	- Added a new configuration option for better handling of response saving in the CLI.
	- Updated project structure to improve testing capabilities and integration definitions.
	- Added retry handling for rate limits in the Calendly integration.

- **Bug Fixes**
	- Adjusted data structure definitions for better accuracy and functionality in event management.
	- Removed trailing comma in YAML configuration to prevent potential parsing errors.


tests/integration-templates/jest.config.ts
```diff
@@ -0,0 +1,218 @@
+/**
```
bodinsamuel: Would be great to use vitest if possible, to limit the number of library we use

integration-templates/calendly/mocks/paginate/get/event-invitees/scheduled_events/29413570-8d98-4001-a340-38abdd2a5d39/invitees.json
```diff
@@ -0,0 +1,41 @@
+[
```
bodinsamuel, line 1: do we really need all of that? Just one file is enough for testing I guess?

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -372,6 +379,25 @@ export class DryRunService {
                         break;
                     }
                 }
+
+                if (options.saveResponses) {
+                    const directoryName = `${process.env['NANGO_MOCKS_RESPONSE_DIRECTORY']}${providerConfigKey}`;
+                    if (!fs.existsSync(directoryName)) {
+                        fs.mkdirSync(`${directoryName}/mocks`, { recursive: true });
+                    }
```
bodinsamuel: Unecessary if you are creating another folder recursively just after

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -372,6 +379,25 @@ export class DryRunService {
...
+                    if (!fs.existsSync(`${directoryName}/mocks/${syncName}`)) {
+                        fs.mkdirSync(`${directoryName}/mocks/${syncName}`, { recursive: true });
+                    }
+                    if (nangoProps.rawSaveOutput) {
+                        const filePath = `${directoryName}/mocks/${syncName}/batchSave.json`;
+                        fs.writeFileSync(filePath, JSON.stringify(nangoProps.rawSaveOutput, null, 2));
+                    }
+
+                    if (nangoProps.rawDeleteOutput) {
+                        const filePath = `${directoryName}/mocks/${syncName}/batchDelete.json`;
```
bodinsamuel, line 405: not sure if batchDelete is appropriate, there are plenty of time where it's not a batch

packages/shared/lib/sdk/sync.ts
```diff
@@ -482,7 +488,40 @@ export class NangoAction {
...
+                            });
+                        }
+                        filePath = `${directoryName}/mocks/paginate/${paginateType}/${this.syncConfig?.sync_name}/${fileName}.json`;
+                    }
+                    const responsePath = config.paginate?.response_path;
+                    const paginateData = responsePath ? proxyResponse.data[responsePath] : proxyResponse.data;
+                    fs.writeFileSync(filePath, JSON.stringify(paginateData, null, 2));
+                }
+            }
+            return proxyResponse;
```
bodinsamuel, line 511: It's probably handy to do it here, but mixing that up in the SDK is a bit annoying. 
Is there any way to decouple this with events or by storing it in memory until further processing?

packages/shared/lib/sdk/sync.ts
```diff
@@ -482,7 +488,40 @@ export class NangoAction {
...
+                    if (!fs.existsSync(directoryName)) {
+                        fs.mkdirSync(directoryName);
+                        fs.mkdirSync(`${directoryName}/mocks`);
+                    }
+                    let filePath = `${directoryName}/mocks/${fileName}.json`;
+                    if (config.paginate) {
+                        const paginateType = config.method.toLowerCase();
+                        if (!fs.existsSync(`${directoryName}/mocks/paginate/${paginateType}`)) {
+                            fs.mkdirSync(`${directoryName}/mocks/paginate/${paginateType}`, { recursive: true });
+                        }
```
bodinsamuel: also unnecessary

packages/shared/lib/sdk/sync.ts
```diff
@@ -580,6 +619,20 @@ export class NangoAction {
         if (!cachedConnection || Date.now() - cachedConnection.timestamp > MEMOIZED_CONNECTION_TTL) {
             const connection = await this.nango.getConnection(providerConfigKey, connectionId);
             this.memoizedConnections.set(credentialsPair, { connection, timestamp: Date.now() });
+            if (this.saveResponses) {
+                const directoryName = `${process.env['NANGO_MOCKS_RESPONSE_DIRECTORY']}${providerConfigKey}`;
+                if (directoryName) {
+                    if (!fs.existsSync(directoryName)) {
+                        fs.mkdirSync(`${directoryName}/mocks`, { recursive: true });
+                    }
```
bodinsamuel: also unnecessary

tests/integration-templates/package.json
```diff
@@ -0,0 +1,30 @@
+{
```
bodinsamuel: imo I would make that part of the workspace to avoid multiplication of package-lock.json and complexifying dependency upgrade

tests/integration-templates/tests/generate-tests.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+function generateSyncTest(integration: string, syncName: string, modelName: string) {
+    const data: Record<string, string | Record<string, string | boolean>> = {
+        integration,
+        syncName,
+        modelName
+    };
+
+    if (fs.existsSync(path.resolve(__dirname, `../nango-integrations/${integration}/mocks/${syncName}/batchDelete.json`))) {
```
bodinsamuel: (comment applicable for a lot of places) storing the base file path would be great for readability

tests/integration-templates/tests/generate-tests.ts
```diff
@@ -0,0 +1,82 @@
...
+
+    const result = ejs.render(templateSource, data);
+
+    if (!fs.existsSync(path.resolve(__dirname, '../nango-integrations/tests'))) {
+        fs.mkdirSync(path.resolve(__dirname, '../nango-integrations/tests'), { recursive: true });
+    }
+    const outputPath = path.resolve(__dirname, `../nango-integrations/tests/${data['integration']}-${data['syncName']}.test.ts`);
+    fs.writeFileSync(outputPath, result);
+
+    console.log(`Test file 'tests/${data['integration']}-${data['syncName']}.test.ts' created successfully.`);
```
bodinsamuel: You didn't push the calendly one?

tests/integration-templates/generate-and-run-tests.bash
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
bodinsamuel: Not chmod correctly

bodinsamuel (commented):
Seems to work well, very nice addition 👌🏻 
A few open comments, plus a few feedback:

- I would simply commit the test files, it's hard to debug and run them with the current setup
- Find a solution to not pollute the SDK too much with this
- I was already worried about the footprint of the templates folder, now it's x10 :D

.github/workflows/integration-flows.yaml
```diff
@@ -33,18 +33,34 @@ jobs:
...
                         cd "$dir"
+                        has_mocks=$([ -d "./mocks" ] && echo true || echo false)
                         mkdir -p nango-integrations/$dir
                         mv nango.yaml nango-integrations/
                         find . -maxdepth 1 ! -name '.' ! -name 'nango-integrations' -exec mv -t nango-integrations/$dir {} +
                         cd nango-integrations
                         NANGO_CLI_UPGRADE_MODE=ignore nango compile
+
+                        # run tests if applicable
+                        if [ "$has_mocks" = true ]; then
```
TBonnin: you can directly do `if [ -d "./mocks" ]; then`

packages/shared/lib/sdk/sync.ts
TBonnin: what about moving this logic to a `proxyWithRecorder()` wrapping function. so here in the sdk with simply have 
```
return proxyWithRecorder(this, config);
``` 
basically having some sort of adapter pattern

tests/integration-templates/tests/generate-tests.ts
TBonnin: I like that

tests/integration-templates/jest.config.ts
khaliqgant: Updated to use vitest

integration-templates/calendly/mocks/paginate/get/event-invitees/scheduled_events/29413570-8d98-4001-a340-38abdd2a5d39/invitees.json
khaliqgant: Will keep the files as I already edited them heavily.

packages/cli/lib/services/dryrun.service.ts
khaliqgant: Updated

packages/cli/lib/services/dryrun.service.ts
khaliqgant: Will keep as it doesn't harm at all.

packages/shared/lib/sdk/sync.ts
khaliqgant: Thanks, updated

packages/shared/lib/sdk/sync.ts
khaliqgant: Updated

packages/shared/lib/sdk/sync.ts
khaliqgant: Updated

tests/integration-templates/package.json
khaliqgant: Updated

tests/integration-templates/tests/generate-tests.ts
khaliqgant: Updated

tests/integration-templates/generate-and-run-tests.bash
khaliqgant: Updated

tests/integration-templates/tests/generate-tests.ts
khaliqgant: Pushed it for calendly. It doesn't get used but is more so there for showcasing that we have tests

tests/integration-templates/generate-and-run-tests.bash
khaliqgant: updated

.github/workflows/integration-flows.yaml
khaliqgant: Store as a variable because the location changed

packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
```diff
@@ -272,7 +272,10 @@ export interface NangoProps {
         };
         messages: unknown[];
     } | undefined;
+    rawSaveOutput?: unknown[] | undefined;
+    rawDeleteOutput?: unknown[] | undefined;
```
TBonnin, line 276: do we need to store the batch operation output in memory? Can we do the same as the proxy and store them in different files that can be concatenated when replaying the execution?

packages/integration-template-tests/lib/sync-template.ejs
```diff
@@ -0,0 +1,116 @@
+import { vi, expect, it, describe } from "vitest";
```
TBonnin, line 1: Going back to the idea of having a recording adapter for proxy, batch, etc... to have all the http mocking logic in one place instead of having to duplicate it in the tests. By this I mean something like 
```
Recorder.proxy({proxyConfig: config, option: this.recordingOption) 
// with recordingOption of type = 'record' | 'replay' | 'passthrough'
```
when dry-run --save-responses is set, set 'record' as the recordingOption. The proxy function then make the real http request and save the content in a file. if --save-responses is not set, the recordingOption is 'passthrough', the default value. In thi case `Recorder.proxy` make the real http request and don't save anything

In the tests just set NangoProps.recordingOptions to 'replay', in this case `Recorder.proxy()` get the response from the file.

Same for `batchSave` or any other operations requiring mocking. In batchSave for example, if `this.recordingOptions` is `record`, save the output in a file and when set to `replay` get the data from the file. 

It would mean:
- adding a single new property `recordingOption` to NangoProps/NangoAction
- changes to sync/sdk will be minimal. Basically in the functions that requires mocking replace `await this.nango.proxy(proxyConfig)` into `Recorder.proxy({proxyConfig, nango: this}) or add `Recorder.batchSave({results, nango: this})`, etc...
- no need for mocking anything in the test, just setting the nangoMock.recordingOption to `replay` and calling `await fetchData(nangoMock);`
- all the file manipulation logic (reading and writing) is in the Recorder

What do you think? I am happy to pair if you prefer

integration-templates/calendly/tests/calendly-event-invitees.test.ts
```diff
@@ -0,0 +1,114 @@
+import { vi, expect, it, describe } from 'vitest';
+import type { NangoSync } from '../models.js';
+
+import fetchData from '../calendly/syncs/event-invitees.js';
```
bodinsamuel: paths could be generated relative to template `../syncs/event-invitees.js`

integration-templates/calendly/tests/calendly-event-invitees.test.ts
```diff
@@ -0,0 +1,114 @@
...
+
+    it('should get, map correctly the data and batchSave the result', async () => {
+        await fetchData(nangoMock);
+
+        // eslint-disable-next-line @typescript-eslint/unbound-method
+        expect(nangoMock.batchSave).toHaveBeenCalledWith(batchSaveData, 'EventInvitee');
+    });
+
+    if (batchDeleteData && batchDeleteData.length > 0) {
+        it('should get, map correctly the data and batchDelete the result', async () => {
```
bodinsamuel: There is a built-in feature to do this and also track when not runned

```suggestion
        it.runIf(batchDeleteData && batchDeleteData.length > 0)('should get, map correctly the data and batchDelete the result', async () => {
```

packages/integration-template-tests/package-lock.json
bodinsamuel, line 1: to delete

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -303,6 +306,10 @@ export class DryRunService {
                 },
                 startedAt: new Date()
             };
+            if (options.saveResponses) {
+                nangoProps.rawSaveOutput = [];
+                nangoProps.rawDeleteOutput = [];
```
bodinsamuel, line 322: Is there a time when we do not need to log everything?

packages/shared/lib/sdk/response.saver.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+function getPaginateFilePath(directoryName: string, config: ProxyConfiguration, syncConfig?: SyncConfig): string {
+    const paginateType = config.method?.toLowerCase() || 'get';
+    const fileName = config.endpoint.replace('/', '');
+    const paginatePath = `${directoryName}/mocks/paginate/${paginateType}/${syncConfig?.sync_name}`;
+
+    ensureDirectoryExists(paginatePath);
+
+    if (fileName.includes('/')) {
```
bodinsamuel, line 19: how can it contain / if you replaced all of them L14?

bodinsamuel (commented):
Almost there, this looks much cleaner now 👌🏻 

Personal opinion: For the tests I would try to have a similar flow than we have right now =  consider those as part of the regular test suites if they can run as unit test, or isolate them with a dedicated vitest config like integrations tests at the root level; so there is only one way to test the codebase. The generation would only be there to check that we haven't missed a tests (generate + check git status)

integration-templates/calendly/tests/calendly-event-invitees.test.ts
khaliqgant: This is throwing an error, can you point to the docs where this feature is described? All I see is something like this https://stackoverflow.com/questions/58264344/conditionally-run-tests-in-jest

packages/integration-template-tests/lib/sync-template.ejs
khaliqgant: I like the idea but unsure I'll have the bandwidth to make the update this week. I follow the high level but would want to discuss the finer details probably via pair but do you feel this is an immediate blocking update or a v2 change?

TBonnin: at the end you decide, especially since it is purely internal and could be fully refactored without too much pain but given this is a long term solution that we are gonna build the integrations testing on, I don't think it is a bad thing to spend a little bit of time to get it right and not impact too much the rest of the codebase.

integration-templates/calendly/tests/calendly-event-invitees.test.ts
bodinsamuel: https://vitest.dev/api/#test-runif

khaliqgant: Thanks, updated!

integration-templates/calendly/tests/calendly-event-invitees.test.ts
khaliqgant: Had to restructure tests some, but this has now been updated.

packages/shared/lib/sdk/response.saver.ts
khaliqgant: Didn't replace all but rather the first occurrence of it. So this string: `scheduled_events/7ab23fe1-3047-4a16-9325-93a19e73cb42/invitees` would enter that block

packages/cli/lib/services/dryrun.service.ts
khaliqgant: Yeah, probably but for now we're saving everything for completeness sake

khaliqgant:
> Almost there, this looks much cleaner now 👌🏻
> 
> Personal opinion: For the tests I would try to have a similar flow than we have right now = consider those as part of the regular test suites if they can run as unit test, or isolate them with a dedicated vitest config like integrations tests at the root level; so there is only one way to test the codebase. The generation would only be there to check that we haven't missed a tests (generate + check git status)

Yeah, good call. Updated as you described

packages/integration-template-tests/lib/sync-template.ejs
khaliqgant: Discussed this with Thomas and I am going to revisit the approach next week to potentially mock at the axios level to simplify things further. However will merge this and make that improvement on the next iteration.

bodinsamuel (approved):
Thanks for your grit on this, looks very good ! 🚀

PR #2685, opened 2024-09-05
fix(internal-integrations): [nan-1629] continue on error

khaliqgant:
## Describe your changes
Not all integrations will have an environment in the shared dev account but that shouldn't fail the entire action

NAN-1629

## Summary by CodeRabbit

- **Chores**
	- Improved the deployment process by ensuring that workflow execution continues even if the deployment command encounters an error, enhancing overall robustness.


.github/workflows/deploy-scripts-internally.yaml
```diff
@@ -34,7 +34,7 @@ jobs:
 
                     cd /tmp/nango-temp/nango-integrations
                     npm install zod soap botbuilder
-                    nango admin:deploy-internal template-$integration
+                    nango admin:deploy-internal template-$integration || true
```
bodinsamuel: This will ignore all legitimate errors :/ 
Also why not all integrations would have an env? That seems to be what we want + solve this issue

khaliqgant: The integration might not exist in the shared account meaning there aren't credentials in the shared dev account. For example we don't have algolia credentials in the shared dev account. Because of that the entire script fails and won't continue past that. 

For the context of this action it is OK to ignore this error and other legitimate errors as the only goal is to deploy the scripts that exist in the shared dev account.

TBonnin: can we `have deploy-internal` return a different error code when integration doesn't exist and check the error code?

khaliqgant: added logic to check the error code and return true if `unknown_provider_config`

.github/workflows/deploy-scripts-internally.yaml
```diff
@@ -34,7 +34,14 @@ jobs:
 
                     cd /tmp/nango-temp/nango-integrations
                     npm install zod soap botbuilder
-                    nango admin:deploy-internal template-$integration
+
+                    response=$(nango admin:deploy-internal template-$integration)
+
+                    echo "$response" | jq '.error.code' | grep -q "unknown_provider_config"
+                    if [ $? -eq 0 ]; then
+                      echo "Error: Unknown provider config for $integration"
```
TBonnin:
```suggestion
                      echo "Warning: Unknown provider config for $integration"
```
since we are not failing

.github/workflows/deploy-scripts-internally.yaml
```diff
@@ -34,7 +34,14 @@ jobs:
...
+                      return true
```
TBonnin: does this return mean the script will exit at the first missing integration? is that what we want?

TBonnin (commented):
worth having a pass with https://www.shellcheck.net/ to see if it notices some potential improvements

.github/workflows/deploy-scripts-internally.yaml
khaliqgant: not at all! Thanks, updated

khaliqgant:
> worth having a pass with https://www.shellcheck.net/ to see if it notices some potential improvements

Updated with https://github.com/NangoHQ/nango/pull/2685/commits/816d2c19f7024951d4460eb24cb3b1cb565e7f85. Mostly had suggestions around double quote and checking if the `cd` command exists

TBonnin (approved):
looks good. is `.jq` always installed?

khaliqgant:
> looks good. is `.jq` always installed?

Yup https://github.com/marketplace/actions/install-jq#force

PR #2690, opened 2024-09-06
chore: lower mininum frequency interval to 30s

TBonnin:
## Describe your changes
We want to lower the minimum sync interval to 30s (instead of 5mins)

Note 1: when sync takes longer than frequency interval, a new execution is not started at the end of the interval but once the previous task is finished. Multiple intervals can be skipped

Note 2: people would need to upgrade their CLI because we validate the frequency from the nango.yaml client side

https://linear.app/nango/issue/NAN-1668/lower-max-sync-frequency-to-30-seconds

### DIFF
--- a/packages/server/lib/controllers/sync.controller.ts
+++ b/packages/server/lib/controllers/sync.controller.ts
@@ -764,7 +764,8 @@ class SyncController {
             if (frequency) {
                 const interval = getInterval(frequency, new Date());
                 if (interval instanceof Error) {
-                    res.status(400).send({ message: 'frequency must have a valid format (https://github.com/vercel/ms)' });
+                    const error = new NangoError(interval.message);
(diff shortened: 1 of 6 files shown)


packages/server/lib/controllers/sync.controller.ts
```diff
@@ -764,7 +764,8 @@ class SyncController {
             if (frequency) {
                 const interval = getInterval(frequency, new Date());
                 if (interval instanceof Error) {
-                    res.status(400).send({ message: 'frequency must have a valid format (https://github.com/vercel/ms)' });
+                    const error = new NangoError(interval.message);
+                    res.status(400).send({ message: error.message });
```
TBonnin, line 768: to return the specific error message instead of a generic one

bodinsamuel (approved):

PR #2702, opened 2024-09-09
docs: scripts v2

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1632/improve-sync-query-param-descriptions 
Fixes https://linear.app/nango/issue/NAN-1510/[document]-scripts-ui-v2

- Replace reference of the old tab
Note, that there are a lot of mentions of scripts everywhere. Not sure how deep we want to go and replace them with endpoints @bastienbeurier 

- Fix some links in the UI

### DIFF
--- a/packages/server/lib/controllers/v1/connection/get.ts
+++ b/packages/server/lib/controllers/v1/connection/get.ts
@@ -71,7 +71,7 @@ export const getConnection = asyncWrapper<GetConnection>(async (req, res) => {
 
             res.status(400).send({
                 errorLog,
-                provider: null,
+                provider: null, // TODO: fix this
                 connection: errorConnection
             });
         } else {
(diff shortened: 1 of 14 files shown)


TBonnin (approved):

packages/server/lib/controllers/v1/connection/get.ts
```diff
@@ -71,7 +71,7 @@ export const getConnection = asyncWrapper<GetConnection>(async (req, res) => {
 
             res.status(400).send({
                 errorLog,
-                provider: null,
+                provider: null, // TODO: fix this
```
bastienbeurier, line 74: Just making sure you're aware of this `TODO`

bastienbeurier (approved):

packages/server/lib/controllers/v1/connection/get.ts
bodinsamuel: Yep, it's a technical problem deeper than this but I found out during this PR https://linear.app/nango/issue/NAN-1683/when-a-credentials-refresh-fails-getconnection-is-failing-without-a

PR #2703, opened 2024-09-10
feat(integration-templates): add woocommerce integration templates

hassan254-prog:
## Describe your changes

- Add Woocommerce (orders and customers) integration templates

[EXT-137](https://linear.app/nango/issue/EXT-137/add-woocommerce-templates-to-public-templates)


integration-templates/woocommerce/types.ts
```diff
@@ -0,0 +1,163 @@
+export interface WooCommerceOrder {
```
bodinsamuel, line 1: Just a general question for everybody, should we prefix / find a way to make those internal types less prone to conflict? 
For example, here we are prefixing with `WooCommerce` (and I also did something similar) but ultimately it's hard to dissociate the types from models than those ones. I don't have a solution but maybe somebody can come up with something @khaliqgant @TBonnin

packages/shared/flows.yaml
```diff
@@ -4025,6 +4025,43 @@ integrations:
...
                 language: string
+    woocommerce:
+        syncs:
+            customers:
+                description: |
+                    Periodically fetches all the Woo customers.
+                endpoint: /customers
+                sync_type: full
+                scopes: read
+                track_deletes: false
```
bodinsamuel: no track_deletes?

bodinsamuel (approved):
Looks good 👌🏻

integration-templates/woocommerce/types.ts
TBonnin: Very good question. I have also used some random naming conventions in integrations (for instance Netsuite where models are called `NetsuiteSomething` and logic related types are prefix with `NS_`
I think we should have a different naming convention but the ideas I came up with so far all suck I think (see above :) ) Happy to do some brainstorming though

TBonnin (approved):

packages/shared/flows.yaml
hassan254-prog: aah yes. Updated.

integration-templates/woocommerce/types.ts
khaliqgant: I often do the same, `Order` vs. `SomeApiOrder`.

Other ideas:
- `Order` vs. `_Order` | `_${API_PROVIDER}_Order`
- `Order` vs. `Internal_Order` | `Internal_${API_PROVIDER_Order`

I'm personally OK with `Order` vs. `WooCommerceOrder` FWIW

bodinsamuel: I think WooCommerceOrder is fine but it's usually how we differentiate stuff in `nango.yaml` when you have multiple integrations. It might be safer to use `Internal_Order` indeed, and it needs to apply to all the types inside the file. Otherwise no other ideas :/

PR #2705, opened 2024-09-10
fix(integration-template-upload): [nan-1629] update exit code specific logic and don't bail on success

khaliqgant:
## Describe your changes
Fix logic so the following happens:
```
if error and if it equals unknown_provider_config
    echo "warning..."
    continue
if error
    echo "error"
    exit

if no error and is successful
    do nothing
```
Previously, even on success it was throwing an error

NAN-1629


TBonnin (approved):
sorry I missed that last time.

PR #2706, opened 2024-09-10
fix(xero-auth): [nan-1684] handle multiple tenants

khaliqgant:
## Describe your changes
Xero allows for multiple organizations but in our post connection script we just grabbed the first one. Decode the JWT as described in https://developer.xero.com/documentation/guides/oauth2/pkce-flow/#5-check-the-tenants-youre-authorized-to-access to find the `authEventId` to use the appropriate tenant to set to the `connectionConfig`

NAN-1684

### DIFF
--- a/packages/server/lib/hooks/connection/xero-post-connection.ts
+++ b/packages/server/lib/hooks/connection/xero-post-connection.ts
@@ -2,6 +2,10 @@ import type { InternalNango as Nango } from './post-connection.js';
 import type { OAuth2Credentials } from '@nangohq/types';
 import jwt from 'jsonwebtoken';
 import { isAxiosError } from 'axios';
+import { getLogger } from '@nangohq/utils';
+import { z } from 'zod';
+
+const logger = getLogger('post-connection:xero');
 
 interface XeroConnection {
     id: string;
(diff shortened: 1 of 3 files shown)


packages/server/lib/hooks/connection/xero-post-connection.ts
```diff
@@ -1,19 +1,69 @@
...
 
-    await nango.updateConnectionConfig({ tenant_id });
+    if (response.data.length === 1) {
+        tenant_id = response.data[0]['tenantId'];
+    } else {
+        // decode the jwt to find the corresponding tenant if there are multiple
+        const credentials = connection.credentials as OAuth2Credentials;
+        const token = credentials.access_token;
+        const decoded = jwt.decode(token, { complete: true });
+        if (!decoded) {
```
TBonnin: Can we also consider that a decoded token that is string is non actionable. `if(!decoded || typeof decoded === 'string')`

packages/server/lib/hooks/connection/xero-post-connection.ts
```diff
@@ -1,19 +1,69 @@
...
+            return;
```
TBonnin, line 62: would it make sense to log something to make it easier to debug?

packages/server/lib/hooks/connection/xero-post-connection.ts
```diff
@@ -1,19 +1,69 @@
...
+        }
+        const payload = decoded.payload as XeroJWTPayload;
```
TBonnin: Can we validate the token payload instead of casting and assuming it is always well-formed. Something like:
```
        const XeroJWTPayloadSchema = z.object({
            authentication_event_id: z.string()
        });
        ...
        const decoded = jwt.decode(token, { complete: true });
        const decodedToken = XeroJWTPayloadSchema.safeParse(decoded);
        if (decodedToken.success) {
            const authentication_event_id = decodedToken.data.authentication_event_id;
            const foundConnection = response.data.find((connection: XeroConnection) => connection.authEventId === authentication_event_id);

            if (foundConnection) {
                tenant_id = foundConnection['tenantId'];
            }
        } else {
            // TODO: logging the error
        }
```

packages/server/lib/hooks/connection/xero-post-connection.ts
```diff
@@ -1,19 +1,69 @@
...
+        const foundConnection = response.data.find((connection: XeroConnection) => connection.authEventId === authentication_event_id);
+
+        if (foundConnection) {
+            await nango.updateConnectionConfig({ tenant_id: foundConnection['tenantId'] });
+        }
+    }
+
+    if (tenant_id) {
+        await nango.updateConnectionConfig({ tenant_id });
+    }
```
TBonnin, line 81: should we log something if there is no tenant?

TBonnin (approved):
made a couple of comments about logging and token payload validation.

packages/server/lib/hooks/connection/xero-post-connection.ts
khaliqgant: Updated

packages/server/lib/hooks/connection/xero-post-connection.ts
khaliqgant: Indeed, updated.

packages/server/lib/hooks/connection/xero-post-connection.ts
khaliqgant: Updated

packages/server/lib/hooks/connection/xero-post-connection.ts
khaliqgant: Updated

PR #1584, opened 2024-01-30
[nan-280] add support for multi part

khaliqgant:
## Describe your changes
We aren't handling multipart/form-data POST requests correctly via the proxy, this adds in support for that and also adds in support for xml (!) requests.

## Issue ticket number and link
NAN-280

## Checklist before requesting a review (skip if just adding/editing APIs & templates)
- [ ] I added tests, otherwise the reason is: 
- [ ] I added observability, otherwise the reason is:
- [ ] I added analytics, otherwise the reason is:

PR #2708, opened 2024-09-10
fix(proxy): file upload via multipart/form-data or application/octet-stream

TBonnin:
- fixing /proxy to allow file upload with `Content-Type: multipart/form-data`
- adding support for uploading file in proxy with `Content-Type: application/octet-stream`
- fixing node client to upload file via `nango.proxy()`

https://linear.app/nango/issue/NAN-1473/enable-file-upload-with-nango-proxy

### DIFF
--- a/packages/node-client/lib/index.ts
+++ b/packages/node-client/lib/index.ts
@@ -705,6 +705,10 @@ export class Nango {
             ...customPrefixedHeaders
         };
 
+        if (customHeaders?.['Content-Type']) {
+            headers['Content-Type'] = customHeaders['Content-Type'];
+        }
+
         if (retries) {
             headers['Retries'] = retries;
         }
--- a/packages/shared/lib/models/Proxy.ts
+++ b/packages/shared/lib/models/Proxy.ts
@@ -4,12 +4,25 @@ import type { BasicApiCredentials, ApiKeyCredentials, AppCredentials, TbaCredent
 import type { Connection } from './Connection.js';
 import type { Template as ProviderTemplate } from '@nangohq/types';
 
+export interface File {
+    fieldname: string;
+    originalname: string;
+    encoding: string;
+    mimetype: string;
+    size: number;
+    destination: string;
+    filename: string;
+    path: string;
+    buffer: Buffer;
+}
+
 interface BaseProxyConfiguration {
     providerConfigKey: string;
     connectionId: string;
     endpoint: string;
     retries?: number;
     data?: unknown;
+    files?: File[];
     headers?: Record<string, string>;
     params?: string | Record<string, string | number>;
     paramsSerializer?: ParamsSerializerOptions;
--- a/packages/shared/lib/services/proxy.service.ts
+++ b/packages/shared/lib/services/proxy.service.ts
@@ -170,6 +170,12 @@ class ProxyService {
             Object.keys(data as any).forEach((key) => {
                 formData.append(key, (data as any)[key]);
             });
+            for (const file of externalConfig.files || []) {
+                formData.append(file.fieldname, file.buffer, {
+                    filename: file.originalname,
+                    contentType: file.mimetype
+                });
+            }
 
             data = formData;
         }
--- a/packages/node-client/bin/proxy-octet-stream.js
+++ b/packages/node-client/bin/proxy-octet-stream.js
@@ -7,8 +7,6 @@ const nango = new Nango({ host: 'http://localhost:3003', secretKey: args[0] });
 const filePath = './test.pdf';
 const fileName = filePath.split('/').pop();
 const buffer = fs.readFileSync(filePath);
-const formData = new FormData();
-formData.append('file', new Blob([buffer]), fileName);
 
 nango
     .proxy({
@@ -16,10 +14,11 @@ nango
         connectionId: 'u',
         baseUrlOverride: 'http://localhost:3009',
         method: 'POST',
-        endpoint: '/upload/multipart',
-        data: formData,
+        endpoint: '/upload/octet-stream',
+        data: buffer,
         headers: {
-            'Content-Type': 'multipart/form-data'
+            'Content-Type': 'application/octet-stream',
+            'X-File-Name': fileName
         }
     })
     .then((response) => {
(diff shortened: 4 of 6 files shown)


packages/node-client/bin/proxy-multi-part-form.js
```diff
@@ -4,23 +4,22 @@ const args = process.argv.slice(2);
...
+formData.append('file', new Blob([buffer]), fileName);
 
 nango
     .proxy({
         providerConfigKey: 'unauthenticated',
         connectionId: 'u',
         baseUrlOverride: 'http://localhost:3009',
         method: 'POST',
-        endpoint: '/',
+        endpoint: '/upload/multipart',
```
TBonnin, line 19: maybe I should commit the python webserver I used to received the uploaded file. Let me know what you think

packages/node-client/lib/index.ts
```diff
@@ -705,6 +705,10 @@ export class Nango {
             ...customPrefixedHeaders
         };
 
+        if (customHeaders?.['Content-Type']) {
+            headers['Content-Type'] = customHeaders['Content-Type'];
+        }
+
```
TBonnin, line 711: without setting the Content-Type, axios below assumes `application/x-www-form-urlencoded` when data is a Buffer

packages/shared/lib/services/proxy.service.ts
```diff
@@ -170,6 +170,12 @@ class ProxyService {
             Object.keys(data as any).forEach((key) => {
                 formData.append(key, (data as any)[key]);
             });
+            for (const file of externalConfig.files || []) {
```
khaliqgant, line 173: I think you can remove the above 

```
            Object.keys(data as any).forEach((key) => {
                formData.append(key, (data as any)[key]);
            });
```
as this was my quick attempt to add in the support for file uploads in https://github.com/NangoHQ/nango/pull/1584

TBonnin: multipart/form-data data can also contain key/value pairs (ie: form input values) that are not files but end up in the req.body and therefore data, so we need to do both.

khaliqgant: Cool

packages/node-client/bin/proxy-multi-part-form.js
bodinsamuel: I think we need a bigger initiative to help test more stuff than just this, a test website would be nice

packages/shared/lib/models/Proxy.ts
```diff
@@ -4,12 +4,25 @@ import type { BasicApiCredentials, ApiKeyCredentials, AppCredentials, TbaCredent
 import type { Connection } from './Connection.js';
 import type { Template as ProviderTemplate } from '@nangohq/types';
 
+export interface File {
```
bodinsamuel, line 7: I'm surprised it's not provided by express?

bodinsamuel (commented):
Looks good and simpler fix than expected. (nb: Not tested yet)

packages/shared/lib/models/Proxy.ts
TBonnin: it is provided by the multer middleware but I didn't want to add it as a dependency of shared. https://expressjs.com/en/resources/middleware/multer.html

bodinsamuel: I think it's fine if it's a dev deps (we tested with `@nangohq/*`) as you want ☺️

bodinsamuel (commented):
I tried with Google Drive but could not make it work (tried in insomnia and local).
I think they expect the actual content-type of the file uploaded (plus some others stuff maybe I don't know). 

Which integration can I try?


```js
import fs from 'fs';

async function uploadBasic() {
    const filePath = './photo.jpg';
    const buffer = fs.readFileSync(filePath);

    const options = {
        method: 'POST',
        headers: {
            'Content-Type': 'image/jpeg', // Do not work with application/octet either
            'Connection-Id': 'test-connection-id',
            'Provider-Config-Key': 'google-drive',
            Authorization: '
        },
        data: buffer
    };

    const res = await fetch('http://localhost:3003/proxy/drive/v3/files?=string&uploadType=media', options);
    console.log(res);
}

await uploadBasic();
```

TBonnin:
> I tried with Google Drive but could not make it work (tried in insomnia and local). I think they expect the actual content-type of the file uploaded (plus some others stuff maybe I don't know).
> 
> Which integration can I try?
> 
> ```js
> import fs from 'fs';
> 
> async function uploadBasic() {
>     const filePath = './photo.jpg';
>     const buffer = fs.readFileSync(filePath);
> 
>     const options = {
>         method: 'POST',
>         headers: {
>             'Content-Type': 'image/jpeg', // Do not work with application/octet either
>             'Connection-Id': 'test-connection-id',
>             'Provider-Config-Key': 'google-drive',
>             Authorization: '
>         },
>         data: buffer
>     };
> 
>     const res = await fetch('http://localhost:3003/proxy/drive/v3/files?=string&uploadType=media', options);
>     console.log(res);
> }
> 
> await uploadBasic();
> ```

the endpoint is `upload/drive/v3/files`. You are missing the `/upload` part

Also one can specify the `Nango-Proxy-Content-Type` to specify the value that should be used by the proxied request if different from the request to nango API. In the case of Google Drive it doesn't seem necessary

This works for me for instance:
```
curl -X POST 'http://localhost:3003/proxy/upload/drive/v3/files?uploadType=media' --header 'Authorization: Bearer XXX' -H 'Connection-Id: gd' -H 'Provider-Config-Key: google-drive' -H 'Content-Type: application/octet-stream' --data-binary @/Users/thomas/Desktop/j.jpg
```

bodinsamuel:
It uploads a json when I do that
```sh
curl -X POST 'http://localhost:3003/proxy/upload/drive/v3/files?uploadType=media' 
--header 'Authorization: Bearer ' 
-H 'Connection-Id: test-connection-id' 
-H 'Provider-Config-Key: google-drive' 
-H 'Content-Type: application/octet-stream' 
--data-binary photo.jpg

{
"kind":"drive#file",
"id":"1Kaf89VsDsTl1WqH2J2von1gwv-dIUeD8",
"name":"Untitled",
"mimeType":"application/json"
}
```
Screenshot 2024-09-12 at 12 00 39

---

If I pass `Nango-Proxy-Content-Type` it's now the correct type, but the image is corrupted (only 2bytes)

TBonnin:
> ```shell
> --data-binary photo.jpg
> ```

You are missing the `@`: `--data-binary @photo.jpg`

bodinsamuel (approved):
indeed my bad 🤦🏻

PR #2709, opened 2024-09-10
chore: add sync_config_id column to sync_config table

TBonnin:
linking sync_config from sync currently requires to check the connection table and match on connectionId, environmentId and syncName. Adding a sync_config_id foreign key to the sync table for more performant queries and easier debugging

There were actually a bug a couple of month ago that was caused by a missing join where fetching syncs: https://github.com/NangoHQ/nango/pull/2518/files. This change should make such bug less likely to happen.

https://linear.app/nango/issue/NAN-1442/add-foreign-key-nango-syncs-nango-sync-configs

### DIFF
--- a/packages/database/lib/migrations/20240910183251_sync_config_id.cjs
+++ b/packages/database/lib/migrations/20240910183251_sync_config_id.cjs
@@ -1,0 +1,30 @@
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = function (knex) {
+    return knex.schema
+        .alterTable('_nango_syncs', function (table) {
+            table.integer('sync_config_id').unsigned().references('id').inTable('_nango_sync_configs').onDelete('CASCADE');
+        })
+        .then(() => {
+            return knex.raw(`
+            UPDATE _nango_syncs AS syncs
+            SET sync_config_id = sync_configs.id
+            FROM _nango_sync_configs AS sync_configs
+            JOIN _nango_connections AS connections ON sync_configs.nango_config_id = connections.config_id
+            WHERE syncs.name = sync_configs.sync_name
+              AND syncs.nango_connection_id = connections.id
+              AND sync_configs.environment_id = connections.environment_id
+              AND syncs.sync_config_id IS NULL;
+        `);
+        });
+};
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.down = function (knex) {
+    return knex.schema.alterTable('_nango_syncs', function (table) {
+        table.dropColumn('sync_config_id');
+    });
+};
--- a/packages/database/lib/migrations/20240910183251_sync_config_id.cjs
+++ b/packages/database/lib/migrations/20240910183251_sync_config_id.cjs
@@ -15,6 +15,7 @@ exports.up = function (knex) {
             WHERE syncs.name = sync_configs.sync_name
               AND syncs.nango_connection_id = connections.id
               AND sync_configs.environment_id = connections.environment_id
+              AND sync_configs.type = 'sync'
               AND syncs.sync_config_id IS NULL;
         `);
         });
--- a/packages/server/lib/controllers/v1/flows/id/patchFrequency.ts
+++ b/packages/server/lib/controllers/v1/flows/id/patchFrequency.ts
@@ -55,7 +55,7 @@ export const patchFlowFrequency = asyncWrapper<PatchFlowFrequency>(async (req, r
         return;
     }
 
-    const syncs = await getSyncsBySyncConfigId(valParams.data.id);
+    const syncs = await getSyncsBySyncConfigId(environment.id, valParams.data.id);
     for (const sync of syncs) {
         const updated = await orchestrator.updateSyncFrequency({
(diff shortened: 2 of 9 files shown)


packages/database/lib/migrations/20240910183251_sync_config_id.cjs
```diff
@@ -0,0 +1,30 @@
...
+        .then(() => {
+            return knex.raw(`
+            UPDATE _nango_syncs AS syncs
+            SET sync_config_id = sync_configs.id
+            FROM _nango_sync_configs AS sync_configs
+            JOIN _nango_connections AS connections ON sync_configs.nango_config_id = connections.config_id
+            WHERE syncs.name = sync_configs.sync_name
+              AND syncs.nango_connection_id = connections.id
+              AND sync_configs.environment_id = connections.environment_id
+              AND syncs.sync_config_id IS NULL;
```
TBonnin, line 19: syncs only contains 25k entries so this update should be quite quick

packages/shared/lib/services/sync/sync.service.ts
```diff
@@ -370,18 +370,12 @@ export const findSyncByConnections = async (connectionIds: number[], sync_name:
...
         .from<Sync>(TABLE)
-        // Sync table doesn't have a unique foreign key to sync config
-        // so we need to join on the name
-        // and verify the sync connection environment
-        .join(SYNC_CONFIG_TABLE, `${TABLE}.name`, `${SYNC_CONFIG_TABLE}.sync_name`)
-        .join(CONNECTIONS_TABLE, `${CONNECTIONS_TABLE}.id`, `${TABLE}.nango_connection_id`)
+        .join(SYNC_CONFIG_TABLE, `${TABLE}.sync_config_id`, `${SYNC_CONFIG_TABLE}.id`)
         .where({
-            [`${CONNECTIONS_TABLE}.environment_id`]: environmentId,
-            [`${SYNC_CONFIG_TABLE}.environment_id`]: environmentId,
```
TBonnin, line 384: simpler join

packages/server/lib/controllers/v1/flows/id/patchFrequency.ts
```diff
@@ -55,7 +55,7 @@ export const patchFlowFrequency = asyncWrapper<PatchFlowFrequency>(async (req, r
         return;
     }
 
-    const syncs = await getSyncsBySyncConfigId(environment.id, valParams.data.id);
+    const syncs = await getSyncsBySyncConfigId(valParams.data.id);
```
bodinsamuel: Since we don't do permissions check yet and syncConfigId is auto-incremented it's not safe to query without env id. Happy to discuss but I would keep it especially since this value is available without join

packages/database/lib/migrations/20240910183251_sync_config_id.cjs
```diff
@@ -0,0 +1,30 @@
...
+        .alterTable('_nango_syncs', function (table) {
+            table.integer('sync_config_id').unsigned().references('id').inTable('_nango_sync_configs').onDelete('CASCADE');
+        })
+        .then(() => {
+            return knex.raw(`
+            UPDATE _nango_syncs AS syncs
+            SET sync_config_id = sync_configs.id
+            FROM _nango_sync_configs AS sync_configs
+            JOIN _nango_connections AS connections ON sync_configs.nango_config_id = connections.config_id
+            WHERE syncs.name = sync_configs.sync_name
```
bodinsamuel, line 15: missing sync type

bodinsamuel (commented):
(A bit optimistic to push code change with the migration)

packages/server/lib/controllers/v1/flows/id/patchFrequency.ts
TBonnin: ok. I'll add it back

TBonnin:
> (A bit optimistic to push code change with the migration)

is it? the code cannot execute unless the migration is successful

packages/server/lib/controllers/v1/flows/id/patchFrequency.ts
TBonnin: [ae30d3e](https://github.com/NangoHQ/nango/pull/2709/commits/ae30d3ebea0799506c6abaea3dec2fb637001cc4)

packages/database/lib/migrations/20240910183251_sync_config_id.cjs
TBonnin: [7691921](https://github.com/NangoHQ/nango/pull/2709/commits/76919213995ebd68a8dd80c773d9cc01755b81dc)

bodinsamuel (approved):

bodinsamuel:
If the migration fails to fill all rows (for some reason) you now have two problems to solve (the db plus some queries failing). 
I think it's just a good practice to split migration and usage

TBonnin:
closing to split migration and code changes:
https://github.com/NangoHQ/nango/pull/2711
https://github.com/NangoHQ/nango/pull/2712

PR #2713, opened 2024-09-11
fix: allow special characters in provider config key

TBonnin:
for legacy reason some customers use special chars in their provider config key

Characters have been identified running the following sql query against the prod replica db
```
SELECT DISTINCT unnest(string_to_array(provider_config_key, NULL)) AS unique_character
FROM nango._nango_connections
WHERE provider_config_key IS NOT NULL
ORDER BY unique_character;
```

### DIFF
--- a/packages/server/lib/helpers/validation.ts
+++ b/packages/server/lib/helpers/validation.ts
@@ -6,7 +6,7 @@ export const providerSchema = z
     .max(255);
 export const providerConfigKeySchema = z
     .string()
-    .regex(/^[a-zA-Z0-9_-]+$/)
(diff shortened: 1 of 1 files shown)


bodinsamuel (approved):
Thanks ! Needs to be fixed in the schema too I think

TBonnin:
> Thanks ! Needs to be fixed in the schema too I think

which schema?
merging for now to unblock customers but I will follow up with any required changes

bodinsamuel:
packages/cli/lib/nango.yaml.schema.v2.json
But turns out I only added key validation for scripts name here so all good

PR #2722, opened 2024-09-13
feat(integration-templates): add discourse integration template

hassan254-prog:
## Describe your changes

- Add discourse integration template


integration-templates/discourse/nango.yaml
```diff
@@ -0,0 +1,17 @@
+integrations:
+    discourse:
+        syncs:
+            users:
+                description: |
+                    Fetches a list of active users from Discourse.
+                endpoint: /discourse/users
+                sync_type: full
+                runs: every 1 hour
```
khaliqgant, line 9: track_deletes: true

hassan254-prog: Thank you Khaliq. Updated with [commit](https://github.com/NangoHQ/nango/pull/2722/commits/4809e695a54651800f1f90a89dbf86d0880b5335)

integration-templates/discourse/nango.yaml
```diff
@@ -0,0 +1,18 @@
+integrations:
+    discourse:
+        syncs:
+            users:
+                description: |
+                    Fetches a list of active users from Discourse.
```
bodinsamuel, line 6: I don't think we filter
```suggestion
                    Fetches a list of users from Discourse.
```

bodinsamuel (approved):

integration-templates/discourse/nango.yaml
hassan254-prog: The implemented endpoint is filtering for active users only  `/active`.

bodinsamuel: ah yeah I didn't saw it was in the path of the endpoint my bad. 
Is it still relevant to have `active` in the User model then? or maybe the sync should be renamed to users-active. Not suggesting to change anything but wondering if we want to be explicit

hassan254-prog: I think that would be nice to explicitly name the script as `active-users` as it only fetches for active users only.

bodinsamuel (approved):

PR #2728, opened 2024-09-15
docs: add tip to google-mail.mdx

tonyd3:
## adding a tip to gmail integration


bodinsamuel:
Hey, thanks for your contribution. 
I'm not sure it's relevant since it must be done for all integrations, but maybe we should actually document that for all integrations, wdyt @bastienbeurier ?

bastienbeurier:
> Hey, thanks for your contribution. I'm not sure it's relevant since it must be done for all integrations, but maybe we should actually document that for all integrations, wdyt @bastienbeurier ?

This is documented in this [guide](https://docs.nango.dev/integrate/guides/authorize-an-api):

> When registering, the API provider will prompt you for the Callback URL.

So I think it's okay to not document in API-specific gotchas as well.

tonyd3:
sg!

PR #2730, opened 2024-09-16
fix(github-app-oauth): [nan-1733] GitHub app missing secret and update iss in case the app-id is missing

khaliqgant:
## Describe your changes
The webhook secret was missing in the case of a Github App. Fallback to the connectionConfig value for app_id if missing in the config.

NAN-1733


bodinsamuel (approved):
Thanks 👌🏻

PR #2731, opened 2024-09-16
fix(sdk): handle circular json in log()

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1735/log-do-not-work-with-circular-ref

- Correctly handle circular ref in json when calling `log()`
It was previously not an issue because we were stringify everything in the message which already handled the circular ref. But now we have split both usage and let axios handle stringification, which apparently doesn't work. 

- Also took the opportunity to redac Authorization header

### DIFF
--- a/packages/shared/lib/sdk/sync.ts
+++ b/packages/shared/lib/sdk/sync.ts
@@ -14,7 +14,7 @@ import type { SyncConfig } from '../models/Sync.js';
 import type { RunnerFlags } from '../services/sync/run.utils.js';
 import { validateData } from './dataValidation.js';
 import { NangoError } from '../utils/error.js';
-import { stringifyAndTruncateLog } from './utils.js';
+import { stringifyAndTruncateLog, stringifyObject } from './utils.js';
 import type { DBTeam, MessageRowInsert } from '@nangohq/types';
 
 const logger = getLogger('SDK');
@@ -824,10 +824,10 @@ export class NangoAction {
                             Authorization: `Bearer ${this.nango.secretKey}`,
                             'Content-Type': 'application/json'
                         },
-                        data: {
+                        data: stringifyObject({
                             activityLogId: this.activityLogId,
                             log
-                        }
+                        })
                     });
                 },
                 { retry: httpRetryStrategy }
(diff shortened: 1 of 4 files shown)


packages/shared/lib/sdk/sync.ts
```diff
@@ -824,10 +824,10 @@ export class NangoAction {
                             Authorization: `Bearer ${this.nango.secretKey}`,
                             'Content-Type': 'application/json'
                         },
-                        data: {
+                        data: stringifyObject({
```
bodinsamuel, line 827: I was not very confident about giving a string to axios but seems to work, would be great if you can try on your side just to confirm 😬

packages/shared/lib/sdk/utils.unit.test.ts
```diff
@@ -31,8 +31,22 @@ describe('stringifyAndTruncateLog', () => {
...
     });
+
+    it('should handle circular', () => {
+        const obj: Record<string, any> = { foo: 'bar' };
+        obj['circular'] = obj;
+        const str = stringifyAndTruncateLog(obj);
+        expect(str).toStrictEqual('{"circular":"[Circular]","foo":"bar"}');
+    });
+
+    it('should redac know keys', () => {
```
TBonnin:
```suggestion
    it('should redac known keys', () => {
```

packages/shared/lib/sdk/sync.ts
TBonnin: It think it works because we explicitly set the content type to `application/json` and don't rely on Axios automatic content-type inference

packages/shared/lib/sdk/sync.unit.test.ts
```diff
@@ -457,25 +457,26 @@ describe('Log', () => {
...
-            },
+            data: expect.any(String),
             headers: {
                 Authorization: 'Bearer ***',
                 'Content-Type': 'application/json'
             },
             method: 'POST',
             url: '/environment/1/log'
         });
+        expect(JSON.parse((mock.mock.calls as any)[0][0].data)).toStrictEqual({
```
TBonnin, line 468: `(mock.mock.calls as any)[0][0].data)` that's scary 😨

TBonnin (approved):

khaliqgant (approved):

packages/shared/lib/sdk/sync.unit.test.ts
bodinsamuel: yep, not a fan but since everything is stringified now there was no way to use partial match without accessing the value directly :/

packages/shared/lib/sdk/sync.ts
bodinsamuel: yep most probably

PR #2732, opened 2024-09-16
feat(api): GET /providers, GET /providers/:provider

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1711/add-providers-to-return-auth-type-and-connectionconfig

I need this endpoint for the upcoming UI. Turns out we already had `/provider` (sic) but undocumented. To avoid breaking something somewhere, I will deprecate this endpoint.

- Add `GET /providers(?search=query)`
- Add `GET /providers/:provider`
- Split config / provider services
We were mixing `providers.yaml` loading and config stuff. I also changed the code to stop throwing for proper error handling

- Rename interface name from `Template` to `Provider`
Not entirely sure why it was named like this, please let me know. 

- Documentation
Add the official page in doc and OpenAPI spec

## 🧪 Test

```
curl --request GET \
  --url http://localhost:3003/providers \
  --header 'Authorization: __API__'
```

Screenshot 2024-09-16 at 15 02 04


docs-v2/reference/api/providers/list.mdx
```diff
@@ -0,0 +1,35 @@
...
+        "paginate": {
+          "type": "cursor",
+          "cursor_path_in_response": "paging.next.after",
+          "limit_name_in_request": "limit",
+          "cursor_name_in_request": "after",
+          "response_path": "results"
+        }
+      },
+      "docs": "https://docs.nango.dev/integrations/all/hubspot"
+    }
```
TBonnin, line 40: I would show a second one, just to make it clearer data is an array

packages/server/lib/controllers/apiAuth.controller.ts
```diff
@@ -86,7 +87,13 @@ class ApiAuthController {
                 return;
             }
 
-            const template = configService.getTemplate(config.provider);
+            const template = getProvider(config.provider);
```
TBonnin: since you are renaming the function I would also rename the variable to `provider` and also the error msg and code below to be consistent

packages/server/lib/controllers/config.controller.ts
```diff
@@ -190,13 +193,17 @@ class ConfigController {
...
 
             const config = await configService.getProviderConfig(providerConfigKey, environment.id);
-
             if (!config) {
                 errorManager.errRes(res, 'unknown_provider_config');
                 return;
             }
 
-            const providerTemplate = configService.getTemplate(config.provider);
+            const providerTemplate = getProvider(config.provider);
```
TBonnin: let's always name this variable the same in all files.

packages/server/lib/controllers/providers/getProviders.ts
```diff
@@ -0,0 +1,43 @@
...
+        for (const providerName of Object.keys(providers)) {
+            if (reg.test(providerName)) {
+                filtered.push({ ...providers[providerName]!, name: providerName });
+            }
+        }
+    } else {
+        filtered = Object.entries(providers).map(([name, provider]) => {
+            return { ...provider, name };
+        });
+    }
```
TBonnin, line 40: you can do something like this I think
```
    const filtered = Object.entries(providers)
        .filter(([name]) => !queries.search || new RegExp(queries.search, 'i').test(name))
        .map(([name, provider]) => ({ ...provider, name }));
```

packages/server/lib/helpers/validation.ts
```diff
@@ -4,6 +4,10 @@ export const providerSchema = z
     .string()
     .regex(/^[a-zA-Z0-9_-]+$/)
     .max(255);
+export const providerNameSchema = z
+    .string()
+    .regex(/^[a-zA-Z0-9 _-]+$/)
```
TBonnin: I don't think we allow space or underscore in provider name. Only alphanumeric + dash

packages/shared/lib/services/providers.ts
```diff
@@ -0,0 +1,75 @@
...
+}
+
+export function checkProviderExists(provider: string) {
+    getProviders();
+    return providers && provider in providers;
+}
+
+export function getProvider(provider: string): Provider | null {
+    getProviders();
+    return providers && provider in providers ? providers[provider]! : null;
```
TBonnin: something like might be easier to read
```
export function getProvider(provider: string): Provider | null {
    const allProviders = getProviders();
    return allProviders?.[provider] ?? null;
}

```

TBonnin (approved):
some naming comments but otherwise looks good

packages/server/lib/helpers/validation.ts
bodinsamuel: indeed bad c/c

packages/server/lib/controllers/apiAuth.controller.ts
bodinsamuel: yeah I wanted to limit changes, but happy to do it 👌🏻

packages/server/lib/controllers/appAuth.controller.ts
```diff
@@ -86,10 +87,17 @@ class AppAuthController {
 
             await logCtx.enrichOperation({ integrationId: config.id!, integrationName: config.unique_key, providerName: config.provider });
 
-            const template = configService.getTemplate(config.provider);
-            const tokenUrl = typeof template.token_url === 'string' ? template.token_url : (template.token_url?.['APP'] as string);
+            const provider = getProvider(config.provider);
+            if (!provider) {
+                await logCtx.error('Unknown provider template');
```
TBonnin: what about simply `Unknown provider`?

TBonnin (approved):

packages/server/lib/controllers/appAuth.controller.ts
bodinsamuel: true changing that

PR #2736, opened 2024-09-16
fix(syncs table): sync_config_id must be updated when new deploy

TBonnin:
I introduced a bug last week when adding the `sync_config_id` column to the `syncs` table. The value is set when a sync is created and never changes. However each time a sync definition is deployed (or prebuild templates are upgraded) a new version of the `sync_config` is added but the `sync.sync_config_id` is currently not updated and keep pointing to the same `sync_config` that is now inactive.

Impact: This new `sync_config_id` is only used by the patch frequency endpoint and demo auto idle feature for now but the bug should be fixed asap

This commit is ensuring that sync_config_id in syncs table is updated on deploy/upgrade as well as fixing the drift of `sync_config_id` values in the db

### DIFF
--- a/packages/database/lib/migrations/20240916181552_fix_sync_sync_config_id.cjs
+++ b/packages/database/lib/migrations/20240916181552_fix_sync_sync_config_id.cjs
@@ -1,0 +1,25 @@
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = function (knex) {
+    // when the sync_config_id column was introduced,
+    // it was populated with the correct values
+    // however the sync_config_id must be updated every time a new version of a sync config is created
+    // this migration will fix the sync_config_id column for all existing syncs
+    return knex.raw(`
+            UPDATE _nango_syncs AS syncs
+            SET sync_config_id = sync_configs.id
+            FROM _nango_sync_configs AS sync_configs
+            JOIN _nango_connections AS connections ON sync_configs.nango_config_id = connections.config_id
+            WHERE syncs.name = sync_configs.sync_name
+              AND syncs.nango_connection_id = connections.id
+              AND sync_configs.environment_id = connections.environment_id
+              AND sync_configs.type = 'sync'
+              AND sync_configs.active = true;
+        `);
+};
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.down = function () {};
--- a/packages/shared/lib/services/sync/config/deploy.service.ts
+++ b/packages/shared/lib/services/sync/config/deploy.service.ts
@@ -6,7 +6,7 @@ import { getSyncAndActionConfigByParams, increment, getSyncAndActionConfigsBySyn
 import connectionService from '../../connection.service.js';
 import { LogActionEnum } from '../../../models/Telemetry.js';
 import type { HTTP_VERB, ServiceResponse } from '../../../models/Generic.js';
-import type { SyncModelSchema, SyncConfig, SyncDeploymentResult, SyncConfigResult, SyncEndpoint, SyncType } from '../../../models/Sync.js';
+import type { SyncModelSchema, SyncConfig, SyncDeploymentResult, SyncConfigResult, SyncEndpoint, SyncType, Sync } from '../../../models/Sync.js';
 import type { DBEnvironment, DBTeam, IncomingFlowConfig, IncomingPreBuiltFlowConfig, NangoModel, PostConnectionScriptByProvider } from '@nangohq/types';
 import { postConnectionScriptService } from '../post-connection.service.js';
 import { NangoError } from '../../../utils/error.js';
@@ -22,6 +22,7 @@ import type { Config } from '../../../models/Provider.js';
 import type { NangoSyncConfig } from '../../../models/NangoConfig.js';
 
 const TABLE = dbNamespace + 'sync_configs';
+const SYNC_TABLE = dbNamespace + 'syncs';
 const ENDPOINT_TABLE = dbNamespace + 'sync_endpoints';
 
 const nameOfType = 'sync/action';
@@ -159,8 +160,8 @@ export async function deploy({
             await postConnectionScriptService.update({ environment, account, postConnectionScriptsByProvider });
         }
 
-        if (idsToMarkAsInactive.length > 0) {
-            await db.knex.from<SyncConfig>(TABLE).update({ active: false }).whereIn('id', idsToMarkAsInactive);
+        for (const id of idsToMarkAsInactive) {
+            await markSyncConfigAsInactive(id);
         }
 
         await logCtx.info(`Successfully deployed ${flows.length} script${flows.length > 1 ? 's' : ''}`, {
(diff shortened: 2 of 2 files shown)


packages/database/lib/migrations/20240916181552_fix_sync_sync_config_id.cjs
```diff
@@ -0,0 +1,25 @@
...
+    return knex.raw(`
+            UPDATE _nango_syncs AS syncs
+            SET sync_config_id = sync_configs.id
+            FROM _nango_sync_configs AS sync_configs
+            JOIN _nango_connections AS connections ON sync_configs.nango_config_id = connections.config_id
+            WHERE syncs.name = sync_configs.sync_name
+              AND syncs.nango_connection_id = connections.id
+              AND sync_configs.environment_id = connections.environment_id
+              AND sync_configs.type = 'sync'
+              AND sync_configs.active = true;
```
TBonnin, line 18: fixing the drift of `sync_config_id` values for all the sync definitions that have been deployed/upgraded since last week

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -268,33 +269,41 @@ export async function upgradePreBuilt({
     delete flowData.id;
 
     try {
-        const [syncId] = await db.knex.from<SyncConfig>(TABLE).insert(flowData).returning('id');
```
TBonnin, line 271: change below is not entirely related but I fixed the naming: the query returns SyncConfigs not Sync. I also added a the `if(newSyncConfigId)` to avoid casting

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -363,7 +372,7 @@ export async function deployPreBuilt({
 
     const logCtx = await logContextGetter.create({ operation: { type: 'deploy', action: 'prebuilt' } }, { account, environment });
 
-    const idsToMarkAsInvactive = [];
+    const idsToMarkAsInactive = [];
```
TBonnin, line 376: unrelated to the change but fixing typo

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -425,19 +434,19 @@ export async function deployPreBuilt({
             bumpedVersion = increment(previousSyncAndActionConfig.version as string | number).toString();
 
             if (runs) {
-                const syncsConfig = await getSyncsByProviderConfigAndSyncName(environment.id, provider_config_key, sync_name);
-                for (const syncConfig of syncsConfig) {
-                    const interval = syncConfig.frequency || runs;
+                const syncs = await getSyncsByProviderConfigAndSyncName(environment.id, provider_config_key, sync_name);
```
TBonnin, line 438: renaming variable since `getSyncsByProviderConfigAndSyncName` returns syncs not syncConfigs

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -701,20 +710,20 @@ async function compileDeployInfo({
         }
 
         if (runs) {
-            const syncsConfig = await getSyncsByProviderConfigAndSyncName(environment_id, providerConfigKey, syncName);
+            const syncs = await getSyncsByProviderConfigAndSyncName(environment_id, providerConfigKey, syncName);
```
TBonnin, line 714: same thing: renaming the variable so name matches type

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -835,6 +844,31 @@ async function compileDeployInfo({
...
+                JOIN nango._nango_sync_configs as active_config
+                ON old_config.sync_name = active_config.sync_name
+                AND old_config.nango_config_id = active_config.nango_config_id
+                AND old_config.environment_id = active_config.environment_id
+                WHERE old_config.id = ?
+                AND active_config.active = true
+            )
+            WHERE sync_config_id = ?`,
+            [oldSyncConfigId, oldSyncConfigId]
+        );
```
TBonnin, line 869: not a huge fan of a complex query like this but to be able to match a old sync config with the new added one it would require a heavy refactor of `deploy` and `deployPreBuilt` functions, which I didn't want to do right now

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -551,17 +560,17 @@ export async function deployPreBuilt({
     const isPublic = configs.every((config) => config.is_public);
 
     try {
-        const syncIds = await db.knex.from<SyncConfig>(TABLE).insert(insertData).returning('id');
+        const syncConfigs = await db.knex.from<SyncConfig>(TABLE).insert(insertData).returning('id');
```
bodinsamuel:
```suggestion
        const syncConfigs = await db.knex.from<SyncConfig>(TABLE).insert(insertData).returning('*');
```

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -268,33 +269,41 @@ export async function upgradePreBuilt({
     delete flowData.id;
 
     try {
-        const [syncId] = await db.knex.from<SyncConfig>(TABLE).insert(flowData).returning('id');
+        const [newSyncConfig] = await db.knex.from<SyncConfig>(TABLE).insert(flowData).returning('id');
```
bodinsamuel:
```suggestion
        const [newSyncConfig] = await db.knex.from<SyncConfig>(TABLE).insert(flowData).returning('*');
```

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -268,33 +269,41 @@ export async function upgradePreBuilt({
...
-                }
-                endpoints.push(res);
-            });
-        }
 
-        if (endpoints.length > 0) {
-            await db.knex.from<SyncEndpoint>(ENDPOINT_TABLE).insert(endpoints);
-        }
+        const newSyncConfigId = newSyncConfig?.id;
+        if (newSyncConfigId) {
```
bodinsamuel: with this if we go straight to `logCtx.success()` which seems incorrect
I would do `if (!newSyncConfig.id) throw new Error()`

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -835,6 +844,31 @@ async function compileDeployInfo({
...
+            UPDATE nango._nango_syncs
+            SET sync_config_id = (
+                SELECT active_config.id
+                FROM nango._nango_sync_configs as old_config
+                JOIN nango._nango_sync_configs as active_config
+                ON old_config.sync_name = active_config.sync_name
+                AND old_config.nango_config_id = active_config.nango_config_id
+                AND old_config.environment_id = active_config.environment_id
+                WHERE old_config.id = ?
+                AND active_config.active = true
```
bodinsamuel:
```suggestion
                SELECT active_config.id
                FROM nango._nango_sync_configs as old_config
                JOIN nango._nango_sync_configs as active_config
                    ON old_config.sync_name = active_config.sync_name
                    AND old_config.nango_config_id = active_config.nango_config_id
                    AND old_config.environment_id = active_config.environment_id
                WHERE old_config.id = ?
                    AND active_config.active = true
```
Nitpicking for readability

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -835,6 +844,31 @@ async function compileDeployInfo({
     };
 }
 
+async function markSyncConfigAsInactive(oldSyncConfigId: number): Promise<void> {
```
bodinsamuel: I would just rename for explicitness, this method switch active state from one sync_config id to another

bodinsamuel (approved):
Works great 👍🏻

packages/shared/lib/services/sync/config/deploy.service.ts
TBonnin: that was the behavior before but yeah I agree that doesn't seem right. will fix

bodinsamuel (approved):

PR #2737, opened 2024-09-16
fix: slow records deletion query

TBonnin:
Deleting records by syncId is unnecessarily slow because the sql query doesn't use the partition key (connectionId/model). It becomes very visible when deleting 100_000s records.
This commit add connection and model parameter to the `records.deleteRecordsBySyncId` function. 

Note: all functions in `records` should take `connectionId/model` as parameter so query can be efficient. Not sure how we could enforce this other than via PR reviews but it is something to keep in mind

Note2: fix for sync_config_id must be merged first: https://github.com/NangoHQ/nango/pull/2736

https://linear.app/nango/issue/NAN-1730/delete-records-is-slow

EXPLAIN before:
 ```
Delete on records  (cost=2026.15..4725.10 rows=0 width=0) 
...
```

EXPLAIN after 
```
Delete on records  (cost=0.28..16.34 rows=0 width=0)
...
```

### DIFF
--- a/packages/shared/lib/clients/orchestrator.ts
+++ b/packages/shared/lib/clients/orchestrator.ts
@@ -27,12 +27,12 @@ import { SyncCommand } from '../models/index.js';
 import tracer from 'dd-trace';
 import { clearLastSyncDate } from '../services/sync/sync.service.js';
 import { isSyncJobRunning } from '../services/sync/job.service.js';
-import { getSyncConfigRaw } from '../services/sync/config/config.service.js';
+import { getSyncConfigRaw, getSyncConfigBySyncId } from '../services/sync/config/config.service.js';
 import environmentService from '../services/environment.service.js';
 import type { DBEnvironment, DBTeam } from '@nangohq/types';
 
 export interface RecordsServiceInterface {
-    deleteRecordsBySyncId({ syncId }: { syncId: string }): Promise<{ totalDeletedRecords: number }>;
+    deleteRecordsBySyncId({ connectionId, model, syncId }: { connectionId: number; model: string; syncId: string }): Promise<{ totalDeletedRecords: number }>;
 }
 
 export interface OrchestratorClientInterface {
@@ -540,6 +540,7 @@ export class Orchestrator {
     }
 
     async runSyncCommand({
+        connectionId,
         syncId,
         command,
         environmentId,
@@ -547,6 +548,7 @@ export class Orchestrator {
         recordsService,
         initiator
     }: {
+        connectionId: number;
         syncId: string;
         command: SyncCommand;
         environmentId: number;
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Review by khaliqgant — APPROVED** [review]

File: packages/shared/lib/clients/orchestrator.ts [hunk 1]
```diff
@@ -584,8 +586,11 @@ export class Orchestrator {
                     await cancelling(syncId);
 
                     await clearLastSyncDate(syncId);
-                    const del = await recordsService.deleteRecordsBySyncId({ syncId });
-                    await logCtx.info(`Records for the sync were deleted successfully`, del);
+                    const syncConfig = await getSyncConfigBySyncId(syncId);
+                    for (const model of syncConfig?.models || []) {
+                        const del = await recordsService.deleteRecordsBySyncId({ syncId, connectionId, model });
+                        await logCtx.info(`Records for model ${model} were deleted successfully`, del);
+                    }
```
bodinsamuel, line 593: maybe this change could have directly been encapsulated in the deleteRecordsBySyncId so it's transparent to the user and no duplicate code

packages/shared/lib/services/sync/sync.service.ts
```diff
@@ -639,11 +639,14 @@ export async function findDemoSyncs(): Promise<PausableSyncs[]> {
...
 }
 
-export async function findRecentlyDeletedSync(): Promise<{ id: string; environmentId: number }[]> {
+export async function findRecentlyDeletedSync(): Promise<{ id: string; environmentId: number; connectionId: number; models: string[] }[]> {
     const q = db.knex
         .from('_nango_syncs')
-        .select<{ id: string; environmentId: number }[]>('_nango_syncs.id', '_nango_connections.environment_id')
+        .select<
+            { id: string; environmentId: number; connectionId: number; models: string[] }[]
+        >('_nango_syncs.id', '_nango_connections.environment_id', '_nango_connections.id', '_nango_sync_configs.models')
```
bodinsamuel: _nango_connections.connection_id
```suggestion
        >('_nango_syncs.id', '_nango_connections.environment_id', '_nango_connections.connection_id', '_nango_sync_configs.models')
```

bodinsamuel (approved):
Looks okay (not tested) 👌🏻

packages/shared/lib/services/sync/sync.service.ts
TBonnin: it is actually `connection.id` (the internal number id, not the user-defined one).

packages/shared/lib/clients/orchestrator.ts
TBonnin: I thought about it but I prefer to have all functions in `records` to follow the same pattern, taking a connection and a model. I even thought about create a explicit `Partition` type.

packages/shared/lib/services/sync/sync.service.ts
bodinsamuel: oh okay my bad, but they are conflicting no? both are returning `id` in the final json

packages/shared/lib/clients/orchestrator.ts
bodinsamuel: understandable in the model, could just be a helper on the side, but not pushing anything it's still manageable like this 👌🏻

packages/shared/lib/services/sync/sync.service.ts
TBonnin: the column name of the result have indeed the same `id` name but knex maps them to the select type `{ id: string; environmentId: number; connectionId: number; models: string[] }` based on column order so it works.
But you are right let's be explicit, I'll add aliases

PR #2740, opened 2024-09-17
fix(log): handle large payload

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1740/handle-large-log-payload-again

- Handle large log payload
I had already fixed that but because we are now logging object + message in the sdk, the check has to be done twice. 
However, I also realized we have the same issue in the `{ input }` object when triggering an action. (it can also happen inside our own codebase but less likely)

### Failing validation

> It should return a number, but returns a massive string

```yaml
      returnTooLarge:
        description: trigger an returnTooLarge
        output: ReturnTooLarge
        endpoint: POST /unauth/act/returnTooLarge
      
    ReturnTooLarge:
      id: number
```
```ts

export default async function runAction(nango: NangoSync,): Promise<{ id: string }> {
  return {
    id: "a".repeat(10000000),   };
}
```


khaliqgant (approved):

packages/utils/lib/json.ts
```diff
@@ -0,0 +1,60 @@
...
+    return JSON.parse(truncateJsonPkg(JSON.stringify(value), maxSize).jsonString);
+}
+
+/**
+ * Truncate a JSON as string
+ * Will entirely remove properties that are too big
+ */
+export function truncateJsonString(value: string, maxSize: number = MAX_LOG_PAYLOAD): string {
+    return truncateJsonPkg(value, maxSize).jsonString;
+}
```
TBonnin, line 60: should this file be called something like `stringUtils.ts` instead of `json`?

TBonnin (approved):

packages/utils/lib/json.ts
bodinsamuel: I asked myself the same question: it depends on how you interpret the name of the files: whether the input is important or the output.
This one takes a value as string but others takes Record, either they all related to json (no matter the encoding) or I move truncateJsonString in string.ts
Wdyt?

TBonnin: I don't feel strongly either side, `json` is not a bad name it is just a bit generic. Another option could be `logFormatUtils.ts` or something similar because those functions are mostly used in the context of logging. Feel free to keep it as it is though

bodinsamuel: I think we have many spots where json stringification could be an issue (that's why I moved it there actually); I'll leave it as is

PR #2744, opened 2024-09-18
fix(api): GET /integrations + GET /integrations/:uniqueKey

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1746/get-integrations

- Add `GET /integrations`
- Add `GET /integrations/:uniqueKey`
I changed the param name to `uniqueKey` it's what we return in the payload, happy to discuss.
I didn't add `?include_creds` like in the previous endpoint, when I announce the deprecation, it will be noted as a difference, and I want people to reach out if necessary. This is leaking stuff, and it would be great if we could get rid of it. I actively reached out to the only person I could identify in the list, but no customer is currently using this feature.

- Deprecate `GET /config` and `GET /config/:providerConfigKey`
Note that there are more endpoints under than prefix.

> [!NOTE]
> There is a little bit of code duplication but that's expected so we don't break legacy

---

Open questions: I haven't added/modified the node-client, I have 2 solutions: 
- rename `getIntegration` to `getIntegrationLegacy` but that would force a migration as soon as they upgrade (or use new name)
- add a new method but I don't know what name should I use.


docs-v2/reference/api/integration/getLegacy.mdx
```diff
@@ -0,0 +1,30 @@
+---
+title: 'Get an integration (Legacy)'
```
bodinsamuel: Note Mintlify do not support `deprecated` property. I have sent them an email to see if they can change that

packages/server/lib/controllers/config/providerConfigKey/deleteIntegration.ts
```diff
@@ -1,6 +1,6 @@
 import { asyncWrapper } from '../../../utils/asyncWrapper.js';
 import { requireEmptyQuery, zodErrorToHTTP } from '@nangohq/utils';
-import type { DeleteIntegrationPublic } from '@nangohq/types';
+import type { DeletePublicIntegration } from '@nangohq/types';
```
bodinsamuel, line 3: Just a nomenclature consistency

docs-v2/reference/api/integration/getLegacy.mdx
TBonnin: can we have DEPRECATED in the title, instead of Legacy then.

packages/server/lib/controllers/integrations/uniqueKey/getIntegration.ts
```diff
@@ -0,0 +1,40 @@
...
+        res.status(400).send({ error: { code: 'invalid_uri_params', errors: zodErrorToHTTP(valParams.error) } });
+        return;
+    }
+
+    const { environment } = res.locals;
+    const params: GetPublicIntegration['Params'] = valParams.data;
+
+    const integration = await configService.getProviderConfig(params.uniqueKey, environment.id);
+    if (!integration) {
+        res.status(404).send({ error: { code: 'not_found', message: 'Integration does not exist' } });
```
TBonnin: can we have the requested unique key in the message? Might be nice to have when reading logs

TBonnin (approved):

packages/server/lib/controllers/integrations/uniqueKey/getIntegration.ts
bodinsamuel: good point

docs-v2/reference/api/integration/getLegacy.mdx
bodinsamuel: I followed what we did for other endpoints but happy to change to something clearer

PR #2745, opened 2024-09-19
fix(logs): sync_type can be empty

khaliqgant:
## Describe your changes
Fixes 
image


bodinsamuel (approved):
This seems like an issue that should be solved. In the DB I see we used to always fill it, we might have broken something (might be me)

khaliqgant:
> This seems like an issue that should be solved. In the DB I see we used to always fill it, we might have broken something (might be me)

See about 400+ syncs that have `NULL` for `sync_type` that stops at August 12th so looks like it was fixed then maybe?

PR #2746, opened 2024-09-19
feat: make connectionId optional when creating a new connection

TBonnin:
## Describe your changes

Making connectionId optional when creating a connection. If connectionId is not provided a uuid is generated.
I haven't updated the documentation which is still relevant. I will do it once we decide not passing the connectionId is the default way. With this PR we are making it possible without promoting it yet.

https://linear.app/nango/issue/NAN-1747/make-connectionid-optional-when-creating-a-new-connection

### DIFF
--- a/packages/frontend/lib/index.ts
+++ b/packages/frontend/lib/index.ts
@@ -27,9 +27,9 @@ export interface AuthResult {
     isPending?: boolean;
 }
 
-interface AuthOptions {
+type AuthOptions = {
     detectClosedAuthWindow?: boolean; // If true, `nango.auth()` would fail if the login window is closed before the authorization flow is completed
-}
+} & (ConnectionConfig | OAuth2ClientCredentials | OAuthCredentialsOverride | BasicApiCredentials | ApiKeyCredentials | AppStoreCredentials);
 
 export default class Nango {
     private hostBaseUrl: string;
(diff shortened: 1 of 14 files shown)


packages/frontend/lib/index.ts
```diff
@@ -82,11 +82,24 @@ export default class Nango {
...
      * @returns A promise that resolves with the authentication result
      */
-    public async create(providerConfigKey: string, connectionId: string, connectionConfig?: ConnectionConfig): Promise<AuthResult> {
+    public async create(providerConfigKey: string, connectionConfig?: ConnectionConfig): Promise<AuthResult>;
+    public async create(providerConfigKey: string, connectionId: string, connectionConfig?: ConnectionConfig): Promise<AuthResult>;
+    public async create(
+        providerConfigKey: string,
+        connectionIdOrConnectionConfig?: string | ConnectionConfig,
+        moreConnectionConfig?: ConnectionConfig
+    ): Promise<AuthResult> {
```
TBonnin, line 95: method overload to allow passing `connectionConfig` without having to specify an `undefined` connectionId

packages/webapp/src/pages/Connection/Create.tsx
```diff
@@ -205,14 +205,19 @@ export default function IntegrationCreate() {
...
             hmac: hmacDigest || '',
             credentials
-        })
+        };
+        const getConnection =
+            authMode === 'NONE'
+                ? nango.create(target.integration_unique_key.value, target.connection_id.value, connectionConfig)
+                : nango.auth(target.integration_unique_key.value, target.connection_id.value, connectionConfig);
+
+        getConnection
```
TBonnin, line 220: the compiler doesn't like the dynamic index method calling (I don't know how this is called) with method overloadding. No logic has changed here

bodinsamuel: honestly more readable like 👌🏻

packages/shared/lib/services/hmac.service.ts
```diff
@@ -21,8 +21,9 @@ class HmacService {
         return key;
     }
 
-    async verify(expectedDigest: string, id: number, ...values: string[]): Promise<boolean> {
-        const actualDigest = await this.digest(id, ...values);
+    async verify(expectedDigest: string, id: number, ...values: (string | undefined)[]): Promise<boolean> {
+        const definedValues: string[] = values.flatMap((v) => (v === undefined ? [] : [v]));
```
bodinsamuel, line 25: filter(Boolean) would not work here?

bodinsamuel (approved):
It works great, tested locally by modifying the frontend 👌🏻

khaliqgant (approved):

packages/shared/lib/services/hmac.service.ts
TBonnin: you are right, that should work. let me check

TBonnin: it does work but it will require to cast to `string[]` because filter doesn't change the type of values. I will leave it like it is then

PR #2747, opened 2024-09-19
feat(sdk): add new endpoints

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1753/document-new-endpoints-integrations-and-providers

- Add new endpoints to the SDK
There is only one method that was different enough, queries and output that even with signature overload it wasn't possible to make it non-breaking. Honestly, I didn't find anything compelling to deprecate without either being ugly or breaking customers' scripts, so very open to suggestions. 

- Document new SDK method
- Add support for `include=webhook` in `GET /integrations/:key`
The SDK needs it, and some customers use it. It's a secret but not credentials, so I didn't put a too vague name. If we need credentials at some point, we can differentiate.

### DIFF
--- a/docs-v2/reference/sdks/node.mdx
+++ b/docs-v2/reference/sdks/node.mdx
@@ -125,7 +125,7 @@ await nango.getProvider(<NAME>)
 Returns a list of integrations.
 
 ```js
-await nango.listIntegrationsV2()
+await nango.listIntegrations()
 ```
 
 **Example Response**
@@ -133,7 +133,7 @@ await nango.listIntegrationsV2()
 <Expandable>
 ```json
 {
-    "data": [
+    "configs": [
         {
             "unique_key": "slack-nango-community",
             "provider": "slack",
--- a/docs-v2/reference/sdks/node.mdx
+++ b/docs-v2/reference/sdks/node.mdx
@@ -153,39 +153,6 @@ await nango.listIntegrationsV2()
 ```
 </Expandable>
 
-### List all integrations (Deprecated)
-
-Returns a list of integrations.
-
-```js
-await nango.listIntegrations()
-```
-
-**Example Response**
-
-<Expandable>
-```json
-{
-    "configs": [
-        {
-            "unique_key": "slack-nango-community",
-            "provider": "slack",
-            "logo": "http://localhost:3003/images/template-logos/slack.svg",
-            "created_at": "2023-10-16T08:45:26.241Z",
-            "updated_at": "2023-10-16T08:45:26.241Z",
-        },
-        {
-            "unique_key": "github-prod",
-            "provider": "github",
-            "logo": "http://localhost:3003/images/template-logos/github.svg",
-            "created_at": "2023-10-16T08:45:26.241Z",
-            "updated_at": "2023-10-16T08:45:26.241Z",
-        },
-    ]
-}
-```
-</Expandable>
-
 ### Get an integration
 
 Returns a specific integration.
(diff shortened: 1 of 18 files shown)


docs-v2/reference/sdks/node.mdx
```diff
@@ -94,6 +157,42 @@ await nango.listIntegrations()
 
 Returns a specific integration.
 
+```js
+await nango.getIntegrationV2(<UNIQUE_KEY>, { include?: ['webhook'] });
```
TBonnin: what about having the new method take an single options object instead, then signature is different and we can use method overloading.
```
type getIntegrationOptions = {uniqueKey: string, query?: ...}

public async getIntegration({uniqueKey, query} : getIntegrationOptions);
public async getIntegration(name: string);
public async getIntegration(nameOrOptions: string | getIntegrationOptions) {
...
```
Same for `getProvider({name}: { name: string})` to keep it consistent

packages/node-client/lib/utils.ts
```diff
@@ -40,3 +40,19 @@ export function getUserAgent(userAgent?: string): string {
...
+        return;
+    }
+
+    Object.entries(queries).forEach(([name, value]) => {
+        if (Array.isArray(value)) {
+            for (const el of value) {
+                url.searchParams.set(name, el);
+            }
+        } else {
+            url.searchParams.set(name, value || '');
```
TBonnin: should the query param be set if value is undefined/null?

packages/server/lib/controllers/integrations/uniqueKey/getIntegration.ts
```diff
@@ -12,10 +12,25 @@ export const validationParams = z
...
+    .object({
+        include: z
+            .union([
+                z
+                    .string()
+                    .transform((v) => [v])
+                    .pipe(z.array(valInclude)),
+                z.array(valInclude)
+            ])
+            .optional()
```
TBonnin, line 21: is something like this easier to read?
```
const validationQuery = z.object({
  include: z
    .union([valInclude, z.array(valInclude)])
    .optional()
    .transform((val) => (Array.isArray(val) ? val : val ? [val] : []))
});
```

docs-v2/reference/sdks/node.mdx
bodinsamuel: Could work indeed, it's a paradigm shift though. Usually in Nodejs SDK the first param to get() is an id, string or number not an object.

We can have naming consistency:
```ts
- getProvider({ id })
- getIntegration({ id })
- getConnection({ id })
```
or still use the name
```ts
- getProvider({ name })
- getIntegration({ unique_key })
- getConnection({ connection_id })
```

wdyt?

packages/node-client/lib/utils.ts
bodinsamuel: yeah good point (I'll still allow false which is a legitimate value)

packages/server/lib/controllers/integrations/uniqueKey/getIntegration.ts
bodinsamuel: 👌🏻

packages/node-client/bin/get-integration.js
```diff
@@ -4,7 +4,7 @@ const args = process.argv.slice(2);
 const nango = new Nango({ host: 'http://localhost:3003', secretKey: args[0] });
 
 nango
-    .getIntegration(args[1], args[2])
+    .getIntegrationV2(args[1])
```
khaliqgant: Would rather this be a separate file until we actually retire the original method

khaliqgant (approved):
non blocking comment inline

docs-v2/reference/sdks/node.mdx
TBonnin: I prefer option 2. It is a bit more self-explanatory what the param is.

bodinsamuel:
@TBonnin Updated with signature change, thanks for the suggestion 💪🏻

TBonnin (approved):

PR #2749, opened 2024-09-20
feat(connect): integrations list

bodinsamuel:
## Describe your changes

Contributes to https://linear.app/nango/issue/NAN-1703/create-ui

- Create the integrations list page [[figma](https://www.figma.com/design/wBPwGMlopg8tcMFRekYLgE/Auth-UI?node-id=4107-456&node-type=canvas&m=dev)]

- Setup react correctly: error boundary, router, fetcher, 

Screenshot 2024-09-20 at 13 54 26

## 🧪 Tests
```
npm run dw
npm run dwa

# temp to be able to use API
touch packages/connect-ui/.env 
VITE_LOCAL_SECRET_KEY=__SECRET_KEY__

npm run connect-ui:dev:watch
```


packages/connect-ui/package.json
```diff
@@ -9,7 +9,11 @@
         "preview": "vite preview"
     },
     "devDependencies": {
+        "@nangohq/types": "0.42.10",
```
khaliqgant: Should add this to the publish.sh `bump_other_pkg` https://github.com/NangoHQ/nango/blob/master/scripts/publish.sh#L83

khaliqgant (approved):
Feedback to add to the publish.sh script

packages/connect-ui/package.json
bodinsamuel: oh yeah thanks 🙏🏻

packages/connect-ui/src/lib/api.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    let json: T | undefined;
+    if (res.status !== 204) {
+        json = (await res.json()) as T;
+    }
+
+    return { res, json: json || {} };
+}
+
+export function isError<TType extends Endpoint<{ Path: any; Success: any; Method: any }>['Reply']>(json: TType): json is Exclude<TType, { data: any }> {
```
TBonnin: you used `T` above for the generic type name. Let's be consistent

packages/connect-ui/src/lib/api.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+}
+
+/******************
+ * ENDPOINTS
+ ********************/
+
+export async function getIntegrations() {
+    const { json, res } = await fetchApi<GetPublicListIntegrations>('/integrations', {});
+    if (res.status !== 200 || isError(json)) {
```
TBonnin: Can this logic be in `fetchApi` and make it returns a result. either `Ok({resp, json})` or `Err(new APIError({res, json})` so we don't need to check in each endpoint.
```
export async function getIntegrations() {
   const res = await fetchApi<GetPublicListIntegrations>('/integrations', {});
   if(res.isErr()) {
      throw res.error
   }
   return res.value
}
``` 
Thinking about it we can even throw in `fetchApi` and only return the json when successful 
```
export async function getIntegrations() {
   return fetchApi<GetPublicListIntegrations>('/integrations', {});
}
```

packages/connect-ui/src/lib/query.ts
```diff
@@ -0,0 +1,17 @@
...
+    defaultOptions: {
+        queries: {
+            refetchInterval: 0,
+            refetchOnWindowFocus: false,
+            refetchOnMount: true,
+            staleTime: 60 * 1000 * 5,
+            retry: 0,
+            retryDelay: (attemptIndex) => {
+                return Math.min(2000 * 2 ** attemptIndex, 30000);
+            }
```
TBonnin, line 14: do you need do define retryDelay if retry = 0?

TBonnin (approved):
small comments but lgtm otherwise

packages/connect-ui/src/lib/query.ts
bodinsamuel: no but I might reenable retry at some point, it was annoying now because the backend is mostly broken :p

packages/connect-ui/src/lib/api.ts
bodinsamuel: yep you are right, got lazy 🤦🏻

packages/connect-ui/src/lib/api.ts
bodinsamuel: indeed, not sure why I complicated things

PR #2750, opened 2024-09-20
docs: connection_id is optional when creating a connection

TBonnin:
## Describe your changes

Documenting that connection_id is now optional when creating a connection.
Keeping it as a draft PR for now until we decide to make it the main way to create a connection. Just making sure we don't forget about it.

### DIFF
--- a/docs-v2/integrate/guides/advanced/secure-frontend-sdk.mdx
+++ b/docs-v2/integrate/guides/advanced/secure-frontend-sdk.mdx
@@ -21,7 +21,8 @@ import * as crypto from 'node:crypto';
 
 // Enforce backend authentication before generating the HMAC digest.
 const hmac = crypto.createHmac('sha256', '<HMAC-KEY>'); // HMAC key set in your environment settings.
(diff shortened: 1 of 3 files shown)


bodinsamuel:
just curious if we still need this PR?

TBonnin:
very good question. What is the strategy regarding the deprecation of `nango.auth()` @bastienbeurier

bastienbeurier:
> very good question. What is the strategy regarding the deprecation of `nango.auth()` @bastienbeurier

We said we will remove support for public key, HMAC & non-generated connection IDs on July 31, 2025 ([announcement](https://nango-community.slack.com/archives/C03QBHSMPUM/p1738339205938909)). 

Currently, if I'm not mistaken: 
1. Using the Connect session token enforces a generated connection ID
2. Using the public key forces to pass in a connection ID (deprecated)

On July 31, 2025, we'll remove support for 2. 1. will become the only way to create a connection.


So I'm in favor of cancelling this PR!

PR #2751, opened 2024-09-20
fix(logs): truncate only non required properties

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1769/logs-truncation-removes-mandatory-fields

- Only truncate `meta` object
When truncating a log's JSON it would randomly remove mandatory properties if it was too big. Now we truncate only selected property, it's not yet bulletproof because people could send a large message and large `meta` object too. We should limit the message size

### DIFF
--- a/packages/shared/lib/sdk/sync.ts
+++ b/packages/shared/lib/sdk/sync.ts
@@ -836,8 +836,13 @@ export class NangoAction {
                     if (data.length > MAX_LOG_PAYLOAD) {
                         log.message += ` ... (truncated, payload was too large)`;
                         // Truncating can remove mandatory field so we only try to truncate meta
-                        const meta = JSON.parse(truncateJsonString(stringifyObject(log.meta), MAX_LOG_PAYLOAD - 1000)) as MessageRowInsert['meta'];
-                        data = stringifyObject({ activityLogId: this.activityLogId, log: { ...log, meta } });
+                        data = stringifyObject({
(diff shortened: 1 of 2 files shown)


packages/shared/lib/sdk/sync.ts
```diff
@@ -835,7 +835,13 @@ export class NangoAction {
...
                     if (data.length > MAX_LOG_PAYLOAD) {
                         log.message += ` ... (truncated, payload was too large)`;
-                        data = truncateJsonString(stringifyObject({ activityLogId: this.activityLogId, log }), MAX_LOG_PAYLOAD);
+                        // Truncating can remove mandatory field so we only try to truncate meta
+                        if (log.meta) {
+                            data = stringifyObject({
+                                activityLogId: this.activityLogId,
+                                log: { ...log, meta: truncateJson(log.meta) as MessageRowInsert['meta'] }
+                            });
+                        }
```
TBonnin, line 844: should we also truncate request and response?

TBonnin (approved):

packages/shared/lib/sdk/sync.ts
bodinsamuel: mmh I didn't think about that indeed. It's annoying because truncating a json object is really slow, I wish we had a way to do path traversal and determine length without having to stringify each node.
Let's see if we need after this PR

PR #2752, opened 2024-09-23
fix: missing FLAG_AUTH_ENABLED in docker-compose.yaml

marcodeltongo:
While trying to run locally as per documentation, it defaulted to Sign In form even if FLAG_AUTH_ENABLED was set in ENV.


bodinsamuel (approved):
Indeed, thanks a lot 🙏🏻

PR #2754, opened 2024-09-23
feat(connect-ui): handle generic auth method

bodinsamuel:
## Describe your changes

Contributes to https://linear.app/nango/issue/NAN-1703/create-ui

- Handle generic auth method
API Key, Basic auth, unauth, and all auth that do not take a connection config (e.g: google drive). If your integration is requiring a connectionConfig it's not yet supported

- Frontend: type errors + fix a typo
- Frontend: add a way to clear SDK's state, it was annoying to do that manually when debugging
- Frontend: slightly change the way the modal is being created
We were creating an empty popup and then changing the URL which made the catching of popup blocker more random

## 🧪 Tests
```ts
npm i
npm run dw
npm run dwa

cd packages/connect-ui
touch .env
> VITE_LOCAL_SECRET_KEY=FILL
> VITE_LOCAL_PUBLIC_KEY=FILL
> VITE_LOCAL_HOSTNAME=http://localhost:3003


npm run connect-ui:dev:watch
```

https://github.com/user-attachments/assets/080d3746-bc1e-457e-a80b-175646fc63a4


packages/frontend/lib/index.ts
```diff
@@ -159,35 +171,38 @@ export default class Nango {
...
 
                 const error = new AuthError(errorDesc, errorType);
-                return reject(error);
+                reject(error);
+                return;
             };
 
             if (this.status === AuthorizationStatus.BUSY) {
-                const error = new AuthError('The authorization window is already opened', 'windowIsOppened');
+                const error = new AuthError('The authorization window is already opened', 'windowIsOpened');
```
bodinsamuel, line 203: fixing this typo might be a breaking change

packages/frontend/lib/index.ts
```diff
@@ -159,35 +171,38 @@ export default class Nango {
...
                 reject(error);
+                return;
```
bodinsamuel, line 205: we were not rejecting but not returning, not sure if it was intentional but if yes it was annoying to get an error but still get a popup. Happy to discuss

packages/server/lib/controllers/config.controller.ts
```diff
@@ -144,11 +144,7 @@ class ConfigController {
             );
 
             res.status(200).send({
-                integrations: integrations.sort((a: Integration, b: Integration) => {
-                    const creationDateA = a.creationDate || new Date(0);
-                    const creationDateB = b.creationDate || new Date(0);
-                    return creationDateB.getTime() - creationDateA.getTime();
-                })
```
bodinsamuel, line 151: Happy to discuss this sort, I'm sure it was intentional but very confusing to have this "random" sort in the UI

packages/connect-ui/src/lib/nango.ts
```diff
@@ -0,0 +1,6 @@
+import Nango from '@nangohq/frontend';
+
+export const nango = new Nango({
+    publicKey: import.meta.env.VITE_LOCAL_PUBLIC_KEY,
```
TBonnin, line 4: tried to run the branch but getting a publicKey is not a uuid v4 error. How should I be passing this env var?

packages/connect-ui/src/views/IntegrationsList.tsx
```diff
@@ -51,20 +53,28 @@ const Integrations: React.FC = () => {
...
 
     async function triggerAuth() {
         setLoading(true);
+
         let provider: GetPublicProvider['Success'] | undefined;
         try {
             provider = await getProvider({ provider: integration.provider });
         } catch (err) {
-            console.log(err); // TODO: handle this
+            console.log(err);
```
TBonnin, line 69: is this still needed?

packages/frontend/lib/index.ts
```diff
@@ -202,25 +217,53 @@ export default class Nango {
...
+                this.win.close();
+            } catch {
+                // do nothing
+            }
+            this.win = null;
+        }
+
+        this.status = AuthorizationStatus.IDLE;
+    }
+
```
TBonnin, line 266: I think that calling `clear` in `nango.auth()` will solve this bug: https://linear.app/nango/issue/NAN-1773/fix-the-authorization-window-is-already-opened

packages/server/lib/controllers/config.controller.ts
TBonnin: I imagine it was to show the integrations by most recently added in the dashboard. The sorting can be done client side I think.

TBonnin (approved):

packages/server/lib/controllers/config.controller.ts
bodinsamuel: yeah, actually that's what I'm changing because it's super annoying to have seemingly random order in the dashboard. 
Maybe @khaliqgant can clarify what were the expectations?

packages/connect-ui/src/lib/nango.ts
bodinsamuel: in the PR desc there are some instructions, you should be able to test just by creating a `.env` in connect-ui, let me know

packages/connect-ui/src/views/IntegrationsList.tsx
bodinsamuel: Since we don't show actual error, I'm usually fine to log errors in the console so we can open the console in prod. wdyt?

packages/server/lib/controllers/config.controller.ts
bodinsamuel: merging, happy to revert if needed

PR #2756, opened 2024-09-23
feat: add keystore package

TBonnin:
## Describe your changes

A new keystore package containing logic to create, store, retrieve and delete private keys.
I decided to make it an independant package instead of adding it to `server` so it can be self-contained and reusable. It means the table migrations are different (which creates new migrations table) but also means the package can be used independently. 
The db client must be provided though so it can use the same as the one used by the service importing `keystore`. 
In this PR I added keystore to `server`. In next PR I will add the endpoint used in the new auth flow to create and check session token

https://linear.app/nango/issue/NAN-1758/implement-keystore-package

- [x] I added observability, otherwise the reason is:

### DIFF
--- a/packages/keystore/lib/db/migrate.ts
+++ b/packages/keystore/lib/db/migrate.ts
@@ -1,0 +1,32 @@
+import path from 'node:path';
+import type knex from 'knex';
+import { fileURLToPath } from 'node:url';
+import { logger } from '../utils/logger.js';
+
+export async function migrate(db: knex.Knex, schema: string): Promise<void> {
+    logger.info('[keystore] migration');
+
+    const runningMigrationOnly = process.argv.some((v) => v === 'migrate:latest');
+    const isJS = !runningMigrationOnly;
+    const migrationsConfig = {
+        extension: isJS ? 'js' : 'ts',
+        directory: 'migrations',
+        tableName: 'migrations_keystore',
+        loadExtensions: [isJS ? '.js' : '.ts'],
+        schemaName: schema
+    };
+
+    const filename = fileURLToPath(import.meta.url);
+    const dirname = path.dirname(path.join(filename, '../../'));
+    const dir = path.join(dirname, 'dist/db/migrations');
+
+    await db.raw(`CREATE SCHEMA IF NOT EXISTS ${schema}`);
+    const [, pendingMigrations] = (await db.migrate.list({ ...migrationsConfig, directory: dir })) as [unknown, string[]];
+    if (pendingMigrations.length === 0) {
+        logger.info('[keystore] nothing to do');
+        return;
+    }
+
+    await db.migrate.latest({ ...migrationsConfig, directory: dir });
+    logger.info('[keystore] migrations completed.');
+}
--- a/packages/keystore/lib/db/migrate.ts
+++ b/packages/keystore/lib/db/migrate.ts
@@ -3,7 +3,7 @@ import type knex from 'knex';
 import { fileURLToPath } from 'node:url';
 import { logger } from '../utils/logger.js';
 
-export async function migrate(db: knex.Knex, schema: string): Promise<void> {
+export async function migrate(db: knex.Knex, migrationSchema: string = 'migrations'): Promise<void> {
     logger.info('[keystore] migration');
 
     const runningMigrationOnly = process.argv.some((v) => v === 'migrate:latest');
@@ -11,16 +11,16 @@ export async function migrate(db: knex.Knex, schema: string): Promise<void> {
     const migrationsConfig = {
         extension: isJS ? 'js' : 'ts',
         directory: 'migrations',
+        schemaName: migrationSchema,
         tableName: 'migrations_keystore',
-        loadExtensions: [isJS ? '.js' : '.ts'],
-        schemaName: schema
+        loadExtensions: [isJS ? '.js' : '.ts']
     };
 
     const filename = fileURLToPath(import.meta.url);
     const dirname = path.dirname(path.join(filename, '../../'));
     const dir = path.join(dirname, 'dist/db/migrations');
 
-    await db.raw(`CREATE SCHEMA IF NOT EXISTS ${schema}`);
+    await db.raw(`CREATE SCHEMA IF NOT EXISTS ${migrationSchema}`);
     const [, pendingMigrations] = (await db.migrate.list({ ...migrationsConfig, directory: dir })) as [unknown, string[]];
     if (pendingMigrations.length === 0) {
         logger.info('[keystore] nothing to do');
--- a/packages/keystore/lib/db/migrations/20240920184658_initial_keystore_model.ts
+++ b/packages/keystore/lib/db/migrations/20240920184658_initial_keystore_model.ts
@@ -1,0 +1,37 @@
+import type { Knex } from 'knex';
+import { PRIVATE_KEYS_TABLE } from '../../models/privatekeys.js';
+
+export const config = {
+    transaction: false
+};
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.transaction(async (trx) => {
+        await trx.raw(`
+            CREATE TYPE entity_types AS ENUM (
+                'session',
+                'connection',
+                'environment'
+            );
+        `);
+        await trx.raw(`
+            CREATE TABLE IF NOT EXISTS ${PRIVATE_KEYS_TABLE} (
+                id SERIAL PRIMARY KEY,
+                display_name VARCHAR(255) NOT NULL,
+                encrypted BYTEA NOT NULL,
+                hash TEXT NOT NULL UNIQUE,
+                created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+                deleted_at TIMESTAMP WITH TIME ZONE,
+                expires_at TIMESTAMP WITH TIME ZONE,
+                last_access_at TIMESTAMP WITH TIME ZONE,
+                entity_type entity_types NOT NULL,
+                entity_id INTEGER NOT NULL
+            );
+        `);
+    });
+    await knex.raw(`CREATE INDEX CONCURRENTLY IF NOT EXISTS "idx_${PRIVATE_KEYS_TABLE}_type_hash" ON ${PRIVATE_KEYS_TABLE} (entity_type, hash);`);
+}
+
+export async function down(knex: Knex): Promise<void> {
+    await knex.raw(`DROP TABLE IF EXISTS ${PRIVATE_KEYS_TABLE}`);
+}
--- a/packages/keystore/lib/db/migrations/20240920184658_initial_keystore_model.ts
+++ b/packages/keystore/lib/db/migrations/20240920184658_initial_keystore_model.ts
@@ -8,7 +8,7 @@ export const config = {
 export async function up(knex: Knex): Promise<void> {
     await knex.transaction(async (trx) => {
         await trx.raw(`
-            CREATE TYPE entity_types AS ENUM (
+            CREATE TYPE private_key_entity_types AS ENUM (
                 'session',
                 'connection',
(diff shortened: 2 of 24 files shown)


packages/keystore/lib/db/migrations/20240920184658_initial_keystore_model.ts
```diff
@@ -0,0 +1,37 @@
...
+    transaction: false
+};
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.transaction(async (trx) => {
+        await trx.raw(`
+            CREATE TYPE entity_types AS ENUM (
+                'session',
+                'connection',
+                'environment'
```
TBonnin, line 14: we might eventually want to use this package to handle environment private keys.

packages/keystore/lib/models/privatekeys.ts
```diff
@@ -0,0 +1,152 @@
...
+        };
+    }
+};
+
+export async function createPrivateKey(
+    db: knex.Knex,
+    { displayName, entityType, entityId, ttlInMs }: Pick<PrivateKey, 'displayName' | 'entityType' | 'entityId'> & { ttlInMs?: number }
+): Promise<Result<string, PrivateKeyError>> {
+    const now = new Date();
+    const prefix = entityType.slice(0, 4);
```
TBonnin: adding a prefix to make it easier to debug

packages/keystore/lib/models/privatekeys.ts
```diff
@@ -0,0 +1,152 @@
...
+    const random = crypto.randomBytes(32).toString('hex');
```
TBonnin, line 82: I picked a size of 32 bytes but I am happy to reconsider

packages/keystore/lib/db/migrations/20240920184658_initial_keystore_model.ts
```diff
@@ -0,0 +1,37 @@
...
+            );
+        `);
+        await trx.raw(`
+            CREATE TABLE IF NOT EXISTS ${PRIVATE_KEYS_TABLE} (
+                id SERIAL PRIMARY KEY,
+                display_name VARCHAR(255) NOT NULL,
```
bodinsamuel, line 20: missing env_id and/or account_id to be able to quickly select / delete all of them. Especially if we have a put environment keys inside

packages/keystore/lib/types.ts
```diff
@@ -0,0 +1,15 @@
...
+    readonly displayName: string;
+    readonly encrypted: Buffer;
+    readonly hash: string;
+    readonly createdAt: Date;
+    readonly deletedAt: Date | null;
+    readonly expiresAt: Date | null;
+    readonly lastAccessAt: Date | null;
+    readonly entityType: EntityType;
+    readonly entityId: number;
+}
```
bodinsamuel: imo it will probably be needed in the sdk and server, might be easier to have that type in `/types`

packages/keystore/lib/models/privatekeys.ts
```diff
@@ -0,0 +1,152 @@
...
+            hash: dbKey.hash,
+            createdAt: dbKey.created_at,
+            deletedAt: dbKey.deleted_at,
+            expiresAt: dbKey.expires_at,
+            lastAccessAt: dbKey.last_access_at,
+            entityType: dbKey.entity_type,
+            entityId: dbKey.entity_id
+        };
+    }
+};
```
bodinsamuel, line 67: it's not really clear that it's just a formatter

packages/keystore/lib/db/migrations/20240920184658_initial_keystore_model.ts
```diff
@@ -0,0 +1,37 @@
...
+import { PRIVATE_KEYS_TABLE } from '../../models/privatekeys.js';
+
+export const config = {
+    transaction: false
+};
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.transaction(async (trx) => {
+        await trx.raw(`
+            CREATE TYPE entity_types AS ENUM (
```
bodinsamuel: entity might be too common since enum are stored globally

packages/keystore/lib/db/migrations/20240920184658_initial_keystore_model.ts
```diff
@@ -0,0 +1,37 @@
...
+                created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+                deleted_at TIMESTAMP WITH TIME ZONE,
+                expires_at TIMESTAMP WITH TIME ZONE,
+                last_access_at TIMESTAMP WITH TIME ZONE,
+                entity_type entity_types NOT NULL,
+                entity_id INTEGER NOT NULL
+            );
+        `);
+    });
+    await knex.raw(`CREATE INDEX CONCURRENTLY IF NOT EXISTS "idx_${PRIVATE_KEYS_TABLE}_type_hash" ON ${PRIVATE_KEYS_TABLE} (entity_type, hash);`);
```
bodinsamuel: I would let the hash define the whole query

```ts 
Authorization: nango_session_kdjfdjfd

privateKeys.get('nango_session_kdjfdjfd')

SELECT * FROM private_keys where hash = HASH(nango_session_kdjfdjfd)
```

packages/keystore/lib/db/migrations/20240920184658_initial_keystore_model.ts
```diff
@@ -0,0 +1,37 @@
...
+                'session',
+                'connection',
+                'environment'
+            );
+        `);
+        await trx.raw(`
+            CREATE TABLE IF NOT EXISTS ${PRIVATE_KEYS_TABLE} (
+                id SERIAL PRIMARY KEY,
+                display_name VARCHAR(255) NOT NULL,
+                encrypted BYTEA NOT NULL,
```
bodinsamuel: We can argue some type of keys (if not all of them) should not be stored at all, just the hash. (expect env I guess)

packages/keystore/lib/models/privatekeys.ts
```diff
@@ -0,0 +1,152 @@
...
+    readonly encrypted: Buffer;
+    readonly hash: string;
+    readonly created_at: Date;
+    readonly deleted_at: Date | null;
+    readonly expires_at: Date | null;
+    readonly last_access_at: Date | null;
+    readonly entity_type: EntityType;
+    readonly entity_id: number;
+}
+type DbInsertPrivateKey = Omit<DbPrivateKey, 'id' | 'created_at' | 'deleted_at' | 'last_access_at'>;
```
bodinsamuel: Could be stored elsewhere for clarity

packages/keystore/lib/models/privatekeys.ts
bodinsamuel: I would try to have well identified non truncated prefix, e.g: 
nango_session
nango_connection
nango_secret_env_
nango_public_env_

For security and clarity reasons

packages/keystore/lib/db/migrations/20240920184658_initial_keystore_model.ts
```diff
@@ -0,0 +1,37 @@
...
+                hash TEXT NOT NULL UNIQUE,
+                created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+                deleted_at TIMESTAMP WITH TIME ZONE,
```
bodinsamuel: Not sure I would bother with soft delete rn, that's a liability if db leaks and needs regular clean up

packages/keystore/lib/models/privatekeys.ts
```diff
@@ -0,0 +1,152 @@
...
+    return Ok(keyValue);
+}
+
+export async function getPrivateKey(
+    db: knex.Knex,
+    { keyValue, entityType }: { keyValue: string; entityType: EntityType }
+): Promise<Result<PrivateKey, PrivateKeyError>> {
+    const now = new Date();
+    const [key] = await db
+        .update({ last_access_at: now })
```
bodinsamuel, line 104: smart

packages/keystore/lib/db/migrate.ts
```diff
@@ -0,0 +1,32 @@
+import path from 'node:path';
```
bodinsamuel, line 1: I understand why you created a dedicated file, however, it's going to create two lock tables for each package that uses this pattern. I don't have a strong opinion because it's definitely handy to split everything, just not sure it's worth it imo

bodinsamuel (commented):
Looks good, a few open questions and suggestions

packages/keystore/lib/db/migrate.ts
TBonnin: It is true the package could assume the tables exist and have a single place that ensure the database is in the correct state. The cost is two extra lock tables, the advantage if we ever use the package in another service  is that we don't have to worry about order of deployment/migration since it is embedded in keystore package.
It also make testing easier, the package can just run the keystore related migrations without having to setup the entire db.

btw we are also creating extra migrations tables for the scheduler.

What if the migrations tables were in their own `migrations` schema?

packages/keystore/lib/db/migrations/20240920184658_initial_keystore_model.ts
TBonnin: good point. I will prefix with `private_key_`

packages/keystore/lib/db/migrations/20240920184658_initial_keystore_model.ts
TBonnin: good point. Even though one can always store some placeholder data it is better to be explicit in the model definition.

packages/keystore/lib/models/privatekeys.ts
TBonnin: it is only used internally and it mostly a convenience type to make it harder to insert the wrong type. I am gonna leave it here

packages/keystore/lib/models/privatekeys.ts
TBonnin: what about `PrivateKeyMapper`?

packages/keystore/lib/types.ts
TBonnin: I'll move it when it is necessary

packages/keystore/lib/db/migrations/20240920184658_initial_keystore_model.ts
TBonnin: also modifed create function to specify if you want to store the encrypted value or not

TBonnin:
@bodinsamuel thank you very much for the review. I think I have addressed all your comments. 
I'll ping you to discuss further about the migration tables

packages/types/lib/keystore/index.ts
```diff
@@ -0,0 +1,16 @@
+export const entityTypes = ['session', 'connection', 'environment'] as const;
```
bodinsamuel: Cant' be in types sorry. You can replicate the same inverted with type-fest

```ts
import type { UnionToTuple } from 'type-fest';

export type EntityType = 'session' | 'connection' | 'environment';

export const entityTypes: UnionToTuple<EntityType> = ['session', 'connection'];
// Will error, enforce all entries to be set once
````

bodinsamuel (approved):

packages/types/lib/keystore/index.ts
bodinsamuel: available in >4.26

packages/keystore/lib/db/migrate.ts
bodinsamuel: Yeah, the main difference is that it's not in the same schema. Let's discuss tomorrow ☺️

packages/types/lib/keystore/index.ts
TBonnin: I actually don't need the entityTypes const. will remove

PR #2758, opened 2024-09-24
fix(integrations): update endpoints so entities are organized via headers

khaliqgant:
## Describe your changes
Update endpoints so they are grouped by entity


bodinsamuel (approved):
We need to work on supporting {Model:id} in URL because `PUT /accounts` is "incorrect"

TBonnin (approved):

khaliqgant:
> We need to work on supporting {Model:id} in URL because `PUT /accounts` is "incorrect"

Yup. Indeed

PR #2762, opened 2024-09-25
feat(providers): add display name, validate doc/svg path

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1786/add-display-name-for-all-providers

- Add display name for each providers
- Add stricter validation for doc path, svg path
- Fix some naming, missing doc, missing svg
- Use them in the dashboard

### DIFF
--- a/docs-v2/integrations/all/unauthenticated.mdx
+++ b/docs-v2/integrations/all/unauthenticated.mdx
@@ -1,0 +1,8 @@
+---
+title: Unauthenticated
+sidebarTitle: Unauthenticated
+---
+
+API configurations: [`unauthenticated`](https://nango.dev/providers.yaml)
+
+This is a testing provider, it is not doing anything and not recommended to use in production.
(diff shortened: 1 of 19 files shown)


docs-v2/integrations/all/unauthenticated.mdx
```diff
@@ -0,0 +1,8 @@
+---
+title: Unauthenticated
+sidebarTitle: Unauthenticated
+---
+
+API configurations: [`unauthenticated`](https://nango.dev/providers.yaml)
+
+This is a testing provider, it is not doing anything and not recommended to use in production.
```
khaliqgant: I don't know if I would say it is not recommended to use in production. It has use cases.

I would rather say,

> This provider doesn't require credentials and no authentication flow is performed when creating connections.

...or something like that.

khaliqgant (approved):
Looks good! Feedback on unauthenticated description

docs-v2/integrations/all/unauthenticated.mdx
bodinsamuel: Yeah makes sense I wasn't sure how we wanted to advertise this one. Changing 👌🏻

PR #2763, opened 2024-09-25
feat(integration-template-tests): [nan-1696] update sdk so that record responses is less instrusive

khaliqgant:
## Describe your changes
Use an axios interceptor to avoid adding logic to the SDK and so that the logic for saving down responses is in one place. This was suggested in https://github.com/NangoHQ/nango/pull/2678#discussion_r1747576571 and discussed further in Slack.

NAN-1696


packages/shared/lib/sdk/sync.ts
```diff
@@ -424,6 +433,14 @@ export class NangoAction {
         if (config.dryRun) {
             this.dryRun = config.dryRun;
             this.saveResponses = config.saveResponses || false;
```
bodinsamuel: Is `saveResponses` still useful?

bodinsamuel (approved):
looks good

packages/shared/lib/sdk/sync.ts
khaliqgant: Nope! Thanks 🙏

Updated with https://github.com/NangoHQ/nango/pull/2763/commits/4cd4273bcd20a91759261ec9f9bbb0a377fb0c93

packages/cli/lib/services/response-saver.service.ts
```diff
@@ -0,0 +1,86 @@
...
+    ensureDirectoryExists(`${directoryName}/mocks`);
+
+    let filePath: string;
+    if (customFilePath) {
+        filePath = path.join(directoryName, customFilePath);
+        ensureDirectoryExists(path.dirname(filePath));
+    } else {
+        const fileName = config.endpoint.replace('/', '');
+        filePath = `${directoryName}/mocks/${config.method?.toLowerCase()}/${fileName}.json`;
+        ensureDirectoryExists(path.dirname(filePath));
```
TBonnin, line 33: you can put this outside of the ifelse

packages/node-client/lib/index.ts
```diff
@@ -107,6 +117,12 @@ export class Nango {
                 'User-Agent': this.userAgent
             }
         });
+
+        if (interceptors) {
+            if (interceptors.response) {
```
TBonnin, line 122: `if(interceptors?.response) {`

packages/shared/lib/sdk/sync.ts
```diff
@@ -613,15 +619,6 @@ export class NangoAction {
...
             this.memoizedConnections.set(credentialsPair, { connection, timestamp: Date.now() });
-            if (this.saveResponses) {
-                const directoryName = `${process.env['NANGO_MOCKS_RESPONSE_DIRECTORY']}${providerConfigKey}`;
-                responseSaver.saveResponse<Pick<Connection, 'metadata' | 'connection_config'>>({
-                    directoryName,
-                    config: { endpoint: 'getConnection', providerConfigKey } as ProxyConfiguration,
-                    data: { metadata: connection.metadata as Metadata, connection_config: connection.connection_config },
-                    customFilePath: 'mocks/nango/getConnection.json'
-                });
-            }
```
TBonnin, line 624: ❤️

TBonnin (commented):
Much cleaner. very good job

PR #2768, opened 2024-09-25
feat(providers): add connection_config, credentials definition

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1792/connection-config-document-in-providersyaml-and-enforce-them

- Add and enforce `connection_config` definitions
This makes sure we have a JSON schema to validate and to display connection config in the UI and backend. 
The values are returned by the endpoint but not yet used in the UI. Some things are incomplete or might be slightly incorrect, it's a daunting task so I have fully checked and vested all of them.

- Add and enforce the `credentials` definitions for API_KEY
Same as `connection_config`, separated to clarify the expectation. 

## Tests
Screenshot 2024-09-25 at 18 10 18


khaliqgant (approved):

scripts/validation/providers/validate.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    for (const [key, value] of Object.entries(obj)) {
+        if (typeof value === 'string') {
+            const matches = value.matchAll(/\$\{connectionConfig\.([^}]+)\}/g);
+            for (const match of matches) {
+                references.push({ path: [...path, key], key: match[1]! });
+            }
+        } else if (typeof value === 'object' && value) {
+            references = references.concat(findConnectionConfigReferences(value, [...path, key]));
```
TBonnin: no need for `concat` I think:
```
references.push(...findConnectionConfigReferences(value, [...path, key]));
```

scripts/validation/providers/validate.ts
```diff
@@ -0,0 +1,129 @@
...
+
+console.log('✅ JSON schema valid');
+
+// Check if files exist
+console.log('Checking values...');
+const docsPath = path.join(__dirname, '../../../docs-v2/integrations/all');
+const svgPath = path.join(__dirname, '../../../packages/webapp/public/images/template-logos');
+
+let error = false;
+for (const [providerKey, provider] of Object.entries(providersJson)) {
```
TBonnin, line 53: not a big deal but we could have the logic in this loop extracted to a `validateProvider(providerKey, provider)` function to make it more readable. That could also allow to process all providers in parrallel :)

packages/shared/providers.yaml
```diff
@@ -1,3 +1,4 @@
+# yaml-language-server: $schema=./../../scripts/validation/providers/schema.json
```
TBonnin, line 1: I haven't checked all changes in providers.yaml. Is it possible to verify that all  patterns for example matches existing examples we have in the database?

TBonnin (approved):

scripts/validation/providers/validate.ts
bodinsamuel: indeed (not going to lie it's chatgpt :whistle:)

scripts/validation/providers/validate.ts
bodinsamuel: Happy to abstract a bit indeed. I won't do it in parallel it's messing up the logs, it's not too slow rn

packages/shared/providers.yaml
bodinsamuel: yep as I said I haven't checked all of them, it's not enforced so not side effects rn

PR #2769, opened 2024-09-25
feat:  add POST /connect/sessions endpoint

TBonnin:
## Describe your changes

Adding the `POST /connect/sessions` endpoint as well as the `end_users` and `connect_sessions` tables.
Happy to discuss the naming 😺 

Spec:
```
POST /connect/sessions
-H "Authorization: Bearer PRIVATE_KEY"
{
  "end_user": {
     "id": "1jkhfz2FEuihojhfsdkjh29842", // Required
     "email": "a@a.com", // Required but for domain
     "display_name": "Mr A." // Optional
  },
   "organization": { // Optional
      "if": "foi", // Optional
      "display_name": "My org.", // Optional
   }
  "allowed_integrations": ['hubspot', 'github'] // Optional
  "integrations_config_defaults": { // Optional
    "hubspot": {
      "connection_config": {
        "extension": "fr",
      }
    },
  },
}

// Response
{
  "token": "nango_connect_session_dfkjdfkdjflksjfoizrjfozijfzoijfz",
  "expires_at": Date
}
```

`GET /connect/session` will come in next PR. Split them so PR are smaller

https://linear.app/nango/issue/NAN-1775/%5Bserver%5D-api-endpoints-to-getpost-sessions


packages/keystore/lib/utils/encryption.ts
```diff
@@ -1,15 +1,32 @@
...
         encryption = new Encryption(encryptionKey);
     }
     return encryption;
 }
+
+export async function hashValue(val: string): Promise<string> {
+    const encryptionKey = getEncryptionKey();
+    if (!encryptionKey) {
+        throw new Error('NANGO_ENCRYPTION_KEY is not set');
+    }
```
bodinsamuel: If both call are mandatory, the throw could live in the getEncryptionKey imo

packages/server/lib/controllers/connect/postSessions.ts
```diff
@@ -0,0 +1,121 @@
...
+                    profileId: z.string().max(255).min(1),
+                    email: z.string().email().optional(),
+                    displayName: z.string().max(255).optional(),
+                    organization: z
+                        .object({
+                            organizationId: z.string().max(255).min(0),
+                            displayName: z.string().max(255).optional()
+                        })
+                        .optional()
+                }),
```
bodinsamuel:
```suggestion
                }).strict(),
```

packages/server/lib/controllers/connect/postSessions.ts
```diff
@@ -0,0 +1,121 @@
...
+            .object({
+                linkedProfile: z.object({
+                    profileId: z.string().max(255).min(1),
+                    email: z.string().email().optional(),
+                    displayName: z.string().max(255).optional(),
+                    organization: z
+                        .object({
+                            organizationId: z.string().max(255).min(0),
+                            displayName: z.string().max(255).optional()
+                        })
```
bodinsamuel:
```suggestion
                        }).strict()
```

packages/server/lib/controllers/connect/postSessions.ts
```diff
@@ -0,0 +1,121 @@
...
+                        .optional()
+                }),
+                allowedIntegrations: z.array(z.string()).optional(),
+                integrationsConfigDefaults: z
+                    .record(
+                        z.object({
+                            connectionConfig: z.record(z.unknown())
+                        })
```
bodinsamuel:
```suggestion
                        }).strict()
```

packages/server/lib/controllers/connect/postSessions.ts
```diff
@@ -0,0 +1,121 @@
...
+        const getLinkedProfile = await linkedProfileService.getLinkedProfile(trx, {
+            profileId: req.body.linkedProfile.profileId,
+            accountId: res.locals.account.id,
+            environmentId: res.locals.environment.id
+        });
+
+        let linkedProfileId: number;
+        if (getLinkedProfile.isErr()) {
+            // create linked profile if it doesn't exist yet
+            if (getLinkedProfile.error.code === 'not_found') {
```
bodinsamuel: opinion: you could reduce the indenting by reversing the condition

packages/server/lib/controllers/connect/postSessions.ts
```diff
@@ -0,0 +1,121 @@
...
+import { validateRequest } from '@nangohq/utils';
+import * as keystore from '@nangohq/keystore';
+import * as linkedProfileService from '../../services/linkedProfile.service.js';
+import * as connectSessionService from '../../services/connectSession.service.js';
+
+const validate = validateRequest<PostConnectSessions>({
+    parseBody: (data) =>
+        z
+            .object({
+                linkedProfile: z.object({
```
bodinsamuel: At this stage I would not bother allowing just sending the `profileId` because it advocate for never sending the full profile again leading to desync

packages/server/lib/controllers/connect/postSessions.ts
```diff
@@ -0,0 +1,121 @@
...
+                    accountId: res.locals.account.id,
+                    environmentId: res.locals.environment.id
+                });
+                if (createLinkedProfile.isErr()) {
+                    res.status(500).send({ error: { code: 'internal_error', message: 'Failed to create linked profile' } });
+                    return;
+                }
+                linkedProfileId = createLinkedProfile.value.id;
+            } else {
+                console.log(getLinkedProfile.error);
```
bodinsamuel: if you want to keep it
```suggestion
                logger.error(getLinkedProfile.error);
```

packages/server/lib/controllers/connect/postSessions.ts
```diff
@@ -0,0 +1,121 @@
...
+                res.status(500).send({ error: { code: 'internal_error', message: 'Failed to get linked profile' } });
+                return;
+            }
+        } else {
+            // TODO: what if linkedProfile exists but email, displayName, or organization is different?
+            linkedProfileId = getLinkedProfile.value.id;
```
bodinsamuel: as we said in the meeting, profileId should probably be `companyId + profileId` 
if the email change then it needs update, if the companyId is different that means it's a different profile (even if it's the same end user)

packages/server/lib/controllers/connect/postSessions.ts
```diff
@@ -0,0 +1,121 @@
+import type { PostConnectSessions } from '@nangohq/types';
+import { z } from 'zod';
+import db from '@nangohq/database';
+import { asyncWrapper } from '../../utils/asyncWrapper.js';
+import { validateRequest } from '@nangohq/utils';
+import * as keystore from '@nangohq/keystore';
+import * as linkedProfileService from '../../services/linkedProfile.service.js';
+import * as connectSessionService from '../../services/connectSession.service.js';
+
+const validate = validateRequest<PostConnectSessions>({
```
bodinsamuel: I know the format is frustrating rn but I wouldn't use this, the error codes are entirely different from the rest of the API

packages/server/lib/services/connectSession.service.ts
```diff
@@ -0,0 +1,102 @@
+import type knex from 'knex';
+import type { ConnectSession } from '@nangohq/types';
+import { Err, Ok } from '@nangohq/utils';
+import type { Result } from '@nangohq/utils';
+
+const CONNECT_SESSIONS_TABLE = 'connect_sessions';
+
+interface DbConnectSession {
```
bodinsamuel: We mostly use `DB` (except DbTask) we need to pick one

packages/server/lib/services/connectSession.service.ts
```diff
@@ -0,0 +1,102 @@
...
+                code: 'creation_failed',
+                message: 'Failed to create connect session',
+                payload: { linkedProfileId, allowedIntegrations, integrationsConfigDefaults }
+            })
+        );
+    }
+    return Ok(ConnectSessionMapper.from(session));
+}
+
+export async function getConnectSession(db: knex.Knex, id: number): Promise<Result<ConnectSession, ConnectSessionError>> {
```
bodinsamuel: should take envId at least

packages/server/lib/services/connectSession.service.ts
```diff
@@ -0,0 +1,102 @@
...
+    const session = await db<DbConnectSession>(CONNECT_SESSIONS_TABLE).where({ id }).first();
+    if (!session) {
+        return Err(new ConnectSessionError({ code: 'not_found', message: `Connect session '${id}' not found` }));
+    }
+    return Ok(ConnectSessionMapper.from(session));
+}
+
+export async function deleteConnectSession(db: knex.Knex, id: number): Promise<Result<void, ConnectSessionError>> {
```
bodinsamuel: same comment

packages/server/lib/services/linkedProfile.service.ts
```diff
@@ -0,0 +1,130 @@
...
+                    accountId,
+                    environmentId
+                }
+            })
+        );
+    }
+    return Ok(LinkedProfileMapper.from(profile));
+}
+
+export async function deleteLinkedProfile(db: knex.Knex, id: number): Promise<Result<void, LinkedProfileError>> {
```
bodinsamuel: same comment

packages/types/lib/connect/api.ts
```diff
@@ -0,0 +1,24 @@
...
+                      organizationId: string;
+                      displayName?: string | undefined;
+                  }
+                | undefined;
+        };
+        allowedIntegrations?: string[] | undefined;
+        integrationsConfigDefaults?: Record<string, { connectionConfig: Record<string, unknown> }> | undefined;
+    };
+    Error: ApiError<'forbidden' | 'invalid_request' | 'internal_error'>;
+    Success: ConnectSessionToken;
```
bodinsamuel: opinion: we never discussed the shape of response, usually I'm advocating to use `data: {} | {}[]` to unify answer and to avoid potential conflict with prop that could be named `error`

packages/types/lib/linkedProfile/index.ts
```diff
@@ -0,0 +1,14 @@
...
+    readonly environmentId: number;
+    readonly email: string;
+    readonly displayName?: string | null;
+    readonly organization?: {
+        readonly organizationId: string;
+        readonly displayName?: string | null;
+    } | null;
+    readonly createdAt: Date;
+    readonly updatedAt: Date | null;
+}
```
bodinsamuel: nb: readonly is nice when you don't manipulate but can be annoying as lib user.
```ts
const lp: Partial<LinkedProfile> = {
    accountId: 1
}

// Not allowed bc it's readonly
lp.email = 'example@dev';
```

Not saying it's an issue in that case but just fyi

packages/keystore/lib/db/migrations/20240925111857_rename_connect_session.ts
```diff
@@ -0,0 +1,13 @@
+import type { Knex } from 'knex';
```
bodinsamuel, line 1: This migration does not run on my local setup

bodinsamuel (commented):
Looks good 🚀 📶 

I can't test because of the migration that don't run (no error just don't run)
Few comments + others stuff (I guess you want to do all of that later but just in case)
- Missing openapi desc, doc
- SDK

packages/server/lib/controllers/connect/postSessions.ts
TBonnin: 🙈

packages/server/lib/controllers/connect/postSessions.ts
TBonnin: changed it but it is a bit ridiculous to have to repeat the same 10 lines on every single endpoint

packages/server/lib/controllers/connect/postSessions.ts
TBonnin: let's not call it profileId if it is not an id then. It might be confusing to users if we consider `profileId + orgId` to be the unique key

EDIT: I had misunderstood your comment. We are on the same page. We leave it to the nango users to decide what's an unique profile and we update profile attributes if they have changed for existing profile ids

packages/types/lib/linkedProfile/index.ts
TBonnin: noted

TBonnin:
@bodinsamuel thank you for the review. I have addressed all your comments I think.

Regarding the missing openapi specs/doc and sdk changes, I wanted to first fully agree on the naming, input and output shapes, etc... before I start documenting things and adding to the sdk. I will do all of that in a subsequent PR

bastienbeurier:
Kicking off the conversation about naming. 

@TBonnin proposed the term "linked profile": 
```
{
  "linkedProfile": {
     "profileId": "1jkhfz2FEuihojhfsdkjh29842", // Required
     "email": "a@a.com", // Required but for domain
     "displayName": "Mr A." // Optional
     "organization": { // Optional
        "organizationId": "foi", // Optional
        "displayName": "My org.", // Optional
     }
  }
}
```

I would propose using the terms "end user" and "end user organization", so the payload would look like: 
```
{
  "end_user_settings": {
     "id": "1jkhfz2FEuihojhfsdkjh29842",
     "email": "a@a.com",
     "display_name": "Mr A.",
     "org_id": "foi",
     "org_display_name": "My org"
  }
}
```
or 
```
{
  "settings": {
     "end_user_id": "1jkhfz2FEuihojhfsdkjh29842",
     "end_user_email": "a@a.com",
     "end_user_display_name": "Mr A.",
     "end_user_org_id": "foi",
     "end_user_org_display_name": "My org"
  }
}
```

@TBonnin @bodinsamuel @rguldener @khaliqgant what do you think?

Other naming detail: since we call the auth UI "Nango Connect", I would use the term "connect session" (and "connect session token" for the actual token) rather than "connection session", because this session only lives while establishing the connection, not for the whole lifespan of the connection.

bodinsamuel:
> linkedProfile | end_user

end_user is confusing because it depends on your POV, if they are also B2B2C then end_user is a different one. That doesn't mean linkedProfile is way more explicit but at least it's POV free, and makes sense to me.

> flat vs deep object

I also prefer deep objects. Flat object is not easier to write, harder to read, annoying to document, and annoying to strictly type (TS type). In our codebase, it will be stored flat so DX over code here. 

> settings 

Too generic.


> Other naming detail: since we call the auth UI "Nango Connect", I would use the term "connect session" (and "connect session token" for the actual token) rather than "connection session", because this session only lives while establishing the connection, not for the whole lifespan of the connection.

Agreed indeed

rguldener:
Agree with the others that `settings` is generic and doesn't fit well here. To me these are `properties` of an object (user or company).

Regarding the entity name: I would go for `user` and `organization`.
I briefly checked and its what [Posthog](https://posthog.com/docs/product-analytics/identify), [Amplitude](https://amplitude.com/docs/apis/analytics/identify), [Intercom](https://developers.intercom.com/installing-intercom/web/attributes-objects) and Mixpanel also do. I think it is likely our users are familiar with these and if it works for them I think it can work for us.

bastienbeurier:
`user` and `organization` works for me. 

No conviction around `flat` vs. `deep` objects. 

@TBonnin the params will be in snake case, right? 

What about: 
```
{
  "user_properties": {
     "id": "",
     "email": "",
     "display_name": ""
  },
  "organization_properties": {
    "id": "",
    "display_name": ""
  }
}
```

or 

```
{
  "user_properties": {
     "id": "",
     "email": "",
     "display_name": "",
     "organization": {
       "id": "",
       "display_name": ""
    }
  },
  
}
```

packages/server/lib/controllers/connect/postSessions.ts
bodinsamuel: yes, very happy to iterate on something similar to validateRequest, just need to follow what we previously set to avoid breaking change. It's mostly a matter of copying the code and changing the response

packages/shared/flows.yaml
```diff
@@ -23,25 +23,25 @@ integrations:
                     Creates an ephemeral transaction in Anrok.
```
bodinsamuel: Merge is incorrect

bodinsamuel (approved):
Still waiting for final decision on naming but looks good 💪🏻

rguldener:
The first one, with user & org on the same level, implies to me that they are independent.

The second, with org in the user, implies to me that the org is linked to the user with a 1:n

On deep vs not: the analytics tools all seem to have pretty flat structures. Not saying we have to copy, but might be worth a quick look at what they do and why.

TBonnin:
the name `user` seems to be the most popular. I'd then prefer to go with @bastienbeurier proposal: `end_user`. 
_Why not simply `user`?_ Because there is also a notion of nango `user` and it might become very confusing very fast, especially for self-hosted customers.

so it will looks like this, without nesting organization to make it more extensible because the data model is simpler now but I wouldn't be surprised if it becomes more complex over time.
```
{
  "end_user": { // required
     "id": "",
     "email": "",
     "display_name": "", 
  }, 
  "organization": { // optional
     "id": "",
     "display_name": ""
  }
}
```

packages/types/lib/connect/api.ts
```diff
@@ -0,0 +1,28 @@
...
+        organization?:
+            | {
+                  id: string;
+                  display_name?: string | undefined;
+              }
+            | undefined;
+        allowed_integrations?: string[] | undefined;
+        integrations_config_defaults?: Record<string, { connection_config: Record<string, unknown> }> | undefined;
+    };
+    Error: ApiError<'forbidden' | 'invalid_body' | 'invalid_query_params' | 'internal_error'>;
```
bodinsamuel, line 21: Old stuff

bodinsamuel (approved):
💯

packages/types/lib/connect/api.ts
TBonnin: yes. It is fixed in the other PR https://github.com/NangoHQ/nango/pull/2776/files#diff-bb1a9436ed436eaa04424f761fb7e7983183af72d751f15bb824395481e4adacL21

PR #2771, opened 2024-09-26
fix(jobs): try to handle knex pool exhaustion

bodinsamuel:
## Describe your changes

- Handle unhandledRejection and uncaughtException
This is something we need to do globally, but it's the opportunity to start. Not entirely sure the error we have is related to uncaught error but we'll see. We don't properly log errors anywhere because we don't have a working sentry and our Datadog error log is shaky. 

- Add proper health checks
Turns out /health is only called on web service, so our current health check (which was not doing much) was useless anyway.  I try to query nothing and timeout quickly

- Remove transaction from Delete Syncs cron
This is the only place where there was an exotic way of using transactions. Honestly, I'm not convinced there was an issue but since the transaction was useless I'm removing it


bodinsamuel:
Tested in staging, didn't notice anything suspicious.

packages/shared/lib/services/sync/job.service.ts
```diff
@@ -100,7 +100,7 @@ export const updateLatestJobSyncStatus = async (sync_id: string, status: SyncSta
  * @desc grab any existing results and add them to the current
  */
 export const updateSyncJobResult = async (id: number, result: SyncResultByModel, model: string): Promise<SyncJob> => {
-    return db.knex.transaction(async (trx) => {
+    return await db.knex.transaction(async (trx) => {
```
TBonnin, line 103: why is it necessary? the calls to `updateSyncJobResult` are themselves awaited

packages/jobs/lib/app.ts
```diff
@@ -16,20 +17,55 @@ try {
     logger.info(`🚀 service ready at http://localhost:${port}`);
     const processor = new Processor(orchestratorUrl);
 
+    // We are using a setTimeout because we don't want overlapping setInterval if the DB is down
+    let healthCheck: NodeJS.Timeout | undefined;
+    const check = async () => {
+        try {
+            await db.knex.raw('SELECT 1').timeout(1000);
```
TBonnin, line 24: that should be in the `/health` endpoint logic

packages/jobs/lib/routes/getHealth.ts
```diff
@@ -14,5 +14,8 @@ export const routeHandler: RouteHandler<Health> = {
     path,
     method,
     validate: (_req, _res, next) => next(), // No extra validation needed
-    handler: (_req, res) => res.status(200).json({ status: 'ok' })
+    handler: (_req, res) => {
+        // Not used in Render
```
TBonnin, line 18: according to the logs this endpoint is called

TBonnin (changes requested):

packages/jobs/lib/app.ts
bodinsamuel: please read PR desc ☺️

packages/jobs/lib/routes/getHealth.ts
bodinsamuel: can you show me the log please? can't find that and their doc is clear on that

TBonnin: oups. I was looking at server (which is not a private service on Render and actually get an healthcheck).
It is ridiculous render doesn't provide healthcheck for private services.
Do we know how Render behaves when the server closes and stopped accepting requests?

bodinsamuel: Yep not good, looking at the doc it's just process liveness I guess 
https://docs.render.com/deploys

packages/shared/lib/services/sync/job.service.ts
bodinsamuel: I don't think there is an issue here, I just wanted to be sure it's awaited somewhere. If on the other end someone is doing `void updateSyncJobResult()` then nobody is awaiting which can be interpreted differently by knex 
https://knexjs.org/guide/transactions.html

TBonnin (approved):
we'll try to check what's the behavior when server is closed but lgtm for now

PR #2781, opened 2024-09-27
fix(jobs): missing timeout value

bodinsamuel:
## Describe your changes

Fix missing default timeout value


TBonnin (approved):

packages/jobs/lib/app.ts
```diff
@@ -31,6 +31,7 @@ try {
     void check();
 
     const close = async () => {
+        console.log('Closing...');
```
TBonnin: logger.info() no?

bodinsamuel: 🤦🏻

PR #2782, opened 2024-09-27
fix(ci): do not ask for upgrade on CI

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1818/do-not-ask-for-upgrade-cli-on-ci

- CLI: do not ask for upgrade on CI

### DIFF
--- a/packages/cli/lib/index.ts
+++ b/packages/cli/lib/index.ts
@@ -17,7 +17,7 @@ import { compileAllFiles } from './services/compile.service.js';
 import verificationService from './services/verification.service.js';
 import { DryRunService } from './services/dryrun.service.js';
 import { v1toV2Migration, directoryMigration } from './services/migration.service.js';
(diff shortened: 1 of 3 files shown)


TBonnin (approved):

khaliqgant:
Why won’t the NANGO_CLI_UPGRADE env variable work here?

packages/cli/lib/index.ts
```diff
@@ -36,7 +36,9 @@ class NangoCommand extends Command {
                 }
             }
 
-            await upgradeAction(debug);
+            if (isCI) {
```
khaliqgant: Don’t we want the upgrade action if it is not CI?

bodinsamuel: yep indeed 🤦🏻

bodinsamuel:
> Why won’t the NANGO_CLI_UPGRADE env variable work here?

It would for sure, it's just unnecessary to ask for customers to know about it

khaliqgant (approved):

PR #2786, opened 2024-09-30
fix(integration-tests): fix overrides

hassan254-prog:
## Describe your changes

- When generating mocks for integration templates, if you have multiple scripts running the same proxy endpoint, it was overriding these mocks, i.e salesforce. Adding the syncName would help distinguish them.


khaliqgant:
Thanks! You'll need a corresponding update to this file as well to include the sync name when fetching the file in this case https://github.com/NangoHQ/integration-templates/blob/main/vitest.setup.ts

khaliqgant (approved):

PR #2787, opened 2024-09-30
feat(connect): require session token to load

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1822/enforce-sessiontoken-usage

- Enforce session token to load UI
We request the sessionToken, open the modal, the modal sends "ready" when react is loaded, and the parent sends back the sessionToken. It's not passed in the UI to avoid leaking the token in logs.

- Add `POST /api/v1/connect/sessions`
We needed a quick way to generate this token for the dashboard, it"s better to assume the UI do not have access to the private key (it's the case but that's not good imo). This endpoint is actually calling the public one by faking the request.

### DIFF
--- a/packages/server/lib/controllers/v1/connect/sessions/postConnectSessions.ts
+++ b/packages/server/lib/controllers/v1/connect/sessions/postConnectSessions.ts
@@ -1,0 +1,31 @@
+import { asyncWrapper } from '../../../../utils/asyncWrapper.js';
+import { requireEmptyBody, requireEmptyQuery, zodErrorToHTTP } from '@nangohq/utils';
+import type { PostConnectSessions, PostInternalConnectSessions } from '@nangohq/types';
+import { postConnectSessions } from '../../../connect/postSessions.js';
+
+export const postInternalConnectSessions = asyncWrapper<PostInternalConnectSessions>((req, res, next) => {
+    const valQuery = requireEmptyQuery(req, { withEnv: true });
+    if (valQuery) {
+        res.status(400).send({ error: { code: 'invalid_query_params', errors: zodErrorToHTTP(valQuery.error) } });
+        return;
+    }
+
+    const valBody = requireEmptyBody(req);
+    if (valBody) {
+        res.status(400).send({ error: { code: 'invalid_body', errors: zodErrorToHTTP(valBody.error) } });
+        return;
+    }
+
+    const { user } = res.locals;
+
+    // @ts-expect-error req.body is never but we want to fake it on purpose
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    req.body = {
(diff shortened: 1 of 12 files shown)


packages/types/lib/connect/api.ts
```diff
@@ -53,3 +53,14 @@ export type DeleteConnectSession = Endpoint<{
...
+
+export type PostInternalConnectSessions = Endpoint<{
+    Method: 'POST';
+    Path: '/api/v1/connect/sessions';
+    Success: {
+        data: {
+            token: string;
+            expires_at: Date;
+        };
+    };
```
TBonnin:
```suggestion
    Success: PostConnectSessions['Success'];
```

packages/server/lib/controllers/v1/connect/sessions/postConnectSessions.ts
```diff
@@ -0,0 +1,31 @@
...
+        res.status(400).send({ error: { code: 'invalid_body', errors: zodErrorToHTTP(valBody.error) } });
+        return;
+    }
+
+    const { user } = res.locals;
+
+    // @ts-expect-error req.body is never but we want to fake it on purpose
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    req.body = {
+        end_user: { id: `nango_${user.id}`, email: user.email, display_name: user.name }
```
TBonnin: should we have a more explicit prefix for the id instead of `nango_`. Something like `from_dashobard_` or similar

packages/server/lib/controllers/v1/connect/sessions/postConnectSessions.ts
```diff
@@ -0,0 +1,31 @@
...
+    } satisfies PostConnectSessions['Body'];
+
+    // @ts-expect-error yes I know
+    req.query = {};
```
TBonnin, line 28: do we need this?

TBonnin (approved):

packages/server/lib/controllers/v1/connect/sessions/postConnectSessions.ts
bodinsamuel: yeah we pass ?env=NAME to load the proper env

packages/server/lib/controllers/v1/connect/sessions/postConnectSessions.ts
bodinsamuel: `nango_dashboard_` ?

packages/server/lib/controllers/v1/connect/sessions/postConnectSessions.ts
bodinsamuel: added a better comment

PR #2788, opened 2024-09-30
fix(connect): fields ordering

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1808/find-a-solution-to-have-a-field-order-in-the-ui

- Providers: Add optional order in the field schema
- UI: Order fields automatically if order is provided
The algo is simple and allows to put some `connectionConfig` before credentials (but not having a mix of both).

- UI: auto generate all fields

### Example

> Algolia where app id is now more logically placed before api key
Screenshot 2024-09-30 at 14 29 18

### DIFF
--- a/packages/shared/providers.yaml
+++ b/packages/shared/providers.yaml
@@ -119,6 +119,7 @@ algolia:
             description: The application ID for your Algolia account
             example: ERBSOWZO32
             pattern: '^[A-Z0-9]{10}$'
+            order: 1
     credentials:
         apiKey:
             type: string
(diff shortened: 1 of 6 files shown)


khaliqgant (approved):

packages/shared/providers.yaml
```diff
@@ -119,6 +119,7 @@ algolia:
             description: The application ID for your Algolia account
             example: ERBSOWZO32
             pattern: '^[A-Z0-9]{10}$'
+            order: 1
```
TBonnin, line 122: could we have made `connection_config` a list of key value and rely on the list ordering?

bodinsamuel: ultimately I think every field will be documented in `connection_config` and `credentials` so it's just one way to not duplicate info. But still, that's a good point indeed, maybe it would be clearer 🤔

PR #2789, opened 2024-09-30
fix(redis): move more into kvstore

bodinsamuel:
## Describe your changes

While investigating the db pool exhaustion I realized we were using db lock for stuff that can be user-generated. While in itself is not an issue, it's an issue if there are a lot of them waiting since we have a limited number of connections.
Ideally we should be using something that can scale like Redis, fortunately, we have already everything in place, except the code is still in `shared`.
This is another step towards this, no new logic is added in this PR, just code moving

- Move FeatureFlags to `kvstore` package
- Copy Locking to `kvstore` package
- Allow reuse of Redis connection pool
Right now each KVstore instantiation was creating a new connection, which could lead to issues later but also increase memory/cpu in Node.

- Move runner flags in jobs so they can use kvstore
We don't want kvstore to be imported by shared


packages/kvstore/lib/index.ts
```diff
@@ -1,17 +1,42 @@
...
 
 export { InMemoryKVStore } from './InMemoryStore.js';
 export { RedisKVStore } from './RedisStore.js';
 export type { KVStore } from './KVStore.js';
+export { FeatureFlags } from './FeatureFlags.js';
+export type { RedisClientType } from 'redis';
 
+export const redisURL = process.env['NANGO_REDIS_URL'];
+let kvStore: RedisKVStore | InMemoryKVStore | undefined;
 export async function createKVStore() {
```
TBonnin, line 14: I would rename `createKVStore` to something like `getKVStore` or similar since it now can return a memoized instance and doesn't always create a new instance

packages/kvstore/lib/RedisStore.ts
```diff
@@ -1,16 +1,12 @@
...
 export class RedisKVStore implements KVStore {
-    private client: RedisClientType;
+    public client: RedisClientType;
 
     constructor(url: string) {
-        this.client = createClient({ url: url });
-
-        this.client.on('error', (err) => {
-            console.error(`Redis (kvstore) error: ${err}`);
-        });
```
TBonnin, line 13: is removing the logging on error on purpose?

packages/kvstore/lib/index.ts
```diff
@@ -1,17 +1,42 @@
...
+        kvStore = new RedisKVStore(redisURL);
+        await kvStore.connect();
+    } else {
+        kvStore = new InMemoryKVStore();
+    }
+    return kvStore;
+}
+
+let redisClient: RedisClientType | undefined;
+export function createRedisClient({ url = redisURL, cache = true }: { url?: string; cache?: boolean } = {}): RedisClientType {
```
TBonnin, line 29: It is not super obvious what the cache parameter is actually doing unless you look at the body of the function.
What about instead having 2 functions:
- `createRedisClient` that is always creating a new client
- `getRedisClient` that is return the memoized one if exist or call createRedisClient if not

packages/server/lib/clients/publisher.client.ts
```diff
@@ -17,28 +16,27 @@ const enum MessageType {
 
 export type WebSocketClientId = string;
 
-export class Redis {
+export class RedisPubSub {
```
TBonnin, line 19: better name 👍

packages/server/lib/clients/publisher.client.unit.test.ts
```diff
@@ -19,27 +20,32 @@ const mockRes = ({ status }: { status: number }) => {
...
+    private pub: RedisClientType | undefined;
+    // @ts-expect-error unused because implements
+    private sub: RedisClientType | undefined;
+
+    public async connect(): Promise<void> {
+        // void
     }
 
-    public override async publish(channel: string, message: string) {
+    public async publish(channel: string, message: string) {
```
TBonnin, line 35: no override anymore?

packages/jobs/lib/env.ts
```diff
@@ -1,3 +1,3 @@
 import { ENVS, parseEnvs } from '@nangohq/utils';
 
-export const envs = parseEnvs(ENVS.required({ ORCHESTRATOR_SERVICE_URL: true }));
+export const envs = parseEnvs(ENVS.required({ ORCHESTRATOR_SERVICE_URL: true, NANGO_REDIS_URL: true }));
```
TBonnin, line 3: this is making Redis mandatory for enterprise setup no? are we ok with that? is it gonna break existing enterprise customer? cc @khaliqgant

bodinsamuel: ah that's why I ask on slack, I thought it was there also for enterprise. 
As I said in the PR I think our usage of PG lock can be problematic, I'm not sure keeping them is wise so hopefully we can setup a redis on entreprise

TBonnin: the locking would still works with an in-memory kvstore if there is a single instance? no?

khaliqgant: They are not using redis fwiw

bodinsamuel:
Will revisit when we have split shared and cli

Issue #2767, opened 2024-09-25
Add fal.ai integration


claudfuen:
Site: https://fal.ai/
Docs: https://fal.ai/docs

Note, you have to create an account in order to view the HTTP endpoint instructions.

# Example Request
```bash
response=$(curl --request POST \
  --url https://queue.fal.run/fal-ai/fast-turbo-diffusion \
  --header "Authorization: Key $FAL_KEY" \
  --header "Content-Type: application/json" \
  --data '{
     "prompt": "A cinematic shot of a baby racoon wearing an intricate italian priest robe."
   }')
REQUEST_ID=$(echo "$response" | grep -o '"request_id": *"[^"]*"' | sed 's/"request_id": *//; s/"//g')
```

```bash
curl --request GET \
  --url https://queue.fal.run/fal-ai/fast-turbo-diffusion/requests/$REQUEST_ID/status \
  --header "Authorization: Key $FAL_KEY"
```

khaliqgant:
Done! 
<img width="1032" alt="image" src="https://github.com/user-attachments/assets/da0408e8-bbb9-4a76-bd52-b662a4e54054">

PR #2791, opened 2024-09-30
feat(integrations): add support for fal.ai

Shubhamai:
## Describe your changes
- Added fal.ai support.

https://github.com/NangoHQ/nango/issues/2767


packages/shared/providers.yaml
```diff
@@ -5151,3 +5151,18 @@ zoominfo:
     token_request_auth_method: custom
     body_format: json
     docs: https://docs.nango.dev/integrations/all/zoominfo
+
+
+falai:
+    display_name: fal.ai
```
bodinsamuel, line 5265: missing categories (I guess we should enforce it in the CI)

packages/shared/providers.yaml
```diff
@@ -5151,3 +5151,18 @@ zoominfo:
...
+    auth_mode: API_KEY
+    proxy:
```
bodinsamuel, line 5270: can you add a `verification` prop please ☺️

bodinsamuel (commented):
🚀

hassan254-prog:
same as here, please use this [logo](https://codefile.io/f/yYgBgU6fHp)

Shubhamai:
Thanks, made the review changes. Except review prop. I don't think there's endpoint in fal.ai to verify the API Key without causing unintended effects.

bodinsamuel:
>Thanks, made the review changes. Except review prop. I don't think there's endpoint in fal.ai to verify the API Key without causing unintended effects.

Mmh indeed that's unfortunate

bodinsamuel (approved):

hassan254-prog (approved):

PR #2792, opened 2024-09-30
feat: add middleware to combine connect session and private key auth

TBonnin:
Combine private key and connect session token auth in one additional authorization middleware because some api endpoint must be now accessible client side by the connect UI

https://linear.app/nango/issue/NAN-1829/combine-connect-session-and-private-key-auth-middleware


packages/server/lib/middleware/access.middleware.ts
```diff
@@ -1,18 +1,49 @@
...
 import tracer from 'dd-trace';
 import type { RequestLocals } from '../utils/express.js';
+import type { ConnectSession, DBEnvironment, DBTeam } from '@nangohq/types';
 
 const logger = getLogger('AccessMiddleware');
 
 const keyRegex = /^[0-9A-F]{8}-[0-9A-F]{4}-[4][0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i;
 const ignoreEnvPaths = ['/api/v1/meta', '/api/v1/user', '/api/v1/user/name', '/api/v1/signin', '/api/v1/invite/:id'];
 
 export class AccessMiddleware {
```
TBonnin, line 19: I wish all the middleware functions were composable but since we return early, sending 401 error it is not as easy as I would like to. I therefore extracted some some of the logic of the 2 middleware so it can be reused in a 3rd one.

packages/server/lib/middleware/access.middleware.ts
```diff
@@ -177,50 +230,77 @@ export class AccessMiddleware {
...
+                environmentId: String(result.value.environment.id),
+                connectSessionId: String(result.value.connectSession.id)
             });
             next();
-        } catch (err) {
-            logger.error(`failed_get_env_by_connect_session_token ${stringifyError(err)}`);
-            return errorManager.errRes(res, 'malformed_auth_header');
+        } catch (e) {
+            logger.error(`failed_get_env_by_connect_session ${stringifyError(e)}`);
+            return errorManager.errRes(res, 'unknown_account');
```
TBonnin, line 260: not sure why the err in the catch was set to `malformed_auth_header`. Maybe a bad copy pasting

packages/shared/lib/utils/error.ts
```diff
@@ -34,7 +34,7 @@ export class NangoError extends Error {
 
             case 'unknown_account':
                 this.status = 401;
-                this.message = 'Authentication failed. The provided secret/public key does not match any account.';
+                this.message = 'Authentication failed. The provided authorization header does not match any account.';
```
TBonnin, line 37: Since my last PRs the message also applies to connect session

packages/server/lib/middleware/access.middleware.ts
bodinsamuel: yep indeed

packages/server/lib/routes.integration.test.ts
```diff
@@ -1,34 +1,86 @@
...
         await multipleMigrations();
         await migrateLogsMapping();
+        await migrateKeystore(db.knex);
 
         api = await runServer();
     });
     afterAll(() => {
         api.server.close();
     });
+    describe('GET /logs', () => {
```
bodinsamuel: still wrong though, it's calling`/api/v1/environment` (my bad)

packages/server/lib/routes.ts
```diff
@@ -187,8 +188,8 @@ publicAPI.route('/config/:providerConfigKey').get(apiAuth, configController.getP
 publicAPI.route('/config').post(apiAuth, configController.createProviderConfig.bind(configController));
 publicAPI.route('/config').put(apiAuth, configController.editProviderConfig.bind(configController));
 publicAPI.route('/config/:providerConfigKey').delete(apiAuth, deletePublicIntegration);
-publicAPI.route('/integrations').get(apiAuth, getPublicListIntegrations);
-publicAPI.route('/integrations/:uniqueKey').get(apiAuth, getPublicIntegration);
+publicAPI.route('/integrations').get(connectSessionOrApiAuth, getPublicListIntegrations);
+publicAPI.route('/integrations/:uniqueKey').get(connectSessionOrApiAuth, getPublicIntegration);
```
bodinsamuel, line 193: forgot to say we also need all the `/oauth/ + /auth/ + /api-auth/` routes

packages/server/lib/routes.integration.test.ts
```diff
@@ -1,34 +1,86 @@
...
+            });
         });
+    });
 
-        expect(await res.json()).toStrictEqual({
-            error: {
-                code: 'invalid_json',
-                message: expect.any(String) // unfortunately the message is different depending on the platform
-            }
+    describe('GET /integrations', () => {
```
bodinsamuel: ideally this is enforced at each endpoint in their respective integration tests (when available), also we need to enforce the opposite now

packages/server/lib/middleware/access.middleware.ts
```diff
@@ -169,6 +184,44 @@ export class AccessMiddleware {
...
+            return Err('missing_auth_header');
+        }
+
+        const token = authorizationHeader.split('Bearer ').pop();
+
+        if (!token) {
+            return Err('malformed_auth_header');
+        }
+
+        const getConnectSession = await connectSessionService.getConnectSessionByToken(db.knex, token);
```
bodinsamuel, line 200: since it's done at every call I would at least check it has the right prefix before looking in DB so you can early exit.

In the case of prefix but wrong key you should also exit but not continue to secretKey check. Because right now it's returning `The provided secret key is not a UUID v4` on invalid sessionToken

bodinsamuel (commented):
Works well 🚁
Just a few open comments and we are almost done 🙏🏻

bodinsamuel (approved):
(meant to approve)

packages/server/lib/routes.integration.test.ts
TBonnin: will fix

packages/server/lib/controllers/connect/postSessions.integration.test.ts
```diff
@@ -14,8 +13,6 @@ describe(`POST ${endpoint}`, () => {
     let seed: { env: DBEnvironment };
 
     beforeAll(async () => {
-        await multipleMigrations();
-        await migrateKeystore(db.knex);
```
TBonnin, line 18: moved the migrations to `runServer` since it is a hard dependency anyway

packages/server/lib/routes.ts
TBonnin: I will change the auth middleware for auth endpoints in the next PR when I modified the frontend sdk where those endpoints are being called from

packages/server/lib/utils/tests.ts
```diff
@@ -121,6 +124,10 @@ export function shouldRequireQueryEnv({ res, json }: { res: Response; json: any
  * Run the API in the test
  */
 export async function runServer(): Promise<{ server: Server; url: string; fetch: ReturnType<typeof apiFetch> }> {
+    await multipleMigrations();
+    await migrateLogsMapping();
+    await migrateKeystore(db.knex);
```
bodinsamuel, line 129: good call

bodinsamuel (approved):
Amazing, massive work 💪🏻

bodinsamuel:
Going to merge this, I'm a bit stuck without this PR

PR #2794, opened 2024-10-01
fix(sdk): sample record validation errors

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1833/reduce-validation-logs-andor-process-in-parallel

- Sample record validation errors
It's spamming the infra, making syncs slower than they should be.

- Send the sample in parallel
Don't remember why I didn't do it like this. It complexifies the code a bit but faster; since we sample it's no longer a big problem but can be useful if we increase the sampling or remove it.

> The wording is not crazy good, if you have suggestions...
Screenshot 2024-10-01 at 11 37 15


khaliqgant (approved):

packages/shared/lib/sdk/sync.ts
```diff
@@ -941,16 +944,31 @@ export class NangoSync extends NangoAction {
                 continue;
             }
 
+            hasErrors.push({ data: record, validation });
             metrics.increment(metrics.Types.RUNNER_INVALID_SYNCS_RECORDS);
 
+            if (this.runnerFlags?.validateSyncRecords) {
+                break;
+            }
```
TBonnin, line 952: what is this `if` for?

bodinsamuel: early break if enabled

bodinsamuel: otherwise it just logs everything*

TBonnin: hasErrors is never gonna have more than 1 element then if we always early break

bodinsamuel: yes sorry it's maybe not clear, this flag is only set to true in CLI

TBonnin: ah ok. so in CLI it breaks out after the first validation error

TBonnin (approved):

PR #2797, opened 2024-10-01
feat(connect): support for domain suffix, hidden fields, default value

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1834/support-for-domain-suffix-hidden-fields-default-value

- Support for domain suffix
- Support for hidden fields 
For fields that are going to be shown in the UI (because they have a default value). Will be useful when I get session token values
- Support for default value
It's useful for some providers with static values (e.g: Freshdesk has a static password because they choose to use basic auth with an api key 🤦🏻 ). Will also be useful when I get session token values

- Backfill domain suffix
- Add some warnings in `validate.ts` so I can track progress

--

Screenshot 2024-10-01 at 15 06 11


scripts/validation/providers/validate.ts
```diff
@@ -117,10 +117,23 @@ function validateProvider(providerKey: string, provider: Provider) {
...
 
     if (provider.auth_mode === 'API_KEY') {
-        if (!provider?.credentials?.['apiKey']) {
+        if (!provider.credentials?.['apiKey']) {
             console.error(chalk.red('error'), chalk.blue(providerKey), `"credentials" > "apiKey" is not defined`);
             error = true;
         }
+        if (!provider.proxy?.verification) {
+            console.warn(chalk.yellow('warning'), chalk.blue(providerKey), `do not have "proxy" > "verification" set`);
+        }
```
TBonnin, line 126: should this warning always exist regardless of the `auth_mode`?

TBonnin (approved):

scripts/validation/providers/validate.ts
bodinsamuel: it's mostly there to verify credentials at save, with oauth we already know the credentials are good

PR #2799, opened 2024-10-01
fix(quickbooks-sandbox): [nan-1824] update quickbooks sandbox

khaliqgant:
## Describe your changes
Add quickbooks sandbox

NAN-1824


TBonnin (approved):

bodinsamuel (approved):
Isn't flow supposed to updated by the other repo?

khaliqgant:
> Isn't flow supposed to updated by the other repo?

Yeah it will. I just updated while testing and it doesn't really matter if it gets edited here as it will get overridden from integration-templates

PR #2594, opened 2024-08-08
feat: Integration UI revamp

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1542/script-v2-integrate-settings
Fixes https://linear.app/nango/issue/NAN-1537/script-v2-integrate-ui
Fixes https://linear.app/nango/issue/NAN-1429/editing-an-integration-name-to-an-existing-integration-silently-errors

### Major Changes

- Rework `PATCH /flows/:id/enable`
- Rework `PATCH /flows/:id/disable`
- Rework `PATCH /flows/:id/frequency`
- Rework `POST /integrations/`
- Rework `GET /integrations/:id`
- Rework `PATCH /integrations/:id`
- Rework `GET /integrations/:id/flows`

- Rework UI /integrations/:id 
New endpoints page, settings page revamp (and split components per credentials type), new endpoint page. The goal of this revamp is to present integrations as a list of endpoints rather than a list of scripts.

### 🧪 How to review and test?

This PR is massive since it touches a lot of components and endpoints, sorry about that. The best way to review that is to begin with the UI and check the code. I have extensively tested it, but there is a lot to go over and there are probably some edge cases.

- Create a new integration
- Enable / Disable actions and syncs
- Edit integration credentials (if you can test: try doing it for multiple types of credentials)
- Check one endpoint, access the URL directly,
- Change frequency of an endpoint 
- Delete an integration

> [!NOTE]
> UI/UX has been reviewed with Bastien but we are not sure about everything, feel free to raise confusing points or things that could be improved

## Summary by CodeRabbit

- **New Features**
	- Introduced new endpoints for managing integrations, including retrieval of integration flows and details based on configuration keys.
	- Added a customizable Select component for enhanced user interactions.
	- Enhanced the user interface with new components for displaying and managing API endpoints.
	- Implemented a new endpoint for updating synchronization flow frequencies.

- **Improvements**
	- Enhanced validation logic for various parameters across multiple components.
	- Updated API specifications to reflect new response structures and improved user feedback mechanisms during integration operations.

- **Bug Fixes**
	- Corrected minor issues in error messages and validation checks.

- **Chores**
	- Removed outdated API utility functions to simplify the codebase.
	- Added new dependencies to support enhanced UI components.

PR #2801, opened 2024-10-01
fix: sync frequency can be edited to be 30s or more

TBonnin:
This is a regression issue
When revamping the integration UI https://github.com/NangoHQ/nango/pull/2594, the [change](https://github.com/NangoHQ/nango/pull/2690/files#diff-5c0e007ca978cb035dcb1b4c38cacc650a268098a324a043d807fc7e2311ff78R203-R242) to allow sync frequency lower than 5mins was lost 

### DIFF
--- a/packages/server/lib/controllers/v1/flows/id/patchFrequency.ts
+++ b/packages/server/lib/controllers/v1/flows/id/patchFrequency.ts
@@ -13,7 +13,7 @@ export const validationBody = validationBodyBase.extend({
     frequency: z
         .string()
         .regex(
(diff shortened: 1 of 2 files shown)


bodinsamuel (approved):
Arg did I completely missed this or we did both change at the same time?

TBonnin:
> Arg did I completely missed this or we did both change at the same time?

I think I made the change while you were working on your refactoring branch and we missed it in the review

bodinsamuel:
> I think I made the change while you were working on your refactoring branch and we missed it in the review

Arf, my bad should have caught that, thanks for fixing

PR #2806, opened 2024-10-02
feat: add support for connect session token auth in frontend sdk

TBonnin:
- Adding support for connect session token in frontend sdk
- Adding auth middleware to authorize via either public key or connnect session token

https://linear.app/nango/issue/NAN-1837/sdk-support-for-sessiontoken

### DIFF
--- a/packages/frontend/lib/index.ts
+++ b/packages/frontend/lib/index.ts
@@ -39,7 +39,8 @@ export default class Nango {
     private hostBaseUrl: string;
     private websocketsBaseUrl: string;
     private status: AuthorizationStatus;
-    private publicKey: string;
+    private publicKey: string | undefined;
+    private connectSessionToken: string | undefined;
     private debug = false;
     private width: number | null = null;
     private height: number | null = null;
@@ -47,7 +48,15 @@ export default class Nango {
 
     public win: AuthorizationModal | null = null;
 
-    constructor(config: { host?: string; websocketsPath?: string; publicKey: string; width?: number; height?: number; debug?: boolean }) {
+    constructor(config: {
+        host?: string;
+        websocketsPath?: string;
+        publicKey?: string;
+        connectSessionToken?: string;
+        width?: number;
+        height?: number;
+        debug?: boolean;
+    }) {
         config.host = config.host || prodHost; // Default to Nango Cloud.
         config.websocketsPath = config.websocketsPath || '/'; // Default to root path.
         this.debug = config.debug || false;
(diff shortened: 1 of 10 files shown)


packages/frontend/lib/index.ts
```diff
@@ -350,6 +353,29 @@ export default class Nango {
...
+            ...(credentials ? { body: JSON.stringify(credentials) } : {})
+        });
+
+        if (!res.ok) {
+            const errorResponse = await res.json();
+            throw new AuthError(errorResponse.error.message, errorResponse.error.code);
+        }
+
+        return res.json();
+    }
```
TBonnin, line 377: refactoring to avoid the fetch duplication for all kind of auth

packages/server/lib/middleware/access.middleware.ts
```diff
@@ -67,6 +67,7 @@ export class AccessMiddleware {
             next();
         } catch (err) {
             logger.error(`failed_get_env_by_secret_key ${stringifyError(err)}`);
+            span.setTag('error', err);
```
TBonnin, line 69: unrelated but we forgot to set the span as errored

packages/server/lib/middleware/access.middleware.ts
```diff
@@ -255,8 +234,9 @@ export class AccessMiddleware {
                 connectSessionId: String(result.value.connectSession.id)
             });
             next();
-        } catch (e) {
-            logger.error(`failed_get_env_by_connect_session ${stringifyError(e)}`);
+        } catch (err) {
```
TBonnin, line 237: renaming to `err` for consistency with the rest of the file

bodinsamuel: It reminds me I need to load a eslint rule for that

packages/server/lib/controllers/auth/postUnauthenticated.ts
```diff
@@ -13,7 +13,8 @@ import { connectionCreated, connectionCreationFailed } from '../../hooks/hooks.j
 const queryStringValidation = z
     .object({
         connection_id: connectionIdSchema.optional(),
-        public_key: z.string().uuid(),
+        public_key: z.string().uuid().optional(),
+        connect_session_token: z.string().startsWith('nango_connect_session_').optional(),
```
bodinsamuel: note you can use `packages/server/lib/helpers/validation.ts` to store reusable zod schema

packages/types/lib/auth/http.api.ts
```diff
@@ -40,7 +39,6 @@ export type PostPublicTableauAuthorization = Endpoint<{
     };
     Querystring: {
         connection_id?: string | undefined;
-        public_key: string;
```
bodinsamuel, line 43: isn't the actual fix public_key + session_token ?
Also I see it was missing in PostPublicUnauthenticatedAuthorization

bodinsamuel (approved):
Looks good, will test more deeply once merged

packages/types/lib/auth/http.api.ts
TBonnin: we can add them both but they are never used in the endpoint logic so decided not to have them. wdyt?

bodinsamuel: yeah I see your point, ideally, we have something always correct but typescript is too lax on partial objects. no need to fix 👍🏻

PR #2807, opened 2024-10-02
feat(integrations): rapidapi support

Shubhamai:
## Describe your changes
Added [RapidAPI](http://rapidapi.com) support.


packages/shared/providers.yaml
```diff
@@ -3826,6 +3826,29 @@ ramp-sandbox:
...
+        apiKey:
+            type: string
+            title: API Key
+            description: The API key for your RapidAPI account
+    connection_config:
+        host:
+            type: string
+            title: Host
+            description: The host of the API (e.g. api-football-v1.p.rapidapi.com)
+            pattern: '^[a-z0-9_-]+$'
```
khaliqgant, line 3890: Would something like `'^[a-z0-9_-]+(\.[a-z0-9_-]+)*$'` be better for the pattern to match on the periods?

Shubhamai: Not sure about this, for ex: here's a sample API key from RapidAPI: `b7c156af2dmgh5c635305f3744bap168553jsp75193c8367ef`

khaliqgant: But I thought this was the pattern for the host, not the value of the API key?

packages/shared/providers.yaml
```diff
@@ -3826,6 +3826,29 @@ ramp-sandbox:
...
+        headers:
+            x-rapidapi-key: ${apiKey}
+            x-rapidapi-host: ${connectionConfig.host}
+        base_url: https://${connectionConfig.host}
+    docs: https://docs.nango.dev/integrations/all/rapidapi
+    credentials:
+        apiKey:
+            type: string
+            title: API Key
+            description: The API key for your RapidAPI account
```
bodinsamuel, line 3882: You can add pattern and example here 
```suggestion
            description: The API key for your RapidAPI account
            example: b7c156af2dmgh5c635305f3744bap168553jsp75193c8367ef
            pattern: '^[a-zA-Z0-9]+$'
```

packages/shared/providers.yaml
```diff
@@ -3826,6 +3826,29 @@ ramp-sandbox:
...
+rapidapi:
+    display_name: RapidAPI
+    categories:
+        - dev-tools
+    auth_mode: API_KEY
+    proxy:
+        headers:
+            x-rapidapi-key: ${apiKey}
+            x-rapidapi-host: ${connectionConfig.host}
+        base_url: https://${connectionConfig.host}
```
bodinsamuel: can you change both base_url and host to be something like `https://${connectionConfig.subdomain}.rapidapi.com`?

packages/shared/providers.yaml
```diff
@@ -3826,6 +3826,29 @@ ramp-sandbox:
...
+    credentials:
+        apiKey:
+            type: string
+            title: API Key
+            description: The API key for your RapidAPI account
+    connection_config:
+        host:
+            type: string
+            title: Host
+            description: The host of the API (e.g. api-football-v1.p.rapidapi.com)
```
bodinsamuel: with the other change in base_url can you add `suffix: rapidapi.com`

packages/shared/providers.yaml
Shubhamai: Sorry, completely missed it. Yes you are right, I think this should work for host url.

bodinsamuel (commented):
Hey, thanks for your contribution. Added some comments regarding the subdomain, let me know if you want us to takeover ☺️

bodinsamuel (approved):
🙏🏻

khaliqgant (approved):

PR #2809, opened 2024-10-03
feat(connect): open beta

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1853/productize

- Enable UI in cloud env
- Allow UI to open Connect UI in an iframe
- Less hardcoding of api hostname
Still not good though, it will be impossible to not pass an env var at compile time. It's the same issue for the current dashboard, the trick I used to load dynamic env with `/env.js` is only working because it's loaded in the same subdomain. I guess that means we'll still need to build two docker image staging + prod unless we remove frontend from the docker image.

- Clear open popup on early exit

It's currently accessible in staging

### DIFF
--- a/packages/connect-ui/src/views/Go.tsx
+++ b/packages/connect-ui/src/views/Go.tsx
@@ -62,8 +62,8 @@ const defaultConfiguration: Record<string, { secret: boolean; title: string; exa
 };
 
 export const Go: React.FC = () => {
-    const { provider, integration, session, sessionToken, setIsDirty } = useGlobal();
-    const nango = useNango(sessionToken);
+    const { provider, integration, session, setIsDirty } = useGlobal();
+    const nango = useNango();
 
     const [loading, setLoading] = useState(false);
     const [result, setResult] = useState<AuthResult>();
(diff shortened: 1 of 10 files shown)


packages/connect-ui/src/views/Go.tsx
```diff
@@ -334,7 +334,9 @@ export const Go: React.FC = () => {
                                 );
                             })}
                         </div>
-                        {shouldAutoTrigger && <div className="text-sm text-dark-500 w-full text-center">{loading && 'A popup is opened...'}</div>}
+                        {shouldAutoTrigger && provider.auth_mode !== 'NONE' && (
```
TBonnin, line 337: can the `provider.auth_mode` be factored into `shouldAutoTrigger`?

packages/webapp/src/pages/Connection/List.tsx
```diff
@@ -187,11 +188,7 @@ export default function ConnectionList() {
             <div className="flex justify-between mb-8 items-center">
                 <h2 className="flex text-left text-3xl font-semibold tracking-tight text-white">Connections</h2>
                 <div className="flex gap-2">
-                    {globalEnv.features.connectUI && (
-                        <Button onClick={onClickConnectUI} variant={'emptyFaded'}>
-                            Open Connect UI (alpha)
-                        </Button>
-                    )}
+                    {globalEnv.features.connectUI && <Button onClick={onClickConnectUI}>Open Connect UI (beta)</Button>}
```
TBonnin, line 191: open beta for everybody?

TBonnin (approved):
> I guess that means we'll still need to build two docker image staging + prod unless we remove frontend from the docker image.

not ideal but would require a bit of effort

packages/connect-ui/src/views/Go.tsx
bodinsamuel: ~true~ actually no, it should auto trigger for unauthenticated anyway just not display that specific message

packages/webapp/src/pages/Connection/List.tsx
bodinsamuel: I guess. we can load a feature flag if Bastien wants it

PR #2812, opened 2024-10-03
fix(connect): defer token creation after iframe open

bodinsamuel:
## Describe your changes

- Defer token creation after iframe open
Allow to pass the token after the iframe is created so we can display a loading screen instead of just nothing

### DIFF
--- a/packages/webapp/src/pages/Connection/List.tsx
+++ b/packages/webapp/src/pages/Connection/List.tsx
@@ -110,13 +110,8 @@ export default function ConnectionList() {
         };
     }, [debouncedSearch]);
 
-    const onClickConnectUI = async () => {
+    const onClickConnectUI = () => {
         if (!environmentAndAccount) {
-            return;
-        }
-
-        const res = await apiConnectSessions(env);
-        if ('error' in res.json) {
             return;
         }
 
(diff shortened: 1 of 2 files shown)


packages/webapp/src/pages/Connection/List.tsx
```diff
@@ -128,14 +123,23 @@ export default function ConnectionList() {
...
+
+        // We defer the token creation so the iframe can open and display a loading screen
+        //   instead of blocking the main loop and no visual clue for the end user
+        setTimeout(async () => {
+            const res = await apiConnectSessions(env);
+            if ('error' in res.json) {
+                return;
+            }
+            connectUI.current!.setSessionToken(res.json.data.token);
+        }, 10);
```
TBonnin, line 142: that's a pretty short value. no? what happen if the call to `connect/sessions` takes longer than that?

TBonnin (approved):
one question about the timeout duration but lgtm otherwise

packages/webapp/src/pages/Connection/List.tsx
bodinsamuel: The call is triggered after 10ms.

The setTimeout is just here to free the main event loop, which means the iframe can be created and the URL loading started, the iframe doesn't have a separate sandbox so it's important that it has time to load and display something. 
Then the setTimeout triggers and it takes as much time as it needs. 
10ms is not a lot but on a good connection waiting for nothing would be detrimental to UX.

PR #2815, opened 2024-10-03
fix(providers): add new token_url_encode parameter to provider config

gorets:
## Describe your changes

This need for specific cases (like aaa/bbb in the subdomain URL)


TBonnin (commented):
Thank you for your contribution @gorets 
Can you share with us an example of provider where token url must be url encoded?

Also can you add this new property to the docs and spec schema? Thank you
- https://github.com/NangoHQ/nango/blob/244b782d7bf9ad4b94a1207275136d346ef04aa8/docs-v2/spec.yaml#L1858
- https://github.com/NangoHQ/nango/blob/244b782d7bf9ad4b94a1207275136d346ef04aa8/scripts/validation/providers/schema.json#L77
- https://github.com/NangoHQ/nango/blob/244b782d7bf9ad4b94a1207275136d346ef04aa8/docs-v2/reference/api-configuration.mdx?plain=1#L252

gorets:
> Thank you for your contribution @gorets Can you share with us an example of provider where token url must be url encoded?

We have very specific case with Salesforce Experience Cloud, where we use fully custom domain with additional tenant.
We have custom authorization URL look like `https://custom.domain.com/tenant`

We changed our provider config for `salesforce-experience-cloud`
```
    authorization_url: https://${connectionConfig.subdomain}/services/oauth2/authorize
    token_url: `https://${connectionConfig.subdomain}/services/oauth2/token
```

and uses next value for subdomain  `custom.domain.com/tenant`

But after last update our hack is't worked :(

I understand that this is a very unique case, but I would be very grateful for a solution to our problem.

bodinsamuel (approved):

khaliqgant (approved):

PR #2818, opened 2024-10-04
feat(integrations): add support for Buildium

hassan254-prog:
## Describe your changes

- Add support for buildium

[EXT-180](https://linear.app/nango/issue/EXT-180/add-support-for-buildium)


packages/shared/providers.yaml
```diff
@@ -733,6 +733,34 @@ builder-io-public:
...
+    connection_config:
+        clientId:
+            type: string
+            title: Client ID
+            description: Your API Key Client ID
+    credentials:
+        apiKey:
+            type: string
+            title: API Key
+            description: The API key for your Buildium account
```
TBonnin, line 764: there is new attributes like pattern or example? can you add them here.

hassan254-prog: updated [with](https://github.com/NangoHQ/nango/pull/2818/commits/ab9594d098ce6ef2c3598ef1eb44aa56545968bc), picked an example of both from their API documentation

packages/shared/providers.yaml
```diff
@@ -755,11 +755,15 @@ buildium:
...
             description: Your API Key Client ID
+            pattern: '^[a-z0-9_-]+$'
+            example: a228f0e7-b4a3-4150-b9ae-8552fc2880d3
     credentials:
         apiKey:
             type: string
             title: API Key
             description: The API key for your Buildium account
+            pattern: '^[a-z0-9_-]+$'
+            example: uOq2p+xlgpFdijfV/HqY+EvYpZKHRwlyhGuAVbJIxXs
```
TBonnin, line 766: the regex doesn't match the value. It contains `+` or `/` for instance. You probably want `^[A-Za-z0-9+/]+$` instead

TBonnin (changes requested):

packages/shared/providers.yaml
hassan254-prog: aah yes, updated with [commit](https://github.com/NangoHQ/nango/pull/2818/commits/457afb07627c43917bf969a1cdaca95a6081af97)

TBonnin (approved):

PR #2826, opened 2024-10-07
fix: enable possibility to run remote runner

TBonnin:
For enterprise customers who want to be able to run remote runner:
- Starting/Getting runner based on REMOTE_RUNNER_TYPE env var

### DIFF
--- a/packages/jobs/lib/runner/runner.ts
+++ b/packages/jobs/lib/runner/runner.ts
@@ -5,6 +5,7 @@ import { isEnterprise, env, getLogger } from '@nangohq/utils';
 import type { ProxyAppRouter } from '@nangohq/nango-runner';
 import type { KVStore } from '@nangohq/kvstore';
 import { createKVStore } from '@nangohq/kvstore';
+import { envs } from '../env.js';
 
 const logger = getLogger('Runner');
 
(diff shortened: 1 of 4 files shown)


packages/jobs/lib/runner/runner.ts
```diff
@@ -54,12 +55,17 @@ export async function getOrStartRunner(runnerId: string): Promise<Runner> {
             logger.error(err);
         }
     }
-    const isRender = process.env['IS_RENDER'] === 'true';
     let runner: Runner;
-    if (isRender) {
-        runner = await RenderRunner.getOrStart(runnerId);
-    } else {
-        runner = await LocalRunner.getOrStart(runnerId);
+    switch (envs.RUNNER_TYPE) {
```
khaliqgant, line 59: should we do toUpperCase to ensure it is consistent? Or is that handled already?

TBonnin: already handled by z.enum. `envs.RUNNER_TYPE` is a union type `'LOCAL' | 'RENDER' | 'REMOTE'`

khaliqgant (approved):

bodinsamuel (approved):

PR #2827, opened 2024-10-07
fix(providers): missing $ in snowflake

bodinsamuel:
## Describe your changes

- Fix bad interpolation in providers.yaml
- Add a way to prevent that in the future


khaliqgant (approved):

scripts/validation/providers/validate.ts
```diff
@@ -41,6 +41,13 @@ if (validator.errors) {
     process.exit(1);
 }
 
+const invalidInterpolation = /(?<!(\$|]))\{/g;
```
TBonnin, line 44: why the `]` condition?

TBonnin (approved):

scripts/validation/providers/validate.ts
bodinsamuel: there are some regex like [a-z]{10} so it should not warn if it matches a ]{

PR #2829, opened 2024-10-07
fix: out of sync between task state and sync_job status

TBonnin:
When user currently cancels a task we only update the sync_job entry if the abort task is successfully processed by jobs.
However if jobs for some reason is having issue, the customer is left with seeing the task still running in the dashboard. With this commit the sync job is always updated when cancellation is clicked and doesn't depend on Job actually aborting the task. Even if jobs is misbehaving the scheduler will eventually recover by cancelling/timing out the task

https://linear.app/nango/issue/NAN-1865/fix-sync-status-out-of-sync-on-connections-page

### DIFF
--- a/packages/jobs/lib/execution/sync.ts
+++ b/packages/jobs/lib/execution/sync.ts
@@ -21,7 +21,7 @@ import {
     configService,
     createSyncJob,
     getSyncConfigRaw,
-    isSyncJobRunning
+    getSyncJobByRunId
 } from '@nangohq/shared';
 import { Err, Ok, metrics } from '@nangohq/utils';
 import type { Result } from '@nangohq/utils';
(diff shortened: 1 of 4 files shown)


packages/jobs/lib/execution/sync.ts
```diff
@@ -463,7 +463,7 @@ export async function abortSync(task: TaskSyncAbort): Promise<Result<void>> {
             logger.error(`failed to abort script for task ${task.abortedTask.id}: ${abortedScript.error}`);
         }
 
-        const syncJob = await isSyncJobRunning(task.syncId);
```
TBonnin, line 466: we cannot get the running sync job since we updated its status to STOPPED as soon as the user clicked on cancel. 
Don't remember why I did it this way originally, I guess I lazily used an existing function without thinking to much of the side effect

bodinsamuel (approved):
Ah that's always a problem when not exposing internal jobs and "ideal state" vs "current state"

PR #2830, opened 2024-10-08
fix: dynamic getOutpout longPolling timeout

TBonnin:
This is not the best solution since it means potentially keeping connection open for a very long time but for now the retry logic is broken. I will fix it in next PR but at least it will fix the action taking longr than 2mins

### DIFF
--- a/packages/orchestrator/lib/routes/v1/tasks/taskId/getOutput.ts
+++ b/packages/orchestrator/lib/routes/v1/tasks/taskId/getOutput.ts
@@ -13,9 +13,9 @@ type GetOutput = Endpoint<{
         taskId: string;
     };
     Querystring: {
-        longPolling?: boolean;
+        longPolling?: number | undefined;
     };
-    Error: ApiError<'task_not_found'>;
+    Error: ApiError<'task_not_found' | 'timeout'>;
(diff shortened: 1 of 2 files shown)


khaliqgant (approved):

packages/orchestrator/lib/routes/v1/tasks/taskId/getOutput.ts
```diff
@@ -55,7 +51,7 @@ const handler = (scheduler: Scheduler, eventEmitter: EventEmitter) => {
             cleanupAndRespond((res) => res.status(200).json({ state: completedTask.state, output: completedTask.output }));
         };
         const timeout = setTimeout(() => {
-            cleanupAndRespond((res) => res.status(204).send());
+            cleanupAndRespond((res) => res.status(500).send({ error: { code: 'timeout', message: 'Long polling timeout' } }));
```
TBonnin, line 54: There is an issue with the API types and express. Since the response body can actually be empty, we can get an undefined result even though the compiler said that's not the case. That's the source of the `mapError` called on undefined. I will fix in another PR. 
Anyway sending 204 is not a very good solution and it is better to explicitely differentiate an empty response vs a timeout

PR #2831, opened 2024-10-08
fix: retry until to fetch task output instead of long-lived request

TBonnin:
## Describe your changes
Instead of keeping the request to orchestrator `GET /task/ID/output` alive for potentially minutes, we do shorter requests and retry until the action task is supposed to expire

### DIFF
--- a/packages/orchestrator/lib/clients/client.ts
+++ b/packages/orchestrator/lib/clients/client.ts
@@ -8,8 +8,8 @@ import { route as postSchedulesSearchRoute } from '../routes/v1/schedules/postSe
 import { route as getOutputRoute } from '../routes/v1/tasks/taskId/getOutput.js';
 import { route as putTaskRoute } from '../routes/v1/tasks/putTaskId.js';
 import { route as postHeartbeatRoute } from '../routes/v1/tasks/taskId/postHeartbeat.js';
-import type { Result, Route } from '@nangohq/utils';
-import { Ok, Err, routeFetch, getLogger, retryWithBackoff, httpRetryStrategy } from '@nangohq/utils';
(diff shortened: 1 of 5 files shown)


packages/shared/lib/clients/orchestrator.ts
```diff
@@ -137,6 +137,7 @@ export class Orchestrator {
                 args
             });
 
+            console.log(actionResult, 'actionResult');
```
khaliqgant: log

packages/utils/lib/express/route.ts
```diff
@@ -79,6 +79,7 @@ export const routeFetch = <E extends Endpoint<any>>(
             }
             return json;
         } catch (error: unknown) {
+            console.log('error', error);
```
khaliqgant: log

packages/orchestrator/lib/clients/client.ts
```diff
@@ -40,11 +40,20 @@ export class OrchestratorClient {
     private routeFetch<E extends Endpoint<any>>(
         route: Route<E>,
         config?: {
-            timeoutMs: number;
+            timeoutMs?: number;
+            retryConfig?: RetryConfig<E['Reply']>;
         }
     ): (props: { query?: E['Querystring']; body?: E['Body']; params?: E['Params'] }) => Promise<E['Reply']> {
         return (props) => {
-            return retryWithBackoff(() => routeFetch(this.baseUrl, route, config)(props), { retry: httpRetryStrategy, numOfAttempts: 3, maxDelay: 50 });
```
TBonnin, line 47: had to refactor `retry.ts` because it only handles function that throw error. The orchestrator typed API isn't throwing error but return them as values

khaliqgant (approved):

PR #2832, opened 2024-10-09
fix: bump jobs/orchestrator server limit for incoming request

TBonnin:
the goal is to allow actions' (or any task) output to be bigger than 10mb

### DIFF
--- a/packages/orchestrator/lib/constants.ts
+++ b/packages/orchestrator/lib/constants.ts
@@ -1,0 +1,1 @@
+export const serverRequestSizeLimit = '30mb';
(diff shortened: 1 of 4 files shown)


packages/orchestrator/lib/constants.ts
```diff
@@ -0,0 +1 @@
+export const serverRequestSizeLimit = '30mb';
```
TBonnin, line 1: making it a constant in orchestrator so it is linked explicitely.

khaliqgant (approved):

PR #2835, opened 2024-10-10
fix(integration): improve guru integration

hassan254-prog:
## Describe your changes

- This change is necessary to properly handle pagination when processing` nextPageLink` in Guru. By using the updated base URL, the endpoint construction avoids redundancy.

### DIFF
--- a/packages/shared/providers.yaml
+++ b/packages/shared/providers.yaml
@@ -1960,7 +1960,7 @@ guru:
         - knowledge-base
     auth_mode: BASIC
     proxy:
-        base_url: https://api.getguru.com/api/v1
+        base_url: https://api.getguru.com
     docs: https://docs.nango.dev/integrations/all/guru
 
 github:


packages/shared/providers.yaml
```diff
@@ -1960,7 +1960,7 @@ guru:
         - knowledge-base
     auth_mode: BASIC
     proxy:
-        base_url: https://api.getguru.com/api/v1
```
khaliqgant, line 1963: We have to be careful with this for users who are using this in the proxy or scripts. This would be a breaking change  that we need to announce etc.

hassan254-prog: I agree, we definitely need to handle this carefully to avoid disrupting users who rely on the proxy or scripts. Maybe we can check on the connections currently running in production environment.

TBonnin: there is only 12 existing guru integrations (including 4 in a prod environment) so it will definitely be a breaking change

bodinsamuel: We could have a second provider, not ideal though
We could introduce a `patch` props: 
```yaml
guru:
  proxy:
      base_url: https://api.getguru.com/api/v1
  patch
    20241022:
      base_url: https://api.getguru.com/
```

or the opposite to optimize for new integrations

```yaml
guru:
  proxy:
      base_url: https://api.getguru.com
  patch
    20241022:
      base_url: https://api.getguru.com/api/v1
```

So in the code we could use patch depending on the `created_at`

TBonnin:
@hassan254-prog is this PR still relevant? thx

hassan254-prog:
> @hassan254-prog is this PR still relevant? thx

The customer hasn't responded yet; perhaps we should hold off on this for a bit.

khaliqgant:
Should we just close this @hassan254-prog ?

hassan254-prog:
@khaliqgant yes.

PR #2838, opened 2024-10-13
fix(connections): update last fetched at to be 33%

khaliqgant:
## Describe your changes
Last fetched at is used to to check for token refresh. Reduce the setting of this value to 33% of the time. This will mean that more connections will be refreshed in the cron but since this call is made every time a connection is retrieved even if it isn't refreshed is a little excessive. Every time the credentials are actually refreshed then update last fetched at.


bodinsamuel (approved):
Lets try, thanks for pushing this

PR #2840, opened 2024-10-14
feat(integrations): platform changes to add JWT as an auth_mode

hassan254-prog:
## Describe your changes

- Platform changes to accommodate support for Ghost Admin API support


TBonnin:
what's the difference between Ghost admin auth and a classic API key?

hassan254-prog:
> what's the difference between Ghost admin auth and a classic API key?

In Ghost, there are two main APIs: the [Content](https://ghost.org/docs/content-api/#token-authentication) API and the [Admin](https://ghost.org/docs/admin-api/#token-authentication) API, each with different methods of authentication. The Content API is designed for read-only access to public content, and therefore uses a less secure method of authentication. The Admin API, on the other hand, provides full read and write access to all of Ghost's content, settings, and data.

TBonnin:
can we make it a `JwtCredentials` instead of something fully specific to Ghost Admin? I don't see how ghost admin auth differs from any JWT based auth flow

khaliqgant:
> can we make it a `JwtCredentials` instead of something fully specific to Ghost Admin? I don't see how ghost admin auth differs from any JWT based auth flow

@hassan254-prog csn you attend to this first thing tomorrow?

hassan254-prog:
Yes, I will make the necessary changes.

khaliqgant (approved):

hassan254-prog:
It might be hard to make this less specific to Ghost Admin, as the `apiKey.split(':')` method relies on the assumption that the API key consists of two parts (ID and secret), which is a specific characteristic of Ghost Admin.

TBonnin (approved):
It looks like the only logic specific to ghost is the split of the api key. The rest seems to be generic jwt logic. That's a lot of code then that cannot be reused, especially since I won't be surprise that we will have to support another jwt based auth soon. Anyway I won't be a blocker.

hassan254-prog:
@TBonnin, I just updated the PR to accommodate this. I was initially skeptical but it seems we have to go this route 😅

The PR now adds support for JWT, and this will provide a solid foundation as we continue to build on it. However, we currently cannot test the generated token with a test hook, as the authentication header must be formatted as `Authorization: Ghost ${token}` for Ghost. Tested the token generated and they were successfully authenticating Ghost Admin APIs. Waiting to have this reviewed so that I can update the [followup](https://github.com/NangoHQ/nango/pull/2841)

```
ghost-admin:
    display_name: Ghost Admin
    categories:
        - dev-tools
        - design
        - cms
    auth_mode: JWT
    token_headers:
        algorithm: HS256
    token_payload:
        expires_in_ms:  300000
        token_audience: /admin/
    proxy:
        headers:
            Accept: application/json
            Accept-Version: ${connectionConfig.version}
        base_url: https://${connectionConfig.adminDomain}/ghost/api/admin/
    docs: https://docs.nango.dev/integrations/all/ghost
    connection_config:
        adminDomain:
            type: string
            title: Ghost Admin Domain
            description: The domain of your Ghost Admin which can be different from your main domain
            example: 'mock.ghost.io'
            pattern: '^([a-z0-9_-]+\.|)(ghost\.io|[a-z0-9_-]+\.[a-z]{2,})$'
        version:
            type: string
            title: API Version
            description: The version of the Ghost Admin API to use
            example: 'v3.0'
            pattern: '^v\d+\.\d+$'

**Comment by TBonnin** [comment]
> However, we currently cannot test the generated token with a test hook, as the authentication header must be formatted as `Authorization: Ghost ${token}` for Ghost. 

can we have the proxy headers authorization defined in providers.yaml like we do for other mode to specify how the token is being passed:
```
    proxy:
        headers:
            Authorization: Ghost ${token}
```

**Comment by TBonnin** [comment]
```
token_headers:
        algorithm: HS256
token_payload:
        expires_in_ms:  300000
        token_audience: /admin/
```

Can we nest and organize the definition like this:
```
token:
        expires_in_ms: 300000
        headers:
                alg: HS256
        payload:
                aud: /admin/
```
Expiration applies to the token not just the payload. I would also use the short attribute name like `alg` and `aud` since there are part of the protocol https://datatracker.ietf.org/doc/html/rfc7519#section-4.1. Like that they can just be inserted as is when building the token without having to do any mapping

**Comment by hassan254-prog** [comment]
> ```
> token_headers:
>         algorithm: HS256
> token_payload:
>         expires_in_ms:  300000
>         token_audience: /admin/
> ```
> 
> Can we nest and organize the definition like this:
> 
> ```
> token:
>         expires_in_ms: 300000
>         headers:
>                 alg: HS256
>         payload:
>                 aud: /admin/
> ```
> 
> Expiration applies to the token not just the payload. I would also use the short attribute name like `alg` and `aud` since there are part of the protocol https://datatracker.ietf.org/doc/html/rfc7519#section-4.1. Like that they can just be inserted as is when building the token without having to do any mapping

Yes, it makes sense to organize and rename this to align it with the protocol defined in RFC 7519. I have updated the PR with the latest commit. I have also added a connection test hook so that we can test other providers using the JWT `auth_mode` as well.

**Comment by hassan254-prog** [comment]
> > However, we currently cannot test the generated token with a test hook, as the authentication header must be formatted as `Authorization: Ghost ${token}` for Ghost.
> 
> can we have the proxy headers authorization defined in providers.yaml like we do for other mode to specify how the token is being passed:
> 
> ```
>     proxy:
>         headers:
>             Authorization: Ghost ${token}
> ```

Quick question: won't this make it specific to Ghost? After token generation, shouldn't the token be passed in the Authorization header as `Bearer ${token}` for authentication? I have this logic implemented in `proxyService`.

File: packages/frontend/lib/index.ts [hunk 1]
```diff
@@ -415,6 +443,13 @@ export default class Nango {
             });
         }
 
+        if ('privateKeyId' in credentials || 'issuerId' in credentials || 'privateKey' in credentials || 'api_key' in credentials) {
+            return await this.triggerAuth({
+                authUrl: this.hostBaseUrl + `/auth/jwt/${providerConfigKey}${this.toQueryString(connectionId, connectionConfig as ConnectionConfig)}`,
+                credentials: credentials as BasicApiCredentials
```
TBonnin: BasicApiCredentials?

packages/shared/lib/sdk/sync.ts
```diff
@@ -248,6 +249,15 @@ interface TableauCredentials extends CredentialsCommon {
     token?: string;
     expires_at?: Date | undefined;
 }
+interface JWTCredentials {
```
TBonnin: we have name `TbaCredentials` so if we follow the same logic it should be `JwtCredentials`. Doesn't look as good but at least it is consistent

packages/shared/lib/services/connection.service.ts
```diff
@@ -253,6 +265,68 @@ class ConnectionService {
...
+                    connection_id: connectionId,
+                    provider_config_key: providerConfigKey,
+                    config_id: config.id as number,
+                    credentials,
+                    metadata: metadata || null,
+                    connection_config: connectionConfig || {},
+                    environment_id: environment.id
+                })
+            )
+            .returning('*');
```
TBonnin, line 305: can we do a upsert with `insert.onConflict` instead?

packages/shared/lib/services/connection.service.ts
```diff
@@ -1451,6 +1544,67 @@ class ConnectionService {
...
+    ): ServiceResponse<JWTCredentials> {
+        if (apiKey) {
+            const apiKeyParts = apiKey.split(':');
+            if (apiKeyParts.length === 2 && apiKeyParts[0] && apiKeyParts[1]) {
+                privateKeyId = apiKeyParts[0];
+                privateKey = apiKeyParts[1];
+            } else {
+                throw new NangoError(`invalid_api_key_format`);
+            }
+        }
```
TBonnin, line 1545: you can do 
```
if (apiKey) {
  const [privateKeyId, privateKey] = apiKey.split(':');
  
  if (!privateKeyId || !privateKey) {
    throw new NangoError('invalid_api_key_format');
  }
}
```

packages/shared/lib/services/connection.service.ts
```diff
@@ -1451,6 +1544,67 @@ class ConnectionService {
...
+
+        if (!privateKey) {
+            throw new NangoError(`invalid_jwt_private_key`);
+        }
+
+        const now = Math.floor(Date.now() / 1000);
+        const payload = {
+            iat: now,
+            exp: now + provider.token.expires_in_ms / 1000,
+            aud: provider.token.payload.aud
```
TBonnin: can we do?
```
const payload = {
            ...provider.token.payload.
            iat: now,
            exp: now + provider.token.expires_in_ms / 1000
```
so any values defined in providers.yaml will be set automatically

packages/shared/lib/services/connection.service.ts
```diff
@@ -1451,6 +1544,67 @@ class ConnectionService {
...
+        };
+
+        const header = {
+            alg: provider.token.headers.alg,
+            kid: privateKeyId
+        };
```
TBonnin, line 1564: same here.

packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
```diff
@@ -202,11 +203,20 @@ interface TableauCredentials extends CredentialsCommon {
     token?: string;
     expires_at?: Date | undefined;
 }
+interface JWTCredentials {
+    type: AuthModes['Jwt'];
+    api_key?: string;
```
TBonnin: can we add a comment to make it clear that `api_key` is to support provider like Ghost that use a composite api_key to pass the id and the privateKeyId and privateKey?

Maybe we can even make the type more explicit to make it clear it is either `api_key` or `privateKey/privateKeyId`. Ex:
```
interface JWTCredentials {
    type: AuthModes['Jwt'];
    privateKey: {
       id: string,
       secret: string
    } | string, // colon separated string: 'id:secret'
    issuerId?: string;
    token?: string;
    expires_at?: Date | undefined;
``` 
We would also need to make the `/auth/jwt` endpoint match this type if possible

packages/shared/lib/services/connection.service.ts
TBonnin: we should also document that when specified we expect this attribute to be the id/secret separated by a colon character

TBonnin (commented):
thank you very much for your hard work @hassan254-prog 
Made a few comments but it looks very good

packages/frontend/lib/index.ts
hassan254-prog: I have updated this with the latest [commit](https://github.com/NangoHQ/nango/pull/2840/commits/6c49f1085afcc716d7d9622373faace0827f0e78)

packages/shared/lib/sdk/sync.ts
hassan254-prog: Yes, makes sense to make it uniform, thanks , updated with the latest [commit](https://github.com/NangoHQ/nango/pull/2840/commits/6c49f1085afcc716d7d9622373faace0827f0e78)

packages/shared/lib/services/connection.service.ts
hassan254-prog: updated with the latest commit

packages/shared/lib/services/connection.service.ts
hassan254-prog: Yes, we should have this documented in the providers docs, but now that `apiKey` is not part of `JwtCredentials`, I have updated the logic to split the `privateKey` at the webapp, that way we can have 

```
interface JWTCredentials {
    type: AuthModes['Jwt'];
    privateKey: {
       id: string,
       secret: string
    } | string, // colon separated string: 'id:secret'
    issuerId?: string;
    token?: string;
    expires_at?: Date | undefined;

packages/shared/lib/services/connection.service.ts
hassan254-prog: yes, updated this with the latest commit.

packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
hassan254-prog: Yes, I have updated this, making `privateKey` required since it can be of `string` or an object with `id` and `secret`. Please have a look and let me know.

khaliqgant (changes requested):
I see a few issues with this:

1. The JWT auth mode seems to basically be the same as app store
2. The name getJwtCredentials is poorly named as there is already a `getJWTCredentials` method in the `connection.service`

khaliqgant:
> I see a few issues with this:
> 
> 1. The JWT auth mode seems to basically be the same as app store
> 2. The name getJwtCredentials is poorly named as there is already a `getJWTCredentials` method in the `connection.service`

We'll need a follow up to this to consolidate app store and jwt credentials. It doesn't seem there have been any app store connections added in production so it still isn't confirmed that the integration type works itself.

PR #2847, opened 2024-10-15
feat(integration): add support for Apollo Oauth2

hassan254-prog:
## Describe your changes

- Add support for Apollo Oauth2
- Made platform changes to accommodate Apollo's `authorization_url_fragment`, ensuring that it is correctly appended to the `authorization_url` during authorization.

[EXT-187](https://linear.app/nango/issue/EXT-187/add-support-for-apollo-oauth2)


docs-v2/spec.yaml
```diff
@@ -1876,6 +1876,8 @@ components:
           type: object
         authorization_url_replacements:
           type: object
+        authorization_url_fragment:
+            type: string
```
TBonnin: indentation looks off

packages/shared/providers.yaml
```diff
@@ -229,6 +229,26 @@ apollo:
...
             description: The API key for your Apollo account
 
+apollo-oauth:
+    display_name: Apollo (OAuth)
+    categories:
+        - marketing
+    auth_mode: OAUTH2
+    authorization_url: https://app.apollo.io
+    token_url: https://app.apollo.io/api/v1/oauth/token
+    authorization_url_fragment: oauth/authorize
```
TBonnin, line 274: not sure I understand the role of `authorization_url_fragment`. why is `the authorization_url` not https://app.apollo.io/oauth/authorize?

docs-v2/spec.yaml
hassan254-prog: Thanks, fixed with [commit](https://github.com/NangoHQ/nango/pull/2847/commits/5ab422b632785b46508dfe3ac6b56c9a21fc184e)

packages/shared/providers.yaml
hassan254-prog: I attempted to access the URL `https://app.apollo.io/oauth/authorize`, but I encountered a 404 error. This issue likely arises because the server does not recognize the `/oauth/authorize` path when it is not prefixed by a fragment identifier. In this case, the fragment identifier (#) is essential for directing the application to the correct route.

image

TBonnin: cannot `authorization_url` be  `https://app.apollo.io/#/oauth/authorize`?

hassan254-prog: Fragments are not sent to the server. In HTTP requests, anything after the `#` in a URL is treated as a fragment identifier and is not transmitted as part of the request to the server. Only the part before the `#` is sent. Therefore, the URL will always be `https://app.apollo.io`.

TBonnin: I am not sure I follow. The fragment is concatenated in oauth.controller.ts (changes above). Why do we need to concatenate in the logic instead of being part of the authoriziation_url?

hassan254-prog: I see the confusion. We concatenate the fragment `#/oauth/authorize` in the logic because fragments are only used by the client and are never transmitted to the server as part of an HTTP request. When configuring the `authorizationUrl`, we use the `origin (host)` and `pathname (path)` separately rather than the full href. This is because the current library `simple-oauth2` [expects](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/93a782948d4ca1b68e686e01fe6d9e819aec2423/types/simple-oauth2/v1/index.d.ts#L6) these two components, not the full href. The fragment is therefore omitted in the pathname, which is why I made the above change. 
image

hassan254-prog: We concatenate the fragment `#/oauth/authorize` in the logic because fragments are only used by the client and are never transmitted to the server as part of an HTTP request. When configuring the `authorizationUrl`, we use the origin (host) and pathname (path) separately rather than the full href. This is because the current library (simple-oauth2) expects these two components, not the full href. The fragment is therefore omitted in the pathname, which explains the reasoning behind this approach. Please let me know if I am missing anything here.
image

TBonnin: got it. It is a shame to have to support fragments even though it is explicitly stated in the oauth2 framework that authorize url should not have fragments: https://www.rfc-editor.org/rfc/rfc6749#section-3.1

TBonnin (approved):

packages/shared/providers.yaml
hassan254-prog: Yes, I think as we add support for more of these, we are more likely to encounter this kind of situation

PR #2850, opened 2024-10-16
fix(node-client): [nan-1915] pass the body along for a delete request if provided

khaliqgant:
## Describe your changes
Follow up of https://github.com/NangoHQ/nango/pull/2229/ -- the node client should also allow `DELETE` to have a body. 

**Note: This is untested as of now**

NAN-1915

### DIFF
--- a/packages/node-client/lib/index.ts
+++ b/packages/node-client/lib/index.ts
@@ -811,7 +811,7 @@ export class Nango {
         } else if (method?.toUpperCase() === 'PUT') {
             return this.http.put(url, config.data, options);
         } else if (method?.toUpperCase() === 'DELETE') {
-            return this.http.delete(url, options);
+            return this.http.delete(url, { ...options, data: config.data || {} });
         } else {
             return this.http.get(url, options);
         }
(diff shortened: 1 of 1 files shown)


packages/node-client/lib/index.ts
```diff
@@ -811,7 +811,7 @@ export class Nango {
         } else if (method?.toUpperCase() === 'PUT') {
             return this.http.put(url, config.data, options);
         } else if (method?.toUpperCase() === 'DELETE') {
-            return this.http.delete(url, options);
+            return this.http.delete(url, { ...options, data: config.data || {} });
```
TBonnin:
```
{ 
...options, 
...(config.data ? { data: config.data } : {})
}
```
so no data is set if no config.data, instead of passing an empty object

TBonnin (approved):

packages/node-client/lib/index.ts
khaliqgant: Yup, thanks. Will adjust, and test tomorrow

PR #2851, opened 2024-10-16
feat: expose auth or sync errors when listing connections

TBonnin:
## Describe your changes

Expose connection errors when listing connections
Errors can be of type `sync` when a sync has failed or `auth` if credentials is not valid (only oauth for now, I have a ticket to check basic and api key auth mode).
Eventually we can add a 'reason` attribute for each error but right now fetching a specific log from ES is not supported

Ex: 
```
{
"connections": [
  {
    "id": 1,
    "connection_id": "test-1",
    "provider": "slack",
    "provider_config_key": "slack-nango-community",
    "created": "2023-06-03T14:53:22.051Z",
    "metadata": null,
    "errors": [{ "type": "auth", "log_id": "VrnbtykXJFckCm3HP93t"}]
  },
  {
    "id": 2,
    "connection_id": "test-2",
    "provider": "slack",
    "provider_config_key": "slack-nango-community",
    "created": "2023-06-03T15:00:14.945Z",
    "metadata": {
      "bot_id": "some-uuid"
    },
    "errors": []
  }
]
}

```

### DIFF
--- a/docs-v2/spec.yaml
+++ b/docs-v2/spec.yaml
@@ -437,6 +437,21 @@ paths:
                                                 metadata:
                                                     type: object
                                                     description: Custom metadata attached to the connection
+                                                errors:
+                                                    type: array
+                                                    items:
+                                                        type: object
+                                                        properties:
+                                                            type:
+                                                                type: string
+                                                                example: "'auth' or 'sync'"
+                                                            log_id:
+                                                                type: string
+                                                                example: "VrnbtykXJFckCm3HP93t"
+                                                    description: |
+                                                        List of connection errors. Ex:
+                                                        - Connection credentials are invalid (type=auth)
+                                                        - Last sync for the connnection has failed (type=sync)
 
         post:
             description: Adds a connection for which you already have credentials.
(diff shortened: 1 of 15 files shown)


docs-v2/spec.yaml
```diff
@@ -437,6 +437,21 @@ paths:
...
+                                                        properties:
+                                                            type:
+                                                                type: string
+                                                                example: "'auth' or 'sync'"
+                                                            log_id:
+                                                                type: string
+                                                                example: "VrnbtykXJFckCm3HP93t"
+                                                    description: |
+                                                        List of connection errors. Ex:
+                                                        - Connection credentials are invalid (type=auth)
```
khaliqgant, line 453: is this description actually correct since we only give back a log id?

khaliqgant (approved):
Question about documentation inline

docs-v2/spec.yaml
TBonnin: I didn't document the log_id field since it can only be used with our UI.
This PR is a small step in the direction of exposing connection level errors so customers know something is wrong for a given connection.  Eventually we might want to add a `reason` but for now it is technically complicated to get the details of one single log from ES

PR #2852, opened 2024-10-17
feat(integrations): add support for bill

hassan254-prog:
## Describe your changes

- Add support for Bill(Connect API) and Bill-sandbox(Connect API) APIs. The flow is a little bit custom as it requires 4 parameters(`Username`, `Password`, `OrganizationId` and `DevKey`) to generate a `sessionId` which are both used together with the `DevKey` in the headers for authentication. The `sessionId` has an expiry time of 35 minutes after inactivity, since it is hard for us to know if the key was used within the 35 minutes period, I have added the logic to always refresh the `sessionId` after every 35 minutes that way we are ensured of a fresh  `sessionId`.

[EXT-194](https://linear.app/nango/issue/EXT-194/support-billcom-integration)


packages/server/lib/controllers/auth/postBill.ts
```diff
@@ -0,0 +1,209 @@
...
+
+    const queryStringVal = queryStringValidation.safeParse(req.query);
+    if (!queryStringVal.success) {
+        res.status(400).send({
+            error: { code: 'invalid_query_params', errors: zodErrorToHTTP(queryStringVal.error) }
+        });
+        return;
+    }
+
+    const paramVal = paramValidation.safeParse(req.params);
```
TBonnin: `paramsVal` and `paramsValidation` with a `s`, no?

packages/server/lib/controllers/auth/postBill.ts
```diff
@@ -0,0 +1,209 @@
...
+            res.status(400).send({ error: { code: 'invalid_auth_mode' } });
+            return;
+        }
+
+        await logCtx.enrichOperation({ integrationId: config.id!, integrationName: config.unique_key, providerName: config.provider });
+
+        const { success, error, response: credentials } = await connectionService.getBillCredentials(provider, userName, password, organizationId, devkey);
+
+        console.log(success);
+        console.log(credentials);
```
TBonnin: console.log 🪓

packages/shared/lib/models/Proxy.ts
```diff
@@ -45,6 +45,8 @@ export interface ApplicationConstructedProxyConfiguration extends BaseProxyConfi
     token: string | BasicApiCredentials | ApiKeyCredentials | AppCredentials | TbaCredentials | TableauCredentials;
     provider: Provider;
     connection: Connection;
+    sessionId?: string;
+    devKey?: string;
```
TBonnin: do we really need to add those 2 attributes which are specific to a single provider?

packages/shared/lib/services/connection.service.ts
```diff
@@ -327,6 +329,68 @@ class ConnectionService {
...
+                    connection_id: connectionId,
+                    provider_config_key: providerConfigKey,
+                    config_id: config.id as number,
+                    credentials,
+                    metadata: metadata || null,
+                    connection_config: connectionConfig || {},
+                    environment_id: environment.id
+                })
+            )
+            .returning('*');
```
TBonnin, line 414: can we not do an upsert with `insert().onConflict()`?

packages/server/lib/controllers/auth/postBill.ts
hassan254-prog: 🤦‍♂️, updated  this with the latest [commit](https://github.com/NangoHQ/nango/pull/2852/commits/2aeacff57d36fa34b3461b7995d1291ec5c1dbd8), thanks.

packages/server/lib/controllers/auth/postBill.ts
hassan254-prog: Yes it should be `paramsVal` and `paramsValidation` , updated this.

packages/shared/lib/services/connection.service.ts
hassan254-prog: Yes this is possible, since we have constraints on the fields `provider_config_key`, `connection_id`, and `environment_id` in the connections table. I think we should also refactor the other `upsert` functions. Updated this with the latest [commit](https://github.com/NangoHQ/nango/pull/2852/commits/054687f6796c3104523a332db412cb0e98c9382c).

packages/shared/lib/models/Proxy.ts
hassan254-prog: I think we can omit this and handle authentication at the script level. Just a thought—since we've implemented a new `auth_mode` for this, would it be wise to do the same for the proxy, or should we authenticate at the script level? What do you think?

TBonnin: I would try to keep the ApplicationConstructedProxyConfiguration as generic as possible. sessionId and devKey are already part of of the `BillCredentials`. Can we get them from there?

hassan254-prog: Yes, we can get it and use it when working with integrations. Thank you for clarifying this. Updating PR now.

hassan254-prog: I have reverted this with the latest commits

TBonnin (approved):

khaliqgant (approved):

khaliqgant:
@hassan254-prog can you resolve conflicts here?

hassan254-prog:
> @hassan254-prog can you resolve conflicts here?

resolving now 👍.

packages/webapp/src/pages/Connection/Create.tsx
```diff
@@ -426,6 +436,20 @@ export default function IntegrationCreate() {
             }
         }
 
+        const billCredentialsString = '';
+        if (integration?.authMode === 'BILL') {
+            if (apiAuthUsername && apiAuthPassword && organizationId && devKey) {
+                tableauCredentialsString = `
```
bodinsamuel: wrong var

bodinsamuel (approved):
Such an error to have standard and non-standard auth mixed, this list is going to grow forever 😢

packages/webapp/src/pages/Connection/Create.tsx
hassan254-prog: 🤦‍♂️,updated with the latest commit

PR #2855, opened 2024-10-17
feat: expose connection error in getConnection response

TBonnin:
## Describe your changes

Follow up of https://github.com/NangoHQ/nango/pull/2851
Adding an `errors` attribute to the connection endpoint to expose sync or auth failure for a given connection


khaliqgant (approved):

TBonnin:
After discussing with @bastienbeurier we agreed not to add the errors to the single connection endpoint, which is the most used endpoint by far. We should be very careful adding any extra db query in this endpoint. I'll close the PR for now. We can reopen it if we change our mind later

PR #2856, opened 2024-10-17
feat(post-connection-scripts): [nan-1901] add post connection scripts

khaliqgant:
## Describe your changes
Adds post connection scripts for gusto, docusign, calendly

NAN-1901

### DIFF
--- a/packages/server/lib/hooks/connection/calendly-post-connection.ts
+++ b/packages/server/lib/hooks/connection/calendly-post-connection.ts
@@ -1,0 +1,25 @@
+import type { InternalNango as Nango } from './post-connection.js';
+import type { CalendlyUser } from '../response-types/calendly.js';
+import { isAxiosError } from 'axios';
+
+export default async function execute(nango: Nango) {
+    const connection = await nango.getConnection();
+
+    const response = await nango.proxy<CalendlyUser>({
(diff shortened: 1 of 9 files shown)


packages/server/lib/hooks/connection/calendly-post-connection.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    if (isAxiosError(response)) {
+        return;
+    }
+
+    const { data } = response;
+
+    const { current_organization } = data.resource;
+
+    const organizationId = current_organization.split('/').pop();
```
TBonnin, line 22: should we be more defensive in case the response is not what we expect? same question for all the providers below

TBonnin (approved):

packages/server/lib/hooks/connection/calendly-post-connection.ts
khaliqgant: Thanks, perhaps, but all the post connection scripts are executed within a try, catch https://github.com/NangoHQ/nango/blob/master/packages/server/lib/hooks/connection/post-connection.ts#L109. Will merge for now and we can revisit if it causes problems. 

The most important thing is that if the post connection script fails it won't block creating the connection.

PR #2857, opened 2024-10-17
fix(response-saver): concat paginated responses

khaliqgant:
## Describe your changes
If a response is paginated and the second page is empty it overwrites the file. As a quick fix concat the result into one file.


packages/cli/lib/services/response-saver.service.ts
```diff
@@ -10,12 +10,29 @@ export function ensureDirectoryExists(directoryName: string): void {
...
+    customFilePath: string;
+    concatenateIfExists: boolean;
+}): void {
     ensureDirectoryExists(`${directoryName}/mocks`);
 
     const filePath = path.join(directoryName, customFilePath);
     ensureDirectoryExists(path.dirname(filePath));
 
+    if (fs.existsSync(filePath)) {
+        const existingData = JSON.parse(fs.readFileSync(filePath, 'utf8'));
```
TBonnin, line 30: hopefully we won't have too big saved response because loading/concat/rewrite could become quite slow

TBonnin (approved):

bodinsamuel (approved):

PR #2858, opened 2024-10-18
fix: lock can be released even when not holding the lock

TBonnin:
## Describe your changes

I wanted to use the redis based locking mechanism to refactor the refresh tokens logic but I found out a pretty big flow where a lock can be released by a process that doesn't hold it. That's a pretty bad design flow but I am not sure of magnitude of the impact since we are using it only to prevent concurrent token refresh and the issue would only materialize in the very specific scenario where the lock holding process is still trying to refresh the token after the lock timeout (10s)

Anyway this fix ensures that to release it, a lock must first be acquired. 
I have also moved the instantiation of the kvstore so it can be reused outside of the connection service. (I am gonna use it in the refreshTokens cron job)

### DIFF
--- a/packages/shared/lib/index.ts
+++ b/packages/shared/lib/index.ts
@@ -33,6 +33,7 @@ export * from './services/invitations.js';
 export * from './services/providers.js';
 
 export * as oauth2Client from './clients/oauth2.client.js';
+export * from './clients/locking.js';
 
 export * from './services/nango-config.service.js';
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -36495,36 +36495,10 @@
                 "acorn": "^8"
             }
         },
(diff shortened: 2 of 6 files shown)


packages/shared/lib/index.ts
```diff
@@ -33,6 +33,7 @@ export * from './services/invitations.js';
 export * from './services/providers.js';
 
 export * as oauth2Client from './clients/oauth2.client.js';
+export * from './clients/locking.js';
```
bodinsamuel, line 36: I should have merged my PR before #2789 😭

bodinsamuel (approved):

packages/shared/lib/index.ts
TBonnin: ah yes sorry about that

PR #2859, opened 2024-10-18
refactor: refreshTokens cron to use redis based locking 

TBonnin:
## Describe your changes
- using the redis-backed locking mechanism to prevent multiple instances to run the refreshTokens jobs. The current locking mechanism (advisory lock in pg) has been causing deadlock and outage lately. Lock now has a TTL.

I have also modified the logic to paginate through all stale connections (instead of 1000 of them) to prevent any scenario where the same 1000 "stucked" stale connections could be returned by the queries. I think we currently assume last_fetched_at is always updated so it is working but iterating over all the stales connection make it more resilient to errors or changes in `getConnectionCredentials`

depends on https://github.com/NangoHQ/nango/pull/2858


packages/shared/lib/services/connection.service.ts
```diff
@@ -931,8 +931,8 @@ class ConnectionService {
                     });
                 }
 
-                // TODO: this leak credentials to the logs
-                const errorWithPayload = new NangoError(error!.type, connection);
+                const { credentials, ...connectionWithoutCredentials } = connection;
+                const errorWithPayload = new NangoError(error!.type, connectionWithoutCredentials);
```
TBonnin, line 1058: unrelated but it was quick to fix

bodinsamuel:
I haven't looked I'm sure it's correct but we can't use redis for this cron since it's needed for self-hosted (happy to enforce Redis though)

packages/shared/lib/services/connection.service.ts
```diff
@@ -945,18 +945,12 @@ class ConnectionService {
...
-                await this.updateLastFetched(connection.id);
             }
 
             connection.credentials = response.credentials as OAuth2Credentials;
         }
 
-        // sample this to reduce writes and load on the db
-        if (Math.random() < 0.33) {
-            await this.updateLastFetched(connection.id);
-        }
```
TBonnin, line 1073: we were already doing it above.

TBonnin:
> I haven't looked I'm sure it's correct but we can't use redis for this cron since it's needed for self-hosted (happy to enforce Redis though)

Took a shortcut when I said Redis. :) In reality the lock is backed by a KVStore which is the `InMemoryKVStore` if Redis is not setup. https://github.com/NangoHQ/nango/pull/2858/files#diff-01d4e5dfd0fed0e438c2ded2cf6bf3168dedbff096b802507afdb53c7256d68eR14

The Redis requirement stays the same: it is only mandatory if running multi-instances server

packages/shared/lib/services/connection.service.ts
bodinsamuel: it was a tmp fix from khaliq, I think we can revert anyway as it was not directly the issue

bodinsamuel (approved):
Should have read before :D

PR #2861, opened 2024-10-21
fix(cli): add credentials to save-responses

hassan254-prog:
## Describe your changes
- Add credentials to `--save-responses`

### DIFF
--- a/packages/cli/lib/services/response-saver.service.ts
+++ b/packages/cli/lib/services/response-saver.service.ts
@@ -57,9 +57,9 @@ export function onAxiosRequestFulfilled({
         const connection = response.data as Connection;
 
         // getConnection could be getMetadata as well which would be cached
-        saveResponse<Pick<Connection, 'metadata' | 'connection_config'>>({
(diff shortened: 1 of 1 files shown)


packages/cli/lib/services/response-saver.service.ts
```diff
@@ -57,9 +57,9 @@ export function onAxiosRequestFulfilled({
         const connection = response.data as Connection;
 
         // getConnection could be getMetadata as well which would be cached
-        saveResponse<Pick<Connection, 'metadata' | 'connection_config'>>({
+        saveResponse<Pick<Connection, 'metadata' | 'connection_config' | 'credentials'>>({
             directoryName,
-            data: { metadata: connection.metadata as Metadata, connection_config: connection.connection_config },
+            data: { metadata: connection.metadata as Metadata, connection_config: connection.connection_config, credentials: connection.credentials },
```
khaliqgant, line 62: Don’t think we want this as to not leak credentials

bodinsamuel (commented):
Seems like a footgun no?

hassan254-prog:
I think for this, we will have to manually add fake mocks.

PR #2862, opened 2024-10-21
feat: basic/apikey credentials check cron job

TBonnin:
## Describe your changes
I decided to piggy-back on the refreshTokens cron job (renamed refreshConnections even though I am not sure with the naming. Recommendations are welcomed) to benefit from logging of errors, etc... and not have another dedicated background jobs that is very similar.

A failed check shows in the logs like this, the same way a token refresh error cc @bastienbeurier :
Screenshot 2024-10-21 at 08 53 34

https://linear.app/nango/issue/NAN-1404/periodically-test-for-invalid-api-keys-and-basic-creds

### DIFF
--- a/packages/server/lib/refreshConnections.ts
+++ b/packages/server/lib/refreshConnections.ts
@@ -1,38 +1,44 @@
 import * as cron from 'node-cron';
-import type { Lock } from '@nangohq/shared';
+import type { Lock, NangoError } from '@nangohq/shared';
 import { errorManager, ErrorSourceEnum, connectionService, locking } from '@nangohq/shared';
+import type { Result } from '@nangohq/utils';
 import { stringifyError, getLogger, metrics } from '@nangohq/utils';
 import { logContextGetter } from '@nangohq/logs';
-import { connectionRefreshFailed as connectionRefreshFailedHook, connectionRefreshSuccess as connectionRefreshSuccessHook } from './hooks/hooks.js';
+import {
+    connectionRefreshFailed as connectionRefreshFailedHook,
+    connectionRefreshSuccess as connectionRefreshSuccessHook,
+    connectionTest
+} from './hooks/hooks.js';
 import tracer from 'dd-trace';
+import type { ApiKeyCredentials, BasicApiCredentials, ConnectionConfig, Provider, TbaCredentials } from '@nangohq/types';
 
 const logger = getLogger('Server');
-const cronName = '[refreshTokens]';
-const cronMinutes = 10;
+const cronName = '[refreshConnections]';
+const cronMinutes = 1;
 
-export function refreshTokens(): void {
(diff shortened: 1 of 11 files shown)


packages/server/lib/refreshConnections.ts
```diff
@@ -99,3 +106,15 @@ export async function exec(): Promise<void> {
...
+    providerName: string,
+    provider: Provider,
+    credentials: ApiKeyCredentials | BasicApiCredentials | TbaCredentials,
+    connectionId: string,
+    providerConfigKey: string,
+    environment_id: number,
+    connection_config: ConnectionConfig
+): Promise<Result<boolean, NangoError>> {
+    return connectionTest(providerName, provider, credentials, connectionId, providerConfigKey, environment_id, connection_config, tracer);
+}
```
TBonnin: needed so I can pass the tracer to `shared`

bastienbeurier:
Nice!

packages/server/lib/refreshConnections.ts
bodinsamuel: I wonder if it's really useful, maybe tracer usage is not required and we can load dd-trace directly here

packages/server/lib/refreshConnections.ts
```diff
@@ -1,38 +1,44 @@
...
+    connectionTest
+} from './hooks/hooks.js';
 import tracer from 'dd-trace';
+import type { ApiKeyCredentials, BasicApiCredentials, ConnectionConfig, Provider, TbaCredentials } from '@nangohq/types';
 
 const logger = getLogger('Server');
-const cronName = '[refreshTokens]';
-const cronMinutes = 10;
+const cronName = '[refreshConnections]';
+const cronMinutes = 1;
```
bodinsamuel: on purpose?

bodinsamuel (approved):
We should probably store an ideal `next_refresh_at` in the DB instead of selecting all connections ever in this cron

packages/server/lib/refreshConnections.ts
TBonnin: good point. I though dd-trace was not a dependency of shard but it actually is.

packages/server/lib/refreshConnections.ts
TBonnin: oups no. that was for my testing

packages/server/lib/refreshConnections.ts
bodinsamuel: I'm not used to tracing so not sure if there is a global state or only one attached to the tracer object

TBonnin:
> We should probably store an ideal `next_refresh_at` in the DB instead of selecting all connections ever in this cron

we could also join with `active_logs` to not retry the connection we know are already failing

PR #2868, opened 2024-10-22
feat(integrations): add TwoStep as a new auth_mode for Perimeter 81

hassan254-prog:
## Describe your changes

- Add support for perimeter81

[EXT-195](https://linear.app/nango/issue/EXT-195/support-permiter81)

The flow is somewhat custom, where we send a single `api_key` and receive an `access_token`, which can later be used for authentication by passing the token in the headers like this:` Authorization: Bearer <ACCESS_TOKEN>`. I anticipate the list will continue to grow as we add support for more of these. I’m open to suggestions on how we can better handle these scenarios.

```
curl -X 'POST' -H 'Content-Type: application/json' \
-d '{
  "grantType": "api_key",
  "apiKey": "<API_KEY>"
}' 'https://api.perimeter81.com/api/v1/auth/authorize'


{
   "data":{
      "tokenType":"bearer",
      "accessToken":"<ACCESS_TOKEN>",
      "accessTokenExpire":<TOKEN_EXPIRE_TIME>
   }
}

### CODE REVIEW

**Review by bodinsamuel — COMMENTED** [review]
I feel like we could go with some `if` in the code instead of entirely dedicated auth strategy, but at the same time it keeps them well isolated. I don't know

**Comment by hassan254-prog** [comment]
I have run tests locally using a mock API server based on the provider's API documentation, and tests were successful.

File: packages/frontend/lib/index.ts [hunk 1]
```diff
@@ -377,6 +379,14 @@ export default class Nango {
             return { params: tableauCredentials } as unknown as ConnectionConfig;
         }
 
+        if ('api_key' in credentials) {
+            const perimeterCredentials: PerimeterCredentials = {
+                api_key: credentials.api_key
+            };
+
+            return { params: perimeterCredentials } as unknown as ConnectionConfig;
+        }
```
TBonnin: this works only because no other credentials has a attribute called api_key in snake case. It is very fragile and might be a source of bugs if more credentials with an api_key attribute are added. This PR is probably not the place to refactor but a good refactoring of this function would be welcome. For instance credentials could bevalidated with the help of zod schemas

TBonnin (commented):
it is yet another flavor of a 2-steps token based authentication. I am a little bit worried about adding a new credentials type every single time. I wonder if we could define a more generic `authMode`. We already have `token_url`, we would need a additional `token_url_headers` but do we need anything else in the providers.yaml

hassan254-prog:
> it is yet another flavor of a 2-steps token based authentication. I am a little bit worried about adding a new credentials type every single time. I wonder if we could define a more generic `authMode`. We already have `token_url`, we would need a additional `token_url_headers` but do we need anything else in the providers.yaml

We can also define the various fields i.e `api_key` for perimeter81, in the providers.

hassan254-prog:
@TBonnin, @bodinsamuel, 

I don't know if I have done too much here, but after working on this, I believe we can comfortably merge `BILL`, `TABLEAU`, and other future integrations with this kind of flavor. I will open another PR to refactor the above `convertCredentialsToConfig`. I will need help with renaming the `auth_mode`.  What do you think about the implementation?

This will now enable us to have such:
```yaml
foo:
    display_name: Foo
    categories:
        - productivity
    auth_mode: TWOSTEP
    proxy:
        base_url: https://api.perimeter81.com/api/rest
    token_url: https://api.${connectionConfig.domain}.com/api/v1/auth/authorize
    token_params: 
        credentials:
            apiKey: ${credential.apiKey}
            grantType: api_key
            site:
                content_url: ${credential.contentUrl}
    token_headers:
        Content-Type: application/json
    token_response:
        token: data.accessToken #Dynamically access the token 
        token_expiration: data.accessTokenExpire #Dynamically access the token expiration date 
        token_expiration_formula: expireAt | expireIn # Same for this we can dynamically pick the formula to calculate the expire at value
    token_expires_in_ms: 36000   # Help us define this if provider has only mentioned in docs but not included in raw cred data
   
   ```
   **Future works for this:**

- Dynamically add authentication in proxy headers for this, i.e:

```yaml
proxy:
    base_url: https://api.perimeter81.com/api/rest
    authentication_header: X-tableau-Auth ${token}
```
- Come up with a way of how we can dynamically add zod validation.

packages/shared/providers.yaml
```diff
@@ -3789,6 +3789,32 @@ perplexity:
...
+    token_url: https://api.${connectionConfig.domain}.com/api/v1/auth/authorize
+    token_params:
+        apiKey: ${credential.apiKey}
+        grantType: api_key
+    token_headers:
+        Content-Type: application/json
+    token_response:
+        token: data.accessToken
+        token_expiration: data.accessTokenExpire
+        token_expiration_formula: expireAt
```
TBonnin: would call that `token_expiration_strategy`

TBonnin (commented):
love it. Regarding the naming I feel it should be called `TBA` for token based authentication and the existing TBA should be renamed NetsuiteTBA because it is very much specific to Netsuite. What do you think?

docs-v2/reference/api-configuration.mdx
```diff
@@ -193,7 +193,7 @@ slack:
 </ResponseField>
 
 <ResponseField name="auth_mode" type="string" >
-    The authentication mode. Must be one of: "API_KEY", "APP", "APP_STORE", "BASIC", "NONE", "OAUTH1", "OAUTH2", "OAUTH2_CC", "CUSTOM", "TBA", "TABLEAU", "JWT", "BILL".
+    The authentication mode. Must be one of: "API_KEY", "APP", "APP_STORE", "BASIC", "NONE", "OAUTH1", "OAUTH2", "OAUTH2_CC", "CUSTOM", "TBA", "TABLEAU", "JWT", "BILL", "PERIMETER".
```
bodinsamuel: missed that

docs-v2/spec.yaml
```diff
@@ -1822,7 +1822,7 @@ components:
             type: string
         auth_mode:
           type: string
-          enum: ['OAUTH1','OAUTH2','OAUTH2_CC','BASIC','API_KEY','APP_STORE','CUSTOM','APP','NONE','TBA','TABLEAU','JWT','BILL']
+          enum: ['OAUTH1','OAUTH2','OAUTH2_CC','BASIC','API_KEY','APP_STORE','CUSTOM','APP','NONE','TBA','TABLEAU','JWT','BILL','PERIMETER']
```
bodinsamuel: same

packages/shared/lib/services/connection.service.ts
```diff
@@ -410,6 +421,50 @@ class ConnectionService {
         return [{ connection: connection!, operation: connection ? 'override' : 'creation' }];
     }
 
+    public async upsertPerimeterConnection({
```
bodinsamuel: upsetTwoStepConnection?

packages/shared/lib/services/connection.service.ts
```diff
@@ -410,6 +421,50 @@ class ConnectionService {
...
+            .from<StoredConnection>(`_nango_connections`)
+            .insert(encryptedConnection)
+            .onConflict(['connection_id', 'provider_config_key', 'environment_id', 'deleted_at'])
+            .merge({
+                ...encryptedConnection,
+                updated_at: new Date()
+            })
+            .returning('*');
+
+        void analytics.track(AnalyticsTypes.PERIMETER_CONNECTION_INSERTED, account.id, { provider: config.provider });
```
bodinsamuel: same comment

packages/server/lib/controllers/auth/postTwoStep.ts
```diff
@@ -0,0 +1,204 @@
...
+                    account,
+                    auth_mode: 'TWOSTEP',
+                    operation: updatedConnection.operation
+                },
+                config.provider,
+                logContextGetter,
+                undefined,
+                logCtx
+            );
+        }
```
bodinsamuel: Sorry to inflict you that, I just changed this block of code in all auth mode can you apply the same changes please? ☺️ 
https://github.com/NangoHQ/nango/pull/2884/files

hassan254-prog:
> love it. Regarding the naming I feel it should be called `TBA` for token based authentication and the existing TBA should be renamed NetsuiteTBA because it is very much specific to Netsuite. What do you think?

Yes I agree, that would be a better naming strategy. My worry is after renaming would it be a breaking change to the existing TBA connections? Maybe we can start by checking if there are any connections in prod environment.

bodinsamuel (commented):
Thanks for the grit on this, I haven't tested but looks correct. Last batch of comments 👌🏻

khaliqgant:
> > love it. Regarding the naming I feel it should be called `TBA` for token based authentication and the existing TBA should be renamed NetsuiteTBA because it is very much specific to Netsuite. What do you think?
> 
> Yes I agree, that would be a better naming strategy. My worry is after renaming would it be a breaking change to the existing TBA connections? Maybe we can start by checking if there are any connections in prod environment.

There are `TBA` integrations on prod, so let's hold off on renaming `TBA` for now.

docs-v2/reference/api-configuration.mdx
hassan254-prog: Thank you 🙏

packages/shared/lib/services/connection.service.ts
hassan254-prog: updated with latest commit.

packages/server/lib/controllers/auth/postTwoStep.ts
hassan254-prog: No worries at all—I initially missed this. Thank you! ☺️ I've updated it with the latest [commit](https://github.com/NangoHQ/nango/pull/2868/commits/05b56ec0584bccafc644ef6191387894f4a39512)

packages/server/lib/controllers/auth/postTwoStep.ts
```diff
@@ -0,0 +1,216 @@
...
+            await logCtx.failed();
+
+            errorManager.errRes(res, 'two_step_error');
+
+            return;
+        }
+
+        const connectionId = receivedConnectionId || connectionService.generateConnectionId();
+
+        const [updatedConnection] = await connectionService.upsetTwoStepConnection({
```
TBonnin: typo: `upsert` instead of `upset`

packages/shared/providers.yaml
TBonnin: @hassan254-prog can you take care of the rename?

docs-v2/reference/api-configuration.mdx
```diff
@@ -193,7 +193,7 @@ slack:
 </ResponseField>
 
 <ResponseField name="auth_mode" type="string" >
-    The authentication mode. Must be one of: "API_KEY", "APP", "APP_STORE", "BASIC", "NONE", "OAUTH1", "OAUTH2", "OAUTH2_CC", "CUSTOM", "TBA", "TABLEAU", "JWT", "BILL".
+    The authentication mode. Must be one of: "API_KEY", "APP", "APP_STORE", "BASIC", "NONE", "OAUTH1", "OAUTH2", "OAUTH2_CC", "CUSTOM", "TBA", "TABLEAU", "JWT", "BILL", "TWOSTEP".
```
TBonnin: are we set on the name `TWOSTEP`? @khaliqgant @bodinsamuel @hassan254-prog It is harder to change once used. if yes should it be `TWO_STEP`?

TBonnin (approved):

packages/server/lib/controllers/auth/postTwoStep.ts
hassan254-prog: updated with the latest commit, thank you 🙏!

docs-v2/reference/api-configuration.mdx
hassan254-prog: Since token-based authentication is a security process involving two steps; authentication and token generation, 
I suggest naming it `TWO_STEP`. Perhaps Khaliq and Samuel have other reservations?

khaliqgant: `TWO_STEP` is good for a name for me.

khaliqgant (approved):

bodinsamuel (approved):

PR #2870, opened 2024-10-22
fix(connection): getConnectionCredentials should return credentials even on error

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1933/fix-ui-crash-when-loading-a-connection-with-token-refresh-error

- Split getConnectionCredentials and initial connection fetch
For the /connections/:id endpoint we still want to return the connection even if the refresh failed but if there was an error we stripped the credentials to avoid leaking stuff. Initially, I had just reverted this and stripped the credentials outside of the function at multiple places but the code was even more crappier.
So I split the function, this allows us to not repeat call to fetch the integration on the API and correctly output the errors. It's a bit less straightforward but more granular

### DIFF
--- a/packages/server/lib/controllers/v1/connection/get.ts
+++ b/packages/server/lib/controllers/v1/connection/get.ts
@@ -106,7 +106,11 @@ export const getConnection = asyncWrapper<GetConnection>(async (req, res) => {
 
     if (credentialResponse.isErr()) {
         const errorLog = await errorNotificationService.auth.get(connectionRes.response.id as number);
+
+        // When we failed to refresh we still return a 200 because the connection is used in the UI
+        // Ultimately this could be a second endpoint so the UI displays faster and no confusion between error code
         res.status(200).send({ errorLog, provider: integration.provider, connection: connectionRes.response });
+
         return;
     }
 
--- a/packages/server/lib/controllers/v1/connection/get.ts
+++ b/packages/server/lib/controllers/v1/connection/get.ts
@@ -1,7 +1,7 @@
 import { z } from 'zod';
 import { asyncWrapper } from '../../../utils/asyncWrapper.js';
 import { requireEmptyBody, zodErrorToHTTP } from '@nangohq/utils';
-import type { Connection, GetConnection, IntegrationConfig } from '@nangohq/types';
+import type { GetConnection, IntegrationConfig } from '@nangohq/types';
 import { connectionService, configService, errorNotificationService } from '@nangohq/shared';
 import { connectionRefreshFailed as connectionRefreshFailedHook, connectionRefreshSuccess as connectionRefreshSuccessHook } from '../../../hooks/hooks.js';
 import { logContextGetter } from '@nangohq/logs';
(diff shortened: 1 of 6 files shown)


packages/server/lib/refreshConnections.ts
```diff
@@ -56,21 +56,22 @@ export async function exec(): Promise<void> {
             const limit = 1000;
             // eslint-disable-next-line no-constant-condition
             while (true) {
-                const staleConnections = await connectionService.getStaleConnections({ days: 0, limit, cursor });
+                const staleConnections = await connectionService.getStaleConnections({ days: 1, limit, cursor });
```
bodinsamuel, line 59: I think it was an error, let me know @TBonnin

packages/server/lib/hooks/connection/post-connection.ts
```diff
@@ -23,22 +22,12 @@ async function execute(createdConnection: RecentlyCreatedConnection, providerNam
...
-            providerConfigKey: upsertedConnection.provider_config_key,
-            logContextGetter,
-            instantRefresh: false,
-            onRefreshSuccess: connectionRefreshSuccessHook,
-            onRefreshFailed: connectionRefreshFailedHook
-        });
-
-        if (credentialResponse.isErr()) {
+        const connectionRes = await connectionService.getConnection(upsertedConnection.connection_id, upsertedConnection.provider_config_key, environment.id);
+        if (connectionRes.error || !connectionRes.response) {
```
bodinsamuel, line 26: On post connection I don't think we need to test again the credentials since we just fetched them, but maybe I forgot an edge case? @khaliqgant

packages/server/lib/refreshConnections.ts
TBonnin: yes, that's a test leftover. Thank you for fixing

packages/server/lib/refreshConnections.ts
```diff
@@ -36,7 +36,7 @@ export function refreshConnectionsCron(): void {
 }
 
 export async function exec(): Promise<void> {
-    return await tracer.trace<Promise<void>>('nango.server.cron.refreshConnections', async (span) => {
+    await tracer.trace<Promise<void>>('nango.server.cron.refreshConnections', async (span) => {
```
TBonnin, line 39: exec is itself awaited above.

packages/shared/lib/services/connection.service.ts
```diff
@@ -985,55 +985,26 @@ class ConnectionService {
...
-        }
-
-        const provider = getProvider(config?.provider);
+        const provider = getProvider(integration.provider);
         if (!provider) {
             const error = new NangoError('unknown_provider_config');
             return Err(error);
         }
 
+        const copy = { ...connection };
```
TBonnin, line 994: why a copy?

packages/server/lib/controllers/v1/connection/get.ts
```diff
@@ -53,87 +53,73 @@ export const getConnection = asyncWrapper<GetConnection>(async (req, res) => {
...
-                        error: {
-                            code: 'unknown_provider_config',
-                            message: credentialResponse.error.message
-                        }
-                    });
-                    break;
-            }
-        }
+        const errorLog = await errorNotificationService.auth.get(connectionRes.response.id as number);
+        res.status(200).send({ errorLog, provider: integration.provider, connection: connectionRes.response });
```
TBonnin, line 112: should we return something else than 200 in case of error?

packages/shared/lib/services/connection.service.ts
```diff
@@ -1063,36 +1034,39 @@ class ConnectionService {
...
+                const errorWithPayload = new NangoError(error!.type, { connection: connectionWithoutCredentials });
 
                 // there was an attempt to refresh the token so clear it from the queue
                 // of connections to refresh if it failed
-                await this.updateLastFetched(connection.id);
+                await this.updateLastFetched(connection.id!);
 
+                /**
+                 * Be careful this Error contains credentials
+                 */
```
TBonnin: is this comment correct? errorWithPayload should not contain credentials
```
                const { credentials, ...connectionWithoutCredentials } = copy;
                const errorWithPayload = new NangoError(error!.type, { connection: connectionWithoutCredentials });
```

packages/server/lib/hooks/connection/post-connection.ts
khaliqgant: No, should be fine to just use `getConnection` here

packages/shared/lib/services/connection.service.ts
bodinsamuel: it's modifying the credentials by reference since we don't fetch anymore inside, it's best practice to copy (even though it's a shallow copy)

packages/server/lib/controllers/v1/connection/get.ts
bodinsamuel: The thing is in the UI it's still used to display the page, before that we returned an error by did not respected that and still displayed the page. My take on this: it should be a totally different endpoint.

packages/shared/lib/services/connection.service.ts
bodinsamuel: oops leftover from the first fix ty

packages/server/lib/controllers/v1/connection/get.ts
TBonnin: ok. can you add a comment?

packages/server/lib/refreshConnections.ts
bodinsamuel: yes, it's just eslint best practice rules when you return void https://typescript-eslint.io/rules/no-confusing-void-expression/

TBonnin (approved):

PR #2874, opened 2024-10-22
feat: Add support for record count

mbiddle153:
## Describe your changes
Adds method to get count of records for a given connection and model (I believe has to be on a per-model basis due to the records table partitioning).

Aggregates and attaches that count to the returned `SyncResponse` object and displays it in the `SyncRow` component.

### DIFF
--- a/packages/records/lib/db/migrations/20241023085055_create_record_counts_table.ts
+++ b/packages/records/lib/db/migrations/20241023085055_create_record_counts_table.ts
@@ -17,8 +17,8 @@ export async function up(knex: Knex): Promise<void> {
 
         // INDEXES
         await knex.schema.alterTable(TABLE, function (table) {
-            table.unique(['connection_id', 'environment_id', 'model']);
-            table.index(['connection_id', 'environment_id', 'model']);
+            table.unique(['environment_id', 'connection_id', 'model']);
+            table.index(['environment_id', 'connection_id', 'model']);
         });
     });
 }
--- a/packages/records/lib/db/migrations/20241023085055_create_record_counts_table.ts
+++ b/packages/records/lib/db/migrations/20241023085055_create_record_counts_table.ts
@@ -1,0 +1,31 @@
+import type { Knex } from 'knex';
+
+const TABLE = 'record_counts';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.transaction(async (trx) => {
+        // TABLE
+        await trx.raw(`
+            CREATE TABLE "${TABLE}" (
+                connection_id integer NOT NULL,
+                environment_id integer NOT NULL,
+                model character varying(255) NOT NULL,
+                object_count integer NOT NULL,
+                updated_at timestamp with time zone NOT NULL DEFAULT CURRENT_TIMESTAMP
+            )
+        `);
+
+        // INDEXES
+        await knex.schema.alterTable(TABLE, function (table) {
+            // Maybe include environment_id on uniqueness and compound index?
+            table.unique(['connection_id', 'model']);
+            table.index(['connection_id', 'model']);
+        });
+    });
+}
+
+export async function down(knex: Knex): Promise<void> {
+    // TABLE
+    // INDEXES are dropped automatically
+    await knex.raw(`DROP TABLE IF EXISTS "${TABLE}"`);
+}
--- a/packages/records/lib/db/migrations/20241023085055_create_record_counts_table.ts
+++ b/packages/records/lib/db/migrations/20241023085055_create_record_counts_table.ts
@@ -17,7 +17,6 @@ export async function up(knex: Knex): Promise<void> {
 
         // INDEXES
         await knex.schema.alterTable(TABLE, function (table) {
-            // Maybe include environment_id on uniqueness and compound index?
             table.unique(['connection_id', 'environment_id', 'model']);
             table.index(['connection_id', 'environment_id', 'model']);
         });
--- a/packages/records/lib/db/migrations/20241023085055_create_record_counts_table.ts
+++ b/packages/records/lib/db/migrations/20241023085055_create_record_counts_table.ts
@@ -18,8 +18,8 @@ export async function up(knex: Knex): Promise<void> {
         // INDEXES
         await knex.schema.alterTable(TABLE, function (table) {
             // Maybe include environment_id on uniqueness and compound index?
-            table.unique(['connection_id', 'model']);
-            table.index(['connection_id', 'model']);
+            table.unique(['connection_id', 'environment_id', 'model']);
+            table.index(['connection_id', 'environment_id', 'model']);
         });
     });
 }
--- a/packages/records/lib/db/migrations/20241023085055_create_record_counts_table.ts
+++ b/packages/records/lib/db/migrations/20241023085055_create_record_counts_table.ts
@@ -6,11 +6,11 @@ export async function up(knex: Knex): Promise<void> {
     await knex.transaction(async (trx) => {
         // TABLE
         await trx.raw(`
-            CREATE TABLE "${TABLE}" (
-                connection_id integer NOT NULL,
+            CREATE TABLE IF NOT EXISTS "${TABLE}" (
                 environment_id integer NOT NULL,
+                connection_id integer NOT NULL,
                 model character varying(255) NOT NULL,
-                object_count integer NOT NULL,
+                count integer NOT NULL,
                 updated_at timestamp with time zone NOT NULL DEFAULT CURRENT_TIMESTAMP
             )
         `);
@@ -24,7 +24,6 @@ export async function up(knex: Knex): Promise<void> {
 }
 
 export async function down(knex: Knex): Promise<void> {
-    // TABLE
-    // INDEXES are dropped automatically
     await knex.raw(`DROP TABLE IF EXISTS "${TABLE}"`);
+    // Note: INDEXES are dropped automatically
 }
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -34,7 +34,7 @@ export async function getRecordsCount({ connectionId, model }: { connectionId: n
 
         return Ok(result[0]?.object_count ?? 0);
     } catch (_error) {
-        const e = new Error(`List records error for model ${model}`);
+        const e = new Error(`Count records error for model ${model}`);
         return Err(e);
     }
 }
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -22,6 +22,23 @@ dayjs.extend(utc);
 
 const BATCH_SIZE = 1000;
 
+export async function getRecordsCount({ connectionId, model }: { connectionId: number; model: string }): Promise<Result<number>> {
+    try {
+        const result = (await db(RECORDS_TABLE).count('* as object_count').where({ connection_id: connectionId, model, deleted_at: null }).select()) as {
+            object_count: number;
+        }[];
+
+        if (result.length === 0) {
+            return Ok(0);
+        }
+
+        return Ok(result[0]?.object_count ?? 0);
+    } catch (_error) {
+        const e = new Error(`List records error for model ${model}`);
+        return Err(e);
+    }
+}
+
 export async function getRecords({
     connectionId,
     model,
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -369,11 +369,13 @@ export async function update({
 
 export async function deleteRecordsBySyncId({
     connectionId,
+    environmentId,
     model,
     syncId,
     limit = 5000
 }: {
     connectionId: number;
+    environmentId: number;
     model: string;
     syncId: string;
     limit?: number;
@@ -392,6 +394,7 @@ export async function deleteRecordsBySyncId({
             .del();
         totalDeletedRecords += deletedRecords;
     } while (deletedRecords >= limit);
+    await deleteRecordCount({ connectionId, environmentId, model });
 
     return { totalDeletedRecords };
 }
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -6,6 +6,7 @@ import type {
     FormattedRecordWithMetadata,
     GetRecordsResponse,
     LastAction,
+    RecordCount,
     ReturnedRecord,
     UnencryptedRecord,
     UpsertSummary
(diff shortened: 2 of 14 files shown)


mbiddle153:
Would appreciate guidance on the PR checklist - what all is expected for something like this?

packages/server/lib/controllers/sync.controller.ts
```diff
@@ -118,14 +118,42 @@ class SyncController {
...
+            if (!(model in objectCountByModel)) {
+                const count = await recordsService.getRecordsCount({ connectionId, model });
+                objectCountByModel[model] = count.isOk() ? count.value : 0;
+            }
+        }
+
+        return syncs.map((sync) => {
+            const sumObjectCount = (sync: Sync & { models: string[] }, objectCountByModel: Record<string, number>) => {
+                return sync.models.reduce((sum: number, model: string) => {
+                    return sum + Number(objectCountByModel[model] || 0);
```
mbiddle153: Wrapped this in `Number()` since otherwise `objectCountByModel[model]` was coming back as a string... unclear to me why

packages/records/lib/models/records.ts
```diff
@@ -22,6 +22,23 @@ dayjs.extend(utc);
...
+    try {
+        const result = (await db(RECORDS_TABLE).count('* as object_count').where({ connection_id: connectionId, model, deleted_at: null }).select()) as {
+            object_count: number;
+        }[];
+
+        if (result.length === 0) {
+            return Ok(0);
+        }
+
+        return Ok(result[0]?.object_count ?? 0);
```
TBonnin: `count` from Postgres are returned as bigInt which are interpreted by knex (the db library we are using) as string. More info https://knexjs.org/guide/query-builder.html#count
Typescript compiler is lying here since it is not necessary returning a `Result<number>` 😉 

I would do something like this and ensure it returns a number:
```
    const [result] = await db
       .from(RECORDS_TABLE).count('*')
       .where({
          connection_id: connectionId,
          model,
          deleted_at: null
       });
    return Ok(result ? Number(result) : 0);
```

packages/server/lib/controllers/sync.controller.ts
```diff
@@ -118,14 +118,42 @@ class SyncController {
...
             next(e);
         }
     }
 
+    private async addObjectCount(syncs: (Sync & { models: string[] })[], connectionId: number) {
+        const models = syncs.map((sync) => sync.models).flat();
+        const objectCountByModel: Record<string, number> = {};
+        for (const model of models) {
+            if (!(model in objectCountByModel)) {
+                const count = await recordsService.getRecordsCount({ connectionId, model });
```
TBonnin: Note: Some connections/models might have millions of records. Counting records inline could take several seconds each time and make the call to the endpoint quite slow.

packages/server/lib/controllers/sync.controller.ts
TBonnin: see my comment above about postgres count result

packages/webapp/src/pages/Connection/components/SyncRow.tsx
```diff
@@ -134,6 +134,7 @@ export const SyncRow: React.FC<{ sync: SyncResponse; connection: Connection; pro
                 </Link>
             </Table.Cell>
             <Table.Cell bordered>{formatFrequency(sync.frequency)}</Table.Cell>
+            <Table.Cell bordered>{sync.object_count ?? '-'}</Table.Cell>
```
TBonnin: should we format the count somewhat? How does it look in the UI when count is big?

TBonnin (commented):
PR checklist is more of a reminder. You don't have to do everything

packages/server/lib/controllers/sync.controller.ts
mbiddle153: Per our discussion will be investigating an alternative approach - store the per-(connection_id, model) count somewhere in the database, and update it at the end of each sync task

bodinsamuel (commented):
Works great locally, I think Thomas already raised the important points, let's see the next version of this 🙏🏻

packages/records/lib/models/records.ts
```diff
@@ -227,6 +240,21 @@ export async function upsert({
...
                 });
+
+                const delta_count = chunkSummary.addedKeys.length - (chunkSummary.deletedKeys?.length ?? 0);
+                await trx
+                    .from(RECORD_COUNTS_TABLE)
+                    .insert({
+                        connection_id: connectionId,
+                        model,
+                        environment_id: environmentId,
+                        object_count: Math.max(0, delta_count)
```
mbiddle153: Doing max here since the first time this runs, if the sync job happens to be deleting more than it adds, we'll end up with a negative count.

packages/records/lib/db/migrations/20241023085055_create_record_counts_table.ts
```diff
@@ -0,0 +1,31 @@
...
+                environment_id integer NOT NULL,
+                model character varying(255) NOT NULL,
+                object_count integer NOT NULL,
+                updated_at timestamp with time zone NOT NULL DEFAULT CURRENT_TIMESTAMP
+            )
+        `);
+
+        // INDEXES
+        await knex.schema.alterTable(TABLE, function (table) {
+            // Maybe include environment_id on uniqueness and compound index?
```
bodinsamuel: yes for both indexes you need the environment id otherwise there is a risk of collision and index will not be used for querying

packages/records/lib/models/records.ts
```diff
@@ -22,6 +22,17 @@ dayjs.extend(utc);
 
 const BATCH_SIZE = 1000;
 
+export async function getRecordsCount({ connectionId, model }: { connectionId: number; model: string }): Promise<Result<number>> {
+    try {
+        const [result] = await db.from(RECORD_COUNTS_TABLE).where({ connection_id: connectionId, model }).select('object_count');
+
+        return Ok(result ? Number(result.object_count) : 0);
+    } catch (_error) {
```
bodinsamuel: note if you are not using the error you can omit entirely the `(err) `

packages/records/lib/models/records.ts
```diff
@@ -22,6 +22,17 @@ dayjs.extend(utc);
 
 const BATCH_SIZE = 1000;
 
+export async function getRecordsCount({ connectionId, model }: { connectionId: number; model: string }): Promise<Result<number>> {
```
bodinsamuel: could be moved to a separate file since we are in a models folder (aka one file = one model = one table)
Also missing the environmentId in the where

packages/server/lib/controllers/sync.controller.ts
bodinsamuel: ideally here you select all count of envId + connectionId and then find the appropriate Model count in the for loop so we can do it a one query instead of N

packages/server/lib/controllers/sync.controller.ts
```diff
@@ -118,14 +118,42 @@ class SyncController {
                 return;
             }
 
-            const syncs = await getSyncs(connection, orchestrator);
+            const rawSyncs = await getSyncs(connection, orchestrator);
+
+            if (!connection.id) {
```
bodinsamuel: note this if is not necessary, the typing is incorrect for legacy reasons, if you have a connection you have its id. You can safely suppress `connect.id!` until we fix

packages/records/lib/models/records.ts
bodinsamuel: is this possible? First run can't deleteKeys (AFAIK)
edit: maybe on the second generation and `markNonCurrentGenerationRecordsAsDeleted` kicks in? @TBonnin

bodinsamuel (commented):
Looks good and works locally 🙏🏻 
Few comments but it's almost ready imo, congrats!

packages/records/lib/db/migrations/20241023085055_create_record_counts_table.ts
mbiddle153: What exactly is the environment? Based on the schema it looks like it corresponds to a customer's environment?

packages/records/lib/models/records.ts
mbiddle153: I was thinking if this is deployed to existing syncs, there will be no `record_counts` entry for the sync, and so this update will be starting from an incorrect count (hence why I think we need the temporary task to get the initial correct count for everybody first).

But for this specifically, I guess it doesn't make much sense to arbitrarily limit it to zero if the reported count can be inaccurate anyway

packages/records/lib/db/migrations/20241023085055_create_record_counts_table.ts
bodinsamuel: yes, 
account = organisation 
environment is like a workspace inside an or, it contains the scripts, the records, the logs, etc.
Everything should be grouped at the environment level for security/isolation reason. Because two envs can have the  same connection_id, scripts names, etc.

packages/records/lib/models/records.ts
mbiddle153: I think returning the error here is still useful though, right? Maybe I am misinterpreting

packages/records/lib/models/records.ts
mbiddle153: Will leave these methods in this file for now, unless you want to push back. It complicates the namespace, and we end up passing the `recordsService` as a prop some places, and would need to add an additional `recordCountsService` prop there, which to me feels heavy for a simple helper that is pretty tightly coupled to the records service

packages/server/lib/controllers/sync.controller.ts
mbiddle153: Good suggestion - much cleaner approach

packages/records/lib/models/records.ts
bodinsamuel: I meant the `catch (_error) {` could be replaced with `catch {

packages/server/lib/controllers/sync.controller.ts
```diff
@@ -118,14 +118,33 @@ class SyncController {
...
         } catch (e) {
             next(e);
         }
     }
 
+    private async addObjectCount(syncs: (Sync & { models: string[] })[], connectionId: number, environmentId: number) {
+        const objectCountByModelResult = await recordsService.getRecordCountsByModel({ connectionId, environmentId });
+
+        if (objectCountByModelResult.isOk()) {
+            const objectCountByModel = objectCountByModelResult.unwrap();
```
TBonnin: we can use `objectCountByModelResult.value` since we are in the `isOk` if

packages/server/lib/controllers/sync.controller.ts
```diff
@@ -118,14 +118,33 @@ class SyncController {
...
+            return syncs.map((sync) => {
+                let objectCountSum = 0;
+                for (const model of sync.models) {
+                    objectCountSum = objectCountSum + (objectCountByModel[model]?.object_count ?? 0);
+                }
+                return { ...sync, object_count: objectCountSum };
+            });
```
TBonnin: can we use reduce instead of the for loop
```
  return syncs.map(sync => ({
    ...sync,
    object_count: sync.models.reduce(
      (sum, model) => sum + (result.value[model]?.object_count ?? 0),
      0
    )
  }));
```

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -590,6 +591,7 @@ export class Orchestrator {
                     const syncConfig = await getSyncConfigBySyncId(syncId);
                     for (const model of syncConfig?.models || []) {
                         const del = await recordsService.deleteRecordsBySyncId({ syncId, connectionId, model });
+                        await recordsService.deleteRecordCount({ connectionId, environmentId, model });
```
TBonnin: I would call `deleteRecordCount` directly inside `deleteRecordsBySyncId` since it always need to happen when deleting records

packages/records/lib/models/records.ts
TBonnin: > hence why I think we need the temporary task to get the initial correct count for everybody first.

I am not asking you to implement this task but how would you go about it?

packages/records/lib/db/migrations/20241023085055_create_record_counts_table.ts
```diff
@@ -0,0 +1,30 @@
...
+                model character varying(255) NOT NULL,
+                object_count integer NOT NULL,
+                updated_at timestamp with time zone NOT NULL DEFAULT CURRENT_TIMESTAMP
+            )
+        `);
+
+        // INDEXES
+        await knex.schema.alterTable(TABLE, function (table) {
+            table.unique(['connection_id', 'environment_id', 'model']);
+            table.index(['connection_id', 'environment_id', 'model']);
```
bodinsamuel: environment_id should be first in index for efficient filtering

packages/records/lib/models/records.ts
```diff
@@ -22,6 +23,30 @@ dayjs.extend(utc);
...
+    environmentId: number;
+}): Promise<Result<Record<string, RecordCount>>> {
+    try {
+        const results = await db
+            .from(RECORD_COUNTS_TABLE)
+            .where({
+                connection_id: connectionId,
+                environment_id: environmentId
+            })
+            .select('*');
```
bodinsamuel: You need to type the select `select<Type>()`

packages/records/lib/types.ts
```diff
@@ -58,3 +58,10 @@ export interface UpsertSummary {
     deletedKeys?: string[];
     nonUniqueKeys: string[];
 }
+
+export interface RecordCount {
+    model: string;
+    connectionId: number;
+    environmentId: number;
+    object_count: number;
+}
```
bodinsamuel, line 68: The type is missing updated_at and has inconsistent casing

bodinsamuel (approved):
Good to go 🚀 
It will probably miss the one-off script to fill the data, only if you have time 👌🏻

packages/records/lib/models/records.ts
mbiddle153: Ah sorry - never got any of my notifications properly set up so missed this comment.

I'll talk about it during the presentation, but my plan was to set up a temporary "admin endpoint" that only we can hit (would need to understand details of how to set something like that up here), that would enqueue a job for every `(environment_id, connection_id)` pair. This job would get the full count for each model on that pair and update the `record_counts` table with that info. These jobs could be spread out over some amount of time (maybe an hour? maybe more) to avoid locking up database resources/locking the tables themselves for an extended period of time.

bodinsamuel (approved):

PR #2880, opened 2024-10-23
feat(notion-scim): Support notion SCIM

khaliqgant:
## Describe your changes
Add Notion scim as a specific provider


packages/shared/providers.yaml
```diff
@@ -3495,6 +3495,23 @@ notion:
...
+    proxy:
+        base_url: https://api.notion.com/scim
+        headers:
+            Authorization: Bearer ${apiKey}
+    docs: https://docs.nango.dev/integrations/all/notion
+    credentials:
+        apiKey:
+            type: string
+            title: API Key
+            description: The API key for your Notion account
```
bodinsamuel, line 3551: It's missing validation and verification, otherwise somebody will need to painfully backfill those 😬

khaliqgant: Yeah. Missing both because we can’t access a Notion enterprise account so I dont know what either look like

bodinsamuel (approved):

PR #2883, opened 2024-10-24
fix(connections): add FK to end users

bodinsamuel:
## Describe your changes

Contributes to https://linear.app/nango/issue/NAN-1944/connect-end-users-to-connections

- Add Foreign Key from connections to end_users
Implementation will come after

- Add Foreign Key from oauth_session to connect_session to be able to get back the session on the callback
Implementation will come after

### DIFF
--- a/packages/database/lib/migrations/20241024094439_connections_end_user.cjs
+++ b/packages/database/lib/migrations/20241024094439_connections_end_user.cjs
@@ -1,0 +1,15 @@
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`ALTER TABLE "_nango_connections" ADD COLUMN "end_user_id" int4`);
+
+    await knex.raw(`ALTER TABLE "_nango_connections" ADD FOREIGN KEY ("end_user_id") REFERENCES "end_users" ("id") ON DELETE SET NULL`);
+};
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.down = async function (knex) {
+    await knex.raw(`ALTER TABLE "_nango_connections" DROP COLUMN "end_user_id"`);
+};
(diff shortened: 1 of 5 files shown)


packages/shared/lib/models/Connection.ts
```diff
@@ -7,6 +7,7 @@ export type ConnectionConfig = Record<string, any>;
 export interface BaseConnection extends TimestampsAndDeleted {
     id?: number;
     config_id?: number;
+    end_user_id?: number;
```
bodinsamuel, line 10: undefined on purpose until I fix the implementation

packages/database/lib/migrations/20241024094439_connections_end_user.cjs
```diff
@@ -0,0 +1,18 @@
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`ALTER TABLE "_nango_connections" ADD COLUMN "end_user_id" int4`);
+    await knex.raw(`ALTER TABLE "_nango_connections" ADD FOREIGN KEY ("end_user_id") REFERENCES "end_users" ("id") ON DELETE SET NULL`);
```
TBonnin, line 6: can we not do that in one single query:
```
ALTER TABLE "_nango_connections" 
  ADD COLUMN "end_user_id" int4,
  ADD FOREIGN KEY ("end_user_id") REFERENCES "end_users" ("id") ON DELETE SET NULL;
```

TBonnin (approved):

packages/database/lib/migrations/20241024094439_connections_end_user.cjs
bodinsamuel: I'm sure I could, not sure what would be the benefit, it's 2 different queries

PR #2884, opened 2024-10-24
feat(connections): link to end_user on success

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1944/connect-end-users-to-connections

- Link end_user on every connection creation (:|) 

## Tests

Easiest is to use the connect-ui right now, it should populate the end_user_id correctly on a successful flow

### DIFF
--- a/packages/server/lib/controllers/appAuth.controller.ts
+++ b/packages/server/lib/controllers/appAuth.controller.ts
@@ -19,6 +19,9 @@ import publisher from '../clients/publisher.client.js';
 import { logContextGetter } from '@nangohq/logs';
 import { stringifyError } from '@nangohq/utils';
 import { connectionCreated as connectionCreatedHook, connectionCreationFailed as connectionCreationFailedHook } from '../hooks/hooks.js';
+import { linkConnection } from '../services/endUser.service.js';
+import db from '@nangohq/database';
+import { getConnectSession } from '../services/connectSession.service.js';
 
 class AppAuthController {
     async connect(req: Request, res: Response<any, never>, _next: NextFunction) {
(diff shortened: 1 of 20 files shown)


packages/server/lib/controllers/appAuth.controller.ts
```diff
@@ -179,24 +182,38 @@ class AppAuthController {
...
-                    logContextGetter,
-                    undefined,
-                    logCtx
-                );
+            if (session.connectSessionId) {
+                const connectSession = await getConnectSession(db.knex, { id: session.connectSessionId, accountId: account.id, environmentId: environment.id });
+                if (connectSession.isErr()) {
+                    await logCtx.error('Failed to get session');
+                    await logCtx.failed();
+                    return publisher.notifyErr(res, wsClientId, providerConfigKey, connectionId, WSErrBuilder.UnknownError('failed to get session'));
```
TBonnin, line 196: should we failed the connection creation in this case? we'll end up returning a failure even though the connection db upsert above was successful

packages/server/lib/controllers/auth/postBill.ts
```diff
@@ -150,24 +149,36 @@ export const postPublicBillAuthorization = asyncWrapper<PostPublicBillAuthorizat
...
+                account,
+                auth_mode: 'BILL',
+                operation: updatedConnection.operation
+            },
+            config.provider,
+            logContextGetter,
+            undefined,
+            logCtx
+        );
+
```
TBonnin, line 181: I really don't like all those provider specific auth modes. So much code must be duplicated 😢

packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -1251,6 +1276,26 @@ class OAuthController {
...
+                if (session.connectSessionId) {
+                    const connectSession = await getConnectSession(db.knex, {
+                        id: session.connectSessionId,
+                        accountId: account.id,
+                        environmentId: environment.id
+                    });
+                    if (connectSession.isErr()) {
+                        await logCtx.error('Failed to get session');
+                        await logCtx.failed();
+                        return publisher.notifyErr(res, channel, providerConfigKey, connectionId, WSErrBuilder.UnknownError('failed to get session'));
```
TBonnin, line 1294: same question about failing after the connection was successfully inserted in the db

packages/shared/lib/utils/encryption.manager.ts
```diff
@@ -64,29 +64,14 @@ export class EncryptionManager extends Encryption {
         return decryptedEnvironment;
     }
 
-    public encryptApiConnection(connection: ApiConnection): Omit<StoredConnection, 'created_at' | 'updated_at'> {
```
TBonnin, line 67: is `ApiConnection` still use elsewhere then? if no it can be deleted

bodinsamuel: I have it cleaned already in a new PR it was just annoying to fix to have it deleted after 👌🏻

packages/server/lib/controllers/appAuth.controller.ts
bodinsamuel: ideally we'll add a transaction that's why I kept the `db` params, I feel like it's a very unlikely event for the moment. 
I'll create a followup issue if that's alright?

packages/server/lib/controllers/oauth.controller.ts
bodinsamuel: answered above

packages/server/lib/controllers/appAuth.controller.ts
TBonnin: sounds reasonable

TBonnin (approved):

PR #2887, opened 2024-10-24
fix(flows): endpoint deduplication conflict

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1952/quickbooks-templates-differ-between-environments

- Fix endpoint deduplication when multiples are returning the same Model name
It was okay for syncs but not for actions

- Also took the opportunity to improve the endpoint sort and grouping which nows works "correctly" even in a single group (Group by endpoint, then sort by endpoints)
Screenshot 2024-10-24 at 18 25 51


khaliqgant (approved):

packages/webapp/src/pages/Integrations/providerConfigKey/Endpoints/Show.tsx
```diff
@@ -54,22 +54,27 @@ export const EndpointsShow: React.FC<{ integration: GetIntegration['Success']['d
...
+                        const lenA = (a.endpoint.path.match(/\//g) || []).length;
+                        const lenB = (b.endpoint.path.match(/\//g) || []).length;
+                        if (lenA > lenB) return 1;
+                        else if (lenA < lenB) return -1;
+
+                        // Sort alphabetically
+                        return a.endpoint.path > b.endpoint.path ? 1 : -1;
+                    })
+                    .sort((a, b) => {
+                        if (a.endpoint.path !== b.endpoint.path) return 0;
```
TBonnin, line 69: not sure I understand the 0 value here when paths are different

TBonnin (approved):

packages/webapp/src/pages/Integrations/providerConfigKey/Endpoints/Show.tsx
bodinsamuel: Basically after the first sort we have grouped by path, e.g: `/accounts -> /customers -> /zoobar` but method are still unordered. We don't want to change the order of the group, we just want to reorder inside a "group" by method so if the path are different we don't up or down the position

PR #2888, opened 2024-10-24
fix: use DD_SITE env var instead of hardcoded value

TBonnin:
## Describe your changes

Using the DD_SITE env var in telemetry.ts instead of hardcoded value. 

https://linear.app/nango/issue/NAN-1957/solve-datadog-issues

### DIFF
--- a/packages/shared/lib/utils/telemetry.ts
+++ b/packages/shared/lib/utils/telemetry.ts
@@ -44,10 +44,10 @@ class Telemetry {
     private logInstance: v2.LogsApi | undefined;
     constructor() {
         try {
-            if ((isCloud || isEnterprise) && process.env['DD_API_KEY'] && process.env['DD_APP_KEY']) {
+            if ((isCloud || isEnterprise) && process.env['DD_API_KEY'] && process.env['DD_APP_KEY'] && process.env['DD_SITE']) {
                 const configuration = client.createConfiguration();
                 configuration.setServerVariables({
-                    site: 'us3.datadoghq.com'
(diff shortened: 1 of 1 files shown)


khaliqgant (approved):

packages/shared/lib/utils/telemetry.ts
```diff
@@ -44,10 +44,10 @@ class Telemetry {
     private logInstance: v2.LogsApi | undefined;
     constructor() {
         try {
-            if ((isCloud || isEnterprise) && process.env['DD_API_KEY'] && process.env['DD_APP_KEY']) {
+            if ((isCloud || isEnterprise) && process.env['DD_API_KEY'] && process.env['DD_APP_KEY'] && process.env['DD_SITE']) {
                 const configuration = client.createConfiguration();
                 configuration.setServerVariables({
-                    site: 'us3.datadoghq.com'
+                    site: process.env['DD_SITE']
```
bodinsamuel, line 50: I'm not sure we enforce this value, I would just make sure it works or has a default value

bodinsamuel (approved):

packages/shared/lib/utils/telemetry.ts
TBonnin: what do you mean by `enforce`?
btw it is an already existing env var that we are already using in other part of the codebase

bodinsamuel: I meant be sure it's there but I forgot there was the if above 🤦🏻

PR #2891, opened 2024-10-24
feat: add migration for otlp settings

TBonnin:
## Describe your changes

Adding column otlp_settings to environment table. Will be used in my next PR
It will contains info about where to send the openTelemetry data. ex:
```
{
   'endpoint': 'http://otlp.agent:4318/v1',
   'headers': { 'x-api-key': 'xxx'}
}
```

https://linear.app/nango/issue/NAN-1782/opentelemetry-export

### DIFF
--- a/packages/types/lib/environment/db.ts
+++ b/packages/types/lib/environment/db.ts
@@ -39,6 +39,7 @@ export interface DBEnvironment extends Timestamps {
     slack_notifications: boolean;
 
     webhook_receive_url?: string;
+    otlp_settings?: { endpoint: string; headers: Record<string, string> } | null;
 }
 
 export interface ExternalWebhook extends Timestamps {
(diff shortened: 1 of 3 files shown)


packages/types/lib/environment/db.ts
```diff
@@ -39,6 +39,7 @@ export interface DBEnvironment extends Timestamps {
     slack_notifications: boolean;
 
     webhook_receive_url?: string;
+    otlp_settings?: { endpoint: string; headers: Record<string, string> } | null;
```
bodinsamuel, line 42: remember to remove `?` on the next PR 👌🏻

bodinsamuel (approved):

packages/types/lib/environment/db.ts
TBonnin: thanks as you saw I had forgotten

PR #2892, opened 2024-10-24
feat: add OpenTelemetry export

TBonnin:
Log operations are sent as OpenTelemetry traces to a otlp backend set by customers on the environment level
It is using a custom SpanProcessor from opentelemetry-js lib with one exporter per customers. I assume this feature will only be available to scale plan customers and therefore the number of exporters will stay low
Another thing to note is that operation must be started and ended on the same service, which is the case today but could become a limitation in the future.
It could be extended later on by also exporting otlp logs and attach them to the traces, the same way we attach logs to operation in Nango

I have tested with an oltp collector like jaeger running on my machine and with Honeycomb. 
To export to Datadog, one must go via a recent version of the datadog-agent

https://linear.app/nango/issue/NAN-1782/opentelemetry-export

### DIFF
--- a/packages/logs/lib/otlp/otlp.ts
+++ b/packages/logs/lib/otlp/otlp.ts
@@ -32,7 +32,9 @@ function registerRoutes(routes: RouteConfig[]) {
             routingProcessor.updateRoutes(routes);
         } else {
             routingProcessor = new RoutingSpanProcessor(routes);
-            const provider = new NodeTracerProvider({ resource: new Resource({ [ATTR_SERVICE_NAME]: 'nango-otlp' }) });
+            const provider = new NodeTracerProvider({
+                resource: new Resource({ [ATTR_SERVICE_NAME]: 'nango-otlp' })
+            });
             provider.addSpanProcessor(routingProcessor);
             provider.register();
         }
--- a/packages/logs/lib/otlp/otlp.ts
+++ b/packages/logs/lib/otlp/otlp.ts
@@ -1,0 +1,60 @@
+import { trace } from '@opentelemetry/api';
+import { NodeTracerProvider } from '@opentelemetry/sdk-trace-node';
+import { Resource } from '@opentelemetry/resources';
+import { ATTR_SERVICE_NAME } from '@opentelemetry/semantic-conventions';
+import { once, stringifyError } from '@nangohq/utils';
+import { logger } from '../utils.js';
+import { RoutingSpanProcessor } from './otlpSpanProcessor.js';
+import { setInterval } from 'timers';
+
+// Enable OpenTelemetry console logging
+// import { DiagLogLevel, DiagConsoleLogger, diag } from '@opentelemetry/api';
+// diag.setLogger(new DiagConsoleLogger(), DiagLogLevel.DEBUG);
+
+export const otlpRoutingAttributeKey = 'otlp.internal.routingKey';
+
+export interface RouteConfig {
+    routingId: string;
+    routingEndpoint: string;
+    routingHeaders: Record<string, string>;
+}
+
+let routingProcessor: RoutingSpanProcessor | null = null;
+process.on('SIGTERM', async () => {
+    if (routingProcessor) {
+        await routingProcessor.shutdown();
+    }
+});
+
+function registerRoutes(routes: RouteConfig[]) {
+    try {
+        if (routingProcessor) {
+            routingProcessor.updateRoutes(routes);
+        } else {
+            routingProcessor = new RoutingSpanProcessor(routes);
+            const provider = new NodeTracerProvider({ resource: new Resource({ [ATTR_SERVICE_NAME]: 'nango-otlp' }) });
+            provider.addSpanProcessor(routingProcessor);
+            provider.register();
+        }
+    } catch (err) {
+        logger.error(`failed_to_register_otlp_routes ${stringifyError(err)}`);
+    }
+}
+
+async function updateRoutes(getRoutes: () => Promise<RouteConfig[]>) {
+    try {
+        const routes = await getRoutes();
+        registerRoutes(routes);
+    } catch (err) {
+        logger.error(`failed_to_update_otlp_routes ${stringifyError(err)}`);
+    }
+}
+
+export const otlp = {
+    register: once(async (getRoutes: () => Promise<RouteConfig[]>) => {
+        await updateRoutes(getRoutes);
+        setInterval(async () => await updateRoutes(getRoutes), 10000);
+    }),
+    tracer: trace.getTracer('nango-otlp'),
+    routingAttributeKey: 'otlp.internal.routingKey'
+};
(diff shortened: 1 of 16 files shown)


packages/logs/lib/otlp/otlp.ts
```diff
@@ -0,0 +1,53 @@
...
+        registerRoutes(routes);
+    } catch (err) {
+        logger.error(`failed_to_update_otlp_routes ${stringifyError(err)}`);
+    }
+}
+
+export const otlp = {
+    register: once(async (getRoutes: () => Promise<RouteConfig[]>) => {
+        await updateRoutes(getRoutes);
+        setInterval(async () => await updateRoutes(getRoutes), 10000);
```
TBonnin, line 58: went for `setInterval` since it is simpler.
A 10s lag between activating the feature and seeing the first traces exported is in my opinion reasonable

packages/logs/lib/otlp/otlpSpan.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+    if (operation.operation.type === 'proxy') {
+        return true;
+    }
+    if (operation.operation.type === 'action') {
+        return true;
+    }
+    if (operation.operation.type === 'webhook') {
+        return true;
+    }
```
TBonnin, line 107: we agreed with @bastienbeurier to only export a subset of the operation, the ones that are the most meaningful. Could be extended to more operations later

packages/logs/lib/otlp/otlpSpanProcessor.ts
```diff
@@ -0,0 +1,122 @@
...
+
+export interface RouteConfig {
+    routingId: string;
+    routingEndpoint: string;
+    routingHeaders: Record<string, string>;
+}
+
+interface BatchSpanProcessorWithRouteHash {
+    processor: BatchSpanProcessor;
+    routeHash: number;
```
TBonnin, line 13: hash is used to skip updating a route if its content hasn't changed

packages/logs/lib/otlp/otlpSpanProcessor.ts
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    onEnd(span: ReadableSpan): void {
+        try {
+            const processor = this.getProcessorForSpan(span);
+            if (processor) {
+                span.spanContext();
+                if (span.attributes[otlpRoutingAttributeKey]) {
+                    // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+                    delete span.attributes[otlpRoutingAttributeKey];
```
TBonnin, line 134: we don't want to expose the attribute we used internally for the routing to the customers

packages/logs/lib/otlp/otlp.ts
```diff
@@ -0,0 +1,53 @@
...
+    try {
+        const routes = await getRoutes();
+        registerRoutes(routes);
+    } catch (err) {
+        logger.error(`failed_to_update_otlp_routes ${stringifyError(err)}`);
+    }
+}
+
+export const otlp = {
+    register: once(async (getRoutes: () => Promise<RouteConfig[]>) => {
```
TBonnin, line 56: once is used to prevent executing register multiple times if called several times by accident

packages/types/lib/environment/db.ts
```diff
@@ -39,6 +39,7 @@ export interface DBEnvironment extends Timestamps {
     slack_notifications: boolean;
 
     webhook_receive_url?: string;
+    otlp_settings?: { endpoint: string; headers: Record<string, string> } | null;
```
bodinsamuel:
```suggestion
    otlp_settings: { endpoint: string; headers: Record<string, string> } | null;
```

packages/logs/lib/otlp/otlpSpanProcessor.ts
```diff
@@ -0,0 +1,116 @@
...
+
+export class RoutingSpanProcessor implements SpanProcessor {
+    private processors = new Map<string, BatchSpanProcessorWithRouteHash>();
+
+    constructor(routes: RouteConfig[]) {
+        try {
+            for (const route of routes) {
+                const processor = new BatchSpanProcessor(
+                    new OTLPTraceExporter({
+                        url: `${route.routingEndpoint.replace(/\/$/, '')}/traces`, // trim trailing slash if present
```
bodinsamuel: should be done on validation not here

packages/logs/lib/otlp/otlpSpanProcessor.ts
```diff
@@ -0,0 +1,116 @@
...
+                        headers: route.routingHeaders
+                    })
+                );
+                const routeHash = stringToHash(JSON.stringify(route));
```
bodinsamuel, line 56: you could store an `id:uuid` inside the object and forget about this hashing

packages/logs/lib/otlp/otlpSpanProcessor.ts
```diff
@@ -0,0 +1,116 @@
...
+                if (existing) {
+                    // Skip if the route hasn't changed
+                    if (existing.routeHash === routeHash) {
+                        continue;
+                    }
+                    toShutdown.push(existing.processor);
+                }
+
+                const traceExporter = new OTLPTraceExporter({
+                    url: `${route.routingEndpoint.replace(/\/$/, '')}/traces`,
```
bodinsamuel: same comment here

packages/logs/lib/otlp/otlpSpan.ts
```diff
@@ -0,0 +1,109 @@
...
+
+    end(state: MessageState): void {
+        if (this.span) {
+            this.span.setAttribute('nango.operation.status', state);
+            this.span.setStatus({ code: ['success', 'waiting', 'running'].includes(state) ? SpanStatusCode.OK : SpanStatusCode.ERROR });
+            this.span.end();
+        }
+    }
+
+    enrich(data: Partial<MessageRow>): void {
```
bodinsamuel, line 60: We need to make sure `enrich` always comes before send, in #2884 I fixed a bunch of those but might be worth checking again

bodinsamuel (approved):
Looks correct 👌🏻 
Not tested on my side

packages/logs/lib/otlp/otlpSpanProcessor.ts
TBonnin: I prefer this solution because even with a `version:uuid` in the otlp_settings a comparison will need to happen anyway to decide to update the version or not since we don't want the version to change if the values entered by the customer are the same. It then save an extra attributes in the settings

packages/logs/lib/otlp/otlpSpan.ts
TBonnin: yes I expect to tweak a few things based on the first feedback. Even the subset of attributes attaches to traces is arbitrary I might have missed some important info but it will be easier to add later on

TBonnin:
@bodinsamuel I made 2 extra commits:
- logging export error so we can debug more easily https://github.com/NangoHQ/nango/pull/2892/commits/4b4cbff56ce43eaa78968b9068f7c358cb0d588c
- fix a bug where span processor would not be remove when the settings are deleted

can you review them? thank you

bodinsamuel (approved):
Let's go 💯

PR #2893, opened 2024-10-25
feat(integrations): add support for keeper

hassan254-prog:
## Describe your changes

- Add support for Keeper

[EXT-202](https://linear.app/nango/issue/EXT-202/add-support-for-keeper)


packages/shared/providers.yaml
```diff
@@ -2891,6 +2891,34 @@ keap:
...
 
+keeper-scim:
+    display_name: Keeper(SCIM)
+    categories:
+        - productivity
+    auth_mode: API_KEY
+    proxy:
+        headers:
+            Authorization: Bearer ${apiKey}
+        base_url: https://keepersecurity.com/api/rest/scim/v2/${connectionConfig.node}
```
bodinsamuel, line 2902: I thought we were trying to not put version in base_url?

bodinsamuel (approved):
🐝

packages/shared/providers.yaml
hassan254-prog: I was trying to avoid that, but we might be unable to get `connectionConfig.node`, which is crucial for requests.

hassan254-prog: @bodinsamuel , would it be better to define the API version in the path `/v2/${connectionConfig.node}` or should we let the user specify the version directly within `connectionConfig`?

bodinsamuel: That's a tough question indeed, we need to think about it

PR #2894, opened 2024-10-25
fix(validations): add provider name and remove credentials and connection_config from interpolation checks

hassan254-prog:
## Describe your changes

- Fixes validation errors for #2893  by adding a provider name for better debugging and also removing `connection_config` and `credentials` from interpolation errors check


bodinsamuel (approved):
It was wrongly matching the regexes I suppose?

hassan254-prog:
> It was wrongly matching the regexes I suppose?

Yes, it was matching for `connection_config` and `credentials`, which would fail if we have `pattern: '^\d{15}$'`

PR #2897, opened 2024-10-25
feat(ui): list connections refactor, show end user profile

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1812/ui-for-userorganization-info-connections-screen

## Major

- Refactor the connections list page
Show end user profile, add avatar, search (connection id, profile display name), filter by integration, filter by errors

- Move endpoint to file `GET /api/v1/connections`
- Move endpoint to file `GET /connections`

## Minor

- Add new endpoint `/connections/count` 
It was used to display the error mark, since the main endpoints has pagination it was no longer good, it's also a bit more efficient now.

- Add Connection button now uses Connect UI by default
We still have the fallback with the old version for a few weeks

- Allow Connect UI for all env
Note the fix for self hosting is in a next PR

https://github.com/user-attachments/assets/02e97ad8-ed3d-4d92-9f16-ef8bf61978d3

--- 

New query seems acceptable (results in prod replica):

```sql
EXPLAIN SELECT
	*
FROM (
	SELECT
		row_to_json(_nango_connections.*) AS connection,
		row_to_json(end_users.*) AS end_user,
		(
			SELECT
				COALESCE(json_agg(json_build_object('type', TYPE, 'log_id', log_id)), '[]'::json)
			FROM
				_nango_active_logs
			WHERE
				_nango_connections.id = _nango_active_logs.connection_id
				AND _nango_active_logs.active = TRUE) AS active_logs, "_nango_configs"."provider"
		FROM
			"_nango_connections"
			INNER JOIN "_nango_configs" ON "_nango_connections"."config_id" = "_nango_configs"."id"
			LEFT JOIN "end_users" ON "end_users"."id" = "_nango_connections"."end_user_id"
		WHERE
			"_nango_connections"."environment_id" = 18
			AND "_nango_connections"."deleted" = FALSE
			and(connection_id ILIKE '%samuel%'
				OR end_users.display_name ILIKE '%samuel%'
				OR end_users.email ILIKE '%samuel%')
		ORDER BY
			"_nango_connections"."created_at" DESC
		LIMIT 20) AS "rows"
```

```sql
Subquery Scan on rows  (cost=44.07..46.75 rows=1 width=104)
  ->  Limit  (cost=44.07..46.74 rows=1 width=112)
        ->  Result  (cost=44.07..46.74 rows=1 width=112)
              ->  Sort  (cost=44.07..44.08 rows=1 width=84)
                    Sort Key: _nango_connections.created_at DESC
                    ->  Nested Loop  (cost=11.99..44.06 rows=1 width=84)
                          ->  Hash Left Join  (cost=11.71..41.80 rows=1 width=1997)
                                Hash Cond: (_nango_connections.end_user_id = end_users.id)
"                                Filter: (((_nango_connections.connection_id)::text ~~* '%samuel%'::text) OR ((end_users.display_name)::text ~~* '%samuel%'::text) OR ((end_users.email)::text ~~* '%samuel%'::text))"
                                ->  Bitmap Heap Scan on _nango_connections  (cost=2.83..32.85 rows=27 width=899)
                                      Recheck Cond: ((environment_id = 18) AND (NOT deleted))
                                      ->  Bitmap Index Scan on idx_connections_envid_connectionid_provider_where_deleted  (cost=0.00..2.82 rows=27 width=0)
                                            Index Cond: (environment_id = 18)
                                ->  Hash  (cost=6.17..6.17 rows=217 width=1177)
                                      ->  Seq Scan on end_users  (cost=0.00..6.17 rows=217 width=1177)
                          ->  Index Scan using _nango_configs_pkey on _nango_configs  (cost=0.29..2.26 rows=1 width=12)
                                Index Cond: (id = _nango_connections.config_id)
              SubPlan 1
                ->  Aggregate  (cost=2.63..2.64 rows=1 width=32)
                      ->  Index Scan using idx_activelogs_connectionid_where_active on _nango_active_logs  (cost=0.41..2.63 rows=1 width=25)
                            Index Cond: (connection_id = _nango_connections.id)
```


packages/shared/lib/services/config.service.ts
```diff
@@ -87,17 +87,25 @@ class ConfigService {
...
     async listProviderConfigsByProvider(environment_id: number, provider: string): Promise<ProviderConfig[]> {
         return (await db.knex.select('*').from<ProviderConfig>(`_nango_configs`).where({ environment_id, provider, deleted: false }))
             .map((config) => encryptionManager.decryptProviderConfig(config))
             .filter((config) => config != null) as ProviderConfig[];
     }
 
-    async getAllNames(environment_id: number): Promise<string[]> {
-        const configs = await this.listProviderConfigs(environment_id);
-        return configs.map((config) => config.unique_key);
-    }
```
bodinsamuel, line 99: Dead code

packages/webapp/src/components/MultiSelect.tsx
```diff
@@ -50,7 +50,7 @@ export const MultiSelect: React.FC<MultiSelectArgs<any>> = ({ label, options, se
     return (
         <Popover open={open} onOpenChange={setOpen}>
             <PopoverTrigger asChild>
-                <Button variant="zombieGray" size={'xs'} className={cn('text-text-light-gray', isDirty && 'text-white')}>
+                <Button variant="zombieGray" size={'sm'} className={cn('text-text-light-gray', isDirty && 'text-white')}>
```
bodinsamuel, line 53: Size were incorrect

packages/shared/lib/services/connection.service.ts
```diff
@@ -980,32 +1015,42 @@ class ConnectionService {
...
+            .orderBy('_nango_connections.created_at', 'desc');
 
-        if (connectionId) {
-            queryBuilder.where({
-                connection_id: connectionId
+        if (search) {
+            subQuery.where(function () {
+                this.whereRaw('connection_id ILIKE ?', `%${search}%`)
+                    .orWhereRaw('end_users.display_name ILIKE ?', `%${search}%`)
+                    .orWhereRaw('end_users.email ILIKE ?', `%${search}%`);
```
bodinsamuel, line 1033: This part is not going to scale forever but looks okay for now

packages/shared/lib/services/connection.service.ts
```diff
@@ -980,32 +1015,42 @@ class ConnectionService {
...
+        if (withError === false) {
+            query.whereRaw("rows.active_logs::jsonb = '[]'");
+        } else if (withError === true) {
+            query.whereRaw("rows.active_logs::jsonb <> '[]'");
+        }
 
-    public async getAllNames(environment_id: number): Promise<string[]> {
-        const connections = await this.listConnections(environment_id);
-        return [...new Set(connections.map((config) => config.connection_id))];
+        return await query;
```
bodinsamuel, line 1053: Check PR desc for query plan

packages/shared/lib/services/config.service.ts
```diff
@@ -87,17 +87,25 @@ class ConfigService {
             .filter(Boolean) as ProviderConfig[];
     }
 
+    async listIntegrationForApi(environment_id: number): Promise<(ProviderConfig & { connection_count: string })[]> {
+        const res = await db.knex
+            .select<
+                (ProviderConfig & { connection_count: string })[]
+            >('*', db.knex.raw(`(SELECT COUNT(*) FROM _nango_connections WHERE _nango_connections.environment_id = _nango_configs.environment_id AND _nango_connections.config_id = _nango_configs.id AND deleted = false) as connection_count`))
```
TBonnin: what about extracting the subquery so it is easier to read:
```
    const connectionCountSubquery = db.knex('_nango_connections')
        .count('* as count')
        .where({
            environment_id: db.knex.ref('_nango_configs.environment_id'),
            config_id: db.knex.ref('_nango_configs.id'),
            deleted: false
        })
        .as('connection_count');

...
.select(
            '_nango_configs.*',
            connectionCountSubquery
        )
...
```

packages/shared/lib/services/connection.service.ts
```diff
@@ -957,20 +959,53 @@ class ConnectionService {
...
+                '_nango_connections.environment_id': environmentId,
+                '_nango_connections.deleted': false
+            })
+            .first();
+        const res = await query;
+        if (!res) {
+            return { total: 0, withError: 0 };
+        }
+
+        return { total: Number(res.total_connection), withError: Number(res.with_error) };
```
TBonnin: should this function return a Result instead of 0s when there is an error?

packages/shared/lib/services/connection.service.ts
```diff
@@ -957,20 +959,53 @@ class ConnectionService {
...
-    ): Promise<
-        { id: number; connection_id: string; provider: string; created: string; metadata: Metadata; active_logs: [{ type: string; log_id: string }] }[]
-    > {
-        const queryBuilder = db.knex
+    public async count({ environmentId }: { environmentId: number }): Promise<{ total: number; withError: number }> {
+        const query = db.knex
+            .from(`_nango_connections`)
+            .select<{ total_connection: string; with_error: string }>(
+                db.knex.raw('COUNT(_nango_connections.*) as total_connection'),
+                db.knex.raw('COUNT(_nango_connections.*) FILTER (WHERE _nango_active_logs.type IS NOT NULL) as with_error')
```
TBonnin: TIL I learn about pg FILTER :)

TBonnin: should `with_error` actually be called `with_auth_error` since we are not accounting for the sync errors?

packages/shared/lib/services/connection.service.ts
```diff
@@ -980,32 +1015,42 @@ class ConnectionService {
                     WHERE _nango_connections.id = ${ACTIVE_LOG_TABLE}.connection_id
                       AND ${ACTIVE_LOG_TABLE}.active = true
                   ) as active_logs
-               `)
+               `),
```
TBonnin, line 1018: do we need this subquery if `withError` is false?

packages/webapp/src/utils/avatar.ts
```diff
@@ -0,0 +1,22 @@
+export function toAcronym(name: string): string {
+    const clean = name
+        // remove all tag-like sequences: [Shared][Prod]
+        .replace(/\[[^\]]*\]/g, '')
+        // replace noisy chars with spaces
+        .replace(/[/._&:-]/g, ' ')
```
TBonnin: what about `#:;#@|[]`?

packages/webapp/src/utils/avatar.ts
```diff
@@ -0,0 +1,22 @@
...
+        return firstLetters.slice(0, 2).toUpperCase();
+    }
+
+    // Rollback to something generic if nothing is left
+    return (
+        name
+            .replace(/[^a-zA-Z]/g, '')
+            .slice(0, 2)
+            .toUpperCase() || name.substring(0, 2)
+    );
```
TBonnin, line 21: not sure this function will cover every single scenario but we'll see

TBonnin (approved):

packages/webapp/src/utils/avatar.ts
bodinsamuel: good point

packages/webapp/src/utils/avatar.ts
bodinsamuel: yeah it's not meant to be perfect for sure

packages/shared/lib/services/connection.service.ts
bodinsamuel: good point

packages/shared/lib/services/connection.service.ts
bodinsamuel: imo this count is not a critical feature, if it fails I don't want to cut the rest of the API call. wdyt?

packages/shared/lib/services/connection.service.ts
bodinsamuel: withError is a filter, not a field selector, so I need the full list to post-filter only rows without active_logs. (not sure if I'm clear ^^)

packages/shared/lib/services/connection.service.ts
TBonnin: The result can be handled so `Err` is not blocking but at least this function semantic will be stronger. For example differentiating an error to having zero conns

packages/shared/lib/services/connection.service.ts
TBonnin: got it. I misread the rest of the query. I thought it was getting all conns OR all conns with errors but it is instead all conns without error OR all conns with errors

TBonnin (approved):

PR #2899, opened 2024-10-25
feat(integrations): add support for Dixa

hassan254-prog:
## Describe your changes

- Add support for Dixa

[EXT-163](https://linear.app/nango/issue/EXT-163/add-support-for-dixa)


packages/shared/providers.yaml
```diff
@@ -1523,6 +1523,29 @@ discourse:
...
 
+dixa:
+    display_name: Dixa
+    categories:
+        - support
+    auth_mode: API_KEY
+    proxy:
+        base_url: https://${connectionConfig.apiType}.dixa.io
+        headers:
+            Authorization: ${apiKey}
```
bodinsamuel, line 1725: Can be tested maybe

hassan254-prog: Should we enforce verification on untested integrations on our end for both [Dixa](https://github.com/NangoHQ/nango/pull/2899) and [Chattermill](https://github.com/NangoHQ/nango/pull/2900). Could this be a case of shooting ourselves in the foot 🤔?

bodinsamuel: I mean when you publish an integration I have no way to know if it's tested or not, if you don't feel confident (that's okay) you should write it in the PR desc ☺️

bodinsamuel (approved):

packages/shared/providers.yaml
hassan254-prog: I am sorry for the miscommunication here🤦‍♂️. Thank you for the clarification 🙏.

PR #2909, opened 2024-10-28
feat(connect): allow optional params, cosmetic fixes

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1947/allow-some-fields-to-be-optional

- Allow optional params
- Fix: do not display back on a single integration scenario
- Fix: move help at the bottom of the screen
- Fix: change button wording when error in credentials

### DIFF
--- a/packages/connect-ui/src/lib/utils.ts
+++ b/packages/connect-ui/src/lib/utils.ts
@@ -5,17 +5,29 @@ import { z } from 'zod';
 import type { SimplifiedJSONSchema } from '@nangohq/types';
 
 import type { ClassValue } from 'clsx';
+import type { ZodTypeAny } from 'zod';
 
 export function cn(...inputs: ClassValue[]) {
     return twMerge(clsx(inputs));
 }
 
-export function jsonSchemaToZod(schema: SimplifiedJSONSchema): z.ZodString {
-    let field = z.string();
-    if (schema.format === 'hostname') field = field.regex(/^[a-zA-Z0-9-]$/);
-    else if (schema.format === 'uuid') field = field.uuid();
-    else if (schema.format === 'uri') field = field.url();
-    if (schema.pattern) field = field.regex(new RegExp(schema.pattern), { message: `Incorrect ${schema.title}` });
+export function jsonSchemaToZod(schema: SimplifiedJSONSchema): ZodTypeAny {
+    let fieldString = z.string();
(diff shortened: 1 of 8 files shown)


packages/connect-ui/src/lib/utils.ts
```diff
@@ -5,17 +5,29 @@ import { z } from 'zod';
...
+        fieldString = fieldString.url();
+    }
+    if (schema.pattern) {
+        fieldString = fieldString.regex(new RegExp(schema.pattern), { message: `Incorrect ${schema.title}` });
+    }
 
-    return field;
+    if ('optional' in schema && schema.optional === true) {
+        // https://stackoverflow.com/questions/73715295/react-hook-form-with-zod-resolver-optional-field
+        return z.union([fieldString, z.literal('')]);
```
TBonnin, line 29: why not `.optional`?

packages/connect-ui/src/lib/utils.ts
```diff
@@ -5,17 +5,29 @@ import { z } from 'zod';
...
+    if (schema.format === 'hostname') {
+        fieldString = fieldString.regex(/^[a-zA-Z0-9-]$/);
+    } else if (schema.format === 'uuid') {
+        fieldString = fieldString.uuid();
+    } else if (schema.format === 'uri') {
+        fieldString = fieldString.url();
+    }
+    if (schema.pattern) {
+        fieldString = fieldString.regex(new RegExp(schema.pattern), { message: `Incorrect ${schema.title}` });
+    }
```
TBonnin, line 25: can we have a pattern with a known format? Should it be `else`?

TBonnin (approved):

packages/connect-ui/src/lib/utils.ts
bodinsamuel: it doesn't work, hence the link

packages/connect-ui/src/lib/utils.ts
bodinsamuel: no, in the case of URI for example: URI is not enforcing https which we with the pattern

packages/connect-ui/src/lib/utils.ts
TBonnin: by bad. I had missed the link 🙈

PR #2910, opened 2024-10-28
feat(shared): Dynamic reload for providers

nalanj:
## Describe your changes

Adds the ability to reload providers dynamically. This is handled by a new function, `launchProvidersSync` that uses `setTimeout` in the background to monitor for updates to the providers file.

To utilize a provider's file over http, set the `PROVIDERS_URL` environment variable. Optionally set `PROVIDERS_RELOAD_INTERVAL` to a number of milliseconds to wait between reload attempts.

- [x] Add launchProvidersSync to services that need it
- [x] Log when providers data updates so that it's clear
- [x] Add ability to load providers over HTTP
- [x] Make the reload frequency configurable

### How to review it

1. Set up a github gist with the content of `providers.yaml`.
2. Add `/raw` to the end of the gist URL to get the raw URL for it. Set that value to `PROVIDERS_URL` in `.env`.
3. Run the app: `npm run dwa`
4. Update the gist. You'll see a log message within the reload frequency time showing that it's been updated.
5. Run a runner script that takes longer than the reload frequency and update the gist while it's running to see a log message saying it has been updated.

### Open Questions

- What do we want the default reload frequency to be? I currently have it set to 30 seconds.
- Is it dangerous to just replace the providers while the app is running? Based on cusory code review it doesn't seem to be, but it'd be helpful to have more experienced feedback to that problem.

### Deployment Notes

- Do we want to simply use the `providers.yaml` from GitHub or do we want to host a specific version somewhere?
- We'll need to set up an ENV var for `PROVIDERS_URL`. If that env var isn't set, the app will work like it did before.

### To Clean Up

- Right now I'm putting a param onto the url since I'm hosting with a gist and need to be able to cache bust. We may want to keep this if we go with the `providers.yaml` out of the repo. I did notice with a gist that it flapped between versions at least once - something to watch for with however we host the file.

### Things I thought about and didn't do

- It doesn't feel like reloads from disk are important to this project since our main goal is to allow for reloads without restarts, and upgrading the default `providers.yaml` would probably happen along with a package update.
- It also doesn't feel like allowing someone to configure a disk path for `providers.yaml` is that important to the current project.
- I didn't write any tests because there were no tests for these files to start with and they'd be better as integration tests but are still pretty likely to be flaky. More than happy to dig into testing this more, though.

### Things I noticed

- I had to run `npm run ts-build` to get `shared` to rebuild. `npm run dwa` didn't seem to pick up code changes there.
- The code feels like we intend to just load `providers.yaml` from the module by default. Would it make sense to remove the file resolution code in `getProvidersPath` and just try to load `providers.yaml` from the default location?


bodinsamuel (commented):
Very quick 🚀 
Seems to work. It's not up to you but I'm not convinced about the whole task tbf.
Ideally we don't have to do that, but I would at least replace the setInterval by a HTTP call, so we can remove the setInterval. 

Also having the URL as a env var means we have to restart the server to set it, defeating the purpose :D I have propositions but since it's trial I'll keep those, happy to discuss today with @TBonnin

nalanj:
Thanks for the comments, @bodinsamuel!

> It's not up to you but I'm not convinced about the whole task tbf.

I can't help you there!

When I was looking at the problem from the outside, my thought was that moving the orchestrator to more of a mark and sweep GC model, where on deploy runners are flagged for expiration and a new runner is brought up on next job, with the expiring runners being allowed to finish their current job(s). Thomas mentioned the P90 for a job is like 7 seconds so I think you'd end up with most expired really fast. But that said, that wasn't the project.

> Ideally we don't have to do that, but I would at least replace the setInterval by a HTTP call, so we can remove the setInterval.

When digging into this, the main decision seemed to be does `providers.yaml` update by push, pull, or some hybrid. It was pretty quickly obvious that pull was the simplest solution, pure push wasn't really feasible due to needing the option to allow self-hosted installations to update automatically, and hybrid would be feasible but much more complex, so I chose the simplest path. More than happy to rewrite it though.

> 
> Also having the URL as a env var means we have to restart the server to set it, defeating the purpose :D I have propositions but since it's trial I'll keep those, happy to discuss today with @TBonnin

That's certainly true if you were to update the URL frequently, but I was assuming we'd have the ability to host the `providers.yaml` on a fixed URL that would seldom change. Is there some constraint in the Nango infra that makes it impossible to host dynamic content over http?

packages/shared/lib/services/providers.ts
```diff
@@ -22,49 +26,96 @@ export function getProvider(providerName: string): Provider | null {
...
+            return findProvidersYaml(parentDir);
         }
-        const parentDir = path.dirname(dir);
-        if (parentDir === dir) {
-            throw new NangoError('providers_yaml_not_found');
-        }
-        return findProvidersYaml(parentDir);
     };
     return findProvidersYaml(dirname());
 }
```
TBonnin, line 73: do you mind giving a quick explanation why this function is now async?

TBonnin:
> Thanks for the comments, @bodinsamuel!
> 
> > It's not up to you but I'm not convinced about the whole task tbf.
> 
> I can't help you there!
> 
> When I was looking at the problem from the outside, my thought was that moving the orchestrator to more of a mark and sweep GC model, where on deploy runners are flagged for expiration and a new runner is brought up on next job, with the expiring runners being allowed to finish their current job(s). Thomas mentioned the P90 for a job is like 7 seconds so I think you'd end up with most expired really fast. But that said, that wasn't the project.
> 
> > Ideally we don't have to do that, but I would at least replace the setInterval by a HTTP call, so we can remove the setInterval.
> 
> When digging into this, the main decision seemed to be does `providers.yaml` update by push, pull, or some hybrid. It was pretty quickly obvious that pull was the simplest solution, pure push wasn't really feasible due to needing the option to allow self-hosted installations to update automatically, and hybrid would be feasible but much more complex, so I chose the simplest path. More than happy to rewrite it though.
> 
> > Also having the URL as a env var means we have to restart the server to set it, defeating the purpose :D I have propositions but since it's trial I'll keep those, happy to discuss today with @TBonnin
> 
> That's certainly true if you were to update the URL frequently, but I was assuming we'd have the ability to host the `providers.yaml` on a fixed URL that would seldom change. Is there some constraint in the Nango infra that makes it impossible to host dynamic content over http?

TBonnin:
> Also having the URL as a env var means we have to restart the server to set it, defeating the purpose :D I have propositions but since it's trial I'll keep those, happy to discuss today with @TBonnin

The env var purpose is not to be able to change the url but act as a way to enable or not the feature for self-hosted. For enterprise customers, the need for such feature is not high in my opinion because providers churn is way lower and because they control everything about the infra and operation. It is then much easier (and also rarer) to deploy new version. 
With the env var they can decide not to use this dynamic loading feature if they don't need it or if they don't want to reach an external resource for security reason.

TBonnin:
my bad. I misclicked the close button

TBonnin:
> > It's not up to you but I'm not convinced about the whole task tbf.
> 
> I can't help you there!
> 
> When I was looking at the problem from the outside, my thought was that moving the orchestrator to more of a mark and sweep GC model, where on deploy runners are flagged for expiration and a new runner is brought up on next job, with the expiring runners being allowed to finish their current job(s). Thomas mentioned the P90 for a job is like 7 seconds so I think you'd end up with most expired really fast. But that said, that wasn't the project.

@bodinsamuel This task objective was definitely not to solve the `deployment kills scripts` issue but to make providers.yaml changes not require a full deploy. Happy to discuss though.

What you propose @nalanj is definitely going in the right direction but as always the devil is in the details and I think such project might be a bit too wide for a trial week. I am happy to pick you brain about it though and discuss about potential solutions

bodinsamuel (approved):
Alright, discussed with Thomas and Alan. 
We agreed about the path forward, let's go with this solution. If we can just increase the interval time, the rest is 👌🏻 
Maybe if we want to expose this file in a S3 buckets, we need a github actions to upload it. Otherwise we serve it from `server` (e.g: app.nango.dev/providers.yaml`)

packages/shared/lib/services/providers.ts
```diff
@@ -22,49 +26,96 @@ export function getProvider(providerName: string): Provider | null {
...
+            rawFile = await response.text();
+        } else {
+            throw new NangoError('provider_template_loading_fetch_failed');
         }
+    }
 
-        for (const key in fileEntries) {
-            const entry = fileEntries[key];
+    return rawFile;
+}
```
TBonnin, line 96: should we back up to load it from file if it fails to fetch the url or fails to load it?

TBonnin (approved):

packages/runner/lib/app.ts
```diff
@@ -2,6 +2,9 @@ import './tracer.js';
 import { server } from './server.js';
 import { stringifyError } from '@nangohq/utils';
 import { logger } from './utils.js';
+import { launchProvidersSync } from '@nangohq/shared';
+
+await launchProvidersSync();
```
khaliqgant: `sync` is a loaded word within the context of Nango so I would advise another name for this method. Maybe something with `monitor`?

packages/shared/lib/services/providers.ts
nalanj: Sure thing. I originally was thinking the path would be configurable as well, but we backed out of that, and this stuck around afterward. That said, since this code is called through an asynchronous function `launchProviderSync` I still think it makes sense to stay async.

packages/shared/lib/services/providers.ts
nalanj: First, I'm open to either way. 

My take on why not to back up to loading it from disk is that it will give a false positive on loading over URL (i.e. it'll be hard to notice that you're just not loading it from the right place), and then the app will look like it's working correctly when it's not.

packages/runner/lib/app.ts
nalanj: I like monitor. Incoming!

nalanj: fixed in [fd34a14](https://github.com/NangoHQ/nango/pull/2910/commits/fd34a143ae87f6faccb1e7939faa046f59bbb2b9)

khaliqgant (approved):

PR #2912, opened 2024-10-28
fix(connect): document connect ui, add apiURL param

bodinsamuel:
## Describe your changes

Contributes https://linear.app/nango/issue/NAN-1923/make-nango-connect-the-only-way-to-create-connection-in-nango-ui

- Document connect ui as the main way to consume Nango
The old doc has been moved to advanced.
- Add ApiURL param so people can self-host


docs-v2/reference/sdks/frontend.mdx
```diff
@@ -100,7 +136,7 @@ const result = nango.auth('<INTEGRATION-ID>', '<CONNECTION-ID>', {
     The integration ID that you can find in the integration settings on the Nango UI.
   </ResponseField>
 
-  <ResponseField name="connectionId" type="string" required>
+  <ResponseField name="connectionId" type="string">
```
TBonnin, line 139: I think we need to also update the `docs-v2/spec.yaml` file to make the connection_id field not required for the POST /connection endpoint

docs-v2/reference/sdks/frontend.mdx
```diff
@@ -180,6 +216,11 @@ const result = nango.auth('<INTEGRATION-ID>', '<CONNECTION-ID>', {
   </ResponseField>
 </Expandable>
 
+By default we are auto generating the `connectionId` which is the unique identifier of this connection but ou can also specify it like this:
+```js
+const result = await nango.auth('<INTEGRATION-ID>', '<CONNECTION-ID>');
+```
+
```
TBonnin, line 223: do we even document this option or not?

TBonnin (approved):

docs-v2/reference/sdks/frontend.mdx
bodinsamuel: For legacy reason, I think it's somewhat required

PR #2914, opened 2024-10-28
fix(connect): serve in docker image

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1838/serve-ui-in-self-hosted-image

- Optionally serve Nango Connect directly from the docker image
It's still the old one but the entrypoint will be compatible with the new image

- Add documentation

## How to test

- Kill your local dev docker 

```diff
diff --git a/docker-compose.yaml b/docker-compose.yaml
index 969c721bb..5605a31f1 100644
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -15,9 +15,11 @@ services:
             - nango

     nango-server:
+        build:
+            context: .
+            dockerfile: packages/server/Dockerfile
         image: nangohq/nango-server:hosted
         container_name: nango-server
-        platform: linux/amd64
         environment:
             - NANGO_ENCRYPTION_KEY=${NANGO_ENCRYPTION_KEY}
             - NANGO_DB_USER=${NANGO_DB_USER}
```
```
docker compose up
```


TBonnin (commented):
lacking some context but can the container serve both apps?

bodinsamuel:
it's not yet ready, should work though but the testing is tedious. 
A docker image can serve two ports yes

TBonnin:
> it's not yet ready, should work though but the testing is tedious. A docker image can serve two ports yes

I am definitely lacking context. If it is not ready why even providing a way to serve connect-ui in self-hosted docker container?

bodinsamuel:
> The PR is not ready sorry :D 

Sorry I thought it was ready for review. let's make it a draft then

packages/connect-ui/package.json
```diff
@@ -6,7 +6,11 @@
     "scripts": {
         "dev": "vite",
         "build": "tsc -b && vite build",
-        "preview": "vite preview"
+        "preview": "vite preview",
+        "prod:unsafe": "serve dist -p ${NANGO_CONNECT_UI_PORT:-3009}"
```
TBonnin: `prod` and `unsafe` might be a bit antinomic. what about `run:unsafe`

packages/server/entrypoint.sh
```diff
@@ -0,0 +1,26 @@
...
+  node "$dir/packages/server/dist/server.js" &
+
+  # This is not recommended, you should server UI from a dedicated static website hosting
+  npm run -w @nangohq/connect-ui prod:unsafe &
+
+  # Wait for any process to exit
+  wait -n
+
+  # Exit with status of process that exited first
+  exit $?
```
TBonnin, line 22: not sure the background processes are being killed when the script exits. We might want to trap signals and ensure proper cleanup

packages/server/entrypoint.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
+
+set -ex
+
+dir=$(pwd)
+
+echo "prout $dir "
```
TBonnin: 💨 🤣

bodinsamuel: 🫡

packages/server/entrypoint.sh
bodinsamuel: that's the recommended way to run multi process in container. I don't know enough about bash to know if it's good enough
https://docs.docker.com/engine/containers/multi-service_container/

TBonnin: it is indeed running in a container so it is maybe an overkill. A less naive approach could be something like this:
```
...
# trap signals and kills processes
trap 'kill $server_pid $ui_pid; exit' SIGINT SIGTERM

# wait for processes to exit
wait -n
exit_code=$?

# clean up processes
kill $server_pid $ui_pid 2>/dev/null || true
exit $exit_code
```

bodinsamuel: not sure how it behaves with multi-process, I'll merge as-is and if get complain I'll investigate

bodinsamuel:
@TBonnin Can I have a last review on this please? ☺️

docs-v2/host/self-host/self-hosting-instructions.mdx
```diff
@@ -27,6 +27,25 @@ The resulting callback URL for OAuth will be `<INSTANCE-URL>/oauth/callback`.
...
 
+## Using Nango Connect
+
+Nango Connect is an aditionnal UI that you can your customers can use to easily connect to integration. [See related documentation](/integrate/guides/authorize-an-api#authorize-users-from-your-app).
+This UI is directly available for self-hosted in the main docker image.
+
+```sh
+FLAG_SERVE_CONNECT_UI=true
+NANGO_CONNECT_UI_PORT=3009
+NANGO_PUBLIC_CONNECT_URL=
```
TBonnin: no value here?

packages/server/entrypoint.sh
```diff
@@ -0,0 +1,25 @@
...
+
+echo "$dir/packages/server/dist/server.js"
+
+# https://docs.docker.com/engine/containers/multi-service_container/
+
+# connect ui
+if [ "$FLAG_SERVE_CONNECT_UI" == "true" ]; then
+  node "$dir/packages/server/dist/server.js" &
+
+  # This is not recommended, you should server UI from a dedicated static website hosting
```
TBonnin:
```suggestion
  # This is not recommended, you should serve UI from a dedicated static website hosting
```

TBonnin: or even maybe `... serve Connect UI ...`

TBonnin (approved):

PR #2916, opened 2024-10-29
feat(webapp): Syncs Table UI

nalanj:
## Describe your changes

Based on the [figma drawing](https://www.figma.com/design/hKL3cmI9vBwDoZyRahQnsq/Connections-Tab?node-id=4149-3559&t=6zV7ZuPy1sVIHo4F-4), this PR updates the Syncs table to clean up styles and make it more nicely spaced.

Because the counts PR is still in flight, this doesn't add the counts column yet, but don't worry, it's coming!

Here's how it came out:

Screenshot 2024-10-29 at 3 22 18 PM

image

Screenshot 2024-10-29 at 3 22 22 PM

A few things to note:

- The duration of the last execution is now in a tooltip under the Last Execution column.
- The function that is used to generate the text for the frequency column now takes our special intervals into account so that they'll be rendered as more concise frequencies.
- The last execution status is more clear because it no longer has a mixed meaning based on the paused status of the row.
- The paused status is now handled in the Next Sync Start column.

## Thing we maybe want to discuss

- I wrote a wrapper for the `Tooltip` component called `SimpleTooltip` that made it more concise to write up a tooltip. We probably want to name it something different.


bastienbeurier:
This looks great @nalanj! Thank you for the detailed description in the PR. I'll let @bodinsamuel review the PR.

packages/webapp/src/components/ui/Tooltip.tsx
```diff
@@ -23,4 +23,23 @@ const TooltipContent = React.forwardRef<React.ElementRef<typeof TooltipPrimitive
...
+
+    return (
+        <TooltipProvider>
+            <Tooltip>
+                <TooltipContent>{props.tooltipContent}</TooltipContent>
+                <TooltipTrigger>{props.children}</TooltipTrigger>
+            </Tooltip>
+        </TooltipProvider>
+    );
+}
```
bodinsamuel: The component inside `ui` are copy pasted from Shadcn, so I would put this component outside this folder

packages/webapp/src/components/ui/Tooltip.tsx
```diff
@@ -23,4 +23,23 @@ const TooltipContent = React.forwardRef<React.ElementRef<typeof TooltipPrimitive
 );
 TooltipContent.displayName = TooltipPrimitive.Content.displayName;
 
-export { Tooltip, TooltipTrigger, TooltipContent, TooltipProvider };
+interface SimpleTooltipProps {
+    tooltipContent: React.ReactElement | string | undefined;
+}
+
+function SimpleTooltip(props: React.PropsWithChildren<SimpleTooltipProps>) {
```
bodinsamuel: We currently use React.FC syntax, even if it doesn't bring a lot at least it's consistent

packages/webapp/src/components/ui/Tooltip.tsx
```diff
@@ -23,4 +23,23 @@ const TooltipContent = React.forwardRef<React.ElementRef<typeof TooltipPrimitive
 );
 TooltipContent.displayName = TooltipPrimitive.Content.displayName;
 
-export { Tooltip, TooltipTrigger, TooltipContent, TooltipProvider };
+interface SimpleTooltipProps {
+    tooltipContent: React.ReactElement | string | undefined;
```
bodinsamuel: I think React.ReactNode is covering all of this

packages/webapp/src/pages/Connection/components/SyncRow.tsx
```diff
@@ -302,8 +264,8 @@ export const SyncRow: React.FC<{ sync: SyncResponse; connection: Connection; pro
                                 )}
 
                                 <Link to={logUrl} className="w-full">
-                                    <Button variant="zombie" className="w-full gap-4">
-                                        <QueueListIcon className="flex h-6 w-6 text-gray-400 cursor-pointer" />
+                                    <Button variant="zombie" className="w-full gap-4 rounded hover:bg-black text-gray-400">
```
bodinsamuel: Ideally button don't take style in className, maybe it's missing a variant for dropDown

packages/webapp/src/pages/Connection/components/SyncRow.tsx
```diff
@@ -104,50 +97,52 @@ export const SyncRow: React.FC<{ sync: SyncResponse; connection: Connection; pro
...
+            <Table.Cell bordered className="text-center">
+                <SimpleTooltip
+                    tooltipContent={
+                        sync.latest_sync?.result && Object.keys(sync.latest_sync?.result).length > 0 ? (
+                            <pre>{parseLatestSyncResult(sync.latest_sync?.result, sync.latest_sync?.models)}</pre>
+                        ) : undefined
+                    }
+                >
+                    <Link to={logUrl}>{formatDateToUSFormat(sync.latest_sync?.updated_at)}</Link>
+                </SimpleTooltip>
```
bodinsamuel, line 146: Could be aligned left
Screenshot 2024-10-30 at 10 57 49

bodinsamuel (commented):
What I could test looks good 👌🏻 

I got this error
```ts
SyncRow.tsx:107 Uncaught TypeError: Cannot read properties of null (reading 'status')
    at SyncRow (SyncRow.tsx:107:1)
```
It happens when there are multiple syncs, and some never than never run. I think the type is incorrect

packages/webapp/src/components/ui/Tooltip.tsx
nalanj: When I peeked at the types for ReactNode I just loved seeing this:

ReactNode types

nalanj: When I peeked at those types, I loved this type signature:

image

packages/webapp/src/components/ui/Tooltip.tsx
nalanj: fixed in [a34a541](https://github.com/NangoHQ/nango/pull/2916/commits/a34a541f8391d1fb725991361f7cb79ddcada95c)

packages/webapp/src/components/ui/Tooltip.tsx
nalanj: fixed in [a1534cf](https://github.com/NangoHQ/nango/pull/2916/commits/a1534cfae116c4ba8b84900424cc8a51b8225ae0)

packages/webapp/src/components/ui/Tooltip.tsx
nalanj: fixed in [a34a541](https://github.com/NangoHQ/nango/pull/2916/commits/a34a541f8391d1fb725991361f7cb79ddcada95c)

TBonnin (approved):

nalanj:
@bodinsamuel what do we think of just having a tag that says `NEVER RUN` when a connection hasn't run yet, to handle that null situation?

image

packages/webapp/src/pages/Connection/components/SyncRow.tsx
nalanj: Fixed in 177e87dc

packages/webapp/src/pages/Connection/components/SyncRow.tsx
nalanj: fixed in 773f593e

nalanj:
@bodinsamuel Fixes for all of your comments are in, I'm pretty sure. Thanks for the great feedback.

I did want to note for you and @bastienbeurier a couple of spots where I moved away from the design. Obviously the object counts aren't in yet. I also centered the Last Sync Start and Next Sync Start columns, rather than left aligning them, because in practice it felt more consistent to me. I'm down to change to match the design though.

Also the tooltips are showing above rather than below the items, because that's what was happening prior. Let me know if we'd rather they were below. I wasn't sure how the tooltip library handles window sizing and all (except that a quick test showed that it goes below when it's appropriate).

bastienbeurier:
> @bodinsamuel Fixes for all of your comments are in, I'm pretty sure. Thanks for the great feedback.
> 
> I did want to note for you and @bastienbeurier a couple of spots where I moved away from the design. Obviously the object counts aren't in yet. I also centered the Last Sync Start and Next Sync Start columns, rather than left aligning them, because in practice it felt more consistent to me. I'm down to change to match the design though.
> 
> Also the tooltips are showing above rather than below the items, because that's what was happening prior. Let me know if we'd rather they were below. I wasn't sure how the tooltip library handles window sizing and all (except that a quick test showed that it goes below when it's appropriate).

Thanks for sharing the details!

All good for the tooltip IMO. For the alignment, I feel everything should be left aligned, similarly to what we do on other tables in the app. I'll comment on the design directly.

nalanj:
@bastienbeurier columns are all left aligned in bd174f98.

Screenshot 2024-10-30 at 11 48 47 AM

packages/webapp/src/pages/Connection/components/SyncRow.tsx
bodinsamuel: Thanks, I know I was lazy when I implemented the first dropdown :D

bodinsamuel (approved):
💯

bodinsamuel:
> @bodinsamuel what do we think of just having a tag that says NEVER RUN when a connection hasn't run yet, to handle that null situation?

Sure, lgtm 👌🏻

nalanj:
@bastienbeurier here's a screenshot of those wider labels:

Screenshot 2024-10-30 at 2 05 14 PM

One thing I don't love about the increase in verbosity there is how the menu looks when it's actually running:

image

But also it's probably fine.

bastienbeurier:
> @bastienbeurier here's a screenshot of those wider labels:
> 
> Screenshot 2024-10-30 at 2 05 14 PM
> One thing I don't love about the increase in verbosity there is how the menu looks when it's actually running:
> 
> image
> But also it's probably fine.

Mhmmm I see. Do you think "Trigger (...)" would be descriptive enough? (to reduce verbosity)  "Trigger" is the syntax we use on this endpoint: https://docs.nango.dev/reference/api/sync/trigger

nalanj:
@bastienbeurier A few options I see:

1. Trigger Incremental & Trigger Refresh
2. Execute & Refresh
3. Combine the two into a menu item called Trigger (or Execute), put them into a modal like the full refresh is now, but put a checkbox to trigger and explain the full refresh.

bastienbeurier:
> @bastienbeurier A few options I see:
> 
> 1. Trigger Incremental & Trigger Refresh
> 2. Execute & Refresh
> 3. Combine the two into a menu item called Trigger (or Execute), put them into a modal like the full refresh is now, but put a checkbox to trigger and explain the full refresh.

Trigger Incremental & Trigger Refresh work for me for now until we find something better!

nalanj:
OK, I think this is it!

image

PR #2918, opened 2024-10-30
fix(ui): Connection Show refactor

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1924/ui-for-userorganization-info-connection-details-authorization-tab

## Major

- Add end user profile in Connection Header
There is a new method in town that fetches the connection and all the data `getConnectionForPrivateApi`

- Add endpoint `POST /connections/:id/refresh` 
Force refresh was part of the GET method, which was hard to manage and also not cool for idempotency (browser and SWR were retrying the call).

## Minor

- Add end user profile in authorization details

https://github.com/user-attachments/assets/c199ae87-1699-40a2-bf76-660a2abe8446


packages/shared/lib/services/connection.service.ts
```diff
@@ -745,6 +745,31 @@ class ConnectionService {
...
+            .from(`_nango_connections`)
+            .leftJoin('end_users', 'end_users.id', '_nango_connections.end_user_id')
+            .where({ connection_id: connectionId, provider_config_key: providerConfigKey, '_nango_connections.environment_id': environmentId, deleted: false })
+            .first();
+        if (!result) {
+            return null;
+        }
+
+        return { connection: encryptionManager.decryptConnection(result.connection)!, end_user: result.end_user };
+    }
```
TBonnin, line 771: can we use `Result` as the return type of all new function that can actually fail?

packages/types/lib/connection/api/get.ts
```diff
@@ -68,19 +62,33 @@ export type GetConnection = Endpoint<{
...
+    Querystring: {
+        env: string;
+        provider_config_key: string;
+    };
+    Path: '/api/v1/connections/:connectionId/refresh';
+    Error: ApiError<'unknown_provider_config'> | ApiError<'failed_to_refresh', any, ActiveLog | null>;
+    Success: {
+        data: {
+            success: boolean;
+        };
```
TBonnin, line 91: I am not super convinced by the type names between `ApiConnectionSimple`, `ApiConnectionFull`, `ApiPublicConnection`, there are not super descriptive but I cannot think of something better.

packages/webapp/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
     "compilerOptions": {
-        "target": "ES5",
-        "lib": ["DOM", "DOM.Iterable", "ESNext"],
+        "target": "ES2021",
+        "lib": ["DOM", "DOM.Iterable", "ES2021"],
```
TBonnin, line 4: i am curious about those changes?

TBonnin (approved):

packages/types/lib/connection/api/get.ts
bodinsamuel: yep I agree, usually I go with some kind of variant like this but never found a satisfying name. 
Alternatively, it could be named ApiGetConnectionsItem and ApiGetConnectionItem or something like that ^^' 

it's either too verbose or too vague

packages/webapp/tsconfig.json
bodinsamuel: ah yes forgot to put a comment . replaceAll is only available for > ES2020 and this
```ts
type 'IterableIterator<[string, string]>' can only be iterated through when using the '--downlevelIteration' flag or with a '--target' of 'es2015' or higher.

19             [...usp.entries()].forEach(([key, value]) => {
```

PR #2921, opened 2024-10-30
fix(api): reup search in /connections

### LINKED ISSUE #2920: Invalid query params in endpoint /connection?connectionId
```
"error":{"code":"invalid_query_params","errors":[{"code":"unrecognized_keys","message":"Unrecognized key(s) in object: 'connectionId'","path":[]}]}}
```

I used the example straight out of the documentation here:
https://docs.nango.dev/reference/api/connection/list

This seemed to be working before.

bodinsamuel:
## Describe your changes

Closes #2920 

- Reup search in `GET /connections`
This feature was forgotten will moving the endpoint code to the new format in #2897
Note that the behavior is slightly different now, not sure if I should introduce a new `search` prop to filter on everything and keep `connectionId` as it was before (filtering only connectionId)

### DIFF
--- a/packages/server/lib/controllers/connection/getConnections.ts
+++ b/packages/server/lib/controllers/connection/getConnections.ts
@@ -1,20 +1,30 @@
 import { asyncWrapper } from '../../utils/asyncWrapper.js';
-import { requireEmptyQuery, zodErrorToHTTP } from '@nangohq/utils';
+import { zodErrorToHTTP } from '@nangohq/utils';
 import type { GetPublicConnections } from '@nangohq/types';
 import { AnalyticsTypes, analytics, connectionService } from '@nangohq/shared';
 import { connectionToPublicApi } from '../../formatters/connection.js';
+import { z } from 'zod';
+
+const validationQuery = z
+    .object({
+        connectionId: z.string().max(255).optional()
+    })
+    .strict();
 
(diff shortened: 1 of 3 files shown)


packages/server/lib/controllers/connection/getConnections.ts
```diff
@@ -1,20 +1,30 @@
...
+        });
         return;
     }
 
     const { environment, account } = res.locals;
+    const queryParam: GetPublicConnections['Querystring'] = queryParamValues.data;
 
     void analytics.track(AnalyticsTypes.CONNECTION_LIST_FETCHED, account.id);
-    const connections = await connectionService.listConnections({ environmentId: environment.id, limit: 10000 });
+    const connections = await connectionService.listConnections({ environmentId: environment.id, search: queryParam.connectionId, limit: 10000 });
```
TBonnin, line 27: as you mention in the description, it might be more flexible and understandable to have `search` be of type `{connection?: {id: string}, end_user?: {name?: string, email?: string}` and only search if attributes is set but that's way more complicated

TBonnin (approved):

packages/server/lib/controllers/connection/getConnections.ts
bodinsamuel: yeah and at the same time search is usually a wildcard, not sure I can fine tune after when Bastien answers

PR #2922, opened 2024-10-30
feat: add settings for OpenTelemetry export

TBonnin:
## Describe your changes

Add environment settings to input OpenTelemetry endpoint and headers
Screenshot 2024-10-30 at 13 22 26

Note: 
- an empty endpoint is considered as disabled. I don't think adding an extra checkmark is actually useful
- The OpenTelemetry feature is only enabled for enterprise setup or via feature flag. I was thinking I would hide the settings if not enabled but I think it is nice to promote the feature even if not enabled. Editing the settings would show `OpenTelemetry export is not enabled for this account` error. I am happy to also make a mention of it being a paid feature in the environment setting page. Let me know @bastienbeurier 

### DIFF
--- a/packages/server/lib/controllers/v1/environment/otlp/postSettings.ts
+++ b/packages/server/lib/controllers/v1/environment/otlp/postSettings.ts
@@ -6,7 +6,7 @@ import { environmentService, featureFlags } from '@nangohq/shared';
 
 const bodyValidation = z
     .object({
-        endpoint: z.string(),
+        endpoint: z.string().url(),
         headers: z.record(z.string())
     })
     .strict();
--- a/packages/server/lib/controllers/v1/environment/otlp/postSettings.ts
+++ b/packages/server/lib/controllers/v1/environment/otlp/postSettings.ts
@@ -1,0 +1,53 @@
+import { z } from 'zod';
+import { asyncWrapper } from '../../../../utils/asyncWrapper.js';
+import type { DBTeam, UpdateOtlpSettings } from '@nangohq/types';
+import { requireEmptyQuery, zodErrorToHTTP, isEnterprise } from '@nangohq/utils';
+import { environmentService, featureFlags } from '@nangohq/shared';
+
+const bodyValidation = z
+    .object({
+        endpoint: z.string(),
+        headers: z.record(z.string())
+    })
+    .strict();
+
+export const postSettings = asyncWrapper<UpdateOtlpSettings>(async (req, res) => {
+    const emptyQuery = requireEmptyQuery(req, { withEnv: true });
+    if (emptyQuery) {
+        res.status(400).send({ error: { code: 'invalid_query_params', errors: zodErrorToHTTP(emptyQuery.error) } });
+        return;
+    }
+
+    const val = bodyValidation.safeParse(req.body);
+
+    if (!val.success) {
+        res.status(400).send({
+            error: { code: 'invalid_body', errors: zodErrorToHTTP(val.error) }
+        });
+        return;
+    }
+
+    const { environment, account } = res.locals;
+
+    const isEnabled = await isOtlpEnabled({ account });
+    if (!isEnabled) {
+        res.status(403).send({ error: { code: 'forbidden', message: 'OpenTelemetry export is not enabled for this account' } });
+        return;
+    }
+
+    const { data: settings } = val;
+
+    settings.endpoint = settings.endpoint.trim().replace(/\/$/, '');
+
+    const newSettings = settings.endpoint.length > 0 ? settings : null;
+    await environmentService.editOtlpSettings(environment.id, newSettings);
+
+    res.send(settings);
+});
+
+const isOtlpEnabled = async ({ account }: { account: DBTeam }): Promise<boolean> => {
+    if (isEnterprise) {
+        return true;
+    }
+    return featureFlags.isEnabled('feature:otlp:account', account.uuid, false);
+};
--- a/packages/shared/lib/services/environment.service.ts
+++ b/packages/shared/lib/services/environment.service.ts
@@ -336,6 +336,10 @@ class EnvironmentService {
         return db.knex.from<DBEnvironment>(TABLE).where({ id }).update({ hmac_key: hmacKey }, ['id']);
     }
 
+    async editOtlpSettings(environmentId: number, otlpSettings: { endpoint: string; headers: Record<string, string> } | null): Promise<DBEnvironment | null> {
+        return db.knex.from<DBEnvironment>(TABLE).where({ id: environmentId }).update({ otlp_settings: otlpSettings }, ['id']);
+    }
+
     async getEnvironmentVariables(environment_id: number): Promise<DBEnvironmentVariable[] | null> {
         const result = await db.knex.select('*').from<DBEnvironmentVariable>(`_nango_environment_variables`).where({ environment_id });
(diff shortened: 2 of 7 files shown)


packages/webapp/src/pages/Environment/Settings.tsx
```diff
@@ -118,6 +124,9 @@ export const EnvironmentSettings: React.FC = () => {
         setSlackConnectedChannel(slack_notifications_channel);
 
         setEnvVariables(env_variables);
+
+        setOtlpEndpoint(environment.otlp_settings?.endpoint || '');
+        setOtlpHeaders(environment.otlp_settings?.headers || {});
```
TBonnin, line 129: caveat: headers are reordered by key automatically. I didn't want to make the model and code more complicated for such a tiny feature and the fact that in most case there will be only 1 header max

packages/server/lib/controllers/v1/environment/otlp/postSettings.ts
```diff
@@ -0,0 +1,53 @@
+import { z } from 'zod';
```
bodinsamuel, line 1: having one endpoint per setting is a bit insane imo
what do you think of taking this opportunity to create a generic endpoint (that only accepts otlp for now)?

packages/shared/lib/services/environment.service.ts
```diff
@@ -336,6 +336,10 @@ class EnvironmentService {
         return db.knex.from<DBEnvironment>(TABLE).where({ id }).update({ hmac_key: hmacKey }, ['id']);
     }
 
+    async editOtlpSettings(environmentId: number, otlpSettings: { endpoint: string; headers: Record<string, string> } | null): Promise<DBEnvironment | null> {
+        return db.knex.from<DBEnvironment>(TABLE).where({ id: environmentId }).update({ otlp_settings: otlpSettings }, ['id']);
+    }
```
bodinsamuel, line 341: same for this, there are 20 update functions while only need one that does this
```
await db.knex.from<DBEnvironment>(TABLE).where({ id: environmentId }).update(updatedEnv);
```

packages/webapp/src/pages/Environment/Settings.tsx
```diff
@@ -988,6 +1058,111 @@ export const EnvironmentSettings: React.FC = () => {
...
+                                                Add OpenTelemetry Endpoint Header
+                                            </button>
+                                            <button type="submit" className="hover:bg-gray-200 bg-white text-gray-700 flex h-11 rounded-md px-4 pt-3 text-sm">
+                                                Save OpenTelemetry Endpoint Headers
+                                            </button>
+                                        </div>
+                                    </form>
+                                </div>
+                            )}
+                        </div>
```
bodinsamuel, line 1166: Would be nice to have an example in the tooltip or better in the placeholder of the input

packages/server/lib/controllers/v1/environment/otlp/postSettings.ts
```diff
@@ -0,0 +1,53 @@
...
+import { asyncWrapper } from '../../../../utils/asyncWrapper.js';
+import type { DBTeam, UpdateOtlpSettings } from '@nangohq/types';
+import { requireEmptyQuery, zodErrorToHTTP, isEnterprise } from '@nangohq/utils';
+import { environmentService, featureFlags } from '@nangohq/shared';
+
+const bodyValidation = z
+    .object({
+        endpoint: z.string(),
```
bodinsamuel: is it a URL?

packages/webapp/src/pages/Environment/Settings.tsx
```diff
@@ -988,6 +1058,111 @@ export const EnvironmentSettings: React.FC = () => {
                                 </form>
                             </div>
                         </div>
+                        <div>
```
bodinsamuel, line 1061: To see with @bastienbeurier but I feel like it would be better for the UX to clearly state that the account has no access to this feature instead of clicking edit, pasting something, and then getting an error.
You can return the flag in the `/meta`.

packages/server/lib/controllers/v1/environment/otlp/postSettings.ts
```diff
@@ -0,0 +1,53 @@
...
+    await environmentService.editOtlpSettings(environment.id, newSettings);
+
+    res.send(settings);
+});
+
+const isOtlpEnabled = async ({ account }: { account: DBTeam }): Promise<boolean> => {
+    if (isEnterprise) {
+        return true;
+    }
+    return featureFlags.isEnabled('feature:otlp:account', account.uuid, false);
```
bodinsamuel, line 52: what about self hosted?

bodinsamuel (commented):
few open questions

packages/server/lib/controllers/v1/environment/otlp/postSettings.ts
TBonnin: it is a bit insane. I also think the design is now overwhelming with all those options and I would vote for not only a endpoints refactor but a UI redesign as well. cc @bastienbeurier 
I would prefer to push this otlp export feature to completion though and not go into a refactor at this point.

packages/webapp/src/pages/Environment/Settings.tsx
TBonnin: good point. will do

packages/server/lib/controllers/v1/environment/otlp/postSettings.ts
TBonnin: yes. good point

packages/server/lib/controllers/v1/environment/otlp/postSettings.ts
TBonnin: it is supposed to be a scale plan only feature so they can only be enterprise or cloud. Am I missing something?

packages/webapp/src/pages/Environment/Settings.tsx
TBonnin: will check with @bastienbeurier

packages/server/lib/controllers/v1/environment/otlp/postSettings.ts
bodinsamuel: it shouldn't be shown in the UI then

packages/webapp/src/pages/Environment/Settings.tsx
TBonnin: as per our discussion with @bastienbeurier: yes this would be a better UX but let's ship it like this for now and we will work on redesigning the env settings page including clearly stating that this option is paid only

packages/server/lib/controllers/v1/environment/otlp/postSettings.ts
TBonnin: Agreed. Just check with @bastienbeurier though and we decided to keep it as is for now (aka: failing unless enterprise or cloud with flag enabled) and improve it when making the design of this page better

bodinsamuel:
let me know when you need a second review

TBonnin:
@bodinsamuel fixed the zod schema and the endpoint placeholder.
The rest (multiple edit endpoints and functions, ux, etc...) will be improved/refactored later.

bodinsamuel (approved):

bodinsamuel (approved):

PR #2924, opened 2024-10-30
fix(records): Fix id comparison check in update

nalanj:
## Describe your changes

In the `update` function within records, if the id over the wire is an integer, the comparison always fails because it checks the external id of the new record with the id of the old record. This small fix compares external ids for both so that they're converted to strings.

### DIFF
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -333,7 +333,7 @@ export async function update({
 
                     const { record: oldRecord, ...oldRecordRest } = rawOldRecord;
 
-                    const record = records.find((record) => record.external_id === oldRecord.id);
+                    const record = records.find((record) => record.external_id === rawOldRecord.external_id);
 
                     const newRecord: FormattedRecord = {
                         ...oldRecordRest,


packages/records/lib/models/records.ts
```diff
@@ -289,7 +289,7 @@ export async function update({
 
                     const { record: oldRecord, ...oldRecordRest } = rawOldRecord;
 
-                    const record = records.find((record) => record.external_id === oldRecord.id);
+                    const record = records.find((record) => record.external_id === rawOldRecord.external_id);
```
khaliqgant, line 336: I wonder why we were comparing the id with the external_id in the first place. That appears to be plain wrong which means this method didn’t work as intended to begin with. 

If so, good catch here!

bodinsamuel (approved):
Nice

packages/records/lib/models/records.ts
TBonnin: @khaliqgant it was more of a naming issue. `record` is the entire record as store in the db. what is called `oldRecord` is actually the record data as defined in the model. In theory `recordData.id` and `record.external_id` is the same thing except when the typescript compiler is lying :)

@nalanj this is fixing the bug for sure but I would also fix the types so we don't trip over the issue somewhere else

TBonnin (approved):
merging because it is fixing a bug but we should also fix the underlying issue around the type

PR #2927, opened 2024-10-31
fix: openapi spec bad merge add CI

bodinsamuel:
## Describe your changes

- Fix openapi corrupted file
A bad merge happened probably in my previous PRs, mintlify is not triggering any error and we don't have CI in place to catch that


khaliqgant (approved):
Thanks. Remember issues like this are pretty annoying difficult to catch. Definitely agree we should have something in the CI to catch issues like these - so thanks for adding!

PR #2929, opened 2024-10-31
feat(integrations): add support for aws-iam

hassan254-prog:
## Describe your changes

- Add support for [AWS IAM](https://docs.aws.amazon.com/iam/)


docs-v2/integrations/all/aws.mdx
```diff
@@ -22,14 +22,48 @@ API configuration: [`aws`](https://nango.dev/providers.yaml)
...
+- AWS uses different domain extensions codes for different regions, i.e `us-east-2` for the US East (Ohio) region, `us-east-1` for the US East (N. Virginia) region, or `eu-west-1` for Europe (Ireland) region.It also uses different subdomains for different `user pools`.
 
 
 ## API gotchas
+- To create a new connection in Nango for AWS IAM, you'll need an Access Key ID and a Secret Access Key. Use these to set up your connection, with the Access Key ID as the username and the Secret Access Key as the password. For more details on generating these keys, visit [Create an access key](https://docs.aws.amazon.com/IAM/latest/UserGuide/access-key-self-managed.html#Using_CreateAccessKey).
+- With this you can now use your basic auth credentials to sign you requests as follows. For more on how to sign your requests, please vist [AWS signature version 4 for API requests](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_sigv.html)
+
+```js
+    if ('username' in connection.credentials && 'password' in connection.credentials && 'region' in connection.connection_config) {
```
khaliqgant, line 39: @bodinsamuel @TBonnin Not saying we should do this now but I wonder if it would make sense to internalize into the platform this kind of logic at some point.

packages/shared/providers.yaml
```diff
@@ -538,6 +538,24 @@ aws:
             example: com
             pattern: '^[a-z.]+$'
 
+aws-iam:
+    display_name: AWS IAM
+    categories:
+        - dev-tools
+    auth_mode: BASIC
+    proxy:
+        base_url: https://iam.amazonaws.com/${connectionConfig.region}
```
hassan254-prog: @bodinsamuel, I've added a `connectionConfig` region here to prompt the user to input the region. This is required for signing the requests but not for making the actual call. Setting `baseUrlOverride` at the script level isn't the most efficient approach. Another option could be to set `aws-iam` as unauthenticated, then save the username, password, and region to `connectionConfig`. However, this would mean storing credentials in an unencrypted format. Do you have any suggestions for a workaround here?

bodinsamuel: mmh maybe we could introduce a `proxy.connection_base_url` to dissociate the usage?

docs-v2/integrations/all/aws.mdx
bodinsamuel: for sure, I'm all for it

khaliqgant: Ok cool, thanks, I'll sync with Bastien from a product perspective on this.

packages/shared/providers.yaml
hassan254-prog: Yes, that works, thank you Samuel will push an update for this 👍.

khaliqgant: I find this confusing to have:
```
proxy:
    base_url: string
    connection_base_url: string
```

I'm also unclear how this solves the issue? Can you explain further @bodinsamuel ?

packages/shared/providers.yaml
```diff
@@ -538,6 +538,24 @@ aws:
             example: com
             pattern: '^[a-z.]+$'
 
+aws-iam:
+    display_name: AWS IAM
```
TBonnin, line 559: This is a bit narrow. Can we have the aws subdomain be a `connection.config` param? Otherwise we are going to need a provider for each separate aws services that has its own subdomain

docs-v2/integrations/all/aws.mdx
TBonnin: 100% Potential solutions: have a `AWS_SIGN_V4` dedicated auth_mode or a proxy subattributes to specify pre-build authorization logic
```
    proxy:
        headers:
            Authorization: AWS_SIGN_V4

```

packages/shared/providers.yaml
hassan254-prog: @bodinsamuel , just to confirm that I understand your suggestion: we should introduce `connection_base_url: ${connectionConfig.region}` to help us get the region, as this is not used anywhere when building proxies. However, we will still maintain `base_url` without region? not clean but better to my initial implementation.

We can as well have this to get these `connectionConfigs` that we can later use in scripts, what do you think?

```
foo:
    display_name: Foo
    categories:
        - productivity
    auth_mode: OAUTH2
    proxy:
        base_url: url
    connection_configs:
        region: ${connectionConfig.region}
        host: ${connectionConfig.host}

Comment (bodinsamuel) [reply packages/shared/providers.yaml]: My understanding is that we only need this url `https://iam.amazonaws.com/${connectionConfig.region}` for connecting, so if we put that by default every call to aws will need to use baseUrlOverride. And if we don't do it we need a IF in our code.

If we dissociate the base_url for regular calls and token calls it could help solve that with zero specific IF in the code and good DX.

File: packages/shared/providers.yaml (hunk 3 above)
Comment (hassan254-prog) [reply packages/shared/providers.yaml:559]: @TBonnin , I might be missing something here, could you please elaborate further? Might you be referring renaming the provider to `aws-services` and have `base_url: https://${connectionConfig.service}.amazonaws.com`.

File: packages/shared/providers.yaml (hunk 2 above)
Comment (bodinsamuel) [reply packages/shared/providers.yaml]: we have `authorization_url` for oauth2 maybe a similar wording could help, or authentication since you are adding `authentication_header`

Comment (hassan254-prog) [reply packages/shared/providers.yaml]: Yes, the wording looks great 👌. Perhaps we could include it in the proxy object, as it's only needed for proxy calls, including it in `proxy.headers`, will automatically include it when constructing headers, something like this;

```
aws-iam:
    display_name: AWS IAM
    categories:
        - dev-tools
    auth_mode: BASIC
    proxy:
        base_url: https://iam.amazonaws.com
        authentication_header: ${connectionConfig.region}
        retry:
            at: 'x-ratelimit-reset'
    docs: https://docs.nango.dev/integrations/all/aws

bodinsamuel: I meant you are adding `authentication_header` in your other PR, not sure this one is a header

hassan254-prog: I think we may be misunderstanding each other because of the wording 😅, so here, I will revert to using `connection_base_url`  to dissociate base_url for regular calls, and `connection_base_url` to prompt the user to input the region. Updated PR with latest [commit](https://github.com/NangoHQ/nango/pull/2929/commits/84394e95f4e3523d53da7b301914596360ab438d)

packages/shared/providers.yaml
TBonnin: It is linked to the the discussion about connection_base_url I think. I was asking if we are gonna need to have a new provider alias for each separate aws service or if we could have a generic `AWS Signature` provider or something similar.

hassan254-prog: I think we could introduce a separate auth mode, `AWS_SIGN_V4`, and then allow multiple AWS providers, such as `aws-iam` and `aws-s3`, to use this same auth mode. This approach would reduce the size of the documentation file for each provider, as we would have separate documentation for each one. Additionally, it would enable us to categorize providers by product areas, such as analytics, dev tools, etc. What do you think?

TBonnin: that would be better I think. What do @khaliqgant and @bodinsamuel think?

khaliqgant: Yes, this works well.

hassan254-prog: Addressed [changes](https://github.com/NangoHQ/nango/pull/2929#discussion_r1827435791) in the latest [commit](https://github.com/NangoHQ/nango/pull/2929/commits/84394e95f4e3523d53da7b301914596360ab438d)

bodinsamuel (approved):

packages/shared/providers.yaml
TBonnin: linking this discussion with https://github.com/NangoHQ/nango/pull/2942#pullrequestreview-2416097872
The AWS_SIGN_V4 and WSSE auth mode seems to basically be BASIC with a special header that must be constructed a certain way. 

What about NOT adding a new auth_mode (and everything that goes with it, endpoints, etc...) but add a new optional `Signature` attributes to a provider. If present, the logic at runtime will call a prebuilt function depending on the `Signature` value (ex: `AWS_SIGN_V4`, `EMARSYS_WSSE` that would generate the signature and add the header but wouldn't require any other code change

khaliqgant: Agree that this should happen but don't think that should block this PR. Tracking in https://linear.app/nango/issue/EXT-221/refactor-providersyaml-to-have-a-signature-attribute

khaliqgant (approved):

PR #2930, opened 2024-10-31
feat(ui): display organization logo automatically

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1986/display-domain-logo-as-avatar-in-connections

- Add https://logo.dev to display organization logo
Based on email domain, it's not perfect but should cover most use case. The value is computed in the frontend, could be better to store it or even allow customer to specify it but it's lightweight so not too critical.
Logo.dev has a free plan with attribution, we'll put a notice on the marketing site. Self-hosted and enterprise can BYOK if needed.
Our api key is stored in 1password, but anybody can create an account.

Screenshot 2024-10-31 at 11 55 17
Screenshot 2024-10-31 at 18 23 02


.env.example
```diff
@@ -103,3 +103,7 @@ MAILGUN_API_KEY=
 
 # Orchestrator
 ORCHESTRATOR_SERVICE_URL="http://localhost:3008"
+
+# Key to display domain logo in the UI
+# https://www.logo.dev/
+PUBLIC_LOGODEV_KEY=
```
TBonnin, line 109: LOGODEV_PUBLIC_KEY?

packages/webapp/src/components/AvatarCustom.tsx
```diff
@@ -20,3 +21,53 @@ export const AvatarCustom: React.FC<Props> = ({ displayName, ...props }) => {
...
+    'mailinator.com',
+    'me.com',
+    'orange.fr',
+    'outlook.com',
+    'proton.me',
+    'protonmail.com',
+    'qq.com',
+    'yahoo.co.uk',
+    'yahoo.com'
+]);
```
TBonnin, line 139: According to Claude we could add a few more:

> Additional Google domains (gmail variants)
> More Yahoo regional domains (.fr, .de, .co.jp)
> Popular European providers (web.de, mail.ru, yandex.ru)
> Major Asian providers (yeah.net)
> Additional Microsoft domains (outlook.de, outlook.fr, msn.com)
> Business-focused providers (zoho.com)
> Additional French providers (laposte.net, sfr.fr)
> German providers (gmx.net, gmx.com)
> More Apple domains (mac.com)

did you get this list from existing customer emails?

TBonnin (approved):
product question: why don't we want to show the email provider logo if end user email is a personal email?

.env.example
bodinsamuel: emphasis on Public because it will end up in the frontend, like all keys there:
packages/server/lib/controllers/v1/getEnvJs.ts

packages/webapp/src/components/AvatarCustom.tsx
bodinsamuel: I got the list from our user base and top providers but not exhaustive, I'll add some more

PR #2934, opened 2024-10-31
fix: records update webhook should be sent when coming from webhook exec

TBonnin:
Now sending a webhook to notify about records update (added, update, deleted) from within a webhook script. 
We used to only notify the customer when records where updated via a sync.

https://linear.app/nango/issue/NAN-2021/[stacker-poc]-webhook-script-updates-should-trigger-a-nango-webhook

### DIFF
--- a/packages/jobs/lib/execution/sync.ts
+++ b/packages/jobs/lib/execution/sync.ts
@@ -208,6 +208,7 @@ export async function handleSyncSuccess({ nangoProps }: { nangoProps: NangoProps
             records: {} as Record<string, SyncResult>,
             runTimeSecs: runTime
         };
+        const webhookSettings = await externalWebhookService.get(nangoProps.environmentId);
         for (const model of nangoProps.syncConfig.models) {
             let deletedKeys: string[] = [];
             if (nangoProps.syncConfig.track_deletes) {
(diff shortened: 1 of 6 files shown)


packages/jobs/lib/execution/sync.ts
```diff
@@ -268,49 +269,45 @@ export async function handleSyncSuccess({ nangoProps }: { nangoProps: NangoProps
 
             syncPayload.records[model] = { added, updated, deleted };
 
-            const webhookSettings = await externalWebhookService.get(nangoProps.environmentId);
```
TBonnin, line 271: moving the `externalWebhookService.get` outside of the models loop. No need to do it multiple times.

packages/jobs/lib/execution/webhook.ts
```diff
@@ -161,11 +166,71 @@ export async function handleWebhookSuccess({ nangoProps }: { nangoProps: NangoPr
...
+                        throw new Error(`Failed to send webhook for webhook: ${nangoProps.syncConfig.sync_name}`);
+                    }
+                } catch (err: unknown) {
+                    span?.setTag('error', err);
+                } finally {
+                    span.finish();
+                }
+            });
+        }
+    }
```
TBonnin, line 220: this looks very much like the logic in `./sync.ts` but I decided it was not worth consolidating the code because there is some small differences. For example webhook result handling doesn't have to deal with full refresh or track_deletes

packages/shared/lib/services/sync/job.service.ts
```diff
@@ -91,18 +91,24 @@ export const getSyncJobByRunId = async (run_id: string): Promise<SyncJob | null>
...
+export const updateSyncJobStatus = async (id: number, status: SyncStatus): Promise<SyncJob | null> => {
+    const [job] = await schema()
+        .from<SyncJob>(SYNC_JOB_TABLE)
+        .where({ id, deleted: false })
+        .update({
+            status,
+            updated_at: new Date()
+        })
+        .returning('*');
+    return job || null;
```
TBonnin, line 103: returning the job to avoid having to make another query if we need the job details after updating

packages/shared/lib/services/sync/job.service.ts
```diff
@@ -91,18 +91,24 @@ export const getSyncJobByRunId = async (run_id: string): Promise<SyncJob | null>
...
 };
 
-export const updateLatestJobSyncStatus = async (sync_id: string, status: SyncStatus): Promise<void> => {
+export const updateLatestJobSyncStatus = async (sync_id: string, status: SyncStatus): Promise<SyncJob | null> => {
```
TBonnin, line 106: not required. Only so this function return type is consistent with `updateSyncJobStatus` above

packages/types/lib/scripts/syncs/api.ts
```diff
@@ -4,4 +4,4 @@ export interface SyncResult {
     deleted: number;
 }
 
-export type SyncType = 'INCREMENTAL' | 'FULL';
+export type SyncType = 'INCREMENTAL' | 'FULL' | 'WEBHOOK';
```
khaliqgant, line 7: `EXTERNAL_WEBHOOK` or maybe `THIRD_PARTY_WEBHOOK`? I feel like there needs to be some specificity since we're sending a webhook and then the `syncType` is webhook -- semi confusing

khaliqgant (approved):
nit on naming.

packages/types/lib/scripts/syncs/api.ts
TBonnin: it is true there are a lot of webhooks involved but I am not sure `sync type = external webhook` bring much more additional info

PR #2935, opened 2024-11-01
fix(records): update records count when marking records as deleted

TBonnin:
## Describe your changes

we forgot to update the records count when marking records as deleted when track_deletes = true

### DIFF
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -263,7 +263,7 @@ export async function upsert({
                             connection_id: connectionId,
                             model,
                             environment_id: environmentId,
-                            count: Math.max(0, delta)
+                            count: delta
                         })
                         .onConflict(['connection_id', 'environment_id', 'model'])
                         .merge({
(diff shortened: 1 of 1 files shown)


packages/records/lib/models/records.ts
```diff
@@ -263,7 +263,7 @@ export async function upsert({
                             connection_id: connectionId,
                             model,
                             environment_id: environmentId,
-                            count: Math.max(0, delta)
```
TBonnin, line 266: unrelated but max was only useful when the counts were not populated yet for historic data. removing it since it doesn't bring anything

packages/records/lib/models/records.ts
```diff
@@ -419,23 +419,40 @@ export async function markNonCurrentGenerationRecordsAsDeleted({
...
+            })
+            .whereNot({
+                sync_job_id: generation
+            })
+            .update({
+                deleted_at: now,
+                updated_at: now,
+                sync_job_id: generation
+            })
+            .returning('id')) as unknown as string[];
```
TBonnin, line 442: this query hasn't changed. it is just wrapped in the transction

khaliqgant (approved):

PR #2936, opened 2024-11-01
feat: show records count on connections page

TBonnin:
Column is called `Records` cc @bastienbeurier
Count is shorten when greater than 1000. Ex: 1.6K, 3M.
Non-shorten count is always shown in the hover tooltip

Screenshot 2024-11-01 at 14 13 41

Depends on this [bugfix](https://github.com/NangoHQ/nango/pull/2935) to be merged first and inaccurate rows in the db to be fixed 

https://linear.app/nango/issue/NAN-1929/[trial]-show-object-count-in-sync-tab

### DIFF
--- a/packages/webapp/src/pages/Connection/Syncs.tsx
+++ b/packages/webapp/src/pages/Connection/Syncs.tsx
@@ -69,8 +69,9 @@ export const Syncs: React.FC<SyncsProps> = ({ syncs, connection, provider }) => 
                             <Table.Head className="w-[120px]">Models</Table.Head>
                             <Table.Head className="w-[120px]">Last Execution</Table.Head>
                             <Table.Head className="w-[80px]">Frequency</Table.Head>
+                            <Table.Head className="w-[80px]">Records</Table.Head>
                             <Table.Head className="w-[120px]">Last Sync Start</Table.Head>
-                            <Table.Head className="w-[130px]">Next Sync Start</Table.Head>
+                            <Table.Head className="w-[140px]">Next Sync Start</Table.Head>
                             <Table.Head className="w-[40px]"></Table.Head>
                         </Table.Row>
                     </Table.Header>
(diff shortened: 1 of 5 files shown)


packages/webapp/src/pages/Connection/Syncs.tsx
```diff
@@ -69,8 +69,9 @@ export const Syncs: React.FC<SyncsProps> = ({ syncs, connection, provider }) =>
                             <Table.Head className="w-[120px]">Models</Table.Head>
                             <Table.Head className="w-[120px]">Last Execution</Table.Head>
                             <Table.Head className="w-[80px]">Frequency</Table.Head>
+                            <Table.Head className="w-[80px]">Records</Table.Head>
                             <Table.Head className="w-[120px]">Last Sync Start</Table.Head>
-                            <Table.Head className="w-[130px]">Next Sync Start</Table.Head>
+                            <Table.Head className="w-[140px]">Next Sync Start</Table.Head>
```
TBonnin, line 74: unrelated but with 130 it would add a new line when value is `Schedule Paused`

packages/webapp/src/utils/utils.tsx
```diff
@@ -265,6 +265,27 @@ export function formatFrequency(frequency: string): string {
...
+        if (count >= value) {
+            const number = count / value;
+            if (Number.isInteger(number)) {
+                return `${number.toLocaleString()}${suffix}`;
+            }
+            return `${Number(number.toFixed(1)).toLocaleString()}${suffix}`;
+        }
+    }
+    return count.toLocaleString();
+}
```
bodinsamuel: You have formatQuantity() that does the same thing but shorter

bodinsamuel (approved):

khaliqgant (approved):

bodinsamuel (approved):

TBonnin:
FYI @khaliqgant @bodinsamuel added a commit to show the count breakdown per model in the tooltip 
https://github.com/NangoHQ/nango/pull/2936/commits/0e88f8ffff54b6c7cf79cfa9cdff00ff0787e589
Screenshot 2024-11-07 at 15 32 40

PR #2940, opened 2024-11-04
feat(nango-yaml): endpoint explicit definition

bodinsamuel:
## Context

Contributes to https://linear.app/nango/issue/NAN-1934/specify-script-endpoint-section-in-nangoyaml

We want to introduce an `entry` props that will define the endpoint category. That means we can no longer store this in a string (or we can but it would be bad). We still need to support the old way but this PR introduce a more explicit mode for endpoint.

> [!NOTE]
> This PR is changing the internal format but should not introduce any blocking changes for customers. If something is no longer working it's a bug

### Major Changes

- Change default format of endpoint in parser and internally
It's just the string split in an object. That will allow us to add `entry` later.
```ts
{ method: HTTP_METHOD, path: string } 
```

- Introduce a way to enforce CLI version at the API level
I ended up not enforcing a very recent CLI version because we want to keep retro-compat for a few months, but at some we will be able to increase the minimum CLI version. 

### Minor

- Rename HTTP_VERB to HTTP_METHOD
- Remove a lot of dead code

## 🧪 How to test?

- Run the dashboard, go to integrations and check that nothing crashes
- Deploy stuff to your local dashboard using the old CLI `nango deploy dev` 
- Deploy stuff to your local dashboard using the new CLI `node nango/packages/cli/dist/index.js deploy dev`


packages/shared/lib/services/flow.service.ts
```diff
@@ -126,76 +124,6 @@ class FlowService {
...
-                            foundAction = action.name;
-                        }
-                    }
-                }
-            }
-        }
-
-        return foundAction;
-    }
-
```
bodinsamuel, line 198: Dead code

packages/shared/lib/services/nango-config.service.ts
```diff
@@ -9,17 +9,18 @@ import type {
...
-
-export function loadStandardConfig(configData: NangoConfig, showMessages = false, isPublic?: boolean | null): ServiceResponse<StandardNangoConfig[] | null> {
+import { determineVersion, getInterval, isJsOrTsType, parseEndpoint } from '@nangohq/nango-yaml';
+import type { NangoSyncEndpointV2 } from '@nangohq/types';
+
+/**
+ * Legacy parser only used for flows.yaml
+ * TODO: kill this in favor of nango-yaml + transformation to StandardNangoConfig
+ */
+export function loadStandardConfig(configData: NangoConfig): ServiceResponse<StandardNangoConfig[] | null> {
```
bodinsamuel, line 23: After all the massive effort regarding nango.yaml parser and code splitting, I realized this file is close to being useless (which is a good thing), so I could remove everything that was used in the CLI before.

packages/webapp/src/pages/Integrations/components/FlowCard.tsx
```diff
@@ -1,85 +0,0 @@
-import { ArrowPathRoundedSquareIcon, BoltIcon } from '@heroicons/react/24/outline';
```
bodinsamuel, line 1: Dead code

packages/webapp/src/types.ts
```diff
@@ -133,48 +58,3 @@ export interface NangoSyncModel {
...
-    has_webhook_user_defined_secret?: boolean;
-    scopes: string;
-    auth_mode: AuthModeType;
-    created_at: string;
-    webhook_secret?: string;
-    custom?: Record<string, string>;
-    connection_count: number;
-    connections: Connection[];
-    docs: string;
-}
```
bodinsamuel, line 180: Everything removed here is dead code

packages/cli/lib/services/__snapshots__/config.service.unit.test.ts.snap
```diff
@@ -156,7 +156,8 @@ exports[`load > should parse a nango.yaml file that is version 2 as expected 1`]
         {
           "description": "Get a GitHub issue.",
           "endpoint": {
-            "GET": "/ticketing/tickets/{GithubCreateIssueInput:id}",
+            "method": "GET",
+            "path": "/ticketing/tickets/{GithubCreateIssueInput:id}",
           },
```
TBonnin, line 161: doing like below wasn't a good idea?
```
"endpoint": {
            "GET": "/ticketing/tickets/{GithubCreateIssueInput:id}",
            "entry": "something"
}
```

packages/server/lib/routes.ts
```diff
@@ -219,8 +220,8 @@ publicAPI.route('/connection/metadata').post(apiAuth, postPublicMetadata);
 publicAPI.route('/connection/metadata').patch(apiAuth, patchPublicMetadata);
 publicAPI.route('/connection').post(apiAuth, connectionController.createConnection.bind(connectionController));
 publicAPI.route('/environment-variables').get(apiAuth, environmentController.getEnvironmentVariables.bind(connectionController));
-publicAPI.route('/sync/deploy').post(apiAuth, postDeploy);
-publicAPI.route('/sync/deploy/confirmation').post(apiAuth, postDeployConfirmation);
+publicAPI.route('/sync/deploy').post(apiAuth, cliMinVersion('0.39.25'), postDeploy);
+publicAPI.route('/sync/deploy/confirmation').post(apiAuth, cliMinVersion('0.39.25'), postDeployConfirmation);
```
TBonnin, line 224: that's a breaking change. Would it make sense to first figure out if people are using older version to see how much of a problem is it going to be?

packages/webapp/src/pages/Integrations/providerConfigKey/Endpoints/components/List.tsx
```diff
@@ -70,7 +70,7 @@ export const EndpointsList: React.FC<{ integration: GetIntegration['Success']['d
                                 </Table.Header>
                                 <Table.Body>
                                     {flows.map((flow) => {
-                                        const usp = new URLSearchParams(flow.endpoint);
+                                        const usp = new URLSearchParams(flow.endpoint as unknown as Record<string, string>);
```
TBonnin, line 73: why the casting here?

TBonnin (commented):
code looks ok but I am worried about the min cli version breaking change that could prevent customers to deploy

packages/webapp/src/pages/Integrations/providerConfigKey/Endpoints/components/List.tsx
bodinsamuel: an interface is not compatible with a Record unfortunately

packages/server/lib/routes.ts
bodinsamuel: The selected version is from 6months ago, I can remove if we are not confident 👌🏻

packages/cli/lib/services/__snapshots__/config.service.unit.test.ts.snap
bodinsamuel: This makes finding the key more complicated, like currently doing Object.entries() and iterating

packages/server/lib/routes.ts
TBonnin: the safest route would be to only add the middleware in `logging-mode` and let it run for a couple of days to see if any customer would be blocked by it.

TBonnin (approved):
approving because from a code point of view I don't see any problem. The question about the breaking change (cli version requirement) still remain. Up to you if you want to take the risk or not

packages/server/lib/routes.ts
bodinsamuel: I'll log you are right 👌🏻

PR #2941, opened 2024-11-04
fix: records update bugs and inefficiency

TBonnin:
## Describe your changes

Found a few small issues in the records update function as well as some inefficiency.

### DIFF
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -1,17 +1,8 @@
 import dayjs from 'dayjs';
 import utc from 'dayjs/plugin/utc.js';
 import { db } from '../db/client.js';
-import type {
-    FormattedRecord,
-    FormattedRecordWithMetadata,
-    GetRecordsResponse,
-    LastAction,
-    RecordCount,
-    ReturnedRecord,
-    UnencryptedRecord,
-    UpsertSummary
-} from '../types.js';
-import { decryptRecord, decryptRecords, encryptRecords } from '../utils/encryption.js';
+import type { FormattedRecord, FormattedRecordWithMetadata, GetRecordsResponse, LastAction, RecordCount, ReturnedRecord, UpsertSummary } from '../types.js';
+import { decryptRecordData, encryptRecords } from '../utils/encryption.js';
 import { RECORDS_TABLE, RECORD_COUNTS_TABLE } from '../constants.js';
 import { removeDuplicateKey, getUniqueId } from '../helpers/uniqueKey.js';
 import { logger } from '../utils/logger.js';
@@ -173,7 +164,7 @@ export async function getRecords({
         }
 
         const results = rawResults.map((item) => {
-            const decryptedData = decryptRecord(item);
+            const decryptedData = decryptRecordData(item);
             const encodedCursor = Buffer.from(`${item.last_modified_at}||${item.id}`).toString('base64');
             return {
                 ...decryptedData,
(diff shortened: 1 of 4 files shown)


packages/records/lib/models/records.ts
```diff
@@ -322,24 +322,26 @@ export async function update({
             for (let i = 0; i < recordsWithoutDuplicates.length; i += BATCH_SIZE) {
                 const chunk = recordsWithoutDuplicates.slice(i, i + BATCH_SIZE);
 
-                updatedKeys.push(...(await getUpdatedKeys({ records: chunk, connectionId, model, trx })));
+                const chunkUpdatedKeys = await getUpdatedKeys({ records: chunk, connectionId, model, trx });
+                if (chunkUpdatedKeys.length === 0) {
+                    logger.warn(`Did not find records to update: ${{ connectionId, model }}`);
```
bodinsamuel: logger.warn does not exists (types are wrong), it's logger.warning

packages/records/lib/models/records.ts
```diff
@@ -322,24 +313,27 @@ export async function update({
             for (let i = 0; i < recordsWithoutDuplicates.length; i += BATCH_SIZE) {
                 const chunk = recordsWithoutDuplicates.slice(i, i + BATCH_SIZE);
 
-                updatedKeys.push(...(await getUpdatedKeys({ records: chunk, connectionId, model, trx })));
+                const oldRecords = await getRecordsToUpdate({ records: chunk, connectionId, model, trx });
```
TBonnin, line 316: we used to make 2 queries to get the existing records to update. :(

packages/records/lib/models/records.ts
```diff
@@ -322,24 +313,27 @@ export async function update({
...
                     }
 
-                    const { record: oldRecord, ...oldRecordRest } = rawOldRecord;
-
-                    const record = records.find((record) => record.external_id === rawOldRecord.external_id);
+                    const { json: newRecordData, ...newRecordRest } = inputRecord;
 
                     const newRecord: FormattedRecord = {
-                        ...oldRecordRest,
+                        ...newRecordRest,
```
TBonnin, line 333: that was a bug. The only attributes that is different is the sync_job_id but was not updated

packages/records/lib/utils/encryption.ts
```diff
@@ -14,7 +14,7 @@ function isEncrypted(data: UnencryptedRecordData | EncryptedRecordData): data is
     return 'encryptedValue' in data;
 }
 
-export function decryptRecord(record: FormattedRecord): UnencryptedRecordData {
+export function decryptRecordData(record: FormattedRecord): UnencryptedRecordData {
```
TBonnin, line 17: renaming to make it clearer that the function only return the unencrypted data not the entire record

packages/webhooks/lib/sync.ts
```diff
@@ -69,7 +69,7 @@ export const sendSync = async ({
             responseResults?.added === 0 && responseResults?.updated === 0 && (responseResults.deleted === 0 || responseResults.deleted === undefined);
 
         if (!webhookSettings.on_sync_completion_always && noChanges) {
-            await logCtx?.info('There were no added, updated, or deleted results. No webhook sent, as per your environment settings');
+            await logCtx?.info(`There were no added, updated, or deleted results for model ${model}. No webhook sent, as per your environment settings`);
```
TBonnin, line 72: unrelated but it is confusing in the logs to see multiple times the same message when sync is multi model

bodinsamuel (approved):

packages/records/lib/models/records.ts
TBonnin: old man yells at cloud

TBonnin: old man yells at cloud (1)

PR #2942, opened 2024-11-05
feat(authmode): add a new signature authmode

hassan254-prog:
## Describe your changes

- Add WSSE auth mode

[EXT-206](https://linear.app/nango/issue/EXT-206/add-emarsys-support)

With the new changes we can now comfortably support [emarsys](https://dev.emarsys.com/docs/core-api-reference/b3c3a1eba8515-authentication#wsse-formata-namewsse-formata) and also found [oroinc](https://doc.oroinc.com/api/authentication/wsse/) that also supports WSSE auth mode.


hassan254-prog:
@bodinsamuel, I wanted to check if there have been any recent updates to the flex properties of the container holding the username and password fields;
image

TBonnin (commented):
Instead of adding yet another auth_mode. Can we simply add a new `signature` attribute to provider.yaml entry to be used with `Basic` auth mode.
Based on this attribute the logic can add the required header with the correct signature based on the provider name instead of having yet another endpoint, etc...
We could also use it for AWS signed APIs

hassan254-prog:
> Instead of adding yet another auth_mode. Can we simply add a new `signature` attribute to provider.yaml entry to be used with `Basic` auth mode. Based on this attribute the logic can add the required header with the correct signature based on the provider name instead of having yet another endpoint, etc... We could also use it for AWS signed APIs

Yes, that would be a better approach. To maintain the properties of `BASIC` auth_mode, specifically, the absence of a `token`, `expire_at`, and `refresh` attributes, I suggest introducing a custom auth_mode that incorporates the `signature` attribute, `WSSE`,  `AWS_V4` e.t.c. This way, we can preserve the properties of `BASIC` auth_mode. What do you think?

TBonnin:
> Yes, that would be a better approach. To maintain the properties of `BASIC` auth_mode, specifically, the absence of a `token`, `expire_at`, and `refresh` attributes, I suggest introducing a custom auth_mode that incorporates the `signature` attribute, `WSSE`, `AWS_V4` e.t.c. This way, we can preserve the properties of `BASIC` auth_mode. What do you think?

Sure. I am trying to avoid the out-of-control increase of AUTH_MODE. If the solution is to group all the signature related ones under a single AUTH_MODE this is already better than having a AUTH_MODE specific to a provider or a signature method

packages/types/lib/providers/provider.ts
```diff
@@ -135,4 +135,10 @@ export interface ProviderTwoStep extends BaseProvider {
     token_expires_in_ms?: number;
     proxy_header_authorization?: string;
 }
-export type Provider = BaseProvider | ProviderOAuth1 | ProviderOAuth2 | ProviderJwt | ProviderTwoStep;
+export interface ProviderSignatureBased extends BaseProvider {
+    signature_protocol: 'WSSE';
```
hassan254-prog: With the new changes, we can now comfortably add support to more than one signature-based auth method, such as `WSSE` and `AWS_V4`, without needing to add a new auth mode. Thanks @TBonnin for suggesting this change.

docs-v2/reference/api-configuration.mdx
```diff
@@ -193,7 +193,7 @@ slack:
 </ResponseField>
 
 <ResponseField name="auth_mode" type="string" >
-    The authentication mode. Must be one of: "API_KEY", "APP", "APP_STORE", "BASIC", "NONE", "OAUTH1", "OAUTH2", "OAUTH2_CC", "CUSTOM", "TBA", "TABLEAU", "JWT", "BILL", "TWO_STEP".
+    The authentication mode. Must be one of: "API_KEY", "APP", "APP_STORE", "BASIC", "NONE", "OAUTH1", "OAUTH2", "OAUTH2_CC", "CUSTOM", "TBA", "TABLEAU", "JWT", "BILL", "TWO_STEP", "SIGNATURE_BASED".
```
TBonnin: what about simply `SIGNATURE`

packages/shared/lib/services/connection.service.ts
```diff
@@ -467,6 +470,50 @@ class ConnectionService {
...
+                ...encryptedConnection,
+                updated_at: new Date()
+            })
+            .returning('*');
+
+        void analytics.track(AnalyticsTypes.SIGNATURE_BASED_CONNECTION_INSERTED, account.id, { provider: config.provider });
+
+        return [{ connection: connection!, operation: connection ? 'override' : 'creation' }];
+    }
+
```
TBonnin, line 213: do we need a new function just for this. All the upsert functions are basically identical

packages/shared/lib/services/connection.service.ts
```diff
@@ -2044,6 +2102,54 @@ class ConnectionService {
...
+            const passwordDigest = Buffer.from(sha1Hash, 'hex').toString('base64');
+
+            const token = `UsernameToken Username="${username}", PasswordDigest="${passwordDigest}", Nonce="${nonce}", Created="${timestamp}"`;
+
+            return token;
+        } catch (err) {
+            throw new NangoError('wsse_token_generation_error', { message: err instanceof Error ? err.message : 'unknown error' });
+        }
+    }
```
TBonnin: I would have this function in its own file in `lib/signature/wsse` for instance. It would be easier to find and once we add a couple of other signature methods, the current file might become a bit messy

TBonnin (commented):
one question about the auth mode name but otherwise looks good to me

scripts/validation/providers/schema.json
```diff
@@ -295,6 +296,10 @@
                     "type": "string",
                     "enum": ["HMAC-SHA1", "PLAINTEXT"]
                 },
+                "signature_protocol": {
+                    "type": "string",
+                    "enum": ["WSSE"]
+                },
```
TBonnin, line 308: what about a signature top level attributes with a `protocol` signature?
```
auth_mode: SIGNATURE
signature: 
   protocol: WSSE
```
just in case we need to add another signature related attribute in the future

docs-v2/reference/api-configuration.mdx
hassan254-prog: I have updated this with the latest [commit](https://github.com/NangoHQ/nango/pull/2942/commits/b4d3dd0ffcd24f1188b251597107f09a4e90eb13)

packages/shared/lib/services/connection.service.ts
hassan254-prog: I have updated this with the latest [commit](https://github.com/NangoHQ/nango/pull/2942/commits/b4d3dd0ffcd24f1188b251597107f09a4e90eb13)

scripts/validation/providers/schema.json
hassan254-prog: I have updated this with the latest [commit](https://github.com/NangoHQ/nango/pull/2942/commits/b4d3dd0ffcd24f1188b251597107f09a4e90eb13)

packages/shared/lib/services/connection.service.ts
hassan254-prog: I have refactored the upsert functions so that we can have one that will upsert, `BASIC`, `API_KEY`, `JWT`, `BILL`, `TABLEAU`, `SIGNATURE`, `TWO_STEP` and `TBA` connections in the latest [commit](https://github.com/NangoHQ/nango/pull/2942/commits/b4d3dd0ffcd24f1188b251597107f09a4e90eb13)

bodinsamuel (approved):
Looks good, might have been even better to reuse basic but not sure 👌🏻

TBonnin (approved):
lgtm. Same question: could we reuse BASIC auth mode?

hassan254-prog:
> lgtm. Same question: could we reuse BASIC auth mode?

To maintain the properties of the `BASIC` auth_mode, specifically, the absence of the token generated, expire_at, and refresh attributes found in the `SIGNATURE` auth_mode. Also to ensure a cleaner approach that improves DX.

PR #2944, opened 2024-11-05
feat(integrations): add support for emarsys core api

hassan254-prog:
## Describe your changes

- Add support for Emarsys Core API

[EXT-206](https://linear.app/nango/issue/EXT-206/add-emarsys-support)


packages/shared/providers.yaml
```diff
@@ -1814,6 +1814,24 @@ dropbox:
...
+        base_url: https://api.emarsys.net
+        headers:
+            Content-Type: application/json
+        verification:
+            method: GET
+            endpoint: /api/v2/settings
+        retry:
+            at: 'X-RateLimit-Reset'
+    token:
+        expires_in_ms: 300000
```
bodinsamuel, line 1871: How do a long sync behave when a token expires during execution?

bodinsamuel (approved):

packages/shared/providers.yaml
hassan254-prog: For every call, Nango checks `credentials.expires_at` to determine whether the credentials should be refreshed before making the call. This ensures that tokens are always fresh before the actual request to the API.

PR #2945, opened 2024-11-05
fix(jobs): use new docker image

bodinsamuel:
## Describe your changes

- Jobs Use new docker image 
Already deployed in staging, nb: requires manual change before deployment on prod


Dockerfile
```diff
@@ -91,12 +91,15 @@ RUN true \
   && apt-get purge -y --auto-remove -o APT::AutoRemove::RecommendsImportant=false
 
 # Do not use root to run the app
-USER node
+# BUT it does not work with secret mount (could not find a solution yet)
```
TBonnin, line 93: when do we need secret mount?

TBonnin (commented):
I don't remember why jobs is not yet using the unified image. Do you remember? @bodinsamuel

Dockerfile
bodinsamuel: To load Google credentials for bigquery

bodinsamuel:
> I don't remember why jobs is not yet using the unified image.

No big reasons, it was just a matter of taking the time

Dockerfile
TBonnin: do we need to find a solution before merging this change?

bodinsamuel: nope, just running as root is good, <i just left the comment to be sure the next person is not trying to put an user back

TBonnin (approved):

PR #2951, opened 2024-11-06
fix(cli): pin all internal packages

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-2088/cli-pin-internal-packages

- Pin all internal packages
We recently run into an issue where people want to downgrade but they can't because npm will happily install internal packages at a higher version thanks to `^`. So it's working as intended but we don't want that since our API is unstable.

- Add `version.ts` in CLI
As we are closer and closer to being able to remove shared from the CLI, I'm taking the opportunity to move this critical part in the CLI


TBonnin (approved):

bodinsamuel:
Fixed an issue with path using a node >20.11.0 syntax

khaliqgant (approved):

PR #2952, opened 2024-11-06
fix: upsert summary race condition

TBonnin:
Doing a first query to calculate which records will be inserted/updated and then upserting them could lead to a race condition when multiple concurrent executions would assume there are about to insert the same record but only the first one to execute the second query is actually doing the insertion. The upsert summary 'added' value would then be wrong, effectively counting the same records multiple times. This is then affecting the records_count table with a saved count higher than the actual count
It is unlikely to happen unless `batchSave` is executed concurrently which we are not recommending but in theory it does happen since there is about a dozen rows in records_count table that are slightly higher than the real counts.

The fix consist in adding a advisory lock to prevent concurrent upsert by the same sync job and a single CTE query comparing before/after to determine which records are inserted/updated/deleted/undeleted

Required to fix discrepancies between real records count and the records_count table
https://linear.app/nango/issue/NAN-1929/[trial]-show-object-count-in-sync-tab

### DIFF
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -6,7 +6,7 @@ import { decryptRecordData, encryptRecords } from '../utils/encryption.js';
 import { RECORDS_TABLE, RECORD_COUNTS_TABLE } from '../constants.js';
 import { removeDuplicateKey, getUniqueId } from '../helpers/uniqueKey.js';
 import { logger } from '../utils/logger.js';
-import { Err, Ok, retry } from '@nangohq/utils';
+import { Err, Ok, retry, stringToHash } from '@nangohq/utils';
 import type { Result } from '@nangohq/utils';
 import type { Knex } from 'knex';
 
@@ -200,14 +200,12 @@ export async function upsert({
     connectionId,
     environmentId,
     model,
-    syncJobId,
     softDelete = false
 }: {
     records: FormattedRecord[];
     connectionId: number;
     environmentId: number;
     model: string;
-    syncJobId: number;
     softDelete?: boolean;
 }): Promise<Result<UpsertSummary>> {
     const { records: recordsWithoutDuplicates, nonUniqueKeys } = removeDuplicateKey(records);
@@ -221,8 +219,8 @@ export async function upsert({
     const summary: UpsertSummary = { addedKeys: [], updatedKeys: [], deletedKeys: [], nonUniqueKeys };
     try {
         await db.transaction(async (trx) => {
-            // Lock on sync job level to prevent concurrent upserts
-            await trx.raw(`SELECT pg_advisory_xact_lock(?) as lock_records_upsert`, [syncJobId]);
+            // Lock to prevent concurrent upserts
+            await trx.raw(`SELECT pg_advisory_xact_lock(?) as lock_records_upsert`, [newLockId(connectionId, model)]);
             for (let i = 0; i < recordsWithoutDuplicates.length; i += BATCH_SIZE) {
                 const chunk = recordsWithoutDuplicates.slice(i, i + BATCH_SIZE);
                 const encryptedRecords = encryptRecords(chunk);
@@ -339,13 +337,11 @@ export async function upsert({
 export async function update({
     records,
     connectionId,
-    model,
-    syncJobId
+    model
 }: {
     records: FormattedRecord[];
     connectionId: number;
     model: string;
-    syncJobId: number;
 }): Promise<Result<UpsertSummary>> {
     const { records: recordsWithoutDuplicates, nonUniqueKeys } = removeDuplicateKey(records);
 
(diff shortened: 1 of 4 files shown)


packages/records/lib/models/records.ts
```diff
@@ -216,53 +216,90 @@ export async function upsert({
...
+
+                // then update records that were not inserted
+                const updateQuery = trx
+                    .with('update', (qb) => {
+                        qb.insert(encryptedRecords)
+                            .into(RECORDS_TABLE)
+                            .onConflict(['connection_id', 'external_id', 'model'])
+                            .merge()
+                            .whereNotIn(`${RECORDS_TABLE}.external_id`, inserted)
+                            .returning('external_id');
```
TBonnin, line 266: the upsert is now 2 queries but the initial query to determine the counts is not there anymore

packages/records/lib/models/records.ts
```diff
@@ -216,53 +216,95 @@ export async function upsert({
...
+                                            WHEN existing.deleted_at IS NOT NULL AND upsert.deleted_at IS NULL THEN 'undeleted'
+                                            WHEN existing.deleted_at IS NULL AND upsert.deleted_at IS NOT NULL THEN 'deleted'
+                                            WHEN existing.data_hash <> upsert.data_hash THEN 'changed'
+                                            ELSE 'unchanged'
+                                        END
+                                END as status`)
+                    )
+                    .from('upsert')
+                    .leftJoin('existing', 'upsert.external_id', 'existing.external_id');
+                const updatedRes = await withRetry(updateQuery);
```
TBonnin, line 280: the query is more complicated for sure but we avoid the initial query in `getUpsertSummary`

packages/records/lib/models/records.ts
```diff
@@ -216,53 +216,97 @@ export async function upsert({
...
+                const updateQuery = trx
+                    .with('existing', (qb) => {
+                        qb.select('external_id', 'data_hash', 'deleted_at')
+                            .from(RECORDS_TABLE)
+                            .where({
+                                connection_id: connectionId,
+                                model
+                            })
+                            .whereIn('external_id', existingIds)
+                            .forUpdate();
```
bodinsamuel: note: this is similar to an advisory lock in the end since you lock all the rows. 
I'm wondering about:
- the perf of locking all rows like this vs one lock.
- if it's working since you can only lock rows that exist. if a webhook is running concurrently the two operations could potentially execute at the same time, would only be an issue with record creation maybe I don't know

packages/records/lib/models/records.ts
```diff
@@ -216,53 +216,97 @@ export async function upsert({
...
+                                CASE
+                                    WHEN existing.external_id IS NULL THEN 'inserted'
+                                    ELSE
+                                        CASE
+                                            WHEN existing.deleted_at IS NOT NULL AND upsert.deleted_at IS NULL THEN 'undeleted'
+                                            WHEN existing.deleted_at IS NULL AND upsert.deleted_at IS NOT NULL THEN 'deleted'
+                                            WHEN existing.data_hash <> upsert.data_hash THEN 'changed'
+                                            ELSE 'unchanged'
+                                        END
+                                END as status`)
```
bodinsamuel, line 276: Found this https://stackoverflow.com/a/38858662
You might already know this since it's very similar to this PR, but could simplify the update/insert query and delete could be done in a second query without the need to lock all rows

bodinsamuel (approved):
Thanks for tackling this, looks okay in theory so I'll approve but I have two open comments

packages/records/lib/models/records.ts
TBonnin: It is locking indeed. There is no real way around it but as you mentioned an explicit advisory lock could be subject to deadlock or connection pool congestion. 
Locking the rows that are about to be modified is not a bad idea I think

Regarding you second point, yes you are right. In case of concurrent records creation, the first one to occur is going to win but whatever strategy we decide to implement we always gonna have to make a choice between one or the other

packages/records/lib/models/records.ts
TBonnin: I have tried to make `xmax` work but without success. I am always getting errors like `cannot retrieve a system column in this context`

bodinsamuel: arg too bad, seems like it was a good solution

packages/records/lib/models/records.ts
bodinsamuel: I think my point boils down to this: it will use the same number of connections since it's also a global lock but in a less explicit way

TBonnin: I am fine using an advisory lock if this is what you prefer

bodinsamuel: I'm not really preferring just thinking out loud. Lock would prevent webhook and sync from executing at the same time and achieve the same thing. If you think it's too not relevant feel free to discard.
We could use Redis here to save pg connection. 

Just noticed we were not using forUpdate in `getRecordsToUpdate` which also could simply explain the issue

bodinsamuel (commented):
sorry I just noticed my comment was not sent

packages/records/lib/models/records.ts
```diff
@@ -200,12 +200,14 @@ export async function upsert({
...
     environmentId,
     model,
+    syncJobId,
     softDelete = false
 }: {
     records: FormattedRecord[];
     connectionId: number;
     environmentId: number;
     model: string;
+    syncJobId: number;
```
bodinsamuel: One last comment, we create a sync job for each webhook. ConnectionId or syncId might be better suited maybe

bodinsamuel (approved):

packages/records/lib/models/records.ts
TBonnin: `forUpdate` doesn't prevent other query to read the rows so if it is happening in a different query it doesn't prevent the race condition.

packages/records/lib/models/records.ts
TBonnin: the original goal was to prevent race condition from the same job execution but you are right that webhooks can affect the same dataset. 
The challenge of webhook concurrency is a whole other topic but since we are trying to prevent race condition let's have a lock on connection/model. I'll change that on Tuesday

PR #2957, opened 2024-11-07
feat(api): enforce max cli version with flag

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-2097/enforce-maximum-cli-version-on-public-api

> [!NOTE]
> This is a proposal

- Enforce a maximum CLI version tied to the backend version
For enterprise only because they are pinned to a version for months, and it's easy to make the mistake of upgrading the CLI but not the backend.

Screenshot 2024-11-07 at 12 24 58


TBonnin (approved):
I like the idea, especially for enterprise customers.

packages/server/lib/middleware/cliVersionCheck.ts
```diff
@@ -33,3 +34,35 @@ export function cliMinVersion(minVersion: string) {
...
+        }
+
+        const match = userAgent.match(VERSION_REGEX);
+        if (!match || match.length <= 1 || !match[1]) {
+            // Could be strictly enforced
+            next();
+            return;
+        }
+
+        console.log(match[1], semver.gt(match[1], NANGO_VERSION));
```
TBonnin: 🪓

bodinsamuel:
As discussed with bastien, I might try to enforce only minor versions instead of patch so we can still provide patches for enterprise (eventually) <=0.42.xxx

TBonnin:
> As discussed with bastien, I might try to enforce only minor versions instead of patch so we can still provide patches for enterprise (eventually) <=0.42.xxx

It would then only be useful if we follow proper semantic versioning, which limits a lot the safeguard capability this change is supposed to provide

bodinsamuel:
> It would then only be useful if we follow proper semantic versioning, which limits a lot the safeguard capability this change is supposed to provide

Yeah that's my fear, also I'm not sure we have a way to backfill patches so it's limited.

khaliqgant (approved):
We should start to follow proper semantic versioning in this case.

bodinsamuel:
I need to think about it a bit more, or maybe just optin for enterprise

packages/server/lib/middleware/cliVersionCheck.ts
```diff
@@ -24,10 +26,42 @@ export function cliMinVersion(minVersion: string) {
 
         if (semver.gt(minVersion, match[1])) {
             logger.info(`This endpoint requires a CLI version >= ${minVersion} (current: ${match[0]})`);
-            // res.status(400).send({
-            //     error: { code: 'invalid_cli_version', message: `This endpoint requires a CLI version >= ${minVersion} (current: ${match[0]})` }
-            // });
-            // return;
```
bodinsamuel, line 30: I forgot to ever enforce it, now it's safe ahah

packages/utils/lib/environment/detection.ts
```diff
@@ -28,3 +28,4 @@ export const flagHasBigQuery = process.env['FLAG_BIG_QUERY_EXPORT_ENABLED'] ===
 export const flagHasSlack = !isHosted;
 export const flagHasPlan = process.env['FLAG_PLAN_ENABLED'] === 'true';
 export const flagHasUsage = process.env['FLAG_USAGE_ENABLED'] === 'true';
+export const flagEnforceCLIVersion = process.env['FLAG_ENFORCE_CLI_VERSION'] === 'true';
```
bodinsamuel, line 31: Note we'll need to add to the enterprise self-hosting default var

bodinsamuel:
Changed to be opt-in only so we can enabled it only for enterprise

TBonnin (approved):
👌

PR #2958, opened 2024-11-07
feat(nango-yaml): support new endpoint format

bodinsamuel:
## Describe your changes

Contributes https://linear.app/nango/issue/NAN-1934/specify-script-endpoint-section-in-nangoyaml

- Introduce a new format in nango-yaml
The parsing was done in a previous PR we can now support and send the new format. It's replacing the old format but I don't think we'll get rid of this one soon.
NB: I also added support for entity but it's not yet store

- Add a new command to migrate for simplicity's sake
`nango migrate-endpoints`

- Fixed a bug with very old configs without endpoints


khaliqgant (approved):
looks good

bodinsamuel:
@TBonnin any comment on this PR?

packages/cli/lib/services/migration.service.ts
```diff
@@ -118,3 +118,24 @@ export const directoryMigration = async (loadLocation: string, debug?: boolean):
...
+        return;
+    }
+
+    if (response.parsed.yamlVersion !== 'v2') {
+        console.log(chalk.red(`nango.yaml is not at v2. New endpoint format is only supported in V2`));
+        return;
+    }
+
+    let dump = response.yaml.replace(
+        /^(\s+)endpoint: ((GET|POST|PUT|PATCH|DELETE)\s)?(\/[a-zA-Z0-9-:{}./_]+)$/gim,
```
TBonnin, line 134: can we make this a const and reuse it in the replace below?

TBonnin (approved):

packages/cli/lib/services/migration.service.ts
bodinsamuel: it's not the same regex, but I'll split it 🆗

bodinsamuel: Cant f making it work, anyway since it's two different regex it's fine (but I would love to understand why concatenation don't work if you have an idea)
```diff
diff --git a/packages/cli/lib/services/migration.service.ts b/packages/cli/lib/services/migration.service.ts
index 82e42d349..9b1190d39 100644
--- a/packages/cli/lib/services/migration.service.ts
+++ b/packages/cli/lib/services/migration.service.ts
@@ -1,3 +1,4 @@
+/* eslint-disable no-useless-escape */
 import fs, { writeFileSync } from 'fs';
 import path from 'path';
 import chalk from 'chalk';
@@ -119,6 +120,7 @@ export const directoryMigration = async (loadLocation: string, debug?: boolean):
     console.log(chalk.green(`Migration to nested directories complete.`));
 };

+const endpointRegex = `((GET|POST|PUT|PATCH|DELETE)\s)?(/[a-zA-Z0-9-:{}./_]+)`;
 export function endpointMigration(loadLocation: string): void {
     const { response, success } = parse(loadLocation);
     if (!success || !response?.parsed) {
@@ -130,11 +132,8 @@ export function endpointMigration(loadLocation: string): void {
         return;
     }

-    let dump = response.yaml.replace(
-        /^(\s+)endpoint: ((GET|POST|PUT|PATCH|DELETE)\s)?(\/[a-zA-Z0-9-:{}./_]+)$/gim,
-        `$1endpoint:\r\n$1  method: $3\r\n$1  path: $4`
-    );
-    dump = dump.replace(/^(\s+)- ((GET|POST|PUT|PATCH|DELETE)\s)?(\/[a-zA-Z0-9-:{}./_]+)$/gim, `$1- method: $3\r\n$1  path: $4`);
+    let dump = response.yaml.replace(new RegExp(`^(\s+)endpoint: ${endpointRegex}$`, 'gim'), `$1endpoint:\r\n$1  method: $3\r\n$1  path: $4`);
+    dump = dump.replace(new RegExp(`(\s+)- ${endpointRegex}$`, 'gim'), `$1- method: $3\r\n$1  path: $4`);

     writeFileSync(`${loadLocation}/nango.yaml`, dump);
     console.log(chalk.green(`Migration complete.`));
```

TBonnin: not sure. do you need to escape backslashes when using `new Regex`? ie: `^(\\s+)endpoint...`

PR #2960, opened 2024-11-07
feat(okta): [nan-2124] add okta preview

khaliqgant:
## Describe your changes
Add in okta preview

### DIFF
--- a/packages/shared/providers.yaml
+++ b/packages/shared/providers.yaml
@@ -3931,22 +3931,12 @@ okta:
             suffix: .okta.com
 
 okta-preview:
+    alias: okta
     display_name: Okta (Preview)
-    auth_mode: OAUTH2
     authorization_url: https://${connectionConfig.subdomain}.oktapreview.com/oauth2/v1/authorize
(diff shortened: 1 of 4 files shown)


packages/shared/providers.yaml
```diff
@@ -3930,6 +3930,32 @@ okta:
...
+    docs: https://docs.nango.dev/integrations/all/okta
+    connection_config:
+        subdomain:
+            type: string
+            title: Okta Preview Domain
+            description: The subdomain of your Okta Preview account
+            pattern: '^[a-z0-9_-]+$'
+            example: domain
+            suffix: .oktapreview.com
+
```
TBonnin, line 3949: cannot it not be an alias of okta just overriding the urls?

khaliqgant: Yup https://github.com/NangoHQ/nango/pull/2960/commits/25d01d3a9b91afc813b7d8b8b72497a1395da419

TBonnin (approved):

PR #2961, opened 2024-11-07
feat: return records count in /sync/status endpoint/sdk

TBonnin:
Screenshot 2024-11-07 at 14 34 38

https://linear.app/nango/issue/NAN-1928/return-sync-object-count-from-api

### DIFF
--- a/packages/shared/lib/services/sync/manager.service.ts
+++ b/packages/shared/lib/services/sync/manager.service.ts
@@ -468,7 +468,7 @@ export class SyncManagerService {
             latestExecutionStatus: latestJob?.status,
             recordCount,
             ...(includeJobStatus ? { jobStatus: latestJob?.status as SyncStatus } : {})
-        } as ReportedSyncJobStatus;
+        };
     }
 }
 
--- a/packages/shared/lib/services/sync/manager.service.ts
+++ b/packages/shared/lib/services/sync/manager.service.ts
@@ -445,7 +445,7 @@ export class SyncManagerService {
 
         const countRes = await recordsService.getRecordCountsByModel({ connectionId: sync.nango_connection_id, environmentId });
         if (countRes.isErr()) {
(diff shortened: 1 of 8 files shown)


packages/shared/lib/services/sync/manager.service.ts
```diff
@@ -432,13 +435,27 @@ export class SyncManagerService {
...
-            const syncConfig = await getSyncConfigByParams(environmentId, sync.name, providerConfigKey);
             frequency = syncConfig?.runs || null;
         }
         if (!schedule) {
             throw new Error(`Schedule for sync ${sync.id} and environment ${environmentId} not found`);
         }
+
+        const countRes = await recordsService.getRecordCountsByModel({ connectionId: sync.nango_connection_id, environmentId });
+        if (countRes.isErr()) {
+            logger.error(`Failed to get record count for sync ${sync.id} in environment ${environmentId}: ${stringifyError(countRes.error)}`);
```
bodinsamuel: do we want this to be optional? Some customers might build logic around those numbers and if they goes to zero suddenly could be problematic

packages/shared/lib/services/sync/manager.service.ts
```diff
@@ -449,6 +466,7 @@ export class SyncManagerService {
             frequency,
             latestResult: latestJob?.result,
             latestExecutionStatus: latestJob?.status,
+            recordCount,
             ...(includeJobStatus ? { jobStatus: latestJob?.status as SyncStatus } : {})
         } as ReportedSyncJobStatus;
```
bodinsamuel: type was not updated. 
not your fault but the usage of `as` here is problematic (as always). I just tried to replace it with `satisfies` and there are more errors :/

bodinsamuel (approved):

packages/shared/lib/services/sync/manager.service.ts
TBonnin: good catch. I removed the casting and fixed the type. There might be underlying issue with the type of some attributes like the `type` one but I didn't want to change the behavior in this PR and at least now the types match the behavior

PR #2966, opened 2024-11-08
feat(endpoints): store and use "group“

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1934/specify-script-endpoint-section-in-nangoyaml

- Store entity
- Use it in the UI to categorize endpoints

*cf my own nango.yaml* 
Screenshot 2024-11-08 at 15 06 34


docs-v2/reference/integration-configuration.mdx
```diff
@@ -20,6 +20,7 @@ integrations:
         endpoint: # Generated endpoint to access the synced model.
           method: GET
           path: /tasks
+          entity: Tasks
```
TBonnin: Not a huge fan of the name `entity`. I don't find it super descriptive. what about `group` or `tag`?

packages/webapp/src/pages/Integrations/providerConfigKey/Endpoints/Show.tsx
```diff
@@ -39,12 +25,7 @@ export const EndpointsShow: React.FC<{ integration: GetIntegration['Success']['d
         const tmp: Record<string, NangoSyncConfigWithEndpoint[]> = {};
         for (const flow of data.flows) {
             for (const endpoint of flow.endpoints) {
-                const paths = endpoint.path.split('/');
-                const firstPath = paths[1];
-                if (!firstPath) {
-                    continue;
-                }
-                const groupName = allowedGroup.includes(firstPath) ? firstPath : 'others';
+                const groupName = endpoint.entity || 'others';
```
TBonnin: should we have a migration query to populate the field with the first part of the path for existing endpoints in order to recreate the current behavior?

TBonnin (approved):

docs-v2/reference/integration-configuration.mdx
bodinsamuel: I don't either, not my call though. @khaliqgant

packages/webapp/src/pages/Integrations/providerConfigKey/Endpoints/Show.tsx
bodinsamuel: mmh good point. We have planned a task to backfill all templates and customers, I don't think we need to go further  honestly but cc @bastienbeurier

bastienbeurier: I think we're okay waiting for the backfill, unless @khaliqgant disagrees.

docs-v2/reference/integration-configuration.mdx
bastienbeurier: CleanShot 2024-11-13 at 17 01 33@2x

Tags seem to be the way OpenAPI calls this. It's also a good idea to make it an array, though we don't need this now.

https://swagger.io/docs/specification/v3_0/grouping-operations-with-tags/

bodinsamuel: indeed tags plural could be "useful" but realistically I have never seen an API needing that and it doesn't convey the same meaning. 
By order of pref: Group, Entity, Tag

packages/webapp/src/pages/Integrations/providerConfigKey/Endpoints/Show.tsx
khaliqgant: All good with waiting for the backfill

docs-v2/reference/integration-configuration.mdx
bodinsamuel: Renamed to group

PR #2976, opened 2024-11-12
fix(cli): clean up some references to shared and unused code

bodinsamuel:
## Describe your changes

Contributes to https://linear.app/nango/issue/NAN-358/cli-upgrade-is-taking-too-long

- Clean up some references to `shared` 
Down to like 5 imports but annoying ones: the SDK (I think we should move that to a published npm package), NangoError (not portable but we also know it's time to get rid of this).

- Remove some unused code
I specifically removed the staging code because I don't think it's useful, you just need to change the `.env`, let me know.


packages/cli/lib/services/deploy.service.ts
```diff
@@ -34,10 +33,7 @@ class DeployService {
         if (!process.env['NANGO_HOSTPORT']) {
             switch (environmentName) {
                 case 'local':
-                    process.env['NANGO_HOSTPORT'] = `http://localhost:${port}`;
-                    break;
-                case 'staging':
-                    process.env['NANGO_HOSTPORT'] = stagingHost;
+                    process.env['NANGO_HOSTPORT'] = localhostUrl;
```
TBonnin, line 36: localhostUrl for `staging` case?

packages/cli/lib/types.ts
```diff
@@ -3,12 +3,11 @@ export interface GlobalOptions {
     debug: boolean;
 }
 
-export type ENV = 'local' | 'staging' | 'production';
+export type ENV = 'local' | 'cloud';
```
TBonnin, line 6: does this change affect Enterprise or self-hosted installs?

packages/types/lib/index.ts
```diff
@@ -38,6 +38,7 @@ export type * from './auth/api.js';
 export type * from './auth/db.js';
 export type * from './auth/http.api.js';
 export type * from './deploy/api.js';
+export type * from './deploy/results.js';
```
TBonnin: I would have a single file name `./deploy/index.js`. I doubt `results` would ever have much more content

TBonnin (approved):

packages/cli/lib/services/deploy.service.ts
bodinsamuel: staging case was deleted

packages/cli/lib/types.ts
bodinsamuel: nope, it's useless, we need to remove that entirely and just enforce NANGO_HOSTPORT

PR #2978, opened 2024-11-12
feat(api): Expose providers.yaml through server

nalanj:
## Describe your changes

See https://linear.app/nango/issue/NAN-2105/dynamically-reload-providersyaml

Adds an unsecured endpoint `/providers.json` that serves up the providers.yaml file data as JSON.

A future PR will add dynamic reloading to runners.


packages/shared/lib/services/providers.ts
```diff
@@ -39,7 +40,7 @@ function getProvidersPath() {
     return findProvidersYaml(dirname());
 }
 
-function loadProvidersYaml(): Record<string, Provider> | undefined {
+export function loadProvidersYaml(): Record<string, Provider> | undefined {
```
TBonnin: does this need to be exported?

packages/server/lib/controllers/v1/getProvidersJSON.ts
```diff
@@ -0,0 +1,13 @@
+import { getProviders } from '@nangohq/shared';
+import { asyncWrapper } from '../../utils/asyncWrapper.js';
+
+let providersJSON: string | undefined = undefined;
+
+export const getProvidersJSON = asyncWrapper<any, any>((_, res) => {
+    if (!providersJSON) {
+        providersJSON = JSON.stringify(getProviders());
+    }
```
TBonnin, line 9: the providers content is already cached in memory. Is it worth caching the json as well? How slow/fast is the stringifying?

TBonnin (approved):

packages/shared/lib/services/providers.ts
nalanj: Oh nope, that was from a previous idea of how to do this. Removing!

nalanj: Fixed in cb23febc

packages/server/lib/controllers/v1/getProvidersJSON.ts
nalanj: Do you mean caching in `shared` like we do with the providers data itself? This is caching the data in memory in the controller here. 

The reason I put this here is that I wasn't sure if we wanted to have that extra chunk of memory always in use by anything that uses providers from shared. It's currently about 190kb, so it's not like it's too big of a deal, but I figured we didn't want side effects that like that consistently leaking out of `shared`.

It takes around 500-700 microseconds to stringify the JSON right now and only has to happen the first time this controller is called, so I don't think it's too big of a deal.

TBonnin: up to you. It doesn't seem to be a big deal either way. 🚢

bodinsamuel: All for caching, but if you do it, you will need to deal with cache invalidation in the next PR 
Here you have no way to clear the cache easily

bodinsamuel (approved):

packages/server/lib/controllers/v1/getProvidersJSON.ts
TBonnin: server restart/deployment would "invalidate" the cache.

bodinsamuel: ah yes true we rely on the deploy, then fine 👌🏻

PR #2980, opened 2024-11-13
fix: deprecate publicKey and HMAC

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-2155/replace-hmac-and-public-api-key-by-session-token

- **Docs:** replace all reference to public key and HMAC by session token
This has been validated by Bastien.
- **UI**: add deprecation warning on public key and HMAC settings
- **Server**: make the type and validation stricter around either public_key or session_token
- **Frontend**: make the type and validation stricter around either public_key or session_token
Also only check credentials at call time because they can be passed asynchronously 
- **Fix** HMAC check that was sending headers on failure but did not stop controller execution
- **Fix** TBA that was missing try/catch + connectionCreationFailedHook
- **Drop** AuthLink which seems unused for many months now and was exposing insecure fonctionnallities 
Please let me know if that's alright


packages/frontend/lib/index.ts
```diff
@@ -78,13 +87,9 @@ export default class Nango {
             this.height = config.height;
         }
 
-        this.hostBaseUrl = config.host.slice(-1) === '/' ? config.host.slice(0, -1) : config.host; // Remove trailing slash.
+        this.hostBaseUrl = config.host.endsWith('/') ? config.host.slice(0, -1) : config.host; // Remove trailing slash.
```
TBonnin: can this be a simple regex `config.host.replace(/\/+$/, '');`

TBonnin (approved):
lgtm. could have been split into smaller PRs imho

PR #2982, opened 2024-11-13
fix(providers): fix whatsapp apiKey regex

hassan254-prog:
## Describe your changes

- Fix whatsapp apiKey regex


bodinsamuel (approved):
Thanks 🙏🏻

PR #2983, opened 2024-11-13
feat(jobs): Pass PROVIDERS_URL and PROVIDERS_RELOAD_INTERVAL to runners

nalanj:
## Describe your changes

Sets up needed ENV vars for providers reloading to render runner and local runner.

@TBonnin No env vars were set in `RemoteRunner`, but it wasn't completely clear to me how `RemoteRunner` works at this point. I may need help figuring out how it works.

See https://linear.app/nango/issue/NAN-2105/dynamically-reload-providersyaml


TBonnin:
Remote runner is for Enterprise customer who are running Nango in their own infrastructure. In this case we don't control where the runner(s) are. 
We should document this env var in case Enterprise customers want to also use the dynamic url (@khaliqgant do we have an Enterprise setup guide somewhere?) but we should also make sure that the providers data is loaded from binary if this env var is not specified. In a single tenant enterprise setup, dynamic loading of providers isn't bringing much value anyway

packages/jobs/lib/runner/render.runner.ts
```diff
@@ -79,7 +80,9 @@ export class RenderRunner implements Runner {
                         { key: 'NANGO_TELEMETRY_SDK', value: process.env['NANGO_TELEMETRY_SDK'] || 'false' },
                         { key: 'DD_ENV', value: process.env['DD_ENV'] || '' },
                         { key: 'DD_SITE', value: process.env['DD_SITE'] || '' },
-                        { key: 'DD_TRACE_AGENT_URL', value: process.env['DD_TRACE_AGENT_URL'] || '' }
+                        { key: 'DD_TRACE_AGENT_URL', value: process.env['DD_TRACE_AGENT_URL'] || '' },
+                        { key: 'PROVIDERS_URL', value: getProvidersUrl() },
+                        { key: 'PROVIDERS_RELOAD_INTERVAL', value: envs.PROVIDERS_RELOAD_INTERVAL.toString() }
```
TBonnin, line 85: Note: the env vars are only passed when the runner is created, which means we are gonna need to backfill those values for all existing runners. I think I have done it once in the past, not sure I pushed the script though. I remember the Render api around env vars was pretty awkward

TBonnin (approved):

bodinsamuel (approved):

packages/jobs/lib/runner/render.runner.ts
nalanj: Wrote up a script here: https://github.com/NangoHQ/nango/pull/2984

PR #2984, opened 2024-11-13
feat(script): Script for rolling out env vars to render runners

nalanj:
## Describe your changes

To roll out providers updating in runners we needed to be able to ship some new env vars to existing runners. I went ahead and ran this against staging with the new variables we needed to ensure it was working correctly.

Usage:

```
ENVIRONMENT=staging \
OWNER_ID=<OWNER_ID> \
RENDER_API_KEY=<RENDER_API_KEY> \
KEY=PROVIDERS_RELOAD_INTERVAL \
VALUE=60000 \
node ./scripts/runner-env.js
```

https://linear.app/nango/issue/NAN-2105/dynamically-reload-providersyaml


scripts/runner-env.js
```diff
@@ -0,0 +1,108 @@
...
+function sleep(ms) {
+    return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+async function fetchRunners() {
+    let services = [];
+    let cursor = '';
+    let running = true;
+
+    while (running) {
```
TBonnin: why not `while(true)` and `break` without the `running` var?

scripts/runner-env.js
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env node
+
+// Update environment variables for all runners in the given environment
+
+const apiUrl = 'https://api.render.com/v1';
+
+if (!process.env.ENVIRONMENT) {
+    console.error('Please set the ENVIRONMENT environment variable');
```
TBonnin: what about a `help()` function showing how to use the script?

scripts/runner-env.js
```diff
@@ -0,0 +1,108 @@
+#!/usr/bin/env node
```
TBonnin: can we call the file `runner-update-env.js`?

scripts/runner-env.js
```diff
@@ -0,0 +1,108 @@
...
+            running = false;
+            break;
+        }
+
+        cursor = fetched[fetched.length - 1].cursor;
+        console.log(cursor);
+
+        const filteredServices = fetched.map((item) => item.service).filter((service) => service.name.startsWith(`${process.env.ENVIRONMENT}-runner-`));
+
+        services = services.concat(...filteredServices);
```
TBonnin: is it not `services.concat(filteredServices)` or `services.push(...filteredServices)`

TBonnin (approved):

scripts/runner-env.js
```diff
@@ -0,0 +1,108 @@
...
+        const fetched = await response.json();
+        if (fetched.length === 0) {
+            running = false;
+            break;
+        }
+
+        cursor = fetched[fetched.length - 1].cursor;
+        console.log(cursor);
+
+        const filteredServices = fetched.map((item) => item.service).filter((service) => service.name.startsWith(`${process.env.ENVIRONMENT}-runner-`));
```
TBonnin: is it the linter keeping everything on one line?

scripts/runner-env.js
nalanj: Indeed

scripts/runner-env.js
nalanj: Sure sure, incoming.

scripts/runner-env.js
nalanj: That was what I had initially and eslint was mad at me. I'll update and put a rule exception there because I think it's easier to read that way anyway.

scripts/runner-env.js
nalanj: You're right - but also concat just appends anything that isn't an array so it ended up with the same result. Good catch.

scripts/runner-env.js
nalanj: Yes. Prettier is configured to 160 chars so it's going nuts there. I'll see what I can do.

TBonnin: don't worry. not a big deal

scripts/runner-env.js
TBonnin: > concat just appends anything that isn't an array

🙄 🙄 🙄

scripts/runner-env.js
nalanj: Fixed in b671438e

nalanj: Fixed in b671438e

scripts/runner-env.js
nalanj: Fixed in 72654332

scripts/runner-env.js
nalanj: Fixed in 8338a3b6

scripts/runner-env.js
nalanj: Fixed in 8338a3b

scripts/runner-env.js
nalanj: Fixed in 17b64d01

bodinsamuel (approved):

PR #2985, opened 2024-11-13
fix(deploy runner): for doesn't split json lines properly

TBonnin:
It led to some of the runners not being deployed

I have checked that all running runners are now being processed

https://linear.app/nango/issue/NAN-1861/3-runners-are-not-being-deployed-when-deploying-runners


khaliqgant (approved):

scripts/deploy/runners.bash
```diff
@@ -24,7 +24,10 @@ while true; do
   parsed_response=$(echo "$response" | jq -c '.[]' 2>/dev/null)
   CURSOR=$(echo "$response" | jq -r '.[-1].cursor' 2>/dev/null)
 
-  for item in $parsed_response; do
```
TBonnin, line 27: this was the problematic line, not splitting the json lines correctly

PR #2988, opened 2024-11-14
fix(scripts): Add more resilient rate limit handling in runner-update-env

nalanj:
## Describe your changes

I ran into an issue while rolling out the updated env vars that I bumped into rate limits with Render against our prod data that I hadn't bumped into in staging due to the lower number of runners in staging.

This PR updates the network calls to sleep against their rate limit so that they'll run fast but not too fast.

### DIFF
--- a/scripts/runner-update-env.js
+++ b/scripts/runner-update-env.js
@@ -73,6 +73,10 @@ async function fetchRunners() {
 
         services = services.concat(filteredServices);
 
+        if (!response.headers.get('ratelimit-remaining') || !response.headers.get('ratelimit-reset')) {
+            throw new Error('Unexpected API format: no rate limit headers');
+        }
+
         const remaining = parseInt(response.headers.get('ratelimit-remaining'));
         const resetMs = (parseInt(response.headers.get('ratelimit-reset')) + 1) * 1000;
 
@@ -98,6 +102,10 @@ async function updateEnvVar(serviceId, key, value) {
     if (!response.ok) {
         const body = await response.text();
         throw new Error(`Failed to update env var ${key} for service ${serviceId}\n${body}`);
+    }
+
+    if (!response.headers.get('ratelimit-remaining') || !response.headers.get('ratelimit-reset')) {
+        throw new Error('Unexpected API format: no rate limit headers');
     }
 
     const remaining = parseInt(response.headers.get('ratelimit-remaining'));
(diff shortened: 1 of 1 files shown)


bodinsamuel (approved):

scripts/runner-update-env.js
```diff
@@ -73,7 +73,10 @@ async function fetchRunners() {
 
         services = services.concat(filteredServices);
 
-        await sleep(1000);
+        const remaining = parseInt(response.headers.get('ratelimit-remaining'));
+        const resetMs = (parseInt(response.headers.get('ratelimit-reset')) + 1) * 1000;
```
TBonnin, line 81: what if the headers are not present?

scripts/runner-update-env.js
```diff
@@ -73,7 +73,10 @@ async function fetchRunners() {
...
+
+        await sleep(Math.ceil((resetMs + 1) / remaining));
```
TBonnin: why this `+ 1`?

nalanj: Rate limit remaining time usually doesn't take the time it takes for the request to arrive on the other side into account, so it's usually a good idea to give it some head room. This will divide across their rate limit utilization so it should be something like an additional 10ms per request on average.

scripts/runner-update-env.js
nalanj: Handled in 97452c72

scripts/runner-update-env.js
nalanj: Oh, I see now. I had a redundant one in there. Removed in 1da6c41b

scripts/runner-update-env.js
```diff
@@ -73,7 +73,14 @@ async function fetchRunners() {
 
         services = services.concat(filteredServices);
 
-        await sleep(1000);
+        if (!response.headers.get('ratelimit-remaining') || !response.headers.get('ratelimit-reset')) {
+            throw new Error('Unexpected API format: no rate limit headers');
```
TBonnin, line 77: would it make sense to just have a warning and a high sleep delay if there is no headers. Since this is not something under our control we might still want the script to execute, even if slow

TBonnin (approved):

scripts/runner-update-env.js
nalanj: Maybe? I figure since it's a simple script we're running ourselves that if something changes with the API we'd dig into it and update the script anyway. I figure failing fast gives us the easiest way to emphasize something weird is happening.

nalanj: I had hit merge when I saw this, but if you want me to update to add the sleep on missing headers bit I'm more than happy to do so.

TBonnin: don't worry. not a big deal

PR #2989, opened 2024-11-14
feat(runner): Add monitorProviders support for runners

nalanj:
## Describe your changes

Adds new monitorProviders function to shared and utilizes it in runners to reload providers.yaml dynamically when a PROVIDERS_URL is set.

https://linear.app/nango/issue/NAN-2105/dynamically-reload-providersyaml


packages/shared/lib/services/providers.ts
```diff
@@ -23,6 +24,48 @@ export function getProvider(providerName: string): Provider | null {
...
+export async function monitorProviders(): Promise<void> {
+    const providersUrl = process.env['PROVIDERS_URL'];
+
+    // fall back to standard disk loading if no URL is provided
+    if (!providersUrl) {
+        return;
+    }
+
+    const maybeReloadInterval = process.env['PROVIDERS_RELOAD_INTERVAL'];
+    const reloadInterval = parseInt(maybeReloadInterval ?? '') || 60000;
```
TBonnin: if you use `parseEnvs(ENVS)` from the `utils` package you'll get all the this for free

packages/runner/lib/app.ts
```diff
@@ -2,6 +2,9 @@ import './tracer.js';
 import { server } from './server.js';
 import { stringifyError } from '@nangohq/utils';
 import { logger } from './utils.js';
+import { monitorProviders } from '@nangohq/shared';
+
+await monitorProviders();
```
TBonnin: let's cleanup the interval timer in the `close` function below

TBonnin (approved):

packages/shared/lib/services/providers.ts
nalanj: Fixed in e70ed3f7

packages/runner/lib/app.ts
nalanj: Fixed in e70ed3f7

PR #2990, opened 2024-11-14
feat: rename post-connection-scripts table to on-event-scripts 

TBonnin:
## Describe your changes

See Linear ticket for more context

Nango currently allows users to define a custom script that is
executed when a connection is created.
We want to extend this feature to also make it possible to execute
a custom script when a connection is deleted.
In order to do so we are going to implement a generic solution where a
custom script can be executed based on a Nango event like connection
creation or deletion
This first PR is not introducing any new behaviour but rename the
`post-connection-scripts table` to `on-event-scripts` and adding an event column
where the type of event associated with the script is being stored.
The PR is also renaming things related to the deployment, making sure
the old `postConnectionScriptsByProvider` param sent by old CLI is still
supported for backward compatibility

In next PRs I will:
- rename all things related to postConnnection in the orchestrator/scheduler
- add support for ON_CONNECTION_DELETED event
- refactor CLI to allow definition of both events/scripts

### DIFF
--- a/packages/shared/lib/services/sync/on-event-scripts.service.ts
+++ b/packages/shared/lib/services/sync/on-event-scripts.service.ts
@@ -1,26 +1,26 @@
 import db from '@nangohq/database';
 import remoteFileService from '../file/remote.service.js';
 import { env } from '@nangohq/utils';
-import type { PostConnectionScriptByProvider, PostConnectionScript, DBTeam, DBEnvironment } from '@nangohq/types';
+import type { OnEventScriptsByProvider, OnEventScript, DBTeam, DBEnvironment } from '@nangohq/types';
 import { increment } from './config/config.service.js';
 import configService from '../config.service.js';
 
-const TABLE = '_nango_post_connection_scripts';
+const TABLE = 'on_event_scripts';
 
-export const postConnectionScriptService = {
+export const onEventScriptService = {
     async update({
         environment,
         account,
-        postConnectionScriptsByProvider
+        onEventScriptsByProvider
     }: {
         environment: DBEnvironment;
         account: DBTeam;
-        postConnectionScriptsByProvider: PostConnectionScriptByProvider[];
+        onEventScriptsByProvider: OnEventScriptsByProvider[];
     }): Promise<void> {
         await db.knex.transaction(async (trx) => {
-            const postConnectionInserts: Omit<PostConnectionScript, 'id' | 'created_at' | 'updated_at'>[] = [];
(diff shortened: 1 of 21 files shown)


packages/shared/lib/services/sync/on-event-scripts.service.ts
```diff
@@ -77,10 +77,10 @@ export const postConnectionScriptService = {
...
                 }
             }
-            await trx.insert(postConnectionInserts).into(TABLE);
+            await trx.insert(onEventInserts).into(TABLE);
         });
     },
-    getByConfig: async (configId: number): Promise<PostConnectionScript[]> => {
-        return db.knex.from<PostConnectionScript>(TABLE).where({ config_id: configId, active: true });
+    getByConfig: async (configId: number): Promise<OnEventScript[]> => {
+        return db.knex.from<OnEventScript>(TABLE).where({ config_id: configId, active: true });
```
TBonnin, line 84: For now all rows are related to connection creation events. In following PR I will add a `eventType` parameter

bodinsamuel:
Didn't had time to review this morning, will look at that asap

packages/types/lib/deploy/api.ts
```diff
@@ -1,15 +1,15 @@
...
+import type { IncomingFlowConfig, OnEventScriptsByProvider } from './incomingFlow.js';
 import type { SyncDeploymentResult } from './index.js';
 
 export type PostDeployConfirmation = Endpoint<{
     Method: 'POST';
     Path: '/sync/deploy/confirmation';
     Body: {
         flowConfigs: IncomingFlowConfig[];
-        postConnectionScriptsByProvider: PostConnectionScriptByProvider[];
+        onEventScriptsByProvider: OnEventScriptsByProvider[];
```
bodinsamuel, line 12: nothing to change: What you did is smart, but a bit bothered that it's not representative of the truth

bodinsamuel (approved):
Looks good and work locally 🙏🏻

bodinsamuel:
Oops forgot one comment, it's missing the openapi schema change

packages/types/lib/deploy/api.ts
TBonnin: backward compatibility is indeed annoying ;)

TBonnin:
> Oops forgot one comment, it's missing the openapi schema change

I was not planning to change the openapi spec until I update the CLI to send the new param. What's described in the spec (ie: with `postConnectionScriptsByProvider` param) is still working.

PR #2991, opened 2024-11-14
feat(integration): support pulling workable subdomain through post-connection hook

nalanj:
## Describe your changes

Adds post connection script for workable to set its subdomain connection data dynamically.

Also fixes what I'm pretty sure was a bug where `baseUrlOverride` wasn't being passed through to the proxy from within post connection scripts.

https://linear.app/nango/issue/NAN-2170/post-connection-script-for-workable


packages/server/lib/hooks/connection/post-connection.ts
```diff
@@ -52,17 +52,23 @@ async function execute(createdConnection: RecentlyCreatedConnection, providerNam
 
                 return connection as Connection;
             },
-            proxy: async ({ method, endpoint, data, headers, params }: UserProvidedProxyConfiguration) => {
+            proxy: async ({ method, endpoint, data, headers, params, baseUrlOverride }: UserProvidedProxyConfiguration) => {
```
nalanj, line 55: While debugging using `baseUrlOverride` within the post-connection hook I traced it back to it not being passed through here.

A couple of the other system post-connection hooks also were using baseUrlOverride, so I'm not sure if they were just broken or if I'm really off base with what I was seeing.

packages/server/lib/hooks/connection/workable-post-connection.ts
```diff
@@ -0,0 +1,27 @@
...
+    const response = await nango.proxy({
+        baseUrlOverride: 'https://workable.com',
+        endpoint: '/spi/v3/accounts',
+        method: 'GET',
+        connectionId: connection.connection_id,
+        providerConfigKey: connection.provider_config_key
+    });
+
+    if (isAxiosError(response) || !response || !response.data) {
+        throw new Error('Failed to retrieve Workable subdomain');
```
nalanj, line 22: Throwing really only gives visibility in our logs, but it's something.

packages/server/lib/hooks/hooks.ts
```diff
@@ -76,15 +76,15 @@ export const connectionCreated = async (
...
-    if (options.initiateSync === true && !isHosted) {
-        await syncManager.createSyncForConnection(connection.id as number, logContextGetter, orchestrator);
-    }
-
     if (options.runPostConnectionScript === true) {
         await postConnection(createdConnectionPayload, provider, logContextGetter);
         await externalPostConnection(createdConnectionPayload, provider, logContextGetter);
     }
 
+    if (options.initiateSync === true && !isHosted) {
```
nalanj, line 84: The previous ordering would cause a sync to fire right before the post connection script, which meant that first execution would run without the data from the post connection script.

packages/shared/providers.yaml
```diff
@@ -6106,6 +6107,7 @@ workable:
             pattern: '^[a-z0-9_-]+$'
             example: domain
             suffix: .workable.com
+            optional: true
```
nalanj, line 6180: I went this route thinking it's better to let someone just hard code the subdomain if they know it, given some of the other complexity we saw with this change. Also the validation fails if you have use a non-specified `connection_config` variable in the `base_url` field.

packages/server/lib/hooks/connection/post-connection.ts
khaliqgant: Confirmed that the baseUrlOverride wasn't working as expected. Just checked the two using it. They were recently added (gusto and docusign) so don't think it has a real negative impact. Thanks for catching!

TBonnin (approved):
lgtm and thanks for the bugfix

khaliqgant (approved):

packages/server/lib/hooks/connection/workable-post-connection.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export default async function execute(nango: Nango) {
+    const connection = await nango.getConnection();
+
+    // rely on the customer provided subdomain if possible
+    if (connection.connection_config?.['subdomain']) {
+        return;
+    }
+
+    const response = await nango.proxy({
```
bodinsamuel: Can you type (even lightly)?  e.g: `nango.proxy<WorkableAccounts>(`

packages/shared/providers.yaml
bodinsamuel: For the validation, you can also improve it if it's in the way 👍🏻

bodinsamuel (approved):
🚚

packages/server/lib/hooks/connection/workable-post-connection.ts
nalanj: Fixed in 3d41db42

bodinsamuel:
Im going to go ahead and bypass protection since everybody is merging before you :D

PR #2992, opened 2024-11-15
fix(flows): fix CI, remove old parsing

bodinsamuel:
## Describe your changes

- **Fix CI for Flows**
New flows format from the other repo is breaking the CI test here (inconvenient but that's it)

- **Finally, drop `nango-config.service.js`**
I realized the CI was broken because the types were not synchronized in the tests, when I tried to fix the types it was still incompatible elsewhere so decided to go all in and finally delete the old parser. 

- **Use `nango-yaml `parser to parse `flows.yaml`**
It's good because there is only one way to parse yaml now BUT:
The types are now slightly wrong in the other way. I still need to move `StandardNangoConfig` to `@nangohq/types` and change the `model` shape to the new shape (the one from nango-yaml parser). 
It's not an issue because this data is only used by the frontend that expects both type of models.

- **Cache `flows.yaml` parsing**


packages/shared/lib/services/flow.service.ts
```diff
@@ -3,65 +3,121 @@ import path from 'path';
...
         try {
             const flowPath = path.join(dirname(import.meta.url), '../../flows.yaml');
-            const flows = yaml.load(fs.readFileSync(flowPath).toString()) as Config;
+            const flows = yaml.load(fs.readFileSync(flowPath).toString()) as FlowsYaml;
 
             if (flows === undefined || !('integrations' in flows) || Object.keys(flows.integrations).length <= 0) {
                 throw new Error('empty_flows');
             }
 
+            this.flowsRaw = flows;
```
TBonnin: why not assigning to this.flowsRaw above directly?

packages/shared/lib/services/flow.service.ts
```diff
@@ -3,65 +3,121 @@ import path from 'path';
...
+                continue;
+            }
 
-            const { success, response } = loadStandardConfig(nangoConfig);
+            const { models, ...rest } = flow;
+            const parser = new NangoYamlParserV2({
+                raw: { integrations: { [providerConfigKey]: rest }, models: models },
+                yaml: ''
+            });
+            parser.parse(); // we assume it's valid
```
TBonnin: why can we assume it is valid?

packages/shared/lib/services/flow.service.ts
```diff
@@ -3,65 +3,121 @@ import path from 'path';
...
+                providerConfigKey,
+                actions: [],
+                syncs: []
+            };
 
-            if (success && response) {
-                if (rawName) {
-                    const responseWithRaw = response.map((standardConfigItem) => {
-                        return { ...standardConfigItem, rawName };
+            for (const item of [...integration.actions, ...integration.syncs]) {
```
TBonnin, line 68: can integration be undefined if not found above?

TBonnin (approved):

packages/shared/lib/services/flow.service.ts
bodinsamuel: it's been parsed already by the integration template repo, the parsing is there just to get a standard format

packages/shared/lib/services/flow.service.ts
bodinsamuel: in theory, everything is possible, in practice no. This file has plenty of issues regarding error handling, we'll need to go over it one more time for sure

packages/shared/lib/services/flow.service.ts
bodinsamuel: I can indeed 👍🏻

PR #2457, opened 2024-07-03
feat: validate sync records

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1156/use-json-schema-to-validate-records-at-run-time
Fixes https://linear.app/nango/issue/NAN-1347/flag-in-cli-to-bypass-validation
Fixes https://linear.app/nango/issue/NAN-1348/improve-cli-display

- Validate records 
- Add flag to bypass validation in the CLI
- Better CLI output for validation

## Test

```sh
node packages/cli/dist/index.js dryrun __SYNC__ __CONNECTION__ --no-validation
```
Screenshot 2024-07-04 at 15 29 46

PR #2993, opened 2024-11-15
fix: post-connection-script

TBonnin:
## Describe your changes
post-connection-script have been failing to execute since the introduction of this change: https://github.com/NangoHQ/nango/pull/2457/files#diff-f52f0d61cd6802499f5a31aca89d3aec3df278a751065390c02b6bbdb24c0840R813-R81 (they don't have a syncId or a syncJobId)
I first thought about removing this check but after reflecting on it I came to the conclusion that post-connection-scripts (soon to become on-event-scripts) should not be `sync` that can save records and should be limited to external side effects like calling some APIs to do some setup/cleanup for instance.

This is a breaking change but I have checked in the db and there is only 2 active post-connection-scripts: one is my own test one and the 2nd one is a customer that is only calling `nango.proxy` and `nango.getMetadata` and anyway those scripts have been failing to execute for the past 4 months so I don't think it is being used.

### DIFF
--- a/packages/runner/lib/exec.ts
+++ b/packages/runner/lib/exec.ts
@@ -18,7 +18,16 @@ export async function exec(
     codeParams?: object,
     abortController: AbortController = new AbortController()
 ): Promise<RunnerOutput> {
-    const rawNango = nangoProps.scriptType === 'action' ? new NangoAction(nangoProps) : new NangoSync(nangoProps);
+    const rawNango = (() => {
+        switch (nangoProps.scriptType) {
+            case 'sync':
+            case 'webhook':
+                return new NangoSync(nangoProps);
+            case 'action':
+            case 'post-connection-script':
+                return new NangoAction(nangoProps);
+        }
+    })();
     const nango = process.env['NANGO_TELEMETRY_SDK'] ? instrumentSDK(rawNango) : rawNango;
     nango.abortSignal = abortController.signal;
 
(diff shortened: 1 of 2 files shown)


packages/runner/lib/exec.ts
```diff
@@ -18,7 +18,16 @@ export async function exec(
...
+    const rawNango = (() => {
+        switch (nangoProps.scriptType) {
+            case 'sync':
+            case 'webhook':
+                return new NangoSync(nangoProps);
+            case 'action':
+            case 'post-connection-script':
+                return new NangoAction(nangoProps);
+        }
+    })();
```
TBonnin, line 30: I don't think making `post-connection-script` a sync was an explicit decision but the result of the code having a default value. If the ternary condition had been written differently I am sure we could have ended up with a `NangoAction`
That's the reason I am here using a switch. If we ever add a new script type we would need to make a decision explicitly otherwise this code won't compile

bodinsamuel: tbf Action is also wrong since we expose a lot of stuff that could not work before post-connection is executed, but we are close to rewrite the sdk 👍🏻

bodinsamuel (approved):

packages/runner/lib/exec.ts
TBonnin: one option is less wrong that the other 😜

PR #2997, opened 2024-11-15
chore(orchestrator): rename PostConnectionTask to OnEventTask

TBonnin:
## Describe your changes

No behavior is changing as part of this PR. It is just renaming the PostConnectionTask into OnEventTask and tighten some types.
We could find an even more generic name because, from the point of view of the orchestrator, it is just a task with a script file attached to it but I couldn't find a better name. I am open to suggestion though. 

The concept of post connection still exists in the codebase. It is the provider-specific piece of logic that is defined in providers.yaml (ex: https://github.com/NangoHQ/nango/blob/master/packages/shared/providers.yaml#L1080) I don't like the fact it is called post_connection_script since it is not a script but some prebuilt logic that is executed server side. Anyway renaming it is out of the scope of this PR but at least now we don't have 2 different things with the same name anymore.

Next PR will refactor the CLI to enable the definition of the events and their associated scripts

https://linear.app/nango/issue/NAN-1962/connect-lifecycle-scripts#comment-7a350835


packages/server/lib/hooks/connection/post-connection.ts
```diff
@@ -123,7 +123,7 @@ async function execute(createdConnection: RecentlyCreatedConnection, providerNam
             }
         }
     } catch (err) {
-        await telemetry.log(LogTypes.POST_CONNECTION_SCRIPT_FAILURE, `Post connection manager failed, ${stringifyError(err)}`, LogActionEnum.AUTH, {
+        await telemetry.log(LogTypes.POST_CONNECTION_FAILURE, `Post connection manager failed, ${stringifyError(err)}`, LogActionEnum.AUTH, {
```
TBonnin, line 126: I sort of lied when I said I didn't rename post_connection things (the one that are provider specific). I at least renamed the telemetry constants

package-lock.json
nalanj, line 1: 🤔 Did you mean to delete package-lock.json here?

nalanj (commented):
Read through all of this and it looks good. I wasn't sure completely about how to test it, and wasn't sure about deleting the package-lock.json though.

package-lock.json
TBonnin: It is only deleting some transient dependencies that were likely not committed earlier. It is not related to this PR to tell the truth, I just squeezed it in here because of lazyness :)

bodinsamuel: This diff always happen after we publish the npm packages, it's going to disappear at some point when we remove shared from the cli

packages/logs/lib/models/helpers.ts
```diff
@@ -124,7 +124,7 @@ export const operationTypeToMessage: Record<ConcatOperationList, string> = {
     'action:run': 'Action execution',
     'admin:impersonation': 'Admin logged into another account',
     'auth:create_connection': 'Create connection',
-    'auth:post_connection': 'Post connection script execution',
+    'auth:post_connection': 'post connection execution',
```
bodinsamuel, line 127: not requiring changes: I just noticed both `post-connection` and `onConnectionCreated` are using `auth:post_connection` maybe it would be better to dissociate them, e.g: `internal_on_connection_created` and `external_on_connection_created` or clearly identify them if possible

Also one question, why not move both into the `on` folder?

bodinsamuel (approved):
Works well 💯

packages/logs/lib/models/helpers.ts
TBonnin: I am planning to change the log types once I introduced the new event. For now it is still post connection related so I think it is fine.
I also want to keep the internal post connection separate from the rest because they are technically very different. They are not schedule custom scripts executed by the runner but some provider specific piece of logic executed by server

PR #2999, opened 2024-11-18
fix(quickbooks): prompt user for a realmId connection config parameter

hassan254-prog:
## Describe your changes

- Prompt user for the `realmId` in quickbooks.


packages/shared/providers.yaml
```diff
@@ -4634,18 +4635,36 @@ quickbooks:
...
+            description: The realmId of your quickbooks account
+            pattern: '^\d{16}$'
+            example: '9341453474484455'
 
 quickbooks-sandbox:
     alias: quickbooks
     display_name: Quickbooks (sandbox)
     proxy:
+        connection_config:
+            realmId: ${connectionConfig.realmId}
```
bodinsamuel, line 4657: I think it's a follow-up question from the aws PR I didn't get, How are we supposed to use it if it's not in a URL or headers?

bodinsamuel (approved):

packages/shared/providers.yaml
hassan254-prog: For the AWS-IAM, since we didn't support the AWS V4 signature authentication mode, we needed a way to prompt the user for the region. This region is then used to generate the signature within the script. For QuickBooks, the `realmId` is required to build the proxy `base_url`. However, after discussing with Khaliq, we decided it was best not to alter the `base_url` as doing so might break compatibility for other users.

bodinsamuel: alright 👌🏻

PR #3004, opened 2024-11-18
fix(auth): incorrect validation

bodinsamuel:
## Describe your changes

- Fix auth validation
While refactoring auth validation I added `strict` which works weirdly when merging (it add stricts but only on the last object, removing all others objects before 😢 ). It was working with Nango Connect because it doesn't take any other params but when using regular headless it was only allowing `public_key`

- Added one simple integration test to avoid regression

Thanks @hassan254-prog for reporting

### DIFF
--- a/packages/types/lib/auth/http.api.ts
+++ b/packages/types/lib/auth/http.api.ts
@@ -24,7 +24,7 @@ export type PostPublicTbaAuthorization = Endpoint<{
     Params: {
         providerConfigKey: string;
     };
-    Path: '/auth/tba';
+    Path: '/auth/tba/:providerConfigKey';
     Error:
         | ApiError<'invalid_body'>
         | ApiError<'invalid_query_params'>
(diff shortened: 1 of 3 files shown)


khaliqgant (approved):

packages/types/lib/auth/http.api.ts
```diff
@@ -174,7 +174,7 @@ export type PostPublicSignatureAuthorization = Endpoint<{
     Params: {
         providerConfigKey: string;
     };
-    Path: '/auth/signature-based';
+    Path: '/auth/signature-based/:providerConfigKey';
```
TBonnin, line 177: can we have a linter or something to test the api definition against the routes?

TBonnin (approved):

nalanj (approved):

packages/types/lib/auth/http.api.ts
bodinsamuel: ideally every endpoints have at least one integration test that would take care of that (that's why I changed it now because suddenly it was screaming at me)

PR #3005, opened 2024-11-18
fix(webapp): Clear any cache from swr/infinite

nalanj:
## Describe your changes

We were seeing the connections tab still populated with stale data after logging out and logging back in. The culprit was that swr/infinite does not clear its state through the same mechanism as swr itself. This adds code to work around that, and a brute force page reload to clear all state in memory.

https://linear.app/nango/issue/NAN-1310/caching-is-too-aggressive-on-the-app


packages/webapp/src/utils/user.tsx
```diff
@@ -30,6 +29,14 @@ export function useSignout() {
...
-        nav('/signin', { replace: true });
+
+        // swr/infinite doesn't currently support clearing cache keys with the
+        // default mechanism. see https://github.com/vercel/swr/issues/2497
+        for (const key of cache.keys()) {
+            await mutate(key, undefined, { revalidate: false });
+        }
+
+        // force a full reload to ensure all state is cleared
+        window.location.href = '/signin';
```
nalanj, line 40: I don't love this, but it's kind of a sledgehammer solution for cache busting as a fallback, and I doubt we have so many manual logouts that it really matters much.

packages/webapp/src/utils/user.tsx
```diff
@@ -30,6 +29,14 @@ export function useSignout() {
         await logoutAPI(); // Destroy server session.
 
         await mutate(() => true, undefined, { revalidate: false }); // clean all cache
-        nav('/signin', { replace: true });
+
+        // swr/infinite doesn't currently support clearing cache keys with the
+        // default mechanism. see https://github.com/vercel/swr/issues/2497
+        for (const key of cache.keys()) {
```
nalanj, line 35: 🍖 Oddly, the default `mutate` recommended snippet for clearing the cache doesn't actually clear all of the cache. It specifically doesn't clear keys related to `swr/infinite`, so this fallback cleans that up.

nalanj:
I'm setting this ready for review since CI failures seem to be related to Docker Hub pull limits rather than anything in the PR.

packages/webapp/src/utils/user.tsx
bodinsamuel: I think it's fair tbh
if we do that we can get rid of the cache busting anyway

bodinsamuel (approved):

packages/webapp/src/utils/user.tsx
nalanj: I figured I would leave the cache busting in there in case we ever decided we wanted a more permanent storage for the cache, but yeah, it's effectively just a few lines of useless code now.

PR #3008, opened 2024-11-18
fix(runner): do not exit the runner process when a sync is cancelled

TBonnin:
## Describe your changes

I think that at some point we were forking/spawning a new subprocess for each script execution. When aborting we will then exit the process. However we don't do that anymore and scripts simply run in a node vm in the same runner process.
This commit modifies the sync sdk so an error is thrown instead of exiting. Jobs then discard this specific error.

https://linear.app/nango/issue/NAN-2195/fix-aborting-a-script-kills-the-runner-process

### DIFF
--- a/packages/jobs/lib/execution/operations/output.ts
+++ b/packages/jobs/lib/execution/operations/output.ts
@@ -2,7 +2,7 @@ import { orchestratorClient } from '../../clients.js';
 import type { JsonValue } from 'type-fest';
 import { logger } from '../../logger.js';
 import { NangoError } from '@nangohq/shared';
-import type { NangoProps } from '@nangohq/shared';
+import type { NangoProps, RunnerOutputError } from '@nangohq/shared';
 import { handleSyncError, handleSyncSuccess } from '../sync.js';
 import { handleActionError, handleActionSuccess } from '../action.js';
 import { handleWebhookError, handleWebhookSuccess } from '../webhook.js';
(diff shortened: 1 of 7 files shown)


packages/jobs/lib/execution/operations/output.ts
```diff
@@ -33,25 +33,25 @@ export async function handleSuccess({ taskId, nangoProps, output }: { taskId: st
...
-}: {
-    taskId: string;
-    nangoProps: NangoProps;
-    error: {
-        type: string;
-        payload: Record<string, unknown>;
-        status: number;
-    };
-}): Promise<void> {
+export async function handleError({ taskId, nangoProps, error }: { taskId: string; nangoProps: NangoProps; error: RunnerOutputError }): Promise<void> {
```
TBonnin, line 36: no real change here. Just making the type more explicit

packages/runner/lib/abort.ts
```diff
@@ -5,6 +5,7 @@ export const abort = (taskId: string): boolean => {
     const ac = abortControllers.get(taskId);
     if (ac) {
         ac.abort();
+        abortControllers.delete(taskId);
```
TBonnin, line 8: unrelated to this issue but I realized we were not very good citizen, not removing the abortController from the global map once it has been aborted

packages/jobs/lib/execution/operations/output.ts
```diff
@@ -33,25 +33,25 @@ export async function handleSuccess({ taskId, nangoProps, output }: { taskId: st
...
     const formattedError = toNangoError({
         err: error,
         defaultErrorType: `${nangoProps.scriptType}_script_failure`,
         scriptName: nangoProps.syncConfig.sync_name
     });
 
+    if (error.type === 'script_aborted') {
```
TBonnin, line 37: I am not a huge fan of NangoError.type being a `string` and not a union type. I tried to refactor it but it is basically opening pandora box so I gave up (for now) :(

nalanj (approved):

packages/jobs/lib/execution/operations/output.ts
```diff
@@ -33,25 +33,25 @@ export async function handleSuccess({ taskId, nangoProps, output }: { taskId: st
...
+        // do nothing, the script was aborted and its state already updated
+        logger.info(`Script was aborted. Ignoring output.`, {
+            taskId,
+            syncId: nangoProps.syncId,
+            environmentId: nangoProps.environmentId,
+            providerConfigKey: nangoProps.providerConfigKey,
+            connectionId: nangoProps.connectionId
+        });
+        return;
+    }
```
bodinsamuel, line 47: Could be placed before formattedError

bodinsamuel (approved):

PR #3009, opened 2024-11-18
fix(auth): remove more strict

bodinsamuel:
## Describe your changes

- Fix remove more strict in the auth controllers
Didn't had the credentials to test so I missed that


TBonnin (approved):

TBonnin (approved):
can we add some tests to make sure it is not re-added by mistake in a couple of months :)

bodinsamuel:
> can we add some tests to make sure it is not re-added by mistake in a couple of months :)

Yes I wish but all of them requires an external connection so we need to think about mocking third party

PR #3010, opened 2024-11-18
fix(webapp): Handle empty state for webhook settings coming over the wire

nalanj:
## Describe your changes

When webhook settings come back over the wire as null, which happens when they've never been configured, we were just skipping updating them, which caused stale data from an environment change in the UI. This fixes that!

https://linear.app/nango/issue/NAN-2104/inconsistent-environment-settings-values-when-switching-environments


packages/webapp/src/pages/Environment/Settings.tsx
```diff
@@ -115,7 +115,17 @@ export const EnvironmentSettings: React.FC = () => {
...
+            setWebhookCheckboxSettings({
+                alwaysSendWebhook: false,
+                sendAuthWebhook: false,
+                sendRefreshFailedWebhook: false,
+                sendSyncFailedWebhook: false
+            });
+            setWebhookUrl('');
+            setWebhookUrlSecondary('');
         }
+
```
TBonnin, line 128: would it make sense to have those settings always returned by the api?

nalanj: @TBonnin I'm pretty sure we could always return them. We would have to either:

1. pre-set an `ExternalWebhook` in the DB when we create the environment 
2. remove the `id` field from the `webhook_settings` part of the response so we can send an empty state without an id 

Technically we could make the id nullable but it seems like it'd be better just to not leak that id since we don't use it to update the webhook anyway.

TBonnin: I would be tempted to go for option 2 if we are actually not using this internal id anywhere. Approving but up to you to stick with this frontend change if you prefer this option

TBonnin (approved):

packages/webapp/src/pages/Environment/Settings.tsx
nalanj: @TBonnin I'll let this sit until tomorrow and we can see what @bodinsamuel says about it. I 100% agree with you about choosing option 2 though.

bodinsamuel: I would go for `1` personally, no changes are required and feels normal to me to pre-populate data

bodinsamuel (approved):

packages/webapp/src/pages/Environment/Settings.tsx
nalanj: I'm going to merge this as-is, and I logged a ticket to come back to for resolving this discussion, but I think this PR solves an existing problem the fastest way we can right now.

PR #3012, opened 2024-11-18
fix(server): Update query to paginate correctly against filter

nalanj:
## Describe your changes

Update the connections list query to filter errored connections appropriately. This was a little tricky! The key element though is the `FILTER` against `json_agg`. The query I'm targeting roughly looks like this:

```
SELECT row_to_json(_nango_connections.*) as connection,
  row_to_json(end_users.*) as end_user,
  _nango_configs.provider,
  COALESCE(
  	json_agg(
  		json_build_object('type', _nango_active_logs.type, 'log_id', _nango_active_logs.log_id)
  	) FILTER (WHERE _nango_active_logs.id IS NOT NULL)
  	, '[]'::json
  ) as active_logs,
  count(_nango_active_logs.id) as active_logs_count
FROM _nango_connections
	INNER JOIN _nango_configs ON "_nango_connections"."config_id" = "_nango_configs"."id"
	LEFT JOIN end_users ON end_users.id = _nango_connections.end_user_id
	LEFT JOIN _nango_active_logs ON _nango_connections.id = _nango_active_logs.connection_id
		AND _nango_active_logs.active = true
WHERE _nango_connections.environment_id = 2
	AND _nango_connections.deleted = false
GROUP BY _nango_connections.id, end_users.id, _nango_configs.provider
	HAVING count(_nango_active_logs.id) = 0
```

https://linear.app/nango/issue/NAN-2134/bug-with-connection-errors


TBonnin (commented):
I think we are due for adding test(s) for this function. The query was already quite involved and I am worried about breaking something here. I am happy to pair if you want to

packages/shared/lib/services/connection.service.integration.test.ts
```diff
@@ -66,4 +67,103 @@ describe('Connection service integration tests', () => {
             expect(updatedMetadata).toEqual({ ...initialMetadata, ...newMetadata });
         });
     });
+
+    describe('listConnections', () => {
```
bodinsamuel, line 74: Nice adding tests 💪🏻  
Could cover more ground by testing the API directly imo

packages/shared/lib/services/connection.service.integration.test.ts
```diff
@@ -66,4 +67,103 @@ describe('Connection service integration tests', () => {
...
+        const googleConfigKey = Math.random().toString(36).substring(7);
+        const notionConfigKey = Math.random().toString(36).substring(7);
```
bodinsamuel: I tried to make tests fully independent by seeding the env and account on each tests, it's debatable but avoid this stuff

packages/shared/lib/services/connection.service.ts
```diff
@@ -785,56 +785,57 @@ class ConnectionService {
...
+            .leftJoin(ACTIVE_LOG_TABLE, function () {
+                this.on(`${ACTIVE_LOG_TABLE}.connection_id`, '_nango_connections.id').andOn(`${ACTIVE_LOG_TABLE}.active`, db.knex.raw(true));
+            })
             .where({
                 '_nango_connections.environment_id': environmentId,
                 '_nango_connections.deleted': false
             })
-            .orderBy('_nango_connections.created_at', 'desc');
+            .orderBy('_nango_connections.created_at', 'desc')
+            .groupBy('_nango_connections.id', 'end_users.id', '_nango_configs.provider')
```
bodinsamuel, line 818: TIL you could groupBy `.id` with a `row_to_json`, that was the main reason for the subQuery, thanks 🙏🏻

bodinsamuel (approved):
Thanks

packages/shared/lib/services/connection.service.integration.test.ts
nalanj: Yeah, I was just following the pattern of tests in the file - it probably is worth a discussion of overall testing philosophy here since there are certainly tests in the repo that follow a lot of different patterns.

packages/shared/lib/services/connection.service.integration.test.ts
nalanj: Sure sure

nalanj: I actually think this is worth a discussion as well, because updating these tests to be isolated would be inconsistent with the rest of the file here and makes it more difficult to read. Feels like a discussion around testing philosophy is in order.

packages/shared/lib/services/connection.service.integration.test.ts
bodinsamuel: There is no philosophy tbh, those are just old tests. My opinion regarding tests: it's either very fine unit tests or e2e tests as much as we can. 
Because here you tests that it does what you want but that doesn't prevent the API from badly using the functions. Not pushing change or anything

packages/shared/lib/services/connection.service.integration.test.ts
bodinsamuel: same answer as above, those are old tests, them being not great is not a reason to not do better ☺️ 
tbf here using `const { env } = await seeders.seedAccountEnvAndUser();` instead of `env` would solve everything (I think), not a huge diff

packages/shared/lib/services/connection.service.integration.test.ts
nalanj: I guess I'm more trying to confirm that we'd rather this be an e2e test rather than a unit test (or integration since it hits the DB, I'm not sure how we're thinking about that name).

packages/shared/lib/services/connection.service.integration.test.ts
```diff
@@ -66,4 +67,103 @@ describe('Connection service integration tests', () => {
...
+        });
+
+        it('should filter connections by integration id', async () => {
+            const dbConnections = await connectionService.listConnections({
+                environmentId: env.id,
+                integrationIds: [googleConfigKey]
+            });
+
+            const connectionIds = dbConnections.map((c) => c.connection.connection_id);
+            expect(connectionIds).toEqual([googleOK.connection_id]);
```
TBonnin: can you avoid the mapping with something like 
```
expect(dbConnections).toEqual([
   expect.objectContaining({ 
      connection: { 
         connection_id: googleOK.connection_id 
      }
   })
]);
```

TBonnin (approved):
thank you very much for adding tests

packages/shared/lib/services/connection.service.integration.test.ts
nalanj: @TBonnin It has to be this because of how connection id is nested:

```
expect(dbConnections).toEqual([
  expect.objectContaining({ connection: expect.objectContaining({ connection_id: notion.connection_id }) }),
  expect.objectContaining({ connection: expect.objectContaining({ connection_id: google.connection_id }) })
]);
```

TBonnin: nvm then. this is ugly

bodinsamuel (approved):

PR #3013, opened 2024-11-18
fix(webapp): Only allow one error filter at a time

nalanj:
## Describe your changes

Right now it's kind of odd that you can select both error states but on the server it only applies the first one chosen. This makes it where the UI only allows one error state to be chosen, which is more consistent with the API.

https://linear.app/nango/issue/NAN-2197/connections-error-filter-shoudnt-be-a-multiselect

### DIFF
--- a/packages/webapp/src/pages/Connection/List.tsx
+++ b/packages/webapp/src/pages/Connection/List.tsx
@@ -120,14 +120,14 @@ export const ConnectionList: React.FC = () => {
     const [selectedIntegration, setSelectedIntegration] = useState<string[]>(defaultFilter);
     const [search, setSearch] = useState<string>('');
     const [debouncedSearch, setDebouncedSearch] = useState<string>('');
-    const [filterWithError, setFilterWithError] = useState<string[]>(defaultFilter);
+    const [filterWithError, setFilterWithError] = useState<string>('');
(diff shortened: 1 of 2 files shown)


nalanj:
Reverted a change from another branch I was working on that I missed here.

packages/webapp/src/pages/Connection/List.tsx
```diff
@@ -150,6 +150,11 @@ export const ConnectionList: React.FC = () => {
         setSelectedIntegration(values);
     };
 
+    const handleFilterErrorChange = (values: string[]) => {
+        const newItems = values.filter((f) => !filterWithError.includes(f));
+        setFilterWithError(newItems);
+    };
```
bodinsamuel, line 156: nitpick: I was lazy to do a proper filter but if you filter it doesn't make sense to store an array anymore.

bodinsamuel (approved):
Thanks 🙏🏻

packages/webapp/src/pages/Connection/List.tsx
nalanj: Fixed in e95d69d8

nalanj:
@bodinsamuel mind giving this another peek to see if that was what you meant?

bodinsamuel (approved):

PR #3016, opened 2024-11-19
fix(docs): [nan-2174] update images and docs link

khaliqgant:
## Describe your changes
Follow up to https://github.com/NangoHQ/nango/pull/3003
NAN-2174


bodinsamuel (approved):

nalanj (approved):
Noticed that the rounded corners in the screenshot are visible. Not sure if it's possible to toss a border radius on there to get it to clip a bit and make them disappear.

khaliqgant:
> Noticed that the rounded corners in the screenshot are visible. Not sure if it's possible to toss a border radius on there to get it to clip a bit and make them disappear.

Yeah, not ideal, but will leave as is for now.

PR #3017, opened 2024-11-19
feat(integrations): add support for grain

hassan254-prog:


packages/shared/providers.yaml
```diff
@@ -2839,6 +2839,41 @@ gorgias:
...
+    proxy:
+        base_url: https://api.grain.com
+        headers:
+            Authorization: Bearer ${apiKey}
+    docs: https://docs.nango.dev/integrations/all/grain
+    credentials:
+        apiKey:
+            type: string
+            title: API Key
+            description: The personal access token to your Grain account
```
TBonnin, line 2875: is the api key considered a secret?

TBonnin (approved):

packages/shared/providers.yaml
bodinsamuel: by default yes

bodinsamuel (approved):

PR #3018, opened 2024-11-19
fix(webapp): Fix filterWithError default value

nalanj:
## Describe your changes

filterWithError didn't like being set to a default of empty after I removed the array handling. This fixes that default state


packages/webapp/src/pages/Connection/List.tsx
```diff
@@ -120,7 +120,7 @@ export const ConnectionList: React.FC = () => {
     const [selectedIntegration, setSelectedIntegration] = useState<string[]>(defaultFilter);
     const [search, setSearch] = useState<string>('');
     const [debouncedSearch, setDebouncedSearch] = useState<string>('');
-    const [filterWithError, setFilterWithError] = useState<string>('');
+    const [filterWithError, setFilterWithError] = useState<string>('all');
```
bodinsamuel, line 123: Ah yes loose typing :D

bodinsamuel (approved):

packages/webapp/src/pages/Connection/List.tsx
nalanj: TypeScript makes it where you can never write any bugs. Right? RIGHT?

PR #3019, opened 2024-11-19
fix(cli): remove more from shared

bodinsamuel:
## Describe your changes

Contributes to https://linear.app/nango/issue/NAN-358/cli-upgrade-is-taking-too-long

- Remove NangoError
Replaced with CLIError when possible
- Remove ServiceResult
Copy pasted `result.ts` to emulate that behavior


packages/cli/lib/cli.ts
```diff
@@ -37,11 +37,11 @@ export function generate({ fullPath, debug = false }: { fullPath: string; debug?
     const postConnectionTemplateContents = fs.readFileSync(path.resolve(__dirname, './templates/post-connection.ejs'), 'utf8');
 
     const res = loadYamlAndGenerate({ fullPath, debug });
-    if (!res.success) {
+    if (!res) {
         return;
     }
 
-    const parsed = res.response!;
+    const parsed = res;
```
TBonnin: why not directly `const parsed = loadYamlAndGenerate()`? 😄

packages/cli/lib/cli.ts
```diff
@@ -223,15 +223,11 @@ NANGO_DEPLOY_AUTO_CONFIRM=false # Default value`
...
-        console.log(chalk.red(res.error?.message));
-        if (res.error?.payload) {
-            console.log(res.error.payload);
-        }
+    if (!res) {
         return;
     }
 
-    const parsed = res.response!;
+    const parsed = res;
```
TBonnin: same here

packages/cli/lib/index.ts
```diff
@@ -277,15 +277,14 @@ program
         const { autoConfirm, debug } = this.opts();
         const fullPath = process.cwd();
         await verificationService.necessaryFilesExist({ fullPath, autoConfirm, debug });
-        const { success, error, response } = parse(path.resolve(fullPath, NANGO_INTEGRATIONS_LOCATION));
-
-        if (!success || !response?.parsed) {
-            console.log(chalk.red(error?.message));
+        const resParsing = parse(path.resolve(fullPath, NANGO_INTEGRATIONS_LOCATION));
```
TBonnin: I would just call it `parsing`

packages/cli/lib/services/compile.service.ts
```diff
@@ -37,11 +37,11 @@ export async function compileAllFiles({
     }
 
     const res = loadYamlAndGenerate({ fullPath, debug });
-    if (!res.success) {
+    if (!res) {
         return false;
     }
 
-    const parsed = res.response!;
+    const parsed = res;
```
TBonnin: same here

packages/cli/lib/utils/result.ts
```diff
@@ -0,0 +1,63 @@
+/*
```
TBonnin, line 1: I am not a huge fan of duplicating code like this. If this is the price to remove `shared` dependency in CLI, it is probably worth it though

What are the other options:
1. make package `utils` a dependency of CLI. The risk is that it becomes another `shared situation` at some point
2. Package `Result` in its own package

any other?

TBonnin (approved):

packages/cli/lib/utils/result.ts
bodinsamuel: `Result` should really be https://github.com/supermacro/neverthrow
But I'm convinced we need a shared package between backend and cli yes

TBonnin: ouh! I like neverthrow. Looks like I didn't do my homework well enough when introducing `Result`.
The API is very similar to ours with a couple of welcome additions. Maybe a christmas break project to switch to it

PR #3021, opened 2024-11-19
feat: add support for on-events syntax in nango.yaml

TBonnin:
We want to be able to support custom scripts for more than just post-connection event. This commit is adding support for the following syntax in nango.yaml:
```
my-integration:
    on-events:
        post-connection-creation:
            - script1
            - script2
        pre-connection-deletion:
            - script1
    syncs:
        ...
```
Events currently supported are `post-connection-creation` and `pre-connection-deletion`
The change is backward compatible and still support the old `post-connection-scripts` even though I don't think it is currently being used.
I will add in next PR the logic to trigger the script when connection is deleted

### DIFF
--- a/packages/cli/lib/nango.yaml.schema.v2.json
+++ b/packages/cli/lib/nango.yaml.schema.v2.json
@@ -18,6 +18,29 @@
                             "_": "post-connection-scripts must be an array of strings."
                         }
                     },
+                    "on-events": {
+                        "type": "object",
+                        "properties": {
+                            "post-connection-creation": {
+                                "type": "array",
+                                "items": {
+                                    "type": "string"
+                                },
+                                "errorMessage": {
+                                    "_": "post-connection-creation must be an array of strings."
+                                }
+                            },
+                            "pre-connection-deletion": {
+                                "type": "array",
+                                "items": {
+                                    "type": "string"
+                                },
+                                "errorMessage": {
+                                    "_": "post-connection-creation must be an array of strings."
+                                }
+                            }
+                        }
+                    },
                     "syncs": {
                         "type": "object",
                         "patternProperties": {
--- a/packages/cli/lib/nango.yaml.schema.v2.json
+++ b/packages/cli/lib/nango.yaml.schema.v2.json
@@ -20,6 +20,7 @@
                     },
                     "on-events": {
                         "type": "object",
+                        "additionalProperties": false,
                         "properties": {
                             "post-connection-creation": {
                                 "type": "array",
--- a/packages/cli/lib/services/deploy.service.ts
+++ b/packages/cli/lib/services/deploy.service.ts
@@ -360,32 +360,34 @@ class DeployService {
         version?: string | undefined;
         optionalSyncName?: string | undefined;
         optionalActionName?: string | undefined;
-    }): { flowConfigs: IncomingFlowConfig[]; onEventScriptsByProvider: OnEventScriptsByProvider[]; jsonSchema: JSONSchema7 } | null {
+    }): { flowConfigs: IncomingFlowConfig[]; onEventScriptsByProvider: OnEventScriptsByProvider[] | undefined; jsonSchema: JSONSchema7 } | null {
         const postData: IncomingFlowConfig[] = [];
-        const onEventScriptsByProvider: OnEventScriptsByProvider[] = [];
+        const onEventScriptsByProvider: OnEventScriptsByProvider[] | undefined = optionalActionName || optionalSyncName ? undefined : []; // only load on-event scripts if we're not deploying a single sync or action
 
         for (const integration of parsed.integrations) {
             const { providerConfigKey, onEventScripts, postConnectionScripts } = integration;
 
-            const scripts: OnEventScriptsByProvider['scripts'] = [];
-            for (const event of Object.keys(onEventScripts) as OnEventType[]) {
-                for (const scriptName of onEventScripts[event]) {
-                    const files = loadScriptFiles({ scriptName: scriptName, providerConfigKey, fullPath, type: 'on-events' });
-                    if (files) {
-                        scripts.push({ name: scriptName, fileBody: files, event });
+            if (onEventScriptsByProvider) {
+                const scripts: OnEventScriptsByProvider['scripts'] = [];
+                for (const event of Object.keys(onEventScripts) as OnEventType[]) {
+                    for (const scriptName of onEventScripts[event]) {
+                        const files = loadScriptFiles({ scriptName: scriptName, providerConfigKey, fullPath, type: 'on-events' });
+                        if (files) {
+                            scripts.push({ name: scriptName, fileBody: files, event });
+                        }
                     }
                 }
-            }
 
-            // for backward compatibility we also load post-connection-creation scripts
-            for (const scriptName of postConnectionScripts || []) {
-                const files = loadScriptFiles({ scriptName: scriptName, providerConfigKey, fullPath, type: 'post-connection-scripts' });
-                if (files) {
-                    scripts.push({ name: scriptName, fileBody: files, event: 'post-connection-creation' });
+                // for backward compatibility we also load post-connection-creation scripts
+                for (const scriptName of postConnectionScripts || []) {
+                    const files = loadScriptFiles({ scriptName: scriptName, providerConfigKey, fullPath, type: 'post-connection-scripts' });
+                    if (files) {
+                        scripts.push({ name: scriptName, fileBody: files, event: 'post-connection-creation' });
+                    }
                 }
-            }
-            if (scripts.length > 0) {
-                onEventScriptsByProvider.push({ providerConfigKey, scripts });
+                if (scripts.length > 0) {
+                    onEventScriptsByProvider.push({ providerConfigKey, scripts });
+                }
             }
 
             if (!optionalActionName) {
--- a/packages/cli/lib/services/deploy.service.ts
+++ b/packages/cli/lib/services/deploy.service.ts
@@ -372,9 +372,11 @@ class DeployService {
                 for (const event of Object.keys(onEventScripts) as OnEventType[]) {
                     for (const scriptName of onEventScripts[event]) {
                         const files = loadScriptFiles({ scriptName: scriptName, providerConfigKey, fullPath, type: 'on-events' });
-                        if (files) {
-                            scripts.push({ name: scriptName, fileBody: files, event });
+                        if (!files) {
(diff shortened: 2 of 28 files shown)


packages/database/lib/migrations/2024111809211759_update_on_event_enum.cjs
```diff
@@ -0,0 +1,25 @@
+exports.config = { transaction: false };
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.schema.raw(`
+        ALTER TYPE script_trigger_event RENAME VALUE 'ON_CONNECTION_CREATED' TO 'POST_CONNECTION_CREATION';
```
TBonnin, line 8: renaming to make it even clearer when the script is triggering in case we one day need a `post` and `pre` creation or deletion.

packages/shared/lib/services/sync/on-event-scripts.service.ts
```diff
@@ -19,39 +28,35 @@ export const onEventScriptService = {
...
+            const previousScriptVersions = await trx
+                .from<OnEventScript>(TABLE)
+                .whereRaw(`config_id IN (SELECT id FROM _nango_configs WHERE environment_id = ?)`, [environment.id])
+                .where({
+                    active: true
+                })
+                .update({
+                    active: false
+                })
+                .returning('*');
```
TBonnin, line 43: here I am also fixing a bug: when a scripts was removed from the nango.yaml it was not deactivated since the execution wouldn't enter the loop below. Now we always deactivate all active scripts for a given environment before inserting the new ones

packages/shared/lib/services/sync/on-event-scripts.service.ts
```diff
@@ -19,39 +28,35 @@ export const onEventScriptService = {
...
-                    const { name, fileBody } = script;
-                    const config = await configService.getProviderConfig(providerConfigKey, environment.id);
-
-                    if (!config || !config.id) {
-                        continue;
-                    }
+                const config = await configService.getProviderConfig(providerConfigKey, environment.id);
+                if (!config || !config.id) {
+                    continue;
+                }
```
TBonnin, line 51: moving it out of the inner loop where it doesn't need to be

packages/cli/lib/nango.yaml.schema.v2.json
```diff
@@ -18,6 +18,29 @@
                             "_": "post-connection-scripts must be an array of strings."
                         }
                     },
+                    "on-events": {
+                        "type": "object",
+                        "properties": {
+                            "post-connection-creation": {
```
bodinsamuel, line 25: opinion: post/pre are harder to understand than before/after

packages/cli/lib/services/deploy.service.ts
```diff
@@ -364,18 +365,27 @@ class DeployService {
...
-            if (postConnectionScripts && postConnectionScripts.length > 0) {
-                const scripts: OnEventScriptsByProvider['scripts'] = [];
-                for (const postConnectionScript of postConnectionScripts) {
-                    const files = loadScriptFiles({ scriptName: postConnectionScript, providerConfigKey, fullPath, type: 'post-connection-scripts' });
-                    if (!files) {
-                        return null;
+            const { providerConfigKey, onEventScripts, postConnectionScripts } = integration;
+
+            const scripts: OnEventScriptsByProvider['scripts'] = [];
+            for (const event in onEventScripts || {}) {
```
bodinsamuel: `if (!onEventScripts) { continue; } ` seems a bit more clear to me

packages/cli/lib/services/deploy.service.ts
```diff
@@ -364,18 +365,27 @@ class DeployService {
...
+                const e = event as OnEventType;
```
bodinsamuel: Why it needs type correction? Also `e` could be better :D

packages/shared/lib/services/sync/on-event-scripts.service.ts
```diff
@@ -19,39 +28,35 @@ export const onEventScriptService = {
     }): Promise<void> {
         await db.knex.transaction(async (trx) => {
             const onEventInserts: Omit<OnEventScript, 'id' | 'created_at' | 'updated_at'>[] = [];
+
+            // Deactivate all previous scripts for the environment
+            // This is done to ensure that we don't have any orphaned scripts when they are removed from nango.yaml
+            const previousScriptVersions = await trx
+                .from<OnEventScript>(TABLE)
+                .whereRaw(`config_id IN (SELECT id FROM _nango_configs WHERE environment_id = ?)`, [environment.id])
```
bodinsamuel, line 36: didn't check but could there be an issue if we only deploy one provider/script?

packages/types/lib/nangoYaml/index.ts
```diff
@@ -36,7 +36,8 @@ export interface NangoYamlV2Integration {
     provider?: string;
     syncs?: Record<string, NangoYamlV2IntegrationSync>;
     actions?: Record<string, NangoYamlV2IntegrationAction>;
-    'post-connection-scripts'?: string[];
+    'on-events'?: Record<string, string[]>;
+    'post-connection-scripts'?: string[]; // DEPRECATED
```
bodinsamuel:
```suggestion
    /**
     * @deprecated
     */
    'post-connection-scripts'?: string[];
```

packages/types/lib/nangoYaml/index.ts
```diff
@@ -85,7 +88,8 @@ export interface NangoYamlParsedIntegration {
     providerConfigKey: string;
     syncs: ParsedNangoSync[];
     actions: ParsedNangoAction[];
-    postConnectionScripts: string[];
+    onEventScripts: Record<OnEventType, string[]>;
+    postConnectionScripts?: string[]; // DEPRECATED
```
bodinsamuel: same

bodinsamuel (commented):
Quick review, I haven't had time to test yet

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -109,7 +109,7 @@ export class DryRunService {
         }
 
         let providerConfigKey: string | undefined;
-        let isPostConnectionScript = false;
+        let isOnventScript = false;
```
nalanj:
```suggestion
        let isOnEventScript = false;
```

nalanj (commented):
Read through all of this and left one comment. I'm not even totally sure how to test this locally at this point, so I'll leave that for more experienced folks for now.

packages/cli/lib/services/deploy.service.ts
TBonnin: the `|| {}` was actually a leftover. `onEventScripts` is always defined now so no need for it

packages/cli/lib/services/deploy.service.ts
TBonnin: https://github.com/microsoft/TypeScript/issues/32811

packages/shared/lib/services/sync/on-event-scripts.service.ts
TBonnin: good point. I missed that

packages/cli/lib/nango.yaml.schema.v2.json
TBonnin: It is shorter and has the same meaning: https://www.oxfordlearnersdictionaries.com/definition/english/post_1

How strong do you feel about it @bodinsamuel @nalanj

nalanj: I don't care too much.

packages/cli/lib/services/deploy.service.ts
bodinsamuel: Ah I see 👍🏻 
Possible alternative without reassigning:
```ts
for (const event of Object.keys(onEventScripts) as OnEventType[]) {
```
But feel free to ignore

TBonnin: I will also rename

packages/cli/lib/nango.yaml.schema.v2.json
```diff
@@ -18,6 +18,29 @@
                             "_": "post-connection-scripts must be an array of strings."
                         }
                     },
+                    "on-events": {
+                        "type": "object",
```
bodinsamuel, line 22:
```suggestion
                        "type": "object",
                        "additionalProperties": false,
```
Had a typo and could not figure out why I didn't got any error

packages/cli/lib/nango.yaml.schema.v2.json
bodinsamuel: it's also much less common https://books.google.com/ngrams/graph?content=post%2Cafter&year_start=1800&year_end=2022&corpus=en&smoothing=3&case_insensitive=true
You decide 👍🏻

bodinsamuel (approved):
Lost all my comment :\
I was saying, when a script is missing it doesn't block the deploy which is weird but probably not related to your PR
Screenshot 2024-11-20 at 17 03 12

And compile doesn't even break ^^' 

Otherwise works well 💪🏻

bodinsamuel (approved):

TBonnin:
> when a script is missing it doesn't block the deploy which is weird but probably not related to your PR

Will look into that because I think it should fail the compilation

TBonnin:
> > when a script is missing it doesn't block the deploy which is weird but probably not related to your PR
> 
> Will look into that because I think it should fail the compilation

it was on me [59f41a4](https://github.com/NangoHQ/nango/pull/3021/commits/59f41a4546539a801f5ecc4e187abaf0edd178a2)

packages/cli/lib/nango.yaml.schema.v2.json
TBonnin: I am gonna stick with post/pre 😸

packages/shared/lib/services/sync/on-event-scripts.service.ts
TBonnin: as discussed in Slack. Yes it was a problem when deploying a single script. Was fixed in this commit https://github.com/NangoHQ/nango/pull/3021/commits/d3d0e345f7b5a74329cb05810db9ce7ebf583275 by not sending the on-events scripts when deploying a single script. There is currently no way to deploy a single `on-events` script but I have a ticket open for that

TBonnin:
Thank you for the thorough review @bodinsamuel I have fixed all issues/comments. let me know if I missed anything

bodinsamuel (approved):
🏎️

PR #1407, opened 2023-12-08
[NAN-20] implement hubspot webhooks

khaliqgant:
## Describe your changes
Adds webhook v1 implementation

Features
- Adds update connection metadata method so users don't have to overwrite metadata with any metadata change. Docsh have been updated
- Adds a hubspot contacts sync
- Abstracts any post connection actions to a hook `connectionCreatedHook`
- Adds a separate task queue for webhook handling which limits `maxActivitiesPerSecond` to 50

## Issue ticket number and link
NAN-20

## Checklist before requesting a review (skip if just adding/editing APIs & templates)
- [x] I added tests, otherwise the reason is: 
- [x] I added observability, otherwise the reason is:
- [ ] I added analytics, otherwise the reason is:

PR #3022, opened 2024-11-20
fix(logs): correctly logs request headers, redac headers and url, log verification

bodinsamuel:
## Describe your changes

While trying to debug verification, I realized we were not logging verification HTTP calls. While doing so found out we didn't log request headers correctly and we didn't redacted everything properly.

Fixes https://linear.app/nango/issue/NAN-2215/proxy-headers-are-not-logged-correctly
Fixes https://linear.app/nango/issue/NAN-2217/logs-connection-verification-http-calls-are-not-logged
Fixes https://linear.app/nango/issue/NAN-2216/logs-urls-are-not-redacted-properly

- **Correctly logs request headers**
We were using `config.headers` which is the config before being populated by specific provider headers (so missing a lot of headers in some cases). I tried to use axios directly but there is no built-in way to get request headers, after struggling I decided to pass `requestConfig` which contains all relevant headers (you can access undocumented raw axios request but it contains a lot of noise like datadog headers and stuff).

- **Correctly redact headers**
After I fixed the headers, I realized that we were not redacting them properly (dug my own graves here). 

- **Logs HTTP verification**
After I fixed everything, I could go back to logging HTTP verification properly (I had an issues with request headers hence the first fix).

- **Correctly redact url**
After fixing everything, I did the last mile and redacted the URL because it could sometimes leak credentials (e.g: `algolia`)

## Tests
Screenshot 2024-11-20 at 15 29 07


packages/server/lib/controllers/proxy.controller.ts
```diff
@@ -425,38 +440,39 @@ class ProxyController {
...
+    }: {
+        error: AxiosError;
+        config: ApplicationConstructedProxyConfiguration;
+        requestConfig: AxiosRequestConfig;
+        errorContent: string | Record<string, string>;
+        logCtx: LogContext;
+    }) {
+        const valuesToFilter = Object.values(config.connection.credentials);
+        const safeHeaders = redactHeaders({ headers: requestConfig.headers, valuesToFilter });
+        const redactedURL = redactURL({ url: requestConfig.url!, valuesToFilter });
```
bodinsamuel, line 467: While doing that I realized it's still suboptimal. A more bulletproof approach would be to set a redaction strategy in the LogContext that would redact the JSON.stringify string just before sending it to elasticsearch. This could lead to false positives but less sweat and mistakes

packages/logs/lib/models/helpers.ts
```diff
@@ -140,3 +141,15 @@ export const operationTypeToMessage: Record<ConcatOperationList, string> = {
     'webhook:incoming': 'Received a webhook',
     'webhook:forward': 'Forwarding Webhook'
 };
+
+/**
+ * Send buffered logs to elasticsearch
+ * Ultimately it would be better to have LogContextBuffer (or an option in LogContext)
+ */
+export async function flushLogsBuffer(logs: MessageRowInsert[], logCtx: LogContext) {
```
TBonnin, line 149: can we not have this function on `LogContext`, for instance `logCtx.flush(logs)` and also have a single call to ES?

packages/server/lib/controllers/proxy.controller.ts
TBonnin: that would be nicer indeed and would require to think about redaction when using `logCtx.http()`

packages/shared/lib/services/connection.service.ts
```diff
@@ -900,15 +902,13 @@ class ConnectionService {
...
-                  environment_id: number,
-                  connection_config: ConnectionConfig
-              ) => Promise<Result<boolean, NangoError>>)
+            | ((args: {
+                  config: Config;
+                  provider: Provider;
+                  credentials: ApiKeyCredentials | BasicApiCredentials | TbaCredentials | JwtCredentials | SignatureCredentials;
+                  connectionId: string;
+                  connectionConfig: ConnectionConfig;
+              }) => Promise<Result<{ logs: MessageRowInsert[] }, NangoError>>)
```
TBonnin, line 921: maybe `connectionTestHook` deserves its own named type. wdyt?

packages/shared/lib/services/proxy.service.ts
```diff
@@ -71,13 +72,6 @@ class ProxyService {
             return { success: false, error: new NangoError('missing_provider_config_key'), response: null, logs };
         }
 
-        logs.push({
-            type: 'log',
-            level: 'debug',
-            createdAt: new Date().toISOString(),
-            message: `Connection id: '${connectionId}' and provider config key: '${providerConfigKey}' parsed and received successfully`
-        });
-
```
TBonnin, line 80: almost 🪓 those ones last time. 😉

TBonnin (approved):

packages/shared/lib/services/connection.service.ts
```diff
@@ -58,6 +59,7 @@ import {
     extractValueByPath,
     stripCredential
 } from '../utils/utils.js';
+import { flushLogsBuffer } from '@nangohq/logs';
```
TBonnin: I don't think `logs` is a dependency of `shared` 😢 You will need to returns the logs so you can flush them outside of shared

packages/shared/lib/services/proxy.service.ts
bodinsamuel: We can agree they are useless :D

packages/shared/lib/services/connection.service.ts
bodinsamuel: aaah true, I was wondering why the CLI tests was failing, thanks

packages/shared/lib/services/connection.service.ts
bodinsamuel: I think the main issue is that hooks are in the server and this specific thing is in shared, wdyt about moving this specific code to server? 
Alternatively we can move all hooks to shared

bodinsamuel: edit: Tried, but it's a massive change to move that to `server`. Wondering what's best tbh, I feel like putting code in `server` is a foot gun, especially regarding credentials management; I think moving hooks to shared is the easiest but also more logical. 
Maybe there was a reason to have hooks on server @khaliqgant ?

packages/logs/lib/models/helpers.ts
bodinsamuel: yes, I think it needs more work as stated in the comment above the function. It's non trivial change so I'll do that later

packages/shared/lib/services/connection.service.ts
khaliqgant: Originally hooks [were in shared](https://github.com/NangoHQ/nango/pull/1407/files#diff-bea6e4872550e34e4234c63f78a893adabfceb768ca41ad82a520dc3edddd591) but @TBonnin [moved them to shared](https://github.com/NangoHQ/nango/pull/2188) actually.

bodinsamuel: arg good catch thanks 👍🏻 
I'll wait for thomas opinion then :D

packages/shared/lib/utils/http.ts
```diff
@@ -0,0 +1,40 @@
...
+            }
+        }
+    }
+
+    return safeHeaders;
+}
+
+export function redactURL({ url, valuesToFilter }: { url: string; valuesToFilter: string[] }) {
+    return valuesToFilter.reduce((curr, value) => {
+        return curr.replace(value, '_REDACTED_');
```
nalanj: Just curious, why `_REDACTED_` here but `REDACTED` above?

packages/shared/lib/services/proxy.service.ts
```diff
@@ -592,19 +546,30 @@ class ProxyService {
...
+        requestConfig,
+        config
+    }: {
+        error: unknown;
+        requestConfig: AxiosRequestConfig;
+        config: ApplicationConstructedProxyConfiguration;
+    }): Logs {
         const logs: MessageRowInsert[] = [];
 
+        const valuesToFilter = Object.values(config.connection.credentials);
```
nalanj, line 560: This isn't for this PR, just noting: Just because I was dealing with the Slack token stuff this is top of mind for me, but in that situation we have extra credential data that comes back, and we know customers are using that but it doesn't show up. May be worth us tweaking that in the future.

bodinsamuel: you mean we need to redact response too? true 😮‍💨

packages/shared/lib/utils/http.ts
bodinsamuel: ah no reason, I initially put underscore I'll change that

packages/shared/lib/services/proxy.service.ts
nalanj: You're right about response, but that actually wasn't what I was saying. I was more saying that right now we have data that comes over the wire that we store and that our customers use that should be redacted, but that we don't categorize as a credential so it wouldn't be able to be redacted.

bodinsamuel: which one are you thinking about?

packages/shared/lib/services/connection.service.ts
TBonnin: My original comment was just about the type so it is less verbose, not moving the hooks logic. 😄 

But if we want to talk about dependency management, the main challenge and question is: how to we want to inject dependencies:
1. hooks outside of shared and inject hook functions into shared that encapsulate their own dependencies (shared doesn't have to care)
2. hooks inside share and inject all the dependencies the hooks needs (shared needs to know about the deps)
3. ~~All the dependencies are direct dependencies of shared so no injection but this is exactly what we have spent a lot of time/energy trying to untangled.~~

i though the first one was cleaner. For instance we don't follow this pattern [here](https://github.com/NangoHQ/nango/blob/72ca1714f9e4e943955550f0d59db4d175592797/packages/shared/lib/services/connection.service.ts#L879-L881). As a result we have to pass the `orchestrator` and `logContextGetter` just because it is required by the side effects executed after a connection is deleted. Instead we could pass a `postConnectionDeletion` hook that is opaque for shared who doesn't need to know anymore about the orchestrator for example

bodinsamuel: Yeah okay I see, I talked about moving code because dissociating the types and the code is always painful. 

---

> i though the first one was cleaner. For instance we don't follow this pattern [here](https://github.com/NangoHQ/nango/blob/72ca1714f9e4e943955550f0d59db4d175592797/packages/shared/lib/services/connection.service.ts#L879-L881). As a result we have to pass the orchestrator and logContextGetter just because it is required by the side effects executed after a connection is deleted.

We mostly did that because of the CLI though not because we really wanted to, at least I wouldn't have bothered with logs.

Opinion: We moved stuff outside shared for multiple reasons, mostly CLI and code cleaning. Once we have cut CLI from shared, there will be fewer reasons to move things outside `shared` things like this code: not big enough to be a package but still shared in various parts of the code. 
We can still do dependency injection even if the code lives inside shared (handy for testing). 
I'll be in favor of actually moving this back since it's reused. Then the way we used is up to the person.

I'll go ahead and merge then

TBonnin: > We mostly did that because of the CLI though not because we really wanted to, at least I wouldn't have bothered with logs.

True and for logs it would make kind of sense. Not sure about the orchestrator though. Anyway it is tangential to this PR so go ahead and merge

PR #3025, opened 2024-11-20
feat(server): Add sync errors to connection count

nalanj:
## Describe your changes

Adds sync errors to response from connection count so we can show it in the UI


TBonnin (approved):

packages/shared/lib/services/connection.service.integration.test.ts
```diff
@@ -209,4 +210,47 @@ describe('Connection service integration tests', () => {
...
+                type: 'sync',
+                action: 'sync_test',
+                connection_id: notionSyncError.id!,
+                log_id: Math.random().toString(36).substring(7),
+                active: true,
+                sync_id: sync.sync.id
+            });
+
+            const countResult = await connectionService.count({ environmentId: env.id });
+            expect(countResult.isOk()).toBeTruthy();
```
TBonnin: you don't really need this one because if it is an error the `unwrap` below will throw

nalanj: Fixed in 49fefb91

PR #3026, opened 2024-11-20
feat(webapp): surface all connection error counts

nalanj:
Show total error connection counts including sync errors and then show new 

Working from this design:

https://www.figma.com/design/X9f9i3bKiQ3gaRhoNTUbXp/Error-Reporting?node-id=4248-1376&t=XPt5fHDj37pclTO8-4

image

Bastien and I discussed removing the red bar because it felt a little too heavy handed. We also discussed having a 3rd state, the x icon, instead of two icons, when a connection has both types of error happening.  And the copy icon shown in the drawing here got removed in favor of just using the connection id copy icon that's already there on hover.

Some screenshots!

image

Hover state for the error counts, displaying error breakdown

image

Tooltip showing a connection that has both sync and authentication errors

image

Tooltip showing a connection that has only sync errors

image

Tooltip showing a connection that has only auth errors

## How I tested it

- Get connections into a bad state
  - To get a Slack connection into a bad auth state I updated the integration config to make the client id incorrect, and then hit refresh on the token
  - To get an Unauthenticated connection into a bad sync state I just used a script that threw
- Once you have that connection in a bad state you have a count you can play with for some of the display logic. I ended up editing the `errorCounts` variable on line 53 to test various states a little more directly without having to break connections manually


packages/webapp/src/components/ErrorCircle.tsx
```diff
@@ -0,0 +1,15 @@
+import { XMarkIcon, ArrowPathIcon, LockClosedIcon } from '@heroicons/react/24/outline';
```
bodinsamuel: Nb: we should only use tabler icons, maybe the figma was old

bodinsamuel (approved):

packages/webapp/src/components/ErrorCircle.tsx
nalanj: @bodinsamuel Fixed in 946d9e36. I went ahead and changed the x icon to use ! instead, because that felt less like some sort of close button to me.

PR #3027, opened 2024-11-20
feat: trigger pre connection deletion script on connection deletion

TBonnin:
## Describe your changes

Trigger pre-deletion-script when connection is deleted


packages/shared/lib/services/connection.service.ts
```diff
@@ -865,6 +869,8 @@ class ConnectionService {
             })
             .update({ deleted: true, credentials: {}, credentials_iv: null, credentials_tag: null, deleted_at: new Date() });
 
+        // TODO: move the following side effects to a post deletion hook
+        // so we can remove the orchestrator and logContextGetter dependencies
```
TBonnin, line 885: would be nice I think to pass those side effects as a `postDeletion` function. It would simplify the dependencies. 
It is out of scope for the PR though

packages/types/lib/logs/messages.ts
```diff
@@ -43,9 +43,10 @@ export interface OperationAction {
     type: 'action';
     action: 'run';
 }
+// TODO: rename to OperationConnection
```
TBonnin, line 46: `OperationConnection` and `type: connection` will  be better names but it would be a breaking change for the log search.

bodinsamuel:
You haven't merged, a bit harder to review

TBonnin:
> You haven't merged, a bit harder to review

Sorry there were some conflicts and github couldn't automerge. It should be fine now @bodinsamuel

nalanj:
Tossing here since I'm not quite to really reviewing yet. When I ran `nango generate` in the directory to get the event script I accidentally called the script: `unauthenticated/pre-connection-deletion` because I was just troubleshooting, and it really didn't like that.

```
Error: ENOENT: no such file or directory, open '/Users/nalanj/Source/NangoHQ/nango-integrations/unauthenticated/on-events/unauthenticated/pre-connection-deletion.ts'
```

May be worth us blocking slashes in those?

nalanj:
In testing this, I'm also seeing that it would be nice if we could somehow surface these scripts existing in the UI so it's easy to know that everything is working as expected.

nalanj:
Also verified this ran! The log entry looks like this:

image

Should that have the name of the on-event script there?

nalanj (approved):
Ran this locally and it worked. Left a couple of small comments above but none of them are deal-breakers to me! :ship:

bodinsamuel (approved):
Works well 👍🏻 
Nitpick, the `onEvent` prop does not reflect the event name but the script name which is confusing (I named mine `pre.ts` ) 
Screenshot 2024-11-21 at 17 00 59

TBonnin:
> Tossing here since I'm not quite to really reviewing yet. When I ran `nango generate` in the directory to get the event script I accidentally called the script: `unauthenticated/pre-connection-deletion` because I was just troubleshooting, and it really didn't like that.
> 
> ```
> Error: ENOENT: no such file or directory, open '/Users/nalanj/Source/NangoHQ/nango-integrations/unauthenticated/on-events/unauthenticated/pre-connection-deletion.ts'
> ```
> 
> May be worth us blocking slashes in those?

good catch. fixed in https://github.com/NangoHQ/nango/pull/3027/commits/e16fb503d388963bbf0416b0a529a44212cd9265

TBonnin:
> Also verified this ran! The log entry looks like this:
> image
> 
> Should that have the name of the on-event script there?

maybe. I have followed the same pattern used in post-connection-scripts where if multiple scripts they are run as part of the same log operation. In this case which scripts should be displayed?
I have a follow-up ticket to first decide if we want to split multiple scripts as multiple logs operations and in this case we could show the script name. Wanted to discuss the UX implications with @bastienbeurier @bodinsamuel

TBonnin:
> Works well 👍🏻 Nitpick, the `onEvent` prop does not reflect the event name but the script name which is confusing (I named mine `pre.ts` ) Screenshot 2024-11-21 at 17 00 59

good catch. renamed the field to `onEventScript` in https://github.com/NangoHQ/nango/pull/3027/commits/566b883c42a8e32a7d2e4325d21f1f6ff7ab24a6

TBonnin:
> In testing this, I'm also seeing that it would be nice if we could somehow surface these scripts existing in the UI so it's easy to know that everything is working as expected.

I know. I will do that in a following PR because it requires a bit of refactoring

PR #3031, opened 2024-11-21
fix(connectionConfig): ensure user's connectionConfig takes precedence

hassan254-prog:
## Describe your changes
- With the `optional` flag for fields in `connect-ui`, the `connectionConfig` can now have an empty string if the user doesn't fill this as it wasn't the case with the headless connection where the `connectionConfig` was mandatory. The PR sorts to fix situations where the user provides the `connnectionConfig` and present in token or callback metadata, ensuring that user's `connectionConfig` takes precedence, if this is an empty string, we will use the metadata from the callback or token, and if not present in either and user hasn't filled this we will have it as an empty string.


packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -1006,7 +1006,20 @@ class OAuthController {
                 return publisher.notifyErr(res, channel, providerConfigKey, connectionId, WSErrBuilder.UnknownError());
             }
 
-            let connectionConfig = { ...session.connectionConfig, ...tokenMetadata, ...callbackMetadata };
+            let connectionConfig = {
```
hassan254-prog, line 1009: This will ensure that users `connecionConfig` takes precedence, since in connect-ui, we have the `optional` field, this makes the users `connectionConfig` an empty string when applied, if the empty string is present in the metadata, we override it, else we use user's `connectionConfg`

khaliqgant (approved):

bodinsamuel (approved):
👌🏻 
Can you make sure to add context to your PR so we understand the initial bug and the fix without digging too much ☺️

hassan254-prog:
> 👌🏻 Can you make sure to add context to your PR so we understand the initial bug and the fix without digging too much ☺️

Updated the PR description 👍 .

PR #3032, opened 2024-11-21
fix(connect-ui): enforce session token integrations

bodinsamuel:
## Describe your changes

Fixes https://linear.app/nango/issue/NAN-1945/enforce-allowed-integrations-in-each-auth-endpoint

- Enforce session token's allowed integrations


nalanj (commented):
Would it make sense to factor this shared code out so future tweaks only change one file?

bodinsamuel:
> Would it make sense to factor this shared code out so future tweaks only change one file?

Yes and no, I am all for not repeating myself but sometimes it's handy. 
In that case factorizing means a lot of `ifs` and possible branching depending on the auth strategy. There are only a few lines that are strictly similar between all of them (e.g: this change, getting the provider/integrations). Basically, we could wrap the bootstrap of each auth and maybe the error handling.

TBonnin:
> > Would it make sense to factor this shared code out so future tweaks only change one file?
> 
> Yes and no, I am all for not repeating myself but sometimes it's handy. In that case factorizing means a lot of `ifs` and possible branching depending on the auth strategy. There are only a few lines that are strictly similar between all of them (e.g: this change, getting the provider/integrations). Basically, we could wrap the bootstrap of each auth and maybe the error handling.

TBonnin:
sorry I fat fingered the close button. Meant to comment

TBonnin:
This change looks very much like it belongs in a middleware. We would have to move the `logCtx` in `RequestLocals` but I am not sure it is possible.

TBonnin (commented):
can we make it a utility function if the middleware option is not feasible:
```
const checkIntegrationAllowed = async (res, logCtx, config) => {
...
}
```
and then
```
if (!await checkIntegrationAllowed(logCtx, res, config)) {
    return;
  }
```

bodinsamuel:
>This change looks very much like it belongs in a middleware. We would have to move the logCtx in RequestLocals but I am not sure it is possible.

This might, but requires a massive change to all auth controllers creating the logCtx beforehand and this makes the logic hard to follow (less self-contained) imo. 
I'll create a helper since both of you advocate for it

bodinsamuel:
Cleaned with an helper and added some integrations tests 😇

packages/server/lib/utils/tests.ts
```diff
@@ -78,7 +78,9 @@ export function apiFetch(baseUrl: string) {
 /**
  * Assert API response is an error
  */
-export function isError(json: any): asserts json is ApiError<any> {
+export function isError<TType extends Record<string, any>>(
+    json: TType extends { json: any } ? never : TType
+): asserts json is Extract<TType extends { json: any } ? never : TType, { error: any }> {
```
TBonnin, line 83: is it possible to define the types in order to get rid of the duplication and make the signature a bit more readable: 
```
type T = Record<string, any>
type TT = T extends { json: any } ? never : T
...
```

packages/server/lib/utils/auth.ts
```diff
@@ -0,0 +1,29 @@
+import type { LogContext } from '@nangohq/logs';
+import type { ApiError, IntegrationConfig } from '@nangohq/types';
+
+import type { Response } from 'express';
+import type { RequestLocals } from './express';
+
+export async function checkIfIntegrationIsAllowed({
```
TBonnin: I know I have proposed this name (sorry) but `isIntegrationAllowed` would be even shorter and better

TBonnin (approved):

packages/server/lib/utils/tests.ts
bodinsamuel: you can't define this types outside the function unless you make them generic, at least I didn't find a proper solution to make that better.

PR #3033, opened 2024-11-21
chore: update pull_request_template

TBonnin:
My take: having a checklist of things that rarely apply is pretty useless. I (and I am sure not the only one) just always ignore the observability and analytics points. I propose this change to focus on the things that matter: description of the changes, testing and ticket. All of them almost always applies.

Also changing to comments so the template instructions don't show once the PR is submitted

- [ ] I added tests, otherwise the reason is: not applicable 🤔 
- [ ] I added observability, otherwise the reason is: not applicable 🙄 
- [ ] I added analytics, otherwise the reason is: not applicable 😝

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,10 +1,6 @@
-<!-- Describe your changes --> 
+<!-- Describe the problem and your solution --> 
 
 <!-- Issue ticket number and link (if applicable) -->
 
-<!-- Testing -->
-- [ ] I have tested those changes 
-- [ ] I have added tests (skip if just adding/editing providers)
-
-<!-- Testing instructions (if applicable) -->
+<!-- Testing instructions (skip if just adding/editing providers) -->
 
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,8 +1,10 @@
-## Describe your changes
+<!-- Describe your changes --> 
 
-## Issue ticket number and link
+<!-- Issue ticket number and link (if applicable) -->
 
-## Checklist before requesting a review (skip if just adding/editing APIs & templates)
-- [ ] I added tests, otherwise the reason is: 
-- [ ] I added observability, otherwise the reason is:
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -1,8 +1,10 @@
...
+<!-- Describe your changes --> 
 
-## Issue ticket number and link
+<!-- Issue ticket number and link (if applicable) -->
+
+<!-- Testing -->
+- [ ] I have tested those changes 
+- [ ] I have added tests (skip if just adding/editing providers)
+
+<!-- Testing instructions (if applicable) -->
```
TBonnin: arguably we could get rid of this one but @nalanj raised a good point where we often make the assumption that the reviewers have all the context to tests the PR.

.github/pull_request_template.md
```diff
@@ -1,8 +1,10 @@
...
 
-## Checklist before requesting a review (skip if just adding/editing APIs & templates)
-- [ ] I added tests, otherwise the reason is: 
-- [ ] I added observability, otherwise the reason is:
-- [ ] I added analytics, otherwise the reason is:
```
bodinsamuel, line 8: lgtm, usually I go simpler. Checkboxes are annoying and honestly, after a while, you stop caring.

```md
# 🤓 Context

# 🧪 How to tests?
```

bodinsamuel (approved):

nalanj (approved):

hassan254-prog (approved):
Yes, I agree with Samuel, after a while you stop caring for the checkboxes.

TBonnin:
nobody likes checkboxes apparently so let's get rid of them: https://github.com/NangoHQ/nango/pull/3033/commits/9a620436dceab04bb5c1e0bd81f38ca5591534e4

.github/pull_request_template.md
TBonnin: I would go for that if it was a private repo but since it is public I would be slightly more verbose

PR #3036, opened 2024-11-21
fix(jobs): execute close only once

TBonnin:
Notice while investigating today's mini incident that the `close` function could be executed multiple times if several unhandledRejections which create some sort of infinite loop

It is not fixing the root cause of the incident, still investigating


packages/utils/lib/once.ts
```diff
@@ -1,11 +1,13 @@
...
     let called = false;
+    let result: R;
 
     return function (...args: T) {
         if (!called) {
             called = true;
-            fn(...args);
+            result = fn(...args);
         }
+        return result;
```
TBonnin, line 11: `once` now handles any sort of return type for `fn` and memoize the result

nalanj (approved):

bodinsamuel (approved):

PR #3037, opened 2024-11-21
feat(server): Add missing field hints for integrations to api

nalanj:
## Describe your changes

Adds a new function to `configService` called `validateProviderConfig` that returns a set of missing fields for the given auth type and provider config. Also updates the api to return those missing fields in our payload for `/api/v1/integrations/:providerId` so that we can display a warning about unfilled fields.

This PR only updates the API, it doesn't handle the fields in the UI yet.

At present it's not handling the CUSTOM auth method because there's some code weirdness I'm digging into there and that auth method is only used for one provider, so it's not too big a deal to keep moving forward and backfill it later.

## How I Tested This

I wrote tests for the validation bits, but I also did a quick check in inspector to ensure it's sending the new data over the API correctly.

- I went to the integration page for a specific integration.
- Opened inspector.
- Found the integration's request.
- Visually confirmed that it's being returned over the API.

### DIFF
--- a/packages/shared/lib/services/config.service.ts
+++ b/packages/shared/lib/services/config.service.ts
@@ -8,6 +8,7 @@ import syncManager from './sync/manager.service.js';
 import { deleteSyncFilesForConfig, deleteByConfigId as deleteSyncConfigByConfigId } from '../services/sync/config/config.service.js';
 import environmentService from '../services/environment.service.js';
 import type { Orchestrator } from '../clients/orchestrator.js';
+import type { AuthModeType } from '@nangohq/types';
 
 class ConfigService {
     async getById(id: number): Promise<ProviderConfig | null> {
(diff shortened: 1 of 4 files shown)


packages/shared/lib/services/config.service.ts
```diff
@@ -74,6 +75,51 @@ class ConfigService {
...
+            case 'APP_STORE':
+            case 'TABLEAU':
+            case 'NONE':
+            case 'OAUTH2_CC':
+                break;
+        }
+
+        return missingFields;
+    }
+
```
TBonnin, line 122: what about a more functional approach that is more extensible. @bodinsamuel is not gonna like it 😆 
```typescript
    interface ValidationRule {
        field: keyof ProviderConfig;
        modes: AuthModeType[];
        validate: (value?: ProviderConfig[keyof ProviderConfig]) => boolean;
    }

    isEmpty = (value?: ProviderConfig[keyof ProviderConfig]) => { ... }

    VALIDATION_RULES: ValidationRule[] = [
        {
            field: 'oauth_client_id',
            modes: ['OAUTH1', 'OAUTH2', 'TBA', 'APP'],
            validate: isEmpty
        },
        {
            field: 'oauth_client_secret',
            modes: ['OAUTH1', 'OAUTH2', 'TBA', 'APP'],
            validate: isEmpty
        },
        {
            field: 'app_link',
            modes: ['APP'],
            validate: isEmpty
        }
    ];

    validateProviderConfig(authMode: AuthModeType, providerConfig: ProviderConfig): string[] {
        return this.VALIDATION_RULES
            .filter((rule) => rule.modes.includes(authMode))
            .filter((rule) => !rule.validate(providerConfig[rule.field]))
            .map((field) => field.toString());
    }

```

TBonnin (approved):

nalanj:
Moved this back to draft because I'm going to have to store these missing fields to handle the list case.

nalanj:
Closing this in favor of a different direction. Sorry for the fake out there, @TBonnin

PR #3039, opened 2024-11-22
fix(connect): improve greenhouse support

bodinsamuel:
# Changes

Fixes https://linear.app/nango/issue/NAN-2234/connect-fix-greenhouse-api-key

- Fix greenhouse using api key as basic auth 
- Add verification
- Allow default empty value in the UI

### DIFF
--- a/packages/connect-ui/src/lib/utils.ts
+++ b/packages/connect-ui/src/lib/utils.ts
@@ -22,7 +22,12 @@ export function jsonSchemaToZod(schema: SimplifiedJSONSchema): ZodTypeAny {
     } else if (schema.format === 'email') {
         fieldString = fieldString.email();
     } else {
-        fieldString = fieldString.min(1, 'This field is required');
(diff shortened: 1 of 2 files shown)


packages/connect-ui/src/lib/utils.ts
```diff
@@ -22,7 +22,12 @@ export function jsonSchemaToZod(schema: SimplifiedJSONSchema): ZodTypeAny {
     } else if (schema.format === 'email') {
         fieldString = fieldString.email();
     } else {
-        fieldString = fieldString.min(1, 'This field is required');
+        // Some providers takes empty default_value
+        // so we need to differentiate empty and not defined
+        // note that "optional" could work but does not exactly mean the same
```
TBonnin, line 27: I want an `Option/Maybe` type. Please typescript, please! Maybe we do our own. It is probably 20 lines of code

TBonnin (approved):

PR #3041, opened 2024-11-22
fix(ui): improve endpoints ordering

bodinsamuel:
# Changes

- Improve endpoints ordering
We were only pushing items up if the comparison was directly different, so when we compared POST to DELETE, POST was not pushed up.

### DIFF
--- a/packages/webapp/src/pages/Integrations/providerConfigKey/Endpoints/Show.tsx
+++ b/packages/webapp/src/pages/Integrations/providerConfigKey/Endpoints/Show.tsx
@@ -9,6 +9,8 @@ import { EndpointsList } from './components/List';
 import { EndpointOne } from './components/One';
 import PageNotFound from '../../../PageNotFound';
 
+const METHOD_PRIORITY = { GET: 1, POST: 2, PUT: 3, PATCH: 4, DELETE: 5 };
+
 export const EndpointsShow: React.FC<{ integration: GetIntegration['Success']['data'] }> = ({ integration }) => {
     const env = useStore((state) => state.env);
     const { providerConfigKey } = useParams();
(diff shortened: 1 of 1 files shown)


packages/webapp/src/pages/Integrations/providerConfigKey/Endpoints/Show.tsx
```diff
@@ -62,8 +62,12 @@ export const EndpointsShow: React.FC<{ integration: GetIntegration['Success']['d
...
                         // Sort by method
                         if (a.endpoint.method === 'GET' && b.endpoint.method !== 'GET') return -1;
-                        else if (a.endpoint.method === 'POST' && b.endpoint.method === 'PUT') return -1;
-                        else if (a.endpoint.method === 'PUT' && b.endpoint.method === 'PATCH') return -1;
+                        else if (
+                            a.endpoint.method === 'POST' &&
+                            (b.endpoint.method === 'PUT' || b.endpoint.method === 'PATCH' || b.endpoint.method === 'DELETE')
+                        )
+                            return -1;
+                        else if (a.endpoint.method === 'PUT' && (b.endpoint.method === 'PATCH' || b.endpoint.method === 'DELETE')) return -1;
```
TBonnin: can we do?
```ts
const METHOD_PRIORITY = {
  'GET': 1,
  'POST': 2,
  'PUT': 3,
  'PATCH': 4,
  'DELETE': 5
};
return METHOD_PRIORITY[a.endpoint.method] - METHOD_PRIORITY[b.endpoint.method];

```

TBonnin: unrelated to this PR but I am not sure why we have 2 `.sort` chained one after each other. 
Also this sorting logic is becoming quite complex and I feel could be refactored like:
```ts
            groups.push({ name: group[0],
                flows: group[1].sort((a, b) => {
    
                    const comparePaths = (pathA, pathB) => {
                        const countPathSegments = (path) => (path.match(/\//g) || []).length;
                        const lenA = countPathSegments(pathA);
                        const lenB = countPathSegments(pathB);
                        if (lenA !== lenB) {
                            return lenA - lenB;
                        }
                        return pathA.localeCompare(pathB);
                    };

                    const compareMethods = (methodA, methodB) => { ... };

                    const pathComparison = comparePaths(a.endpoint.path, b.endpoint.path);

                    if (pathComparison !== 0) {
                        return pathComparison;
                    }

                    return compareMethods(a.endpoint.method, b.endpoint.method);
                })
            });
```

TBonnin (approved):

packages/webapp/src/pages/Integrations/providerConfigKey/Endpoints/Show.tsx
bodinsamuel: Changed with your first suggestion 👌🏻 
It split in two because I had issues containing the logic of grouping and sorting in the same sort 
(grouping by paths, then sorting alphabetically, then sorting by method)

PR #3044, opened 2024-11-22
feat(webapp): Show integration issues in their settings

nalanj:
Waiting on https://github.com/NangoHQ/nango/pull/3050

Shows missing field errors on the integrations page and highlights the errors on the tab:

image

And in the integrations list:

image

- Set up a new integration that needs configuration, but don't configure it
- View the integration page. You should see a yellow dot in the settings tab and a banner saying what fields need to be configured


packages/server/lib/controllers/config.controller.ts
```diff
@@ -131,7 +132,8 @@ class ConfigController {
                         provider: config.provider,
                         scripts: activeFlows.length,
                         connection_count: Number(config.connection_count),
-                        creationDate: config.created_at
+                        creationDate: config.created_at,
+                        missing_fields_count: config.missing_fields.length
```
TBonnin, line 136: why not returning the missing fields and doing the count in the FE?

packages/webapp/src/pages/Integrations/providerConfigKey/Settings/Show.tsx
```diff
@@ -41,6 +41,32 @@ export const SettingsSwitch: React.FC<{ data: GetIntegration['Success']['data'];
...
+        case 'TABLEAU':
+        case 'NONE':
+        case 'OAUTH2_CC':
+            return null;
+
+        default:
+            return null;
+    }
+};
+
```
TBonnin: can this function and the `settingsMissingX` ones be replaced by a map like this:
```
const FIELD_DISPLAY_NAMES = {
  OAUTH1: {
    oauth_client_id: 'Client ID',
    oauth_client_secret: 'Client Secret'
  },
  OAUTH2: {
    oauth_client_id: 'Client ID',
    oauth_client_secret: 'Client Secret'
  },
  TBA: {
    oauth_client_id: 'Client ID',
    oauth_client_secret: 'Client Secret'
  },
  APP: {
    oauth_client_id: 'App ID',
    oauth_client_secret: 'App Private Key',
    app_link: 'App Public Link'
  },
  CUSTOM: {
    oauth_client_id: 'Client ID',
    oauth_client_secret: 'Client Secret',
    app_link: 'App Public Link',
    app_id: 'App ID',
    private_key: 'App Private Key'
  }
} as const;

function missingFieldsMessage(data: GetIntegration['Success']['data']): string | null {
  const mappings = FIELD_DISPLAY_NAMES[data.template.auth_mode];
  if (!mappings) return null;
  
  return data.integration.missing_fields
    .map(field => mappings[field] || field)
    .join(', ');
};
```

TBonnin (approved):
I would try to simplify all those `switches` but looks good otherwise

packages/server/lib/controllers/config.controller.ts
nalanj: I mostly just figured since we're only indicating if there is or isn't an error that it'd be better and more compact to just return a count there. I'm down if you think we should swap it for the full array.

packages/webapp/src/pages/Integrations/providerConfigKey/Settings/Show.tsx
nalanj: See c2f84977

packages/webapp/src/pages/Integrations/providerConfigKey/Settings/components/General.tsx
```diff
@@ -13,11 +13,49 @@ import { InfoBloc } from '../../../../../components/InfoBloc';
...
+
+function missingFieldsMessage(
+    template: GetIntegration['Success']['data']['template'],
+    integration: GetIntegration['Success']['data']['integration']
+): string | null {
+    const mappings = FIELD_DISPLAY_NAMES[template.auth_mode];
+    if (!mappings) return null;
+
+    return integration.missing_fields.map((field) => mappings[field] || field).join(', ');
+}
```
bodinsamuel, line 53: Might have been more user friendly to display a global message and a message near each fields instead

PR #3045, opened 2024-11-22
feat(server): Add missing fields column for integrations to db

nalanj:
We need a stored set of issues on integrations so that we can show those issues in the UI. This will add the new column and ensure that it's handled appropriately in the app.

- [x] Add new column
- [x] Set up validation function for integrations
- [x] Update creating integrations to fill in fields
- [x] Update updating integrations to fill in fields

A future PR will migrate existing providers to note their missing fields correctly since, and then we'll be able to utilize those in the dashboard.

## How I Tested It
- Run the tests
- Click through adding a new oauth based integration
- Note that missing_fields is set for it in the DB
- Update that integration
- Note that missing_fields has been updated accordingly


packages/shared/lib/services/config.service.ts
```diff
@@ -236,6 +259,30 @@ class ConfigService {
 
         return { copiedToId: providerConfigResponse.id!, copiedFromId: foundConfigId };
     }
+
+    VALIDATION_RULES: ValidationRule[] = [
```
nalanj, line 268: @TBonnin 🥇

packages/shared/lib/services/config.service.ts
```diff
@@ -236,6 +259,30 @@ class ConfigService {
...
+            field: 'app_link',
+            modes: ['APP'],
+            isValid: (config) => !!config.app_link
+        }
+    ];
+
+    validateProviderConfig(authMode: AuthModeType, providerConfig: ProviderConfig): string[] {
+        return this.VALIDATION_RULES.filter((rule) => rule.modes.includes(authMode))
+            .filter((rule) => !rule.isValid(providerConfig))
+            .map((rule) => rule.field.toString());
```
TBonnin: I know I wrote this and I wish js was a real functional language but if we are worried about perf and want to avoid iterating through the rules several times we could used a single `flatMap` here. Rules list is small and unlikely to ever grow to an unmanageable size so I personally find the multiple filter/map easier to read but `flatMap` wouldn't be so bad

nalanj: See eaefdc9c. It feels a little cryptic to me, but I think it's mainly because I don't use `flatMap` much.

packages/shared/lib/services/config.service.ts
```diff
@@ -114,7 +122,14 @@ class ConfigService {
     }
 
     async createProviderConfig(config: ProviderConfig): Promise<ProviderConfig | null> {
+        const provider = getProvider(config.provider);
+        if (!provider) {
+            throw new NangoError('unknown_provider');
+        }
```
bodinsamuel: Is it possible (if not too annoying) to pass provider as params instead of loading it here? So it's side-effect "free"

packages/shared/lib/services/config.service.ts
```diff
@@ -236,6 +259,38 @@ class ConfigService {
...
+        {
+            field: 'private_key',
+            modes: ['CUSTOM'],
+            isValid: (config) => !!config.custom?.['private_key']
+        }
+    ];
+
+    validateProviderConfig(authMode: AuthModeType, providerConfig: ProviderConfig): string[] {
+        return this.VALIDATION_RULES.flatMap((rule) => (rule.modes.includes(authMode) && !rule.isValid(providerConfig) ? [rule.field] : []));
+    }
```
bodinsamuel, line 298: not for change: Could have been Zod imo, not necessarily quicker but more flexible
```ts
const val = z.discriminatedUnion('type', [
   z.object({ 
     type: z.enum('OAUTH1', 'OAUTH2'),
     oauth_client_id: z.string(), 
     oauth_client_secret: z.string() 
   })
  ]
)
val.safeParse({ type: authMode, ...providerConfig });
```

bodinsamuel (approved):

packages/shared/lib/services/config.service.ts
nalanj: Oh cool, I didn't know about `discriminatedUnion` in zod.

packages/shared/lib/services/config.service.ts
nalanj: See f428e437.

packages/database/lib/migrations/20241122164425_add_missing_fields_to_config.cjs
```diff
@@ -0,0 +1,13 @@
+exports.up = async function (knex) {
+    await knex.schema.raw(`
+        ALTER TABLE "_nango_configs" 
+        ADD COLUMN IF NOT EXISTS "missing_fields" text[] NOT NULL DEFAULT array[]::text[];
```
TBonnin, line 4: I have the feeling we might at some point need to store more than the missing fields but also the type of issues, etc...and maybe json would be more flexible than an array. No need to change now though, I am just playing nostradamus

packages/shared/lib/services/config.service.ts
TBonnin: oh yeah using `zod` for validation is also a good idea

TBonnin (approved):

packages/database/lib/migrations/20241122164425_add_missing_fields_to_config.cjs
nalanj: Your visions of the future are likely correct, good Seer.

PR #3046, opened 2024-11-22
fix: db pool acquireTimeoutMillis cannot be zero

TBonnin:
for migration we don't enforce any timeout (ie: timeout = 0) but acquireTimeoutMillis cannot be zero.
This commit specify a default value if timeout = 0


TBonnin:
@NangoHQ/engineers FYI force merging since it is a low risk change and server deployment is currently broken

PR #3048, opened 2024-11-22
fix(server): Remove secrets from in app integration api response

nalanj:
I realized were were sending some fields that we'd prefer not to send in the internal api responses for integrations. This PR makes sure those aren't returned.

## How I Tested It

- Go to an oauth based integration in the UI
- Ensure secrets aren't included in the API response for the integration


TBonnin (approved):

packages/server/lib/formatters/integration.ts
```diff
@@ -2,11 +2,16 @@ import type { ApiIntegration, ApiPublicIntegration, ApiPublicIntegrationInclude,
...
+    const apiIntegration = {
         ...data,
         created_at: data.created_at.toISOString(),
         updated_at: data.updated_at.toISOString()
     };
+
+    delete apiIntegration.oauth_client_secret_iv;
+    delete apiIntegration.oauth_client_secret_tag;
+
+    return apiIntegration;
```
bodinsamuel: The type `IntegrationConfig` is invalid, it states that those values are undefined but they are not, they can be nullable though. It can't be fixed yet because of legacy code (it's more or less always the issue each time). Right now, the best fix could look like this for the type:
```ts
export type ApiIntegration = Omit<Merge<IntegrationConfig, ApiTimestamps>, 'oauth_client_secret_iv' | 'oauth_client_secret_tag'>;
```

And then instead of spreading which does not trigger a Typescript error because the interface is still compatible, it should be assigned field by field (like endUser, connection). I forgot about this annoying "issue" for this formatter and a few others (team, invitation, etc.)

bodinsamuel (commented):
Thanks for checking that 👍🏻

bodinsamuel (approved):

PR #3049, opened 2024-11-22
fix: improve on-events script logs

TBonnin:
- each scripts has its own logs operation
- can filter by `Type = Events`

Screenshot 2024-11-22 at 16 36 59

https://linear.app/nango/issue/NAN-2219/on-events-script-follow-up

### How to tests:
1. Set `on-events` scripts in your nango.yaml
```
        on-events:
            post-connection-creation:
                - setup
            pre-connection-deletion:
                - cleanup

```
2. run `nango generate` and modify the scripts
3. run `nango deploy dev`
4. Go to the dashboard and create/delete a connection for the integration you added the scripts into
5. Check the logs

### DIFF
--- a/packages/logs/lib/models/helpers.ts
+++ b/packages/logs/lib/models/helpers.ts
@@ -140,7 +140,8 @@ export const operationTypeToMessage: Record<ConcatOperationList, string> = {
     'sync:unpause': 'Sync schedule started',
     'webhook:incoming': 'Received a webhook',
     'webhook:forward': 'Forwarding Webhook',
-    'events:run': 'On-event script execution'
+    'events:post_connection_creation': 'Post connection creation script execution',
+    'events:pre_connection_deletion': 'Pre connection creation script execution'
 };
 
 /**
--- a/packages/logs/lib/models/helpers.ts
+++ b/packages/logs/lib/models/helpers.ts
@@ -13,7 +13,7 @@ export interface FormatMessageData {
     environment?: { id: number; name: string } | undefined;
     connection?: { id: number; name: string } | undefined;
     integration?: { id: number; name: string; provider: string } | undefined;
-    syncConfig?: { id: number; name: string } | undefined;
+    syncConfig?: { id: number; name: string } | undefined; // TODO: rename to script or something similar because it also apply to actions and on-events scripts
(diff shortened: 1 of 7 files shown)


packages/logs/lib/models/helpers.ts
```diff
@@ -13,7 +13,7 @@ export interface FormatMessageData {
     environment?: { id: number; name: string } | undefined;
     connection?: { id: number; name: string } | undefined;
     integration?: { id: number; name: string; provider: string } | undefined;
-    syncConfig?: { id: number; name: string } | undefined;
```
TBonnin, line 16: `syncConfig` is already misnamed since it also applies to `action`. A better name could be simply `script`.
I attempted to rename it but it has a lot of implications including modifying ES models so I gave up cc @bodinsamuel

bodinsamuel: it was referencing the table which worked with action but indeed does not scale anymore. We can create a new field yes 👍🏻

packages/types/lib/logs/messages.ts
```diff
@@ -43,10 +43,16 @@ export interface OperationAction {
     type: 'action';
     action: 'run';
 }
+
+export interface OperationOnEvents {
+    type: 'events';
+    action: 'run';
```
bodinsamuel: why not put the actual event name here?

bodinsamuel (approved):

packages/types/lib/logs/messages.ts
TBonnin: at first I though that we wouldn't need to update the Operation when adding new event types but yes we could differentiate based on the event type. fixed in https://github.com/NangoHQ/nango/pull/3049/commits/7fe5d861dbfbc3509508e75c59a1d30d062ec6a3

nalanj (approved):
Looks good. For me in the logs these showed up under auth events:

image

TBonnin:
> Looks good. For me in the logs these showed up under auth events:

@nalanj Are you sure you have checkout/built correctly because it should show as `EVENTS` with the name of the script correctly set

nalanj:
@TBonnin Yeah, something weird was going on there. Got EVENT after another runthrough.

PR #3050, opened 2024-11-22
feat(server): Backfill missing_fields on _nango_config

nalanj:
Fills out `missing_fields` as best as we can based on `_nango_config`. Unfortunately the `CUSTOM` `auth_method` won't be able to be backfilled performantly, but all others were possible.

For all of these, the prior PR will cause them to show errors in the future when created or edited, so I think we're safe here without having to decrypt all of the `CUSTOM` auth based ones.

## How I tested

- Run the migration against dev db with oauth and app auth_method integrations that don't have their settings filled and don't have errors set yet. Feel free to just clear `missing_fields` in the DB as an easy way to handle that.
- Confirm in the database that the fields were correctly filled after the migration runs.


packages/database/lib/migrations/20241122212401_integration_backfill_missing_fields.cjs
```diff
@@ -0,0 +1,49 @@
...
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-var-requires */
+const path = require('node:path');
+const fs = require('node:fs');
+const yaml = require('js-yaml');
+
+exports.up = async function (knex) {
+    const providersPath = path.join(__dirname, '..', '..', '..', 'shared', 'providers.yaml');
+    const providers = yaml.load(fs.readFileSync(providersPath, 'utf8'));
```
nalanj: Is this reasonable? I had to be able to figure out the `auth_mode` and this was the most direct path I could come up with.

TBonnin: If providers.yaml is ever moved, this migration would fail. Probably not a problem for cloud but for fresh new install of Enterprise it would be problematic.  
At least we'll need to handle the missing files

I know we don't store the auth_mode in the db but could we infer it from the content of `_nango_configs` table. For instance if the `oauth_client_id` is defined or not.

packages/database/lib/migrations/20241122212401_integration_backfill_missing_fields.cjs
```diff
@@ -0,0 +1,49 @@
...
+
+    const needsClientId = ['OAUTH1', 'OAUTH2', 'TBA', 'APP'];
+    const clientIdProviders = Object.entries(providers)
+        .filter(([_, config]) => needsClientId.includes(config.auth_mode))
+        .map(([name]) => name);
```
TBonnin, line 24: we could take this opportunity to add a `auth_mode` column to `configs`. Bastien and I recently needed this info for instance.

packages/database/lib/migrations/20241122212401_integration_backfill_missing_fields.cjs
nalanj: Unfortunately we overload the usage of some of these fields in terms of types of auth so it's not the easiest to infer. See https://github.com/NangoHQ/nango/blob/9ddb80e8a9fc8e19dcf77f57d1e02042c4adf5f7/packages/webapp/src/pages/Integrations/providerConfigKey/Settings/components/App.tsx#L22-L23

packages/database/lib/migrations/20241122212401_integration_backfill_missing_fields.cjs
nalanj: OK cool. I wasn't sure if we had purposefully left it off or not.

packages/database/lib/migrations/20241122212401_integration_backfill_missing_fields.cjs
nalanj: @TBonnin See 28357288

nalanj:
In Slack we discussed adding `auth_mode` to `_nango_configs`, and it's more complicated than is desirable right now, with not much downside in terms of future work to write a migration and do it later, so we'll handle that in the future if needed.

packages/database/lib/migrations/20241122212401_integration_backfill_missing_fields.cjs
```diff
@@ -0,0 +1,55 @@
...
+const fs = require('node:fs');
+const yaml = require('js-yaml');
+
+exports.up = async function (knex) {
+    let providers;
+    const providersPath = path.join(__dirname, '..', '..', '..', 'shared', 'providers.yaml');
+    try {
+        providers = yaml.load(fs.readFileSync(providersPath, 'utf8'));
+    } catch (e) {
+        console.error(`Warning: Failed to load providers.yaml. ${e.message}. Skipping migration. Configuration warnings will not be accurate.`);
```
TBonnin: not sure what `Configuration warnings` are to be honest

packages/database/lib/migrations/20241122212401_integration_backfill_missing_fields.cjs
```diff
@@ -0,0 +1,55 @@
...
+    const needsAppLink = ['APP'];
+    const appLinkProviders = Object.entries(providers)
+        .filter(([_, config]) => needsAppLink.includes(config.auth_mode))
+        .map(([name]) => name);
+    await knex
+        .queryBuilder()
+        .from('_nango_configs')
+        .whereIn('provider', appLinkProviders)
+        .whereRaw("NOT (missing_fields @> '{app_link}')")
+        .update({ missing_fields: knex.raw("array_append(missing_fields, 'app_link')") });
```
TBonnin, line 52: I guess you can make the query a function to avoid the repetition. Not a big deal though

packages/database/lib/migrations/20241122212401_integration_backfill_missing_fields.cjs
```diff
@@ -0,0 +1,55 @@
...
+/* eslint-disable @typescript-eslint/no-var-requires */
+const path = require('node:path');
+const fs = require('node:fs');
+const yaml = require('js-yaml');
+
+exports.up = async function (knex) {
+    let providers;
+    const providersPath = path.join(__dirname, '..', '..', '..', 'shared', 'providers.yaml');
+    try {
+        providers = yaml.load(fs.readFileSync(providersPath, 'utf8'));
```
TBonnin, line 13: I wish we could have a compile time `fileExist` check so we get an error here next time we move the providers file

TBonnin (approved):

TBonnin (commented):
Can we estimate how fast the migration gonna be in prod? How many integrations do we have to go through?

packages/database/lib/migrations/20241122212401_integration_backfill_missing_fields.cjs
```diff
@@ -0,0 +1,55 @@
...
+    const appLinkProviders = Object.entries(providers)
+        .filter(([_, config]) => needsAppLink.includes(config.auth_mode))
+        .map(([name]) => name);
+    await knex
+        .queryBuilder()
+        .from('_nango_configs')
+        .whereIn('provider', appLinkProviders)
+        .whereRaw("NOT (missing_fields @> '{app_link}')")
+        .update({ missing_fields: knex.raw("array_append(missing_fields, 'app_link')") });
+};
```
bodinsamuel, line 53: You can make the script faster by skipping deleted one (about 25% in prod)

bodinsamuel (approved):

packages/database/lib/migrations/20241122212401_integration_backfill_missing_fields.cjs
nalanj: Yeah, I figured with it being in a migration it wasn't too big of a deal to DRY it up.

packages/database/lib/migrations/20241122212401_integration_backfill_missing_fields.cjs
nalanj: 🤔 yeah me neither. What was I thinking?

packages/database/lib/migrations/20241122212401_integration_backfill_missing_fields.cjs
nalanj: Yeah, I think it would actually make sense for us to make sure we're testing migrations from a clean state if possible to ensure that we catch that situation for self-installs.

nalanj:
@TBonnin There are only 16240 in prod, so this should run pretty quick, especially since I was able to remove any need to decrypt fields.

packages/database/lib/migrations/20241122212401_integration_backfill_missing_fields.cjs
nalanj: Updated in 4e49189b

PR #3056, opened 2024-11-25
fix: not using setInterval for otlp export config refresh

TBonnin:
We want to make sure no more than one updateRoutes occurs simultaneously.

Logs seems to indicate that otlp.updateRoutes was running multiple times concurrently on jobs service.
This commit gets rid of `setInterval` in favor of a simple loop with sleep

### DIFF
--- a/packages/logs/lib/otlp/otlp.ts
+++ b/packages/logs/lib/otlp/otlp.ts
@@ -5,7 +5,6 @@ import { ATTR_SERVICE_NAME } from '@opentelemetry/semantic-conventions';
 import { once, stringifyError } from '@nangohq/utils';
 import { logger } from '../utils.js';
 import { RoutingSpanProcessor } from './otlpSpanProcessor.js';
-import { setInterval } from 'timers';
 
 // Enable OpenTelemetry console logging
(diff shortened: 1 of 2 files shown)


packages/logs/lib/otlp/otlp.ts
```diff
@@ -53,10 +52,17 @@ async function updateRoutes(getRoutes: () => Promise<RouteConfig[]>) {
...
 
 export const otlp = {
+    running: false,
     register: once(async (getRoutes: () => Promise<RouteConfig[]>) => {
-        await updateRoutes(getRoutes);
-        setInterval(async () => await updateRoutes(getRoutes), 10000);
+        otlp.running = true;
+        while (otlp.running) {
+            await updateRoutes(getRoutes);
+            await new Promise((resolve) => setTimeout(resolve, 15000));
```
bodinsamuel, line 60: nb: you can use `node:timers` for this natively

bodinsamuel (approved):

nalanj (approved):

packages/logs/lib/otlp/otlp.ts
TBonnin: always forget about it, even though that what I used in the scheduler package. Trying to deploy asap though so leaving it

PR #3057, opened 2024-11-25
fix(runner): Remove setInterval from providers reload

nalanj:
`setInterval` can cause problems because it doesn't ensure that multiple callbacks don't run at the same time. This removes `setInterval` from provider reloading.

Note: I'm not sure if it's worth it in this case to deploy runners or just wait until we deploy them next time.

See https://linear.app/nango/issue/NAN-2245/fix-do-not-use-setinterval

## How I tested it:

- Make sure a runner has started
- Edit the providers.yaml
- Wait for the reloadInterval (60s by default) to ensure you get a message that it was updated

### DIFF
--- a/packages/shared/lib/services/providers.ts
+++ b/packages/shared/lib/services/providers.ts
@@ -6,6 +6,7 @@ import { NangoError } from '../utils/error.js';
 import { dirname } from '../utils/utils.js';
 import { getLogger, ENVS, parseEnvs } from '@nangohq/utils';
 import { createHash } from 'node:crypto';
+import { setTimeout } from 'node:timers/promises';
 
 const logger = getLogger('providers');
 
@@ -62,7 +63,7 @@ async function pollProviders() {
     const reloadInterval = envs.PROVIDERS_RELOAD_INTERVAL;
 
     while (polling) {
-        await new Promise((resolve) => setTimeout(resolve, reloadInterval));
+        await setTimeout(reloadInterval);
 
         try {
             const providersRaw = await fetchProvidersRaw(providersUrl);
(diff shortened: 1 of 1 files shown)


packages/shared/lib/services/providers.ts
```diff
@@ -42,22 +42,29 @@ export async function monitorProviders(): Promise<() => void> {
...
-        try {
-            const providersRaw = await fetchProvidersRaw(providersUrl);
-            const newProvidersHash = createHash('sha1').update(providersRaw).digest('hex');
-
-            if (newProvidersHash !== providersHash) {
-                providersHash = newProvidersHash;
-                providers = JSON.parse(providersRaw) as Record<string, Provider>;
-                logger.info(`Providers reloaded (${providersHash})`);
+    let running = true;
+    void (async function () {
```
TBonnin: I would be tempted to name the function instead of anonymous call to be a bit more descriptive
```
const pollProviders = async ...
pollProviders()
```
I don't feel strongly about it (was not named before) so up to you

TBonnin (approved):

packages/shared/lib/services/providers.ts
nalanj: I waffled on that - I'll go ahead and do it

packages/shared/lib/services/providers.ts
```diff
@@ -42,22 +42,29 @@ export async function monitorProviders(): Promise<() => void> {
...
+        while (running) {
+            await new Promise((resolve) => setTimeout(resolve, reloadInterval));
```
bodinsamuel: import { setTimeout } from node:timers  can help

bodinsamuel (approved):

packages/shared/lib/services/providers.ts
nalanj: Fixed in 9eeef416

packages/shared/lib/services/providers.ts
nalanj: Did that in dbcd7164. Also I didn't realize that was a thing. Thanks!

PR #3058, opened 2024-11-25
chore: Improve Salesforce API docs

rguldener:
Addresses the common confusion that people think Salesforce Developer Edition accounts are Salesforce sandboxes (they are accounts on the production infrastructure)


nalanj (approved):

bodinsamuel (approved):
Thanks 🙏🏻

PR #3060, opened 2024-11-26
fix(sdk): remove axiosError.config to avoid huge payload

TBonnin:
When proxying in a script fails, the axiosError being thrown and potentially returned by the runner could be huge. Jobs then struggle to deserialize this huge output from the runner.
This commit removes the axiosError.config before throwing it so we avoid huge runner output payload and logs


khaliqgant (approved):

TBonnin:
closing in favor of https://github.com/NangoHQ/nango/pull/3063

PR #3062, opened 2024-11-26
fix(authorization): authorization url encoding

hassan254-prog:
### Describe the problem and your solution
Turns out some of the providers don't like it when we encode some of the url parts, i.e `scopes`, `base_url` when we are interpolating with `connectionConfigs`. To help with this, the PR introduces a way we can handle this within the various parts. Currently we can skip authorization_url encoding within the `scopes`, or when interpolating the `base_url` with `connectionConfig`.

### Testing instructions
When creating a new connection for `Pennylane`, the scopes aren't encoded `customer_invoices+supplier_invoices+ledger+accounting`

### DIFF
--- a/packages/server/lib/controllers/oauth.controller.ts
+++ b/packages/server/lib/controllers/oauth.controller.ts
@@ -533,12 +533,27 @@ class OAuthController {
                     oauth2Client.getSimpleOAuth2ClientConfig(providerConfig, provider, connectionConfig)
                 );
 
+                const scopeSeparator = provider.scope_separator || ' ';
+                const scopes = providerConfig.oauth_scopes ? providerConfig.oauth_scopes.split(',').join(scopeSeparator) : '';
+
                 let authorizationUri = simpleOAuthClient.authorizeURL({
                     redirect_uri: callbackUrl,
-                    scope: providerConfig.oauth_scopes ? providerConfig.oauth_scopes.split(',').join(provider.scope_separator || ' ') : '',
+                    scope: scopes,
                     state: session.id,
                     ...allAuthParams
                 });
+
+                if (provider?.authorization_url_skip_encode?.includes('scopes')) {
+                    const url = new URL(authorizationUri);
+                    const queryParams = new URLSearchParams(url.search);
+                    queryParams.delete('scope');
+                    let newQuery = queryParams.toString();
+                    if (scopes) {
+                        newQuery = newQuery ? `${newQuery}&scope=${scopes}` : `scope=${scopes}`;
+                    }
+                    url.search = newQuery;
+                    authorizationUri = url.toString();
+                }
 
                 if (provider.authorization_url_fragment) {
                     const urlObj = new URL(authorizationUri);
--- a/packages/shared/providers.yaml
+++ b/packages/shared/providers.yaml
@@ -2591,7 +2591,8 @@ github-app-oauth:
     alias: github
     auth_mode: CUSTOM
     authorization_url: ${connectionConfig.appPublicLink}/installations/new
-    authorization_url_encode: false
+    authorization_url_skip_encode:
+        - base_url
     token_url:
         OAUTH2: https://github.com/login/oauth/access_token
         APP: https://api.github.com/app/installations/${connectionConfig.installation_id}/access_tokens
@@ -4469,6 +4470,8 @@ pennylane:
         grant_type: authorization_code
     refresh_params:
         grant_type: refresh_token
+    authorization_url_skip_encode:
+        - scopes
     docs: https://docs.nango.dev/integrations/all/pennylane
 
 peopledatalabs:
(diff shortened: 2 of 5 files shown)


packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -533,13 +533,28 @@ class OAuthController {
...
+                    const url = new URL(authorizationUri);
+                    const queryParams = new URLSearchParams(url.search);
+                    queryParams.delete('scope');
+                    let newQuery = queryParams.toString();
+                    if (scopes) {
+                        newQuery = newQuery ? `${newQuery}&scope=${scopes}` : `scope=${scopes}`;
+                    }
+                    url.search = newQuery;
+                    authorizationUri = url.toString();
+                }
```
bodinsamuel, line 556: ideally at some point we move all of this logic outside so it's unit testable

packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -533,13 +533,28 @@ class OAuthController {
...
+                const scopeSeparator = provider.scope_separator || ' ';
+                const scopes = providerConfig.oauth_scopes ? providerConfig.oauth_scopes.split(',').join(scopeSeparator) : '';
+
                 let authorizationUri = simpleOAuthClient.authorizeURL({
                     redirect_uri: callbackUrl,
-                    scope: providerConfig.oauth_scopes ? providerConfig.oauth_scopes.split(',').join(provider.scope_separator || ' ') : '',
+                    scope: scopes,
                     state: session.id,
                     ...allAuthParams
                 });
```
bodinsamuel, line 544: Might be easier at some point to not use `authorizeURL` and just iterate over the query params

bodinsamuel (approved):
👌🏻

packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -533,13 +533,28 @@ class OAuthController {
...
+                if (provider?.authorization_url_skip_encode?.includes('scopes')) {
+                    const url = new URL(authorizationUri);
+                    const queryParams = new URLSearchParams(url.search);
+                    queryParams.delete('scope');
+                    let newQuery = queryParams.toString();
+                    if (scopes) {
+                        newQuery = newQuery ? `${newQuery}&scope=${scopes}` : `scope=${scopes}`;
+                    }
+                    url.search = newQuery;
+                    authorizationUri = url.toString();
```
TBonnin, line 555: can this be simplified like this?
```ts
    const url = new URL(authorizationUri);
    url.searchParams.delete('scope');
    if (scopes) {
        url.searchParams.set('scope', scopes);
    }
    authorizationUri = url.toString();
```

packages/shared/providers.yaml
```diff
@@ -4469,6 +4470,8 @@ pennylane:
         grant_type: authorization_code
     refresh_params:
         grant_type: refresh_token
+    authorization_url_skip_encode:
+        - scopes
```
TBonnin, line 4474: I would personally prefer to bundle it with the `authorization_url` or `token_url` attributes. Something like:
```
authorization_url: 
   url: https://app.pennylane.com/oauth/authorize
   skip_encode: scopes
```
but it would require changing the type of an existing attributes 😓

TBonnin (approved):
I have the feeling that continuing to add more and more parameters to providers.yaml is going to haunt us at some point but this PR solves an immediate problem so go for it

packages/server/lib/controllers/oauth.controller.ts
hassan254-prog: Yes, the simplification works functionally, but it will still URL-encode the `scopes` value.

packages/shared/providers.yaml
hassan254-prog: I agree. I'm not sure, but to me, this might be the best time to clean up and bundle all related fields together for both `authorization_url` and `token_url`. Currently, we have loosely grouped fields for `authorization_url`, such as:
```
authorization_url_replacements
authorization_url_fragment
```

Bundling them together would make the `providers.yaml` file cleaner. @TBonnin what do you think?

packages/server/lib/controllers/oauth.controller.ts
hassan254-prog: Yes as the function is actually growing with addition of more parameters to providers.yaml.

packages/server/lib/controllers/oauth.controller.ts
hassan254-prog: Directly manipulating query parameters provides for a more controlled approach.

packages/shared/providers.yaml
TBonnin: grouping them together would make sense. Not in this PR of course. Maybe adding an 
```
authorization:
    url: ...
    url_replacements:...
    ...
```
and same thing with token. We would need to ensure backward compatibilities though. 😓

packages/server/lib/controllers/oauth.controller.ts
TBonnin: 🤦 nvm then

PR #3063, opened 2024-11-26
fix(runner): always truncate errors

bodinsamuel:
# Changes

We can't control what the content of errors, even the one feel like they are correct might still be too big.

- Fix runner should always truncate errors
- Remove more keys by default

### DIFF
--- a/packages/utils/lib/json.ts
+++ b/packages/utils/lib/json.ts
@@ -15,6 +15,10 @@ export function stringifyObject(value: any): string {
             }
             if (key === 'Authorization') {
                 return '[Redacted]';
+            } else if (key === 'httpAgent' || key === 'httpsAgent') {
+                return undefined;
+            } else if (key === 'stack') {
+                return undefined;
             }
             return value;
         },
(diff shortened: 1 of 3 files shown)


packages/utils/lib/json.ts
```diff
@@ -15,6 +15,10 @@ export function stringifyObject(value: any): string {
             }
             if (key === 'Authorization') {
                 return '[Redacted]';
+            } else if (key === 'httpAgent' || key === 'httpsAgent') {
+                return undefined;
+            } else if (key === 'stack') {
+                return undefined;
```
TBonnin, line 20:
```ts
} else if (['httpAgent', 'httpsAgent', 'trace'].includes(key)) {
    return undefined;
}
``` 
no?

TBonnin (approved):
should we also truncate the `NangoError` payload? Even if NangoErrors are internal we might by mistake fill it with something big

bodinsamuel:
> should we also truncate the NangoError payload? Even if NangoErrors are internal we might by mistake fill it with something big

yeah good point

PR #3065, opened 2024-11-26
feat(webhooks): send end user on connection creation

bodinsamuel:
## 🤓☝️ Changes

Fixes https://linear.app/nango/issue/NAN-1880/send-back-the-end-user-profile-in-the-webhook

- Send `endUser` in webhook on connection creation
When we use a session token, we pass metadata, but when we receive a webhook, we get a random connectionId that is impossible to link to an actual user. Sending back the userId (or orgId) will allow customers to link this connectionId to an actual user without needing to do any additional work.

NB: we don't send it in any other webhook since the connectionId should have been linked already

## 🧪 How to tests?

- Go to dashboard
- Modify webhook url to your favorite webhook dumper
- Allow webhook to be sent on connection creation
- Create a new connection with the connect UI

Screenshot 2024-11-26 at 17 32 59


packages/database/lib/getDbConfig.ts
bodinsamuel, line 1: Probably a leftover

packages/server/lib/utils/express.ts
```diff
@@ -1,9 +1,31 @@
...
+//           authType: 'publicKey';
+//           account: DBTeam;
+//           environment: DBEnvironment;
+//       }
+//     | {
+//           authType: 'basic' | 'session' | 'secretKey' | 'adminKey' | 'none';
+//           account?: DBTeam;
+//           environment?: DBEnvironment;
+//           user: Pick<DBUser, 'id' | 'email' | 'name'>;
+//       };
```
bodinsamuel, line 22: Tried to fix this type but opened a trap door (unfortunately I have wrote this type so I can only blame myself)
Will fix later

packages/server/lib/webhook/github-app-oauth-webhook-routing.ts
```diff
@@ -98,7 +98,8 @@ async function handleCreateWebhook(integration: ProviderConfig, body: any, logCo
                     environment,
                     account,
                     auth_mode: 'APP',
-                    operation: res.operation
+                    operation: res.operation,
+                    endUser: undefined // TODO fix this
```
bodinsamuel, line 102: Github App webhook routing is weird, but it's an edge case

packages/shared/lib/models/Webhook.ts
bodinsamuel, line 1: left over

nalanj (approved):
Looks good!

packages/server/lib/controllers/apiAuth.controller.ts
```diff
@@ -156,7 +157,8 @@ class ApiAuthController {
                     environment,
                     account,
                     auth_mode: 'API_KEY',
-                    operation: updatedConnection.operation
+                    operation: updatedConnection.operation,
+                    endUser: isConnectSession ? res.locals['endUser'] : undefined
```
TBonnin, line 161: why not directly `res.locals['endUser']`? is it not `undefined` if not `isConnectSession`?

packages/server/lib/controllers/apiAuth.controller.ts
```diff
@@ -28,10 +28,11 @@ import type { MessageRowInsert } from '@nangohq/types';
 
 class ApiAuthController {
     async apiKey(req: Request, res: Response<any, Required<RequestLocals>>, next: NextFunction) {
-        const { account, environment, authType } = res.locals;
+        const { account, environment } = res.locals;
         const { providerConfigKey } = req.params;
         const receivedConnectionId = req.query['connection_id'] as string | undefined;
         const connectionConfig = req.query['params'] != null ? getConnectionConfig(req.query['params']) : {};
+        const isConnectSession = res.locals['authType'] === 'connectSession';
```
TBonnin, line 35: `isConnectSession` could almost be an attribute of `RequestLocals`

packages/server/lib/utils/express.ts
TBonnin: not sure I follow but I trust you

TBonnin (approved):

packages/server/lib/controllers/apiAuth.controller.ts
bodinsamuel: true, I think I like the explicitness here and if (big if) I manage to fix the locals types `endUser` will be never in most case.

packages/server/lib/controllers/apiAuth.controller.ts
bodinsamuel: maybe indeed

packages/server/lib/utils/express.ts
bodinsamuel: This type is too loose meaning we can have connectSession with auth mode basic, which could lead to mistakes. Ideally, we have something stricter but when I wrote it I didn't anticipate this issue and now it seems some middleware does not enjoy the strictness when I tried to change it

PR #3067, opened 2024-11-26
fix(cli): show on-events scripts in nango deploy output

TBonnin:
`on-events` scripts were not shown in the CLI `nango deploy` output.
Now:
Screenshot 2024-11-26 at 13 26 38
`setup` and `cleanup` are the on-event scripts. Note that the versioning scheme for `on-events` is different from syncs/actions, not sure why.

How to test:
1. install cli local package: `npm install PATH/TO/REPO/packages/cli -g`. Might need to uninstall it before.
2. Have a integration setup with some `on-events` scripts
3. Run `nango deploy dev`

### DIFF
--- a/packages/shared/lib/models/NangoConfig.ts
+++ b/packages/shared/lib/models/NangoConfig.ts
@@ -139,4 +139,5 @@ export interface StandardNangoConfig {
     provider?: string;
     syncs: NangoSyncConfig[];
     actions: NangoSyncConfig[];
+    [`on-events`]: NangoSyncConfig[];
 }
--- a/packages/shared/lib/services/on-event-scripts.service.ts
+++ b/packages/shared/lib/services/on-event-scripts.service.ts
@@ -85,12 +85,14 @@ export const onEventScriptService = {
             }
             if (onEventInserts.length > 0) {
                 type R = Awaited<ReturnType<typeof onEventScriptService.update>>;
-                return await trx
-                    .insert<OnEventScript[]>(onEventInserts)
-                    .into(TABLE)
-                    .returning('*')
-                    .join('_nango_configs', `${TABLE}.config_id`, '_nango_configs.id')
-                    .select<R>([`${TABLE}.*`, '_nango_configs.provider_config_key as providerConfigKey']);
+                const res = await trx
+                    .with('inserted', (qb) => {
+                        qb.insert(onEventInserts).into(TABLE).returning('*');
+                    })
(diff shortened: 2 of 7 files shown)


packages/shared/lib/models/NangoConfig.ts
```diff
@@ -139,4 +139,5 @@ export interface StandardNangoConfig {
     provider?: string;
     syncs: NangoSyncConfig[];
     actions: NangoSyncConfig[];
+    [`on-events`]: NangoSyncConfig[];
```
TBonnin, line 142: Note that there is more work to make `on-events` script first-class citizen like syncs or actions. They are not shown in the UI, they are not supported for pre-built, etc... To keep in mind when we are gonna need to add support for those things cc @bastienbeurier 
For now I just needed to to add an attribute to `StandardNangoConfig` for things to compile. It would always be empty though.

packages/shared/lib/services/on-event-scripts.service.ts
```diff
@@ -84,8 +84,15 @@ export const onEventScriptService = {
...
             }
             if (onEventInserts.length > 0) {
-                await trx.insert(onEventInserts).into(TABLE);
+                type R = Awaited<ReturnType<typeof onEventScriptService.update>>;
+                return await trx
+                    .insert<OnEventScript[]>(onEventInserts)
+                    .into(TABLE)
+                    .returning('*')
+                    .join('_nango_configs', `${TABLE}.config_id`, '_nango_configs.id')
+                    .select<R>([`${TABLE}.*`, '_nango_configs.provider_config_key as providerConfigKey']);
```
TBonnin: must return the `providerConfigKey` and since it is not directly stored in the table but in the `nango_configs` one, I needed to add the join

nalanj (approved):
🚢 Looks good! Only weird thing I noticed was these versions:

image

TBonnin:
> 🚢 Looks good! Only weird thing I noticed was these versions:
> image

if by weird you refer to the different versioning system, I mentioned it in the PR description. I am not sure why it was decided to use a different versioning system when implementing the post-connection-script but I haven't revisited it.

nalanj:
> if by weird you refer to the different versioning system, I mentioned it in the PR description. I am not sure why it was decided to use a different versioning system when implementing the post-connection-script but I haven't revisited it.

Ah yep, that sentence had slipped my mind by the time I made it to the end of the review. Still looks good though 😁

PR #3068, opened 2024-11-26
fix(webapp): Invalidate cache on deleting integrations and adding connection from integration page

nalanj:
I noticed a couple of places where the connection list cache wasn't being correctly invalidated:

1. When creating a new connection from the integrations page (like right after creating a new integration).
2. When deleting an integration.

## How I Tested This

1. Go to Connections so the cache is warmed up
1. Create a new Unauthenticated Integration
2. Create a new connection for the new Unauthenticated integration
3. Return to Connections. The new connection should be in the list.
4. Go back to the new integration, go to the settings tab, and delete it.
5. Go back to connections. The new connection should now be gone.


packages/webapp/src/pages/Integrations/providerConfigKey/Show.tsx
```diff
@@ -59,16 +63,20 @@ export const ShowIntegration: React.FC = () => {
     const onEvent: OnConnectEvent = useCallback(
         (event) => {
             if (event.type === 'close') {
+                invalidateConnections(cache, mutate);
```
bodinsamuel: should not invalidate here, just on connect should be enough because you can close without connecting

packages/webapp/src/hooks/useConnections.tsx
```diff
@@ -38,6 +39,15 @@ export function useConnections(queries: GetConnections['Querystring']) {
     return { loading, error: error?.json, data, hasNext, offset: size, setOffset: setSize, mutate };
 }
 
+export function invalidateConnections(cache: Cache, mutate: ReturnType<typeof useSWRConfig>['mutate']) {
```
bodinsamuel: note this is more than invalidate, this is nuking the cache which will directly trigger a fetch call (vs soft cache invalidation which only triggers on next page change)

bodinsamuel (approved):
🧹 💨

packages/webapp/src/hooks/useConnections.tsx
nalanj: Yeah, I tried to only mutate it with an empty value so that it would invalidate, I cold never get that to work with `swr/infinite`. Renamed the function to `clearConnectionsCache` in 48b412cc

packages/webapp/src/pages/Integrations/providerConfigKey/Show.tsx
nalanj: Fixed in 48b412cc

PR #3069, opened 2024-11-26
fix(CLI): show on-events scripts in `nango deploy` confirmation message

TBonnin:
Depends on https://github.com/NangoHQ/nango/pull/3067

The CLI should show the added/deleted on-event scripts in the `nango deploy` confirmation message.

Screenshot 2024-11-26 at 15 52 20

How to test:
1. Install local CLI: `nango install -g PATH/TO/REPO/packages/cli`
2. Add/Remove `on-events` scripts in your nango.yaml
3. run `nango deploy dev` and check the confirmation message show the added/removed scripts


packages/shared/lib/services/on-event-scripts.service.ts
```diff
@@ -84,20 +124,88 @@ export const onEventScriptService = {
...
+            for (const script of provider.scripts) {
+                const key = `${config.id}:${script.name}:${script.event}`;
+
+                const maybeScript = existingMap.get(key);
+                if (maybeScript) {
+                    // Script already exists - it's an update
+                    res.updated.push(maybeScript);
+
+                    // Remove from map to track deletions
+                    existingMap.delete(key);
```
bodinsamuel: could named `previousMap` (or something like this) for clarity, I didn't understood why we were removing from existing if it was indeed existing

packages/shared/lib/services/on-event-scripts.service.ts
```diff
@@ -84,20 +124,88 @@ export const onEventScriptService = {
...
+                } else {
+                    // Script doesn't exist - it's new
+                    res.added.push({
+                        configId: config.id,
+                        name: script.name,
+                        version: '0.0.1',
```
bodinsamuel, line 197: Out of curiosity:
Did you talk with Bastien about adding on-events-scripts in the templates, managing their versioning, and being able to enable/disable them via the UI?

bodinsamuel (approved):

packages/shared/lib/services/on-event-scripts.service.ts
TBonnin: good idea. will fix

packages/shared/lib/services/on-event-scripts.service.ts
TBonnin: No but that's why I pinged him in a previous PR to mention that there would still be a fair amount of work to support the features you mentioned (pre-built, dasbhoard, etc...) for on-events scripts

PR #3072, opened 2024-11-27
feat(ui): new create connection page

bodinsamuel:
## 😘 Changes

Fixes https://linear.app/nango/issue/NAN-2188/test-connection-ui

- UI: New create connection page
~The goal is to give the opportunity for customers to change the `EndUser` (also education time) and display links to documentation. It's one more click but yeah.~

- UI: Keep the legacy page to a new endpoint for a moment
- UI: add new style for Button, Input, Select
https://www.figma.com/design/pMnSC7rhSM79Dqwhm7MFhd

- `POST /api/v1/connect/sessions` now accepts the same payload as the public one
- `GET /api/v1/meta` expose user UUID 
~Not sure what the use of id but it's handy to have static-random id for the endUser profile without exposing our auto-incremented number.~ unused in the end

## 🧪 Tests

- Go to UI > Connections > Create
- Also go to Integrations > any > Add connection

Screenshot 2024-12-02 at 13 17 52


packages/webapp/src/pages/Connection/Create.tsx
bodinsamuel, line 1: The diff is incorrect, it's a brand new file

tests/setupFiles.ts
```diff
@@ -1,7 +1,7 @@
 import { expect } from 'vitest';
 
-const dateRegex = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{1,6}Z$/;
-const dateRegexWithTZ = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.\d{1,6}\+\d{2}:\d{2}$/;
+const dateRegex = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(.\d{1,6})?Z$/;
+const dateRegexWithTZ = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(.\d{1,6})?\+\d{2}:\d{2}$/;
```
bodinsamuel: Had an issue with integrations tests, if you are very unlucky and the test is on the clock there will be no microsecond

nalanj:
image

This is a textual nitpick, but would "All configured integrations" make more sense here?

nalanj:
And another textual nitpick?

image

"Add to your app directly" feels like it skips on what you're adding. Would something like "Connect from your app" be clearer?

nalanj:
😄 Just ran though the UI (haven't read the code yet) but wanted to say that I really like this change!

packages/server/lib/controllers/v1/connect/sessions/postConnectSessions.ts
```diff
@@ -1,12 +1,14 @@
 import { asyncWrapper } from '../../../../utils/asyncWrapper.js';
 import { requireEmptyQuery, zodErrorToHTTP } from '@nangohq/utils';
 import type { PostConnectSessions, PostInternalConnectSessions } from '@nangohq/types';
-import { postConnectSessions } from '../../../connect/postSessions.js';
+import { bodySchema as originalBodySchema, postConnectSessions } from '../../../connect/postSessions.js';
```
nalanj, line 4: I'm not completely sure, but would `postConnectBodySchema` be clearer here?

nalanj (approved):
:+1:

TBonnin (commented):
looks good from a code point of view.
UX wise I find 2 things a bit confusing:
1. The Integration ID and dropdown is the first thing I see but even for me who know what it is about it is not clear why I am asked about it first and what it means. Since it is kind of optional I would put it below the `create test connection` with an explanation about what it is. Same thing about `the test user info`
I understand it is here for educational purpose but in this context it doesn't make much sense since me, the user, is the same person selecting the integration constraint and actually logging in. I think it is more confusing than anything and I would even vote in favor of not showing this option.

2. on the right part, the 3 components (docs button, outer Connect/Custom UI and the bulb/tag) all look like different clickable buttons but they all lead to the same url, the docs. I find that confusing

bodinsamuel:
Putting in draft until we improve the UX

bodinsamuel:
We have revamped the revamp: removed the end user profile (it's not entirely disabled just in case we change our mind again), clarify docs cards, improved the wording

TBonnin:
> We have revamped the revamp: removed the end user profile (it's not entirely disabled just in case we change our mind again), clarify docs cards, improved the wording

It is definitely better. 
The info block is quite big though. Maybe we could have a simple `Test connection user: user info is automatically populated with your Nango account details`.  and have the rest as a (?) button hover.

Also I feel like the dropdown and the authorize button should be next to each other or at least close. Right now drop down and authorize button are quite far away even though they are related while the 2 authorize buttons are close to each other even though they are not (`OR` situation). What about something like to mark the difference between new and old flow:
```
Create a test connection
Pick an integration
|------------------------|  |-----------| 
|      dropdown          |  | Authorize |
|------------------------|  |-----------|

(i) For test connections, the user info is automatically populated with your Nango account details. (?)

OR DEPRECATED FLOW
|-----------| 
| Authorize |
|-----------|
```

bodinsamuel:
@bastienbeurier for the UX feedback

bastienbeurier:
> > We have revamped the revamp: removed the end user profile (it's not entirely disabled just in case we change our mind again), clarify docs cards, improved the wording
> 
> It is definitely better. The info block is quite big though. Maybe we could have a simple `Test connection user: user info is automatically populated with your Nango account details`. and have the rest as a (?) button hover.
> 
> Also I feel like the dropdown and the authorize button should be next to each other or at least close. Right now drop down and authorize button are quite far away even though they are related while the 2 authorize buttons are close to each other even though they are not (`OR` situation). What about something like to mark the difference between new and old flow:
> 
> ```
> Create a test connection
> Pick an integration
> |------------------------|  |-----------| 
> |      dropdown          |  | Authorize |
> |------------------------|  |-----------|
> 
> (i) For test connections, the user info is automatically populated with your Nango account details. (?)
> 
> OR DEPRECATED FLOW
> |-----------| 
> | Authorize |
> |-----------|
> ```

Haha, @TBonnin, you're going to have to be involved in the design phase.

TBonnin:
> Haha, @TBonnin, you're going to have to be involved in the design phase.

No please no! 🤣

bastienbeurier:
> > Haha, @TBonnin, you're going to have to be involved in the design phase.
> 
> No please no! 🤣

@TBonnin @bodinsamuel the latest design proposal: https://www.figma.com/proto/hKL3cmI9vBwDoZyRahQnsq/Connections-Tab?node-id=4702-27943&node-type=instance&viewport=-3125%2C312%2C0.41&t=b11nWvdOcBhhizyh-0&scaling=min-zoom&content-scaling=fixed&starting-point-node-id=4702%3A27943&show-proto-sidebar=1

TBonnin (commented):
🚢

TBonnin (approved):

PR #3074, opened 2024-11-27
feat(api): expose credentials in GET /integrations/:uniqueKey

bodinsamuel:
## Changes

- `GET /integrations/:uniqueKey` now expose credentials (optional)
Some customers need the credentials for internal operations or oauth revoke. This expose optionally the credentials with some added security.

- Try to fix IntegrationConfig type by creating a smaller scope for integration create
This type is annoying, it's used a bit everywhere and everything is marked as undefined except it's null not undefined. Anyway, first steps since I needed better types.

### DIFF
--- a/packages/types/lib/utils.ts
+++ b/packages/types/lib/utils.ts
@@ -13,8 +13,7 @@ export interface Logger {
 type ValidateSelection<T, U> = U extends T ? U : never;
 export type PickFromUnion<T, U extends T> = ValidateSelection<T, U>;
 
-export type NullablePartial<
-    TBase,
-    NK extends keyof TBase = { [K in keyof TBase]: null extends TBase[K] ? K : never }[keyof TBase],
-    NP = Partial<Pick<TBase, NK>> & Pick<TBase, Exclude<keyof TBase, NK>>
-> = { [K in keyof NP]: NP[K] };
+export type NullablePartial<TBase, TNullableKey extends keyof TBase = { [K in keyof TBase]: null extends TBase[K] ? K : never }[keyof TBase]> = Partial<
+    Pick<TBase, TNullableKey>
+> &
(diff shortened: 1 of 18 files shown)


nalanj (approved):
Looks good. Tested it locally and it worked great too.

packages/types/lib/utils.ts
```diff
@@ -12,3 +12,9 @@ export interface Logger {
 
 type ValidateSelection<T, U> = U extends T ? U : never;
 export type PickFromUnion<T, U extends T> = ValidateSelection<T, U>;
+
+export type NullablePartial<
+    TBase,
+    NK extends keyof TBase = { [K in keyof TBase]: null extends TBase[K] ? K : never }[keyof TBase],
```
TBonnin: I would rename `NK` `TNullableKey` to follow the same pattern and make it a bit more descriptive

packages/types/lib/utils.ts
```diff
@@ -12,3 +12,9 @@ export interface Logger {
...
+    NP = Partial<Pick<TBase, NK>> & Pick<TBase, Exclude<keyof TBase, NK>>
+> = { [K in keyof NP]: NP[K] };
```
TBonnin: not sure you need NP. can we have 
```
export type NullablePartial<
   TBase, 
   TNullableKeys ...
> = Omit<TBase, TNullableKeys> & Partial<Pick<TBase, TNullableKeys>>
```

TBonnin (approved):
code looks good but I am not sure I understand the purpose of exposing integration credentials to the users when they are the ones who are providing those credentials in the first place

packages/types/lib/utils.ts
bodinsamuel: You forgot to add the non nullable keys, but yes I can simplify indeed

PR #3075, opened 2024-11-27
fix: do not wait for on-event script to finish when triggering them

TBonnin:
on-event scripts don't have output so waiting for them to finish is not necessary.
Also modifying the max duration timeout to 5mins for on-event script

How to test:
- modify your on-event script to sleep. Ex:
```
    const waitTime = 10;
    for (let i = 0; i < waitTime; i++) {
        await nango.log(`Countdown: ${waitTime - i}`);
        await new Promise((resolve) => setTimeout(resolve, 1000));
    }
```
- create/delete a connection and then check in the logs dashboard that the script is still running and you didn't have to wait for it

### DIFF
--- a/packages/orchestrator/lib/clients/client.ts
+++ b/packages/orchestrator/lib/clients/client.ts
@@ -254,17 +254,28 @@ export class OrchestratorClient {
         return this.execute(schedulingProps);
     }
 
-    public async executeOnEvent(props: ExecuteOnEventProps): Promise<ExecuteReturn> {
+    public async executeOnEvent(props: ExecuteOnEventProps): Promise<VoidReturn> {
         const { args, ...rest } = props;
         const schedulingProps = {
+            retry: { count: 0, max: 0 },
+            timeoutSettingsInSecs: {
+                createdToStarted: 30,
+                startedToCompleted: 5 * 60,
+                heartbeat: 99999
+            },
             ...rest,
             args: {
                 ...args,
                 type: 'on-event' as const
             }
         };
-        return this.execute(schedulingProps);
+        const res = await this.immediate(schedulingProps);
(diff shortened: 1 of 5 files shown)


packages/orchestrator/lib/clients/client.ts
```diff
@@ -254,17 +254,28 @@ export class OrchestratorClient {
...
     }
 
-    public async executeOnEvent(props: ExecuteOnEventProps): Promise<ExecuteReturn> {
+    public async executeOnEvent(props: ExecuteOnEventProps): Promise<VoidReturn> {
         const { args, ...rest } = props;
         const schedulingProps = {
+            retry: { count: 0, max: 0 },
+            timeoutSettingsInSecs: {
+                createdToStarted: 30,
+                startedToCompleted: 5 * 60,
```
TBonnin, line 263: like all our task timeouts, this value is arbitrary. 5 mins is very generous imho

nalanj (approved):
Ran it, looks good!

packages/server/lib/hooks/connection/on/connection-deleted.ts
```diff
@@ -52,8 +52,6 @@ export async function preConnectionDeletion({
         });
         if (res.isErr()) {
             await logCtx.failed();
-        } else {
-            await logCtx.success();
```
bodinsamuel, line 56: I don't see where you `success()` now?

nalanj: It's in that `handleOnEventSuccess` function, I think.

bodinsamuel: indeed, might be that I must have missed it 👍🏻

PR #3076, opened 2024-11-28
fix(api): improve user model usage

bodinsamuel:
## Context

Extracted from #3072

- Clarify our use of user's session
`req.user` is painful because session can be out of date, and it's hard to add/remove fields from session. It's easier to always use `res.locals[user] that's always fresh.

- Pass UUID, will be used later
- Fix regex for integration tests


packages/server/lib/utils/utils.ts
```diff
@@ -10,6 +10,7 @@ import { OrchestratorClient } from '@nangohq/nango-orchestrator';
 
 const logger = getLogger('Server.Utils');
 
+/** @deprecated TODO delete this */
```
nalanj, line 13: Would it make sense to just knock this out? It's only used in a couple of spots right now.

nalanj (approved):
Just one tiny comment that's not a deal-breaker.

packages/server/lib/utils/utils.ts
bodinsamuel: Yes but I wasn't sure about the fix, just noted for later

TBonnin (approved):

PR #3079, opened 2024-11-28
docs: change sample app port

bodinsamuel:
## Changes

- Change sample-app port that was conflicting with local env


nalanj (approved):
:shipit:

PR #3080, opened 2024-11-28
fix(sdk): expose getIntegration()

bodinsamuel:
## Changes

- Expose `getIntegration()` in the SDK
The shape of the response is determined by the include, which is maybe a bad idea I never know. On one hand it's pretty standard, on the other it makes caching and types hard :\


packages/shared/lib/sdk/sync.ts
```diff
@@ -443,11 +435,8 @@ export class NangoAction {
 
     public ActionError = ActionError;
 
-    private memoizedConnections: Map<string, { connection: Connection; timestamp: number } | undefined> = new Map<
-        string,
-        { connection: Connection; timestamp: number }
-    >();
-    private memoizedIntegration: GetPublicIntegration['Success']['data'] | undefined;
+    private memoizedConnections = new Map<string, { connection: Connection; timestamp: number }>();
+    private memoizedIntegration = new Map<string, { integration: GetPublicIntegration['Success']['data']; timestamp: number }>();
```
nalanj, line 438: Just curious here - what's the use case around needing to access any multiple integrations?

packages/shared/lib/sdk/sync.ts
```diff
@@ -659,6 +648,23 @@ export class NangoAction {
         return this.nango.getToken(this.providerConfigKey, this.connectionId);
     }
 
+    /**
+     * Get current integration
+     */
+    public async getIntegration(queries?: GetPublicIntegration['Querystring']): Promise<GetPublicIntegration['Success']['data']> {
```
nalanj, line 653: This is just an aside: rather than sync providers.yaml it feels like we could just pull this data at runtime for each job and not have to load the whole providers file that way.

nalanj (approved):
Looks good!

packages/shared/lib/sdk/sync.ts
bodinsamuel: yep agreed, there is a lot to improve regarding the task payload

packages/shared/lib/sdk/sync.ts
bodinsamuel: It's not the case here, the key is the sum of query params which can change the shape so it's going to store different shape of the same integration depending on the params. Not ideal (cf PR desc)

TBonnin: maybe this explaination should be in a comment. I am sure in 3 months I am gonna have forgotten why

bodinsamuel:
@TBonnin Haven't had the chance to get a review, I'll go ahead and merge but feel free to add comments after

TBonnin (commented):
Sorry. I reviewed it yesterday but forgot to submit it. :(

PR #3081, opened 2024-11-28
feat(sdk): automatically log http calls to API

bodinsamuel:
## Changes

- Automatically log http calls to API
While doing #3080 I noticed we don't log calls to our own API which can lead to confusion when you expect something to happen but don't manually log the result. Thanks to recent change, we now have a way to plug into axios and just listen to requests. 
It will log everything (i.e: triggerAction, getConnection, getIntegration, etc.), except proxy which doesn't go through node-client.

NB: This will for sure increase the log amount but just slightly since we do cache a lot. 

Screenshot 2024-11-28 at 16 11 02


packages/shared/lib/sdk/sync.ts
```diff
@@ -912,10 +925,49 @@ export class NangoAction {
...
             throw new Error(`Failed to log: ${JSON.stringify(response.data)}`);
         }
     }
+
+    private logAPICall(res: AxiosResponse): AxiosResponse {
+        if (!res.config.url) {
+            return res;
+        }
+
+        const valuesToFilter: string[] = [
```
nalanj, line 944: Just a small thought for the future: Could we compute our filter array once up front before any runner work and have the logger lean on it across all logging?

packages/shared/lib/sdk/sync.ts
```diff
@@ -912,10 +925,49 @@ export class NangoAction {
...
+
+        const method = res.config.method?.toLocaleUpperCase(); // axios put it in lowercase;
+        void this.log(
+            `${method} ${res.config.url}`,
+            {
+                type: 'http',
+                request: {
+                    method: method,
+                    url: redactURL({ url: res.config.url, valuesToFilter }),
+                    headers: redactHeaders({ headers: res.config.headers })
```
nalanj: Would this be more future-proof if we went ahead and included `valuesToFilter`?

nalanj (approved):
Just a few small comments :shipit:

packages/shared/lib/sdk/sync.ts
```diff
@@ -487,6 +488,13 @@ export class NangoAction {
                 };
             }
         }
+        if (!config.axios?.response) {
+            axiosSettings.interceptors = {
+                response: {
+                    onFulfilled: this.logAPICall.bind(this)
+                }
+            };
+        }
```
TBonnin, line 486: do I read correctly that it doesn't override the interceptors defined a few lines above that is used for snapshoting in dry-mode?

packages/shared/lib/sdk/sync.ts
```diff
@@ -763,7 +771,12 @@ export class NangoAction {
         const level = userDefinedLevel?.level ?? 'info';
 
         if (this.dryRun) {
-            logger[logLevelToLogger[level] ?? 'info'].apply(null, args as any);
+            // TODO: control this logger inside dryrun instead
+            if (args.length > 1 && 'type' in args[1] && args[1].type === 'http') {
+                logger[logLevelToLogger[level] ?? 'info'].apply(null, [args[0], { status: args[1]?.response?.code || 'xxx' }] as any);
+            } else {
+                logger[logLevelToLogger[level] ?? 'info'].apply(null, args as any);
+            }
```
TBonnin, line 779: Would something like this be easier to read
```
const status = args[1]?.type == 'http' ? args[1].response?.code: undefined
const logArgs =  status ? [args[0], { status }] : args
logger[logLevelToLogger[level] ?? 'info'].apply(null, logArgs as any);
```

packages/shared/lib/sdk/sync.ts
TBonnin: agreed. The valuesToFilter should not changed within a single execution

TBonnin (approved):
small comments. The most important is about precomputing `valuesToFilter`

packages/shared/lib/sdk/sync.ts
bodinsamuel: indeed

packages/shared/lib/sdk/sync.ts
bodinsamuel: yes exactly, I'll add a comment

packages/shared/lib/sdk/sync.ts
bodinsamuel: actually it's changing because connection's credentials can change during execution, but the value could be cached indeed. I'll add a comment

packages/shared/lib/sdk/sync.ts
bodinsamuel: honestly I don't know ahah
I have split the repetitive part

PR #3086, opened 2024-11-29
fix(data-ingestion): log end user

bodinsamuel:
## Changes

- Log end user to BigQuery
- Move endUser.service in shared because it's needed there
- Try to have more typesafety in the schema
- Try to add schema migration
- Add feature flag so we can test locally and enable this feature more easily


bodinsamuel:
Tested in staging

nalanj (commented):
I'm not able to run this because I don't have access to BigQuery (I don't think), but code looks reasonable. It feels like we have so much duplicated code under `jobs/lib/execution`, but I'm not really sure I can speak to it needing to be refactored or not.

packages/jobs/lib/execution/action.ts
```diff
@@ -47,6 +51,11 @@ export async function startAction(task: TaskAction): Promise<Result<void>> {
             throw new Error(`Action config not found: ${task.id}`);
         }
 
+        const getEndUser = await getEndUserByConnectionId(db.knex, { connectionId: task.connection.id });
+        endUser = getEndUser.isOk()
+            ? { id: getEndUser.value.id, endUserId: getEndUser.value.endUserId, orgId: getEndUser.value.organization?.organizationId || null }
+            : null;
```
TBonnin: `endUser` is already null so why not `if(getEndUser.isOk()) { endUser = ... }`

packages/shared/lib/services/endUser.service.ts
```diff
@@ -158,3 +158,16 @@ export async function linkConnection(db: Knex, { endUserId, connection }: { endU
 export async function unlinkConnection(db: Knex, { connection }: { connection: Pick<StoredConnection, 'id'> }) {
     await db<StoredConnection>('_nango_connections').where({ id: connection.id! }).update({ end_user_id: null });
 }
+
+export async function getEndUserByConnectionId(db: Knex, props: { connectionId: number }): Promise<Result<EndUser, EndUserError>> {
+    const endUser = await db<DBEndUser>(END_USERS_TABLE)
```
TBonnin: do we need the generic here since it is specified by the select below?

packages/shared/lib/sdk/sync.ts
```diff
@@ -392,6 +392,7 @@ export interface NangoProps {
     runnerFlags: RunnerFlags;
     debug: boolean;
     startedAt: Date;
+    endUser: { id: number; endUserId: string | null; orgId: string | null } | null;
```
TBonnin, line 386: why not `endUser: EndUser | null`

TBonnin, line 386: we passing the endUser only so we can have it when handling the job result. Have you thought about fetching it from the db when starting the jobs AND again when handling the job result?

TBonnin (approved):
haven't test the big query schema migration but the rest looks ok. Made a few comments but nothing blocking

packages/shared/lib/sdk/sync.ts
bodinsamuel: Because the schema will change at some point and I don't need anything else, plus it's passing through zod which would be annoying to verify. 
We do the same for team (and environment but differently)

bodinsamuel: yes, I'm not a fan. The way I see it, a job should be self-contained you send a payload and you receive the same thing, plus querying a new thing is one more possible way to fail. Tbf almost everything we put in NangoProps is not for execution per se.
Wdyt?

packages/jobs/lib/execution/action.ts
bodinsamuel: indeed

packages/shared/lib/services/endUser.service.ts
bodinsamuel: true

bodinsamuel:
Im going to go ahead and merge since it's for billing, but happy to iterate if you find it useful ☺️ 👍🏻

packages/shared/lib/sdk/sync.ts
TBonnin: I like the self-contained aspect of NangoProps but that's a lot of data doing the round trip. I am not sure what makes the most sense tbh

PR #3089, opened 2024-11-30
feat: Adds personio_v2 provider

tonibardina:
The base url for the `Personio` provider was including the `v1` path so I was not able to make calls to the the `v2` of the api. To not break it for the current users I have followed the same approach that you have used with the twitter api and decided to create the provider for the api `v2` version.


packages/webapp/public/images/template-logos/personio-v2.svg
bodinsamuel, line 1: If you can just symlink the v1 instead 👍🏻

packages/shared/providers.yaml
```diff
@@ -4582,6 +4582,21 @@ personio:
...
+personio-v2:
+    display_name: Personio (v2)
+    categories:
+        - hr
+    auth_mode: OAUTH2_CC
+    proxy:
+        base_url: https://api.personio.de/v2
+        headers:
+            Content-Type: application/x-www-form-urlencoded
+    token_url: https://api.personio.de/v2/auth/token
```
bodinsamuel, line 4644: @NangoHQ/engineers 
Should make the `version` a `connectionConfig` instead with a default value? so we don't need a v3, v4 etc.

docs-v2/integrations/all/personio-v2.mdx
```diff
@@ -0,0 +1,46 @@
+---
+title: Personio (v2)
+sidebarTitle: Personio (v2)
+---
+
+API configurations: [`personio`](https://nango.dev/providers.yaml)
```
bodinsamuel: I think you can remove this page and just add `personio-v2` in personio.mdx

bodinsamuel (commented):
Thanks for you contribution 🙏🏻 
I think we need to determine the best strategy around this issue

docs-v2/integrations/all/personio-v2.mdx
tonibardina: Hey @bodinsamuel I just wonder if we can merge both since they are slightly different:

```yaml

personio:
    display_name: Personio
    categories:
        - hr
    auth_mode: OAUTH2_CC
    proxy:
        base_url: https://api.personio.de/v1
        headers:
            x-personio-partner-id: ${connectionConfig.partnerId}
            x-personio-app-id: ${connectionConfig.appId}
    token_url: https://api.personio.de/v1/auth
    body_format: json
    docs: https://docs.nango.dev/integrations/all/personio
    connection_config:
        partnerId:
            type: string
            title: Partner ID
            description: The partner ID of your Personio account
        appId:
            type: string
            title: App ID
            description: The app ID of your Personio account

personio-v2:
    display_name: Personio (v2)
    categories:
        - hr
    auth_mode: OAUTH2_CC
    proxy:
        base_url: https://api.personio.de/v2
        headers:
            Content-Type: application/x-www-form-urlencoded
    token_url: https://api.personio.de/v2/auth/token
    scope_separator: ' '
    docs: https://docs.nango.dev/integrations/all/personio
    token_params:
        grant_type: client_credentials
```

bodinsamuel: Good point. Maybe we should create a v2 section in the same doc.

tonibardina: Do you have any example of an integration where multiple versions are explained so I can make sure we are being consistent on how things are shown?

packages/shared/providers.yaml
khaliqgant: I think that would be a bit troublesome in the event they want to upgrade to a newer version of the API. I think we should at some point deprecate al the providers that have a version in the `base_url` and make all customers specify a version in their integration scripts. 

The base url IMO for this should ideally just be

```
base_url: https://api.personio.de
```

bodinsamuel: the problem is the token_url, we need the version there

khaliqgant: 😑 didn't see that. That's unfortunate. Guess a version in the connectionConfig is the best approach in that case.

docs-v2/integrations/all/personio-v2.mdx
bodinsamuel: We never did it so I guess something simple like

```
## API v2
[... requirements]
```

packages/shared/providers.yaml
bodinsamuel: Team agreed that it's too challenging to solve right now

docs-v2/integrations/all/personio-v2.mdx
tonibardina: Hey @bodinsamuel actually there were no big differences between both versions from the nango perspective. What I have done is adding a `Api V1` tag to the requirements that are specific for that version

bodinsamuel (approved):

PR #3096, opened 2024-12-02
ci: split CLI/Node tests 

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2181/ci-split-all-cli-node-client-out-of-main-tests-suite

- Split CLI and node client tests and execute in a dedicated matrix 
They run 6 tests (Node 18, 20, 22 x windows or ubuntu). It's a lot but seems relatively fast. Alternatively we could make windows matrix only execute on master.  
We could further optimize this and run them only types or node or cli has changed but it's a pain to create this kind of logic.

- Integration and regular tests only run on Ubuntu + .nvmrc
This doesn't save much time since the matrix was running in parallel but reduce potential for flaky tests


.github/workflows/tests.yaml
```diff
@@ -14,25 +14,21 @@ concurrency:
...
+            - name: Use Node.js
               uses: actions/setup-node@v4
               with:
                   cache: 'npm'
-                  node-version: ${{ matrix.node-version }}
+                  node-version-file: '.nvmrc'
 
             - run: npm ci
             - run: npm run ts-build
             - run: npm run test:unit
```
TBonnin: can we exclude `package/cli` here since it is already tested by the other action? Would it matter?

TBonnin (approved):
Thank you very much for doing this. Did you time the speed gain?

.github/workflows/tests.yaml
bodinsamuel: true

bodinsamuel:
> Thank you very much for doing this. Did you time the speed gain?

I don't think there is much speed gain here since stuff were running in parallel. Windows still takes a long time.

PR #3100, opened 2024-12-03
fix(proxy): handle more content-disposition edge cases

hassan254-prog:
Refactored the response handling to cover more edge cases. Instead of checking if the `content-disposition` header is exactly equal to `attachment`, we now check if it includes `attachment` or `inline`, allowing for more flexibility. This change improves the handling of different content types by considering both attachment and inline dispositions.

Initially this was producing a malformed response with the dropbox and front providers, but after the fix, it works as expected.


packages/server/lib/controllers/proxy.controller.ts
```diff
@@ -265,13 +265,17 @@ class ProxyController {
...
+        const contentType = responseStream.headers['content-type'] || '';
+        const contentDisposition = responseStream.headers['content-disposition'] || '';
+        const transferEncoding = responseStream.headers['transfer-encoding'] || '';
+        const contentEncoding = responseStream.headers['content-encoding'] || '';
 
-        if (isChunked || isEncoded || isAttachment) {
+        const isJsonResponse = contentType.includes('application/json');
+        const isChunked = transferEncoding === 'chunked';
+        const isEncoded = Boolean(contentEncoding);
+        const isAttachmentOrInline = /(attachment|inline)/i.test(contentDisposition);
```
bodinsamuel: Actually I realized I didn't take into account that content-disposition can also contain a filename (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) 

```suggestion
        const isAttachmentOrInline = /^(attachment|inline)(;|\s|$)/i.test(contentDisposition);
```
https://regex101.com/r/se1kms/1

hassan254-prog: updated with the latest [commit](https://github.com/NangoHQ/nango/pull/3100/commits/13028615fc81763ec56eea7bd865790c067b0322). Yes, it was causing providers that return a `Content-Disposition` header with a filename to return malformed responses.

bodinsamuel (approved):

khaliqgant (approved):

PR #3103, opened 2024-12-03
fix(server): Load errorLog regardless of credential refresh result

nalanj:
When viewing a connection we had a strange state that was possible where there was an auth error on a connection but because it was not a connection that could be refreshed it wasn't showing the error in the UI once viewing the settings.

See https://linear.app/nango/issue/NAN-2255/authorization-tab-for-connections-sometimes-doesnt-show-auth-errors

## How I tested it

I set up an auth error row in my test db on a connection that didn't refresh (I used Unauthenticated) and confirmed that it displayed correctly. Here's a sample insert (you'll need to tweak it to get it to work with a connection you have set up):

```
INSERT INTO "nango"."_nango_active_logs"("id","type","action","connection_id","log_id","active","sync_id","created_at","updated_at")
VALUES
(391,E'auth',E'token_refresh',73,E'begpVdnwQLenQDHNVicf',TRUE,NULL,E'2024-11-26 16:53:33.628375+00',E'2024-11-26 16:53:33.628375+00');
```

Before this change you would see the auth error in the connections list, but visiting the connection would not show a dot on the auth tab or show an error on that page. After this change you'll see the dot and the error.

### DIFF
--- a/packages/server/lib/controllers/v1/connections/connectionId/getConnection.ts
+++ b/packages/server/lib/controllers/v1/connections/connectionId/getConnection.ts
@@ -83,31 +83,18 @@ export const getConnection = asyncWrapper<GetConnection>(async (req, res) => {
         onRefreshSuccess: connectionRefreshSuccessHook,
         onRefreshFailed: connectionRefreshFailedHook
     });
-    if (credentialResponse.isErr()) {
-        const errorLog = await errorNotificationService.auth.get(connection.id!);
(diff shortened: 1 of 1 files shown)


bodinsamuel (commented):
Indeed, the whole get and/or refresh credentials situation is really a pain to manage

bodinsamuel (approved):

PR #3105, opened 2024-12-03
feat: introducing the fleet package

TBonnin:
The fleet package will be used to managed the lifecycle of runners 

This commit includes:
- migration for deployments and nodes tables
- deployments and nodes models (with tests) 

This commit DOESN'T include (I though that would be enough for a first PR):
- the control loop to handle nodes state transition
- the routing logic

# How to test
The code is not used so it is not possible to run it manually. There is tests though :)

### DIFF
--- a/packages/fleet/lib/db/migrations/20241102115903_initial_fleet_models.ts
+++ b/packages/fleet/lib/db/migrations/20241102115903_initial_fleet_models.ts
@@ -44,10 +44,6 @@ export async function up(knex: Knex): Promise<void> {
             );
         `);
         await trx.raw(`
-            CREATE UNIQUE INDEX idx_${NODES_TABLE}_routingId_deploymentId
-                ON ${NODES_TABLE}(routing_id, deployment_id);
-        `);
-        await trx.raw(`
             CREATE INDEX idx_${NODES_TABLE}_routingId_state
                 ON ${NODES_TABLE}(routing_id, state)
                 WHERE state IN ('PENDING', 'STARTING', 'RUNNING', 'OUTDATED');
--- a/packages/fleet/lib/db/migrations/20241102115903_initial_fleet_models.ts
+++ b/packages/fleet/lib/db/migrations/20241102115903_initial_fleet_models.ts
@@ -43,6 +43,10 @@ export async function up(knex: Knex): Promise<void> {
             );
         `);
         await trx.raw(`
+            CREATE UNIQUE INDEX idx_${NODES_TABLE}_routingId_deploymentId
+                ON ${NODES_TABLE}(routing_id, deployment_id);
+        `);
+        await trx.raw(`
             CREATE INDEX idx_${NODES_TABLE}_routingId_state
                 ON ${NODES_TABLE}(routing_id, state)
                 WHERE state IN ('PENDING', 'STARTING', 'RUNNING', 'OUTDATED');
--- a/packages/fleet/lib/db/migrations/20241102115903_initial_fleet_models.ts
+++ b/packages/fleet/lib/db/migrations/20241102115903_initial_fleet_models.ts
@@ -6,7 +6,8 @@ export async function up(knex: Knex): Promise<void> {
     await knex.transaction(async (trx) => {
         await trx.raw(`
             CREATE TABLE ${DEPLOYMENTS_TABLE} (
-                commit_id char(40) PRIMARY KEY,
+                id SERIAL PRIMARY KEY,
+                commit_id char(40) NOT NULL,
                 created_at timestamp with time zone NOT NULL DEFAULT CURRENT_TIMESTAMP,
                 superseded_at timestamp with time zone
             );
@@ -30,7 +31,7 @@ export async function up(knex: Knex): Promise<void> {
             CREATE TABLE ${NODES_TABLE} (
                 id SERIAL PRIMARY KEY,
                 routing_id varchar(255) NOT NULL,
-                deployment_id char(40) NOT NULL REFERENCES deployments(commit_id) ON DELETE CASCADE,
+                deployment_id int NOT NULL REFERENCES deployments(id) ON DELETE CASCADE,
                 url varchar(1024),
                 state node_states NOT NULL,
                 image varchar(255) NOT NULL,
--- a/packages/fleet/lib/db/migrations/20241102115903_initial_fleet_models.ts
+++ b/packages/fleet/lib/db/migrations/20241102115903_initial_fleet_models.ts
@@ -1,0 +1,57 @@
+import type { Knex } from 'knex';
+import { NODES_TABLE } from '../../models/nodes.js';
+import { DEPLOYMENTS_TABLE } from '../../models/deployments.js';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.transaction(async (trx) => {
+        await trx.raw(`
+            CREATE TABLE ${DEPLOYMENTS_TABLE} (
+                commit_id char(40) PRIMARY KEY,
+                created_at timestamp with time zone NOT NULL DEFAULT CURRENT_TIMESTAMP,
+                superseded_at timestamp with time zone
+            );
+        `);
+        await trx.raw(`
+            CREATE INDEX idx_${DEPLOYMENTS_TABLE}_active ON ${DEPLOYMENTS_TABLE}(superseded_at) WHERE superseded_at IS NULL;
+        `);
+        await trx.raw(`
+            CREATE TYPE node_states AS ENUM (
+                'PENDING',
+                'STARTING',
+                'RUNNING',
+                'OUTDATED',
+                'FINISHING',
+                'IDLE',
+                'TERMINATED',
+                'ERROR'
+            );
+        `);
+        await trx.raw(`
+            CREATE TABLE ${NODES_TABLE} (
+                id SERIAL PRIMARY KEY,
+                routing_id varchar(255) NOT NULL,
+                deployment_id char(40) NOT NULL REFERENCES deployments(commit_id) ON DELETE CASCADE,
+                url varchar(1024),
+                state node_states NOT NULL,
+                image varchar(255) NOT NULL,
+                cpu_milli int NOT NULL,
+                memory_mb int NOT NULL,
+                storage_mb int NOT NULL,
+                error text,
+                created_at timestamp with time zone NOT NULL DEFAULT CURRENT_TIMESTAMP,
+                last_state_transition_at timestamp with time zone NOT NULL DEFAULT CURRENT_TIMESTAMP
+            );
+        `);
+        await trx.raw(`
+            CREATE INDEX idx_${NODES_TABLE}_routingId_state
+                ON ${NODES_TABLE}(routing_id, state)
+                WHERE state IN ('PENDING', 'STARTING', 'RUNNING', 'OUTDATED');
+        `);
+    });
+}
+
+export async function down(knex: Knex): Promise<void> {
+    await knex.raw(`DROP TABLE IF EXISTS ${DEPLOYMENTS_TABLE}`);
+    await knex.raw(`DROP TABLE IF EXISTS ${NODES_TABLE}`);
+    await knex.raw(`DROP TYPE IF EXISTS node_states`);
+}
--- a/packages/fleet/lib/fleet.integration.test.ts
+++ b/packages/fleet/lib/fleet.integration.test.ts
@@ -1,0 +1,29 @@
+import { expect, describe, it, beforeAll, afterAll, afterEach } from 'vitest';
+import { Fleet } from './fleet.js';
+import { getTestDbClient, testDbUrl } from './db/helpers.test.js';
+import { generateCommitHash } from './models/helpers.test.js';
+
+describe('fleet', () => {
+    const fleetId = 'my_fleet';
+    const fleet = new Fleet({ fleetId, dbUrl: testDbUrl });
+
+    beforeAll(async () => {
+        await fleet.migrate();
+    });
+
+    afterEach(() => {});
+
+    afterAll(async () => {
+        await getTestDbClient(fleetId).clearDatabase();
+    });
+
+    describe('deploy', () => {
+        it('should create a new deployment', async () => {
+            const commitId = generateCommitHash();
+            const deployment = (await fleet.deploy(commitId)).unwrap();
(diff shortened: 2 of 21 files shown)


packages/fleet/lib/db/migrations/20241102115903_initial_fleet_models.ts
```diff
@@ -0,0 +1,57 @@
...
+                'IDLE',
+                'TERMINATED',
+                'ERROR'
+            );
+        `);
+        await trx.raw(`
+            CREATE TABLE ${NODES_TABLE} (
+                id SERIAL PRIMARY KEY,
+                routing_id varchar(255) NOT NULL,
+                deployment_id char(40) NOT NULL REFERENCES deployments(commit_id) ON DELETE CASCADE,
```
bodinsamuel: Need a unique index for those two no?

packages/fleet/lib/fleet.ts
```diff
@@ -0,0 +1,19 @@
+import type { Result } from '@nangohq/utils';
+import { DatabaseClient } from './db/client.js';
+import * as deployments from './models/deployments.js';
+import type { CommitHash, Deployment } from './types.js';
+
+export class Fleet {
+    private dbClient: DatabaseClient;
+    constructor({ fleetId, dbUrl }: { fleetId: string; dbUrl: string }) {
+        this.dbClient = new DatabaseClient({ url: dbUrl, schema: fleetId });
```
bodinsamuel, line 9: why not passing the DBClient directly?

packages/fleet/lib/models/deployments.ts
```diff
@@ -0,0 +1,78 @@
...
+                commit_id: commitId,
+                created_at: now,
+                superseded_at: null
+            };
+            const [inserted] = await trx.into<DBDeployment>(DEPLOYMENTS_TABLE).insert(dbDeployment).returning('*');
+            if (!inserted) {
+                return Err(new Error(`Error: no deployment '${commitId}' created`));
+            }
+            return Ok(DBDeployment.to(inserted));
+        });
```
bodinsamuel, line 57: We don't really do it rn but there is no way to rollback a deploy? or maybe reinsert the same commitHash maybe?

packages/fleet/lib/models/deployments.ts
```diff
@@ -0,0 +1,78 @@
...
+    } catch (err: unknown) {
+        return Err(new Error(`Error creating deployment '${commitId}': ${stringifyError(err)}`));
```
bodinsamuel: Instead of stringifying you can use `new Error(msg, { cause })` which was made for this.
Also could be great to have a dedicated Error type (e.g: FleetError) just for clarity

packages/fleet/lib/models/nodes.integration.test.ts
```diff
@@ -0,0 +1,151 @@
...
+                routingId: 'my-routing-id',
+                deploymentId: activeDeployment,
+                url: 'http://localhost:3000',
+                image: 'nangohq/my-image:latest',
+                cpuMilli: 500,
+                memoryMb: 1024,
+                storageMb: 512
+            })
+        ).unwrap();
+        expect(node).toMatchObject({
```
bodinsamuel: nb: just in case you don't know, toMatchObject is partial matching, usually I prefer to use strictEqual so there is no missing fields

packages/fleet/lib/models/nodes.ts
```diff
@@ -0,0 +1,218 @@
...
+        created_at: now,
+        last_state_transition_at: now
+    };
+    try {
+        const inserted = await db.from<DBNode>(NODES_TABLE).insert(newNode).returning('*');
+        if (!inserted?.[0]) {
+            return Err(new Error(`Error: no node '${nodeProps.routingId}' created`));
+        }
+        return Ok(DBNode.from(inserted[0]));
+    } catch (err: unknown) {
```
bodinsamuel: it's unknown by default
```suggestion
    } catch (err) {
```

packages/fleet/lib/models/nodes.ts
```diff
@@ -0,0 +1,218 @@
...
+        return Err(new Error(`Error creating node '${nodeProps.routingId}': ${stringifyError(err)}`));
```
bodinsamuel: Sometimes you try/catch db sometimes not, are we sure it's throwing? maybe only for insert? or also select? if yes should you try/catch everywhere?

packages/fleet/lib/models/nodes.ts
```diff
@@ -0,0 +1,218 @@
...
+): Promise<Result<Node>> {
+    return db.transaction(async (trx) => {
+        const node = await get(trx, props.nodeId, { forUpdate: true });
+        if (node.isErr()) {
+            return Err(new Error(`Node '${props.nodeId}' not found`));
+        }
+
+        const transition = NodeStateTransition.validate({ from: node.value.state, to: props.newState });
+        if (transition.isErr()) {
+            return Err(transition.error);
```
bodinsamuel:
```suggestion
            return transition.error;
```

packages/fleet/lib/models/nodes.ts
```diff
@@ -0,0 +1,218 @@
...
+    db: knex.Knex,
+    props: {
+        nodeId: number;
+        newState: Omit<NodeState, 'ERROR'>;
+    }
+): Promise<Result<Node>> {
+    return db.transaction(async (trx) => {
+        const node = await get(trx, props.nodeId, { forUpdate: true });
+        if (node.isErr()) {
+            return Err(new Error(`Node '${props.nodeId}' not found`));
```
bodinsamuel:
```suggestion
            return node.error;
```

bodinsamuel (approved):

packages/fleet/lib/db/migrations/20241102115903_initial_fleet_models.ts
TBonnin: indeed

packages/fleet/lib/fleet.ts
TBonnin: because I want the db schema to reflect the name/id of the fleet. I could set the schema for the DBClient and then pass the client but it is more encapsulated this way

packages/fleet/lib/models/deployments.ts
TBonnin: very good point. I forgot about rolling back. commitHash is indeed not unique. I will fix

packages/fleet/lib/models/nodes.ts
TBonnin: A result must be returned. I could maybe do `return transition.mapError(...)` but I don't think it is more readable

packages/fleet/lib/models/nodes.ts
TBonnin: you are right. I should try/catch everywhere

packages/fleet/lib/models/deployments.ts
TBonnin: added a `FleetError` in https://github.com/NangoHQ/nango/pull/3105/commits/ee4d0e9f16dbc9b44ab5f58ea537a4526118a897 let me know what you think?

bodinsamuel:
let me know when you want another pass

TBonnin:
> let me know when you want another pass

I have addressed all your comments I think so happy to get another pass

packages/fleet/lib/models/deployments.integration.test.ts
```diff
@@ -0,0 +1,43 @@
+import { expect, describe, it, beforeEach, afterEach } from 'vitest';
+import * as deployments from './deployments.js';
+import { getTestDbClient } from '../db/helpers.test.js';
+import { generateCommitHash } from './helpers.test.js';
+
+describe('Deployments', () => {
+    const dbClient = getTestDbClient('deployments');
+    const db = dbClient.db;
+    beforeEach(async () => {
```
nalanj, line 9: Just in general, I've seen as test suites grow having the entire database clear and rebuild is almost impossible to make performant. One pattern I've usually used is ensuring there are clear techniques for maintaining a consistent schema in the test suite and ensuring that each test assumes the state of the database is pure chaos, so the test must succeed no matter what else is going on.

That's definitely not something we do elsewhere, but it's very hard to unwind counting on a clean schema for each test.

packages/fleet/lib/fleet.integration.test.ts
```diff
@@ -0,0 +1,29 @@
...
+    });
+
+    afterEach(() => {});
+
+    afterAll(async () => {
+        await getTestDbClient(fleetId).clearDatabase();
+    });
+
+    describe('deploy', () => {
+        it('should create a new deployment', async () => {
```
nalanj, line 21: I'm not 100% sure about the right way for us to isolate this, and it's certainly a philosophical thing, but this test effectively runs the same code as the other deployment creation test. It might be a nice spot for us to use a test stub knowing that  fleet can trust that deploy is well tested.

packages/fleet/lib/fleet.ts
nalanj: Are you thinking we would have multiple fleet schemas in production?

packages/fleet/lib/models/deployments.ts
nalanj: I'm not sure where rolling back is headed, so we may already align, but I think it will be clearer if we're just allowing duplicate commitHash entries with different ids, so that we have a linear history. So a rollback would just be us deploying an old version again, rather than removing anything.

nalanj (approved):
Pretty cool to see it taking shape. None of the things I commented on were things that I feel too too strongly about, more just asking questions and pondering.

packages/fleet/lib/fleet.integration.test.ts
TBonnin: You are right. On the other hand stub everywhere could introduce more issues down the line when underlying logic is changed. 

I've added this test to bootstrap the fleet.integration.test file but I might remove it since it is not very useful

packages/fleet/lib/fleet.ts
TBonnin: not planning to have multiple fleet anytime soon no but nothing would prevent it in the code

packages/fleet/lib/models/deployments.integration.test.ts
TBonnin: I like to follow the `Make it work, make it right, make it fast` principle. I know we are almost at the stage where `make it fast` matters but I have too often fought with state management in tests that were trying to introduce "clever" things, including race conditions and flaky tests.

packages/fleet/lib/models/deployments.ts
TBonnin: that's exactly what's this is doing. Deployments are immutable and rolling back means inserting a new entry with the same commitHash. No update, no deletion

packages/fleet/lib/models/deployments.integration.test.ts
nalanj: I would typically say as-is our tests are written to be clever by assuming the environment is clean, which is almost never the case in production, but to each their own.

packages/fleet/lib/db/migrations/20241102115903_initial_fleet_models.ts
TBonnin: I now remember why I didn't make it unique. Because any node can end up in ERROR state and therefore several entry can have the same routingId/deploymentId. 
Ex:
1. Node is created as PENDING
2. Request to Render to start the new runner fails for some reason. Node is set to ERROR
3. A new node with same routingID and deploymentId is created as PENDING

Another example: if runner is idled because of inactivity, another runner might potentially be created with the same routingId/deployment

packages/fleet/lib/models/deployments.integration.test.ts
TBonnin: I will make this tests to initialize/clean things only once with beforeAll/afterAll if I can in next PR

packages/fleet/lib/db/migrations/20241102115903_initial_fleet_models.ts
bodinsamuel: I see, could be nice to have this clarified somewhere in the models 👍🏻

packages/fleet/lib/db/migrations/20241102115903_initial_fleet_models.ts
```diff
@@ -0,0 +1,58 @@
+import type { Knex } from 'knex';
+import { NODES_TABLE } from '../../models/nodes.js';
+import { DEPLOYMENTS_TABLE } from '../../models/deployments.js';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.transaction(async (trx) => {
+        await trx.raw(`
+            CREATE TABLE ${DEPLOYMENTS_TABLE} (
+                id SERIAL PRIMARY KEY,
+                commit_id char(40) NOT NULL,
```
bodinsamuel, line 10: commit_id is the primary key no? or just unique for simplicity sake

packages/fleet/lib/utils/errors.ts
```diff
@@ -0,0 +1,23 @@
+type Jsonable = string | number | boolean | null | undefined | readonly Jsonable[] | { readonly [key: string]: Jsonable } | { toJSON(): Jsonable };
```
bodinsamuel: could be type-fest or could be stored globally in types/utils

bodinsamuel (approved):
Looking good 🦖

packages/fleet/lib/utils/errors.ts
TBonnin: didn't want to bring type-fest just for that. Where in types/utils would you put it?

packages/fleet/lib/db/migrations/20241102115903_initial_fleet_models.ts
TBonnin: commit id is not unique anymore so we can rollback and deploy the same commit multiple times

packages/fleet/lib/utils/errors.ts
bodinsamuel: packages/types/lib/utils.ts

packages/fleet/lib/db/migrations/20241102115903_initial_fleet_models.ts
bodinsamuel: ah yes sorry 👌🏻

PR #3108, opened 2024-12-04
fix(linear-retries): [nan-2309] handle linear case

khaliqgant:
Linear sends back a 4xx status code with a `RATELIMITED` string in the body. Currently this doesn't get picked up by our retry logic. Searching using the headers will retry even if it isn't a strict 429.

`NAN-2309`

Test by mimicing a response that is like the linear response
```
{
  "request": {
    "headers": {
      "authorization": "REDACTED",
      "user-agent": "nango-node-client/0.45.1 (linux/6.8.0-1018-aws; node.js/20.12.2); sdk"
    },
    "method": "POST",
    "url": "https://api.linear.app/graphql"
  },
  "response": {
    "headers": {
      "x-ratelimit-complexity-remaining": "971913",
      "x-request-id": "8eca3a6e066a3076-SEA",
      "content-length": "535",
      "vary": "Accept-Encoding",
      "x-complexity": "520",
      "x-ratelimit-complexity-limit": "1000000",
      "x-ratelimit-requests-remaining": "0",
      "x-ratelimit-requests-reset": "1733302464386",
      "x-ratelimit-complexity-reset": "1733302464386",
      "content-type": "application/json; charset=utf-8",
      "connection": "close",
      "x-ratelimit-requests-limit": "1200"
    },
    "code": 400
  },
  "error": {
    "message": "Request failed with status code 400",
    "payload": {
      "safeHeaders": {
        "authorization": "REDACTED",
        "user-agent": "nango-node-client/0.45.1 (linux/6.8.0-1018-aws; node.js/20.12.2); sdk"
      },
      "code": "ERR_BAD_REQUEST",
      "data": {
        "errors": [
          {
            "extensions": {
              "userError": true,
              "code": "RATELIMITED",
              "meta": {},
              "type": "ratelimited",
              "userPresentableMessage": "Rate limit exceeded. Only 1200 requests are allowed per 1 hour. For more information see our developer docs at: https://developers.linear.app/docs/graphql/working-with-the-graphql-api/rate-limiting."
            },
            "message": "Rate limit exceeded. Only 1200 requests are allowed per 1 hour. For more information see our developer docs at: https://developers.linear.app/docs/graphql/working-with-the-graphql-api/rate-limiting"
          }
        ]
      },
      "method": "POST",
      "url": "https://api.linear.app/graphql"
    }
  }
}

```


packages/shared/lib/services/proxy.service.ts
```diff
@@ -298,6 +299,32 @@ class ProxyService {
...
 
+    private isProviderRetryTriggered(retryConfig: RetryHeaderConfig | undefined, response: AxiosError): boolean {
+        if (!retryConfig) {
+            return false;
+        }
+
+        const { status_code, body_contains } = retryConfig;
+
+        const statusCodeValid =
+            status_code && status_code.includes('x')
```
bodinsamuel: really not sure we need that feature, they return a 400 let's support a 400 and see if we need more

scripts/validation/providers/schema.json
```diff
@@ -209,6 +209,12 @@
                                 },
                                 "at": {
                                     "type": "string"
+                                },
+                                "status_code": {
+                                    "type": "string"
```
bodinsamuel, line 214: let's stick to number

bodinsamuel (commented):
If we indeed have headers I would suggest something else
```
"x-complexity": "520",
      "x-ratelimit-complexity-limit": "1000000",
      "x-ratelimit-requests-remaining": "0",
      "x-ratelimit-requests-reset": "1733302464386",
      "x-ratelimit-complexity-reset": "1733302464386",
```

We should change the if in `retry` to search for headers whatever the status code

packages/shared/lib/services/proxy.service.ts
khaliqgant: Specifically added this because Linear's docs say:

> Once you actually exceed rate limits, Linear API will start returning rate limit error responses. You can catch these by inspecting response http status code to be 4xx (>=400 and <500) and response body containing RATELIMITED error code.

https://developers.linear.app/docs/graphql/working-with-the-graphql-api/rate-limiting

khaliqgant:
> If we indeed have headers I would suggest something else
> 
> ```
> "x-complexity": "520",
>       "x-ratelimit-complexity-limit": "1000000",
>       "x-ratelimit-requests-remaining": "0",
>       "x-ratelimit-requests-reset": "1733302464386",
>       "x-ratelimit-complexity-reset": "1733302464386",
> ```
> 
> We should change the if in `retry` to search for headers whatever the status code

Thanks, agreed + less code. Updated with https://github.com/NangoHQ/nango/pull/3108/commits/47becfa641d65796cbe6830c5f6291d76953a573

packages/shared/lib/services/proxy.service.ts
```diff
@@ -298,6 +299,24 @@ class ProxyService {
...
 
+    private isRetryInHeader(retryConfig: RetryHeaderConfig | undefined, error: AxiosError): boolean {
+        if (!retryConfig) {
+            return false;
+        }
+
+        const { at, after } = retryConfig;
+
+        if (at) {
+            return Boolean(error.response?.headers[at] || error.response?.headers[at.toLowerCase()]);
```
bodinsamuel: for later: maybe we should just enforce that headers in providers.yaml are lowercase since headers are case-insensitive

packages/shared/lib/services/proxy.service.ts
```diff
@@ -259,7 +259,8 @@ class ProxyService {
             (error.response?.status === 403 && error.response.headers['x-ratelimit-remaining'] && error.response.headers['x-ratelimit-remaining'] === '0') ||
             error.response?.status === 429 ||
             ['ECONNRESET', 'ETIMEDOUT', 'ECONNABORTED'].includes(error.code as string) ||
-            config.retryOn?.includes(Number(error.response?.status))
+            config.retryOn?.includes(Number(error.response?.status)) ||
+            this.isRetryInHeader(config.provider.proxy?.retry, error)
```
bodinsamuel: Just one last question (sorry my suggestion also had some shortcomings):
Are the headers always set or just when we reach rate limit? Here if its always set then all requests will be retried even on regular 400

I guess one missing check that does not exist yet in checking a "remaining headers"

packages/shared/lib/services/proxy.service.ts
bodinsamuel: edit: since I'm modifying the validation I'll do that

packages/shared/lib/services/proxy.service.ts
TBonnin: is isRetryInHeader supporting the special case for github 3 lines above?

packages/shared/lib/services/proxy.service.unit.test.ts
```diff
@@ -505,6 +505,97 @@ describe('Proxy service Construct URL Tests', () => {
...
+                statusText: 'Bad Request',
+                config: {} as InternalAxiosRequestConfig
+            } as AxiosResponse
+        } as AxiosError;
+        const config = {
+            provider: {
+                auth_mode: 'OAUTH2',
+                proxy: {
+                    retry: {
+                        at: 'x-ratelimit-requests-resets'
```
TBonnin: I would make it very different from the real header name. It took me a few seconds to spot the extra `s`

packages/shared/lib/services/proxy.service.unit.test.ts
```diff
@@ -505,6 +505,97 @@ describe('Proxy service Construct URL Tests', () => {
         expect(diff).toBeGreaterThan(1000);
         expect(diff).toBeLessThan(2000);
     });
+
+    it('Should retry based on the header even if the error code is not a 429', async () => {
```
TBonnin, line 526: can we wrap those new tests in their own `describe` so you can instanciate the mockAxiosError only once?

packages/shared/lib/services/proxy.service.ts
khaliqgant: > Are the headers always set or just when we reach rate limit? Here if its always set then all requests will be retried even on regular 400

Unsure tbh which is why in my initial implementation kept this isolated to Linear to keep the impact low since I don't know how most APIs behave. 

> I guess one missing check that does not exist yet in checking a "remaining headers"

I don't know what you mean here? Can you clarify?

khaliqgant: > is isRetryInHeader supporting the special case for github 3 lines above?

No, but I guess we could encapsulate both into updated logic. Few proposals:

1. just keep them both hard coded
```
# github
(error.response?.status === 403 && error.response.headers['x-ratelimit-remaining'] && error.response.headers['x-ratelimit-remaining'] === '0') ||

# linear 
(error.response?.status === 400 && error.response.headers['x-ratelimit-requests-remaining'] && error.response.headers['x-ratelimit-requests-remaining'] === '0') ||
```
2. Codify it in the yaml
```
# github
proxy:
        retry:
            at: 'x-ratelimit-reset'
            remaining: x-ratelimit-remaining
            error_code: 403

# linear
proxy:
           retry:
              at: 'X-RateLimit-Requests-Reset'
              remaining: x-ratelimit-requests-remaining
              error_code: 400
```

any thoughts @NangoHQ/engineers?

TBonnin: the second one looks cleaner but it make providers.yaml more complex. I don't feel strongly either way

bodinsamuel: Option 2 seems correct and not too hard to implement, I would love it 👍🏻

packages/shared/lib/services/proxy.service.unit.test.ts
khaliqgant: Updated

packages/shared/lib/services/proxy.service.unit.test.ts
khaliqgant: Updated

packages/shared/lib/services/proxy.service.ts
khaliqgant: Implemented option 2 with https://github.com/NangoHQ/nango/pull/3108/commits/b94e844d6144e9330c71373c28f4ffdf4feb7a08

bodinsamuel (approved):
Super clean 👌🏻

PR #3110, opened 2024-12-04
fix(providers): enforce lowercase headers, enforce categories

bodinsamuel:
## Changes

- enforce lowercase headers
Headers are case insensitive, but that's annoying when we manipulate those strings against axios/fetch that always lowercase headers
- enforce categories name


TBonnin (approved):
fine by me. Are we sure no logic relies on the case though?

bodinsamuel:
> fine by me. Are we sure no logic relies on the case though?

It's in reaction of #3108 this is the only place those headers are used (and proxy obv)

nalanj (approved):
🙌🏻 I bumped into header casing on this testing PR I've been working on.

khaliqgant (approved):

PR #3112, opened 2024-12-04
docs(sdk): add getIntegration

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2298/document-getintegration-in-sdk

- Document getIntegration in the sdk

### DIFF
--- a/docs-v2/reference/scripts.mdx
+++ b/docs-v2/reference/scripts.mdx
@@ -202,6 +202,28 @@ await nango.triggerAction('<ACTION-NAME>', { 'custom_key1': 'custom_value1' });
 
 _TbD_
 
+# Get Integration
+
+Returns the current integration information
+
+```js
+await nango.getIntegration();
+```
+
+With credentials
+
+```js
+await nango.getIntegration({ include: ['credentials'] });
+```
+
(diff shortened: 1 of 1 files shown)


docs-v2/reference/scripts.mdx
```diff
@@ -202,6 +202,28 @@ await nango.triggerAction('<ACTION-NAME>', { 'custom_key1': 'custom_value1' });
...
+
+With credentials
+
+```js
+await nango.getIntegration({ include: ['credentials'] });
+```
+
+**Parameters**
+
+See `GET /integrations/{uniqueKey}` query parameters: [documentation](/reference/api/integration/get)
```
TBonnin, line 221: Do we use this `see API ENDPOINT doc page ...` pattern for any other method? Should we?

TBonnin (approved):

docs-v2/reference/scripts.mdx
bodinsamuel: I don't think so, and I don't know.
Honestly, those docs are annoying to maintain and got lazy but it's also the first one where the api and the method matches perfectly

PR #3115, opened 2024-12-04
fix(tests): Fix broken integration test that never fails in CI

nalanj:
This was a weird one. Since I've joined I've always had some tests that fail on my computer but never fail in CI. There's another that's flaky for me, but I can't seem to get it to fail right now.

This test was failing for me:

https://github.com/NangoHQ/nango/blob/053d636f2f8d616f2ea32d343830367dcbf34f4c/packages/server/lib/controllers/connect/postSessions.integration.test.ts#L101-L139

The source of the issue, I found, was that we were replying in the server inside of a transaction block, so the transaction hadn't finalized by the time that my request replied and the test checked that the database was updated. This PR pulls the http response sending outside of the transaction block to make sure everything times out correctly.

I couldn't find any other examples where we did this.

## How I tested it

Run the tests!


TBonnin (approved):
lgtm. This is probably not the only place in the codebase where we send response back from inside a db transaction

nalanj:
@TBonnin I actually dug around some and couldn't find any other instances, but that was also my assumption.

bodinsamuel (approved):
Indeed

PR #3116, opened 2024-12-05
feat(fleet): add control loop and state management

TBonnin:
- Implement Supervisor class to manage node states and transitions
- Add methods to start, fail, outdate, terminate, and remove nodes
- Introduce NodeProvider interface for node operations
- Update node state transitions and error handling
- Add integration tests for Supervisor functionality

What's not yet implemented:
- routing logic
- Render NodeProvider
- wiring fleet in jobs

There are still a few TODOs that I will address in following PRs

# How to test
Still not wired. You can run the tests ;)

### DIFF
--- a/packages/fleet/lib/models/nodes.integration.test.ts
+++ b/packages/fleet/lib/models/nodes.integration.test.ts
@@ -54,7 +54,7 @@ describe('Nodes', () => {
         const routingId = 'my-routing-id';
         await nodes.create(db, {
             routingId,
-            deploymentId: activeDeployment,
+            deploymentId: activeDeployment.id,
             url: 'http://localhost:3000',
             image: 'nangohq/my-image:latest',
             cpuMilli: 500,
@@ -63,7 +63,7 @@ describe('Nodes', () => {
         });
         const result = await nodes.create(db, {
             routingId,
-            deploymentId: activeDeployment,
+            deploymentId: activeDeployment.id,
             url: 'http://localhost:3000',
             image: 'nangohq/my-image:latest',
             cpuMilli: 500,
--- a/packages/fleet/lib/models/nodes.integration.test.ts
+++ b/packages/fleet/lib/models/nodes.integration.test.ts
@@ -50,28 +50,6 @@ describe('Nodes', () => {
             lastStateTransitionAt: expect.any(Date)
         });
     });
-    it('should fail to create a node with the same routingId and deploymentId', async () => {
-        const routingId = 'my-routing-id';
-        await nodes.create(db, {
-            routingId,
-            deploymentId: activeDeployment.id,
-            url: 'http://localhost:3000',
-            image: 'nangohq/my-image:latest',
-            cpuMilli: 500,
-            memoryMb: 1024,
-            storageMb: 512
-        });
-        const result = await nodes.create(db, {
-            routingId,
-            deploymentId: activeDeployment.id,
-            url: 'http://localhost:3000',
-            image: 'nangohq/my-image:latest',
-            cpuMilli: 500,
-            memoryMb: 1024,
-            storageMb: 512
-        });
-        expect(result.isErr()).toBe(true);
-    });
     it('should transition between valid states and error when transitioning between invalid states', async () => {
         for (const from of nodeStates) {
             for (const to of nodeStates) {
--- a/packages/fleet/lib/models/nodes.integration.test.ts
+++ b/packages/fleet/lib/models/nodes.integration.test.ts
@@ -2,15 +2,14 @@ import { expect, describe, it, beforeEach, afterEach } from 'vitest';
 import * as nodes from './nodes.js';
 import * as deployments from './deployments.js';
 import { nodeStates } from '../types.js';
-import type { NodeState, Node, RoutingId, Deployment } from '../types.js';
+import type { NodeState, Deployment } from '../types.js';
 import { getTestDbClient } from '../db/helpers.test.js';
-import type { knex } from 'knex';
-import { nanoid } from '@nangohq/utils';
-import { generateCommitHash } from './helpers.test.js';
+import { generateCommitHash, createNodeWithAttributes } from './helpers.test.js';
 
 describe('Nodes', () => {
     const dbClient = getTestDbClient('nodes');
     const db = dbClient.db;
+
     let previousDeployment: Deployment;
     let activeDeployment: Deployment;
     beforeEach(async () => {
@@ -28,7 +27,6 @@ describe('Nodes', () => {
             await nodes.create(db, {
                 routingId: 'my-routing-id',
                 deploymentId: activeDeployment.id,
-                url: 'http://localhost:3000',
                 image: 'nangohq/my-image:latest',
                 cpuMilli: 500,
                 memoryMb: 1024,
@@ -39,7 +37,7 @@ describe('Nodes', () => {
             id: expect.any(Number),
             routingId: 'my-routing-id',
             deploymentId: activeDeployment.id,
-            url: 'http://localhost:3000',
+            url: null,
             state: 'PENDING',
             image: 'nangohq/my-image:latest',
             cpuMilli: 500,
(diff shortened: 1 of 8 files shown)


packages/fleet/lib/supervisor.ts
```diff
@@ -0,0 +1,278 @@
...
+import * as nodes from './models/nodes.js';
+import * as deployments from './models/deployments.js';
+import { Err, Ok } from '@nangohq/utils';
+import type { Result } from '@nangohq/utils';
+import { FleetError } from './utils/errors.js';
+import type { Deployment, Node } from './types.js';
+import { setTimeout } from 'node:timers/promises';
+import type { NodeProvider } from './node-providers/node_provider.js';
+
+type Action =
```
TBonnin: I know Action is an overloaded term in Nango but I couldn't think of a better name. Suggestions are welcomed

packages/fleet/lib/supervisor.ts
```diff
@@ -0,0 +1,278 @@
...
+
+export class Supervisor {
+    private state: SupervisorState = 'stopped';
+    private dbClient: DatabaseClient;
+    private nodeProvider: NodeProvider;
+    public STATE_TIMEOUT_MS = {
+        STARTING: 5 * 60 * 1000,
+        FINISHING: 24 * 60 * 60 * 1000,
+        TERMINATED: 7 * 24 * 60 * 60 * 1000,
+        ERROR: 7 * 24 * 60 * 60 * 1000
```
TBonnin: arbitrary values. Happy to change

packages/fleet/lib/models/nodes.ts
```diff
@@ -215,20 +250,53 @@ export async function fail(
...
+        // only TERMINATED and ERROR nodes can be deleted
+        const getNode = await get(db, props.nodeId);
+        if (getNode.isErr()) {
+            return getNode;
+        }
+        if (!['TERMINATED', 'ERROR'].includes(getNode.value.state)) {
+            return Err(new FleetError(`node_delete_non_terminated`, { context: { nodeId: props.nodeId } }));
+        }
+
+        const deleted = await db.from<DBNode>(NODES_TABLE).where('id', props.nodeId).del().returning('*');
```
bodinsamuel, line 294: there is no need for returning, just checking row_count should be enough no?

packages/fleet/lib/models/nodes.ts
```diff
@@ -215,20 +250,53 @@ export async function fail(
...
+        if (!deleted?.[0]) {
+            return Err(new FleetError(`node_delete_error`, { context: props }));
         }
-        return Ok(DBNode.from(updated[0]));
+        return Ok(DBNode.from(deleted[0]));
```
bodinsamuel, line 298: Not sure why you would want to return a deleted row

packages/fleet/lib/supervisor.integration.test.ts
```diff
@@ -0,0 +1,165 @@
...
+    it('should not have more than one instance processing at a time', async () => {
+        const supervisor1 = new Supervisor({ dbClient, nodeProvider: mockNodeProvider });
+        const supervisor2 = new Supervisor({ dbClient, nodeProvider: mockNodeProvider });
+        supervisor1.start();
+        supervisor2.start();
+        await new Promise((resolve) => setTimeout(resolve, 100));
+        expect(true).toBe(true);
+        await supervisor1.stop();
+        await supervisor2.stop();
+    });
```
bodinsamuel: Not entirely what this does; that stuff does not throw?

packages/fleet/lib/supervisor.ts
bodinsamuel: Operation is more accurate I feel, but still very generic

packages/fleet/lib/supervisor.ts
```diff
@@ -0,0 +1,278 @@
...
+    | { type: 'REMOVE'; node: Node }
+    | { type: 'WARN'; node: Node };
+
+type SupervisorState = 'stopped' | 'running' | 'stopping';
+
+export class Supervisor {
+    private state: SupervisorState = 'stopped';
+    private dbClient: DatabaseClient;
+    private nodeProvider: NodeProvider;
+    public STATE_TIMEOUT_MS = {
```
bodinsamuel: Not sure it belongs to the class, could a readonly const

packages/fleet/lib/models/nodes.integration.test.ts
```diff
@@ -125,37 +132,22 @@ describe('Nodes', () => {
...
-        cpuMilli: 500,
-        memoryMb: 1024,
-        storageMb: 512
+    it('should be able to fail a node', async () => {
+        const node = await createNodeWithAttributes(db, { state: 'PENDING', deploymentId: activeDeployment.id });
+        const failedNode = (await nodes.fail(db, { nodeId: node.id, reason: 'my error' })).unwrap();
+        expect(failedNode.state).toBe('ERROR');
+        expect(failedNode.error).toBe('my error');
+    });
+    it('should be able to register a node', async () => {
```
bodinsamuel, line 146: I'm not sure what you have against line break 🤣

packages/fleet/lib/supervisor.ts
```diff
@@ -0,0 +1,278 @@
...
+        const plan: Action[] = [];
+
+        const search = await nodes.search(this.dbClient.db, {
+            states: ['PENDING', 'STARTING', 'RUNNING', 'OUTDATED', 'FINISHING', 'IDLE', 'TERMINATED', 'ERROR'],
+            ...(cursor ? { cursor } : {})
+        });
+        if (search.isErr()) {
+            return Err(search.error);
+        }
+        for (const [routingId, nodes] of search.value.nodes) {
```
bodinsamuel, line 104: this part is confusing to me, why would group nodes by `routingId`, what's the use? 
Not grouping would make the loop look more regular, e.g: `if (starting) else if (error) etc.`

packages/fleet/lib/supervisor.ts
```diff
@@ -0,0 +1,278 @@
...
+            // Remove old error nodes
+            plan.push(
+                ...(nodes.ERROR || []).flatMap((node) => {
+                    if (Date.now() - node.lastStateTransitionAt.getTime() > this.STATE_TIMEOUT_MS.ERROR) {
+                        return [{ type: 'REMOVE' as const, node }];
+                    }
+                    return [];
+                })
+            );
+        }
```
bodinsamuel, line 165: This plan strategy assume the supervisor is on top of everything, but machine could go down or up during this plan. Is there a plan for runners to reach out to the supervisor?

bodinsamuel (approved):
Not going to lie it's a lot of code and a lot of "hypothetical" code, not everything makes sense to me rn. But looks okay overall

packages/fleet/lib/supervisor.ts
```diff
@@ -0,0 +1,278 @@
...
+        for (const [routingId, nodes] of search.value.nodes) {
+            // Start pending nodes
+            plan.push(...(nodes.PENDING || []).map((node) => ({ type: 'START' as const, node })));
+
+            // Timeout STARTING nodes if they are taking too long
+            plan.push(
+                ...(nodes.STARTING || []).flatMap((node) => {
+                    if (Date.now() - node.lastStateTransitionAt.getTime() > this.STATE_TIMEOUT_MS.STARTING) {
+                        return [{ type: 'FAIL' as const, node, reason: 'starting_timeout_reached' as const }];
```
bodinsamuel: you can remove the const if you cast `flatMap<Action>`

packages/fleet/lib/models/nodes.integration.test.ts
TBonnin: do you mean empty line between the tests? :)

packages/fleet/lib/models/nodes.ts
TBonnin: the idea was to have every function to return a `Result<Node>`. I agree it is not necessary for deletion but it is more for consistency

packages/fleet/lib/supervisor.integration.test.ts
TBonnin: I was just preparing the test for my next PR. That should not be in there. Removing.

packages/fleet/lib/supervisor.ts
TBonnin: good idea. changing for operation

packages/fleet/lib/supervisor.ts
TBonnin: because the decision to create a new node if there is no RUNNING or upcoming ones is per `routingId`. 
True that the rest of the decisions are not dependent on the `routingId` though.

packages/fleet/lib/supervisor.ts
TBonnin: runner would `register` (via jobs) when they start up to say they are online and ready.

packages/fleet/lib/models/nodes.integration.test.ts
bodinsamuel: yes even in general, between logical code blocks but I guess it's keyboard-driven development :p

bodinsamuel (approved):

packages/fleet/lib/supervisor.ts
```diff
@@ -0,0 +1,279 @@
...
+
+    private async loop(): Promise<void> {
+        while (this.state === 'running') {
+            // TODO: Lock to prevent multiple instances from running
+            await this.tick();
+        }
+        this.state = 'stopped';
+    }
+
+    private async plan(cursor?: number): Promise<Result<Operation[]>> {
```
nalanj, line 86: Would it be easier going forward for operations to be an object with a type and a function attached, and then executing the plan just loops and calls the functions?

TBonnin: why not. The advantage of having values instead of opaque functions is that you can optimize the plan before to execute it. I don't think Render API can do it and even if it does I am not planning to support it but we could for instance takes all the CREATE operations fromthe plan and combine them in a unique call to the node provider instead of one by one.

PR #3118, opened 2024-12-05
feat(docs): New docs format

nalanj:
Updates our docs for integrations to be in the new format described in [the ticket](https://linear.app/nango/issue/NAN-2320/backfill-api-specific-docs
).

We're leaning on Mintlify snippets for the always generated portions of the document. See `scripts/docs-gen-snippets.mjs` for that.

https://linear.app/nango/issue/NAN-2320/backfill-api-specific-docs

## How I tested it
- Run `npm run docs`
- Click through the updated documents and make sure everything looks right


bastienbeurier:
CleanShot 2024-12-12 at 11 05 41@2x
This callout should always appear in the `Pre-built use-cases` section, not only in the empty case.

bastienbeurier:
Checking something: the sections `Pre-built tooling` and `Access requirements` will be edited manually passed the initial generation. I want to make sure they won't be overridden by the following generations. IMO only the `pre-built use-cases` should be repeatedly generated.

bastienbeurier:
CleanShot 2024-12-12 at 11 10 15@2x
Nit and optional: instead of showing `github.com` as a link label, can we have something prettier & less prominent? E.g. a github icon, an emoji.

scripts/docs-gen-snippets.mjs
```diff
@@ -0,0 +1,202 @@
...
+                .join('\n')}
+        </AccordionGroup>
+    `
+        .split('\n')
+        .map((line) => line.trim())
+        .join('\n');
+}
+
+function emptyUseCases() {
+    return `## Pre-built use-cases
```
bodinsamuel: Why is it called use-cases and not just endpoints? That would resonate with the dashboard API ref since it's presenting the same data, cc @bastienbeurier

scripts/docs-gen-snippets.mjs
```diff
@@ -0,0 +1,202 @@
...
+        const authMode = providers[provider].auth_mode;
+        const prettyAuthMode = prettyAuthModes[authMode] || authMode;
+        const hasUseCases = useCases[provider] && useCases[provider].length > 0;
+        const toolingSnippet = preBuiltToolingSnippet({ prettyAuthMode, hasUseCases });
+
+        await fs.mkdir(`${snippetsPath}/${path.basename(file, '.mdx')}`, { recursive: true });
+        await fs.writeFile(`${snippetsPath}/${path.basename(file, '.mdx')}/PreBuiltTooling.mdx`, toolingSnippet, 'utf-8');
+
+        const casesSnippet = useCasesSnippet({ useCases: useCases[provider] });
+        await fs.writeFile(`${snippetsPath}/${path.basename(file, '.mdx')}/PreBuiltUseCases.mdx`, casesSnippet, 'utf-8');
```
bodinsamuel: All the repeated path could be stored `const base = ${snippetsPath}/${path.basename(file, '.mdx')}`

scripts/docs-gen-snippets.mjs
```diff
@@ -0,0 +1,202 @@
+import fs from 'node:fs/promises';
```
bodinsamuel: Could have been a ts file

docs-v2/snippets/pre-built-tooling.mdx
```diff
@@ -0,0 +1,168 @@
...
+        OAUTH2_CC: "OAuth",
+        BASIC: "Basic",
+        API_KEY: "API Key",
+        APP_STORE: "Custom",
+        BILL: "Bill",
+        SIGNATURE: "Signature",
+        JWT: "JWT",
+        TWO_STEP: "Two Step",
+        TABLEAU: "Tableau",
+    }
```
bodinsamuel: If you generate all the mdx why not use this value this is also defined in the script?

docs-v2/snippets/providers.mdx
```diff
@@ -0,0 +1,1410 @@
+export const providers = {
```
bodinsamuel: Not sure I get why this file exists?

docs-v2/snippets/generated/zoominfo/PreBuiltUseCases.mdx
```diff
@@ -0,0 +1,5 @@
+## Pre-built use-cases
+
+_No pre-built use-cases yet (time to contribute: &lt;48h)_
+
+<Tip>Not seeing the use-case you need? [Build your own](https://nango.dev/slack) independently.</Tip>
```
bodinsamuel: You could remove 239 files from the repo by reusing a component when it's empty or creating a symlink

docs-v2/integrations/all/accelo.mdx
```diff
@@ -1,29 +1,43 @@
...
-| [Sync data](/guides/getting-started/read-from-an-api) | ✅ |
-| [Perform workflows](/guides/getting-started/write-to-an-api) | ✅ |
-| [Proxy requests](/guides/proxy-requests-to-an-api) | ✅ |
-| [Receive webhooks](/guides/receive-webhooks-from-an-api) | 🚫 (time to contribute: &lt;48h) |
+<Overview />
+<PreBuiltTooling />
+<PreBuiltUseCases />
 
-<Tip>We can implement missing features in &lt;48h, just ask for it in the [community](https://nango.dev/slack).</Tip>
+## Access requirements
```
bodinsamuel, line 60: I feel like this section is really not clear. I thought it was "Accessing this integration on Nango" and it took me while to get that it's "can you create an app in the third party?" 

Not sure how to improve, maybe some introduction text, removing the confusing `?`, changing the label names or removing the section altogether 😅

docs-v2/integrations/all/accelo.mdx
```diff
@@ -1,29 +1,43 @@
...
+| Pre-Requisites | Status | Comment|
+| - | - | - |
+| Paid dev account | ❓ |  |
+| Paid test account | ❓ |  |
+| Partnership | ❓ | |
+| App review | ❓ |  |
+| Security audit | ❓ | |
```
bodinsamuel, line 68: I think like the readmes, it will be much easier to split the file in two and always regenerate the top.
Otherwise, if you/we need to change something at scale it will be painful

bodinsamuel (commented):
Looks good overall but lots of open questions.
Also made my browser crash multiple times 😮‍💨

bodinsamuel:
> Nit and optional: instead of showing github.com as a link label, can we have something prettier & less prominent? E.g. a github icon, an emoji.

Emojis are not accessible, it's really not clear that it's a button and where it will lead you (especially on mobile)

nalanj:
> This callout should always appear in the Pre-built use-cases section, not only in the empty case.

@bastienbeurier Fixed in 6fbb3f0e

docs-v2/snippets/pre-built-tooling.mdx
nalanj: Oops, this is a leftover file from when I was trying to reuse the component. Unfortunately Mintlify's build system couldn't handle much complexity in a snippet because it inlines everything in the build process, so there's no way to have a snippet import our data efficiently.

nalanj:
> Checking something: the sections Pre-built tooling and Access requirements will be edited manually passed the initial generation. I want to make sure they won't be overridden by the following generations. IMO only the pre-built use-cases should be repeatedly generated.

That is the case for access requirements right now, but I didn't realize it was the case for pre-built tooling. I'll go back through and manually insert those.

docs-v2/snippets/providers.mdx
nalanj: I just deleted it. It was from a prior attempt to get mintlify to use data for generation, which it turns out isn't really feasible since mintlify inlines all imports and then slows to a crawl if your imports are too large.

docs-v2/integrations/all/accelo.mdx
nalanj: The only section that is generated is the overview and the pre-built use-cases section. Unfortunately the titles of the pages don't always align with providers so it'd take a lot more logic to generate the top of the doc.

scripts/docs-gen-snippets.mjs
nalanj: Was trying to follow the pattern that other scripts in this directory were written as js, but I'll go back and move to ts looping over any arrays with basically no type safety.

docs-v2/snippets/generated/zoominfo/PreBuiltUseCases.mdx
nalanj: The problem with doing that without also generating each of the integration docs (which isn't happening, yet, maybe it should be) is that adding use cases in the integrations repo wouldn't then show up in this repo without also updating the markdown file for the integration to change the import from the emtpy use cases import to the per-integration one.

docs-v2/integrations/all/accelo.mdx
nalanj: I'll wait to hear from @bastienbeurier on this one.

scripts/docs-gen-snippets.mjs
nalanj: Fixed in 9c01c8d3. Good catch, thanks!

nalanj:
> Nit and optional: instead of showing github.com as a link label, can we have something prettier & less prominent? E.g. a github icon, an emoji.

@bastienbeurier reverted back to the link emoji in d9f59085, but I think @bodinsamuel has a point about it. Easy enough for me to do whatever we decide on that.

docs-v2/integrations/all/accelo.mdx
bastienbeurier: This section needs to be backfilled manually anyway, so we have time to iterate on it. Currently, it's an empty state, so inherently insufficient.

bastienbeurier:
> > Nit and optional: instead of showing github.com as a link label, can we have something prettier & less prominent? E.g. a github icon, an emoji.
> 
> @bastienbeurier reverted back to the link emoji in [d9f5908](https://github.com/NangoHQ/nango/commit/d9f59085d53f78e75f1e63cd4d134f04416a7cec), but I think @bodinsamuel has a point about it. Easy enough for me to do whatever we decide on that.

I don't have a strong opinion on this. We can easily generate this differently at any point, right?

nalanj:
> I don't have a strong opinion on this. We can easily generate this differently at any point, right?

Absolutely. It's just a quick text edit and then re-generating the docs.

scripts/docs-gen-snippets.mjs
bodinsamuel: I can sense a little frustration, if you think it has no value then fine 👍🏻 
Otherwise, you can use `Provider` type

docs-v2/integrations/all/accelo.mdx
bodinsamuel: maybe it can just be left blank when not relevant? Bad messaging is worse than no messaging imo

docs-v2/integrations/all/accelo.mdx
bodinsamuel: with what bastien said about modifying everything manually, you can dismiss my comment. 
But ultimately we could still generate a big part of it, would requires more component though

docs-v2/snippets/generated/zoominfo/PreBuiltUseCases.mdx
bodinsamuel: No sure I get your message. But I guess since it's currently a "one time" generator you can discard

nalanj:
@bastienbeurier f7fb697f pulls Pre-Built tooling back into each doc instead of being a snippet, so that it's maintained by the team instead of automatically.

scripts/docs-gen-snippets.mjs
nalanj: I'm just annoyed at myself for seeing js files and adding another one. I'll go back and port it to ts.

scripts/docs-gen-snippets.mjs
nalanj: Confirmed with @bastienbeurier (https://nangohq.slack.com/archives/C067NRBH1LL/p1734018441493369) that we'll ship now as-is and change later if needed.

bodinsamuel (approved):

TBonnin (approved):

PR #3119, opened 2024-12-05
feat: reconnect

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2302/connect-refuse-connectionid-when-auth-is-session-token
Fixes https://linear.app/nango/issue/NAN-2249/connect-reconnect

- Ability to create a session token to reconnect a `connection_id`
After debate with the team, we settled with a new endpoint dedicated to creating a reconnect token.

- Refuse `connection_id` has a query params if a session token is used

- Update all endpoints to add the reconnect ability

- Add SDK, doc, openapi 

## 🧪 How to test?

There is no built-in way to reconnect from the UI so it's a bit manual

- Use your favorite curl/api tool to create a token from 
```sh
curl --request POST \
  --url http://localhost:3003/connect/sessions/reconnect \
  --header 'Authorization: _PRIVATE_KEY_' \
  --data '{
	"connection_id": "test-connection-id",
	"integration_id": "google-mail",
  "end_user": {
    "id": "test_insomnia",
    "email": "samuel@nango.dev"
  },
}'
```
- Access Connect UI directly with the token `http://localhost:3009/?session_token=SESSION_TOKEN`
- Connect
- Check that no connection has been created and the selected `connection_id` has been updated


packages/server/lib/controllers/auth/postApiKey.ts
```diff
@@ -72,10 +72,15 @@ export const postPublicApiKeyAuthorization = asyncWrapper<PostPublicApiKeyAuthor
     const queryString: PostPublicApiKeyAuthorization['Querystring'] = queryStringVal.data;
     const { providerConfigKey }: PostPublicApiKeyAuthorization['Params'] = paramsVal.data;
     const connectionConfig = queryString.params ? getConnectionConfig(queryString.params) : {};
-    const connectionId = queryString.connection_id || connectionService.generateConnectionId();
+    let connectionId = queryString.connection_id || connectionService.generateConnectionId();
     const hmac = 'hmac' in queryString ? queryString.hmac : undefined;
     const isConnectSession = res.locals['authType'] === 'connectSession';
 
+    if (isConnectSession && queryString.connection_id) {
+        errorRestrictConnectionId(res);
```
bodinsamuel, line 80: This error code was repeated N times, I just figured it would be cleaner like this but I'm not super happy

packages/server/lib/controllers/connect/postSessions.ts
```diff
@@ -64,95 +67,37 @@ export const postConnectSessions = asyncWrapper<PostConnectSessions>(async (req,
...
     const { status, response }: Reply = await db.knex.transaction(async (trx) => {
-        // Check if the endUser exists in the database
-        const endUserRes = await getEndUser(trx, {
-            endUserId: body.end_user.id,
-            accountId: account.id,
-            environmentId: environment.id
-        });
-
-        let endUserInternalId: number;
+        const endUserRes = await upsertEndUser(trx, { account, environment, endUserPayload: body.end_user, organization: body.organization });
```
bodinsamuel, line 70: Put this logic in a function to reuse in postReconnect

docs-v2/reference/sdks/node.mdx
```diff
@@ -1255,6 +1255,87 @@ const { data } = await nango.createConnectSession({
 </Expandable>
 
 
+### Create a reconnect session
+
+Create a reconnect session for a given `connection_id`. Optionally, you can set `end_user` and `organization` to update those information.
```
nalanj:
```suggestion
Create a reconnect session for a given `connection_id`. Optionally, you can set `end_user` and `organization` to update that information.
```

nalanj: Or maybe:

```suggestion
Create a reconnect session for a given `connection_id`. Optionally, you can set `end_user` and `organization` to update those attributes of the connection.
```

packages/connect-ui/src/lib/store.ts
```diff
@@ -30,7 +30,7 @@ export const useGlobal = create<State>((set) => ({
     isSingleIntegration: false,
     session: null,
     nango: null,
-    apiURL: 'https://api.nango.dev',
+    apiURL: 'http://localhost:3003/',
```
nalanj: Just confirming this was intended

packages/frontend/lib/index.ts
```diff
@@ -258,6 +258,14 @@ export default class Nango {
         });
     }
 
+    public reconnect(providerConfigKey: string, options?: AuthOptions): Promise<AuthResult> {
+        if (!this.connectSessionToken) {
+            throw new AuthError('Reconnect is only possible with a session token', 'missing_connect_session_token');
```
nalanj:
```suggestion
            throw new AuthError('Reconnect requires a session token', 'missing_connect_session_token');
```

packages/server/lib/controllers/connect/postReconnect.ts
```diff
@@ -0,0 +1,121 @@
...
+            entityId: createConnectSession.value.id,
+            ttlInMs: 30 * 60 * 1000 // 30 minutes
+        });
+        if (createPrivateKey.isErr()) {
+            res.status(500).send({ error: { code: 'server_error', message: 'Failed to create session token' } });
+            return;
+        }
+
+        const [token, privateKey] = createPrivateKey.value;
+        res.status(201).send({ data: { token, expires_at: privateKey.expiresAt!.toISOString() } });
```
nalanj: This response will be sent before the transaction has been finalized, like that issue I PRed a fix for last week on another controller.

nalanj (approved):
Worked well. 

Left a few comments, and only the transaction related one is one I think definitely needs to be cleaned up.

docs-v2/reference/sdks/node.mdx
```diff
@@ -1255,6 +1255,87 @@ const { data } = await nango.createConnectSession({
 </Expandable>
 
 
+### Create a reconnect session
```
TBonnin, line 1258: maybe we should explain what a reconnect session means or when it should be used? wdyt?

packages/server/lib/controllers/auth/postApiKey.ts
TBonnin: at some point we should stop for a minute and reassess all the auth functions and see if we could consolidate some of the shared logic.
we also repeat the reconnect mechanism below so not sure trying to avoid the repetition here and not below is necessary

packages/shared/lib/services/endUser.service.ts
```diff
@@ -171,3 +171,76 @@ export async function getEndUserByConnectionId(db: Knex, props: { connectionId:
...
+export async function upsertEndUser(
+    db: Knex,
+    {
+        account,
+        environment,
+        endUserPayload,
+        organization
+    }: { account: DBTeam; environment: DBEnvironment; endUserPayload: ConnectSessionInput['end_user']; organization?: ConnectSessionInput['organization'] }
+): Promise<Result<EndUser, EndUserError>> {
+    // Check if the endUser exists in the database
```
TBonnin, line 184: can we not use pg `INSERT ON CONFLICT` to handle the upsert?

packages/server/lib/controllers/connect/postSessions.ts
```diff
@@ -188,3 +133,21 @@ export const postConnectSessions = asyncWrapper<PostConnectSessions>(async (req,
...
+ */
+export function checkIntegrationsDefault(body: Pick<PostConnectSessions['Body'], 'integrations_config_defaults'>, integrations: Config[]): ZodIssue[] | false {
+    if (!body.integrations_config_defaults) {
+        return false;
+    }
+
+    const errors: ZodIssue[] = [];
+    for (const uniqueKey of Object.keys(body.integrations_config_defaults)) {
+        if (!integrations.find((v) => v.unique_key === uniqueKey)) {
+            errors.push({ path: ['integrations_config_defaults', uniqueKey], code: 'custom', message: 'Integration does not exist' });
```
TBonnin, line 148: can  we not return false as soon as we encounter an error?

TBonnin (approved):

docs-v2/reference/sdks/node.mdx
bodinsamuel: yeah might be clearer like this

packages/connect-ui/src/lib/store.ts
bodinsamuel: yes! (absolutely not, thanks)

packages/server/lib/controllers/connect/postReconnect.ts
bodinsamuel: 👍🏻 I yes I started before so I copied the code before the fix, thanks

packages/server/lib/controllers/connect/postSessions.ts
bodinsamuel: I'm team fail exhaustive instead of fail fast, like zod actually

docs-v2/reference/sdks/node.mdx
bodinsamuel: Yes I can add a comment 👌🏻

packages/shared/lib/services/endUser.service.ts
bodinsamuel: I just copied your code actually, we can note to improve that for sure

PR #3120, opened 2024-12-05
fix(api): move more auth endpoints to new format

bodinsamuel:
## Changes

- Move `POST /api-auth/api-key/:providerConfigKey` to new format
- Move `POST /api-auth/basic/:providerConfigKey` to new format
- Move `POST /app-store-auth/:providerConfigKey` to new format


packages/server/lib/controllers/auth/postApiKey.ts
```diff
@@ -0,0 +1,220 @@
...
+        const provider = getProvider(config.provider);
+        if (!provider) {
+            await logCtx.error('Unknown provider');
+            await logCtx.failed();
+            res.status(404).send({ error: { code: 'unknown_provider_template' } });
+            return;
+        }
+
+        if (provider.auth_mode !== 'API_KEY') {
+            await logCtx.error('Provider does not support API key auth', { provider: config.provider });
```
TBonnin, line 115: not from this PR but is it me or the message doesn't match the condition?

packages/server/lib/controllers/auth/postBasic.ts
```diff
@@ -0,0 +1,222 @@
...
+import { hmacCheck } from '../../utils/hmac.js';
+import { connectionCreated as connectionCreatedHook, connectionCreationFailed as connectionCreationFailedHook, connectionTest } from '../../hooks/hooks.js';
+import { connectionCredential, connectionIdSchema, providerConfigKeySchema } from '../../helpers/validation.js';
+import db from '@nangohq/database';
+import { isIntegrationAllowed } from '../../utils/auth.js';
+
+const bodyValidation = z
+    .object({
+        username: z.string(), // no .min() because some basic auth are wrong
+        password: z.string() // no .min() because some basic auth are wrong
```
TBonnin: can we be more explicit about wrong means. Are we talking about some provider using basic with an apikey and empty password value?

packages/server/lib/controllers/apiAuth.controller.ts
```diff
@@ -1,382 +0,0 @@
...
-                }
-            }
-
-            const config = await configService.getProviderConfig(providerConfigKey, environment.id);
-
-            if (config == null) {
-                await logCtx.error('Unknown provider config');
-                await logCtx.failed();
-
-                errorManager.errRes(res, 'unknown_provider_config');
```
nalanj, line 72: Just for understanding sake - is this a pattern we're moving away from?

packages/server/lib/controllers/auth/postApiKey.ts
```diff
@@ -0,0 +1,220 @@
...
+            await logCtx.failed();
+            res.status(400).send({ error: { code: 'invalid_auth_mode' } });
+            return;
+        }
+
+        if (!(await isIntegrationAllowed({ config, res, logCtx }))) {
+            return;
```
nalanj, line 122: Do we have a default response when there's no response set in our handler?

TBonnin (approved):

nalanj (approved):
Just left a couple of comments that were more for my understanding than for feedback.

Looks good overall. Should we have some basic integration tests for these endpoints?

packages/server/lib/controllers/auth/postBasic.ts
bodinsamuel: yeah, wrong in the sense that don't require password or username, I'll add a comment

packages/server/lib/controllers/auth/postApiKey.ts
bodinsamuel: the response is done in the method itself, honestly not ideal but handy for repeating code like this

packages/server/lib/controllers/apiAuth.controller.ts
bodinsamuel: The full context: 
The are multiple reasons for us to move to a different controller format. 
There was no input validation or typing. We could still use the old files but it was also a good opportunity to split files that were super long and hard to maintain. The API grew organically super rapidly so understandable but it's easier like this.
And for this particular line, errorManager is doing too much and not very nicely for a "stable public API" context, at some point hopefully we can get rid of it entirely.

packages/server/lib/controllers/auth/postApiKey.ts
bodinsamuel: mmh no it's correct, we get a provider based on `:providerConfigKey` in the URL. If it's a provider that does not match the auth mode then we raise an error. The message could be "Invalid provider for this endpoint" or we could have one generic endpoint too :D

PR #3124, opened 2024-12-06
feat(integrations): add support for freshteam

hassan254-prog:
Add support for Freshteam
The api key is scoped, thus the reason for omitting the `proxy.verification` field.


packages/shared/providers.yaml
```diff
@@ -2581,6 +2581,37 @@ freshservice:
...
+    docs: https://docs.nango.dev/integrations/all/freshteam
+    docs_connect: https://docs.nango.dev/integrations/all/freshteam/connect
+    connection_config:
+        accountName:
+            type: string
+            title: Account Name
+            description: The account name of your Freshteam account
+            example: nango
+            format: hostname
+            suffix: .freshteam.com
```
bodinsamuel, line 2603: prefix https

bodinsamuel (approved):

packages/shared/providers.yaml
hassan254-prog: still learning 😅 , updated with latest [commit](https://github.com/NangoHQ/nango/pull/3124/commits/60b2eb7c72936faf0f4ecc20981196eb35b77679)

khaliqgant (approved):

PR #3128, opened 2024-12-06
fix(proxy): lowercase incoming headers

bodinsamuel:
## Changes

After #3110 we knew some stuff could happen but didn't expect that bug. Passing a custom `Authorization` (uppercase)  header was not working because `authorization` (lowercase) was now duplicated in the list of headers and ultimately overriding.

- Lowercase incoming headers
- Add tests 
At the end of the file.
- Remove all wrong type-casting in this test file

### DIFF
--- a/packages/shared/lib/services/proxy.service.ts
+++ b/packages/shared/lib/services/proxy.service.ts
@@ -146,7 +146,14 @@ class ProxyService {
             endpoint = endpoint.replace(provider.proxy.base_url, '');
         }
 
-        if (headers && headers['Content-Type'] === 'multipart/form-data') {
+        const headersCleaned: Record<string, string> = {};
+        if (headers) {
+            for (const [key, value] of Object.entries(headers)) {
+                headersCleaned[key.toLocaleLowerCase()] = value;
+            }
+        }
+
+        if (headersCleaned['content-type'] === 'multipart/form-data') {
             const formData = new FormData();
 
             Object.keys(data as any).forEach((key) => {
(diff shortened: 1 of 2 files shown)


packages/shared/lib/services/proxy.service.ts
```diff
@@ -346,8 +351,7 @@ class ProxyService {
         options.url = this.constructUrl(configBody);
         options.method = method;
 
-        const headers = this.constructHeaders(configBody, method, options.url);
-        options.headers = { ...options.headers, ...headers };
```
bodinsamuel, line 350: We were already merging headers in constructHeaders

packages/shared/lib/services/proxy.service.ts
```diff
@@ -146,7 +146,14 @@ class ProxyService {
             endpoint = endpoint.replace(provider.proxy.base_url, '');
         }
 
-        if (headers && headers['Content-Type'] === 'multipart/form-data') {
+        const headersCleaned: Record<string, string> = {};
+        if (headers) {
+            for (const [key, value] of Object.entries(headers)) {
+                headersCleaned[key.toLocaleLowerCase()] = value;
+            }
+        }
```
TBonnin, line 154: reduce would do the job as well
```
cons headersCleaned = Object.entries(headers || {}).reduce(
    (acc, [key, value]) => ({ ...acc, [key.toLowerCase()]: value }), 
    {}
);
```

TBonnin (approved):
:shipit:

nalanj (approved):

packages/shared/lib/services/proxy.service.ts
bodinsamuel: If I could remove one feature from javascript it would be this one, I hate this function so much I found it unreadable :D 
plus it's slow for no reason https://www.measurethat.net/Benchmarks/Show/32914/2/object-for-of-vs-reduce

PR #3129, opened 2024-12-09
feat: fleet into jobs/server

TBonnin:
I apologize in advance for the big PR.
Fleet is now integrated in jobs and server. It is being a feature flag because it only support local runners for now. (The migration creating the tables would run in staging/prod though)

This PR:
- adds runner routing logic in jobs (behind flag)
- implement local runner logic (pretty much copied from previous logic)
- adds register and idle mechanism/endpoints between jobs and runners
- fixes bugs with the fleet supervisor 

Note: I want fleet to manage local runner as well so it is the same code (almost) running locally and in prod. In order to do so every time jobs starts locally a new fake deployment is added to the db so fleet automatically rolls out a new runner and manage its lifecycle. Other than that there are tiny adjustments of the logic that only exist for local execution (ex: shorter timeouts, finishing timeout = idle)

What's not in this PR but will come next:
- /deploy endpoint 
- render node provider
- locking mechanism to ensure only one supervisor is running
- CI script to call /deploy endpoint
- runner config override mechanism for customers that must get more runner resources

# How to test
1. enable feature flag locally
 a. `redis-cli`
 b. `set flag:fleet:global true`
2. start nango locally


packages/jobs/lib/runner/local.ts
```diff
@@ -0,0 +1,81 @@
...
+import type { NodeProvider } from '@nangohq/fleet';
+import { execSync, spawn } from 'child_process';
+import { logger } from '../logger.js';
+import { envs } from '../env.js';
+import { Err, Ok, stringifyError } from '@nangohq/utils';
+import { getProvidersUrl } from '@nangohq/shared';
+
+const localRunnerPids = new Map<number, number>(); // Mapping Node.id to process PID
+
+export const localNodeProvider: NodeProvider = {
```
TBonnin, line 12: that's pretty much a copy of current local runner logic that would be deleted once fleet is fully deployed

packages/jobs/lib/runner/local.ts
```diff
@@ -0,0 +1,81 @@
...
+    start: async (node) => {
+        try {
+            // Random port to avoid conflicts with other fleet runners accross local execution
+            const rndPort = Math.floor(Math.random() * 10000) + 10000;
+            const port = await getPort({ port: portNumbers(rndPort, rndPort + 100) });
```
TBonnin, line 17: The main change (other than the pids map) is the random port since runners live in the db across execution. Always assigning the same port would lead to runners to have the same url even though they are actually different runners

packages/runner/lib/monitor.ts
```diff
@@ -49,6 +51,10 @@ export class RunnerMonitor {
         }
     }
 
+    resetIdleMaxDurationMs(): void {
+        this.idleMaxDurationMs = 1; // 0 is a special value that disables idle tracking
```
bodinsamuel, line 55: ooc what's the use case of disabling idling (if you know)

packages/jobs/lib/routes/runners/postRegister.ts
```diff
@@ -0,0 +1,49 @@
...
+    parseBody: (data) =>
+        z
+            .object({ url: z.string().min(1) })
+            .strict()
+            .parse(data)
+});
+
+const handler = async (req: EndpointRequest<PostRegister>, res: EndpointResponse<PostRegister>) => {
+    try {
+        const register = await runnersFleet.registerNode({ nodeId: req.params.nodeId, url: req.body.url });
```
bodinsamuel, line 34: why would the runner know its URL? 
I expected the runner to be spawned and says 'I'm up' and that's it no?

packages/fleet/lib/models/helpers.ts
```diff
@@ -0,0 +1,17 @@
...
+    const charset = '0123456789abcdef';
+    const length = 40;
+    const randomBytes = new Uint8Array(length);
+    crypto.getRandomValues(randomBytes);
+
+    const value = Array.from(randomBytes)
+        .map((byte) => charset[byte % charset.length])
+        .join('');
+    if (value.length !== 40) {
+        throw new Error('CommitHash must be exactly 40 characters');
```
bodinsamuel: not returning Result?

packages/fleet/lib/instances.ts
```diff
@@ -0,0 +1 @@
+export type FleetId = 'nango_runners';
```
bodinsamuel, line 1: still not sure why we need this?

packages/fleet/lib/fleet.ts
```diff
@@ -1,19 +1,89 @@
...
+        const recurse = async (start: Date): Promise<Result<Node>> => {
+            if (new Date().getTime() - start.getTime() > timeout) {
+                return Err(new FleetError('fleet_node_not_ready_timeout', { context: { routingId } }));
+            }
+            const search = await nodes.search(this.dbClient.db, {
+                states: ['PENDING', 'STARTING', 'RUNNING'],
+                routingId
+            });
+            if (search.isErr()) {
+                return Err(search.error);
```
bodinsamuel, line 63: It's a bit annoying that Error and Error are not compatible

packages/fleet/lib/supervisor.ts
```diff
@@ -14,15 +14,16 @@ type Operation =
...
 
 type SupervisorState = 'stopped' | 'running' | 'stopping';
 
 export const STATE_TIMEOUT_MS = {
-    STARTING: 5 * 60 * 1000,
-    FINISHING: 24 * 60 * 60 * 1000,
+    STARTING: (isLocal ? 30 : 5 * 60) * 1000, // 30 secs locally, 5 minutes otherwise
+    FINISHING: (isLocal ? 30 : 24 * 60 * 60) * 1000, // 30 secs locally, 24 hours otherwise
     TERMINATED: 7 * 24 * 60 * 60 * 1000,
     ERROR: 7 * 24 * 60 * 60 * 1000
```
bodinsamuel: env var would be beneficial here

packages/fleet/lib/supervisor.ts
```diff
@@ -251,6 +281,23 @@ export class Supervisor {
...
 
+    private async finishingNode({ node }: { type: 'FINISHING'; node: Node }): Promise<Result<Node>> {
+        if (!node.url) {
+            return Err(new FleetError('fleet_node_url_not_found', { context: { nodeId: node.id } }));
+        }
+
+        try {
+            await fetch(`${node.url}/notifyWhenIdle`, { method: 'POST', body: JSON.stringify({ nodeId: node.id }) });
+        } catch (error) {
+            logger.warning(`Failed to notify node ${node.id} to notifyWhenIdle: ${error}`);
```
bodinsamuel: no retry?

packages/fleet/lib/supervisor.ts
```diff
@@ -271,9 +318,29 @@ export class Supervisor {
         return nodes.remove(this.dbClient.db, { nodeId: node.id });
     }
 
-    private async warnNode({ node }: { type: 'WARN'; node: Node }): Promise<Result<Node>> {
+    private async finishingTimeout({ node }: { type: 'FINISHING_TIMEOUT'; node: Node }): Promise<Result<Node>> {
+        // Locally we assume the node is IDLE
+        // since the process is likely already killed
+        // and the node is not able to notify back that it is idle
+        if (isLocal) {
```
bodinsamuel: I would love to have a dedicated env var instead of using isLocal

packages/fleet/lib/fleet.ts
```diff
@@ -1,19 +1,89 @@
...
+import { setTimeout } from 'node:timers/promises';
+import { Supervisor } from './supervisor.js';
+import type { NodeProvider } from './node-providers/node_provider.js';
+import type { FleetId } from './instances.js';
+import { noopNodeProvider } from './node-providers/noop.js';
+import { envs } from './env.js';
+
+const defaultDbUrl =
+    envs.NANGO_DATABASE_URL ||
+    `postgres://${encodeURIComponent(envs.NANGO_DB_USER)}:${encodeURIComponent(envs.NANGO_DB_PASSWORD)}@${envs.NANGO_DB_HOST}:${envs.NANGO_DB_PORT}/${envs.NANGO_DB_NAME}`;
```
bodinsamuel, line 17: We are going to use a dedicated DB for that or just a different schema?
because for the keystore we choose a different path of setting migrations in a different schema but using the same schema for the table

packages/jobs/lib/app.ts
```diff
@@ -63,6 +66,16 @@ try {
         // not closing on purpose
     });
 
+    if (envs.RUNNER_TYPE !== 'REMOTE') {
+        if (isLocal) {
+            // when running locally, the runners (running as processes) are being killed
+            // when the main process is killed and the fleet entries are therefore not associated with any running process
+            // we then must fake a new deployment so fleet replaces runners with new ones
+            await runnersFleet.deploy(generateCommitHash());
```
bodinsamuel: it's handy but will create a lot of rows, out of curiosity why not strip the `deployments` and just use current hash?

bodinsamuel (commented):
Works locally 👌🏻 
Really not sure maintaining this code will be easy, maybe at some point using automated solution (whatever the name) would be nice.

packages/runner/lib/monitor.ts
TBonnin: that's one of the settings we override for some scale plan customers so we prevent their runner to be suspended under any circumstances

packages/jobs/lib/routes/runners/postRegister.ts
TBonnin: To allow for more flexibility. It is not necessary for Render but if we decide to move to a new system that is allocating network addresses more dynamically we will have need it.

packages/fleet/lib/supervisor.ts
TBonnin: you are right we should retry

packages/jobs/lib/routes/runners/postRegister.ts
bodinsamuel: mmh I feel like it's dangerous. Currently, this API is not protected so it's a gapping hole, but even then I wouldn't trust a ping with a random URL.

packages/fleet/lib/supervisor.ts
TBonnin: actually we have a RUNNER_TYPE env vars. Let's use this

packages/jobs/lib/app.ts
TBonnin: because each local execution (even from the same current hash) is in reality a different local runner so it has to be different for each execution.
wrt to the number of rows, `supervisor` is cleaning up terminated runners after 7 days so rows won't be increasing forever

packages/fleet/lib/fleet.ts
TBonnin: it is not a problem of compatibility of Error. It is `Result<Search, Error>` that is not compatible with `Result<Node, Error>. A bit annoying indeed when you are only dealing with the Error case though

packages/fleet/lib/instances.ts
TBonnin: It is not mandatory. It is just to give an explicit `id` to things in general and, in this case, not assuming fleet is a singleton for ever. I know it is right now but for the cost of 3 lines of code it could save a lot of pain later if we ever need to split the fleet for some reason

packages/fleet/lib/fleet.ts
TBonnin: Right now it is setup to use different schema for the tables and the migrations. My reasoning is that fleet is somehow independent from the rest of the codebase, both codewise and conceptually. Wdyt?

packages/fleet/lib/models/nodes.ts
```diff
@@ -107,11 +107,11 @@ export async function create(
     try {
         const inserted = await db.from<DBNode>(NODES_TABLE).insert(newNode).returning('*');
         if (!inserted?.[0]) {
-            return Err(new FleetError(`node_creation_error`, { context: nodeProps }));
```
nalanj, line 110: Reading all of these, I feel like it could be nice to have a wrapper pattern on `Err` of `FleetErr` that takes the args for `FleetError` and returns it since it gets typed so often.

Then this would look like:

```
return FleetErr(`node_creation_error`, { context: nodeProps });
```

packages/fleet/lib/supervisor.ts
```diff
@@ -251,6 +281,23 @@ export class Supervisor {
         });
     }
 
+    private async finishingNode({ node }: { type: 'FINISHING'; node: Node }): Promise<Result<Node>> {
+        if (!node.url) {
+            return Err(new FleetError('fleet_node_url_not_found', { context: { nodeId: node.id } }));
+        }
+
+        try {
+            await fetch(`${node.url}/notifyWhenIdle`, { method: 'POST', body: JSON.stringify({ nodeId: node.id }) });
```
nalanj: I think you need an `ok` check on the fetch here as well.

packages/jobs/lib/runner/local.ts
```diff
@@ -0,0 +1,81 @@
...
+            let nodePath = '';
+            try {
+                nodePath = execSync('which node', { encoding: 'utf-8' }).trim();
```
nalanj: I think you could just use `process.argv[0]` here, which is guaranteed to be the absolute path to the currently running version of nodejs.

packages/jobs/lib/runner/local.ts
nalanj: Could we just use the socket native port handling to pick an unused port?

nalanj (approved):
Left a few notes, but looks good overall!

packages/jobs/lib/runner/local.ts
TBonnin: that's what `getPort` do: finding the first unused port in the provided range.

packages/jobs/lib/routes/runners/postRegister.ts
TBonnin: good point about trusting random URL. I've added a check when registering to ensure the received url is acceptable. 15982da4f

TBonnin: regarding the non secure endpoint. That's applicable to all jobs endpoint and should be addressed at some point I agree

TBonnin:
> Really not sure maintaining this code will be easy, maybe at some point using automated solution (whatever the name) would be nice.

@bodinsamuel I wish I didn't have to write this code to be honest and that a AWS lambda-like product with no cold start and infinite runtime existed. That would solve most of our problems but unfortunately it looks like our use-case is not easy after all, hence the not-easy-to-maintain code :(
Even if we were using k8s for example for the orchestration we would still need to write code to handle talking to k8s api and also have code to manage the termination of runner pods. It would probably be less code (leveraging some k8s primitives) but it will live half in the infra and half in the codebase, still not easy to maintain.

packages/fleet/lib/models/nodes.ts
TBonnin: Wrapping the function or wrapping the Error like it is right now is pretty much the same. no? Unless you are talking about implicit type conversion but I am not sure typescript can do that.

packages/jobs/lib/runner/local.ts
nalanj: Yeah, I was more just saying anything like that by default has a race condition in the OS where you're picking an unused port but it can be grabbed before you start the process using the port. Any TCP based service usually has an option where you start it without giving it a port and the OS will give you one that's unused and guarantee there's no race condition.

packages/fleet/lib/models/nodes.ts
nalanj: Yeah, it's effectively the same, but I think it reads a little nicer if it's wrapped up, especially given how often we use the pattern.

packages/jobs/lib/runner/local.ts
TBonnin: the race condition you mention exists. I think they even mention it in the readme of the getPort npm package. Not worth spending too much time on this imho given how unlikely it is to happen and that it will be fixed next time it is run

packages/fleet/lib/fleet.ts
bodinsamuel: yeah that's what I meant. Just suggesting there might be a way to make TS happy when return the same kind of error

packages/fleet/lib/fleet.ts
bodinsamuel: Ideally infra is not directly tied to the code but since it's now a first-citizen I'm not sure the project will be usable without this, aka not independent from the rest of the code. So I wouldn't bother splitting. But doesn't change much, as you wish

bodinsamuel:
>  I wish

Yeah not saying it's bad, just saying it has a big footprint in the codebase. Until we change our current paradigm of "one machine per user" it will stay this way; but as soon as we can execute jobs on any machine it will be easier to just plug a queue (or an http load balancer) and let the orchestration to the cloud provider.
With a queue and Kubernetes (or cloud functions or cloud jobs and maybe alternative we don't know on AWS) it's possible to run tasks without having to deal with infra at all.

packages/fleet/lib/fleet.ts
```diff
@@ -1,19 +1,95 @@
...
+    public async stop(): Promise<void> {
+        await this.supervisor.stop();
+    }
+
     public async deploy(commitId: CommitHash): Promise<Result<Deployment>> {
         return deployments.create(this.dbClient.db, commitId);
     }
+
+    public async getRunningNode(routingId: RoutingId): Promise<Result<Node>> {
+        const timeout = 1000 * 60; // 1 minute
```
bodinsamuel: process env

bodinsamuel (approved):
🏃🏻

PR #3130, opened 2024-12-09
fix(regex): allow special characters

khaliqgant:
Allow special characters for freshteams

### DIFF
--- a/packages/shared/providers.yaml
+++ b/packages/shared/providers.yaml
@@ -2721,7 +2721,7 @@ freshteam:
             description: The API key to your Freshteam account
             doc_section: '#step-2-finding-your-api-key'
             example: 'XtoF94LMNyKPbd5AvR1QJh'
-            pattern: '^[a-zA-Z0-9]+$'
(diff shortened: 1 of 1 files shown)


packages/shared/providers.yaml
```diff
@@ -2721,7 +2721,7 @@ freshteam:
             description: The API key to your Freshteam account
             doc_section: '#step-2-finding-your-api-key'
             example: 'XtoF94LMNyKPbd5AvR1QJh'
-            pattern: '^[a-zA-Z0-9]+$'
+            pattern: '^[a-zA-Z0-9_!@#$%^&*()\\-+=]+$'
```
bodinsamuel, line 2724: what kind of api key is this? 🤔

bodinsamuel (approved):

packages/shared/providers.yaml
khaliqgant: Not sure, just the client reported that our regex was too restrictive for their API key. The provider docs don't have any info on the structure unfortunately.

bodinsamuel: would have been nice to confirm with the customer because a too broad validation is not very useful

PR #3135, opened 2024-12-09
fix(ui): do not identify user on debug mode

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2330/do-not-identify-user-in-debug-mode

- Do not identify user on debug mode

### DIFF
--- a/packages/webapp/src/components/PrivateRoute.tsx
+++ b/packages/webapp/src/components/PrivateRoute.tsx
@@ -66,11 +66,11 @@ export const PrivateRoute: React.FC = () => {
     }, [meta, loadingMeta, env, error, setEnv]);
 
     useEffect(() => {
-        if (user && environmentAndAccount) {
+        if (user && environmentAndAccount && meta && !meta.debugMode) {
             identify(user);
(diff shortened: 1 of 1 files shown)


packages/webapp/src/components/PrivateRoute.tsx
```diff
@@ -66,11 +66,11 @@ export const PrivateRoute: React.FC = () => {
     }, [meta, loadingMeta, env, error, setEnv]);
 
     useEffect(() => {
-        if (user && environmentAndAccount) {
+        if (user && environmentAndAccount && meta && !meta.debugMode) {
```
TBonnin, line 69: if `meta` is undefined, we will not enter the if block, which is a change of behavior. Is that ok?

TBonnin (approved):

packages/webapp/src/components/PrivateRoute.tsx
bodinsamuel: meta was already mandatory for user, it's just ts pleasing

PR #3136, opened 2024-12-09
feat(integrations): add support for Fortnox

bodinsamuel:
## Changes

- Add support for Fortnox 
on behalf of a customer because there was an issue, but then it turns out there was no issue just something miss-configured.


packages/shared/providers.yaml
```diff
@@ -2601,6 +2601,27 @@ fitbit:
...
     docs: https://docs.nango.dev/integrations/all/fitbit
 
+fortnox:
+    display_name: Fortnox
+    categories:
+        - accounting
+        - invoicing
+    auth_mode: OAUTH2
+    authorization_url: https://apps.fortnox.se/oauth-v1/auth
+    token_url: https://apps.fortnox.se/oauth-v1/token
```
nalanj, line 2611: Pretty wild that their oauth2 flow is under a path named `oauth-v1`.

nalanj (approved):
:shipit:

packages/shared/providers.yaml
bodinsamuel: yes it took me a while to stop debugging oauth v1 :D

TBonnin: 😮‍💨

TBonnin (approved):

PR #3138, opened 2024-12-09
feat(cli): Support saving error responses in dryrun

nalanj:
This adds support for `response-saver` dumping failed requests to disk, so that we can at least manually write tests against failed requests. It unfortunately required me to add `onRejected` to our axios interceptors, but I think it came out pretty OK.

## How I tested it:

Run a dry-run that includes failed requests with the CLI. Khaliq and I were running this one in the particular customer issue we were dealing with:

```
npm run dryrun -- anrok create-ephemeral-transaction a -i @anrok/fixtures/invoice.json --save-responses
```


nalanj:
Wanted to note that @khaliqgant was already able to use this to capture errors for a test for a customer.

packages/node-client/lib/index.ts
```diff
@@ -66,6 +66,7 @@ export interface AdminAxiosProps {
         request?: AxiosInterceptorManager<AxiosRequestConfig>;
         response?: {
             onFulfilled: (value: AxiosResponse) => AxiosResponse | Promise<AxiosResponse>;
+            onRejected?: (error: unknown) => void;
```
TBonnin: Is it not returning an `AxiosError | Promise<AxiosError>?`

bodinsamuel (approved):

packages/node-client/lib/index.ts
nalanj: We can. I'm pretty sure Axios only needs it back if you change it, because just returning nothing was working fine, but interceptors in Axios are pretty poorly documented so I'll update it to be consistent with what we're doing on the request. I was avoiding it from the principle of the best code being the code you never write.

nalanj: I have an update incoming once tests pass, but I wanted to note here that the types from Axios for the rejected interceptor are unknown => unknown

nalanj: @TBonnin see 6a1b015

TBonnin (approved):
I haven't tested it but code lgtm

PR #3139, opened 2024-12-09
fix: Return 400 when /connections/:connetionId missing provider_config_key

nalanj:
We had a customer reach out saying they were unable to get /connections/:connectionId to work and were just getting an error. It turns out they had left the required `provider_config_key` param off but it wasn't being validated so they were getting a generic 500 error instead of a clear 400 error. This fixes that.

## How I tested it

```
curl --request GET \
  --url 'http://localhost:3003/connection/test-connection-id' \
  --header 'Authorization: Bearer [bearer]'
```


packages/server/lib/controllers/connection.controller.ts
```diff
@@ -34,6 +34,11 @@ class ConnectionController {
             const instantRefresh = req.query['force_refresh'] === 'true';
             const isSync = (req.get('Nango-Is-Sync') as string) === 'true';
 
+            if (!providerConfigKey) {
```
TBonnin, line 37: should we also error if it is an empty string?

TBonnin (approved):

packages/server/lib/controllers/connection.controller.ts
bodinsamuel: it's already taken care of thanks to javascript

bodinsamuel (approved):

PR #3141, opened 2024-12-09
feat(fleet): ensure only one fleet supervisor is running

TBonnin:
only one fleet supervisor must run at anytime, otherwise it can lead to race conditions when managing the lifecycle of nodes.
This PR adds a locking mechanism using pg advisory lock to ensure only one supervisor is running at any given point in time.

### DIFF
--- a/packages/fleet/lib/utils/locking.ts
+++ b/packages/fleet/lib/utils/locking.ts
@@ -1,5 +1,5 @@
 import type { Knex } from 'knex';
-import { Ok, Err, stringToHash } from '@nangohq/utils';
+import { Err, stringToHash } from '@nangohq/utils';
 import { setTimeout } from 'node:timers/promises';
 import type { Result } from '@nangohq/utils';
 import { FleetError } from './errors.js';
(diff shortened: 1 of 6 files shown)


nalanj (approved):
Failing tests, but code looks good.

packages/fleet/lib/utils/locking.ts
```diff
@@ -0,0 +1,45 @@
...
+}): Promise<Result<void>> {
+    let trx: Knex.Transaction | undefined = undefined;
+    try {
+        trx = await db.transaction();
+        const { rows } = await trx.raw<{ rows: { lock: boolean }[] }>(`SELECT pg_try_advisory_xact_lock(?) as lock`, [stringToHash(lockKey)]);
+
+        if (!rows?.[0]?.lock) {
+            await trx.rollback();
+            await setTimeout(1000);
+            return Ok(undefined);
```
bodinsamuel: Why not returning an error if higher up you are waiting 1s on error?

bodinsamuel (approved):

packages/fleet/lib/utils/locking.ts
TBonnin: you are right. I am doing twice the same thing.

PR #3145, opened 2024-12-10
feat(ui): connection trigger sync, new components

bodinsamuel:
## Changes 

Fixes https://linear.app/nango/issue/NAN-2264/trigger-sync-ui
Figma https://www.figma.com/design/hKL3cmI9vBwDoZyRahQnsq/Connections-Tab?node-id=4844-10056&node-type=frame&m=dev

### Major

- Change UI to trigger sync, move everything into a modal
- Allow full refresh to not clear cache

### Minor

- New button style
- New table style
- New tag style
- New select style
- New dialog style
- Add ButtonLink to fix double layer `a > button` which makes links less accessible

##🧪 Tests

- go to your favorite dashboard > connections > pick one > trigger sync

Screenshot 2024-12-10 at 17 44 28
Screenshot 2024-12-10 at 17 44 34


packages/webapp/src/hooks/useSyncs.tsx
```diff
@@ -12,3 +12,26 @@ export function useSyncs(queries: { env: string; connection_id: string; provider
...
+    }
+) {
+    const res = await apiFetch(`/api/v1/sync/command?env=${env}`, {
+        method: 'POST',
+        body: JSON.stringify(body)
+    });
+
+    return {
+        res,
+        json: await res.json()
```
nalanj, line 35: What happens here if an LB or something else along the way sends back non-json?

packages/webapp/src/pages/Connection/components/SyncRow.tsx
```diff
@@ -69,24 +88,23 @@ export const SyncRow: React.FC<{ sync: SyncResponse; connection: ApiConnectionFu
...
             return;
         }
 
         setSyncCommandButtonsDisabled(true);
-        const res = await runCommandSyncAPI(command, scheduleId, nango_connection_id, syncId, syncName, provider || '');
+        const res = await apiRunSyncCommand(env, { command, schedule_id: scheduleId, nango_connection_id, sync_id: syncId, sync_name: syncName, provider });
 
-        if (res?.status === 200) {
-            await mutate((key) => typeof key === 'string' && key.startsWith(`/api/v1/sync`), undefined);
+        if (res.res.status === 200) {
```
nalanj, line 101: Why not just return the fetch result so we don't have `res.res` here?

packages/webapp/src/pages/Connection/components/SyncRow.tsx
```diff
@@ -7,48 +7,67 @@ import { formatFrequency, getRunTime, parseLatestSyncResult, formatDateToUSForma
...
+            await mutate((key) => typeof key === 'string' && key.startsWith(`/api/v1/sync`));
             toast({ title: `The full resync was successfully triggered`, variant: 'success' });
         } else {
-            const data = await res?.json();
-            toast({ title: data.error, variant: 'error' });
+            const data = await res.json;
```
nalanj: At https://github.com/NangoHQ/nango/pull/3145/files#diff-5e93070819979f0cf4bf249786ba5c96e1cb3e4f0c27d4d3913cc8bfcb249847R35 `res.json()` is already being awaited into the value, so I don't think you need an await here, although it won't hurt anything. Should we have types around all of this?

packages/webapp/src/pages/Connection/components/SyncRow.tsx
```diff
@@ -7,48 +7,67 @@ import { formatFrequency, getRunTime, parseLatestSyncResult, formatDateToUSForma
...
+                  })
+                : await apiRunSyncCommand(env, {
+                      command: 'RUN_FULL',
+                      schedule_id: sync.schedule_id,
+                      nango_connection_id: sync.nango_connection_id,
+                      sync_id: sync.id,
+                      sync_name: sync.name,
+                      provider,
+                      empty_cache: emptyCache
+                  });
```
nalanj: This is hard to scan visually and see what's different between the two calls. Would something like this be clearer?

```suggestion
        const res = await apiRunSyncCommand(env, {
                      command: 'RUN',
                      schedule_id: sync.schedule_id,
                      nango_connection_id: sync.nango_connection_id,
                      sync_id: sync.id,
                      sync_name: sync.name,
                      provider,
                      ...(triggerMode === 'incremental' ? {
                        command: 'RUN_FULL',
                        empty_cache: emptyCache,
                      } : {})
                  });
```

nalanj (commented):
Looks great overall, and I think the dialog is a big improvement on handling manual syncs. Left a few comments around those http calls.

packages/server/lib/controllers/sync.controller.ts
```diff
@@ -546,7 +546,7 @@ class SyncController {
         try {
             const { account, environment } = res.locals;
 
-            const { schedule_id, command, nango_connection_id, sync_id, sync_name, provider } = req.body;
+            const { schedule_id, command, nango_connection_id, sync_id, sync_name, provider, empty_cache } = req.body;
```
TBonnin: not sure about calling this option  `empty_cache`. why not `delete_records` which is straight to the point?

TBonnin (approved):
The action menu doesn't disappear after I click confirm. I don't know if this is on purpose.

Screenshot 2024-12-10 at 23 37 57

packages/webapp/src/pages/Connection/components/SyncRow.tsx
bodinsamuel: Ah yes right.
This is a legacy endpoint that does not have types so it's broken right now

packages/webapp/src/hooks/useSyncs.tsx
bodinsamuel: that's a problem for a future us :D 
no, but we don't handle this anywhere else, we probably need to test how it behaves and have a global fix. I guess rn it will throw and display the ErrorFallback component in the UI

packages/server/lib/controllers/sync.controller.ts
bodinsamuel: yeah true, let me change that

packages/webapp/src/pages/Connection/components/SyncRow.tsx
bodinsamuel: we return the content in `json` but some calls need to check the status for legacy reason (like this one) so we added that. As soon as it's not not needed anymore we can just return the fetch content.

packages/webapp/src/pages/Connection/components/SyncRow.tsx
bodinsamuel: not a huge fan tbh, both have their own drawbacks

packages/webapp/src/hooks/useSyncs.tsx
nalanj: Yeah, was just curious. There's also probably an issue here of it returning a valid json body with a non-200 response code that could also be a challenge.

PR #3147, opened 2024-12-10
feat(integration): add docs and post connection script for adobe umapi

hassan254-prog:

- Add `connect-ui` docs for adobe umapi
- Add a post connection script to grab and update the `connectionConfig` with a `clientId` which is crucial while building proxies

## Testing instructions

- This has not been tested with `adobe-umapi` as we don't have test credentials, but I have tested this with a similar provider, `tapclicks` that uses the same auth mode `oauth2_cc`.
- One can have the, `post_connection_script: adobeUmapiPostConnection` logic within `tapclicks` and create a new connection, the `clientId` is automatically populated.


packages/shared/providers.yaml
```diff
@@ -73,11 +73,14 @@ adobe-umapi:
...
     docs: https://docs.nango.dev/integrations/all/adobe
+    post_connection_script: adobeUmapiPostConnection
+    docs_connect: https://docs.nango.dev/integrations/all/adobe-umapi/connect
     connection_config:
         clientId:
             type: string
-            title: Client ID
-            description: The Client ID from the OAuth Server-to-Server enabled project on developer portal
+            title: ''
+            description: ''
```
TBonnin, line 82: why emptying the title and description?

hassan254-prog: They are both needed in order for validation to pass, but at the same time we are not going to display it to the user hence no need of having any text as it is automated with the `postConnection` script.

bodinsamuel: yeah indeed. At some point we might still want to document those values in the dashboard so we'll see then

bodinsamuel (approved):

khaliqgant (approved):

PR #3148, opened 2024-12-10
feat(integrations): add harvest connect ui docs and a post connection script

hassan254-prog:

- Added harvest `connect-ui` docs
- Added a post connection script to update the `connectionConfig` with a `emailConnection` value this way if the `appDetails` is not provided nango will automatically use this value to build proxies which is crucial. The provider recommends a unique name and means of communication, hence the pattern `connection_id.email` in the `postConnection` script.
- 

## Testing instructions

- Create a new harvest connection, the `emailConnection` is automatically populated.


packages/shared/lib/services/user.service.ts
```diff
@@ -51,8 +51,12 @@ class UserService {
         return result[0] || null;
     }
 
-    async getUsersByAccountId(accountId: number): Promise<DBUser[]> {
-        const result = await db.knex.select('*').from<DBUser>(`_nango_users`).where({ account_id: accountId, suspended: false });
+    async getUsersByAccountId(accountId: number, requireEmail: boolean = false): Promise<DBUser[]> {
+        let query = db.knex.select('*').from<DBUser>(`_nango_users`).where({ account_id: accountId, suspended: false });
+        if (requireEmail) {
+            query = query.orderBy('created_at', 'asc').limit(1);
+        }
```
bodinsamuel: the `if` does not match the query condition. I'm not sure what it's really getting here.

packages/server/lib/hooks/connection/post-connection.ts
```diff
@@ -52,6 +53,13 @@ async function execute(createdConnection: RecentlyCreatedConnection, providerNam
 
                 return connection as Connection;
             },
+            getDbUser: async () => {
+                const users = await userService.getUsersByAccountId(account.id, true);
```
bodinsamuel: You are just getting a random db user from this account, the name doesn't reflect that and it might not be really nice for testing.
I guess that's why you ordered by creation to get the "admin" but there is a good chance it's incorrect. 

Can we think of something else? @NangoHQ/engineers

packages/server/lib/hooks/connection/harvest-post-connection.ts
```diff
@@ -0,0 +1,16 @@
...
+    const user = await nango.getDbUser();
+    const connection = await nango.getConnection();
+
+    if (user === null) {
+        const logger = getLogger('post-connection:harvest');
+        logger.info('No user with email found for post-connection logic');
+        return;
+    }
+
+    const emailConnection = `${connection.connection_id}-${user.email}`;
```
bodinsamuel: shouldn't it be `MyApp (yourname@example.com)` --> `${connection.connection_id} (user.email)`?

khaliqgant:
@bodinsamuel curious on your thoughts here if we need connect UI docs at all for the App Details portion. To me, it makes no sense that the end user would put in any info or even know what to put in for the "App Details". If anything it should be set programmatically and then we fall back to the email via the post connection script as this PR adds in. 

With regards to

> You are just getting a random db user from this account, the name doesn't reflect that and it might not be really nice for testing.
I guess that's why you ordered by creation to get the "admin" but there is a good chance it's incorrect.

As a fallback I think it is OK to have some value inputted as this "App Details" value is required, and even if the email is potentially incorrect I don't see any other way to fill this fallback value. So IMO this potentially incorrect fallback is OK.

Let me know your thoughts!

bodinsamuel:
> @bodinsamuel curious on your thoughts here if we need connect UI docs at all for the App Details portion. 

I don't have enough knowledge on this unfortunately, that's why I have more questions :D 
It seems unnecessary tbh (see next point)

> As a fallback I think it is OK to have some value inputted as this "App Details" value is required, and even if the email is potentially incorrect I don't see any other way to fill this fallback value. So IMO this potentially incorrect fallback is OK.

Why not just `no-reply@nango.dev`? 
It contradicts what I said in an other PR but that way we can get rid of this value altogether and we don't query something random from the DB. Not ideal for reporting but Harvest should have an email registered so it's probably useless. 
For this use case (and probably others) we could add a new field in environment "Main email contact for third party"

khaliqgant:
Replaced by https://github.com/NangoHQ/nango/pull/3160

khaliqgant (approved):

bodinsamuel (approved):
As always thanks @hassan254-prog for the grit and pushing for the right thing 💪🏻

hassan254-prog:
> As always thanks @hassan254-prog for the grit and pushing for the right thing 💪🏻

Thank you @bodinsamuel ! I really appreciate  🙏

PR #3149, opened 2024-12-10
feat(fleet): add Render NodeProvider

TBonnin:
- adding Render node provider. This is highly dependent on lifting the rate limit for creating service. Otherwise we are gonna need to go to AWS directly
- do not fail node when node provider errors and rely on the nature of the control loop to retry. Adding some timeout to deal with never ending errors
- adding a lock around the node creation to prevent race condition that could happen the first time a runner is created if multiple scripts are executed around the same time
- fixing some bugs

# How to test
Unfortunately you cannot test locally, only in staging

### DIFF
--- a/packages/jobs/lib/runner/fleet.ts
+++ b/packages/jobs/lib/runner/fleet.ts
@@ -1,8 +1,23 @@
-import type { NodeProvider } from '@nangohq/fleet';
 import { Fleet } from '@nangohq/fleet';
 import { envs } from '../env.js';
 import { localNodeProvider } from './local.js';
+import { renderNodeProvider } from './render.js';
 
-const nodeProvider: NodeProvider | undefined = envs.RUNNER_TYPE === 'LOCAL' ? localNodeProvider : undefined;
+const defaultNodeConfig = {
+    image: 'nangohq/nango-runner',
+    cpuMilli: 500,
+    memoryMb: 512,
+    storageMb: 20000
+};
 
-export const runnersFleet = new Fleet({ fleetId: 'nango_runners', nodeProvider });
+const fleetId = 'nango_runners';
+export const runnersFleet = (() => {
+    switch (envs.RUNNER_TYPE) {
+        case 'LOCAL':
+            return new Fleet({ fleetId, nodeSetup: { nodeProvider: localNodeProvider, defaultNodeConfig } });
+        case 'RENDER':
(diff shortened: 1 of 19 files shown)


packages/jobs/lib/runner/fleet.ts
```diff
@@ -1,8 +1,23 @@
...
+    memoryMb: 512,
+    storageMb: 20000
+};
 
-export const runnersFleet = new Fleet({ fleetId: 'nango_runners', nodeProvider });
+const fleetId = 'nango_runners';
+export const runnersFleet = (() => {
+    switch (envs.RUNNER_TYPE) {
+        case 'LOCAL':
+            return new Fleet({ fleetId, nodeSetup: { nodeProvider: localNodeProvider, defaultNodeConfig } });
```
bodinsamuel: Local Runner does not really have default node config, it should be optional no or simply stored in the provider?

edit: it's not used right now, no?

bodinsamuel (approved):
Seems okay

--- 

Not related to this PR, could we generate an internal api key (using `private_keys`) linked to single deployment?
So when you receive an HTTP call you know who did it and you can safely assume it's coming from inside and can only touch one serviceId.
Simple,  efficient and reuse something we already knew could benefits other features

packages/jobs/lib/runner/fleet.ts
TBonnin: Good point. There are indeed linked and maybe should be stored in the provider.

TBonnin:
> Not related to this PR, could we generate an internal api key (using `private_keys`) linked to single deployment? So when you receive an HTTP call you know who did it and you can safely assume it's coming from inside and can only touch one serviceId. Simple, efficient and reuse something we already knew could benefits other features

Are we talking about securing `jobs` endpoints? Creating a private key per runner?

packages/jobs/lib/runner/fleet.ts
TBonnin: fixed in 7b86be260

bodinsamuel:
> Are we talking about securing jobs endpoints? Creating a private key per runner?

Yes, it's completely unrelated but just thought about that :D

TBonnin:
> > Are we talking about securing jobs endpoints? Creating a private key per runner?
> 
> Yes, it's completely unrelated but just thought about that :D

That's a good idea. At least one key is limited to the scope of one runner. Will keep this solution in mind

PR #3151, opened 2024-12-11
docs: authorize and reconnect

bodinsamuel:
## Changes

- Document how to reconnect


khaliqgant (approved):

docs-v2/guides/getting-started/authorize-an-api-from-your-app.mdx
```diff
@@ -141,15 +141,62 @@ Successful authorization webhooks sent by Nango are `POST` requests with the fol
...
 ```
 
-For each successful authorization, persist the `connectionId` value alongside its corresponding user or organization, designated by `endUser.endUserId` and `endUser.organizationId`. 
+For each successful authorization, persist the `connectionId` value alongside its corresponding user or organization, designated by `endUser.endUserId` and `endUser.organizationId`.
 
 # Troubleshoot authorization errors
 
 If an authorization request fails, you can analyze the relevant log in the _Logs_ tab of the Nango UI.
 
+# Reconnect a Connection ID
```
bastienbeurier: Would rename to: Reconnect an existing connection

docs-v2/guides/getting-started/authorize-an-api-from-your-app.mdx
```diff
@@ -141,15 +141,62 @@ Successful authorization webhooks sent by Nango are `POST` requests with the fol
...
+
+Nango automatically refreshes OAuth tokens. However, other types of credentials can expire, or permissions may change, so you might need to reconnect your users periodically.
```
bastienbeurier: Would replace this paragraph with: 

When testing or troubleshooting an issue, you may need a user to re-authorize an API. Nango allows you to do this without deleting and recreating the connection. Instead, you can reconnect an existing connection while preserving its metadata and configuration.

In **your backend**, use the `/connect/sessions/reconnect` endpoint ([API](https://google.com/)/[SDK](https://google.com/) reference) to generate a session token specifically for reconnecting a user. This token is then used on the frontend in the same way as when creating a connection, but it will reconnect the existing connection instead.

PR #3153, opened 2024-12-11
chore(eslint): enforce err name, upgrade eslint plugins

bodinsamuel:
## Changes

- Enforce try/catch err variable name
It's been in my backlog for a long time. I have chosen `err` because that's what I like the most, it's neither too short nor too common. Happy to change to error if someone prefer.

- Upgrade eslint plugins


nalanj (approved):
I pretty apathetic about if catch var names are all consistent, so sure, why not.

TBonnin (approved):
I like `err` 😄

PR #3155, opened 2024-12-11
feat(integration): back fill missing base url

hassan254-prog:

- Backfill missing `proxy.base_url`

### DIFF
--- a/packages/shared/providers.yaml
+++ b/packages/shared/providers.yaml
@@ -514,7 +514,7 @@ auth0:
         grant_type: refresh_token
     docs: https://docs.nango.dev/integrations/all/auth0
     proxy:
-        base_url: https://${connectionConfig.subdomain}
+        base_url: https://${connectionConfig.subdomain}.auth0.com
     connection_config:
         subdomain:
             type: string
@@ -748,7 +748,7 @@ battlenet:
     token_params:
         grant_type: authorization_code
     proxy:
-        base_url: https://${connectionConfig.apiBaseUrl}
+        base_url: https://${connectionConfig.apiDomain}
     docs: https://docs.nango.dev/integrations/all/battlenet
     connection_config:
         extension:
@@ -758,10 +758,10 @@ battlenet:
             example: com
             pattern: '^[a-z.]+$'
             order: 1
-        apiBaseUrl:
+        apiDomain:
             type: string
-            title: API Base Url
-            description: The base url to where you will access your API
+            title: API Domain
+            description: The domain to where you will access your API
             example: us.api.blizzard.com
             pattern: '^[a-z.]+$'
             prefix: https://
@@ -1794,7 +1794,7 @@ contentstack:
     token_url: https://${connectionConfig.subdomain}.contentstack.com/apps-api/apps/token
     docs: https://docs.nango.dev/integrations/all/contentstack
     proxy:
-        base_url: https://${connectionConfig.apiBaseUrl}
+        base_url: https://${connectionConfig.apiDomain}
     connection_config:
         subdomain:
             type: string
@@ -1810,10 +1810,10 @@ contentstack:
             title: App ID
             description: The app ID of your Contentstack account
             order: 1
-        apiBaseUrl:
+        apiDomain:
             type: string
-            title: API Base Url
-            description: The base URL to where you will access your API
+            title: API Domain
+            description: The domain to where you will access your API
             pattern: '^[a-z0-9_-]+$'
             example: eu-api.contentstack.com
             prefix: https://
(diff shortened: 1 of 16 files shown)


hassan254-prog:
The remaining provider would be the `unauthenticated` provider and the `proxy.base_url` validations should now pass.  I think we can skip this in the validation script itself.

packages/shared/providers.yaml
```diff
@@ -505,6 +513,8 @@ auth0:
     refresh_params:
         grant_type: refresh_token
     docs: https://docs.nango.dev/integrations/all/auth0
+    proxy:
+        base_url: https://${connectionConfig.subdomain}
```
bodinsamuel: `https://${connectionConfig.subdomain}.auth0.com`?

packages/shared/providers.yaml
```diff
@@ -727,6 +747,8 @@ battlenet:
         response_type: code
     token_params:
         grant_type: authorization_code
+    proxy:
+        base_url: https://${connectionConfig.apiBaseUrl}
```
bodinsamuel: could we always put `.api.blizzard.com`?

packages/shared/providers.yaml
```diff
@@ -1759,6 +1793,8 @@ contentstack:
     authorization_url: https://${connectionConfig.subdomain}.contentstack.com/apps/${connectionConfig.appId}/authorize
     token_url: https://${connectionConfig.subdomain}.contentstack.com/apps-api/apps/token
     docs: https://docs.nango.dev/integrations/all/contentstack
+    proxy:
+        base_url: https://${connectionConfig.apiBaseUrl}
```
bodinsamuel: could we always put `.contentstack.com`?

packages/shared/providers.yaml
```diff
@@ -6821,7 +6910,17 @@ yahoo:
     auth_mode: OAUTH2
     authorization_url: https://api.login.yahoo.com/oauth2/request_auth
     token_url: https://api.login.yahoo.com/oauth2/get_token
+    proxy:
+        base_url: https://${connectionConfig.apiBaseUrl}
```
bodinsamuel: Could we always put `.yahooapis.com`?

packages/shared/providers.yaml
hassan254-prog: according to the docs it’s `https://${connectionConfig.subdomain}`  😅 
Screenshot 2024-12-12 at 00 15 34
They share the same with the authorize and token urls

packages/shared/providers.yaml
hassan254-prog: It may hugely vary depending on the region. For regions outside of China, the base url follows the format `{region}.api.blizzard.com`, where `{region}` can be `us`, `eu`, `kr`, or `tw`. For China, the  api base url is `gateway.battlenet.com.cn`.

packages/shared/providers.yaml
hassan254-prog: This may also hugely vary, for Knowledge vault api we can have `https://gcp-na-ai.contentstack.com/brand-kits`, while for brand kit management api we can have `https://gcp-na-brand-kits-api.contentstack.com`

packages/shared/providers.yaml
hassan254-prog: We have 2 apis for yahoo, Fantacy Sports API with `fantasysports.yahooapis.com` and `api.login.yahoo.com` for the `sign in with Yahoo` that uses the same oauth2 currently supported.

packages/shared/providers.yaml
bodinsamuel: but we put `https://${connectionConfig.subdomain}.auth0.com` in the token_url

packages/shared/providers.yaml
bodinsamuel: I see, can it be named `apiDomain` then or you also expect some paths?

packages/shared/providers.yaml
bodinsamuel: can it be `apiDomain` then?

bodinsamuel: we should probably have a way to tell customers that this is their responsibility not the user'

bodinsamuel (approved):

packages/shared/providers.yaml
hassan254-prog: I have updated this with the latest [commit](https://github.com/NangoHQ/nango/pull/3155/commits/5114e7cb33009b4b0ff2259cc8aec1ec949d3073). Yes after the API domain we can now have the endpoint url `
/data/wow/achievement/index`

packages/shared/providers.yaml
hassan254-prog: I have updated this with the latest [commit](https://github.com/NangoHQ/nango/pull/3155/commits/5114e7cb33009b4b0ff2259cc8aec1ec949d3073)

packages/shared/providers.yaml
hassan254-prog: Makes sense, to have the suffix `auth0.com`, I have updated with the latest commit.

hassan254-prog:
I’m not sure how we will approach this. Telling customers that it’s their responsibility, not the end user's—similar to what we have with Harvest. How can we make it clear that they should have the `appDetails` configured? Is having this information on the specific provider's docs pages sufficient?

bodinsamuel (approved):

khaliqgant (approved):

PR #3156, opened 2024-12-11
feat(docs): Webflow sync

nalanj:
This PR adds a script to sync data in our repo over to webflow. We'll use a github action to run the sync on each commit to `master`. I'll set that action up in a future PR.

See https://linear.app/nango/issue/NAN-2318/backfill-api-catalog-in-webflow-cms

## How I tested it:

- Get the webflow token from 1Password: search for "Webflow API Token"
- `WEBFLOW_CMS_API_TOKEN=<TOKEN> node scripts/webflow-api-sync.mjs`


scripts/webflow-api-sync.ts
```diff
@@ -0,0 +1,180 @@
...
+import { setTimeout } from 'node:timers/promises';
+import util from 'node:util';
+import type { CollectionItem, CollectionItemList } from 'webflow-api/api';
+
+const rateLimitSleep = 1000;
+
+const apiCollectionId = '66f2c79a523a1055b3da7339';
+const apiCategoriesId = '6751672fcfb952c066685384';
+const siteId = '63c092e946f9b71ff6874169';
+const domainIds = ['63d392e62ba46692fb9e082e', '63d392e62ba4666ba69e082d'];
```
bodinsamuel, line 15: I assume those are public ids?

scripts/webflow-api-sync.ts
```diff
@@ -0,0 +1,180 @@
...
+
+const providersPath = 'packages/shared/providers.yaml';
+// eslint-disable-next-line import/no-named-as-default-member, @typescript-eslint/no-explicit-any
+const providers: any = yaml.load(await fs.readFile(providersPath, 'utf8'));
+
+const docsPath = 'docs-v2/integrations/all';
+const files = await fs.readdir(docsPath);
+
+// we only need a subset of providers based on how our docs are written
+const neededProviders: Record<string, any> = {};
```
bodinsamuel:
```suggestion
const providers = yaml.load(await fs.readFile(providersPath, 'utf8')) as Record<string, Provider>;

const docsPath = 'docs-v2/integrations/all';
const files = await fs.readdir(docsPath);

// we only need a subset of providers based on how our docs are written
const neededProviders: Record<string, Provider> = {};
```

scripts/webflow-api-sync.ts
```diff
@@ -0,0 +1,180 @@
...
+        const providerLine = lines.find((line) => line.startsWith('provider: '));
+        if (!providerLine) {
+            throw new Error(`Unable to find provider entry in ${file}`);
+        }
+
+        const provider = providerLine.split('provider: ')[1]?.trim();
+        if (!provider) {
+            throw new Error(`Unable to parse provider in ${file}`);
+        }
+        neededProviders[provider] = { ...providers[provider], url: `https://docs.nango.dev/integrations/all/${path.basename(file, '.mdx')}` };
```
bodinsamuel: Not sure why you need url when you have `docs` inside provider?

scripts/webflow-api-sync.ts
```diff
@@ -0,0 +1,180 @@
...
+        const lines = content.split('\n');
+
+        // find the integration line
+        const providerLine = lines.find((line) => line.startsWith('provider: '));
+        if (!providerLine) {
+            throw new Error(`Unable to find provider entry in ${file}`);
+        }
+
+        const provider = providerLine.split('provider: ')[1]?.trim();
+        if (!provider) {
```
bodinsamuel: you could replace the 3 ifs by `content.match(/^provider: ([a-z-])$/)`

scripts/webflow-api-sync.ts
```diff
@@ -0,0 +1,180 @@
...
+        await webflow.collections.items.deleteItem(apiCollectionId, apiItemsBySlug[toDelete].id);
+        console.log(`Deleted ${toDelete}`);
+        await setTimeout(rateLimitSleep);
+    } catch (err) {
+        console.error(`Failed to delete ${toDelete}`, err);
+        process.exit(1);
+    }
+}
+
+await webflow.sites.publish(siteId, { customDomains: domainIds });
```
bodinsamuel, line 182: Every operations (deleteItem, updateItem)  are pending until published?

bodinsamuel (approved):

scripts/webflow-api-sync.ts
nalanj: Yes - webflow kind of keeps a next state and then when you publish it posts whatever has changed since the last publish.

scripts/webflow-api-sync.ts
nalanj: I wouldn't classify them as secrets since they're just keys within webflow, but I also wouldn't say they're public per-se, since they don't get published anywhere.

scripts/webflow-api-sync.ts
nalanj: Since it's based on human input, docs inside of provider has a higher likelihood of being incorrect than the structure of the repo that we just read.

bodinsamuel: not sure I get your reply, but just to clarify the `docs` value is already checked by another linter, so it's safe

nalanj: Oh, I didn't realize that the other linter checked for broken links. I'll change it.

nalanj: Fixed in 01615c660

scripts/webflow-api-sync.ts
nalanj: Fixed in 01615c660

scripts/webflow-api-sync.ts
nalanj: Fixed in 01615c660

TBonnin (approved):

PR #3158, opened 2024-12-12
feat(fleet): add node config overrides

TBonnin:
Some customers are assigned a bigger runner instance. 

We discussed several options on Slack and I went for the pg table option

This commit adds a node_config_overrides table that allow overriding the default node config if an entry exists in this table for a given routingId.
For now there is no tooling to add/update/remove entries in this table. At first we'll update it by hand.
This mechanism can also be used to rollout a single runner for testing before rolling out all the runners, something I currently do by updating the image tag and redeploying my test runner before hitting the deploy runner button. From now it would mean adding an entry in the table.

# How to test
1. Run Nango locally with `flag:fleet:global` enabled in Redis
2. Add and/or update an entry in the `nango_runners.node_config_overrides` table
3. check that current runner is rolled out and created with correct override


packages/fleet/lib/fleet.ts
```diff
@@ -53,7 +57,17 @@ export class Fleet {
...
     public async rollout(commitId: CommitHash): Promise<Result<Deployment>> {
-        return deployments.create(this.dbClient.db, commitId);
+        return this.dbClient.db.transaction(async (trx) => {
+            const deployment = await deployments.create(trx, commitId);
+            if (deployment.isErr()) {
+                throw deployment.error;
+            }
+
+            // rolling out cancels all nodeConfigOverrides images
+            await nodeConfigOverrides.setForAll(trx, { image: this.nodeProvider.defaultNodeConfig.image });
```
TBonnin: I think it is better not to have long running image override. We can use it to test on a single runner before rolling out all runners but the override would be short leaved and be removed next time the runners are rolled out.

packages/fleet/lib/models/deployments.ts
```diff
@@ -6,7 +6,7 @@ import { FleetError } from '../utils/errors.js';
 
 export const DEPLOYMENTS_TABLE = 'deployments';
 
-export interface DBDeployment {
+interface DBDeployment {
```
TBonnin, line 9: unrelated. some types/functions don't need to be visible outside of their file

packages/fleet/lib/db/migrations/20241222195951_create_node_config_overrides.ts
```diff
@@ -0,0 +1,19 @@
+import type { Knex } from 'knex';
+import { NODE_CONFIG_OVERRIDES_TABLE } from '../../models/node_config_overrides.js';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.raw(`
+            CREATE TABLE ${NODE_CONFIG_OVERRIDES_TABLE} (
+                id SERIAL PRIMARY KEY,
```
bodinsamuel: Note: serial is not "recommended" anymore

packages/fleet/lib/db/migrations/20241222195951_create_node_config_overrides.ts
```diff
@@ -0,0 +1,19 @@
...
+                routing_id varchar(255) NOT NULL UNIQUE,
+                image varchar(255) NOT NULL,
+                cpu_milli int NOT NULL,
+                memory_mb int NOT NULL,
+                storage_mb int NOT NULL,
```
bodinsamuel, line 12: maybe when switching to another cloud provider we won't be able to fill storage, but I guess we can put a default value 0

packages/fleet/lib/models/node_config_overrides.ts
```diff
@@ -0,0 +1,153 @@
+import type knex from 'knex';
+import type { Result } from '@nangohq/utils';
+import { Ok, Err } from '@nangohq/utils';
+import { FleetError } from '../utils/errors.js';
+import type { NodeConfigOverride } from '../types.js';
+import type { RoutingId } from '@nangohq/types';
+
+export const NODE_CONFIG_OVERRIDES_TABLE = 'node_config_overrides';
+
+export interface DBNodeConfigOverride {
```
bodinsamuel: Following your own comment :D
```suggestion
interface DBNodeConfigOverride {
```

packages/fleet/lib/models/node_config_overrides.ts
```diff
@@ -0,0 +1,153 @@
...
+        for (const nodeConfigOverride of nodeConfigOverrides) {
+            nodeConfigOverridesMap.set(nodeConfigOverride.routing_id, DBNodeConfigOverride.from(nodeConfigOverride));
+        }
+        return Ok(nodeConfigOverridesMap);
+    } catch (e) {
+        return Err(new FleetError('node_config_override_search_failed', { cause: e, context: { params } }));
+    }
+}
+
+export async function setForAll(
```
bodinsamuel: imo Seems like a footgun to be able to update more than just the image

bodinsamuel (approved):

packages/fleet/lib/db/migrations/20241222195951_create_node_config_overrides.ts
TBonnin: what's recommended nowadays then?

packages/fleet/lib/models/node_config_overrides.ts
TBonnin: I think it could be handy but I see your point. I will change

packages/fleet/lib/db/migrations/20241222195951_create_node_config_overrides.ts
TBonnin: storage might end up being a useless column but I am gonna keep it for now. We can reassess after moving to another cloud provider

packages/fleet/lib/db/migrations/20241222195951_create_node_config_overrides.ts
TBonnin: got it. `GENERATED AS IDENTITY`

PR #3159, opened 2024-12-12
fix(endUser): make email optional

bodinsamuel:
## Changes

All according to plan 🫣

Fixes https://linear.app/nango/issue/NAN-2371/end-user-change-email-to-be-optional

- Make email optional 
Some customers do not have email in their system and even if they could put garbage it's confusing enough.


packages/webapp/src/pages/Connection/List.tsx
```diff
@@ -58,18 +59,11 @@ const columns: ColumnDef<ApiConnectionSimple>[] = [
                 <div className="flex gap-3 items-center">
                     <AvatarOrganization
                         email={data.endUser?.email ? data.endUser.email : null}
-                        displayName={data.endUser ? data.endUser.displayName || data.endUser.email : data.connection_id}
+                        displayName={data.endUser?.displayName || data.endUser?.email || data.connection_id}
```
nalanj: Would it make sense to factor out a `connectionDisplayName` function and replace a few cases of basically the same conditions?

packages/server/lib/controllers/connect/postSessions.integration.test.ts
```diff
@@ -62,10 +62,7 @@ describe(`POST ${endpoint}`, () => {
         expect(res.json).toStrictEqual({
```
nalanj, line 62: Should test name be changed up at [line 51](https://github.com/NangoHQ/nango/pull/3159/files#diff-58987026fb92134e7b129490ae17232672346cd946c11896a962e5bbd0b73d81R51)

nalanj (approved):
Just left a couple of small comments. Looks good!

packages/server/lib/controllers/connect/postSessions.integration.test.ts
bodinsamuel: good catch

PR #3160, opened 2024-12-12
fix(harvest): default harvest app details

khaliqgant:
Harvest requires an app details header to be sent. This can be set programmatically but should not be set by the end user or even be exposed to the end user. This PR replaces https://github.com/NangoHQ/nango/pull/3148

### DIFF
--- a/packages/shared/providers.yaml
+++ b/packages/shared/providers.yaml
@@ -3374,7 +3374,7 @@ harvest:
             type: string
             title: App Details
             description: The details of your app
-            hidden: true
+            automated: true
 
 health-gorilla:
     display_name: Health Gorilla
--- a/packages/shared/providers.yaml
+++ b/packages/shared/providers.yaml
@@ -3364,7 +3364,7 @@ harvest:
         grant_type: refresh_token
     proxy:
         headers:
-            user-agent: ${connectionConfig.appDetails}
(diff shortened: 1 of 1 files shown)


bodinsamuel (approved):

bodinsamuel:
Note the logic to parse `||` is still in #3148

khaliqgant:
> Note the logic to parse `||` is still in #3148

That logic was added in with https://github.com/NangoHQ/nango/pull/1873

bodinsamuel:
For base url not user agent otherwise I don't think @hassan254-prog would have bothered to add a new logic and tests

hassan254-prog:
Yes the one in https://github.com/NangoHQ/nango/pull/1873 was for the `base_url` for headers it is still in https://github.com/NangoHQ/nango/pull/3148 , cleaned the pr up to now accommodate current PR changes.

khaliqgant:
> For base url not user agent otherwise I don't think @hassan254-prog would have bothered to add a new logic and tests

Hmm, yeah, 

> Yes the one in #1873 was for the `base_url` for headers it is still in #3148 , cleaned the pr up to now accommodate current PR changes.

Ok great, thanks for the quick reaction Kelvin and for pointing out Samuel. Will close this PR in favor of the edited #3148

PR #3161, opened 2024-12-12
chore: call /rollout endpoint in deploy github action

TBonnin:
Calling the /rollout endpoint when deploying runner via github action.
I've tested in staging: https://github.com/NangoHQ/nango/actions/runs/12300905315

### DIFF
--- a/scripts/deploy/runners.bash
+++ b/scripts/deploy/runners.bash
@@ -29,7 +29,7 @@ while true; do
       continue
     fi
     name=$(echo "$item" | jq -r '.service.name' 2>/dev/null)
-    if [[ "$name" != "$ENVIRONMENT-runner-"* ]]; then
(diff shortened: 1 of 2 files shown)


nalanj (approved):
:shipit:

scripts/deploy/runners.bash
```diff
@@ -29,7 +29,7 @@ while true; do
       continue
     fi
     name=$(echo "$item" | jq -r '.service.name' 2>/dev/null)
-    if [[ "$name" != "$ENVIRONMENT-runner-"* ]]; then
+    if [[ ! "$name" =~ ^$ENVIRONMENT-runner-account-(default|[0-9]+)$ ]]; then
```
TBonnin, line 32: old runners are following this format `production-runner-account-123`, new runners are going to be `production-runner-account-123-N`. We don't want the script to deploy new runners

PR #3162, opened 2024-12-12
fix(docs): Fix undefined in endpoint description

nalanj:
Noticed that endpoints sometimes had no description, and it was being rendered as `undefined`.

## How I tested

- Run `npm run docs` and look at the docs.


bodinsamuel (approved):
Description should probably be required in the integrations templates repo

PR #3163, opened 2024-12-12
chore(fleet): add tracing around supervisor plan/execute

TBonnin:
also adding a try/catch around getRunner logic

Ex: https://us3.datadoghq.com/apm/traces?query=env%3Astaging%20service%3Anango-jobs%20resource_name%3Afleet.supervisor.operation.%2A&agg_m=count&agg_m_source=base&agg_t=count&cols=service%2Cresource_name%2C%40duration%2C%40http.method%2C%40http.status_code&fromUser=false&graphType=waterfall&historicalData=true&panel_tab=flamegraph&query_translation_version=v0&shouldShowLegend=true&sort=time&spanID=2784596772656159614&spanType=all&spanViewType=metadata&storage=hot&timeHint=1734038729494&trace=AwAAAZO8wzMW-LOLIwAAABhBWk84dzBweUFBRGl2WlZhLTB0dUhGeGcAAAAkMDE5M2JjYzMtYTMxMC00NWU1LWFjOTQtM2NkMWY3OGYyNTU3AAAidQ&traceID=675b54c900000000605c5601b90448c3&traceQuery=&view=spans&viz=stream&start=1734038206501&end=1734039106501&paused=false

# How to test
The easiest way is to deploy to staging and check the traces in DD

### DIFF
--- a/packages/fleet/lib/supervisor.ts
+++ b/packages/fleet/lib/supervisor.ts
@@ -116,8 +116,7 @@ export class Supervisor {
     }
 
     public async tick(): Promise<Result<void>> {
-        const span = tracer.startSpan('fleet.supervisor.tick');
-        return tracer.scope().activate(span, async () => {
+        return tracer.trace('fleet.supervisor.tick', async (span) => {
             try {
                 this.tickCancelled = false;
 
@@ -131,8 +130,6 @@ export class Supervisor {
             } catch (err) {
                 span.setTag('error', err);
                 return Err(new FleetError('supervisor_tick_failed', { cause: err }));
-            } finally {
-                span.finish();
             }
         });
     }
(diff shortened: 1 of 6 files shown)


packages/fleet/lib/supervisor.ts
```diff
@@ -125,184 +165,208 @@ export class Supervisor {
...
+                        operationSpan.setTag('error', result.error);
+                        logger.error('Failed to execute operation:', result.error, result.error.cause);
+                    }
+                } finally {
+                    operationSpan.finish();
+                }
             }
+        } finally {
+            span.finish();
         }
```
bodinsamuel: Why not just use a span as callbacks? 

If I remember well (but maybe not) it's not mandatory to call finish() ourself. The try/finally would be unnecessary, would result in the same indentation, and would put the span in the right parent?
Screenshot 2024-12-13 at 10 16 14

packages/fleet/lib/supervisor.ts
```diff
@@ -24,6 +25,39 @@ type Operation =
...
+            case 'FINISHING_TIMEOUT':
+            case 'TERMINATE':
+            case 'REMOVE':
+                return {
+                    operation: o.type,
+                    nodeId: o.node.id
+                };
+        }
+    }
+};
```
bodinsamuel: nit: could be placed elsewhere to limit filesize and incr readability

bodinsamuel (approved):

packages/fleet/lib/supervisor.ts
TBonnin: done in [b9bee1c](https://github.com/NangoHQ/nango/pull/3163/commits/b9bee1cf920bb7bc3df810de0beb76a39e6ab5be)

nalanj (approved):

bodinsamuel (approved):

PR #3165, opened 2024-12-12
fix(runner): set render PORT

TBonnin:
https://render.com/docs/web-services#port-binding
Render default port for new service is 10000 and at startup they detect any app running on another port and switch the primary port to it. However the switch triggers a restart to reconfigure the network which is problematic for us since the runner has already registered as RUNNING and is being potentially sent task
This commit set the PORT explicitely in order to avoid the automatic detection and restart

I have tested in staging. No awkward restart when setting the PORT explicitely

### DIFF
--- a/packages/jobs/lib/runner/render.ts
+++ b/packages/jobs/lib/runner/render.ts
@@ -33,6 +33,7 @@ export const renderNodeProvider: NodeProvider = {
                     plan: getPlan(node)
                 },
                 envVars: [
+                    { key: 'PORT', value: '80' },
                     { key: 'NODE_ENV', value: envs.NODE_ENV },
                     { key: 'NANGO_CLOUD', value: String(envs.NANGO_CLOUD) },
                     { key: 'NODE_OPTIONS', value: `--max-old-space-size=${Math.floor((node.memoryMb / 4) * 3)}` },


packages/jobs/lib/runner/render.ts
```diff
@@ -33,6 +33,7 @@ export const renderNodeProvider: NodeProvider = {
                     plan: getPlan(node)
                 },
                 envVars: [
+                    { key: 'PORT', value: '80' },
```
bodinsamuel, line 36: We need to check if it's compatible with the unified docker image, I exposed 8080

bodinsamuel (approved):

packages/jobs/lib/runner/render.ts
TBonnin: that's for the runner which has its own dockerfile https://github.com/NangoHQ/nango/blob/3f1dac3c55a1fd7917065b9ef01a857c27b53a2c/packages/runner/Dockerfile#L22

bodinsamuel: yes I know but one day it won't, anyway it was just a reminder

TBonnin: I see. You mean when we move runner to use the unified image we must change this value to be 8080 👍

PR #3166, opened 2024-12-12
feat(server): Add instance identifier when logging in cloud

nalanj:
When reviewing logs in Render right now it's difficult to know what instance of server logged what. This adds a random identifier at startup that will help us filter logs to specific instances.

I thought about using hostnames but the hostnames on render are super long and not that helpful, and I'm not 100% sure they're guaranteed to be unique on their infra.

### DIFF
--- a/packages/utils/lib/logger.ts
+++ b/packages/utils/lib/logger.ts
@@ -3,6 +3,7 @@ import winston from 'winston';
 import type { Logform, Logger } from 'winston';
 import colors from '@colors/colors';
 import { isCloud, isEnterprise, isTest } from './environment/detection.js';
+import { nanoid } from 'nanoid';
 
 const SPLAT = Symbol.for('splat');
 const level = process.env['LOG_LEVEL'] ? process.env['LOG_LEVEL'] : isTest ? 'error' : 'info';
@@ -41,10 +42,12 @@ if (!isCloud && !isEnterprise) {
         })
     ];
 } else {
+    const instanceId = isCloud ? nanoid(10) : '';
+
     formatters = [
         winston.format.printf((info) => {
             const splat = info[SPLAT] && info[SPLAT].length > 0 ? JSON.stringify(info[SPLAT]) : '';
-            return `[${info.level.toUpperCase()}]${info['service'] ? ` [${info['service']}] ` : ''}${info.message} ${splat}`;
+            return `[${info.level.toUpperCase()}] ${instanceId}${info['service'] ? ` [${info['service']}] ` : ''}${info.message} ${splat}`;
         })
     ];
 }
--- a/packages/utils/lib/logger.ts
+++ b/packages/utils/lib/logger.ts
@@ -42,12 +42,12 @@ if (!isCloud && !isEnterprise) {
         })
     ];
 } else {
-    const instanceId = isCloud ? nanoid(10) : '';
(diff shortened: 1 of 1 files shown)


TBonnin (approved):
good idea. How does it look like in the logs?

packages/utils/lib/logger.ts
```diff
@@ -41,10 +42,12 @@ if (!isCloud && !isEnterprise) {
         })
     ];
 } else {
+    const instanceId = isCloud ? nanoid(10) : '';
```
TBonnin: we can go smaller I think. 4? 6?

packages/utils/lib/logger.ts
```diff
@@ -41,10 +42,12 @@ if (!isCloud && !isEnterprise) {
...
+
     formatters = [
         winston.format.printf((info) => {
             const splat = info[SPLAT] && info[SPLAT].length > 0 ? JSON.stringify(info[SPLAT]) : '';
-            return `[${info.level.toUpperCase()}]${info['service'] ? ` [${info['service']}] ` : ''}${info.message} ${splat}`;
+            return `[${info.level.toUpperCase()}] ${instanceId}${info['service'] ? ` [${info['service']}] ` : ''}${info.message} ${splat}`;
```
TBonnin: does this add an extra space when not in cloud but no space after when cloud?

nalanj: Oh good catch, I'll fix that.

nalanj:
@TBonnin here's how it looks running on my computer with `NANGO_CLOUD` set:

image

packages/utils/lib/logger.ts
nalanj: Fixed in aac9eb623

packages/utils/lib/logger.ts
nalanj: Set to 6 in aac9eb623

packages/utils/lib/logger.ts
```diff
@@ -41,10 +42,12 @@ if (!isCloud && !isEnterprise) {
         })
     ];
 } else {
+    const instanceId = isCloud ? ` ${nanoid(6)}` : '';
+
     formatters = [
         winston.format.printf((info) => {
             const splat = info[SPLAT] && info[SPLAT].length > 0 ? JSON.stringify(info[SPLAT]) : '';
-            return `[${info.level.toUpperCase()}]${info['service'] ? ` [${info['service']}] ` : ''}${info.message} ${splat}`;
+            return `[${info.level.toUpperCase()}]${instanceId}${info['service'] ? ` [${info['service']}] ` : ''}${info.message} ${splat}`;
```
bodinsamuel, line 50: You have access to `RENDER_INSTANCE_ID` that does exactly what you want 
https://render.com/docs/environment-variables#all-runtimes

bodinsamuel (approved):

packages/utils/lib/logger.ts
nalanj: It's just the hostname, which is `srv-cfifsvha6gdq1k7u46a0-99959bc5b-6mcdc` on one of our current instances. Not sure if we want to dump something like that into every log message we have vs. a 6 character identifier.

bodinsamuel (approved):

PR #3167, opened 2024-12-12
chore: keep only 10 days of deleted/terminated schedules/tasks

TBonnin:
We only keep those for debugging purpose. We haven't needed 30 days of history so far


bodinsamuel (approved):

bodinsamuel (commented):
NB: you could make that an env var

nalanj (approved):

TBonnin:
> NB: you could make that an env var

Noted. 
The effect of changing the value being quite slow to materialize it is less useful to be able to quickly change this value though

PR #3169, opened 2024-12-13
feat(api): GET /connection returns end_user

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2372/list-connections-returns-user-and-org-info

- `GET /connection` (sic) now returns end_user
Customers wants to get this information back, I'll add way to filter in an other PR to keep this one short


nalanj (approved):
🚢 🇮🇹

PR #3170, opened 2024-12-13
feat(api): GET /connection search in end user

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2373/list-connections-ability-to-filter-by-user-andor-org-id

- `GET /connection` search exactly in end-user
Since we are moving from connection_id, this is the simplest way to get "all the connections for my user"


docs-v2/spec.yaml
```diff
@@ -406,7 +406,17 @@ paths:
...
                   schema:
                       type: string
-                  description: Search connections. Will search in connection ID or end user profile.
+                  description: Will partially match connection IDs or end user profiles.
+                - name: endUserId
+                  in: query
+                  schema:
+                      type: string
+                  description: Filter the list of connections based on the given end user id.
+                - name: endUserId
```
nalanj:
```suggestion
                - name: endUserOrganizationId
```

...I think?

nalanj (approved):
Just the one comment, looks good once that's fixed (if I understood it correctly).

PR #3171, opened 2024-12-13
feat(integrations): add support for lastpass

hassan254-prog:
Add support for LastPass

There is no need to have the `proxy.verification` method since the provider returns success 200 even when the verification is wrong. 

Nango will supply header `Authorizaton: Basic (username:password).toString('base64')` since it is of `BASIC` auth_mode but seems the provider ingores this so we don't need to worry about it.


TBonnin:
> There is no need to have the `proxy.verification` method since the provider returns success 200 even when the verification is wrong.

🤣 😭

TBonnin (approved):

khaliqgant (approved):

PR #3172, opened 2024-12-13
fix(server): Drastically reduce memory usage in proxy

nalanj:
This does a couple of things:

- Logs when we have a response body > 50MB just for our knowledge in the logs. Will be interesting to see if it comes up around our OOM issues.
- Switches to building an array of buffers rather than doing tons of string conversions. Should drastically reduce GC churn in there, especially depending on how often our data callback was happening.

Benchmarking locally against a 52MB json file:

Before the update:
Pre-request used heap: 179977928
Post-request used heap: 425877152
Memory from request(ish): 245899224

After the update:
Pre-request used heap: 186414032
Post-request used heap: 339980184
Memory from request(ish): 153566152

## How I tested it:

So, maybe this is janky, maybe not, but:

In `providers.ts` I added a fake provider pointing to `localhost:4000`:

providers.ts, line 19
```
if (providers) {
  providers['__local'] = {
    display_name: 'Local Dev',
    auth_mode: 'NONE',
    proxy: {
      base_url: 'http://localhost:4000'
    },
    docs: 'https://google.com'
  };
}
```

Then I ran an http server on port 4000 with a 52MB json file under `biggie.json`, set up the integration and a connection, and made proxy requests against the dev server with that:

```
curl 'http://localhost:3003/proxy/biggie.json' --header 'Authorization: Bearer 07d4a5e1-cf98-49a2-bf86-7481aa258cab' --header 'Connection-Id: 90b41599-37aa-4ec0-95ec-88c6b546a21e' --header 'Provider-Config-Key: __local'
```


packages/server/lib/controllers/proxy.controller.ts
```diff
@@ -286,13 +286,19 @@ class ProxyController {
...
+        let responseLen = 0;
 
         responseStream.data.on('data', (chunk: Buffer) => {
-            responseData += chunk.toString();
+            responseData.push(chunk);
+            responseLen += chunk.length;
         });
 
         responseStream.data.on('end', async () => {
+            if (responseLen > 50_000_000) {
```
TBonnin: i would log everything > 5-10mb to have a better idea. I would be surprised if we have a lot of those but maybe I am wrong

nalanj: Cool cool, I waffled on the issue but based on what I saw we probably have to have someone doing something massive for this to be the cause of our issue.

packages/server/lib/controllers/proxy.controller.ts
```diff
@@ -301,14 +307,14 @@ class ProxyController {
...
-                res.send(responseData);
+                res.send(Buffer.concat(responseData));
                 await logCtx.success();
                 metrics.increment(metrics.Types.PROXY_SUCCESS);
                 return;
             }
 
             try {
-                const parsedResponse = JSON.parse(responseData);
+                const parsedResponse = JSON.parse(Buffer.concat(responseData).toString());
```
TBonnin, line 317: not sure why we parse the json instead of just forwarding the bytes. maybe @khaliqgant knows

nalanj: Yeah, I figured I would come back to that next week since folks are starting to wind down, but it feels like we could just use that block on line 278 as our handling for everything and it would feasibly shut down any issues here.

TBonnin (approved):

packages/server/lib/controllers/proxy.controller.ts
nalanj: Fixed in dd759bf

packages/server/lib/controllers/proxy.controller.ts
khaliqgant: Really don’t recall. Remember going through a lot of iterations of this to handle edge cases though. Not to be alarmist but would just say to express caution with changes without proper testing. Also if it breaks we’ll probably hear about it pretty quickly since the proxy is pretty core

bodinsamuel: It's not so long ago, but I can't really find for which integration it was. 
I would say the parsing/logging part is purely arbitrary and we could do without but we'll see
https://github.com/NangoHQ/nango/pull/2228/files

bodinsamuel (approved):
Very nice, tested locally all good 🙏🏻

PR #3173, opened 2024-12-13
feat(ci): Add action for generating docs

nalanj:
Updates our docs when a file dependency for them changes. The main dependency here is flows.yaml, but also editing the `provider: [whatever]` line in `integrations/all/*.md` could also require updates in snippet generation.

## How I tested it

You can see my test run and automated commits to this PR. If you want to test it yourself you'll need to make an edit to flows.yaml that causes a docs update and ensure that this branch's name is one of the actions triggers.

### DIFF
--- a/.github/workflows/docs-update.yaml
+++ b/.github/workflows/docs-update.yaml
@@ -1,0 +1,27 @@
+name: Docs Update
+
+on:
+    push:
+        branches:
+            - master
+        paths:
+            - packages/shared/flows.yaml
+            - docs-v2/**/*.md
+    pull_request:
+
+jobs:
+    update-docs-job:
+        runs-on: ubuntu-latest
+        steps:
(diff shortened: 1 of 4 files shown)


bodinsamuel (commented):
So now when when we receive a commit from integration templates we will commit a second time to do that? 😮‍💨
Since we are cloning the repo can we just execute this code in the other repo instead?
https://github.com/NangoHQ/integration-templates/blob/main/.github/workflows/push-flows-to-nango-repo.yaml

nalanj:
@bodinsamuel There's another issue that we collate the data based on the provider line in each integration doc, so a commit on this repo can also cause it to need an update.

nalanj:
Closing in favor of https://github.com/NangoHQ/integration-templates/pull/156

PR #3174, opened 2024-12-13
feat(server): Proposed proxy simplification

nalanj:
I'm not sure if this is feasible, but from reading the code I can't see why it wouldn't be.

## How I tested it:

With an unauthenticated integration:

```
curl -v 'http://localhost:3003/proxy/biggie.json' --header 'Authorization: Bearer [token]' --header 'Connection-Id: e3a835e0-a694-444b-93b7-dc5390779f35' --header 'Provider-Config-Key: unauthenticated' --header 'Base-Url-Override: http://localhost:4000/'
```

### DIFF
--- a/packages/server/lib/controllers/proxy.controller.ts
+++ b/packages/server/lib/controllers/proxy.controller.ts
@@ -20,7 +20,7 @@ import {
     redactHeaders,
     redactURL
 } from '@nangohq/shared';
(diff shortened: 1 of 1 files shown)


nalanj:
This is how the proxy used to work before https://github.com/NangoHQ/nango/pull/2228/files. Need to dig into it a bit more and figure out if we can improve memory even more here.

PR #3177, opened 2024-12-16
fix(api): GET /connection/:id new format

bodinsamuel:
## Changes

Contributes to https://linear.app/nango/issue/NAN-2363/session-token-polish
Fixes https://linear.app/nango/issue/NAN-2396/get-connection-returns-user-and-org-info

- `GET /connection/:id` new format
Turns out we were also leaking iv but even documented it, don't think it had any side-effects since we were returning the credentials unencrypted anyway.

- Add missing OpenAPI schema


docs-v2/reference/api/connection/get.mdx
```diff
@@ -3,43 +3,9 @@ title: 'Get connection & credentials'
...
-    },                      
-    "account_id": 0,                              // ID of your Nango account (Nango Cloud only)
-    "metadata": {                                 // Custom metadata stored by you
-        "myProperty": "yes",
-        "filter": "closed=true"
-    }
-}                              
-  ```
-</ResponseExample>
-
```
bodinsamuel, line 39: Unnecessary with openapi

docs-v2/snippets/generated/ramp/PreBuiltUseCases.mdx
```diff
@@ -1,5 +1,17 @@
...
+
+<Accordion title="Users">
+| Endpoint | Description | Readme |
+| - | - | - |
+| `POST /users` | Creates a user in Ramp | [🔗](https://github.com/NangoHQ/integration-templates/blob/main/integrations/ramp/actions/create-user.md) |
+| `DELETE /users` | Deletes a user in Ramp by id | [🔗](https://github.com/NangoHQ/integration-templates/blob/main/integrations/ramp/actions/disable-user.md) |
+| `GET /users` | Fetches a list of users from Ramp | [🔗](https://github.com/NangoHQ/integration-templates/blob/main/integrations/ramp/actions/users.md) |
+</Accordion>
+
+</AccordionGroup>
```
bodinsamuel: I guess it was missing? 🤔

packages/node-client/lib/types.ts
```diff
@@ -156,21 +158,6 @@ export interface MetadataChangeResponse {
...
-    updated_at: Date;
-    provider_config_key: string;
-    connection_id: string;
-    connection_config: Record<string, string>;
-    environment_id: number;
-    metadata?: Metadata | null;
-    credentials_iv?: string | null;
-    credentials_tag?: string | null;
-    credentials: AllAuthCredentials;
-}
```
bodinsamuel, line 172: We were returning all fieds from the DB, I have removed `end_user_id`, `environment_id` `credentials_iv`, `credentials_tag`

packages/server/lib/controllers/connection/connectionId/getConnection.ts
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+
+    // We get connection one last time to get endUser, errors
+    // This is very unoptimized unfortunately
+    const finalConnections = await connectionService.listConnections({ environmentId: environment.id, connectionId });
+    if (finalConnections.length !== 1 || !finalConnections[0]) {
+        res.status(500).send({ error: { code: 'server_error', message: 'Failed to get connection' } });
+        return;
+    }
```
bodinsamuel, line 101: In this endpoint, if you have to refresh, we query 3-4 times the same connection. We can do better for sure but requires more time to consolidate

packages/shared/lib/services/connection.service.ts
```diff
@@ -384,49 +384,15 @@ class ConnectionService {
...
-
-            await telemetry.log(LogTypes.GET_CONNECTION_FAILURE, error.message, LogActionEnum.AUTH, {
-                environmentId: String(environment_id),
-                connectionId,
-                providerConfigKey,
-                level: 'error'
-            });
-
-            return { success: false, error, response: null };
-        }
```
bodinsamuel, line 417: Unnecessary calls when you have typescript and payload validation

packages/shared/lib/services/connection.service.ts
```diff
@@ -438,67 +404,14 @@ class ConnectionService {
...
-                const creds = credentials;
-                creds.expires_at = creds.expires_at != null ? parseTokenExpirationDate(creds.expires_at) : undefined;
-                connection.credentials = creds;
-            }
+        const credentials = connection.credentials;
+        if (credentials.type && 'expires_at' in credentials) {
+            const creds = credentials;
+            creds.expires_at = creds.expires_at != null ? parseTokenExpirationDate(creds.expires_at) : undefined;
+            connection.credentials = creds;
         }
```
bodinsamuel, line 415: I have re-read this section 20times, I can't understand why we copied the code. But please let me know

packages/shared/lib/sdk/sync.ts
```diff
@@ -312,18 +312,18 @@ interface MetadataChangeResponse {
...
+    created_at: string;
+    updated_at: string;
+    last_fetched_at: string;
+    metadata: Record<string, unknown> | null;
+    provider: string;
+    errors: { type: string; log_id: string }[];
+    end_user: ApiEndUser | null;
     credentials: AuthCredentials;
-    end_user_id: number | null;
 }
```
bodinsamuel, line 327: I haven't removed the whole typings but it would break everybody that does not install nango in a package.json

docs-v2/spec.yaml
```diff
@@ -2239,3 +2238,423 @@ components:
...
+                type:
+                    type: string
+                    enum: [TWO_STEP]
+            required:
+                - raw
+                - type
+        UnauthCredentials:
+            type: object
+            title: Unauthenticated
+            additionalProperties: false
```
TBonnin, line 2574: I feel like it is gonna be hard to keep those definitions in sync with the code

packages/server/lib/controllers/connection/connectionId/getConnection.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (connection && connection.credentials && connection.credentials.type === 'OAUTH2' && !returnRefreshToken) {
+        if (connection.credentials.refresh_token) {
+            delete connection.credentials.refresh_token;
+        }
+
+        if (connection.credentials.raw && connection.credentials.raw['refresh_token']) {
+            const rawCreds = { ...connection.credentials.raw }; // Properties from 'raw' are not mutable so we need to create a new object.
+            // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+            delete rawCreds['refresh_token'];
+            connection.credentials.raw = rawCreds;
```
TBonnin: can we do 
```ts
const { refresh_token, ...rest } = connection.credentials.raw;
connection.credentials.raw = rest;
```

packages/server/lib/controllers/connection/connectionId/getConnection.ts
TBonnin: can we not replace the `getConnection` above with `listConnections` or their types are incompatible?

packages/server/lib/formatters/connection.ts
```diff
@@ -68,3 +76,37 @@ export function connectionToPublicApi({
         created: String(data.created_at)
     };
 }
+
+export function connectionFullToPublicApi({
```
TBonnin, line 80: what about `DBConnectionToPublicAPI` instead of introducing a new term 'full connection`

packages/shared/lib/services/connection.service.ts
TBonnin: cc @hassan254-prog

TBonnin (approved):

packages/shared/lib/services/connection.service.ts
hassan254-prog: I believe it may be a good time to refractor it, like we did with the `upsertConnection` functions.

packages/server/lib/controllers/connection/connectionId/getConnection.ts
nalanj: Fixed in 347269aa9

packages/server/lib/controllers/connection/connectionId/getConnection.ts
nalanj: I'm going to skip this for now for the sake of getting the PR to land. Seems like there's a deeper need to clean this up around all of those reloads.

packages/server/lib/formatters/connection.ts
nalanj: Fixed in 60173b63a

bodinsamuel: I specifically dissociated the two because the listing does not return the credentials since they can be outdated.
Edit: it was the same question here: https://github.com/NangoHQ/nango/pull/2918#discussion_r1824643680

docs-v2/spec.yaml
bodinsamuel: yes, ultimately we need to autogenerate one or the other to have something sustainable because the whole openapi spec is hard to maintain

packages/server/lib/formatters/connection.ts
nalanj: And reverted.

PR #3181, opened 2024-12-16
fix(base-url): fix lastpass base url

hassan254-prog:

All requests from LastPass postman collection have the same `base-url`, since it is not possible to have an empty string as an endpoint value in Nango, this PR fixes that for LastPass.


TBonnin (approved):

khaliqgant:
Could the endpoint just be “/“

hassan254-prog:
> Could the endpoint just be “/“

This throws a `406 not acceptable` error.

PR #3183, opened 2024-12-16
chore(ci): Generate docs dynamically in pre-commit hook

nalanj:
This will run `docs:generate` and then add updates to `docs-v2/snippets/generated` in our pre-commit hooks so that we'll keep docs up to date.


.husky/pre-commit
```diff
@@ -2,3 +2,6 @@
 . "$(dirname -- "$0")/_/husky.sh"
 
 npx lint-staged --allow-empty
+
+npm run docs:generate
+git add ./docs-v2/snippets/generated/
```
bodinsamuel: can we just run it when stuff has changed inside docs-v2?

bodinsamuel (approved):

.husky/pre-commit
nalanj: The task is what changes stuff inside of `docs-v2`.

bodinsamuel: not sure I get what you mean

nalanj: Nothing has changed inside `docs-v2` until we run the command

bodinsamuel: If I read the code correctly: it changes stuff if flows.yaml is modified (should be done in the other repo), a `provider:` has been changed, so it's a change in docs-v2. Am I missing something?

nalanj: Sure sure.

bodinsamuel (approved):

PR #3189, opened 2024-12-17
feat(webhooks): enable all types by default

bodinsamuel:
## Changes 

Fixes https://linear.app/nango/issue/NAN-2251/enable-all-webhook-types-by-default

- Enable all webhooks types by default for new customers


packages/shared/lib/utils/utils.ts
```diff
@@ -147,17 +145,6 @@ export function getGlobalWebhookReceiveUrl() {
...
-export async function getOauthCallbackUrl(environmentId?: number) {
-    const globalCallbackUrl = getGlobalOAuthCallbackUrl();
-
-    if (environmentId != null) {
-        const environment: DBEnvironment | null = await environmentService.getById(environmentId);
-        return environment?.callback_url || globalCallbackUrl;
-    }
-
-    return globalCallbackUrl;
-}
```
bodinsamuel, line 159: Had to move this because of circular import bug

nalanj (approved):

PR #3190, opened 2024-12-17
feat(integrations): Update basic auth configuration field names

hassan254-prog:
- With the Connect UI we now can alias fields to have the proper names for Basic Auth. This PR adds that to some of the providers with BASIC auth mode

- Also added the `proxy.retry-after` header where the provider has mentioned it in their docs

### DIFF
--- a/packages/shared/providers.yaml
+++ b/packages/shared/providers.yaml
@@ -2788,6 +2788,8 @@ freshservice:
     auth_mode: BASIC
     proxy:
         base_url: https://${connectionConfig.subdomain}.freshservice.com
+        retry:
+            after: 'retry-after'
     docs: https://docs.nango.dev/integrations/all/freshservice
     connection_config:
         subdomain:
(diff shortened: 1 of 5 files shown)


packages/shared/providers.yaml
```diff
@@ -4009,6 +4085,17 @@ lever-basic-sandbox:
     proxy:
         base_url: https://api.sandbox.lever.co
     docs: https://docs.nango.dev/integrations/all/lever
+    credentials:
+        username:
+            type: string
+            title: User name
+            description: The API Key of your lever sandbox account
```
TBonnin:
```suggestion
            description: The API Key of your Lever sandbox account
```

TBonnin (approved):

packages/shared/providers.yaml
hassan254-prog: Thanks Thomas, updated with the latest [commit](https://github.com/NangoHQ/nango/pull/3190/commits/1b0d7e01c8c4395b3d0f687cbce67e3932849877)

khaliqgant (approved):

PR #3191, opened 2024-12-17
feat(docs): Generate pre-built tooling table

nalanj:
Generates a snippet for the Pre-built tooling part of each integration's doc.

https://linear.app/nango/issue/NAN-2399/improve-generation-of-api-pre-built-tooling-make-periodic

## How I tested it:
- Run `npm run docs` and review the resulting docs


scripts/docs-gen-snippets.ts
```diff
@@ -34,11 +51,79 @@ for (const file of files) {
 
         await fs.mkdir(snippetPath, { recursive: true });
 
+        let providerConfig: Provider | undefined = providers[provider];
+        if (providerConfig && (providerConfig as any)['alias']) {
+            providerConfig = providers[(providerConfig as any)['alias']];
+        }
```
TBonnin: can we do:
```
const providerConfig = providers[(providerConfig as any)['alias']] || providers[provider]
```

TBonnin (approved):

scripts/docs-gen-snippets.ts
nalanj: Fixed in be6c5efe0

PR #3197, opened 2024-12-18
feat(docs): add lattice connect ui docs

Maina-Francis:
Adds support for Lattice API and Lattice SCIM API.

[EXT-335](https://linear.app/nango/issue/EXT-335/add-support-for-lattice)

I ran npm run docs to verify the docs compile as expected.


khaliqgant (changes requested):
The provider should just be lattice, not lattice-scim. The extra logo isn't needed lattice-scim.svg.

The connect docs should include instructions for creating an integration using the lattice public api and also separate section of docs to use the scim API and how to obtain credentials for both.

khaliqgant (approved):

PR #3198, opened 2024-12-18
fix(server): Fix case of duplicate connectionId with different providers for /connections/:id

nalanj:
Reverts the previous revert and adds test for that test case.

### DIFF
--- a/packages/server/lib/controllers/connection/connectionId/getConnection.integration.test.ts
+++ b/packages/server/lib/controllers/connection/connectionId/getConnection.integration.test.ts
@@ -100,4 +100,56 @@ describe(`GET ${endpoint}`, () => {
             updated_at: expect.toBeIsoDateTimezone()
         });
     });
+
+    it('should get a connection despite another connection with same name on a different provider', async () => {
+        const { env, account } = await seeders.seedAccountEnvAndUser();
+
+        await seeders.createConfigSeed(env, 'algolia', 'algolia');
+        const endUser = await seeders.createEndUser({ environment: env, account });
+        const conn = await seeders.createConnectionSeed(env, 'algolia', endUser, {
+            rawCredentials: { type: 'API_KEY', apiKey: 'test_api_key' },
+            connectionConfig: { APP_ID: 'TEST' }
+        });
+
+        await seeders.createConfigSeed(env, 'google', 'google');
+        await seeders.createConnectionSeed(env, 'google', endUser, {
+            connectionId: conn.connection_id,
+            rawCredentials: { type: 'API_KEY', apiKey: 'test_api_key' },
+            connectionConfig: { APP_ID: 'TEST' }
+        });
+
+        const res = await api.fetch(endpoint, {
+            method: 'GET',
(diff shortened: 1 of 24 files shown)


packages/shared/lib/seeders/connection.seeder.ts
```diff
@@ -24,14 +24,23 @@ export const createConnectionSeeds = async (env: DBEnvironment): Promise<number[
...
+    env: DBEnvironment,
+    provider: string,
+    endUser?: EndUser,
+    rest?: {
+        connectionId?: string;
+        rawCredentials?: AuthCredentials;
+        connectionConfig?: any;
+    }
+): Promise<NangoConnection> => {
+    const name = rest?.connectionId ? rest.connectionId : Math.random().toString(36).substring(7);
```
nalanj, line 37: 🍖 make connectionId configurable through `rest`

packages/server/lib/controllers/connection/connectionId/getConnection.integration.test.ts
```diff
@@ -0,0 +1,155 @@
...
+            id: expect.any(Number),
+            last_fetched_at: expect.toBeIsoDateTimezone(),
+            metadata: null,
+            provider: 'algolia',
+            provider_config_key: 'algolia',
+            updated_at: expect.toBeIsoDateTimezone()
+        });
+    });
+
+    it('should get a connection despite another connection with same name on a different provider', async () => {
```
nalanj, line 104: 🍖 Add test with multiple providers. I can remove the prior test if we want but it was helpful to have a case I could run with the regression and the fix not applied yet.

packages/server/lib/controllers/connection/connectionId/getConnection.ts
```diff
@@ -0,0 +1,111 @@
...
+            // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+
+            const { refresh_token, ...rest } = rawCreds;
+            connection.credentials.raw = rest;
+        }
+    }
+
+    // We get connection one last time to get endUser, errors
+    // This is very unoptimized unfortunately
+    const finalConnections = await connectionService.listConnections({ environmentId: environment.id, connectionId, integrationIds: [providerConfigKey] });
```
nalanj, line 97: 🍖 THE FIX

packages/shared/lib/seeders/connection.seeder.ts
TBonnin: not a huge fan of this pattern. I am ok going with it because this change is time sensitive as long as we follow up immediately with a PR to use a object as parameter

TBonnin (approved):

packages/shared/lib/seeders/connection.seeder.ts
bodinsamuel: Its not a pattern its just like that for historical reason, changing to object will update a lot of tests so it was easier like this (we cant fix everything all the time unfortunately)
Happy to see it changed to object 👌🏻

PR #3200, opened 2024-12-19
fix(server): /connections/:connectionId New Format

nalanj:
This is another attempt to get the new /connections/:connectionId PR to land, with some updates to remove the status code changes the previous attempts included. That should help us see how much we've actually changed.


packages/server/lib/controllers/connection/connectionId/getConnection.ts
```diff
@@ -0,0 +1,113 @@
...
+        environment,
+        connection: connectionRes.response,
+        integration,
+        logContextGetter,
+        instantRefresh: instantRefresh ?? false,
+        onRefreshSuccess: connectionRefreshSuccessHook,
+        onRefreshFailed: connectionRefreshFailedHook
+    });
+    if (credentialResponse.isErr()) {
+        res.status(credentialResponse.error.status).send({
```
nalanj, line 75: 🍖 the core of this update

packages/server/lib/controllers/connection/connectionId/getConnection.ts
```diff
@@ -0,0 +1,114 @@
...
+            const { refresh_token, ...rest } = rawCreds;
+            connection.credentials.raw = rest;
+        }
+    }
+
+    // We get connection one last time to get endUser, errors
+    // This is very unoptimized unfortunately
+    const finalConnections = await connectionService.listConnections({ environmentId: environment.id, connectionId, integrationIds: [providerConfigKey] });
+    if (finalConnections.length !== 1 || !finalConnections[0]) {
+        serverError(res, { code: 'server_error', message: 'Failed to get connection' });
```
nalanj: 🍖 usage of new `serverError` function for wrapping a 500 with a trace error and a log message

packages/server/lib/utils/response-error.ts
```diff
@@ -0,0 +1,24 @@
+import type { Response } from 'express';
```
nalanj: 🍖 new helper function wrapping up sending a 500 with a trace and a log. We technically already log the body on the 500, but it happens after the response which I'm pretty sure is what is making it harder to trace, and I don't love that we output the error so far from where it happens.

packages/server/lib/controllers/connection/connectionId/getConnection.ts
TBonnin: are we sure `credentialResponse.error.status` is always defined, if not let's `|| 5XX`

TBonnin (approved):
3rd time the charm

nalanj:
@TBonnin `NangoError` always has status set, and the `credentialResponse` error type is `NangoError`, so I think we're safe.

PR #3201, opened 2024-12-19
fix(build): Bump docker image versions to 20.18

nalanj:
We needed to update to a more recent base docker image since the ones we were using were no longer supported.

https://linear.app/nango/issue/NAN-2423/fix-docker-containers-high-and-severe-vulnerabilities

## How we'll test

Most likely we'll stage them and make sure everything looks good


TBonnin (approved):
probably worth running in staging for a few hours/day before to push to prod

PR #3203, opened 2024-12-19
fix(logs): correctly truncate logs based on bytes size

TBonnin:
truncation logic was based on characters length which led to long string containing unicode characters being truncated but still be above the size limit in bytes and therefore be rejected by 'persist' with 'entity too large' error. 

This commit modifies the truncation logic to take into account the string size in bytes and correctly handled 4 bytes characters

# how to test
Run a sync with the following code before and after this patch. Doesn't fail anymore
```
const longString = Array.from({ length: 500_000 }, () => '😀').join('');
await nango.log(longString);
```

### DIFF
--- a/packages/utils/lib/json.ts
+++ b/packages/utils/lib/json.ts
@@ -1,5 +1,6 @@
 import safeStringify from 'fast-safe-stringify';
 import truncateJsonPkg from 'truncate-json';
+import { truncateBytes } from './string.js';
 
 export const MAX_LOG_PAYLOAD = 99_000; // in  bytes
 
@@ -37,10 +38,11 @@ export function stringifyAndTruncateValue(value: any, maxSize: number = MAX_LOG_
         return 'undefined';
     }
 
-    let msg = typeof value === 'string' ? value : truncateJsonString(stringifyObject(value), maxSize);
+    const msg = typeof value === 'string' ? value : truncateJsonString(stringifyObject(value), maxSize);
 
-    if (msg && msg.length > maxSize) {
-        msg = `${msg.substring(0, maxSize)}... (truncated)`;
+    const truncated = truncateBytes(msg, maxSize);
+    if (truncated !== msg) {
+        return `${truncated}... (truncated)`;
     }
 
     return msg;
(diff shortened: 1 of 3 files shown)


packages/utils/lib/string.ts
```diff
@@ -20,3 +20,42 @@ export function stringTimingSafeEqual(a: string, b: string): boolean {
...
+    const encoder = new TextEncoder();
+    if (encoder.encode(str).length <= maxBytes) {
+        return str;
+    }
+
+    let left = 0;
+    let right = str.length;
+
+    // Find how many characters fit within the byte limit
+    while (right - left > 1) {
```
nalanj, line 34: Love it!

nalanj (approved):
:ship:

packages/utils/lib/json.ts
```diff
@@ -37,10 +38,11 @@ export function stringifyAndTruncateValue(value: any, maxSize: number = MAX_LOG_
...
     }
 
-    let msg = typeof value === 'string' ? value : truncateJsonString(stringifyObject(value), maxSize);
+    const msg = typeof value === 'string' ? value : truncateJsonString(stringifyObject(value), maxSize);
 
-    if (msg && msg.length > maxSize) {
-        msg = `${msg.substring(0, maxSize)}... (truncated)`;
+    const truncated = truncateBytes(msg, maxSize);
+    if (truncated !== msg) {
+        return `${truncated}... (truncated)`;
```
bodinsamuel, line 45: Note if its truncated at max length then its now above max length again with the added msg

bodinsamuel (approved):
Looks okay, wonder about the perf impact at scale

packages/utils/lib/json.ts
TBonnin: that's how it was before. I imagine that's why the `MAX_LOG_PAYLOAD` is `99_000` when the actual limit is 100k

TBonnin:
> Looks okay, wonder about the perf impact at scale

same but at least since the truncation happens in the runner the cost is only borne by the customer doing the logging.

packages/utils/lib/json.ts
bodinsamuel: Yes indeed 👍🏻

PR #3204, opened 2024-12-19
fix(environment-id): environmentId not required for getEnvironmentVariables call

khaliqgant:
Reported by a customer: https://nangohq.slack.com/archives/C07242ZTXCK/p1734637795666279?thread_ts=1734536836.183739&cid=C07242ZTXCK -- environmnentId isn't required to make the call

### DIFF
--- a/packages/shared/lib/sdk/sync.ts
+++ b/packages/shared/lib/sdk/sync.ts
@@ -801,10 +801,6 @@ export class NangoAction {
     }
 
     public async getEnvironmentVariables(): Promise<EnvironmentVariable[] | null> {
-        if (!this.environmentId) {
-            throw new Error('There is no current environment to get variables from');
-        }
-
         return await this.nango.getEnvironmentVariables();
     }
 


TBonnin (approved):

packages/shared/lib/sdk/sync.ts
```diff
@@ -801,10 +801,6 @@ export class NangoAction {
     }
 
     public async getEnvironmentVariables(): Promise<EnvironmentVariable[] | null> {
-        if (!this.environmentId) {
```
nalanj, line 804: I don't know enough to know for sure, but would it be better for us to return an empty array if there's no environmentId set?

khaliqgant: Well, for a getConnection call environmentId is never going to be set https://docs.nango.dev/reference/api/connection/get -- I believe at some point we did get back an environmentId. In order to get an environment there is no need to have an environment id present either https://docs.nango.dev/reference/api/sync/environment-variables

TBonnin: the environment is deducted from the secretKey that is used to authenticate the call to `getEnvironmentVariables`. As you can see below `this.environmentId` is not even passed to `getEnvironmentVariables`

PR #3208, opened 2024-12-20
feat(verification): adds api verification endpoint priority integrations

Maina-Francis:
Adds api verification endpoint for some priority integrations

[EXT-281](https://linear.app/nango/issue/EXT-281/add-an-api-verification-endpoint-priority-integrations)


packages/shared/providers.yaml
```diff
@@ -431,6 +434,9 @@ asana-scim:
     auth_mode: API_KEY
     proxy:
         base_url: https://app.asana.com/api
+        verification:
```
khaliqgant, line 434: This wouldn't work as a POST would require a body and this API endpoint looks incorrect for SCIM https://developers.asana.com/docs/scim#user-endpoints

khaliqgant (approved):

PR #3211, opened 2024-12-20
feat(stack-trace): [nan-2433] script execution add stack trace

khaliqgant:
Add a stack trace to the compile js file

NAN-2433


packages/cli/lib/services/dryrun.service.ts
```diff
@@ -477,17 +477,16 @@ export class DryRunService {
...
-                const wrappedScript = `
-                    (function() {
-                        var module = { exports: {} };
-                        var exports = module.exports;
-                        ${script}
-                        return module.exports;
-                    })();
+                const wrappedCode = `(function() { var module = { exports: {} }; var exports = module.exports; ${script}
+                    return module.exports;
+                })();
```
TBonnin, line 484: to make it a bit more readable we could also add a `// SCRIPT START` comment in the wrapped code and look for its line to calculate the offset. Up to you

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -604,11 +603,21 @@ export class DryRunService {
...
+                              .map((s) => s.trim())
+                              .slice(0, 5) // max 5 lines
+                        : [];
+
                     return {
                         success: false,
                         error: {
                             type: 'script_internal_error',
                             payload: { name: tmp.name || 'Error', code: tmp.code, message: tmp.message },
+                            ...(stacktrace.length > 0 ? { stacktrace } : {}),
```
TBonnin, line 620: why does it need to be outside of the payload in dryrun?

TBonnin (approved):

packages/cli/lib/services/dryrun.service.ts
khaliqgant: Just to make it more obvious and visible to the user. Plus it reads a bit weird inside because the error is repeated with the payload so imo it’s a bit cleaner to be outside

packages/cli/lib/services/dryrun.service.ts
khaliqgant: Yeah, I’m fine with this for now. In my initial implementation I tried to be clever and calculate the lines based on line breaks but this code isn’t complicated and also I feel like a random comment would be strange.

PR #3222, opened 2024-12-23
fix(sync status): [nan-2141] also only grab enabled syncs

khaliqgant:
Problem: Disabled syncs still show in the sync status table. This is only the case for public templates because custom ones lost their active status. For public templates we also need to filter on enabled.

Ticket: NAN-2141

# Testing
1. Enable a sync that is a public template
2. Kick off a sync run
3. Disable the sync
4. See that the sync result no longer shows in the summary table

@TBonnin might want to specifically take a look at this as I know he's done a lot of work optimizing this query to make sure it is performant as this table gets loaded quite often.

### DIFF
--- a/packages/shared/lib/services/sync/sync.service.ts
+++ b/packages/shared/lib/services/sync/sync.service.ts
@@ -197,7 +197,8 @@ export const getSyncs = async (
             this.on(`${SYNC_CONFIG_TABLE}.sync_name`, `${TABLE}.name`)
                 .andOn(`${SYNC_CONFIG_TABLE}.deleted`, '=', db.knex.raw('FALSE'))
                 .andOn(`${SYNC_CONFIG_TABLE}.active`, '=', db.knex.raw('TRUE'))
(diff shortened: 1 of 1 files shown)


nalanj (commented):
One small bug I noticed back in the connections list is that the badge for a failed sync sticks around.

khaliqgant:
> One small bug I noticed back in the connections list is that the badge for a failed sync sticks around.

Hmm. Can you post a screenshot? Not sure exactly what you mean

nalanj:
@khaliqgant oh oops, forgot to paste it! This one:

Screenshot 2024-12-23 at 2 35 05 PM

khaliqgant:
> @khaliqgant oh oops, forgot to paste it! This one:
> 
> Screenshot 2024-12-23 at 2 35 05 PM

Ah got it, thanks! Tracking in https://linear.app/nango/issue/NAN-2448/disabled-failed-syncs-still-show-in-the-error-notification-for-a as this is a separate issue and should be handled in a different place IMO. Will try and fix Thursday

nalanj (approved):

bodinsamuel (approved):

PR #3224, opened 2024-12-23
fix(sync-status): [nan-2207] add connection_id to response

khaliqgant:
Currently, we can request sync statuses from all connection IDs, but we don't know to which connection they belong to, making them unusable. This adds the `connection_id` to the response:
```
{
    "syncs": [
        {
            "id": "4d8069c7-71d7-4147-b36a-01b7c0327718",
            "connection_id": "u-1",
            "type": "INITIAL",
            "nextScheduledSyncAt": null,
            "name": "fake-data",
            "status": "PAUSED",
            "frequency": "every 5m",
            "recordCount": {
                "Test2": 0
            }
        }
    ]
}
```
Updates docs to reflect the update.

NAN-2207

### DIFF
--- a/docs-v2/reference/sdks/node.mdx
+++ b/docs-v2/reference/sdks/node.mdx
@@ -1002,6 +1002,7 @@ await nango.syncStatus('<INTEGRATION-ID>', ['SYNC_NAME1', 'SYNC_NAME2'], '<CONNE
     "syncs": [
         {
             "id": "<string>",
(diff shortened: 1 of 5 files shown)


nalanj (commented):
Code looks good here, but I'm pretty sure something's janky locally for me because it errors every time I try to access the /sync/status endpoint for any integration that has any syncs.

bodinsamuel (approved):

PR #3226, opened 2024-12-24
fix(json-schema): [nan-1680] parse the json schema to push to the configs table

khaliqgant:
JSON schema was not previously being pushed to the DB so public templates models could not be validated.

NAN-1680


nalanj:
@khaliqgant Code looks good - how would I test this?

khaliqgant:
> @khaliqgant Code looks good - how would I test this?

@bodinsamuel when you're back – what were the symptoms of this issue?

bodinsamuel (approved):

PR #3228, opened 2024-12-26
feat(integrations): add support for Scrape.do

gkhngyk:
What is Scrape.do ?
[Scrape.do](https://scrape.do) is a service that allows you to access raw data before the target website understands that you are sending bot traffic, bypassing the blocking problems you experience while scraping your target website, thanks to its rotating proxy, headless browser and captcha handling mechanisms. All the requests you send are upgraded and compared to the requests of a real user. The firewalls of the target websites have a hard time distinguishing these requests, and your success rate will thus exceed 99%.

With Scrape.do, you can use a datacenter, residential or mobile proxy from the region or country you want, manage the entire process, including headless browsers, according to the needs of the target website, and save you from spending your time on crawling processes.


nalanj (approved):
Looks great! Thank you for the contribution.

PR #3231, opened 2025-01-01
feat(integrations): Readwise APIs

sarimabbas:
Added Readwise integration with support for both Readwise API (v2) and Readwise Reader API (v3)

 - API key authentication for both APIs
 - Documentation
 - Provider logo

The Readwise API (v2) enables access to highlights and books data, while the Reader API (v3) allows saving and fetching documents in Readwise Reader.

- https://readwise.io/api_deets
- https://readwise.io/reader_api


packages/shared/providers.yaml
```diff
@@ -5675,6 +5675,49 @@ rapidapi:
...
+    categories:
+        - productivity
+    auth_mode: API_KEY
+    proxy:
+        base_url: https://readwise.io/api/v3
+        headers:
+            authorization: Token ${apiKey}
+        verification:
+            method: GET
+            endpoint: https://readwise.io/api/v2/auth
```
sarimabbas: the v3 api still uses the v2/auth for verification, is it ok to have the full url like this? or will it lead to a bug i.e. `https://readwise.io/api/v3/https://readwise.io/api/v2/auth` ?

packages/webapp/public/images/template-logos/readwise-reader.svg
bodinsamuel, line 1: can you symlink instead of duplicating please? ☺️

packages/shared/providers.yaml
bodinsamuel: good question, we need to check

docs-v2/integrations/all/readwise.mdx
```diff
@@ -0,0 +1,61 @@
+---
+title: Readwise
+sidebarTitle: Readwise
+provider: readwise
+---
+
+import Overview from "/snippets/overview.mdx";
+import PreBuiltTooling from "/snippets/generated/blandai/PreBuiltTooling.mdx";
+
+import PreBuiltUseCases from "/snippets/generated/blandai/PreBuiltUseCases.mdx"
```
bodinsamuel: @nalanj  what do people need to run to generate those files correctly?

(also it's pointing to blandai instead of readwise)

sarimabbas: Sorry about that, I think `npm run docs` will generate these snippet files. I have done that in the latest commit.

packages/webapp/public/images/template-logos/readwise-reader.svg
sarimabbas: Done!

packages/shared/providers.yaml
sarimabbas: I tried running Nango locally (using `docker compose up`). Readwise works but Readwise Reader does not work

CleanShot 2025-01-11 at 12 52 06@2x

Perhaps someone from the team would be willing to make a change that allows the endpoint to be a full URL or point me to the code, I am happy to take a stab at it

docs-v2/integrations/all/readwise.mdx
nalanj: You can also run `npm run docs:generate` to skip running the docs server and just generate the dynamic parts.

packages/shared/providers.yaml
nalanj: From reading the code I think on the verification block you can set `base_url_override` and it will use that on verify.

See https://github.com/NangoHQ/nango/blob/b61728bf82e27631eb9ec73fd9b6e771ec9f38dd/packages/server/lib/hooks/hooks.ts#L229

bodinsamuel:
Sorry about the delay @sarimabbas , can you check again if that works for you for both integration 👍🏻

khaliqgant:
Handled in https://github.com/NangoHQ/nango/pull/3669

PR #3234, opened 2025-01-02
fix(all): Fix deprecated tsconfig error

nalanj:
Since I joined the team I've had errors in tsconfig files in VSCode because of a deprecated setting. I finally traced it down to being that `importsNotUsedAsValues` is deprecated in typescript but included in the tsconfig that we import from `@tsconfig/node18-strictest-esm/tsconfig.json`.

This disables that config since it's deprecated in TypeScript.


bodinsamuel (approved):
We should upgrade this template (or get rid of it)

PR #3235, opened 2025-01-02
fix(server): Clean up createConnectionSeed signature

nalanj:
The signature for `createConnectionSeed` is overly complicated. This simplifies it to just accept an object as a parameter.

See https://linear.app/nango/issue/NAN-2422/clean-up-signature-for-createconnectionseed


packages/shared/lib/seeders/connection.seeder.ts
```diff
@@ -19,40 +19,52 @@ export const createConnectionSeeds = async (env: DBEnvironment): Promise<number[
             environmentId: env.id,
             accountId: 0
         });
-        connectionIds.push(...result.map((res) => res.connection.id!));
+
+        for (const res of result) {
+            if (!res.connection.id) {
```
bodinsamuel, line 24: Note this is just a type error not an actual error that can happen

bodinsamuel (approved):
Thanks

packages/shared/lib/seeders/connection.seeder.ts
nalanj: Yeah, I was just trying to get eslint to stop fussing at me. I figure if we're going to have warnings on we should structure the code such that they don't show up.

bodinsamuel: I think most importantly we are going to fix the type at some point

PR #3238, opened 2025-01-03
chore: Upgrade vitest

nalanj:
While trying to improve CI performance I figured I should upgrade vitest to see if any benefits came from it. They didn't. But still, this passes and why not take the chance to upgrade.


bodinsamuel (approved):
Thanks, it was on my todo 👌🏻

PR #3239, opened 2025-01-03
chore: Only truncate records table once for those tests

nalanj:
While looking into test perf I found that truncating the records table takes about 500ms per time, or about 3.5s across our tests. This updates ids so it can use the same un-truncated table across all of the tests.

BEFORE:

```
> vitest --config ./vite.integration.config.ts records.integration.test.ts --run


 RUN  v2.1.8 /Users/nalanj/Source/NangoHQ/nango

Starting Elasticsearch...
ES running at http://localhost:55215
 ✓ packages/records/lib/models/records.integration.test.ts (7) 6538ms
   ✓ Records service (7) 6538ms
     ✓ Should write records 808ms
     ✓ Should be able to encrypt and insert 2000 records under 2 seconds 744ms
     ✓ Should delete records 604ms
     ✓ Should retrieve records 549ms
     ✓ Should paginate the records to retrieve all records 759ms
     ✓ Should be able to retrieve 20K records in under 5s with a cursor 1580ms
     ✓ Should return correct added records count when upserting concurrently 575ms

 Test Files  1 passed (1)
      Tests  7 passed (7)
   Start at  19:23:49
   Duration  12.78s (transform 96ms, setup 7ms, collect 431ms, tests 6.54s, environment 0ms, prepare 28ms)
```

AFTER:

```
❯ npm run test:integration -- records.integration.test.ts --run

> @nangohq/nango@0.39.32 test:integration
> vitest --config ./vite.integration.config.ts records.integration.test.ts --run


 RUN  v2.1.8 /Users/nalanj/Source/NangoHQ/nango

Starting Elasticsearch...
ES running at http://localhost:55221
 ✓ packages/records/lib/models/records.integration.test.ts (7) 3007ms
   ✓ Records service (7) 3007ms
     ✓ Should write records
     ✓ Should be able to encrypt and insert 2000 records under 2 seconds
     ✓ Should delete records
     ✓ Should retrieve records
     ✓ Should paginate the records to retrieve all records
     ✓ Should be able to retrieve 20K records in under 5s with a cursor 1034ms
     ✓ Should return correct added records count when upserting concurrently

 Test Files  1 passed (1)
      Tests  7 passed (7)
   Start at  19:25:40
   Duration  8.82s (transform 89ms, setup 6ms, collect 288ms, tests 3.01s, environment 0ms, prepare 24ms)
```


nalanj:
Also here the test task just ran in 6m42s. It's been running over 7m in CI most of the time here lately.

packages/records/lib/models/records.integration.test.ts
```diff
@@ -13,13 +13,13 @@ describe('Records service', () => {
...
 
-    afterEach(async () => {
+    afterAll(async () => {
         await db(RECORDS_TABLE).truncate();
     });
 
     it('Should write records', async () => {
-        const connectionId = 1;
-        const environmentId = 2;
+        const connectionId = Math.floor(Math.random() * 1000000);
```
bodinsamuel, line 21: you can use nanoid (from nango utils) instead for clarity

bodinsamuel (approved):

packages/records/lib/models/records.integration.test.ts
nalanj: I don't think we can - nanoid generates strings and these are both integers.

bodinsamuel: ah yes didn't realised that 👍🏻

PR #3240, opened 2025-01-03
chore: Add option to skip creating an env for account seeder

nalanj:
We call `seedAccountEnvAndUser` a lot. When we call it, it creates the default environments and an additional environment. This PR adds a flag to `accountService.getOrCreateAccount` that lets us bypass creating the default environments. All tests succeed when doing so, but it also shaves quite a few seconds off of the integration test suite:

Before:
```
Duration  162.23s (transform 1.09s, setup 132ms, collect 32.28s, tests 83.09s, environment 3ms, prepare 1.32s)
```

After:
```
Duration  144.34s (transform 1.10s, setup 131ms, collect 33.03s, tests 65.14s, environment 3ms, prepare 1.35s)
```


packages/shared/lib/services/account.service.ts
```diff
@@ -61,7 +61,7 @@ class AccountService {
         return account[0].uuid;
     }
 
-    async getOrCreateAccount(name: string): Promise<DBTeam> {
+    async getOrCreateAccount(name: string, createDefaultEnvironments = true): Promise<DBTeam> {
```
bodinsamuel: Can probably be an object for clarity

packages/shared/lib/seeders/account.seeder.ts
```diff
@@ -3,7 +3,7 @@ import accountService from '../services/account.service.js';
 import type { DBTeam } from '@nangohq/types';
 
 export async function createAccount(): Promise<DBTeam> {
-    const acc = await accountService.getOrCreateAccount(uuid());
+    const acc = await accountService.getOrCreateAccount(uuid(), false);
```
bodinsamuel: I feel like there is no reason for us to use getsert here, we should probably just use a new method `accountService.create()`

bodinsamuel (approved):
Open comments, but looks good

packages/shared/lib/seeders/account.seeder.ts
nalanj: @bodinsamuel What do you think about e5e54fb92?

bodinsamuel: sure

PR #3241, opened 2025-01-03
chore: Use singleFork in vitest

nalanj:
While doing some more test performance work I bumped into the singleFork option. 

https://vitest.dev/guide/cli#pooloptions-forks-singlefork

It runs all tests in a single child process, rather than isolating a child process per test file. It turns out that it reduces our run time pretty drastically:

Before:
```
Duration  163.67s (transform 1.05s, setup 127ms, collect 32.85s, tests 83.95s, environment 3ms, prepare 1.34s)

```

After:
```
Duration  94.92s (transform 1.06s, setup 44ms, collect 6.47s, tests 81.63s, environment 0ms, prepare 35ms)
```


nalanj:
And this one finished in 5m52s in CI, our new record during this process of optimizing test perf.

bodinsamuel (approved):
We should be really careful about global state in that case. I'm sure some integration test will assume memory cache is not there or else. But since to be running fine like this 👌🏻

PR #3250, opened 2025-01-06
fix(datadog-api-key): [nan-2459] prefix with api

khaliqgant:
Datadog's API endpoint should be prefixed with `api`. It works without it for some site parameters but is required for `us1`. See https://docs.datadoghq.com/api/latest/users/?site=us3. Reported by https://nangohq.slack.com/archives/C06FQHQJ97Z/p1735932339354609

NAN-2459


packages/shared/providers.yaml
```diff
@@ -2048,7 +2048,7 @@ datadog:
         - dev-tools
     auth_mode: API_KEY
     proxy:
-        base_url: https://${connectionConfig.siteParameter}/api
+        base_url: https://api.${connectionConfig.siteParameter}/api
```
bodinsamuel, line 2051: are we sure it's not going to break existing integration?
other question is `(ap1\.)` in the regex correct or a typo?

khaliqgant: > other question is (ap1\.) in the regex correct or a typo?

Yes, that is correct. 
image

> are we sure it's not going to break existing integration?

`api` is always included in the base_url so we were incorrect in not including it. I tested with it with our current configuration and it still works as expected.

bodinsamuel (approved):

PR #3253, opened 2025-01-06
chore: Update koala snippet

nalanj:
Updates koala snippet per Robin's request.


packages/webapp/src/App.tsx
```diff
@@ -59,7 +59,10 @@ const App = () => {
...
-                            innerHTML: `!function (t) { if (window.ko) return; window.ko = [], ["identify", "track", "removeListeners", "on", "off", "qualify", "ready"].forEach(function (t) { ko[t] = function () { var n = [].slice.call(arguments); return n.unshift(t), ko.push(n), ko } }); var n = document.createElement("script"); n.async = !0, n.setAttribute("src", "https://cdn.getkoala.com/v1/${globalEnv.publicKoalaKey}/sdk.js"), (document.body || document.head).appendChild(n) }();`
+                            innerHTML: `
+                                window.koalaSettings = { host: 'https://bluegrass.nango.dev' };
+                                !function(t){var k="ko",i=(window.globalKoalaKey=window.globalKoalaKey||k);if(window[i])return;var ko=(window[i]=[]);["identify","track","removeListeners","on","off","qualify","ready"].forEach(function(t){ko[t]=function(){var n=[].slice.call(arguments);return n.unshift(t),ko.push(n),ko}});var n=document.createElement("script");n.async=!0,n.setAttribute("src","https://shoegaze.nango.dev/v1/pk_799998f91877afef040206962ce1af164de9/sdk.js"),(document.body || document.head).appendChild(n)}();
+                            `
```
bodinsamuel, line 65: you should keep `${globalEnv.publicKoalaKey}` and add a way to parametrize the domain I guess

nalanj: Oh yep, missed that in the super long template string there.

nalanj: @bodinsamuel see 6f13eb81c

packages/server/lib/controllers/v1/getEnvJs.ts
```diff
@@ -13,6 +13,8 @@ export const getEnvJs = asyncWrapper<any, any>((_, res) => {
         publicPosthogHost: process.env['PUBLIC_POSTHOG_HOST'] || '',
         publicLogoDevKey: process.env['PUBLIC_LOGODEV_KEY'] || '',
         publicKoalaKey: process.env['PUBLIC_KOALA_KEY'] || '',
```
bodinsamuel: We can remove the key then 👌🏻

bodinsamuel (approved):

packages/server/lib/controllers/v1/getEnvJs.ts
nalanj: Ah yep yep, was thinking I'd loop back through and do a cleanup PR but you're right, won't matter to keep here anyway.

nalanj: Fixed in 03b15257a

PR #3259, opened 2025-01-07
chore: remove sentry

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2111/remove-sentry

- Remove sentry from backend
While it's good to have that, the current implementation was not useful. We might want to reintroduce it later more intentionally.
- Frontend use /env.js for Sentry (and PostHog) so there is no need to compile a different image per environment
This docker image was not used yet so it's mostly reducing noise and CI time


packages/jobs/lib/crons/deleteSyncsData.ts
```diff
@@ -24,7 +23,7 @@ export function deleteSyncsData(): void {
             logger.info('[deleteSyncs] ✅ done');
         } catch (err) {
             const e = new Error('failed_to_hard_delete_syncs_data', { cause: err instanceof Error ? err.message : err });
-            errorManager.report(e, { source: ErrorSourceEnum.PLATFORM }, tracer);
+            errorManager.report(e, { source: ErrorSourceEnum.PLATFORM });
```
bodinsamuel, line 26: Passing tracer was not useful since it's a global object, it was a mistake from me a long time ago

packages/shared/lib/utils/analytics.ts
```diff
@@ -99,12 +101,22 @@ class Analytics {
     packageVersion: string | undefined;
 
     constructor() {
+        const hasTelemetry = process.env['TELEMETRY'] !== 'false' && !isStaging;
+        if (!hasTelemetry) {
+            return;
+        }
+
+        // hardcoded for OSS telemetry
+        const key = process.env['PUBLIC_POSTHOG_KEY'] || 'phc_4S2pWFTyPYT1i7zwC8YYQqABvGgSAzNHubUkdEFvcTl';
```
bodinsamuel, line 110: We might want to change that at some point to log OSS in a different account

packages/shared/lib/utils/error.manager.ts
```diff
@@ -141,41 +58,14 @@ class ErrorManager {
...
-        metadata['originalUrl'] = req.originalUrl;
-        metadata['subdomains'] = req.subdomains;
-        metadata['body'] = req.body;
-        metadata['hostname'] = req.hostname;
-        metadata['method'] = req.method;
-        metadata['params'] = req.params;
-        metadata['query'] = req.query;
-
-        return metadata;
-    }
```
bodinsamuel, line 178: it was useless since it's already documented in upper span

nalanj (approved):
🙌🏻

PR #3260, opened 2025-01-07
fix(webapp): Add new koala endpoints to CSP

nalanj:
The new koala hosts also needed to be added to our csp config

### DIFF
--- a/packages/server/lib/middleware/security.ts
+++ b/packages/server/lib/middleware/security.ts
@@ -24,7 +24,16 @@ export function securityMiddlewares(): RequestHandler[] {
             directives: {
                 defaultSrc: ["'self'", hostPublic, hostApi],
                 childSrc: "'self'",
-                connectSrc: ["'self'", 'https://*.google-analytics.com', 'https://*.sentry.io', hostPublic, hostApi, hostWs.href, 'https://*.posthog.com'],
+                connectSrc: [
+                    "'self'",
+                    'https://*.google-analytics.com',
+                    'https://*.sentry.io',
+                    hostPublic,
+                    hostApi,
+                    hostWs.href,
+                    'https://*.posthog.com',
+                    'https://bluegrass.nango.dev'
+                ],
                 fontSrc: ["'self'", 'https://*.googleapis.com', 'https://*.gstatic.com'],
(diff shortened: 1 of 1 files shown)


packages/server/lib/middleware/security.ts
```diff
@@ -24,7 +24,16 @@ export function securityMiddlewares(): RequestHandler[] {
...
-                connectSrc: ["'self'", 'https://*.google-analytics.com', 'https://*.sentry.io', hostPublic, hostApi, hostWs.href, 'https://*.posthog.com'],
+                connectSrc: [
+                    "'self'",
+                    'https://*.google-analytics.com',
+                    'https://*.sentry.io',
+                    hostPublic,
+                    hostApi,
+                    hostWs.href,
+                    'https://*.posthog.com',
+                    'https://bluegrass.nango.dev'
```
bodinsamuel: *.nango.dev or use the PUBLIC_KOALA_CDN_URL hostname (cf hostWs) 
don't forget the code is shared with opensource so stuff like this should not be hardcoded (even if there is no security problem)

packages/server/lib/middleware/security.ts
```diff
@@ -51,7 +60,8 @@ export function securityMiddlewares(): RequestHandler[] {
                     'https://*.googleapis.com',
                     'https://apis.google.com',
                     'https://*.posthog.com',
-                    'https://www.youtube.com'
+                    'https://www.youtube.com',
+                    'https://shoegaze.nango.dev'
```
bodinsamuel: same

bodinsamuel (approved):

packages/server/lib/middleware/security.ts
nalanj: :+1: I was thinking that but then was also surprised about it opening up to Google and posthog.

bodinsamuel: Right, I guess it was less a question because those are "generic" ones but indeed might already be not perfect

nalanj: See 96c4eb4c3

packages/server/lib/middleware/security.ts
nalanj: See 96c4eb4c3

bodinsamuel (approved):

PR #3261, opened 2025-01-07
fix(ui): environment picker v2

bodinsamuel:
## Changes

Contributes to https://linear.app/nango/issue/NAN-2224/create-environments-in-nango-ui

- New environment picker
Search appears only when you have more than 5 (arbitrary number)

Screenshot 2025-01-07 at 17 04 56


nalanj:
@bodinsamuel ~it's not showing up for me:~

nalanj:
@bodinsamuel figured it out

packages/webapp/src/components/EnvironmentPicker.tsx
```diff
@@ -0,0 +1,92 @@
...
+        let newPath = `/${pathSegments.join('/')}`;
+
+        // If on 'integration' or 'connections' subpages beyond the second level, redirect to their parent page
+        if (pathSegments[1] === 'integrations' && pathSegments.length > 2) {
+            newPath = `/${selected}/integrations`;
+        } else if (pathSegments[1] === 'connections' && pathSegments.length > 2) {
+            newPath = `/${selected}/connections`;
+        }
+
+        navigate(newPath);
```
nalanj, line 43: Should the popover close when selecting an item?

nalanj (approved):
Just one small comment. Looks good!

packages/webapp/src/components/EnvironmentPicker.tsx
bodinsamuel: I found that handy to quickly view stuff from different environments, and when you click the wrong one you don't need to reopen it. Don't have a strong opinion.
@bastienbeurier what do you think

https://github.com/user-attachments/assets/7bc5d599-fc37-4a9f-b0e2-9a041014eb95

bastienbeurier: I have no strong feelings about this; I think both can work. I would say priority to the implementer's opinion :D

PR #3262, opened 2025-01-07
chore: upgrade eslint (+ plugins)

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2473/upgrade-eslint

- Upgrade to eslint 9 and all plugins
It's a non-trivial upgrade since eslint chose to have a massive breaking change for their config format. Plus we were behind many months of updates for the plugins.
I haven't added/deleted rules, the lines that are changed in this PR are either: previous rules catching more stuff or new rules inside recommended plugins.

The only big diff is that I enabled Typescript rules only on ts files which removes some false negative (and makes it slightly faster)


packages/fleet/lib/models/deployments.ts
```diff
@@ -34,7 +34,7 @@ const DBDeployment = {
 
 export async function create(db: knex.Knex, commitId: CommitHash): Promise<Result<Deployment>> {
     try {
-        return db.transaction(async (trx) => {
+        return await db.transaction(async (trx) => {
```
bodinsamuel, line 37: New rule https://typescript-eslint.io/rules/return-await/ state that returning await is recommended and as-fast (or even faster in some case) than not awaiting

nalanj (approved):
:+1: I dream of a future where we have no warnings.

bodinsamuel:
> 👍 I dream of a future where we have no warnings.

Well, we need to fix them first :D

PR #3268, opened 2025-01-08
fix: track_deletes only deletes records from previous jobs

TBonnin:
If records were added via a webhook while the full sync is running we don't want to delete those records.
This commit ensures that only records from jobs older than the current one are deleted when track_deletes=true

### DIFF
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -447,9 +447,9 @@ export async function deleteRecordCount({ connectionId, environmentId, model }: 
     await db.from(RECORD_COUNTS_TABLE).where({ connection_id: connectionId, environment_id: environmentId, model }).del();
 }
 
-// Mark all non-deleted records that don't belong to currentGeneration as deleted
+// Mark all non-deleted records from previous generations as deleted
 // returns the ids of records being deleted
-export async function markNonCurrentGenerationRecordsAsDeleted({
+export async function markPreviousGenerationRecordsAsDeleted({
     connectionId,
     model,
     syncId,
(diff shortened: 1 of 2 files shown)


nalanj (approved):

packages/records/lib/models/records.ts
```diff
@@ -471,9 +471,7 @@ export async function markNonCurrentGenerationRecordsAsDeleted({
                 sync_id: syncId,
                 deleted_at: null
             })
-            .whereNot({
-                sync_job_id: generation
-            })
```
bodinsamuel, line 476: note it can have an impact, I don't know if PG can use the index the same way. Did you check with a dummy query in prod?

bodinsamuel (approved):
Amazing 🙏🏻

packages/records/lib/models/records.ts
TBonnin: query plan is the same:

before 
```
Update on records  (cost=0.42..8.45 rows=1 width=30)
  Update on records_p205 records_1
  ->  Index Scan using records_p205_connection_id_model_id_key on records_p205 records_1  (cost=0.42..8.45 rows=1 width=30)
        Index Cond: ((connection_id = 102093) AND ((model)::text = 'SlackMessage'::text))
        Filter: ((deleted_at IS NULL) AND (sync_job_id <> 124692014) AND (sync_id = 'cb89122a-7db0-471c-871a-87610ad9118b'::uuid))
```

after
```
Update on records  (cost=0.42..8.45 rows=1 width=30)
  Update on records_p205 records_1
  ->  Index Scan using records_p205_connection_id_model_id_key on records_p205 records_1  (cost=0.42..8.45 rows=1 width=30)
        Index Cond: ((connection_id = 102093) AND ((model)::text = 'SlackMessage'::text))
        Filter: ((deleted_at IS NULL) AND (sync_job_id < 124692014) AND (sync_id = 'cb89122a-7db0-471c-871a-87610ad9118b'::uuid))
```

PR #3270, opened 2025-01-08
fix(docker): Fix up docker-compose in root of repo to handle empty vars correctly

nalanj:
Empty strings were being passed through when some variables weren't set, which was causing them to flow as empty values through the system.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -19,29 +19,29 @@ services:
         container_name: nango-server
         platform: linux/amd64
         environment:
-            - NANGO_ENCRYPTION_KEY=${NANGO_ENCRYPTION_KEY}
-            - NANGO_DB_USER=${NANGO_DB_USER}
-            - NANGO_DB_PASSWORD=${NANGO_DB_PASSWORD}
-            - NANGO_DB_HOST=${NANGO_DB_HOST}
-            - NANGO_DB_NAME=${NANGO_DB_NAME}
-            - NANGO_DB_SSL=${NANGO_DB_SSL}
-            - NANGO_DB_ADDITIONAL_SCHEMAS=${NANGO_DB_ADDITIONAL_SCHEMAS}
-            - NANGO_DB_POOL_MIN=${NANGO_DB_POOL_MIN}
-            - NANGO_DB_POOL_MAX=${NANGO_DB_POOL_MAX}
+            - NANGO_ENCRYPTION_KEY
+            - NANGO_DB_USER
+            - NANGO_DB_PASSWORD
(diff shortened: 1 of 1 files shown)


docker-compose.yaml
```diff
@@ -19,29 +19,29 @@ services:
...
-            - NANGO_DB_HOST=${NANGO_DB_HOST}
-            - NANGO_DB_NAME=${NANGO_DB_NAME}
-            - NANGO_DB_SSL=${NANGO_DB_SSL}
-            - NANGO_DB_ADDITIONAL_SCHEMAS=${NANGO_DB_ADDITIONAL_SCHEMAS}
-            - NANGO_DB_POOL_MIN=${NANGO_DB_POOL_MIN}
-            - NANGO_DB_POOL_MAX=${NANGO_DB_POOL_MAX}
+            - NANGO_ENCRYPTION_KEY
+            - NANGO_DB_USER
+            - NANGO_DB_PASSWORD
+            - NANGO_DB_HOST=${NANGO_DB_HOST:-nango-db}
```
nalanj, line 25: I had issues with connections failing without setting this to a default, so I went ahead and just pointed it at the docker hostname for the db.

TBonnin: we could do the same for the other NANGO_DB env vars and set the default. Up to you

TBonnin (approved):
this is changing the current behavior. I think it makes sense not to set the env var if they don't exist on the host but just in case, what the worst thing that can happen with this change?

nalanj:
@TBonnin barring some really bizarre programming that I haven't seen, I'm pretty sure the worst case is that we'd have an env var go through as unset rather than an empty string. From what I saw most of the places where we check for empty strings are just `![envVar]` kind of checks so I don't think it should cause any damage to anything. Can you think of anywhere that we would behave drastically different in the case of an unset variable rather than an empty string?

harryadwani:
> Empty strings were being passed through when some variables weren't set, which was causing them to flow as empty values through the system.

Thanks for looking into this. Not sure if this is relevant here, I passed the mailgun api key in vars, but didn't get the verification email. I proceeded anyways by getting the verification url from logs.

` 18:23:53.252 info (Server) POST /api/v1/account/resend-verification-email/by-email -> 200 
nango-server  | 18:25:44.073 warning (Server) POST /api/v1/account/signin -> 400 '{"error":{"code":"email_not_verified"}}'
nango-server  | 18:25:47.180 info (Server.EmailClient) Email client not configured 
nango-server  | 18:25:47.181 info (Server.EmailClient) The following email would have been sent: `

PR #3271, opened 2025-01-09
feat(integrations): add airtable scim support

viictoo:
This PR updates the provider and docs for Airtable SCIM support


packages/shared/providers.yaml
```diff
@@ -198,6 +198,13 @@ airtable:
     webhook_routing_script: airtableWebhookRouting
     docs: https://docs.nango.dev/integrations/all/airtable
 
+airtable-scim:
+    display_name: Airtable (SCIM API)
+    alias: airtable
+    proxy:
+        base_url: https://airtable.com/scim/
+    docs: https://docs.nango.dev/integrations/all/airtable
```
bodinsamuel: I think SCIM needs an auth_mode `API_KEY` and some headers configuration.
Can you confirm @hassan254-prog ? ☺️

hassan254-prog: From the docs, the provider accepts a token which can also be generated via the `Oauth flow` with a specific scope, i.e for list-groups endpoint the scope is `enterprise.scim.usersAndGroups:manage`, 
Screenshot 2025-01-09 at 16 59 00
I think the best way to be sure is to test the actual token generated via the Oauth flow against a simple SCIM endpoint. 🤔

khaliqgant (approved):

PR #3273, opened 2025-01-09
feat(api): POST /environments

bodinsamuel:
## Changes

Contributes to https://linear.app/nango/issue/NAN-2224/create-environments-in-nango-ui

- Add new endpoint `POST /environments`
Tried to keep this short, the UI will come later. 

## 🧪 Tests

- Go to dash
```js
var t = await fetch('/api/v1/environments', { method: 'POST', headers: { 'content-type': 'application/json' }, body: JSON.stringify({name: 'top'})});
```


packages/server/lib/controllers/v1/environment/postEnvironment.ts
```diff
@@ -0,0 +1,51 @@
...
+        return;
+    }
+
+    const body: PostEnvironment['Body'] = valBody.data;
+
+    const accountId = res.locals.user.account_id;
+
+    const exists = await environmentService.getAccountAndEnvironment({ accountId, envName: body.name });
+    if (exists) {
+        res.status(400).send({ error: { code: 'invalid_body', message: 'Environment already exists' } });
```
TBonnin: should we return 409 Conflict?

packages/shared/lib/services/environment.service.ts
```diff
@@ -224,27 +224,22 @@ class EnvironmentService {
...
-            await db.knex.from<DBEnvironment>(TABLE).where({ id: environmentId }).update(encryptedEnvironment);
-
-            const env = encryptionManager.decryptEnvironment(encryptedEnvironment);
-            return env;
+        if (result.length <= 0 || !result[0]) {
+            return null;
         }
 
-        return null;
+        const environment = result[0];
```
TBonnin: can you do `const [environment] = await db.knex...` above to simplify the `if` condition and avoid this assignment?

packages/shared/lib/services/environment.service.ts
```diff
@@ -224,27 +224,22 @@ class EnvironmentService {
...
+        const encryptedEnvironment = await encryptionManager.encryptEnvironment({
+            ...environment,
+            secret_key_hashed: await hashSecretKey(environment.secret_key)
+        });
+        await db.knex.from<DBEnvironment>(TABLE).where({ id: environment.id }).update(encryptedEnvironment);
+
+        const env = encryptionManager.decryptEnvironment(encryptedEnvironment);
+        return env;
```
TBonnin, line 241: shoud we try catch and handle any unexpected exeception?

packages/server/lib/controllers/v1/environment/postEnvironment.ts
```diff
@@ -0,0 +1,51 @@
...
+        res.status(500).send({ error: { code: 'server_error', message: 'Failed to create environment' } });
+        return;
+    }
+
+    await externalWebhookService.update(created.id, {
+        alwaysSendWebhook: true,
+        sendAuthWebhook: true,
+        sendRefreshFailedWebhook: true,
+        sendSyncFailedWebhook: true
+    });
```
TBonnin, line 48: can this fail? what should we do if it does?

TBonnin (approved):

packages/server/lib/controllers/v1/environment/postEnvironment.ts
bodinsamuel: good point

packages/shared/lib/services/environment.service.ts
bodinsamuel: we can but that would mean changing the method signature and updating all use of it. I don't remember did we decided what we do with pure sql calls, let them throw or use ok/err?

packages/server/lib/controllers/v1/environment/postEnvironment.ts
bodinsamuel: I guess it can fail, like the previous message it will throw and return a 500, which is not very different than catching it and returning a 500 imo

packages/shared/lib/services/environment.service.ts
TBonnin: I prefer explicit `Result`, always, but yeah I understand here you need to be backward compatible.

packages/server/lib/controllers/v1/environment/postEnvironment.ts
TBonnin: isn't the environment already created though, even when the webhook service function fails? Should the request be considered a failure then?

bodinsamuel: Yeah indeed. In that case, instead of throwing or return a fake 200, we should actually use transactions but like in many endpoints it's not the case.
We probably need another initiative to find a good way to do that (e.g: how to pass the transaction param, naming, etc.) so we can move forward consistently.

bodinsamuel:
I'll merge like this, but I do agree we are still in a suboptimal situation in term of code quality

PR #3275, opened 2025-01-09
fix(webapp): remove the need for env

bodinsamuel:
## Changes 

- Remove the need to specify the env in the webapp at build time
`/env.js` gives us the missing feature flag, there wasn't much to remove. I'm keeping the docker image compilation because I need to update the deploy too but we are (hopefully) one PR away to remove entreprise, prod and staging compilation

## 🧪 Tests

- Go to dash
- Check that slack integration is working in env settings and code snippets in integrations > endpoint are showing the protected key

> tested in staging


TBonnin (approved):
very excited for next PR

PR #3279, opened 2025-01-09
feat(server, orchestrator, jobs, runner): More descriptive action errors

nalanj:
This adds richer errors when an action is triggered and responds with a 424 when an upstream response is the cause of an error.

Before:
```
❯ curl --request POST --url http://localhost:3000/action/trigger --header "Authorization: Bearer fe212ac9-0832-4210-b868-400db33b3863" --header "Connection-Id: 612d3d37-0a6c-40a5-96f2-f65df6e8fd98" --header "Provider-Config-Key: unauthenticated" --header "Content-Type: application/json" --data '{"action_name": "rate-limited"}'
{"error":{"message":"An error occurred during an HTTP call","code":"script_http_error","payload":{"error":"Rate limit exceeded"}}}%
```

After:
```
❯ curl --request POST --url http://localhost:3000/action/trigger --header "Authorization: Bearer fe212ac9-0832-4210-b868-400db33b3863" --header "Connection-Id: 612d3d37-0a6c-40a5-96f2-f65df6e8fd98" --header "Provider-Config-Key: unauthenticated" --header "Content-Type: application/json" --data '{"action_name": "rate-limited"}'
{"error":{"message":"An error occurred during an HTTP call","code":"script_http_error","payload":{"error":"Rate limit exceeded"},"additional_properties":{"upstream_response":{"status":429,"headers":{"content-type":"application/json; charset=utf-8"},"body":{"error":"Rate limit exceeded"}}}}}%
```

https://linear.app/nango/issue/NAN-2478/action-error-refinement

## How I tested it:

- Call an action that makes a request that errors
- The server will return the more detailed response

## Deployment

Deployment here is going to be tricky - it's all the main back-end services!


nalanj:
Does anyone have any recommendations on how to document this new status code and let customers know about it? Mintlify doesn't seem to render docs for status codes other than 200 for us.

packages/runner/lib/exec.ts
```diff
@@ -192,7 +215,7 @@ export async function exec(
                     return {
                         success: false,
                         error: {
-                            type: 'script_http_error',
+                            type: 'script_network_error',
```
bodinsamuel, line 228: why changing the error code?

packages/runner/lib/exec.unit.test.ts
```diff
@@ -126,14 +131,15 @@ describe('Exec', () => {
...
         fn = async (nango) => {
-            await nango.get({
-                endpoint: '/',
-                baseUrl: 'https://example.dev/'
-            })
+            const err = new Error("Something broke");
+            err.isAxiosError = true;
+            err.code = "ECONNREFUSED";
+
+            throw err;
```
bodinsamuel, line 143: I'm not sure why would you fake this test that was working before?

packages/runner/lib/exec.unit.test.ts
```diff
@@ -146,6 +152,52 @@ describe('Exec', () => {
...
+                data: { error: "Not found" }
+            }
+            throw err;
+        };
+        exports.default = fn
+        `;
+        const res = await exec(nangoProps, jsCode);
+
+        // NB: it will fail because Nango is not running not because the website is not reachable
+        // NB2: the message is different depending on the system running Node
```
bodinsamuel, line 182: Comment not applicable

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -323,12 +330,6 @@ export class Orchestrator {
                 throw res.error;
             }
 
-            await logCtx.info('The webhook was successfully run', {
-                action: webhookName,
-                connection: connection.connection_id,
-                integration: connection.provider_config_key
-            });
```
bodinsamuel: Why is it removed?

nalanj: Weird. I'm asking the same question.

nalanj: Fixed in 5ecaa067b

packages/runner/lib/exec.unit.test.ts
nalanj: It's a unit test. Unit tests aren't supposed to make network calls. Also relying on a network failure to a domain that could appear is kind of asking for a flaky test.

packages/runner/lib/exec.ts
nalanj: Previously we weren't including any payload information, so it made sense to have the same error code on a network error or an error where we actually received an HTTP response. Once we're including response information it makes sense to have a case where you can expect response data to be included with the error (`script_http_error`) and a case where you can't (`script_network_error`).

packages/runner/lib/exec.unit.test.ts
nalanj: Fixed in b61728bf8

packages/runner/lib/exec.unit.test.ts
bodinsamuel: It is debatable tbh, if the error changes we are not sure to catch it anymore but fine

bodinsamuel (approved):

packages/runner/lib/exec.unit.test.ts
nalanj: Yeah, I have my doubts on us relying on that result anyway. From one side ECONNREFUSED, etc. come out of the node http stack pretty consistently and a lot will have changed in the ecosystem if that output changes, but from the other side I'm not sure we want to rely on that output from the API anyway, since it could change. At this point though we are relying on whether a response was received or not, which seems to be the Axios recommended method for knowing if it was an actual response with an error status or if it was a network error.

docs-v2/spec.yaml
```diff
@@ -1434,6 +1434,48 @@ paths:
...
+                                    error:
+                                        type: object
+                                        properties:
+                                            message:
+                                                type: string
+                                            code:
+                                                type: string
+                                            payload:
+                                                type: object
+                                            additional_properties:
```
TBonnin: why not the upstream attribute directly instead of the additional level of nesting:
```
{
error:
   message:
   code:
   payload:
   upstream: {
      status_code:
      headers: 
      response:
   }
}
      
```

packages/runner/lib/exec.ts
TBonnin: does not data always means network error though?

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -117,13 +118,16 @@ export class Orchestrator {
             'connection.provider_config_key': connection.provider_config_key,
             'connection.environment_id': connection.environment_id
         };
+        if (!connection.id) {
+            throw new NangoError('invalid_input', { connection });
```
TBonnin: this is going to throw the error instead of returning an Err

packages/shared/lib/utils/error.manager.ts
```diff
@@ -44,9 +44,13 @@ class ErrorManager {
...
             logger.error(`Response error`, errorToObject(err));
             if (!err.message) {
-                res.status(err.status || 500).send({ error: { code: err.type || 'unknown_error', payload: err.payload } });
+                res.status(err.status || 500).send({
+                    error: { code: err.type || 'unknown_error', payload: err.payload, additional_properties: err.additional_properties }
+                });
             } else {
-                res.status(err.status || 500).send({ error: { message: err.message, code: err.type, payload: err.payload } });
+                res.status(err.status || 500).send({
+                    error: { message: err.message, code: err.type, payload: err.payload, additional_properties: err.additional_properties }
```
TBonnin, line 52: additional properties can be undefined. should we not return it at all in this case?

packages/types/lib/runner/index.ts
```diff
@@ -1,12 +1,14 @@
 export interface RunnerOutputError {
     type: string;
-    payload: Record<string, unknown>;
+    payload: Record<string, unknown> | unknown[];
```
TBonnin, line 3: where can payload be an array?

nalanj: This was a super weird one once I started cleaning up types in `exec.ts`. 

See https://github.com/NangoHQ/nango/blob/b61728bf82e27631eb9ec73fd9b6e771ec9f38dd/packages/runner/lib/exec.ts#L117. `valInput` is a `ValidateDataError[]`. I'm pretty stumped as to why typescript wasn't picking that up as an error before these changes, though.

packages/shared/lib/utils/error.manager.ts
nalanj: It just gets pruned when it's undefined, since it serializes:

```
❯ curl --request GET \
  --url http://localhost:3003/v1/exception \
  --header 'Authorization: Bearer fe212ac9-0832-4210-b868-400db33b3863' \
  --header 'Connection-Id: 612d3d37-0a6c-40a5-96f2-f65df6e8fd98' \
  --header 'Provider-Config-Key: unauthenticated'
{"error":{"message":"The action script failed with a runtime error","code":"action_script_runtime_error","payload":{"error":"Something broke"}}}%
```

docs-v2/spec.yaml
nalanj: I could see three directions:

1. What you mentioned, just add more attributes as needed. My worry there was that we could end up adding more down the road, and if we do we have to update the typing for the response that comes out of actions each time. Not a big deal but it was kind of tricky to get all the types straight through all the services.
2. We have a grab bag kind of property, `additional_properties`, like I built. In that case the error object stays consistent and we don't have to add all sorts of properties to it down the road. It's less pretty, but is also easier to deal with. And since it's a `Record<string,unknown>` it's easier to change going forward because it has to have nested properties.
3. We shove it into payload. This one mostly doesn't work because I found cases where error payload is an array and I think remember finding other cases where it's just a string.

packages/shared/lib/clients/orchestrator.ts
nalanj: See line 242 in this file. This function has a big try/catch and throws errors above which then get caught and returned as Err. 😬

nalanj: Fixed in a39804e00

docs-v2/spec.yaml
nalanj: Updated in 6b665e0c7, but also waiting for feedback from @bodinsamuel and @khaliqgant and it's an easy thing to revert if we decide to go a different direction.

TBonnin (approved):
works for me with the upstream top level attributes in the json api response.
you didn't answer this comment (it is in a resolved thread): https://github.com/NangoHQ/nango/pull/3279/files#r1913242024

packages/runner/lib/exec.ts
nalanj: Oh right, we should check on no response, not no data. Fixing now!

nalanj: @TBonnin See 3e2cb1b6e. Requesting review since this was a little non-trivial.

bodinsamuel (approved):

nalanj:
Wanted to note this is going to sit here a bit while we notify customers that it's coming.

TBonnin (approved):

docs-v2/spec.yaml
bodinsamuel: Payload should disappear imo, it's too generic to be programmatically useful 
upstream is nice if it's dedicated to HTTP response, I would maybe use a more descriptive name but I have trouble finding something that would convey all the meaning e.g: `upstream_http_error` or `script_http_error` 
Since I'm answering late feel free to discard if customers are already moving to upstream

nalanj: I think since `payload` is already typical to our errors we should just let it hang out longer and we could eventually deprecate it in favor of a less generic error schema, like the `upstream` property here.

PR #3280, opened 2025-01-09
feat(ui): create environment UI

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2224/create-environments-in-nango-ui

- Add UI to create environment
As stated by the spec, it's only available for paying customers and up to 10 env.

- Fix Dropdown checkbox
- Fix some error check

## 🧪 Tests
- Go to dash
- Create an env
Screenshot 2025-01-09 at 18 51 37


packages/server/lib/helpers/validation.ts
```diff
@@ -22,7 +22,7 @@ export const connectionIdSchema = z
     .max(255);
 export const envSchema = z
     .string()
-    .regex(/^[a-zA-Z0-9_-]+$/)
+    .regex(/^[a-z0-9_-]+$/)
```
bodinsamuel, line 25: It was an oversight, prod has no env with an uppercase

packages/server/lib/controllers/v1/environment/postEnvironment.ts
```diff
@@ -28,9 +28,21 @@ export const postEnvironment = asyncWrapper<PostEnvironment>(async (req, res) =>
...
-    const exists = await environmentService.getAccountAndEnvironment({ accountId, envName: body.name });
+    const account = await accountService.getAccountById(accountId);
+    if (account?.is_capped) {
+        res.status(400).send({ error: { code: 'feature_disabled', message: 'Creating environment is only available for paying customer' } });
+        return;
+    }
+
+    const environments = await environmentService.getEnvironmentsByAccountId(accountId);
+    if (environments.length >= 10) {
+        res.status(400).send({ error: { code: 'resource_capped', message: "Can't create more environment" } });
```
nalanj:
```suggestion
        res.status(400).send({ error: { code: 'resource_capped', message: "Can't create more environments" } });
```

nalanj (approved):
Works great! Lots of small comments below that should in no-way block this:

- One silly issue we may want to handle is overflows on the environment name. This would apply in the scrolling list as well. See:

image

- It may also make sense to put an overflow scroll on the list in the dropdown so that the button always shows:

image

- Maybe just a future issue, but should we let you rename an environment from the Environment Settings UI?

- Also should we handle deleting environments? No clue how awful that would be but someone's going to ask for it once they can easily create them.

bodinsamuel:
Thanks, fixed that scroll issue and fixed width



>Maybe just a future issue, but should we let you rename an environment from the Environment Settings UI?

Not planned yet

> Also should we handle deleting environments?

Not planned yet

Since it's for paying customer they will still have a direct line of contact with us, butI'm sure those 2 features will come later 👌🏻

PR #3281, opened 2025-01-09
fix(docker): stop building staging, prod, enterprise in favor of unified

bodinsamuel:
## Changes

- Stop building staging, prod, enterprise in favor of unified 
This means the unified `dockerfile` is built only once and not twice, and I have removed enterprise staging and prod for server (aka we only build one server). 
After this is merged, the images won't be tagged by their env, it will also save time on the CI and we might be able to deploy prod directly from PR (handy for hot fixes but I need to test it first).

We still need to build `server-hosted` separately until we officially support this use case. This already works but needs official communication and deprecation period.


bodinsamuel:
Special ping, for our enterprise customer we need to communicate new images without the `enterprise` and `prod` prefix
```diff
-SERVER: nangohq/nango-server:enterprise-4ff76826fe8f082bed7a0f7a2fbebdd558889c45
+SERVER: nangohq/nango-server:4ff76826fe8f082bed7a0f7a2fbebdd558889c45

-JOBS: nangohq/nango:prod-4ff76826fe8f082bed7a0f7a2fbebdd558889c45
+JOBS: nangohq/nango:4ff76826fe8f082bed7a0f7a2fbebdd558889c45

-RUNNER: nangohq/nango-runner:enterprise-4ff76826fe8f082bed7a0f7a2fbebdd558889c45
+RUNNER: nangohq/nango-runner:4ff76826fe8f082bed7a0f7a2fbebdd558889c45

-PERSIST: nangohq/nango:prod-4ff76826fe8f082bed7a0f7a2fbebdd558889c45
+PERSIST: nangohq/nango:4ff76826fe8f082bed7a0f7a2fbebdd558889c45

-ORCHESTRATOR: nangohq/nango:prod-4ff76826fe8f082bed7a0f7a2fbebdd558889c45
+ORCHESTRATOR: nangohq/nango:4ff76826fe8f082bed7a0f7a2fbebdd558889c45
```

nalanj (approved):

bodinsamuel:
@TBonnin would love your review on this (no urgency just explicitly waiting)

.github/workflows/build-image.yaml
```diff
@@ -10,18 +10,37 @@ on:
...
-                      if: ${{ github.ref == 'refs/heads/master' }}
-
-        secrets: inherit
-        uses: ./.github/workflows/build-image-reusable.yaml
-        with:
-            if: ${{ matrix.group.if }}
-            name: ${{ matrix.group.name }}
+        runs-on: ubuntu-latest
+        env:
+            CAN_PUSH: "${{ secrets.DOCKER_PASSWORD != ' && secrets.DOCKER_USERNAME != ' }}"
```
TBonnin: how does this work? missing a quote no? `secrets.DOCKER_PASSWORD != ''`

TBonnin (approved):

.github/workflows/build-image.yaml
bodinsamuel: good question, it was already like this and working so I guess github actions is flexible. Will fix :p

bodinsamuel:
Will merge monday

PR #3282, opened 2025-01-09
chore: Fix up handler type

nalanj:
I kept noticing spots where I was getting eslint errors because of our handlers being async but the handler type expected `void` and not `Promise<void>`. This updates the type for handlers and cleans up the resulting type errors.

If you want to see this in action, before switching to this branch check out line 125 in `packages/jobs/lib/routes/tasks/putTask.ts`. It has an eslint warning because of a misused promise. After this update that eslint warning is gone.


packages/utils/lib/express/route.ts
```diff
@@ -12,7 +12,7 @@ export interface Route<E extends Endpoint<any>> {
 
 export interface RouteHandler<E extends Endpoint<any>> extends Route<E> {
     validate: (req: EndpointRequest<E>, res: EndpointResponse<E>, next: NextFunction) => void;
-    handler: (req: EndpointRequest<E>, res: EndpointResponse<E>, next: NextFunction) => void;
+    handler: (req: EndpointRequest<E>, res: EndpointResponse<E>, next: NextFunction) => void | Promise<void>;
```
nalanj, line 15: 🍖 You can see down at line 22 that we generally treat it as a promise anyway.

bodinsamuel (approved):

PR #3286, opened 2025-01-10
fix(connect): safari content disappearing bug 

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2497/connect-ui-safari-bug

- Fix bug in Safari where part of the content does not appear until a click event 
The text, and border are white for some reasons, like if the paint did not trigger correctly. Did not find any bug related to this nor any stackoverflow answer so I went for simple and added an empty animation around the faulty content to force trigger a paint on DOM mount.


TBonnin (approved):
I trust you and approve blindly 😉

PR #3288, opened 2025-01-10
feat(integrations): add support for shopify api key

hassan254-prog:

- Add support for Shopify (Api Key).
- Improve on Shopify `connect-ui` docs.


packages/shared/providers.yaml
```diff
@@ -6286,6 +6287,39 @@ shopify:
...
             suffix: .myshopify.com
             prefix: https://
+            doc_section: '#step-1-finding-your-shopify-domain'
+
+shopify-api-key:
+    display_name: Shopify (api key)
+    categories:
+        - e-commerce
+    auth_mode: API_KEY
+    proxy:
```
hassan254-prog, line 6297: @NangoHQ/engineers  I was unable to add the `proxy.verification` field for shopify since the provider expects us to send a non-empty body for the `POST` request. Perhaps we could also add that functionality within the proxy verification method? Tested this and the provider returns `error 200` response even if we pass an `apiKey` with limited permissions. So we should be good here with the `proxy.verification` method. 

```
verification:
    method: POST
    endpoint: /admin/api/2024-10/graphql.json
    data: { "query": "query { products(first: 5) { edges { node { id handle } } pageInfo { hasNextPage } } }" }

khaliqgant: Could you try a `GET` with a graphql introspection query like done with plain: https://github.com/NangoHQ/nango/blob/master/packages/shared/providers.yaml#L5542

hassan254-prog: Thanks @khaliqgant! That worked with the `POST` method. updated with latest commit.

khaliqgant (approved):

PR #3292, opened 2025-01-10
fix(fleet): image override might contain the commitId tag

TBonnin:
the entire point of image override in fleet is to be able to create a node running a specific commit (for pre-deployment testing or verification purpose) This commit ensures that an image override with a commitId tag is properly handled

tested in staging


packages/fleet/lib/supervisor/supervisor.ts
```diff
@@ -320,31 +323,28 @@ export class Supervisor {
...
         {
             routingId,
-            deployment,
-            nodeConfig
+            deployment
         }: {
             type: 'CREATE';
             routingId: Node['routingId'];
             deployment: Deployment;
-            nodeConfig?: NodeConfig | undefined;
```
TBonnin, line 329: unrelated to this change but it was not used anymore

bodinsamuel (approved):
Looks good

PR #3294, opened 2025-01-10
chore: Unit test env fix

nalanj:
We rely on our unit tests not hitting the DB, but I realized while running tests that they were in fact hitting the db. This updates the code to set a DB_NAME that does't exist in the scope of the unit tests, so that they can't write anywhere and will error instead.


TBonnin (approved):
should we also ensure the unit tests that are writing to the db don't write to it anymore?

nalanj:
@TBonnin weirdly I'm pretty sure the test that started failing was because it was able to write to the db - it was counting on not being able to. It was this test: https://github.com/NangoHQ/nango/blob/d06af85dd4527deb3ca02f7d33cbd597cc31e618/packages/shared/lib/services/sync/config/deploy.service.unit.test.ts#L101

Thinking about how to have it fail when it tries to write to the db without actually hitting the db driver, since it's on knex calls.

nalanj:
@TBonnin ad7dac7e6 has the fix

PR #3295, opened 2025-01-10
fix: add application_name to fleet db url + db pool max

TBonnin:
Setting db application name for fleet and setting connection pool max to 15

https://linear.app/nango/issue/NAN-2464/add-application-name-to-database-url-for-fleet

### DIFF
--- a/packages/fleet/lib/db/client.ts
+++ b/packages/fleet/lib/db/client.ts
@@ -13,7 +13,7 @@ export class DatabaseClient {
     public url: string;
     private config: knex.Knex.Config;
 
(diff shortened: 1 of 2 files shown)


packages/fleet/lib/db/client.ts
```diff
@@ -13,7 +13,7 @@ export class DatabaseClient {
     public url: string;
     private config: knex.Knex.Config;
 
-    constructor({ url, schema, poolMax = 50 }: { url: string; schema: string; poolMax?: number }) {
+    constructor({ url, schema, poolMax = 15 }: { url: string; schema: string; poolMax?: number }) {
```
TBonnin, line 16: fleet doesn't need this many connections

nalanj (approved):

PR #3297, opened 2025-01-13
feat(runner): use new SDK

bodinsamuel:
## Changes

- Runner uses the new `runner-sdk`
CLI is already using the new package, it's the last step.

- Drop old code

### 🧪 Tests

Launch your local setup and tests some scripts. 
Output should be strictly similar to before

#### Old PR

<details>
## Context

We need to stop using `shared` in the CLI because it contains a lot (e.g: unrelated database models, internal packages, unrelated npm packages, etc.). We imported shared because it contains the runner SDK which is used in the CLI and in the runner + `providers.yaml`. 
The goal of this project is stop importing shared but still be able to share some code between CLI and Runner.

## Changes

Untitled-2024-02-19-1625 excalidraw

- **New package `@nangohq/runner-sdk`**
It contains the shared SDK, data validation, and paginate service. The SDK is now an abstract class. It includes the common logic, but if a method is different between the CLI and Runner, it is not implemented.
The code is mostly similar, except for:
  - Replaced NangoError by SDKError 
  - Removed all the backend logic (e.g: proxy, pagination, instrumentation, persist, etc.)

- **New package `@nangohq/shared-public`**
This package contains logic shared between the CLI, Runner SDK and Backend. And also `providers.yaml`!
It's relatively small for now, the goal is to limit this as much as we can, so it doesn't end up like `shared`. 
I could not find a good name. I often use `core` or `common` but there would be no difference with `utils` which could create some confusion. Happy to change if needed. 

- **CLI and Runner now implement their own SDK**
Using the new package, they implement the custom part (what was inside `sync.ts`). The footprint of duplicated code is small right now so I feel like it works but happy to discuss this choice. 

- **Simplify npm publish**
Since we don't need to pack things up anymore, it's faster and simpler (but still takes a bit of time)

---

## Todo
<details>

- [x] Make SDK works in Runner
- [x] Fix/inject logger 
- [x] Build models.ts
- [x] Re-up integrations tests
- [x] Move monitor providers
- [x] Publish
</details>

</details>


packages/runner-sdk/models.d.ts
bodinsamuel, line 1: This is a temporary fix, until we find a way to inline all types inside a d.ts or better, get rid of models.ts altogether
That means if you modify the SDK you need to manually change this file for the next few weeks.

packages/runner/lib/sdk/sdk.unit.test.ts
```diff
@@ -151,50 +145,44 @@ describe('Pagination', () => {
             token_url: '',
             docs: ''
         };
-        (await import('@nangohq/node')).Nango.prototype.getIntegration = vi.fn().mockReturnValue({ config: { provider: 'github' } });
-        vi.spyOn(providerService, 'getProvider').mockImplementation(() => provider);
+        vi.spyOn(await import('@nangohq/shared-public'), 'getProvider').mockImplementation(() => provider);
```
bodinsamuel: Since dryRun is no longer available in the backend, I had to change the mocking strategy, thankfully it was not too bad

packages/shared-public/lib/providers.ts
```diff
@@ -0,0 +1,55 @@
+import path from 'node:path';
```
bodinsamuel: This is more or less, providers.ts that was in shared

bodinsamuel:
Seems to be working on staging

nalanj (approved):

TBonnin (approved):
Ran it locally and things seem to work 😄

PR #3298, opened 2025-01-13
fix(cli): Remove header prefix on dry-run

nalanj:
Dry run was previously not taking into account the `Nango-Proxy-` prefix on headers, so they would be logged but then tests would fail. This should resolve that.

See https://linear.app/nango/issue/NAN-2498/failing-integration-template-tests


packages/cli/lib/services/response-saver.service.ts
```diff
@@ -153,9 +153,12 @@ function computeConfigIdentity(config: AxiosRequestConfig): ConfigIdentity {
 
     let headers: [string, string][] = [];
     if (config.headers !== undefined) {
-        const filteredHeaders = Object.entries(config.headers).filter(([key]) => !FILTER_HEADERS.includes(key.toLowerCase()));
+        const filteredHeaders = Object.entries(config.headers)
+            .filter(([key]) => !FILTER_HEADERS.includes(key.toLowerCase()))
+            .map<[string, string]>(([key, value]) => (key.toLowerCase().startsWith('nango-proxy-') ? [key.slice(12), String(value)] : [key, String(value)]));
         sortEntries(filteredHeaders);
         headers = filteredHeaders;
+        console.log(JSON.stringify(headers));
```
khaliqgant: Assume this is left over from debugging?

nalanj: Oops, yep. Fixed in 0e612b381

bodinsamuel (approved):

PR #3301, opened 2025-01-14
docs: restructure docs

bastienbeurier:
This change entails: 
- Restructure the documentation guides
- Rename most files
- Set up redirects from old URLs to new ones
- Regenerate API docs snippets
- Update links in the docs
- Update links in the web app


bodinsamuel (approved):
Won't read everything but looks good 👏🏻 Just a few feedback:
- The mention `(free)` is confusing since you are not always in a "buying context", plus it makes everything else "paying" by comparison which is not true. 
- `Individual requests` is confusing. Proxy was not working? Or PassThrough?

bastienbeurier:
> Won't read everything but looks good 👏🏻 Just a few feedback:
> 
> * The mention `(free)` is confusing since it's you are not always in a "buying context", plus it makes everything else "paying" by comparison which is not true.
> * `Individual requests` is confusing. Proxy was not working? Or PassThrough?

Thanks for reviewing and for the relevant feedback. I suspect there's more similar feedback to gather on the new structure/naming. If that's okay, I'll get this out and submit the new docs to the team for deeper review, and we can discuss these two items then.

PR #3302, opened 2025-01-14
fix(types): copy SDK types to @types

bodinsamuel:
## Changes

- Copy SDK types to `@nangohq/types` 
It's to prepare the field when I introduce a dedicated package for the SDK, so we can reuse types across the codebase. Those types are not used at the moment. (see #3297 for full PR)

- Fix types for pagination 
It wasn't using a discriminated object which makes type narrowing impossible 

- Fix types for response saver


TBonnin (approved):

nalanj (approved):
Like it!

PR #3303, opened 2025-01-14
fix(fleet): render error cause is swallowed by the logger 

TBonnin:


bodinsamuel (approved):
Weird

TBonnin:
Probably related to the winston issue https://github.com/winstonjs/winston/issues/2533

PR #3304, opened 2025-01-14
fix(types): NangoProps use DBSyncConfig

bodinsamuel:
## Changes

Split of #3297 

- NangoProps use DBSyncConfig
In order to stop using shared I need to stop using the types from shared, so instead of copying an invalid type I took the time to use the proper one across the board.


packages/runner/lib/exec.ts
```diff
@@ -122,7 +122,7 @@ export async function exec(
                 const valOutput = validateData({
                     version: nangoProps.syncConfig.version || '1',
                     input: output,
-                    modelName: nangoProps.syncConfig.models.length > 0 ? nangoProps.syncConfig.models[0] : undefined,
+                    modelName: nangoProps.syncConfig.models!.length > 0 ? nangoProps.syncConfig.models![0] : undefined,
```
nalanj: I get eslint non-null assertion errors here. 

```suggestion
                    modelName: nangoProps.syncConfig.models && nangoProps.syncConfig.models.length > 0 ? nangoProps.syncConfig.models[0] : undefined,
```

packages/jobs/lib/execution/webhook.ts
```diff
@@ -188,7 +188,7 @@ export async function handleWebhookSuccess({ nangoProps }: { nangoProps: NangoPr
     const logCtx = await logContextGetter.get({ id: String(nangoProps.activityLogId) });
     const environment = await environmentService.getById(nangoProps.environmentId);
     if (environment) {
-        for (const model of nangoProps.syncConfig.models) {
+        for (const model of nangoProps.syncConfig.models!) {
```
nalanj, line 191: Is there any way for us to remove these non-null assertions since they come through as eslint errors?

nalanj (approved):
Looks good. Just some small feedback that is totally your call.

packages/jobs/lib/execution/webhook.ts
bodinsamuel: yes you are right, this has been noted here 👍🏻 
https://github.com/NangoHQ/nango/blob/c62d19625cc409fcfb7618a0b9811cdfddd6eb66/packages/types/lib/syncConfigs/db.ts#L12

packages/runner/lib/exec.ts
bodinsamuel: Good point

PR #3308, opened 2025-01-15
feat: providers package 

bodinsamuel:
## Changes

Split of #3297 

- New package `@nangohq/providers`
Only contains providers.yaml. Compiled and included in tooling but not used in code yet


nalanj (approved):

packages/providers/package.json
```diff
@@ -0,0 +1,21 @@
+{
+    "name": "@nangohq/providers",
+    "version": "0.48.3",
+    "description": "Nango's shared components.",
```
khaliqgant: This description should be more accurate

khaliqgant (changes requested):
Small nit comment made inline.

💯 to having it in its own package.

khaliqgant:
Can you be sure to ping me when this moves permanently? I have a couple of scripts that are watching the providers.yaml for changes to take some actions on the new provider that was added. Thanks!

bodinsamuel:
> Can you be sure to ping me when this moves permanently? I have a couple of scripts that are watching the providers.yaml for changes to take some actions on the new provider that was added. Thanks!

Yes probably not before last week of january, until then nothing changes

khaliqgant (approved):

PR #3309, opened 2025-01-15
feat(integrations): add support for Microsoft business central

hassan254-prog:

- Add support for Microsoft Business Central


hassan254-prog:
@khaliqgant , this is same to `Oauth2_CC` for `UMAPI (Adobe User Management API)` where we need to also pass in a scope for us to get back the `access_token`. Tested this with the headless connection.

khaliqgant (approved):

PR #3310, opened 2025-01-15
fix(jobs): closing logic

TBonnin:
- close the app if db healthCheck fails 5 times in a row (not after the first failure)
- when closing first close the webserver and then destroy the db client


bodinsamuel (approved):

bodinsamuel:
I wonder if we should just close server then everything else. Could api calls not finish without the rest?

nalanj (approved):

TBonnin:
> I wonder if we should just close server then everything else. Could api calls not finish without the rest?

requests received by jobs are from the runners so they are not dependent on the task processor or anything else (other than the db)
it is a good point though, some otlp operations could be lost for instance. let me modify and close everything after the server is done

PR #3311, opened 2025-01-15
fix(render): add throttling so fleet never reaches render api limit

TBonnin:
Render has some rate limit in place regarding the creation of new service.
Current limits are 1000 per hours and recommends to stay around 20 creations per minute.
This commit adds a throttling mechanism to the Render node provider

### DIFF
--- a/packages/jobs/lib/runner/render.ts
+++ b/packages/jobs/lib/runner/render.ts
@@ -1,11 +1,16 @@
 import type { Node, NodeProvider } from '@nangohq/fleet';
 import type { Result } from '@nangohq/utils';
-import { Err, Ok } from '@nangohq/utils';
+import { Err, Ok, getLogger } from '@nangohq/utils';
 import { RenderAPI } from './render.api.js';
 import { envs } from '../env.js';
-import { getPersistAPIUrl, getProvidersUrl } from '@nangohq/shared';
+import { getPersistAPIUrl, getProvidersUrl, getRedisUrl } from '@nangohq/shared';
 import type { AxiosResponse } from 'axios';
 import { isAxiosError } from 'axios';
+import type { RateLimiterAbstract } from 'rate-limiter-flexible';
+import { RateLimiterRedis, RateLimiterMemory } from 'rate-limiter-flexible';
+import { createClient } from 'redis';
+
+const logger = getLogger('Render');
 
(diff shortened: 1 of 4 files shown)


packages/jobs/lib/runner/render.ts
```diff
@@ -137,3 +149,53 @@ function getPlan(node: Node): 'starter' | 'standard' | 'pro' {
...
+const serviceCreationThrottler = await (async () => {
+    const minuteThrottlerOpts = {
+        keyPrefix: 'minute',
+        points: envs.RENDER_SERVICE_CREATION_MAX_PER_MINUTE || 50,
+        duration: 60,
+        blockDuration: 0
+    };
+    const hourThrottlerOpts = {
+        keyPrefix: 'hour',
+        points: envs.RENDER_SERVICE_CREATION_MAX_PER_HOUR || 600,
```
TBonnin: the hard limit is 1000, keeping some buffer to be able to create some new runners and allow enough room for an emergency full deploy. When needed we could change the env var and redeploy jobs

packages/jobs/lib/runner/render.ts
```diff
@@ -1,11 +1,16 @@
...
+import { Err, Ok, getLogger } from '@nangohq/utils';
 import { RenderAPI } from './render.api.js';
 import { envs } from '../env.js';
-import { getPersistAPIUrl, getProvidersUrl } from '@nangohq/shared';
+import { getPersistAPIUrl, getProvidersUrl, getRedisUrl } from '@nangohq/shared';
 import type { AxiosResponse } from 'axios';
 import { isAxiosError } from 'axios';
+import type { RateLimiterAbstract } from 'rate-limiter-flexible';
+import { RateLimiterRedis, RateLimiterMemory } from 'rate-limiter-flexible';
+import { createClient } from 'redis';
```
bodinsamuel, line 11: too bad I never had the time to finish redis migration to kvstore

packages/jobs/lib/runner/render.ts
bodinsamuel: 600 is currently 1.8 deploy, maybe it's a bit too short if we deploy and rollback

bodinsamuel (approved):

packages/jobs/lib/runner/render.ts
TBonnin: bumping it to 700

PR #3312, opened 2025-01-15
feat(server): Log webhooks from syncs separate from the sync

nalanj:
Log webhooks from syncs in a different log context than the sync itself.

I'll be putting together future PR(s) to handle webhooks from refresh token failures and new connections.

See https://linear.app/nango/issue/NAN-1547/log-nango-webhooks

## How I tested it

- Set up webhooks
- Trigger a sync
- You'll see the sync run, and then you'll see a separate webhook event.


TBonnin:
since we are improving the logging. Can we check if we could add some http logging that would help customers debug why calling the webhook endpoint failed?

nalanj:
@TBonnin absolutely - I'll dig into that too

nalanj:
@TBonnin better errors are covered in #3319

package.json
```diff
@@ -51,8 +51,8 @@
     "devDependencies": {
         "@eslint/js": "9.17.0",
         "@types/node": "20.12.2",
-        "@typescript-eslint/eslint-plugin": "8.19.1",
-        "@typescript-eslint/parser": "8.19.1",
+        "@typescript-eslint/eslint-plugin": "8.20.0",
```
nalanj, line 54: Bumped these to try to fix the build failure you can see in the commit history. I figured I might as well keep the upgrade around.

packages/jobs/lib/execution/sync.ts
```diff
@@ -295,6 +300,34 @@ export async function handleSyncSuccess({ nangoProps }: { nangoProps: NangoProps
...
+
+                        if (!providerConfig.id) {
+                            throw new Error('Missing providerConfig.id');
+                        }
+
+                        if (!connection.id) {
+                            throw new Error('Missing connection.id');
+                        }
+
+                        if (!nangoProps.syncConfig.id) {
```
nalanj: These checks are how I ended up resolving the build issue in CI. More than happy to discuss further, but basically the lint error I was seeing in the build history on commits in this PR wasn't happening on my computer, and I could never get it to reproduce. This worked around it.

packages/jobs/lib/execution/sync.ts
```diff
@@ -216,6 +216,11 @@ export async function handleSyncSuccess({ nangoProps }: { nangoProps: NangoProps
             provider_config_key: nangoProps.providerConfigKey
         };
 
+        const providerConfig = await configService.getProviderConfig(connection.provider_config_key, connection.environment_id);
+        if (providerConfig === null) {
+            throw new Error(`Provider config not found for connection: ${connection.connection_id} in handleSyncSuccess`);
```
bodinsamuel: Not sure it's required since we have the stack trace 
```suggestion
            throw new Error(`Provider config not found for connection: ${connection.connection_id}`);
```

packages/jobs/lib/execution/sync.ts
bodinsamuel: I think you are just not up to date

packages/jobs/lib/execution/sync.ts
```diff
@@ -309,11 +342,14 @@ export async function handleSyncSuccess({ nangoProps }: { nangoProps: NangoProps
...
                             operation: lastSyncDate ? SyncType.INCREMENTAL : SyncType.FULL,
-                            logCtx
+                            logCtx: webhookLogCtx
                         });
 
                         if (res.isErr()) {
+                            await webhookLogCtx.failed();
                             throw new Error(`Failed to send webhook for sync: ${nangoProps.syncConfig.sync_name}`);
+                        } else {
+                            await webhookLogCtx.success();
```
bodinsamuel: do you think it could be self-contained inside sendSyncWebhook?

bodinsamuel (approved):
Looks good 👍🏻 
I feel like it's missing the same code for `onFailure`. Also `sendSyncWebhook` is used in webhook execution but not sure if it's part of this change?

packages/jobs/lib/execution/sync.ts
nalanj: Just merged and am running through CI! 🤞🏻

packages/jobs/lib/execution/sync.ts
nalanj: Definitely - we'd just need to send a handful more parameters to `sendSyncWebhook` to help it have all the log context params it needs. I'm not sure if that's ideal with the package separation, but we've already got a function called `sendSync` so it's not like the webhook package doesn't know about syncs.

nalanj: @bodinsamuel See ee9c652 for version where the log context is created inside of sendSync.

bodinsamuel (approved):

packages/jobs/lib/execution/sync.ts
```diff
@@ -216,12 +220,17 @@ export async function handleSyncSuccess({ nangoProps }: { nangoProps: NangoProps
             provider_config_key: nangoProps.providerConfigKey
         };
 
+        providerConfig = await configService.getProviderConfig(connection.provider_config_key, connection.environment_id);
+        if (!providerConfig) {
+            throw new Error(`Provider config not found for connection: ${connection.connection_id}`);
+        }
+
```
TBonnin, line 227: was it too much of a pain to pass the providerConfig in NangoProps to avoid having to call the DB when handling the output?

packages/jobs/lib/execution/sync.ts
```diff
@@ -293,32 +303,36 @@ export async function handleSyncSuccess({ nangoProps }: { nangoProps: NangoProps
...
-                                updated,
-                                deleted
-                            },
-                            operation: lastSyncDate ? SyncType.INCREMENTAL : SyncType.FULL,
-                            logCtx
-                        });
-
-                        if (res.isErr()) {
-                            throw new Error(`Failed to send webhook for sync: ${nangoProps.syncConfig.sync_name}`);
+                    if (team && environment && providerConfig) {
```
TBonnin: should we report or log an error if any of this attributes are missing?

TBonnin (approved):
looks good. quick question about providerConfig in NangoProps

packages/jobs/lib/execution/sync.ts
nalanj: I didn't even think about that since `NangoProps` just doesn't have that property currently. Do you think it's something we'd want to add?

packages/jobs/lib/execution/sync.ts
nalanj: See 7510a1da5

packages/jobs/lib/execution/sync.ts
TBonnin: Wee could save a few queries to the db by putting all static data in NangoProps, like for instance the full account, environment and providerConfig. It might be easier to do that in another PR though

nalanj: Yeah, I'll land this and log that.

PR #3313, opened 2025-01-15
fix(jobs): Try to use a check loop to make that one supervisor test not fail

nalanj:
We've had a supervisor test acting flaky. This attempts to put it into a loop in the test to give it longer to succeed or fail.


bodinsamuel (approved):

TBonnin:
@nalanj thank you for fixing this

nalanj:
@TBonnin 🙏🏻 hopefully it did fix it. I got it to happen one more time in my dev env after I made the change but am not sure if it was a build issue because it never happened again after that.

PR #3314, opened 2025-01-16
fix(proxy): fix proxy header construction for two_step

hassan254-prog:
Fix Proxy header construction for `two_step` auth mode.

### DIFF
--- a/packages/shared/lib/services/proxy.service.ts
+++ b/packages/shared/lib/services/proxy.service.ts
@@ -108,7 +108,7 @@ class ProxyService {
                 break;
             }
             default: {
-                throw new Error(`Unhandled connection.credentials for: ${connection.credentials}`);
+                throw new Error(`Unhandled connection.credentials for: ${(connection.credentials as any).type}`);
             }
         }
 
--- a/packages/shared/lib/services/proxy.service.ts
+++ b/packages/shared/lib/services/proxy.service.ts
@@ -100,6 +100,15 @@ class ProxyService {
                 const nangoError = new NangoError('pass_through_error', error);
                 return { success: false, error: nangoError, response: null, logs };
             }
+            case 'APP_STORE':
+            case 'CUSTOM':
+            case 'TBA':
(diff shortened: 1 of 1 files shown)


packages/shared/lib/services/proxy.service.ts
```diff
@@ -77,52 +77,29 @@ class ProxyService {
...
+            case 'BASIC':
+            case 'API_KEY':
+                token = connection.credentials;
                 break;
+            case 'OAUTH1': {
+                const error = new Error('OAuth1 is not supported yet in the proxy.');
+                const nangoError = new NangoError('pass_through_error', error);
+                return { success: false, error: nangoError, response: null, logs };
+            }
         }
```
bodinsamuel, line 113: we need to enable the check on eslint for switch/case completeness.
In the meantime can you add all the missing ones (with just a break) and a `default` case too with a throw so it's clear when it's missing?

bodinsamuel (approved):

packages/shared/lib/services/proxy.service.ts
```diff
@@ -77,52 +77,39 @@ class ProxyService {
...
+            }
+            case 'APP_STORE':
+            case 'CUSTOM':
+            case 'TBA':
+            case undefined:
+            case 'BILL': {
+                break;
+            }
+            default: {
+                throw new Error(`Unhandled connection.credentials for: ${connection.credentials}`);
```
bodinsamuel: Passing this could work but would also leak credentials, typescript is complaining about never but I think it's acceptable to type it as any in that case `((connection.credentials as any).type`

packages/shared/lib/services/proxy.service.ts
hassan254-prog: I agree 💯, checking with eslint would be better. I have updated with latest commit.

bodinsamuel (approved):
🙏🏻

TBonnin (approved):

PR #3316, opened 2025-01-16
feat(api): track request content length

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2540/track-request-content-length

- API track request content length
Inside traces as tag and as a metric to be able to graph it properly. I'm not the best at math so hopefully the computation is correct.

Screenshot 2025-01-16 at 11 31 05
Screenshot 2025-01-16 at 11 24 19

### DIFF
--- a/packages/server/lib/utils/asyncWrapper.ts
+++ b/packages/server/lib/utils/asyncWrapper.ts
@@ -19,7 +19,7 @@ export function asyncWrapper<TEndpoint extends Endpoint<any>, Locals extends Rec
             const contentLength = req.header('content-length');
             if (contentLength) {
                 const int = parseInt(contentLength, 10);
-                active.setTag('http.request.content_length', `${(int / 1024).toPrecision(2)}kb`);
+                active.setTag('http.request.body.size', int);
                 metrics.histogram(metrics.Types.API_REQUEST_CONTENT_LENGTH, int);
             }
         }
--- a/packages/server/lib/utils/asyncWrapper.ts
+++ b/packages/server/lib/utils/asyncWrapper.ts
@@ -3,6 +3,7 @@ import type { Endpoint } from '@nangohq/types';
 import type { RequestHandler, Request, Response, NextFunction } from 'express';
 import { isAsyncFunction } from 'util/types';
 import type { RequestLocals } from './express.js';
+import { metrics } from '@nangohq/utils';
 
 export function asyncWrapper<TEndpoint extends Endpoint<any>, Locals extends Record<string, any> = Required<RequestLocals>>(
     fn: (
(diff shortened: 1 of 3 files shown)


packages/server/lib/utils/asyncWrapper.ts
```diff
@@ -13,7 +14,15 @@ export function asyncWrapper<TEndpoint extends Endpoint<any>, Locals extends Rec
 ): RequestHandler<any, TEndpoint['Reply'], any, any, any> {
     return (req, res, next) => {
         const active = tracer.scope().active();
-        active?.setTag('http.route', req.route?.path || req.originalUrl);
+        if (active) {
+            active.setTag('http.route', req.route?.path || req.originalUrl);
+            const contentLength = req.header('content-length');
+            if (contentLength) {
+                const int = parseInt(contentLength, 10);
+                active.setTag('http.request.content_length', `${(int / 1024).toPrecision(2)}kb`);
```
nalanj: Just curious - why not just dump bytes?

nalanj (approved):
🙌🏻 Excited that we'll have this.

Just the one question about dumping bytes. It's usually the more standard way to track metrics on data transfer, and I noticed in the datapoints for the traces in the top of the page that the traces showed a lot of < 1 values.

packages/server/lib/utils/asyncWrapper.ts
bodinsamuel: I have formatted it for readability, reading bytes can be confusing IMO. Happy to log the raw value

bodinsamuel:
> It's usually the more standard way to track metrics on data transfer

You mean listening at express level `req.on('data')` and computing the length ourselves?

nalanj:
@bodinsamuel No, I just meant usually in o11y pipelines you'd always dump bytes, not kilobytes. Like in the otel semantic conventions you can see that here: https://opentelemetry.io/docs/specs/semconv/http/http-metrics/#metric-httpclientresponsebodysize

Adding the suffix definitely helps there, and I'm down for whatever. Was just a trivial bit of feedback.

packages/server/lib/utils/asyncWrapper.ts
```diff
@@ -13,7 +14,15 @@ export function asyncWrapper<TEndpoint extends Endpoint<any>, Locals extends Rec
 ): RequestHandler<any, TEndpoint['Reply'], any, any, any> {
     return (req, res, next) => {
         const active = tracer.scope().active();
-        active?.setTag('http.route', req.route?.path || req.originalUrl);
+        if (active) {
```
TBonnin, line 17: there is a hooks attributes for the ddtrace express plugin https://datadoghq.dev/dd-trace-js/interfaces/export_.plugins.express.html#hooks 
could we use this instead?

packages/server/lib/utils/asyncWrapper.ts
```diff
@@ -13,7 +14,15 @@ export function asyncWrapper<TEndpoint extends Endpoint<any>, Locals extends Rec
...
+            active.setTag('http.route', req.route?.path || req.originalUrl);
+            const contentLength = req.header('content-length');
+            if (contentLength) {
+                const int = parseInt(contentLength, 10);
+                active.setTag('http.request.content_length', `${(int / 1024).toPrecision(2)}kb`);
+                metrics.histogram(metrics.Types.API_REQUEST_CONTENT_LENGTH, int);
```
TBonnin, line 23: not sure what the metric gives us. when would it be useful?

bodinsamuel: it's not possible to graph the tag, but we can graph the volume of this metric so we can tie it to outage (if any)

packages/server/lib/utils/asyncWrapper.ts
bodinsamuel: it's handy to have it here (or at least in the other code) because it's shared across multiple services, otherwise, I need to copy the same code in each `tracer.ts`. Wdyt?

TBonnin: do we need it for all services (jobs, orchestrator) or just for server though?

TBonnin (approved):

packages/server/lib/utils/asyncWrapper.ts
bodinsamuel: regarding yesterday incidents I think it would be great to log them everywhere, we are a bit in the blind regarding the payload that are passed by customers

PR #3317, opened 2025-01-16
feat: package runner-sdk

bodinsamuel:
## Changes 

Split of #3297 

- **New package `@nangohq/runner-sdk`**
It contains the shared SDK, data validation, and paginate service. The SDK is now an abstract class. It includes the common logic, but if a method is different between the CLI and Runner, it is not implemented.
The code is mostly similar, except for:

  - Replaced NangoError by SDKError
  - Removed all the backend logic (e.g: proxy, instrumentation, persist, etc.)

- Hardcoded `models.ts` 
It can't be computed like it used to (it wasn't really working anyway), it's hardcoded and delivered as-is for the CLI, so it needs to be maintained manually. We will change that in the coming weeks.


packages/providers/providers.yaml
bodinsamuel, line 1: made that a symlink until we migrate so we don't need to maintain it manually

TBonnin (approved):
looks good. Hard to compare with current code since it is all new files though

PR #3319, opened 2025-01-16
feat(server): Enrich webhook errors

nalanj:
Enriches webhook errors with much more information about the request that was made, to help customers with debugging.

Of course they should log these requests themselves, but there are certainly also cases like gateway errors where they wouldn't receive these and it could be good to see the full request from our side:

image

In that screenshot above, I fixed this small bug:
image

## How I tested it:

- Set up webhooks to whatever test endpoint you like to use, but make sure they fail
- Run a sync so it delivers webhooks
- Check out the logs for failed webhook requests


packages/webhooks/tsconfig.json
```diff
@@ -13,6 +13,9 @@
         },
         {
             "path": "../types"
+        },
+        {
+            "path": "../shared"
```
TBonnin: I am not sure we want to bring shared as a dependency of webhook.
Would it make sense to move the redactHeaders function into `utils`. It doesn't have any dependency and no business domain logic. I think it would be a good candidate to live in `utils`

packages/webhooks/lib/utils.ts
```diff
@@ -27,12 +29,39 @@ export const NON_FORWARDABLE_HEADERS = [
...
+
+                const metaRequest: Record<string, unknown> = {};
+                metaRequest['url'] = `${request.protocol}//${request.host}${request.path}`;
+                metaRequest['method'] = request.method;
+                metaRequest['headers'] = redactHeaders({ headers: request.getHeaders() });
+                meta['request'] = metaRequest;
+            }
+        }
+
+        await logCtx?.error(content, truncateJson(meta));
```
TBonnin: there is a convenience method `logCtx.http()` to log http request.

nalanj: @TBonnin Actually I'm not in love with that method because it doesn't write the response body, which could be helpful in these cases. Would we want to add that?

bodinsamuel: Good point, I think we can definitely add a new field to the `type: http`, alternatively you can use `http()` and log the body in the meta field freely. One of the goal with this method is to have a std way to log HTTP across the codebase, let's try to make that fit 💪🏻

bodinsamuel (commented):
Same points already raised by Thomas, otherwise looks good 👍🏻 

Can you take the opportunity to log success and error the same way in `deliver()`?

packages/webhooks/tsconfig.json
bodinsamuel: agreed 👍🏻

nalanj: Fixed in https://github.com/NangoHQ/nango/pull/3321

nalanj: And 4db0e0c77

nalanj:
@bodinsamuel handled deliver errors in a36990b

packages/webhooks/lib/utils.ts
nalanj: Added `body`. I'm not going to mess with it here, but I really wish the `code` field on those http response logs was `status` or `status_code` instead.

packages/webhooks/lib/utils.ts
```diff
@@ -126,36 +139,61 @@ export const deliver = async ({
...
+                { numOfAttempts: RETRY_ATTEMPTS, retry: retry.bind(this, logRequest, logCtx) }
             );
 
             if (logCtx) {
+                const logResponse = formatLogResponse(response);
+
                 if (response.status >= 200 && response.status < 300) {
-                    await logCtx.info(
+                    await logCtx.http(
                         `${webhookType} webhook sent successfully to the ${type} ${url} and received with a ${response.status} response code${endingMessage ? ` ${endingMessage}` : ''}.`,
```
bodinsamuel, line 167: message for HTTP log should be `METHOD URL`  (yes it's not good everywhere :D)

bodinsamuel (approved):
💪🏻

packages/types/lib/logs/messages.ts
```diff
@@ -134,6 +134,7 @@ export interface MessageRow {
     response: {
         code: number;
         headers: Record<string, string>;
+        body?: unknown;
```
bodinsamuel, line 138: Just had a thought before going to bed, you need to add this prop to schema.ts as keyword "index": false and "doc_values": false otherwise it will blow up

nalanj: Fixed in 2c9e4f96

bodinsamuel (approved):

packages/webhooks/package.json
```diff
@@ -16,6 +16,7 @@
     "keywords": [],
     "dependencies": {
         "@nangohq/logs": "file:../logs",
+        "@nangohq/shared": "file:../shared",
```
TBonnin: Do you still need this?

nalanj: Good catch.

nalanj: Fixed in 411bf4584

TBonnin (approved):

PR #3322, opened 2025-01-16
feat(providers): use new package

bodinsamuel:
## Changes

Split of #3297

- Use new package for providers
Figured we can actually use it straight away since we publish it the CLI can use it without packing.


docker-compose.yaml
```diff
@@ -44,7 +44,7 @@ services:
             - NANGO_LOGS_ES_PWD
             - FLAG_SERVE_CONNECT_UI=${FLAG_SERVE_CONNECT_UI:-true}
         volumes:
-            - './packages/shared/providers.yaml:/usr/nango-server/src/packages/shared/providers.yaml'
+            - './packages/providers/providers.yaml:/usr/nango-server/src/packages/providers/providers.yaml'
```
bodinsamuel, line 47: This needs a breaking changes notice (I don't think a lot of people use this but this)

nalanj (approved):

PR #3323, opened 2025-01-16
fix(fleet): RUNNING to IDLE is a valid state transition

TBonnin:
RUNNING node should be terminated when they haven't executed a script in the last 24h. 
Right now the RUNNING -> IDLE transition results in an error


TBonnin:
I would like to release this change before I am off for the weekend. Since it is low risk, I am gonna merge it now.

nalanj (commented):
👍🏻

PR #3330, opened 2025-01-17
fix(webapp): Handle null sync_type to stop ui error

nalanj:
Came from a customer report. We had a `sync_type` coming over the API as null, and that caused errors in this condition.

image


nalanj:
Went ahead and merged since everyone is out of office right now.

nalanj:
@bodinsamuel would love your eyes on this regardless of it already being merged to make sure I didn't mess something up

PR #3331, opened 2025-01-20
feat(integrations): split greenhouse apis into seperate providers

hassan254-prog:
- Split greenhouse apis into seperate providers


packages/shared/providers.yaml
```diff
@@ -3542,6 +3543,130 @@ greenhouse-basic:
...
+            default_value: ''
+            hidden: true
+
+greenhouse-job-board:
+    display_name: Greenhouse (Job Board API)
+    auth_mode: BASIC
+    categories:
+        - ats
+    proxy:
+        base_url: https://boards-api.greenhouse.io
```
TBonnin: no `retry`

packages/shared/providers.yaml
```diff
@@ -3542,6 +3543,130 @@ greenhouse-basic:
...
+            secret: true
+            doc_section: '#step-2-obtaining-your-api-key'
+        password:
+            type: string
+            title: Password
+            description: Password
+            # https://developers.greenhouse.io/harvest.html#authentication
+            # Greenhouse is using basic auth with an api key
+            default_value: ''
+            hidden: true
```
TBonnin: are the 3 BASIC provider the same with different proxy params? can they be aliased?

TBonnin (approved):
small question about aliases. lgtm otherwise

packages/shared/providers.yaml
```diff
@@ -3542,6 +3543,130 @@ greenhouse-basic:
...
+    proxy:
+        base_url: https://${connectionConfig.baseUrl}
+    docs: https://docs.nango.dev/integrations/all/greenhouse-assessment
+    docs_connect: https://docs.nango.dev/integrations/all/greenhouse-assessment/connect
+    connection_config:
+        baseUrl:
+            type: string
+            title: Assessment Base Url
+            description: The Assessment API Base Url you want to connect to
+            pattern: '^[a-z0-9_-]+$'
```
bodinsamuel: could be just `type: hostname` no?

bodinsamuel (approved):
Massive 💯

khaliqgant (changes requested):
Looks good, could you just move to the new providers.yaml file which is now in the `provider` package

packages/shared/providers.yaml
hassan254-prog: Yes, I noticed that as well. Since each API has its own documentation, some do not explicitly mention rate limits. Some Greenhouse APIs are rate-limited, while others are not.

packages/shared/providers.yaml
hassan254-prog: Yes, I tried aliasing and overriding the `base_url`, but the platform still required the user to provide the `connectionConfigs` from the main provider in this case `greenhouse-basic`. @TBonnin, Perhaps this behavior is intentional by design?

packages/shared/providers.yaml
hassan254-prog: yes, updated with the latest commit 🙏 .

khaliqgant (approved):

PR #3335, opened 2025-01-20
fix(github-app): public links are set in the integration

khaliqgant:
As reported in the community the public link is set in the integration so isn't needed in the modal.

image

### DIFF
--- a/packages/shared/providers.yaml
+++ b/packages/shared/providers.yaml
@@ -3116,6 +3116,7 @@ github-app:
             description: The public link of your GitHub App
             format: uri
             pattern: '^https?://.*$'
+            automated: true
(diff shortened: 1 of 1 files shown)


bodinsamuel (approved):

bodinsamuel:
Should we change `connectionConfig.something` to be `integration.something` in that case? Because that's confusing, I'm not even sure how it works rn

khaliqgant:
> Should we change `connectionConfig.something` to be `integration.something` in that case? Because that's confusing, I'm not even sure how it works rn

Yeah, agreed that `integration.someValue` would be clearer, but that would be an improvement to make down the line IMO.

PR #3338, opened 2025-01-21
feat(cli): Allow init to create any directory name

nalanj:
- Allows any directory name. If no directory is given, init creates against the current directory, like git does.
- Uses .nango directory as a sentinel for determining if it's a nango integration folder or not
- Moves init into services directory for consistency
- Moves several of the files that were created in init with template strings to be files in templates, also for consistency
- Moves some tests around to be beside the init service
- Updates some docs


nalanj:
@khaliqgant here's what I have so far on allowing init with any directory

package.json
```diff
@@ -13,7 +13,7 @@
         "prettier-watch": "onchange './**/*.{ts,tsx}' -- prettier --write {{changed}}",
         "lint": "eslint . ",
         "lint:fix": "eslint . --fix",
-        "ts-build": "tsc -b tsconfig.build.json",
+        "ts-build": "npm run -w nango copy:files && tsc -b tsconfig.build.json",
```
bodinsamuel: it's a primitive (:D) script that only compiles typescript, it's cool that it's side-effects-free, what's the reason behind this?

packages/cli/lib/services/init.service.ts
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Init
+ * If we're not currently in the nango-integrations directory create one
+ * and create an example nango.yaml file
+ */
+export function init({ absolutePath, debug = false }: { absolutePath: string; debug?: boolean }): boolean {
+    const yamlData = fs.readFileSync(path.resolve(__dirname, `../templates/${nangoConfigFile}`), 'utf8');
+    const envData = fs.readFileSync(path.resolve(__dirname, '../templates/env'), 'utf8');
+    const gitIgnoreData = fs.readFileSync(path.resolve(__dirname, '../templates/gitignore'), 'utf8');
```
bodinsamuel: is this really necessary considering the file is super small?
also it should not be empty but should contains ```dist
.env```

packages/cli/lib/services/init.service.unit.cli-test.ts
```diff
@@ -0,0 +1,43 @@
+import { expect, describe, it, afterEach, vi } from 'vitest';
```
bodinsamuel: Does it really need to be `cli-test` (vs unit.test)?

bodinsamuel (approved):
Let's gooo

packages/cli/lib/services/init.service.unit.cli-test.ts
nalanj: Definitely not - was just following the prior pattern for where those files were. Will fix.

nalanj: Fixed in 704bf3c92

package.json
nalanj: :+1: Removed in 141646cdc

packages/cli/lib/services/init.service.ts
nalanj: I was trying to move these all to file copies just so they were a little easier to modify and more consistent to look at, but I can revert.

nalanj: Fixed in 3986c3553

docs-v2/reference/cli.mdx
```diff
@@ -56,7 +56,7 @@ nango [command] --help
 
 | Command                        | Description                                                                                                                                       | Details                                                                                              |
 | -                              | -                                                                                                                                                 | -                                                                                                    |
-| `nango init`                   | Creates the `nango-integrations` directory with a demo Github integration.                                                                        | Generates `models.ts`, but not the compiled `.js` files.                                             |
+| `nango init`                   | Creates an integrations directory with a demo Github integration.                                                                        | Generates `models.ts`, but not the compiled `.js` files.                                             |
```
TBonnin: `nango init` without additional param doesn't create any directory. Should it be `nango init DIR | Creates an DIR  integrations directory....`

TBonnin (approved):

docs-v2/reference/cli.mdx
nalanj: Good call. Updated in ac94019b7.

packages/cli/lib/services/init.service.ts
bodinsamuel: I was just referring to the gitignore, I was perfectly okay with the rest au contraire 👌🏻 sorry for the confusion
I need to be more careful when suggesting, but you don't have to revert when you get a suggestion ☺️

nalanj: 😄 it was trivial and I was apathetic, so figured why not.

khaliqgant (approved):

PR #3354, opened 2025-01-23
fix: first requests to runner fails when new runner comes online

TBonnin:
Network configuration of the newly created service in Render can take up to 30s. 
This commit ensures the runner is reachable from the outside before it registers

### DIFF
--- a/packages/runner/lib/app.ts
+++ b/packages/runner/lib/app.ts
@@ -10,8 +10,13 @@ const providersMonitorCleanup = await monitorProviders();
 try {
     const port = parseInt(process.argv[2] || '') || 3006;
     const id = process.argv[3] || process.env['RUNNER_ID'] || 'unknown-id';
-    const srv = server.listen(port, () => {
+    const srv = server.listen(port, async () => {
         logger.info(`🏃‍♀️ '${id}' ready at http://localhost:${port}`);
+
+        const res = await register();
+        if (res.isErr()) {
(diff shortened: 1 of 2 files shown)


packages/runner/lib/app.ts
```diff
@@ -10,8 +10,13 @@ const providersMonitorCleanup = await monitorProviders();
...
     const port = parseInt(process.argv[2] || '') || 3006;
     const id = process.argv[3] || process.env['RUNNER_ID'] || 'unknown-id';
-    const srv = server.listen(port, () => {
+    const srv = server.listen(port, async () => {
         logger.info(`🏃‍♀️ '${id}' ready at http://localhost:${port}`);
+
+        const res = await register();
+        if (res.isErr()) {
+            logger.error(`${id} Unable to register`, res.error);
+        }
```
bodinsamuel, line 19: Worth investigating how to kill orphan runners

bodinsamuel (approved):

packages/runner/lib/app.ts
TBonnin: as mentioned offline, a runner that cannot register will be mark as failed after 5mins. Before marking it as failed, jobs is going to make an attempt at terminating the node. https://github.com/NangoHQ/nango/blob/6214a69e403bdd2985f8c779da97cff1bc7d9e9f/packages/fleet/lib/supervisor/supervisor.ts#L368

PR #3357, opened 2025-01-23
feat(integrations): add support for recharge

hassan254-prog:
##  Describe the problem and your solution

- Add support for Recharge


hassan254-prog:
I couldn't add a verification method in here since the `api token` is scoped. Tested this locally and it works as expected.

khaliqgant (approved):

PR #3361, opened 2025-01-23
fix(proxy): apply query params outside axios

bodinsamuel:
## Changes 

Fixes https://linear.app/nango/issue/NAN-2574/axios-params-dont-get-passed-correctly-when-using-params-with-netsuite

We had issues with Axios not serializing query params correctly.

- Apply query params outside Axios
In the `proxy.service` and the node client it's handled similarly. 

- Removed `paramsSerializer` 
It seems unused (not in our codebase, not in the templates and not in customers) but hard to check everything


khaliqgant:
@bodinsamuel can this be announced as a breaking change and then this released whenever possible? It is blocking for a prospect

packages/node-client/lib/index.ts
```diff
@@ -834,11 +834,18 @@ export class Nango {
...
-        }
-
-        if (config.paramsSerializer) {
-            options.paramsSerializer = config.paramsSerializer;
+            if (typeof config.params === 'string') {
+                if (url.includes('?')) {
+                    throw new Error('Can not set query params in endpoint and in params');
+                }
+                url = new URL(`${url}${config.params.startsWith('?') ? config.params : `?${config.params}`}`).href;
+            } else {
```
TBonnin, line 842: can `config.params` be neither a string nr an object?

packages/node-client/lib/index.ts
```diff
@@ -834,11 +834,18 @@ export class Nango {
...
-            options.params = config.params;
-        }
-
-        if (config.paramsSerializer) {
-            options.paramsSerializer = config.paramsSerializer;
+            if (typeof config.params === 'string') {
+                if (url.includes('?')) {
+                    throw new Error('Can not set query params in endpoint and in params');
+                }
+                url = new URL(`${url}${config.params.startsWith('?') ? config.params : `?${config.params}`}`).href;
```
TBonnin, line 841: this assumes url is a valid url and config.params is a valid query string. Is that always true?

TBonnin (approved):
code looks good. 
I am worried about some weird edge cases that wouldn't be handled though. I wonder if we could feature flag the change and ramp it up slowly with some proper monitoring

packages/node-client/lib/index.ts
bodinsamuel: I guess it can if they don't respect the types. I would trust the types until we talk about adding runtime type checking

packages/node-client/lib/index.ts
bodinsamuel: Both can be invalid I guess. This new URL will throw if invalid so that's covered. For query params, it's harder to check since almost anything can be a valid query param. 

Do you have a suggestion?

TBonnin: my concern is: is this code behave the same as before if url or query params are invalid? if same behavior as axios I am fine with it

bodinsamuel: it's slightly different on handling query params (otherwise this PR would not exist),
Axios is already sending an invalid url error so this part is similar

bodinsamuel:
@khaliqgant nobody came forward, will merge and publish tomorrow morning

khaliqgant:
> @khaliqgant nobody came forward, will merge and publish tomorrow morning

Ok thanks!

PR #3363, opened 2025-01-24
feat(integrations): add support for salesforce cdp

hassan254-prog:
Add support for Saleforce Data Cloud

Here’s a more grammatically polished version:

This provider supports the `TWO_STEP` authentication mode but requires some changes to fully implement it. Specifically, a second request is needed after obtaining the initial `access_token` from Salesforce to generate the Salesforce Data Cloud (CDP) `access_token`. Since I couldn't test this locally with Data Cloud, I created a mock server to simulate responses from both Salesforce and Data Cloud, similar to the approach with Perimeter81. The process of generating the encoded JWT (assertion) is somewhat technical for users, so I also included a script to enable offline generation, as we did with SharePoint Online v1.


TBonnin (commented):
Yet another auth flow :/ 
I wish that instead of TWO_STEP we introduced a MULTI_STEP mode that takes a list of request params because now we have a TWO_STEPS provider which has 3 steps :(
Is it too late to refactor TWO_STEP? @hassan254-prog

hassan254-prog:
> Yet another auth flow :/ I wish that instead of TWO_STEP we introduced a MULTI_STEP mode that takes a list of request params because now we have a TWO_STEPS provider which has 3 steps :( Is it too late to refactor TWO_STEP? @hassan254-prog

`TWO_STEP` has proven to be quite useful for some of the more complex authentication modes from the provider 😅 🙌 . No, it’s not too late to refactor, as the main logic revolves around how we save the `access_token` and retrieve the `expireAt` value from the response. So, we should be fine to proceed with the refactor. However, I will need to review and test this further on my end. Do we anticipate having a provider with more than three steps in the future? Or work with the max current steps we have at the moment, three.

TBonnin:
> Do we anticipate having a provider with more than three steps in the future? Or work with the max current steps we have at the moment, three.

I stopped trying to anticipate how providers implement authentication. 😆  So many slightly different implementations it is hard to foresee anything. 
I don't know how busy you are and how urgent it is to support this flow. I'll let you and @khaliqgant decide about the necessity of the refactor.

packages/providers/providers.yaml
```diff
@@ -6192,6 +6192,47 @@ salesforce-experience-cloud:
...
+        content-type: application/x-www-form-urlencoded
+    second_request_params:
+        body_format: form
+        token_params:
+            grant_type: urn:salesforce:grant-type:external:cdp
+            subject_token: ${tokenResponse.access_token}
+            subject_token_type: urn:ietf:params:oauth:token-type:access_token
+        token_headers:
+            content-type: application/x-www-form-urlencoded
+        token_url: ${tokenResponse.instance_url}/services/a360/token
```
bodinsamuel: is it a second request or a third one?
We can't change the name of the auth but we can still make that more flexible
suggestion
```yaml
auth_mode: TWO_STEP
additional_steps:
  - body_format: form
    token_params: 
    [...]
```

So we can have N ordered steps in case we have a 4steps providers coming

hassan254-prog: It is a second_request that uses the `access_token` from the first response to now generate the final `access_token` to be used when making proxy requests within Nango.
Yes that is 💯  perfect, dynamically building this will minimize limitations of adding new providers in the future. 

cc @khaliqgant

khaliqgant: Please proceed with this Kelvin after netsuite is done

hassan254-prog: Thank you @bodinsamuel, @TBonnin  and @khaliqgant ! 🙌  `TWO_STEP` is now fully dynamic and robust, it can handle any number of additional steps before parsing the `access_token`. The additional steps are processed sequentially, picking parameters from any step as needed and in the end parse the token response from the last step. However, I haven’t tested this with Salesforce since we don’t have a test account. Instead, I set up mock servers to simulate the flow and successfully tested up to four steps. The solution is now solid. 🚀  An example of a more complex step if there need be:

```
salesforce-cdp:
    display_name: Salesforce (Data Cloud)
    categories:
        - storage
    auth_mode: TWO_STEP
    token_url: http://localhost:3090/first-server
    body_format: form
    token_params:
        grant_type: urn:ietf:params:oauth:grant-type:jwt-bearer
        assertion: ${credential.jwt}
    token_headers:
        content-type: application/x-www-form-urlencoded
    additional_steps:
        - body_format: form
          token_params:
            grant_type: urn:salesforce:grant-type:external:cdp
            subject_token: ${step1.access_token}
            subject_token_type: urn:ietf:params:oauth:token-type:access_token
          token_headers:
            content-type: application/x-www-form-urlencoded
          token_url: ${step1.instance_url}/services/a360/token
        
        - body_format: form
          token_params:
            grant_type: urn:salesforce:grant-type:external:cdp
            subject_token: ${step2.access_token}
            subject_token_type: urn:ietf:params:oauth:token-type:access_token
          token_headers:
            content-type: application/x-www-form-urlencoded
          token_url: ${step2.instance_url}/services/a360/token

        - body_format: form
          token_params:
            grant_type: urn:salesforce:grant-type:external:cdp
            subject_token: ${step3.access_token}
            subject_token_type: urn:ietf:params:oauth:token-type:access_token
          token_headers:
            content-type: application/x-www-form-urlencoded
          token_url: ${step1.instance_url}/services/a360/token

    token_response_metadata:
        - instance_url
    proxy:
        base_url: ${connectionConfig.instance_url}
    token_response:
        token: access_token
        token_expiration: expires_in
        token_expiration_strategy: expireIn
    docs: https://docs.nango.dev/integrations/all/salesforce-cdp
    docs_connect: https://docs.nango.dev/integrations/all/salesforce-cdp/connect
    credentials:
        jwt:
            type: string
            title: Encoded JWT
            description: This is your pre-generated, encoded JSON Web Token (JWT)
            secret: true
            doc_section: '#step-1-generating-your-encoded-jwt'

File: packages/shared/lib/services/connection.service.ts [hunk 2]
```diff
@@ -1769,7 +1772,59 @@ class ConnectionService {
                 responseData = parser.parse(response.data);
             }
 
-            const parsedCreds = this.parseRawCredentials(responseData, 'TWO_STEP', provider) as TwoStepCredentials;
+            const stepResponses: any[] = [responseData];
+            if (provider.additional_steps) {
+                for (let stepIndex = 1; stepIndex <= provider.additional_steps.length; stepIndex++) {
```
bodinsamuel, line 1796: why not simply start at 0?

packages/shared/lib/utils/utils.ts
```diff
@@ -220,6 +221,39 @@ export function stripCredential(obj: any): any {
     return obj;
 }
 
+export function stripStepesponse(obj: any, step: Record<string, any>): any {
```
bodinsamuel:
```suggestion
export function stripStepResponse(obj: any, step: Record<string, any>): any {
```

packages/providers/providers.yaml
```diff
@@ -6542,7 +6542,44 @@ salesforce-experience-cloud:
...
+        content-type: application/x-www-form-urlencoded
+    additional_steps:
+        - body_format: form
+          token_params:
+              grant_type: urn:salesforce:grant-type:external:cdp
+              subject_token: ${step1.access_token}
+              subject_token_type: urn:ietf:params:oauth:token-type:access_token
+          token_headers:
+              content-type: application/x-www-form-urlencoded
+          token_url: ${step1.instance_url}/services/a360/token
```
bodinsamuel, line 6633: step1 is what in this context?

bodinsamuel (approved):
Auth is definitely becoming harder. Looks good, some open questions

packages/shared/lib/utils/utils.ts
hassan254-prog: updated with the latest commit.

packages/providers/providers.yaml
hassan254-prog: step1 refers to the response from the initial request made before any additional steps, if provided. In subsequent steps, parameters are accessed dynamically using `step{request_number}`.

packages/shared/lib/services/connection.service.ts
hassan254-prog: The loop starts at `1` and accesses `provider.additional_steps[stepIndex - 1]` to align with the step numbering convention used in `providers.yaml`. Starting from `1` ensures consistency in configuration, making it more intuitive for human readability—e.g., referencing `step1` instead of `step0`. This approach maintains a natural mapping between the loop index and the defined steps in `providers.yaml`.

PR #3372, opened 2025-01-27
docs: fix list connections params in Node SDK

bastienbeurier:


docs-v2/reference/sdks/node.mdx
```diff
@@ -348,16 +348,10 @@ await nango.listConnections();
...
     </ResponseField>
     <ResponseField name="search" type="string">
-        Will partially match connection IDs or end user profiles.
-    </ResponseField>
-    <ResponseField name="connectionId" type="string" required>
-        Filter the list of connections based on the given end user id.
-    </ResponseField>
-    <ResponseField name="connectionId" type="string" required>
-        Filter the list of connections based on the given end user's organization id.
+        Optional. Search connections. Will search in connection ID or end user profile.
```
bodinsamuel: Sorry it was a wrong c/c, the third connectionId should be `userId`, and the fourth is `endUserOrganizationId`

bastienbeurier: > Sorry it was a wrong c/c, the third connectionId should be `userId`, and the fourth is `endUserOrganizationId`

Is it better like this? Is redundancy between `search` and `queries` needed? Also, does it make sense to nest the end user/org in `queries` (vs. top-level parameters)?

bodinsamuel: I mean this does not fit the actual shape of the HTTP API so as you want but some people might be confused. NB: All the params here are in the query params

bastienbeurier: > I mean this does not fit the actual shape of the HTTP API so as you want but some people might be confused. NB: All the params here are in the query params

Sorry, I don't understand.

I was asking 3 separate questions: 
- Is what I documented accurate (in the Node SDK)? 
- Is there a need for a `search` param and a `queries` param (in the Node SDK)?
- Should the user/org params be nested or top level (in the Node SDK)?

bodinsamuel: tl;dr last commit is correct, my bad for the confusion

Yes actually nevermind
- you are absolutely right, it needs to be in `queries`. 
- search and queries are just a legacy issue (they are all in query params but instead of making the signature grow too much we decided to shove the new params in a `queries` object)
- they are in queries, you were right 
https://github.com/NangoHQ/nango/blob/a5b0eb9cbc8a2f1a6d00b71af048e0c02c77945f/packages/node-client/lib/index.ts#L276-L280

bastienbeurier: Great, thanks! Let me know if there's any other change to make to the PR or if it's good to be approved.

bodinsamuel (approved):
🚀

PR #3378, opened 2025-01-27
feat(integrations): add support for drupal

hassan254-prog:
Add support for Drupal. Drupal returns the `expire_in` value as `human_readable`  format i.e `60 minutes`, need a way to support that.


packages/shared/lib/services/connection.service.ts
```diff
@@ -1127,8 +1128,23 @@ class ConnectionService {
...
+                                throw new NangoError(`Unsupported human-readable unit: ${unit}`);
+                            }
+                            expiresAt = new Date(Date.now() + Number.parseInt(value, 10) * multiplier);
+                        } else {
+                            throw new NangoError(`Unsupported expiration unit: ${expirationUnit}`);
+                        }
+                    } else {
+                        const expiresIn = Number.parseInt(expiration, 10);
+                        expiresAt = new Date(Date.now() + expiresIn);
+                    }
```
TBonnin, line 1139: instead of introduce a new `expires_in_unit` can we simply check if the value is an integer. If not we fallback to parsing the human readable format. We should also use `ms` for the parsing. It is already a dependency of `shared`
```ts
if(Number.isSafeInteger(expiration)) {
   expiresAt = new Date(Date.now() + expiresIn * 1000);
} else {
   expiresAt = ms(expiration)
}
```

packages/providers/providers.yaml
```diff
@@ -2324,6 +2324,60 @@ dropbox:
...
+            secret: true
+            doc_section: '#step-3-generating-your-drupal-client-credentials'
+        userName:
+            type: string
+            title: User Name
+            description: Your Drupal User Name
+            doc_section: '#step-2-finding-your-drupal-user-name'
+    connection_config:
+        baseUrl:
+            type: string
```
bodinsamuel, line 2389: could be format: hostname

packages/providers/providers.yaml
```diff
@@ -2324,6 +2324,60 @@ dropbox:
...
+            title: Base Url
+            description: Your Drupal base url
+            example: example.drupal.com
```
bodinsamuel, line 2392: I suppose it's to accommodate self-hosted?

bodinsamuel (approved):

packages/shared/lib/services/connection.service.ts
bodinsamuel: agreed 👍🏻

hassan254-prog: thanks , this simplifies the logic 🙏 , updated with the latest [commit](https://github.com/NangoHQ/nango/pull/3378/commits/8bda81c938e9659562d10c5abbbd0245bce0c0ca)

packages/providers/providers.yaml
hassan254-prog: thanks @bodinsamuel, updated with the latest commit.

packages/providers/providers.yaml
hassan254-prog: yes to accommodate self hosting

PR #3380, opened 2025-01-27
fix(node-client): updateSyncConnectionFrequency parameters

TBonnin:
should error if sync or connectionId is NOT a string


bodinsamuel (approved):
Fixing a bug exactly 1
year on the clock after I wrote it ✌🏻

PR #3383, opened 2025-01-28
chore: improve publish script

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2621/fix-npm-publish-script

- Fix publish script
I haven't found why but `npm install` behaved differently between my terminal and the bash script, no matter what I did I couldn't make the `npm install` fail or succeed appropriately.
Decided to go all in, rewrite this shit in Javascript, and bump packages using the good old regex replace.

Incidentally, it's way faster now, 
before: 5m30
after: 3m30

Hopefully, it works the first time 🤞🏻


scripts/publish.mjs
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        await $`npm version ${nextVersion} -w "${packageName}"`;
+        await $`npm publish --access public -w "${packageName}"`;
+
+        echo(chalk.green(`${figures.tick} Published ${packageName}      `));
+    });
+}
+
+async function bumpPackageFor(packageName, folders) {
```
TBonnin: we could be smarter here and find all the packages.json in `packages` subfolders and do a `replace(new RegExp(`${packageName}": "${versionRegex}"`), `${packageName}": "${nextVersion}"`)`. 
Then we don't have to pass the folders list and modify this script when adding new packages

TBonnin (approved):
I was wondering how long until you'd rewrite it in ts 😝

scripts/publish.mjs
```diff
@@ -0,0 +1,136 @@
...
+            echo`${chalk.red(`An error occurred`)}`;
+            console.log(res);
+            echo`${res.stderr}`;
+            process.exit(1);
+        }
+
+        await $`npm version ${nextVersion} -w "${packageName}"`;
+        await $`npm publish --access public -w "${packageName}"`;
+
+        echo(chalk.green(`${figures.tick} Published ${packageName}      `));
```
nalanj, line 118: Just curious - why the extra whitespace here?

nalanj (approved):

scripts/publish.mjs
bodinsamuel: the spinner text is kept when we echo inside, so it output `old[newtext]text` (if you know what I mean)
they just want us to log outside the spinner, which is annoying when you have multiple outcomes, decided it was an okay fix :p

scripts/publish.mjs
bodinsamuel: true, let me check that

bodinsamuel: done in 65851d7bc8fbf20785f94a93912b2ba2dc26c88c

PR #3385, opened 2025-01-28
fix(tests): Fix a few flaky tests

nalanj:
Went through and fixed some tests with timing issues that I've run into issues with here lately.


packages/orchestrator/lib/clients/processor.ts
```diff
@@ -41,6 +41,10 @@ export class OrchestratorProcessor {
         }
     }
 
+    public queueSize() {
```
nalanj, line 44: I don't love exposing this but it's the easiest way to help the test know if the queue has emptied.

bodinsamuel (approved):
setTimeout... the inevitable evolution of all integration tests :D

PR #3386, opened 2025-01-28
feat(persist): Deep merge records in batchUpdate

nalanj:
Currently `batchUpdate` only shallow merges. This switches it to deep merge, as request by some customers.

I want to get @bastienbeurier's take on docs strategy here, but code is ready for review.

See https://linear.app/nango/issue/NAN-1982/partial-records-update

## How I tested it

- Wrote tests
- Wrote a sync that dumped a lot of objects into the store and then used `batchUpdate` in a subsequent run to merge them.


packages/records/lib/models/records.ts
```diff
@@ -20,6 +20,7 @@ import { Err, Ok, retry, stringToHash } from '@nangohq/utils';
 import type { Result } from '@nangohq/utils';
 import type { Knex } from 'knex';
 import { Cursor } from '../cursor.js';
+import merge from 'lodash-es/merge.js';
```
TBonnin: not a huge fine of lodash. It is huge and haven't seen a release in years. If we don't want to rollout our own merge function can we find a smaller lib?

nalanj: Yep, fair. I used it because we had lodash elsewhere, but I'll dig around for a better option.

packages/fleet/lib/supervisor/supervisor.integration.test.ts
```diff
@@ -59,7 +59,7 @@ describe('Supervisor', () => {
             void supervisor2.start();
 
             await vi.waitUntil(() => tickSpy1.mock.calls.length > 0, {
-                timeout: 5000
+                timeout: 10000
```
nalanj, line 62: Had this one fail on me in dev again, so I went ahead and bumped it up a good bit more.

packages/records/lib/models/records.ts
nalanj: See 51b0cd166 - switched to using `@fastify/deepmerge`. It's probably better overall since it'll let us tweak array merging as well.

packages/records/lib/models/records.integration.test.ts
```diff
@@ -211,6 +211,55 @@ describe('Records service', () => {
...
+                person: {
+                    name: 'John Doe',
+                    age: 36,
+                    children: [
+                        { name: 'Jenny Doe', age: 3 },
+                        { name: 'Maurice Doe', age: 1 }
+                    ]
+                }
+            });
+        });
```
bodinsamuel, line 272: Would love more test:
- where you swap array items
- where you add an item 
- where you remove an item

bodinsamuel (commented):
Looks good and simple 👍🏻 
I'm not sure it's really clear for an end user, there is no way to optin or optout, this might break some expectation (or be unnoticed)

packages/records/lib/models/records.integration.test.ts
nalanj: :+1: I'll probably extract the merge code to another function and unit test that, just so it's not so heavy a test suite there, if that's cool.

TBonnin (approved):

PR #3389, opened 2025-01-28
feat(persist): add support for merging strategy

TBonnin:
- add an endpoint to fetch the last cursor for a given connection/model
- pass merging strategy to records.upsert/update
- modify records.upsert/update to return the next merging strategy

Next step will be to 
- add a `setMergingStrategy()` function to the runner SDK. This function would potentially fetches the last cursor.
- modify `batch...` functions to send the merging strategy and update its current value from persist response.
- 
Note: I am gonna wait for you @bodinsamuel to merge the runner SDK changes

### DIFF
--- a/packages/persist/lib/routes/environment/environmentId/connection/connectionId/sync/syncId/job/jobId/deleteRecords.ts
+++ b/packages/persist/lib/routes/environment/environmentId/connection/connectionId/sync/syncId/job/jobId/deleteRecords.ts
@@ -2,6 +2,7 @@ import type { ApiError, Endpoint } from '@nangohq/types';
 import type { EndpointRequest, EndpointResponse, RouteHandler, Route } from '@nangohq/utils';
 import { persistRecords, recordsPath } from '../../../../../../../../../records.js';
 import { validateRecords } from './validate.js';
+import type { MergingStrategy } from '@nangohq/records';
 
 type DeleteRecords = Endpoint<{
     Method: typeof method;
@@ -18,9 +19,12 @@ type DeleteRecords = Endpoint<{
         providerConfigKey: string;
         connectionId: string;
         activityLogId: string;
+        merging: MergingStrategy;
     };
     Error: ApiError<'delete_records_failed'>;
-    Success: never;
+    Success: {
+        nextMerging: MergingStrategy;
+    };
 }>;
 
 const path = recordsPath;
@@ -31,7 +35,7 @@ const validate = validateRecords<DeleteRecords>();
 const handler = async (req: EndpointRequest<DeleteRecords>, res: EndpointResponse<DeleteRecords>) => {
     const {
         params: { environmentId, nangoConnectionId, syncId, syncJobId },
-        body: { model, records, providerConfigKey, connectionId, activityLogId }
+        body: { model, records, providerConfigKey, connectionId, activityLogId, merging }
     } = req;
     const result = await persistRecords({
         persistType: 'delete',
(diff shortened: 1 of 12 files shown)


packages/persist/lib/routes/environment/environmentId/connection/connectionId/sync/syncId/job/jobId/deleteRecords.ts
```diff
@@ -43,10 +47,11 @@ const handler = async (req: EndpointRequest<DeleteRecords>, res: EndpointRespons
         syncJobId,
         model,
         records,
-        activityLogId
+        activityLogId,
+        merging
     });
     if (result.isOk()) {
-        res.status(204).send();
+        res.status(200).send({ nextMerging: result.value });
```
TBonnin, line 54: runner only cares about status > 299 so there is no issue changing the status code and the response content

packages/records/lib/models/records.ts
```diff
@@ -441,31 +459,53 @@ export async function update({
...
+                            {
+                                external_id: string;
+                                id: string;
+                                last_modified_at: string;
+                            }[]
+                        >(
+                            trx.raw(`
+                            upsert.id as id,
+                            upsert.external_id as external_id,
+                            to_json(upsert.updated_at) as last_modified_at`)
```
TBonnin, line 489: I must add a CTE with an extra select because we need `to_json(upsert.updated_at)` to construct the cursor value

packages/records/lib/models/records.ts
```diff
@@ -313,19 +321,23 @@ export async function upsert({
...
+                            CASE
+                                WHEN existing.external_id IS NULL THEN 'inserted'
+                                ELSE
+                                    CASE
+                                        WHEN existing.deleted_at IS NOT NULL AND upsert.deleted_at IS NULL THEN 'undeleted'
+                                        WHEN existing.deleted_at IS NULL AND upsert.deleted_at IS NOT NULL THEN 'deleted'
+                                        WHEN existing.data_hash <> upsert.data_hash THEN 'changed'
+                                        ELSE 'unchanged'
+                                    END
+                            END as status`)
```
TBonnin, line 340: returning `id` and `last_modified_at` in order to be able to construct a cursor

packages/persist/lib/server.ts
```diff
@@ -20,7 +21,7 @@ if (process.env['ENABLE_REQUEST_LOG'] !== 'false') {
     server.use(requestLoggerMiddleware({ logger }));
 }
 
-server.use('/environment/:environmentId/*', authMiddleware);
+server.use('/environment/:environmentId/*', authMiddleware, express.json());
```
bodinsamuel, line 24: Not sure about the default, I think it might be 1mb, you could put a lower limit that fits those endpoints

bodinsamuel (approved):
Looks good
(just asking to be sure I understood) Does that means we'll need to call the cursor endpoint at the beginning of every sync?

nalanj (approved):

TBonnin:
> Looks good (just asking to be sure I understood) Does that means we'll need to call the cursor endpoint at the beginning of every sync?

I am gonna add a `setMergingStrategy(...)` function to the runner SDK. This function will indeed call the `/cursor` endpoint. When customer wants records not to be overwritten, they will have to call this function explicitly at the beginning of the sync (ie: before batchSave). If this function is not called, there would be no extra request to `persist` and things will behave exactly as it does currently

packages/persist/lib/server.ts
TBonnin: default is 100kb http://expressjs.com/uk/api.html#express.json
I think this value is fine

PR #3396, opened 2025-01-29
docs: default to dark theme

bastienbeurier:


bodinsamuel (approved):
Not approving that 👎🏻 🙄

bastienbeurier:
> Not approving that 👎🏻 🙄

I think it was already the case, just making it explicit :D

bastienbeurier:
> Not approving that 👎🏻 🙄

My bad, it was set to the user's operating system settings before.

PR #3397, opened 2025-01-29
fix(jobs): wait until runner is reachable before registering

TBonnin:
Network configuration can take a long time to be applied and propagate to jobs when a new runner spins up.
This commit ensures that when jobs receives a registrating request from a runner it wait until it becomes reachable before to proceed with the registration.


packages/fleet/lib/fleet.ts
```diff
@@ -126,6 +127,12 @@ export class Fleet {
         if (valid.isErr()) {
             return Err(valid.error);
         }
+        // in Render, network configuration can take a long time to be applied and accessible to other services
+        // we therefore wait until the health url is reachable
+        const healthy = await waithUntilHealthy({ url: `${url}/health`, timeoutMs: 120_000 });
```
bodinsamuel: Do you think it's worth putting that in an env var?

packages/runner/lib/register.ts
```diff
@@ -12,27 +11,6 @@ export async function register(): Promise<Result<void>> {
...
-                } else {
-                    await setTimeout(waitMs);
-                }
-            } catch {
-                await setTimeout(waitMs);
-            }
-        }
-        if (!isReachable) {
-            return Err(new Error(`Runner at ${envs.RUNNER_URL} is not reachable after ${timeoutMs}ms`));
-        }
```
TBonnin, line 35: we were previously waiting in the runner itself but it looks like it is possible for the runner to reach itself before it is actually reachable from jobs :(

packages/fleet/lib/utils/url.ts
```diff
@@ -0,0 +1,21 @@
+import type { Result } from '@nangohq/utils';
+import { Err, Ok } from '@nangohq/utils';
+import { setTimeout } from 'node:timers/promises';
+
+export async function waithUntilHealthy({ url, timeoutMs }: { url: string; timeoutMs: number }): Promise<Result<void>> {
+    const startTime = Date.now();
+    const waitMs = 1000;
+    while (Date.now() - startTime < timeoutMs) {
+        try {
+            const res = await fetch(url);
```
bodinsamuel, line 10: Hopefully, nodejs does not cache dns failure that could slow down this check (but I think they are notà

bodinsamuel (approved):
Let's try 👍🏻

bodinsamuel (approved):

PR #3398, opened 2025-01-29
docs: add changelog assets

bastienbeurier:


nalanj (approved):
Looks great. I really wish each month could be listed in the sidebar but I'm pretty sure Mintlify just won't do that.

bastienbeurier:
> Looks great. I really wish each month could be listed in the sidebar but I'm pretty sure Mintlify just won't do that.

It is on the right side bar, but it's not super discoverable:
CleanShot 2025-01-29 at 16 50 56@2x

nalanj:
@bastienbeurier Oh awesome - I think I missed that because it's responsive and my browser was hanging out a little too narrow.

PR #3399, opened 2025-01-29
chore(logs): Bump elasticsearch to 8.17.0

nalanj:
There was an issue with a memory leak before 8.15.0 that could map to what we're seeing on server crashes. I'm skeptical but it's worth an upgrade.

Logging tests all passed and local runs seemed to log correctly.


bodinsamuel (approved):
👍🏻

TBonnin (approved):

PR #3400, opened 2025-01-29
feat(zapier-scim): add integration guide and configuration for Zapier SCIM

viictoo:
Add integration guide and configuration for Zapier SCIM

[EXT-290](https://linear.app/nango/issue/EXT-290/zapier-user-provisioning)


docs-v2/integrations/all/zapier-scim/connect.mdx
```diff
@@ -0,0 +1,65 @@
...
+2. Under COMPANY SETTINGS, select **User provisioning**.
+<img src="/integrations/all/zapier/zapier-user-provisioning.png" />
+3. Copy the **SCIM Bearer Token**.
+<img src="/integrations/all/zapier-scim/zapier-api-token.png" />
+
+**Important**: Store your API token securely, as it will be required to connect your Zapier account.
+
+---
+
+## Step 2: Base URL for SCIM API Integration
```
khaliqgant: You can remove step 2 since this is set by Nango so a user won't have to worry about this. You can remove step 3 as well since Nango handles this part.

khaliqgant (changes requested):
Documentation update requested

docs-v2/integrations/all/zapier-scim/connect.mdx
viictoo: updated.

packages/providers/providers.yaml
```diff
@@ -7987,6 +7987,30 @@ zapier-nla:
...
+        verification:
+            method: GET
+            endpoint: /Users
+    docs: https://docs.nango.dev/integrations/all/zapier
+    credentials:
+        apiKey:
+            type: string
+            title: SCIM API Token
+            description: The SCIM API token generated from the Provision tab in your Zapier account.
+            doc_section: '#step-1-generating-your-scim-api-token'
```
khaliqgant: what is this link now? There is no longer a step one right?

viictoo: yes. thats true. updated as well

khaliqgant (approved):

PR #3401, opened 2025-01-29
fix(persist): fix express.json limit

TBonnin:
bug introduced here https://github.com/NangoHQ/nango/commit/fa17d3f72eb25f0658e1c8a471aeeeb47b8396f0#diff-4608f3177ceaf5a108fff1fe6472895820f0344f0ba9cee44979155b65a949e3R24

applying the default 100kb limit to all endpoints and preventing big records to be persisted

### DIFF
--- a/packages/runner/lib/sdk/sdk.ts
+++ b/packages/runner/lib/sdk/sdk.ts
@@ -324,14 +324,8 @@ export class NangoSyncRunner extends NangoSyncBase {
                     this.stringify()
                 );
 
-                if (response.status === 400) {
-                    throw new Error(
-                        `Records invalid format. Please make sure you are sending an array of objects that each contain an 'id' property with type string`
-                    );
-                } else {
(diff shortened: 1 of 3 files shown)


nalanj (approved):

khaliqgant (approved):

packages/runner/lib/sdk/sdk.ts
```diff
@@ -324,14 +324,8 @@ export class NangoSyncRunner extends NangoSyncBase {
                     this.stringify()
                 );
 
-                if (response.status === 400) {
```
TBonnin, line 327: the 400 is coming from there https://github.com/NangoHQ/nango/blob/master/packages/persist/lib/server.ts#L42-L45
Not sure why it is handled specifically here and even less sure why the error message doesn't match with the error.
just removing the special case and displaying the error provided by persist

bodinsamuel: I added that so customers could understand a bit better the validation error when they would send invalid records.
When I did it there was no other 400 I think, but should have checked the code inside the payload

PR #3403, opened 2025-01-29
feat(docs): Backfill missing docs due to multiple providers for same API

nalanj:
Backfills provider docs that are missing from our prior decision to not create multiple docs when we had multiple providers for single apis.

Removes the `provider` gray matter in each of the provider mdx files since we can now just rely on the filename for the provider.

See https://linear.app/nango/issue/NAN-2581/improve-api-docs-page-with-multiple-providersyaml-entries


bastienbeurier (approved):
This looks all good to me even if it's hard to check all the details. Great step and makes the catalog more complete!

bodinsamuel (approved):

khaliqgant (approved):

PR #3404, opened 2025-01-30
fix(pagination): add flag to optionally add params to body

hassan254-prog:
Some providers require pagination parameters to be passed as query parameters for methods other than `GET`. This PR introduces a flag in the `proxyConfiguration` that, when set to `false`, ensures pagination parameters are included in the request body for `POST`, `PUT`, and `PATCH` requests, preserving the original behavior.


packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
```diff
@@ -81,6 +81,7 @@ interface Pagination {
     limit?: number;
     response_path?: string;
     limit_name_in_request: string;
+    pass_pagination_params_in_body?: boolean;
```
bodinsamuel: Might be possible to choose a simpler name since this is already in the Pagination interface, wdyt?
```suggestion
    format?: 'body';
    in?: 'body';
    in_body?: true;
```

bodinsamuel (approved):
Did you make sure it's actually possible to send a Body with GET?

packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
hassan254-prog: Yes makes sense to have a simpler name, `in_body` is much simpler. Updated with the latest commit.

hassan254-prog:
> Did you make sure it's actually possible to send a Body with GET?

hmmm, it's generally not a good practice to include a body in a `GET` request, as `GET` is intended to be primarily used for retrieving data without side effects. The flag `in_body` is only designed to work with `POST`, `PUT`, and `PATCH` requests. Ultimately, it's the responsibility of the engineer implementing the integration to ensure that this flag is not set for `GET` requests.

bodinsamuel:
Ha I see, I misread the PR desc all good 👍🏻

khaliqgant (approved):

PR #3406, opened 2025-01-30
fix: eslint pass

bodinsamuel:
## Changes

- Fix more rules


eslint.config.mjs
```diff
@@ -161,6 +159,8 @@ export default tseslint.config(
             '@typescript-eslint/no-deprecated': 'warn',
             '@typescript-eslint/no-floating-promises': 'warn',
             '@typescript-eslint/no-unsafe-assignment': 'warn',
+            '@typescript-eslint/no-non-null-assertion': 'warn',
+            '@typescript-eslint/no-explicit-any': 'warn',
```
TBonnin, line 163: 👌

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -27,8 +26,8 @@ interface RunArgs extends GlobalOptions {
     connectionId: string;
     lastSyncDate?: string;
     useServerLastSyncDate?: boolean;
-    input?: unknown;
-    metadata?: Metadata;
+    input?: string;
+    metadata?: string;
```
TBonnin, line 30: was the type wrong before?

packages/webapp/src/pages/Environment/Settings.tsx
```diff
@@ -310,7 +310,7 @@ export const EnvironmentSettings: React.FC = () => {
             const [[, header], [, value]] = entries.slice(i, i + 2);
             newOtlpHeaders = {
                 ...newOtlpHeaders,
-                [header.toString()]: value
+                [JSON.stringify(header)]: value
```
TBonnin: why `stringify`? can header be some json value?

TBonnin (approved):
thanks for doing all of this cleanup

packages/cli/lib/services/dryrun.service.ts
bodinsamuel: yes we receive string since it comes from the CLI params

packages/webapp/src/pages/Environment/Settings.tsx
bodinsamuel: nope not JSON, but it can technically be other stuff like number or symbol. Now that I think about, technically the proper fix is to do an `if type === string`

nalanj (approved):
🏃🏻‍♂️

PR #3410, opened 2025-01-30
fix(api): enforce content-type when possible

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2578/enforce-applicationjson-header-for-regulard-api-endpoints

- Enforce content-type when possible
I specifically did not enforce that on the proxy. Tested everything I could, and hopefully haven't missed anything

## 🧪Tests

- Run the dashboard UI and do some stuff
- Call the public API

> It's not a breaking change per say, since it would just not work before but I will still announce that in the community before merging


nalanj (approved):

bodinsamuel:
Can you have a look @TBonnin?

packages/server/lib/middleware/json.middleware.ts
```diff
@@ -0,0 +1,11 @@
+import type { RequestHandler } from 'express';
+
+export const jsonContentTypeMiddleware: RequestHandler = (req, res, next) => {
+    if (req.headers['content-type'] && req.headers['content-type'] !== 'application/json') {
+        // Send error here
+        res.status(415).json({ error: { code: 'invalid_content_type', message: 'Header Content-Type must be application/json' } });
```
TBonnin: nit: 'Content-type header must be ...' no?

packages/server/lib/routes.integration.test.ts
```diff
@@ -10,6 +10,35 @@ describe('route', () => {
     afterAll(() => {
         api.server.close();
     });
+
+    describe('Content-type', () => {
+        it.each(['GET', 'POST'] as const)('should enforce content-type %s', async (val) => {
```
TBonnin, line 15: we support GET with body? 🙃

packages/server/lib/routes.ts
```diff
@@ -192,33 +197,43 @@ publicAPI.route('/auth/bill/:providerConfigKey').post(connectSessionOrPublicAuth
 publicAPI.route('/auth/signature/:providerConfigKey').post(connectSessionOrPublicAuth, postPublicSignatureAuthorization);
 publicAPI.route('/auth/unauthenticated/:providerConfigKey').post(connectSessionOrPublicAuth, postPublicUnauthenticated);
 
+publicAPI.use('/unauth', jsonContentTypeMiddleware);
 // @deprecated
 publicAPI.route('/unauth/:providerConfigKey').post(connectSessionOrPublicAuth, postPublicUnauthenticated);
 
+publicAPI.use('/webhook', jsonContentTypeMiddleware);
```
TBonnin: unfortunately it looks like not all webhooks send json :( https://github.com/NangoHQ/nango/blob/b1c0f9a2f44d594dd29153f0c2e8ae72c99b550e/packages/server/lib/webhook/slack-webhook-routing.ts#L8

TBonnin (commented):
an important comment about the /webhook endpoint. The rest lgtm

packages/server/lib/routes.integration.test.ts
bodinsamuel: It was mostly to ensure it works because we are always putting the header no matter the method (in the UI and probably other customers too)

packages/server/lib/routes.ts
bodinsamuel: ah yes thanks for catching that

PR #3412, opened 2025-01-30
fix(ui): properly debounce/abort logs search

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2629/date-filter-issue-when-selecting-past-dates

Screenshot 2025-01-30 at 14 41 03

- Properly debounce/abort logs search
When you start selecting a day in the date picker it triggers an initial search that would be done by the time you finished picking the end day of your date range. Which then triggered a second query but with the previous invalid data not being removed.
I had bad cancellation/concurrency strategies:
- with the `isLoading` debounce which was not triggering some search = not aborting and not fetching
- the signal was passed as reference which could be replaced by another signal controller before being checked

## 🧪 Tests
- Go to UI
- Pick a first day, try to wait 100-200ms before clicking another day (it's not that simple though)
- At the end you should have proper data
e.g: have the last 2 days picked, start clicking Jan 22, wait, then click Jan 23 as the end day. You should not see Jan 30 as the last operation


nalanj (approved):
🍻

PR #3415, opened 2025-01-30
fix(webapp): fix code generation for two_step

hassan254-prog:
The code being generate was missing  the `type: 'TWO_STEP'` property within credentials, which is crucial when creating a new connection using the frontend sdk in `TWO_STEP`.


hassan254-prog:
@NangoHQ/engineers, is there a way to achieve this without explicitly defining the type? I'm open to suggestions on how to handle this, given that `TWO_STEP` accommodates dynamic inputs, unlike the other auth modes. Specifically, we can't rely on checking for a specific field within the `credentials` object to determine the `auth_mode` as we do with the others.

bodinsamuel (approved):

TBonnin (approved):

PR #3420, opened 2025-01-31
feat: add support for merging strategy into the runner sdk

TBonnin:
- adding `setMergingStrategy` runnder-sdk function
    - if strategy is `ignore-if-modified-since` the function calls persist `/cursor` and store it in a private variable. Its value is then forwarded when calling `batchSave/Update/Delete` which are now sending the next cursor. Again the updated cursor is forwarded to batchSave...
    - One limitation of this approach is that it makes concurrent batchSave undefined behavior. We already recommend not to do that anyway because it doesn't play well with rate-limiting
- refactor persist API related code in runner-sdk into its own file
- send merging strategy to persist and handle response with next merging strategy with new cursor
- fix a couple of bugs from previous PR
- modify CLI compiler to disallow setting merging strategy multiple times or setting it after bachSave


packages/cli/lib/services/parser.service.ts
```diff
@@ -70,15 +70,18 @@ class ParserService {
             'triggerAction'
         ];
 
-        const disallowedActionCalls = ['batchSend', 'batchSave', 'batchDelete'];
+        const disallowedActionCalls = ['batchSend', 'batchSave', 'batchDelete', 'batchUpdate'];
```
TBonnin, line 74: unrelated but `batchUpdate` was missing

packages/cli/lib/services/parser.service.ts
```diff
@@ -117,20 +120,25 @@ class ParserService {
...
+                                            `"${
+                                                model.value
+                                            }" is not a valid model name. Please check "${filePath}:${lineNumber}". The possible model names are: ${modelNames.join(
+                                                ', '
+                                            )}`
+                                        )
+                                    );
+                                    usedCorrectly = false;
+                                }
+                            }
```
TBonnin, line 139: adding support for functions like `setMergingStrategy` that take multiple models as argument

packages/persist/lib/routes/environment/environmentId/connection/connectionId/getCursor.ts
```diff
@@ -16,9 +15,7 @@ type GetCursor = Endpoint<{
         offset: CursorOffset;
     };
     Error: ApiError<'get_cursor_failed' | 'cursor_not_found'>;
-    Success: {
-        cursor?: string;
-    };
+    Success: GetCursorSuccess;
```
TBonnin, line 18: moving the type to `types` package so it can be used by the runner-sdk

packages/persist/lib/routes/environment/environmentId/connection/connectionId/sync/syncId/job/jobId/validate.ts
```diff
@@ -8,7 +8,7 @@ const mergingStrategySchema = z.discriminatedUnion('strategy', [
     }),
     z.object({
         strategy: z.literal('ignore_if_modified_after_cursor'),
-        cursor: z.string()
+        cursor: z.string().optional()
```
TBonnin, line 11: forgot the scenario in my previous PR where there is not records yet (first sync or after emptying the cache). In this case there is no cursor yet

packages/persist/lib/routes/environment/environmentId/connection/connectionId/sync/syncId/job/jobId/validate.ts
```diff
@@ -31,8 +32,7 @@ export const validateRecords = <E extends Endpoint<any>>() =>
                     environmentId: z.coerce.number().int().positive(),
                     nangoConnectionId: z.coerce.number().int().positive(),
                     syncId: z.string(),
-                    syncJobId: z.coerce.number().int().positive(),
-                    merging: mergingStrategySchema.default({ strategy: 'override' })
+                    syncJobId: z.coerce.number().int().positive()
```
TBonnin, line 35: 🤦

packages/records/lib/models/records.integration.test.ts
```diff
@@ -69,7 +70,7 @@ describe('Records service', () => {
         ];
         const inserted = await upsertRecords({ records, connectionId, environmentId, model, syncId, syncJobId: 1 });
         expect(inserted).toStrictEqual({
-            addedKeys: ['1', '2', '3', '4'],
+            addedKeys: expect.arrayContaining(['1', '2', '3', '4']),
```
TBonnin, line 73: see comment below about ordering of upsert/update

packages/records/lib/models/records.ts
```diff
@@ -311,7 +302,7 @@ export async function upsert({
                             .returning(['id', 'external_id', 'data_hash', 'deleted_at', 'updated_at'])
                             .onConflict(['connection_id', 'external_id', 'model'])
                             .merge();
-                        if (merging.strategy === 'ignore_if_modified_after_cursor') {
+                        if (merging.strategy === 'ignore_if_modified_after_cursor' && 'cursor' in merging) {
```
TBonnin, line 306: see above. When 1st sync or empty cache, there is not cursor yet

packages/records/lib/models/records.ts
```diff
@@ -340,7 +331,11 @@ export async function upsert({
                             END as status`)
                     )
                     .from('upsert')
-                    .leftJoin('existing', 'upsert.external_id', 'existing.external_id');
+                    .leftJoin('existing', 'upsert.external_id', 'existing.external_id')
+                    .orderBy([
+                        { column: 'updated_at', order: 'asc' },
+                        { column: 'id', order: 'asc' }
+                    ]);
```
TBonnin, line 339: insertion order is different from cursor order since cursor is based on id which is a generated uuid. so we need to order based on cursor to get the latest cursor below

packages/records/lib/types.ts
```diff
@@ -65,7 +67,3 @@ export interface RecordCount {
     count: number;
     updated_at: string;
 }
-
-export type MergingStrategy = { strategy: 'override' } | { strategy: 'ignore_if_modified_after_cursor'; cursor: string };
-
-export type CursorOffset = 'first' | 'last';
```
TBonnin, line 71: moved to `types` package

packages/runner-sdk/lib/action.ts
```diff
@@ -106,15 +106,6 @@ export abstract class NangoActionBase {
...
 
-    protected stringify(): string {
-        return JSON.stringify(this, (key, value) => {
-            if (key === 'secretKey') {
-                return '********';
-            }
-            return value;
-        });
-    }
-
```
TBonnin, line 117: not necessary anymore. Was used by the persistApi related code in `NangoSync` runner-sdk class. 
PersistApi related code now lives in its own file and the secret is not logged anymore

packages/runner/lib/sdk/persist.ts
```diff
@@ -0,0 +1,312 @@
...
+import type { AxiosInstance, AxiosRequestConfig, AxiosResponse } from 'axios';
+import axios, { AxiosError, isAxiosError } from 'axios';
+import { getPersistAPIUrl } from '@nangohq/shared';
+import { getUserAgent } from '@nangohq/node';
+import { httpRetryStrategy, retryWithBackoff, Ok, Err } from '@nangohq/utils';
+import { logger } from '../logger.js';
+import type { Result } from '@nangohq/utils';
+import type { CursorOffset, DeleteRecordsSuccess, GetCursorSuccess, MergingStrategy, PostRecordsSuccess, PutRecordsSuccess } from '@nangohq/types';
+
+export interface PersistApi {
```
TBonnin: persist related code in runner-sdk `NangoSync` was starting to get out of control. Moving all the persist related code to its own file

packages/types/lib/persist/api.ts
```diff
@@ -0,0 +1,17 @@
...
+    nextMerging: MergingStrategy;
+}
+
+export interface PutRecordsSuccess {
+    nextMerging: MergingStrategy;
+}
+
+export interface DeleteRecordsSuccess {
+    nextMerging: MergingStrategy;
+}
```
TBonnin, line 13: I could have a single one. wdyt?

packages/cli/lib/services/parser.service.ts
```diff
@@ -70,15 +71,20 @@ class ParserService {
...
             setFieldMapping: 'setMetadata'
         };
 
-        const callsReferencingModelsToCheck = ['batchSave', 'batchDelete'];
+        const callsProxy = ['proxy', 'get', 'post', 'put', 'patch', 'delete'];
+        const callsBatchingRecords = ['batchSave', 'batchDelete', 'batchUpdate'];
+        const callsReferencingModelsToCheck = callsBatchingRecords.concat('setMergingStrategy');
+        const proxyLines: number[] = [];
+        const batchingRecordsLines: number[] = [];
+        const setMergingStrategyLines: number[] = [];
```
TBonnin, line 87: see below. addressing footguns by raising compilation errors and enforce a couple of rules:
- setMergingStrategy must be called only once
- setMergingStrategy must be called before any batch operation
- setMergingStrategy must be called before any proxy/get/post/...

packages/types/lib/persist/api.ts
bodinsamuel: I'm good like that, stuff can evolve differently

packages/cli/lib/services/parser.service.ts
bodinsamuel: I don't even think it's possible anyway

packages/cli/lib/services/parser.service.ts
bodinsamuel: very nice

packages/runner/lib/sdk/sdk.ts
```diff
@@ -263,6 +215,58 @@ export class NangoSyncRunner extends NangoSyncBase {
...
+        }
+    }
+
+    getMergingStategy(model: string): MergingStrategy {
+        return this.mergingByModel.get(model) || { strategy: 'override' };
+    }
+
+    setMergingStategyByModel(model: string, merging: MergingStrategy): void {
+        this.mergingByModel.set(model, merging);
+    }
```
bodinsamuel: Can you make them private for clarity?

packages/runner/lib/sdk/sdk.ts
```diff
@@ -263,6 +215,58 @@ export class NangoSyncRunner extends NangoSyncBase {
     sendLogToPersist = NangoActionRunner['prototype']['sendLogToPersist'];
     logAPICall = NangoActionRunner['prototype']['logAPICall'];
 
+    public async setMergingStrategy(merging: 'ignore_if_modified_after' | 'override', ...models: string[]): Promise<void> {
```
bodinsamuel: sugg: I would use a json params, it's not clear to understand after the fact especially with cryptic record name
Screenshot 2025-01-31 at 11 06 35

packages/runner/lib/sdk/sdk.ts
```diff
@@ -263,6 +215,58 @@ export class NangoSyncRunner extends NangoSyncBase {
...
+                    });
+                    if (res.isErr()) {
+                        throw res.error;
+                    }
+                    this.mergingByModel.set(model, { strategy: 'ignore_if_modified_after_cursor', ...(res.value ? { cursor: res.value.cursor } : {}) });
+                    break;
+                }
+                case 'override':
+                    this.mergingByModel.set(model, { strategy: 'override' });
+                    break;
```
bodinsamuel: `default: throw new Error('unsupported');` just in case

packages/runner/lib/sdk/persist.ts
```diff
@@ -0,0 +1,312 @@
...
+        nangoConnectionId,
+        model,
+        offset
+    }: {
+        environmentId: number;
+        nangoConnectionId: number;
+        model: string;
+        offset: CursorOffset;
+    }): Promise<Result<GetCursorSuccess>>;
+}
```
bodinsamuel: cusiosity: Why make an interface when you could just have the class used as a type?

bodinsamuel (approved):
Tested locally 👌🏻 
Glad you could refactor some stuff

nalanj (approved):
🚀

packages/runner/lib/sdk/sdk.ts
TBonnin: totally

packages/runner/lib/sdk/persist.ts
TBonnin: to have a mock implementation in the test. If you have a better solution I would be happy to go with it

packages/runner/lib/sdk/sdk.ts
TBonnin: you mean?
```
public async setMergingStrategy({strategy, models}: {strategy: 'ignore_if_modified_after' | 'override', models: string[]})
```
I initially choose to stay close to the interface of `batchSave` with a first argument being the value to save and the second one the model, in the case of setMergingStrategy being able to pass more than one. Using a object as a param would mean models must be an array which is slightly less clean since in mos cases it would be a single model. 
```
await nango.setMergingStrategy({strategy: 'ignore_if_modified_after', models: [IdTooLarge])
``` 
wdyt?

bodinsamuel: Ah that's a good point, if we are forward-thinking then reversing the model name could help typing and autocompletion with the change I have done in the hackathon:
- `await nango.setMergingStrategy(modelName: TModel, strategy: union); ` 

 - We could allow the first param to be an array too. 
 `await nango.setMergingStrategy(modelName: TModel | TModel[], strategy: union); ` 

--- 

If we go JSON you could allow models to be a string or array. 

Side note: the code is not doing parallelism so it's not different than having N calls. We could not care about multi-model syncs (which is already an exception rather than the norm), and only allow a string. So if you have multiple models you "just" call this multiple times.

bodinsamuel: So I'm not sure tbh, I don't have strong opinion, I just feels like it can be confusing to have free-form strings like this

packages/runner/lib/sdk/persist.ts
bodinsamuel: I believe you can just use the class as a regular interface

packages/runner/lib/sdk/sdk.ts
nalanj: > We could allow the first param to be an array too.
> `await nango.setMergingStrategy(modelName: TModel | TModel[], strategy: union); `

This is where my brain immediately went as well.

TBonnin: > We could allow the first param to be an array too.
> await nango.setMergingStrategy(modelName: TModel | TModel[], strategy: union); 

if I was starting from scratch I would definitely go for that but it would be very different from the batch operation signature. What about a middle ground with first param is a object and 2nd param is model name (abandoning the support for multiple models like you said @bodinsamuel)
```
await nango.setMergingStrategy({ strategy: 'ignore_if_modified_after'}, 'MyModel'); 
```

PR #3421, opened 2025-01-31
docs: public key deprecation guide

bastienbeurier:


rguldener:
@bastienbeurier I think it could be useful to mention that user & company details can now be passed as part of the connection flow and stored in Nango.

This should replace most prior needs for being able to specify the connection id, and is IMO a nice addition

bodinsamuel (approved):
looks good

bastienbeurier:
> @bastienbeurier I think it could be useful to mention that user & company details can now be passed as part of the connection flow and stored in Nango.
> 
> This should replace most prior needs for being able to specify the connection id, and is IMO a nice addition

I had this: 

> You must specify a **user ID** (and optionally an **organization ID**) when fetching the Connect session token to correctly attribute webhooks. You can also attach additional information to connections for better UI display.

But maybe you have something more specific in mind?

PR #3423, opened 2025-01-31
feat(zod-validation): add built in zod validation helper

khaliqgant:
I got tired of writing the same zod validation code logic all the time.

Turns this:
```
  const parsedInput = createUserSchema.safeParse(input);

  if (!parsedInput.success) {
    for (const error of parsedInput.error.errors) {
      await nango.log(
        `Invalid input provided to create a user: ${error.message} at path ${error.path.join(".")}`,
        { level: "error" },
      );
    }
    throw new nango.ActionError({
      message: "Invalid input provided to create a user",
    });
  }
```

into this:
```
await nango.zodValidateInput({ zodSchema: createUserSchema, input });
```


bodinsamuel (commented):
On the paper why not, but maybe it's a good time to actually finish the data/input validation feature already in place instead @bastienbeurier

bastienbeurier:
> On the paper why not, but maybe it's a good time to actually finish the data/input validation feature already in place instead @bastienbeurier

Let's discuss today

khaliqgant:
> On the paper why not, but maybe it's a good time to actually finish the data/input validation feature already in place instead @bastienbeurier

Brought this back as given the current crunch it doesn't seem like we'll get to adding this via validation any time soon.

bodinsamuel (approved):
Just asking in case, isn't this almost equivalent to create an helper file that you import in the templates? I guess it's less convenient for customers' template but at least it wouldn't be part of the api

khaliqgant:
> Just asking in case, isn't this almost equivalent to create an helper file that you import in the templates? I guess it's less convenient for customers' template but at least it wouldn't be part of the api

Yeah could work but quite redundant to create for every single customer repo we have

PR #3425, opened 2025-01-31
chore: cleanup legacy runner management

TBonnin:
Now fleet is fully released we can cleanup all the legacy logic used to manage runners.


.github/workflows/deploy.yaml
```diff
@@ -86,8 +86,6 @@ jobs:
                   RUNNER_OWNER_ID: ${{ secrets.RENDER_RUNNER_OWNER_ID }}
                   INTERNAL_API_KEY: ${{ inputs.stage == 'production' && secrets.PROD_INTERNAL_API_KEY || secrets.STAGING_INTERNAL_API_KEY }}
               run: |
-                  bash ./scripts/deploy/runners.bash
```
TBonnin, line 89: 🪓 🪓 🪓

packages/runner/lib/env.ts
```diff
@@ -1,10 +1,9 @@
 import { ENVS, parseEnvs } from '@nangohq/utils';
 import { getJobsUrl, getPersistAPIUrl } from '@nangohq/shared';
 
-export const envs = parseEnvs(ENVS);
+export const envs = parseEnvs(ENVS.required({ RUNNER_NODE_ID: true }));
 
 export const heartbeatIntervalMs = 30_000;
-export const runnerId = envs.RUNNER_ID || `${envs.RUNNER_NODE_ID}`;
 
-export const jobsServiceUrl = process.env['NOTIFY_IDLE_ENDPOINT']?.replace(/\/idle$/, '') || getJobsUrl(); // TODO: remove legacy NOTIFY_IDLE_ENDPOINT once all runners are updated with JOBS_SERVICE_URL env var
```
TBonnin, line 9: this should have been cleaned up a looong time ago

nalanj (approved):

nalanj:
Just ran everything locally and seemed like it was all running smoothly.

packages/runner/lib/app.ts
```diff
@@ -4,12 +4,13 @@ import { stringifyError } from '@nangohq/utils';
 import { logger } from './utils.js';
 import { monitorProviders } from '@nangohq/shared';
 import { register } from './register.js';
+import { envs } from './env.js';
 
 const providersMonitorCleanup = await monitorProviders();
 
 try {
     const port = parseInt(process.argv[2] || '') || 3006;
-    const id = process.argv[3] || process.env['RUNNER_ID'] || 'unknown-id';
```
bodinsamuel, line 12: I think we are missing this value for onprem / in the helm chart
see last thread from the onprem customers

bodinsamuel (approved):
Nice ♻️

packages/runner/lib/app.ts
TBonnin: good point. RUNNER_ID is irrelevant when using a remote runner which is why this value is not set. maybe we could come up with a better id than `unknown-id` but I don't think it is a big deal right now

bodinsamuel: I'm just worried it's going to be undefined, instead of a string, which could break things I guess

PR #3427, opened 2025-01-31
feat(server): Separate auth related webhooks into their own log context

nalanj:
Isolates webhook logs for auth token refresh failure and connection creation and gives those webhook logs new types

See https://linear.app/nango/issue/NAN-1547/log-nango-webhooks

## How I tested it:

- Make sure webhooks are configured and enabled locally
- Break an integration and trigger a token refresh to see a failed refresh webhook
- Create a new connection to see a new connection


TBonnin (approved):
a lot of data passthrough just to create a log context 🙊

nalanj:
@TBonnin This made me very sad as I did it.

bodinsamuel (approved):
👌🏻

PR #3428, opened 2025-01-31
fix(api): better check for json content-type 

bodinsamuel:
## Changes

- better check for json content-type 
Header can have the charset specified but we only checked exact match, express has a function to handle this check so it simplifies this condition `content-type: application/json; charset=utf-8`


TBonnin (approved):
didn't know about 'is'

bodinsamuel:
me neither, very poorly named function https://badgerbadgerbadgerbadger.medium.com/new-tricks-req-is-type-in-expressjs-9f1504156d52

PR #3429, opened 2025-02-01
fix: fleet to fetch all nodes at once

TBonnin:
Fleet supervisor must know about all active nodes to be able to plan correctly. Before this fix, supervisor was planning one page at a time which could lead to starting operations based on missing information (ex: page includes the OUTDATED node for a given routingId but the PENDING entry is not in the same page, leading the supervisor to believe it must create a PENDING node)

### DIFF
--- a/packages/fleet/lib/models/nodes.ts
+++ b/packages/fleet/lib/models/nodes.ts
@@ -137,50 +137,63 @@ export async function search(
     params: {
         states: [NodeState, ...NodeState[]]; // non-empty array
         routingId?: RoutingId;
-        cursor?: number;
         limit?: number;
     }
-): Promise<
-    Result<{
-        nodes: Map<RoutingId, Record<NodeState, Node[]>>;
-        nextCursor?: number;
-    }>
-> {
-    try {
-        const limit = params.limit || 1000;
-        const query = db
-            .select<DBNode[]>('*')
-            .from(NODES_TABLE)
-            .whereIn('state', params.states)
-            .orderBy('id')
-            .limit(limit + 1); // fetch one more than limit to determine if there are more results
+): Promise<Result<Map<RoutingId, Record<NodeState, Node[]>>>> {
+    const recursiveSearch = async ({
+        cursor,
+        nodesMap,
+        remaining
+    }: {
+        cursor?: number;
+        nodesMap: Map<RoutingId, Record<NodeState, Node[]>>;
+        remaining: number;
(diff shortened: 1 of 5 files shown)


packages/fleet/lib/models/nodes.ts
```diff
@@ -137,50 +137,63 @@ export async function search(
     params: {
         states: [NodeState, ...NodeState[]]; // non-empty array
         routingId?: RoutingId;
-        cursor?: number;
```
TBonnin, line 140: since supervisor must have a full picture of all nodes, let's have search paginate internally and return all nodes

packages/fleet/lib/supervisor/supervisor.integration.test.ts
```diff
@@ -224,12 +224,11 @@ describe('Supervisor', () => {
     });
 
     it('should remove old TERMINATED nodes', async () => {
-        const sevenDaysAgo = new Date(Date.now() - STATE_TIMEOUT_MS.TERMINATED - 1);
         const terminatedNode = await createNodeWithAttributes(dbClient.db, { state: 'TERMINATED', deploymentId: activeDeployment.id });
         const oldTerminatedNode = await createNodeWithAttributes(dbClient.db, {
             state: 'TERMINATED',
             deploymentId: activeDeployment.id,
-            lastStateTransitionAt: sevenDaysAgo
+            lastStateTransitionAt: new Date(Date.now() - STATE_TIMEOUT_MS.TERMINATED - 1)
```
TBonnin, line 231: unrelated but `sevenDaysAgo` name was not up to date. Timeout is actually 24h

bodinsamuel (approved):

packages/fleet/lib/models/nodes.ts
```diff
@@ -137,50 +137,63 @@ export async function search(
...
+        });
+    };
+
+    try {
+        const res = await recursiveSearch({
+            nodesMap: new Map<RoutingId, Record<NodeState, Node[]>>(),
+            remaining: params.limit || Infinity
         });
+        return Ok(res);
     } catch (err) {
```
bodinsamuel, line 197: I guess there can be a race condition if a node unregister/register during the pagination. Can a transaction help?

TBonnin: yes but that's no different from fleet supervisor planning based on a state of the nodes that is not fully up-to-date. (ie: registration happens while it is planning). Fleet control loop is designed around it and will get the correct state and act accordingly in the next tick

PR #3432, opened 2025-02-03
feat(ui): environment settings revamp

bodinsamuel:
## Changes

Fix https://linear.app/nango/issue/NAN-2468/revamp-environment-settings
Figma https://www.figma.com/design/tqlwMbFuBCcMxktQqyl3t9/Environment-Settings?node-id=194-63531&m=dev

>  Please note the design and UX is slightly different than the figma following a discussion from Bastien. I had to improvise some stuff. (NB: I'm not happy with the secret key rotation UX)

- UI: New environment settings page 

- New endpoint `PATCH /api/v1/environments`
Handle all env update

- New endpoint `POST /api/v1/environments/variables`
Handle all updates to env variables

- New endpoint `POST /api/v1/environments/webhook`
Handle all updates to webhook

- **Delete** old and deprecated endpoint
`POST /api/v1/environment/callback`
`POST /api/v1/environment/hmac-enabled`
`POST /api/v1/environment/slack-notifications-enabled`
`POST /api/v1/environment/hmac-key`
`POST /api/v1/environment/environment-variables`
`POST /api/v1/environment/webhook/primary-url`
`POST /api/v1/environment/webhook/secondary-url`
`POST /api/v1/environment/webhook/settings`
`POST /api/v1/environment/otlp/settings`

- Fix DBExternalWebhook type
- Delete old code in services
- UI: Add Switch and Accordion component from shadcn

## 🧪 Tests
- Launch your dashboard 
- Edit all the settings
Please report any bug or UX issue.

https://github.com/user-attachments/assets/1282989e-e39e-4860-9016-b90bbef185cc


bodinsamuel:
Somehow this happened 
Screenshot 2025-02-03 at 19 18 28

packages/server/lib/controllers/v1/environment/patchEnvironment.ts
```diff
@@ -0,0 +1,74 @@
...
+
+const validationBody = z
+    .object({
+        callback_url: z.string().url().optional(),
+        hmac_key: z.string().min(0).max(1000).optional(),
+        hmac_enabled: z.boolean().optional(),
+        slack_notifications: z.boolean().optional(),
+        otlp_endpoint: z.string().url().or(z.literal('')).optional(),
+        otlp_headers: z
+            .array(z.object({ name: z.string().min(1).max(256), value: z.string().min(1).max(4000) }))
```
TBonnin, line 16: > 4000
generous 😉

packages/server/lib/controllers/v1/environment/variables/postVariables.ts
```diff
@@ -0,0 +1,37 @@
...
+        return;
+    }
+
+    const { environment } = res.locals;
+
+    const body: PostEnvironmentVariables['Body'] = val.data;
+
+    const updated = await environmentService.editEnvironmentVariable(environment.id, body.variables);
+    if (!updated && body.variables.length > 0) {
+        res.status(500).send({ error: { code: 'server_error', message: 'Failed to update environment variable' } });
```
TBonnin:
```suggestion
        res.status(500).send({ error: { code: 'server_error', message: 'Failed to update environment variables' } });
```

packages/webapp/src/pages/Environment/Settings/Backend.tsx
```diff
@@ -0,0 +1,171 @@
...
+                        Secret Key
+                    </label>
+                    {hasNewSecretKey && (
+                        <label htmlFor="secretKey" className={'text-s'}>
+                            Current secret
+                        </label>
+                    )}
+                    <div className="flex gap-2">
+                        <SecretInput
+                            view={false}
```
TBonnin, line 98: we can only copy the secretKey now. We cannot see it in the UI anymore?

TBonnin (approved):
Tested it locally, it looks much better. 
The only thing I find a bit weird is the new secret key activation is done by deleting one of the key instead of a 'cancel/rotate' which is much more explicit

packages/webapp/src/pages/Environment/Settings/Backend.tsx
bodinsamuel: yeah that's the part we kinda improvised, I'm not sure if it's good or bad.

packages/server/lib/controllers/v1/environment/patchEnvironment.ts
bodinsamuel: ahah yes, I wasn't sure what people were putting there but we still need a limit :D

bodinsamuel:
> The only thing I find a bit weird is the new secret key activation is done by deleting one of the key instead of a 'cancel/rotate' which is much more explicit

Yes agreed, that's something we quickly changed with @bastienbeurier maybe we can rediscuss

bastienbeurier:
> > The only thing I find a bit weird is the new secret key activation is done by deleting one of the key instead of a 'cancel/rotate' which is much more explicit
> 
> Yes agreed, that's something we quickly changed with @bastienbeurier maybe we can rediscuss

I also agree. Maybe we can show these two buttons instead: 
1. Confirm key rotation
2. Cancel key rotation

Clicking button 1 shows a dialog saying: 
Confirm key rotation? The old key will be deactivated. Only do this when the new key is used in production. This is not reversible.

Clicking button 2 shows a dialog saying: 
Cancel key rotation? The new key will be deactivated. Only do this when the old key is used in production. This is not reversible.

But IMO we should also show this callout that we're not showing at the moment:
CleanShot 2025-02-04 at 11 03 40@2x

bodinsamuel:
> But IMO we should also show this callout that we're not showing at the moment:

I'm not showing this because it's not true (https://linear.app/nango/issue/NAN-2645/secret-key-rotation-make-both-key-available)

bastienbeurier:
> > But IMO we should also show this callout that we're not showing at the moment:
> 
> I'm not showing this because it's not true (https://linear.app/nango/issue/NAN-2645/secret-key-rotation-make-both-key-available)

I see. And I guess it's painful to bake that in so that we have the final UI directly?

bodinsamuel:
> I see. And I guess it's painful to bake that in so that we have the final UI directly?

Yes, we need to change the auth layer which has bigger implication

bodinsamuel:
Not sure that's exactly what you had in mind, but looks like this now
Screenshot 2025-02-04 at 11 16 52

bastienbeurier:
> Not sure that's exactly what you had in mind, but looks like this now Screenshot 2025-02-04 at 11 16 52

Let's add a callout:
> The current secret is still active, the new secret is not. Confirm key rotation to activate new secret and deactivate current secret.

bodinsamuel:
Reviewed everything with bastien, we'll probably have feedback from customers anyway :D

PR #3433, opened 2025-02-03
fix(server): Log instance id from render instead of random id and catch unhandled

nalanj:
Should help with tracking down the crashes we've seen. I also went ahead and dumped RENDER_INSTANCE_ID to our logs instead of the random nanoid id we were using, so that it's easier to correlate metrics graphs to logs.

Also limits depth for dumping errors to logs.


packages/server/lib/server.ts
```diff
@@ -22,9 +22,18 @@ import { refreshConnectionsCron } from './refreshConnections.js';
...
+});
+
+process.on('uncaughtException', (e) => {
+    logger.error('Received uncaughtException...', e);
+    // not closing on purpose
+});
+
 const app = express();
 app.disable('x-powered-by');
-app.set('trust proxy', 1);
```
bodinsamuel, line 27: Are you sure about that, it's usually mandatory as you as you have a load balancer

bodinsamuel (approved):

packages/server/lib/server.ts
nalanj: Gah, there's something crazy that keeps happening with eslint where it prunes lines randomly when I save.

bodinsamuel (approved):

TBonnin (approved):

PR #3436, opened 2025-02-03
docs: show connect UI gif

bastienbeurier:


docs-v2/images/connect-ui.gif
bodinsamuel, line 1: nitpick: it would be cleaner with a less+white background
you could open the box and modify the semi-transparent bg with `background-color: white` (the drop shadow is a bit harsh though)

bodinsamuel (approved):

docs-v2/images/connect-ui.gif
bastienbeurier: @bodinsamuel I think I dont understand! Which white background? In the docs it looks like: 
CleanShot 2025-02-03 at 21 47 52@2x

bastienbeurier: Just tried with a white background but it doesn't look great: 

CleanShot 2025-02-04 at 09 24 05@2x

bodinsamuel: ah I guess it depends if you have dark mode or not. 
My main point is that the background is noisy and people could still think it's only available through the dashboard. 
anyway not a big deal

PR #3438, opened 2025-02-03
chore: fleet improvements

TBonnin:
- adding 'image' column to deployment
    - it will make it easy to switch to another image without having to change fleet code (for example if we move runner to the unified image).
    - Note: the column is not used by fleet supervisor yet. Coming in next PR :)
- checking if the image exist when rolling out to avoid rolling out while master is building
- add 'notes' column to node_overrides
    - won't be used by the app but I wanted a text column where we can manually enter useful information like the name of the customer or 'Thomas's runner'
- making image, cpu_mill, memory_mb, storage_mb columns nullable in the node_overrides table. I will change the logic: if value is null, fleet will pick default values instead of having to insert specific values when overriding, especially useful for the image that changes with each deployment

### DIFF
--- a/packages/fleet/lib/models/deployments.ts
+++ b/packages/fleet/lib/models/deployments.ts
@@ -73,7 +73,6 @@ export async function create(db: knex.Knex, image: string): Promise<Result<Deplo
             return Ok(DBDeployment.to(inserted));
         });
     } catch (err) {
-        console.log(err);
         return Err(new FleetError(`deployment_creation_failed`, { cause: err, context: { image } }));
     }
 }
--- a/packages/fleet/lib/models/deployments.ts
+++ b/packages/fleet/lib/models/deployments.ts
@@ -9,6 +9,7 @@ export const DEPLOYMENTS_TABLE = 'deployments';
 interface DBDeployment {
     readonly id: number;
     readonly commit_id: CommitHash;
+    readonly image: string;
     readonly created_at: Date;
     readonly superseded_at: Date | null;
 }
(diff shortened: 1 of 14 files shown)


bodinsamuel (approved):
🙌🏻

packages/fleet/lib/models/deployments.ts
```diff
@@ -53,19 +56,25 @@ export async function create(db: knex.Knex, commitId: CommitHash): Promise<Resul
...
             const [inserted] = await trx.into<DBDeployment>(DEPLOYMENTS_TABLE).insert(dbDeployment).returning('*');
             if (!inserted) {
-                return Err(new Error(`Error: no deployment '${commitId}' created`));
+                return Err(new Error(`Error: no deployment for '${image}' created`));
             }
             return Ok(DBDeployment.to(inserted));
         });
     } catch (err) {
-        return Err(new FleetError(`deployment_creation_failed`, { cause: err, context: { commitId } }));
+        console.log(err);
```
nalanj: Not sure if this was intended or a leftover

packages/fleet/lib/supervisor/supervisor.integration.test.ts
```diff
@@ -34,8 +34,8 @@ describe('Supervisor', () => {
 
     beforeEach(async () => {
         await dbClient.migrate();
-        previousDeployment = (await deployments.create(dbClient.db, generateCommitHash().unwrap())).unwrap();
-        activeDeployment = (await deployments.create(dbClient.db, generateCommitHash().unwrap())).unwrap();
+        previousDeployment = (await deployments.create(dbClient.db, generateImage().unwrap())).unwrap();
+        activeDeployment = (await deployments.create(dbClient.db, generateImage().unwrap())).unwrap();
```
nalanj, line 38: `.unwrap().unwrap().unwrap().unwrap()` 😆

nalanj (approved):

packages/fleet/lib/models/deployments.ts
TBonnin: 🤦

PR #3441, opened 2025-02-04
fix(server): Override agent toJSON

nalanj:
Overnight I had the realization that I could just override toJSON and not have to do anything more complicated to handle dealing with agent serialization. I'm hoping it'll clear up those really hot profiles we're seeing in prod.


bodinsamuel (approved):
Tested locally👌🏻

TBonnin:
👌

PR #3445, opened 2025-02-04
feat(integrations): whitelist unzipper package

hassan254-prog:
We are  currently working on an integration where the provider exclusively delivers user data as a ZIP archive. Specifically, when requesting a user list, the API responds with a link to a compressed `.zip` file containing x number of `.txt` files that need to be extracted and processed.  The provider uses the **DEFLATE** compression method, which our current implementation does not natively support for in-memory decompression. To handle this efficiently, we need to introduce a package like`unzipper`, which supports **decompression and streaming extraction**. This allows us to process files on the fly without loading the entire archive file into memory, significantly improving performance and reducing memory overhead.


TBonnin (commented):
as usual I am not a huge fan of adding npm packages to our runtime. It basically means we are now responsible for upgrading/managing it. @khaliqgant your call.

khaliqgant:
> as usual I am not a huge fan of adding npm packages to our runtime. It basically means we are now responsible for upgrading/managing it. @khaliqgant your call.

Thanks. Yes, unfortunately think we need this.

khaliqgant (approved):

PR #3447, opened 2025-02-04
fix(shared): Keep agents from cloning in simple-oauth2

nalanj:
Relating to request blocking we were seeing in profiles.

## How I tested it
- made sure oauth2 interactions were still working
- I manually updated `node_modules/simple-oauth2/lib/client/client.js` to print the copied config and make sure it included the resulting getter.


bodinsamuel (approved):
let's try

packages/shared/lib/clients/oauth2.client.ts
```diff
@@ -36,25 +36,22 @@ export function getSimpleOAuth2ClientConfig(
...
             // @ts-expect-error seems unused ?
             scopeSeparator: provider.scope_separator || ' '
         }
     };
+
+    // define getters for agents so that they aren't clone within simple-oauth2
+    const httpConfig = { headers, agents: {} };
+    Object.defineProperty(httpConfig.agents, 'http', { get: () => httpAgent, enumerable: true });
+    Object.defineProperty(httpConfig.agents, 'https', { get: () => httpsAgent, enumerable: true });
+    Object.defineProperty(httpConfig.agents, 'httpsAllowUnauthorized', { get: () => httpsAgent, enumerable: true });
```
TBonnin, line 51: I think we can simplify and avoid having to use `defineProperty` like this:
```
const getAgentConfig = (): { agents: AgentConfig } => ({
  agents: {
    get http() { return httpAgent; },
    get https() { return httpsAgent; },
    get httpsAllowUnauthorized() { return httpsAgent; }
  }
});

const clientConfig: Merge<ModuleOptions, { http: WreckHttpOptions }> = {
    ...,
    http: {
      headers,
      ...getAgentConfig()
    }
  };
```
And adding a comment to mention the getter dynamic lookup trick used here. I am sure next week I am gonna have forgotten about it already

TBonnin (approved):

packages/shared/lib/clients/oauth2.client.ts
nalanj: I had already hit merge but I'll put a follow-up in real quick and merge it if this works out for that hot spot in the profiles, otherwise we'll just revert.

nalanj: https://github.com/NangoHQ/nango/pull/3457

PR #3448, opened 2025-02-04
fix(api): GET /records new format

bodinsamuel:
## Changes

Contributes to https://linear.app/nango/issue/NAN-1985/get-record-by-id

- `GET /records` new format
Added some well deserved integration tests (not covering all params)


packages/server/lib/controllers/sync.controller.ts
```diff
@@ -24,71 +24,21 @@ import {
...
 import { records as recordsService } from '@nangohq/records';
 import type { RequestLocals } from '../utils/express.js';
 import { getOrchestrator } from '../utils/utils.js';
 import { getInterval } from '@nangohq/nango-yaml';
+import { getPublicRecords } from './records/getRecords.js';
 
 const orchestrator = getOrchestrator();
 
 class SyncController {
-    public async getAllRecords(req: Request, res: Response<any, Required<RequestLocals>>, next: NextFunction) {
```
nalanj, line 43: ❤️

nalanj (commented):
Looks great. Was able to curl it locally and everything seems to be working well!

This is another spot where we're deserializing and reserializing, but also I'm not sure how to get around that with encryption.

nalanj (approved):

packages/server/lib/controllers/records/getRecords.integration.test.ts
```diff
@@ -0,0 +1,183 @@
...
+        });
+        isSuccess(res.json);
+        expect(res.json).toStrictEqual({
+            next_cursor: null,
+            records: []
+        });
+        expect(res.res.status).toBe(200);
+    });
+
+    it('should get multiple page of records', async () => {
```
TBonnin, line 98: i feel like here we are testing the `Records.getRecords` function that is already tested https://github.com/NangoHQ/nango/blob/cce5c2736d23830d3f0fdaf6f4f2c9cc32bf7582/packages/records/lib/models/records.integration.test.ts

TBonnin (approved):

packages/server/lib/controllers/records/getRecords.integration.test.ts
bodinsamuel: I think we already had the debate, imho it's still relevant if the controller is not passing the query params to the method properly then there is nothing to catch that

bodinsamuel:
> This is another spot where we're deserializing and reserializing, but also I'm not sure how to get around that with encryption.

Yes this one might be tricky

PR #3450, opened 2025-02-04
chore(fleet): simplify node config overrides

TBonnin:
NodeConfigOverride values are now either null (ie: don't override) or a given value. It simplifies adding new override since we don't have to worry about what are the default values, just enter null for the attributes that don't need to be overridden. Also no special default value for image (nangohq/nango-runner without commitHash)

I still need to remove the commit_id column in deployment which won't be used anymore once this commit is deployed. Will do it in next PR.

# How to test:
Add/modify node_config_overrides rows in your dev db and check that fleet supervisor correctly rolls out new node

### DIFF
--- a/packages/fleet/lib/models/deployments.ts
+++ b/packages/fleet/lib/models/deployments.ts
@@ -18,7 +18,6 @@ const DBDeployment = {
     to(dbDeployment: DBDeployment): Deployment {
         return {
             id: dbDeployment.id,
-            commitId: dbDeployment.commit_id,
             image: dbDeployment.image,
             createdAt: dbDeployment.created_at,
             supersededAt: dbDeployment.superseded_at
@@ -28,7 +27,7 @@ const DBDeployment = {
         return {
             id: deployment.id,
             image: deployment.image,
-            commit_id: deployment.commitId,
+            commit_id: '0000000000000000000000000000000000000000' as CommitHash, //TODO: remove
             created_at: deployment.createdAt,
             superseded_at: deployment.supersededAt
         };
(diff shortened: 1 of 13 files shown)


packages/fleet/lib/models/deployments.ts
```diff
@@ -28,7 +27,7 @@ const DBDeployment = {
         return {
             id: deployment.id,
             image: deployment.image,
-            commit_id: deployment.commitId,
+            commit_id: '0000000000000000000000000000000000000000' as CommitHash, //TODO: remove
```
TBonnin, line 30: keeping the `commit_id` column to avoid issue when deploying. Old jobs instances would still relies on its existence. Once this commit is deployed and no running code relies on this column I will add a migration and cleanup the code

packages/fleet/lib/supervisor/supervisor.ts
```diff
@@ -182,15 +182,21 @@ export class Supervisor {
...
-                    // so we need to check if the image override with commitId is the same as the node's image
-                    // or if the image override (without commitId) + current deployment commitId is the same as the node's image
-                    return !(
-                        (configOverride.image === node.image || `${configOverride.image}:${deployment.commitId}` === node.image) &&
-                        node.cpuMilli === configOverride.cpuMilli &&
-                        node.memoryMb === configOverride.memoryMb &&
-                        node.storageMb === configOverride.storageMb
-                    );
+                    // TODO: remove once values in node_config_overrides are updated to NULL in the db
+                    const imageOverride = configOverride.image?.includes(':') ? configOverride.image : undefined;
```
TBonnin, line 186: the TODO only applies to this line since at the moment the default value for image is `nangohq/nango-runner` without the commitHash

bodinsamuel (approved):

PR #3458, opened 2025-02-05
feat(persist): Enable filtering records by external id

nalanj:
We'll use this for a few new features where we can load records by id.

## How I tested it

- Ran the tests!
- Also dumped the generated query with the new filter to a string and ran it through EXPLAIN ANALYZE to make sure there were no sequential scans. Looks good.

### DIFF
--- a/packages/records/lib/models/records.integration.test.ts
+++ b/packages/records/lib/models/records.integration.test.ts
@@ -524,7 +524,7 @@ describe('Records service', () => {
             expect(allFetchedRecords.length).toBe(numOfRecords);
         });
 
-        it('Should retreive records by external_id', async () => {
+        it('Should retrieve records by external_id', async () => {
             const { connectionId, model } = await upsertNRecords(10);
 
             const response = await Records.getRecords({
--- a/packages/records/lib/models/records.integration.test.ts
+++ b/packages/records/lib/models/records.integration.test.ts
@@ -218,6 +218,41 @@ describe('Records service', () => {
                 });
             });
         });
+
+        it('Should return correct added records count when upserting concurrently', async () => {
+            const connectionId = rnd.number();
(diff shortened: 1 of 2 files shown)


packages/records/lib/models/records.integration.test.ts
```diff
@@ -218,6 +218,41 @@ describe('Records service', () => {
                 });
             });
         });
+
```
nalanj, line 221: Just moving this test to be inside of the upsert section

packages/records/lib/models/records.integration.test.ts
```diff
@@ -431,134 +466,120 @@ describe('Records service', () => {
...
-        const limit = 1000;
-        const { connectionId, model } = await upsertNRecords(numOfRecords);
+                expect(currentRecordDate.isAfter(previousRecordDate) || currentRecordDate.isSame(previousRecordDate)).toBe(true);
+            }
+            expect(allFetchedRecords.length).toBe(numOfRecords);
+        });
 
-        let cursor: string | undefined | null = null;
-        let allRecordsLength = 0;
+        it('Should retreive records by external_id', async () => {
```
nalanj: Sorry, git really went insane on this diff. The blue below is the actual new test, but it may be easier to look at it side by side or in an editor.

packages/records/lib/models/records.integration.test.ts
bodinsamuel: it's easier to review with "hide whitespace"

packages/records/lib/models/records.integration.test.ts
bodinsamuel:
```suggestion
        it('Should retrieve records by external_id', async () => {
```

bodinsamuel (approved):

PR #3460, opened 2025-02-05
fix(api): get records filter should accept uppercase and combined

bodinsamuel:
## Changes

- `GET /records` filter param should accept uppercase and combined
I missed that when checking the openapi spec, it's not documented there but it's in the node-client. e.g: `ADDED,updated` 

```
curl --request GET \
  --url 'http://localhost:3003/records?model=LastSyncDate&limit=100&filter=UPDATED%2Cadded' \
  --header 'Authorization: ' \
  --header 'Connection-Id: 025d9a38-f31c-4a64-b034-57a425761a90' \
  --header 'content-type: application/json'
```

### DIFF
--- a/packages/records/lib/types.ts
+++ b/packages/records/lib/types.ts
@@ -38,6 +38,7 @@ export type ReturnedRecord = {
 } & Record<string, RecordValue> & { id: string };
 
 export type LastAction = 'ADDED' | 'UPDATED' | 'DELETED' | 'added' | 'updated' | 'deleted';
+export type CombinedFilterAction = `${LastAction},${LastAction}`;
 
 interface RecordMetadata {
     first_seen_at: string;
--- a/packages/types/lib/record/api.ts
+++ b/packages/types/lib/record/api.ts
@@ -1,6 +1,7 @@
 import type { ApiError, Endpoint } from '../api';
 
 export type RecordLastAction = 'ADDED' | 'UPDATED' | 'DELETED' | 'added' | 'updated' | 'deleted';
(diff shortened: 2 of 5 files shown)


packages/types/lib/record/api.ts
```diff
@@ -1,6 +1,7 @@
 import type { ApiError, Endpoint } from '../api';
 
 export type RecordLastAction = 'ADDED' | 'UPDATED' | 'DELETED' | 'added' | 'updated' | 'deleted';
+export type CombinedFilterAction = `${RecordLastAction},${RecordLastAction}`;
```
bodinsamuel, line 4: Those are duplicated type, not sure why we are not using @types pkg here

packages/records/lib/types.ts
```diff
@@ -38,6 +38,7 @@ export type ReturnedRecord = {
 } & Record<string, RecordValue> & { id: string };
 
 export type LastAction = 'ADDED' | 'UPDATED' | 'DELETED' | 'added' | 'updated' | 'deleted';
+export type CombinedFilterAction = `${LastAction},${LastAction}`;
```
bodinsamuel, line 41: Those are duplicated type, not sure why we are not using `@types` pkg here

TBonnin: maybe @types didn't exist back then?

TBonnin (approved):

packages/records/lib/types.ts
bodinsamuel: ah maybe, I will make a following PR to sync

PR #3461, opened 2025-02-05
fix(ui): correct field when updating webhooks secondary url

bodinsamuel:
## Changes

- Fix correct field when updating webhooks secondary url


TBonnin (approved):
🙈

PR #3463, opened 2025-02-05
feat(persist): Add endpoint to get records

nalanj:
Adds a new endpoint to persist for getting records. Includes tests.


bodinsamuel (approved):

packages/persist/lib/routes/environment/environmentId/connection/connectionId/getRecords.ts
```diff
@@ -0,0 +1,55 @@
...
+    const {
+        params: { nangoConnectionId },
+        query: { model, externalIds, cursor }
+    } = req;
+    const result = await records.getRecords({
+        connectionId: nangoConnectionId,
+        model,
+        cursor,
+        externalIds,
+        limit: 100
```
TBonnin, line 38: should the caller pass the limit (with a max value allowed)?

packages/persist/lib/server.integration.test.ts
```diff
@@ -356,6 +356,98 @@ describe('Persist API', () => {
...
+                    records: records,
+                    providerConfigKey: seed.connection.provider_config_key,
+                    connectionId: seed.connection.connection_id,
+                    activityLogId: seed.activityLogId
+                }),
+                headers: {
+                    Authorization: `Bearer ${mockSecretKey}`,
+                    'Content-Type': 'application/json'
+                }
+            });
```
TBonnin: you could abstract the insertion logic into a private function (calling the records package directly)

TBonnin (approved):

packages/persist/lib/routes/environment/environmentId/connection/connectionId/getRecords.ts
nalanj: @TBonnin I was thinking since this will be wrapped up behind a function we're exposing that we could always keep limit at 100 and then abstract it away in the sdk. What do you think?

packages/persist/lib/server.integration.test.ts
nalanj: Fixed in 245b84c

packages/persist/lib/routes/environment/environmentId/connection/connectionId/getRecords.ts
nalanj: Fixed in ef46920b8

PR #3466, opened 2025-02-06
feat(api): get records by IDs

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-1985/get-record-by-id

- Allow `GET /records` to be filtered by records' ids
I choose to have an array instead of a comma-separated string because the IDs can contain any string, making them susceptible to bad parsing, plus, it's easier to validate. However, it's a bit more annoying to add to a URL, specially that `UrlSearchParams` does comma-separated by default.

## 🧪 Tests

try filtering with your favorite tool 
```
curl --request GET \
  --url 'http://localhost:3003/records?model=LastSyncDate&ids=1&ids=2' \
  --header 'Authorization: REDACTED' \
  --header 'Connection-Id: REDACTED' \
  --header 'content-type: application/json'
```


nalanj (approved):
Tested locally, looks great!

Also your id handling is going to be inspiration for my new endpoint in persist! 🙏🏻

khaliqgant:
@nalanj with this change realized that we should also update the auto generated documentation for https://github.com/NangoHQ/integration-templates/blob/main/scripts/generate-readmes.ts#L116

Also @bodinsamuel the spec within the webapp should probably list this as a param as well 
image

PR #3468, opened 2025-02-06
feat: runner use new image

bodinsamuel:
## Changes

Contributes to https://linear.app/nango/issue/NAN-605/use-image-for-runner

- Runner: Use unified docker image
Tested in staging https://dashboard.render.com/pserv/srv-cui9s49u0jms739cd95g/settings
It's less transparent since we need to specify a docker command, happy to get suggestion here. I expect that after the migration, we can set it as the default

- Fix deprecated runtime configuration
see doc https://api-docs.render.com/reference/create-service

- Remove docker login leftovers


nalanj (approved):
Looks good. Definitely think Thomas should review too.

packages/jobs/lib/runner/render.ts
```diff
@@ -34,8 +35,11 @@ export const renderNodeProvider: NodeProvider = {
...
-                    env: 'image',
-                    plan: getPlan(node)
+                    runtime: 'image',
+                    plan: getPlan(node),
+                    ...(node.image.includes('-runner')
+                        ? {}
+                        : { envSpecificDetails: { dockerCommand: 'node packages/runner/dist/app.js 80 dockerized-runner' } })
                 },
                 envVars: [
                     { key: 'PORT', value: '80' },
```
TBonnin: you could have a `port` variable and use it here and in the dockerCommand so they are always in sync

TBonnin (approved):
should we also remove building the older image in build-images.yaml?

bodinsamuel:
> should we also remove building the older image in build-images.yaml?

Yes I'm just keeping it until the next successful deploy just in case I need to revert

packages/jobs/lib/runner/render.ts
bodinsamuel: good point

PR #3470, opened 2025-02-06
fix(types): Connection -> DBConnection

bodinsamuel:
## Changes

> [!NOTE]
> Lot of files have been modified but it's mostly due to setting `id` as not undefined
> 98% of this PR is type change only which can be scary but not as serious as it seems

- **Move connection types** to `types` package
- **Replaced** `Connection` -> `DBConnection`
Fixed all undefined with nullable or not undefined
- **Added** `DBConnectionDecrypted` to differentiate between raw connection and decrypted connection
This difference used to be marked by Connection vs StoredConnection which was not very clear, not used correctly everywhere and was not incompatible so it lead to bad argument being passed

- **Replaced** NangoConnection by ConnectionJobs and ConnectionInternal
`ConnectionJobs` is used by Jobs (yes) and `ConnectionInternal` is some kind of weird shape that is used somewhere. Ideally this last type should disappear  

## 🤓 How to review?

You will see most changes are super trivial but if you want to start with the right files:
- [types/connection/db.ts](https://github.com/NangoHQ/nango/pull/3470/files#diff-a819f17557dd02b45e15723a24aba70c7847c50d6c4fedb928de1e9b8c0ae04f)
- [connection.service.ts](https://github.com/NangoHQ/nango/pull/3470/files#diff-00f759b1e40c244e1f3258270374f4fb3e6ddf4e13ced14e12794ff30131efdd)
- then everything else

## 🧪 Tests

While it's only types change, it's also non-trivial. I would be very glad to have more local tests
- Launch the UI and tests all pages
- Trigger some syncs and actions
- Use the local CLI to dryrun and deploy


packages/cli/lib/services/dryrun.service.ts
```diff
@@ -310,12 +310,12 @@ export class DryRunService {
                 scriptType: scriptInfo?.type || 'sync',
                 host: process.env['NANGO_HOSTPORT'],
                 connectionId: nangoConnection.connection_id,
-                environmentId: nangoConnection.environment_id,
```
bodinsamuel, line 313: This call never returned environment_id

packages/server/lib/controllers/connection.controller.ts
```diff
@@ -10,8 +10,6 @@ import { connectionCreated as connectionCreatedHook, connectionCreationStartCapC
 import { getOrchestrator } from '../utils/utils.js';
 import { preConnectionDeletion } from '../hooks/connection/on/connection-deleted.js';
 
-export type { ConnectionList };
```
bodinsamuel, line 13: Left over

packages/server/lib/formatters/connection.ts
```diff
@@ -86,7 +86,7 @@ export function connectionFullToPublicApi({
         connection_config: data.connection_config || {},
         created_at: String(data.created_at),
         updated_at: String(data.updated_at),
-        last_fetched_at: String(data.last_fetched_at),
-        credentials: data.credentials as AllAuthCredentials
+        last_fetched_at: data.last_fetched_at ? String(data.last_fetched_at) : null,
```
bodinsamuel, line 89: It could be nullable

packages/shared/lib/services/connection.service.ts
```diff
@@ -355,25 +372,16 @@ class ConnectionService {
...
-            .where({ id: id, deleted: false });
-
-        if (!result || result.length == 0 || !result[0]) {
-            return null;
-        }
+    public async getConnectionById(id: number): Promise<DBConnection | null> {
+        const result = await db.knex.from<DBConnection>('_nango_connections').select<DBConnection>('*').where({ id: id, deleted: false }).first();
 
-        return result[0];
+        return result || null;
```
bodinsamuel, line 379: This had no reason to filter some fields

packages/types/lib/db.ts
```diff
@@ -7,5 +7,10 @@ export interface Deleted {
     deleted_at?: Date | null;
     deleted?: boolean;
 }
+export interface DeletedCorrect {
+    deleted_at: Date | null;
+    deleted: boolean;
+}
```
bodinsamuel, line 13: annoyingly deleted and deleted_at were also set to undefined but only realised that now so all previous DB* types are kind of wrong too

packages/shared/lib/services/connection.service.ts
```diff
@@ -135,20 +136,22 @@ class ConnectionService {
...
+            connection_id: connectionId,
+            provider_config_key: providerConfigKey,
+            config_id: config_id as number,
+            credentials: parsedRawCredentials,
+            connection_config: connectionConfig || {},
+            environment_id: environmentId,
+            metadata: metadata || null,
+            created_at: new Date(),
+            updated_at: new Date(),
+            id: -1,
```
TBonnin, line 150: id: -1?

packages/types/lib/connection/db.ts
```diff
@@ -1,66 +1,56 @@
...
-    id: number;
-    connection_id: string;
-    provider_config_key: string;
-    provider: string;
-    created: string;
-    metadata?: Metadata | null;
-    error_log_id?: number | string | null;
-}
+export type ConnectionInternal = Pick<DBConnection, 'id' | 'connection_id' | 'provider_config_key' | 'environment_id' | 'connection_config'>;
+export type ConnectionJobs = Pick<DBConnection, 'id' | 'connection_id' | 'provider_config_key' | 'environment_id'>;
```
TBonnin, line 56: I would prefer a name that describes what it is rather than where it is being used. Since it is in `types` packages it can end up being used anywhere which would make its name confusing. 
It is hard to find a good name though. What about `ConnectionIds`?

packages/types/lib/db.ts
```diff
@@ -7,5 +7,10 @@ export interface Deleted {
...
 
 export interface TimestampsAndDeleted extends Timestamps, Deleted {}
+export interface TimestampsAndDeletedCorrect extends Timestamps, DeletedCorrect {}
```
TBonnin, line 16: can we have `TimestampsAndDeleted` be the correct one and rename the old one to something like `TimestampsAndDeletedLegacy`

nalanj (approved):

TBonnin (commented):
much needed refactoring. Thank you very much. I left a few comments about naming 😬

packages/types/lib/connection/db.ts
bodinsamuel: Funny because I have the opposite view on this matter; to me, it's like naming a variable with what it contains; it inevitably ends up being wrong, or you need to rename everything. 
Whereas when you name a var for it's usage, you can adapt the shape and you understand when you should not use it.

packages/shared/lib/services/connection.service.ts
bodinsamuel: yeah I'm not sure how to handle that. this connection does not exist, but I need a valid type for the encryption (I'm removing the id just after).
If I Omit the id `encryptConnection`, there is a chance we will forget it later when it matters. Maybe the only good solution is to create more functions, e.g: `encryptNewConnection`

packages/types/lib/db.ts
bodinsamuel: Yes, I will fix that in a next PR (otherwise it's 15more files changed here ahah)

PR #3472, opened 2025-02-06
fix(cron): automatic refresh token was not passing decrypted connection

bodinsamuel:
## Changes 

Noticed something weird while improving the type for Connection in #3470 

- Automatic refresh token was not passing decrypted connection

When doing this refactor https://github.com/NangoHQ/nango/pull/2870 I split the refresh and the getConnection because it was inefficient. I forgot to decrypt the connection, and due to bad typing, there was no difference between Connection and (decrypted)Connection. The really permissive code completely allowed this to pass, and still mark connection has refreshed while not being refreshed.

NB: this was not impacting the regular refresh on fetch

### DIFF
--- a/packages/shared/lib/services/connection.service.ts
+++ b/packages/shared/lib/services/connection.service.ts
@@ -861,6 +861,10 @@ class ConnectionService {
 
         const copy = { ...connection };
 
+        if (!connection.credentials || 'encrypted_credentials' in connection.credentials) {
+            return Err(new NangoError('invalid_crypted_connection'));
+        }
+
         if (
             connection?.credentials?.type === 'OAUTH2' ||
             connection?.credentials?.type === 'APP' ||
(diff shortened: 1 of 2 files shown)


nalanj (approved):

packages/shared/lib/services/connection.service.ts
```diff
@@ -861,6 +861,10 @@ class ConnectionService {
 
         const copy = { ...connection };
 
+        if (!connection.credentials || 'encrypted_credentials' in connection.credentials) {
```
TBonnin, line 864: Not sure I follow the condition here

bodinsamuel: if credentials are empty or still encrypted, it's mostly to avoid regression. When the type refactor will lend it will be more safely typed to avoid this error.

PR #3475, opened 2025-02-06
feat(runner): getObjectsById

nalanj:
Adds new getObjectsById function in syncs for reading existing records.

See https://linear.app/nango/issue/NAN-2646/script-sdk-get-a-list-of-records


packages/cli/lib/services/sdk.ts
```diff
@@ -207,4 +207,8 @@ export class NangoSyncCLI extends NangoSyncBase {
 
         return super.getMetadata<TMetadata>();
     }
+
+    public override getObjectsByIds<K = any, T = any>(_ids: K[], _model: string): Promise<Map<K, T>> {
+        return Promise.resolve(new Map<K, T>());
```
bodinsamuel: I guess the CLI could reach the API wdyt?

packages/runner/lib/sdk/sdk.ts
```diff
@@ -375,12 +376,55 @@ export class NangoSyncRunner extends NangoSyncBase {
...
+
+            const { nextCursor, records } = res.unwrap();
+            cursor = nextCursor;
+
+            for (const record of records) {
+                const stringId = String(record.id);
+                const realId = externalIdMap.get(stringId);
+                if (realId !== undefined) {
+                    const { _nango_metadata, ...recordWithoutMetadata } = record;
+                    objects.set(realId, recordWithoutMetadata as T);
```
bodinsamuel: I'm worried somebody will ask for the metadata, and then we will have a either breaking change or a new argument that changes the return type. I guess most people want to merge, which is why you did it. 
Do you think it's possible to have this `new Map<id, { record, metadata }>`.
```
const records = nango.getRecords(['foobar']);
const myNew = { ...records.get('foobar').record, updated: true };
```

packages/runner/lib/sdk/persist.ts
```diff
@@ -225,4 +233,32 @@ export class PersistClient {
...
+        model,
+        cursor,
+        externalIds
+    }: {
+        environmentId: number;
+        nangoConnectionId: number;
+        model: string;
+        cursor?: string | undefined;
+        externalIds?: string[] | undefined;
+    }): Promise<Result<GetRecordsSuccess>> {
```
bodinsamuel, line 249: There is no way to import and reuse type like we do with the public api?

bodinsamuel (approved):
Looks good 👌🏻

packages/cli/lib/services/sdk.ts
nalanj: Yeah, I was leaning toward getting this through to the finish line and then doing a CLI pass. What do you think?

packages/runner/lib/sdk/sdk.ts
nalanj: Yeah, that was indeed why I did it. I don't love doing it as `Map<{record, meta}>` since it kind of complicates DX for the default use case. I could see two good angles forward:

1. Return `{records: Map<id, T>, metadata: Map<id, Meta>}`. Then at least the usage would look like `const {records} = await nango.getObjectsById(...);`
2. Clean up `batchSave` and `batchUpdate` to ignore `_nango_metadata` which is the issue I ran into that caused me to build this the way I did.
3. Have another function that returns records with metadata.

I know it would technically be a breaking change, but my take is #2 is the best option and I can't imagine that anyone's loading and then saving `_nango_metadata` on purpose.

packages/runner/lib/sdk/persist.ts
nalanj: Not sure what you mean. I was just following the pattern for `getCursor` above.

packages/cli/lib/services/sdk.ts
nalanj: I think there are two options for this, though, and we may need to build both:

1. Load objects over the api
2. Load objects from a fixture file

bodinsamuel: alright to me, it wasn't clear you wanted to finish this later

packages/runner/lib/sdk/persist.ts
bodinsamuel: yeah I'm just pointing out we are not using shared types which could lead to discrepancy

packages/cli/lib/services/sdk.ts
nalanj: I wasn't 100% sure either 😄

bodinsamuel: for the actual options, I think it's fair to load objects from api otherwise dryrun would never work but not sure

packages/runner/lib/sdk/sdk.ts
bodinsamuel: that would be nice actually to ignore `_nango_metadata` because it might already happen in the wild, otherwise I'm not sure. Maybe 1 is the easiest for customers

packages/cli/lib/services/sdk.ts
nalanj: Yeah, I think you're right. I'll probably just build that in dryrun today and include on this PR. This one was very "I don't know what I don't know"

packages/runner/lib/sdk/sdk.ts
nalanj: It's my fave option. Wasn't sure if you or @TBonnin would have a reason it won't work to do that, but thinking through it I'm pretty sure if someone were to save an object with `_nango_metadata` they couldn't access it again because it would be overwritten by the record load.

packages/cli/lib/services/sdk.ts
```diff
@@ -207,4 +207,35 @@ export class NangoSyncCLI extends NangoSyncBase {
 
         return super.getMetadata<TMetadata>();
     }
+
+    public override async getObjectsByIds<K = any, T = any>(ids: K[], model: string): Promise<Map<K, T>> {
```
TBonnin: wrt to `K = any`. should it only allow `string | number` and not any?

TBonnin: are we sure we want to return a Map and not a List?

packages/runner/lib/sdk/persist.ts
TBonnin: GetRecordsSuccess is still a shared types defined in `types` package. 

I think I didn't put the persist `Endpoint` in `types` because then it is impossible to access the method and path since they are types and not values. But maybe it would be better to have the Endpoint in `types` and hardcode the path/method in persist server.ts

packages/runner/lib/sdk/sdk.ts
TBonnin: pretty sure customers are gonna want the metadata to for example have some updating logic that depends on the `last_modified_at`. Personally I would match the return type of `getRecordsByIds` with the one of `getRecords` and return a list. Are customers explicitly asking for a Map?

packages/runner/lib/sdk/persist.ts
bodinsamuel: Ah yes I remember, always annoying this kind of decisions

packages/cli/lib/services/sdk.ts
nalanj: > wrt to K = any. should it only allow string | number and not any?

Right now we allow an incoming id in persist to be anything and we convert to a string - which has some funny side effects for sure. I'm down to say this has to be `string | number` but didn't want to lock it down if someone has other things going in.

> are we sure we want to return a Map and not a List?

My thinking here was that for the main use-case we've heard about customers will want to take some updates they receive over an API or webhook, take the existing record, and map the updates to that. If we don't hand it off as a map customers will probably very often shove it into a map (good!) or find on each record (bad!), so I was hoping this would bias customers toward doing the right thing.

nalanj:
> pretty sure customers are gonna want the metadata to for example have some updating logic that depends on the last_modified_at. Personally I would match the return type of getRecordsByIds with the one of getRecords and return a list. Are customers explicitly asking for a Map?

My plan here is to update `batchSave` and `batchUpdate` to just ignore `_nango_metadata`. Checked in with @khaliqgant and that seems reasonable, and it lets us include metadata here.

> pretty sure customers are gonna want the metadata to for example have some updating logic that depends on the last_modified_at. Personally I would match the return type of getRecordsByIds with the one of getRecords and return a list. Are customers explicitly asking for a Map?

No, but they're saying they need to `getRecordsById` and grab records from that set as they iterate another set of updates they've received. I'm fine with making it a list though, and just letting them do whatever they want with it.

nalanj:
@TBonnin I think I have this finished up. Here's the unauthenticated sync I've been testing against - works in dryrun and in the platform.

packages/runner/lib/sdk/sdk.ts
```diff
@@ -381,12 +382,55 @@ export class NangoSyncRunner extends NangoSyncBase {
         }
         return true;
     }
+
+    public async getRecordsByIds<K = any, T = any>(ids: K[], model: string): Promise<Map<K, T>> {
```
TBonnin: should we limit the number of records that can be requested at once. As it is a customer can pass an unlimited number of ids

packages/runner-sdk/lib/sync.ts
```diff
@@ -32,6 +32,8 @@ export abstract class NangoSyncBase extends NangoActionBase {
 
     public abstract batchUpdate<T extends object>(results: T[], model: string): MaybePromise<boolean>;
 
+    public abstract getRecordsByIds<K = any, T = any>(ids: K[], model: string): MaybePromise<Map<K, T>>;
```
TBonnin: `K = any`: we actually only support `string | number`

packages/runner/lib/sdk/sdk.ts
TBonnin: also would it nicer to make it a generator so customer can paginate over the result?

packages/runner-sdk/lib/sync.ts
nalanj: Weirdly we don't enforce that on the typescript and `nango.yaml` side for customers, but it does indeed blow up if you try to give it a complex id.

nalanj: Fixed in 427257ea9

packages/runner/lib/sdk/sdk.ts
nalanj: I could see doing a generator on a record by record basis so that they're minimizing the resident set. Good call. I'm thinking it would look like:

```
for await (const record of nango.getRecordsByIds(['a', 'b', 'c'], 'Record')) {
  record.newKey = 'biz';
  nextBatch.push(record);
}

...
await nango.batchSave(nextBatch, 'Record');
```

nalanj: I'm not that concerned about allowing unlimited inputs. Customers could technically also write a tight loop that pushes into an array that would cause an OOM.

packages/cli/lib/services/sdk.ts
```diff
@@ -213,4 +213,45 @@ export class NangoSyncCLI extends NangoSyncBase {
...
+        if (ids.length === 0) {
+            return objects;
+        }
+
+        const externalIds = ids.map((id) => String(id).replaceAll('\x00', ''));
+        const externalIdMap = new Map<string, K>(ids.map((id) => [String(id), id]));
+
+        let cursor: string | null = null;
+        for (let i = 0; i < ids.length; i += 100) {
+            const batchIds = externalIds.slice(i, i + 100);
```
TBonnin, line 229: why 100? should it be higher to save a few round trip?

TBonnin (approved):

packages/cli/lib/services/sdk.ts
nalanj: Sure, I was just setting it to some arbitrary value - what would you like it to be?

nalanj:
@TBonnin did you want me to do the generator method? I never heard back from you on that one

packages/cli/lib/services/sdk.ts
nalanj: And my basis for 100 is that records can be up to 100kb, right?, so 100 records at that size would be 10mb which seems kind of big for a response.

TBonnin: good point. we can tweak the value later.

TBonnin:
> @TBonnin did you want me to do the generator method? I never heard back from you on that one

it is not mandatory for now. Let's see how customers use it.

PR #3476, opened 2025-02-06
feat(persist): Add support for activityLogId to get records endpoint

nalanj:
Realized it's ideal to be able to log from persist consistent with what we do on the other endpoints for saving records.

### DIFF
--- a/packages/persist/lib/routes/environment/environmentId/connection/connectionId/getRecords.ts
+++ b/packages/persist/lib/routes/environment/environmentId/connection/connectionId/getRecords.ts
@@ -2,6 +2,8 @@ import type { ApiError, Endpoint, GetRecordsSuccess } from '@nangohq/types';
 import type { EndpointRequest, EndpointResponse, RouteHandler, Route } from '@nangohq/utils';
 import { records } from '@nangohq/records';
 import { validateGetRecords } from './validate.js';
+import type { LogContext } from '@nangohq/logs';
+import { logContextGetter } from '@nangohq/logs';
 
 type GetRecords = Endpoint<{
     Method: typeof method;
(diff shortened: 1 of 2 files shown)


packages/persist/lib/routes/environment/environmentId/connection/connectionId/getRecords.ts
```diff
@@ -28,17 +31,28 @@ const validate = validateGetRecords<GetRecords>();
...
-        limit: 100
+        limit
     });
 
     if (result.isOk()) {
+        await logCtx?.info(`Successfully found ${result.value.records.length} records`, {
+            model,
+            externalIds,
+            limit
+        });
```
bodinsamuel, line 55: Note this HTTP call will also be logged from the runner

packages/persist/lib/routes/environment/environmentId/connection/connectionId/getRecords.ts
```diff
@@ -15,6 +17,7 @@ type GetRecords = Endpoint<{
         externalIds: string[];
         cursor: string;
         limit: number;
+        activityLogId: string;
```
bodinsamuel: I'm surprised it doesn't complain. Same for cursor
```suggestion
        activityLogId: string | undefined;
```

@TBonnin I feel like the type and zod validation are not properly checked against each other

bodinsamuel (approved):

packages/persist/lib/routes/environment/environmentId/connection/connectionId/getRecords.ts
nalanj: Yeah, from the runner was the point of it.

packages/persist/lib/routes/environment/environmentId/connection/connectionId/getRecords.ts
TBonnin: indeed. looks like my `validateGetRecords` function is swallowing the types. Will fix

TBonnin (approved):

PR #3478, opened 2025-02-07
docs: Standardize providers.yaml display names

bastienbeurier:
Guidelines I used: 
- Only specify auth type when there are multiple auth types for a given API
- Capitalize auth type
- Uppercase for acronyms

### DIFF
--- a/packages/providers/providers.yaml
+++ b/packages/providers/providers.yaml
@@ -146,7 +146,7 @@ affinity:
             doc_section: '#step-1-finding-your-api-key'
 
 aircall:
-    display_name: Aircall
+    display_name: Aircall (OAuth)
     categories:
         - support
     auth_mode: OAUTH2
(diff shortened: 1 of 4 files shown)


bastienbeurier:
@nalanj just checking that this will sync automatically with Webflow when merged.

bodinsamuel (approved):
Could be added to scripts/validation/providers/validate.ts so it's enforced

bastienbeurier:
> Could be added to scripts/validation/providers/validate.ts so it's enforced

Agreed, though some edge cases might be challenging to systematize.

nalanj:
@bastienbeurier it will indeed sync after merge

nalanj:
Worth noting - I've bumped into a few cases where the document title doesn't match the provider name. When I created all the new ones I tried to set them correctly, but it's another kind of grunt work thing that probably should be knocked out.

bastienbeurier:
> Worth noting - I've bumped into a few cases where the document title doesn't match the provider name. When I created all the new ones I tried to set them correctly, but it's another kind of grunt work thing that probably should be knocked out.

@nalanj the API docs page title is not synced with the `providers.yaml` display name?

nalanj:
@bastienbeurier it is not

PR #3485, opened 2025-02-08
fix: decrypt provider config in refreshConnections

TBonnin:
We were passing the encrypted oauth client secret to refreshOrTestCredentials because unlike the other method in config.service the getStaleConnections function doesn't decrypt the oauth client secret :(

Ideally we should have different types for encrypted/decrypted entities to prevent passing the wrong one

### DIFF
--- a/packages/server/lib/refreshConnections.ts
+++ b/packages/server/lib/refreshConnections.ts
@@ -65,14 +65,25 @@ export async function exec(): Promise<void> {
                     }
 
                     const { connection, account, environment, integration } = staleConnection;
-                    const decrypted = encryptionManager.decryptConnection(connection);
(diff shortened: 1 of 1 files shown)


packages/server/lib/refreshConnections.ts
```diff
@@ -65,14 +65,25 @@ export async function exec(): Promise<void> {
                     }
 
                     const { connection, account, environment, integration } = staleConnection;
-                    const decrypted = encryptionManager.decryptConnection(connection);
-                    logger.info(`${cronName} refreshing connection '${connection.connection_id}' for accountId '${account.id}'`);
```
TBonnin, line 69: oups I removed this log. Will fix on Monday

PR #3493, opened 2025-02-10
fix(jobs): logs track_deletes results

bodinsamuel:
## Changes

- logs track_deletes results per model
- throw if we can't find a job


packages/shared/lib/services/sync/job.service.ts
```diff
@@ -133,21 +137,17 @@ export const updateSyncJobResult = async (id: number, result: SyncResultByModel,
...
-            const finalResult = {
+            const deletedValue = Number(deleted) || 0;
+            const incomingDeletedValue = Number(incomingResult?.deleted) || 0;
+            const finalResult: SyncResultByModel = {
                 ...existingResult,
                 [model]: {
                     added: Number(added) + Number(incomingResult?.added),
-                    updated: Number(updated) + Number(incomingResult?.updated)
+                    updated: Number(updated) + Number(incomingResult?.updated),
+                    deleted: deletedValue + incomingDeletedValue
```
bodinsamuel, line 147: Not sure why it was done conditionally, maybe it's wrong, let me know

khaliqgant: Think this was from legacy as we didn't want to show deletes at all since that wasn't a thing in the initial implementation.

khaliqgant (approved):

nalanj (approved):

packages/shared/lib/services/sync/job.service.ts
TBonnin: is it also what we send back in the webhook? Can customers have logic not expecting `deleted` attributes?

TBonnin (approved):

packages/shared/lib/services/sync/job.service.ts
bodinsamuel: We always send at least `deleted: 0` in the webhook by default (see: webhooks/lib/sync.ts:101)
So I don't think it's an issue here

PR #3496, opened 2025-02-10
feat(sync): logs unchanged in results

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2916/log-unchanged-records

> [!NOTE]
> This is a proposal: it's hard to debug how many records have been sent and the logs only say "0 saved" which adds to the confusion. Same for the webhook. 

- Logs unchanged in results and webhook
If this PR is accepted, we will need to announce a breaking change in the webhook shape just in case, because people might be using strict struct to parse the json.

### DIFF
--- a/packages/jobs/lib/execution/sync.integration.test.ts
+++ b/packages/jobs/lib/execution/sync.integration.test.ts
@@ -4,7 +4,7 @@ import type { UnencryptedRecordData, ReturnedRecord } from '@nangohq/records';
 import { records as recordsService, format as recordsFormatter, migrate as migrateRecords, clearDbTestsOnly as clearRecordsDb } from '@nangohq/records';
 import { handleSyncSuccess, startSync } from './sync.js';
 import type { TaskAction, TaskOnEvent, TaskSync, TaskSyncAbort, TaskWebhook } from '@nangohq/nango-orchestrator';
-import type { Sync, SyncResult, Job as SyncJob } from '@nangohq/shared';
+import type { Sync, SyncResult, Job as SyncJob, SyncResultByModel } from '@nangohq/shared';
 import { isSyncJobRunning, seeders, getLatestSyncJob, updateSyncJobResult } from '@nangohq/shared';
 import { Ok, stringifyError } from '@nangohq/utils';
 import { envs } from '../env.js';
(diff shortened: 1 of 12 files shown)


TBonnin (commented):
I think it makes sense cc @bastienbeurier

TBonnin (approved):

bastienbeurier:
@bodinsamuel quick question: unchanged records are: 
1. The ones that have been fetched from the API but are the same as in the cache, so unchanged in the cache or
2. The ones that have not been fetched from the API, so unchanged in the cache or
3. Both

bodinsamuel:
Unchanged = 
> The ones that have been fetched from the API but are the same as in the cache, so unchanged in the cache 

---
Those ones are deleted when using track_deletes. If you don't have track_deletes those are indeed unchanged (untouched is most accurate)
> The ones that have not been fetched from the API, so unchanged in the cache

bastienbeurier:
> Unchanged =
> 
> > The ones that have been fetched from the API but are the same as in the cache, so unchanged in the cache
> 
> Those ones are deleted when using track_deletes. If you don't have track_deletes those are indeed unchanged (untouched is most accurate)
> 
> > The ones that have not been fetched from the API, so unchanged in the cache

Got it. There's something unintuitive to this in my opinion. Can we leverage logging to surface this information? That way we avoid a breaking change too.

bodinsamuel:
> Got it. There's something unintuitive to this in my opinion. Can we leverage logging to surface this information? That way we avoid a breaking change too.

This feature is almost entirely for logging so yes. the only "breaking change" is adding this information to the webhook

>  There's something unintuitive to this in my opinion.

Do you have a suggestion? Keep in mind there are different context around this:
- We call batchSave, it can contain new record, updated record, same record = added, updated, unchanged
  You can't have untouched record here
- At the end of the sync, we aggregate all those metrics. At this stage we could compute the "untouched" record. 
Happy to add this information at this stage

bastienbeurier:
> > Got it. There's something unintuitive to this in my opinion. Can we leverage logging to surface this information? That way we avoid a breaking change too.
> 
> This feature is almost entirely for logging so yes. the only "breaking change" is adding this information to the webhook
> 
> > There's something unintuitive to this in my opinion.
> 
> Do you have a suggestion? Keep in mind there are different context around this:
> 
> * We call batchSave, it can contain new record, updated record, same record = added, updated, unchanged
>   You can't have untouched record here
> * At the end of the sync, we aggregate all those metrics. At this stage we could compute the "untouched" record.
>   Happy to add this information at this stage

If we want to account for every case, it would be: 
- Added – New record fetched and saved.
- Updated – Record fetched, modified, and saved.
- Re-saved (unchanged) – Record fetched, but no changes detected; still saved.
- Unfetched (unchanged) – Record not fetched in this sync but remains in cache (incremental sync).
- Deleted (explicit) – Record was marked as deleted by the external API and removed.
- Deleted (implicit) – Record was not fetched and is assumed deleted (full sync).

IMO I would only add this in the Logs tab execution operation, but wouldn't necessarily change the webhook, since I'm not sure teams would take programmatic action on this.

PR #3498, opened 2025-02-10
fix(persist): Filter external ids to remove 0x00

nalanj:
Persist was having issues where if an external id included `0x00` inserts would fail. This fixes that.

- for `getRecords` there's a new test
- for upsert and updating in persist I set up a sync that had `0x00` in its id because the code path where the external id is extracted currently isn't tested


packages/records/lib/helpers/format.ts
```diff
@@ -46,7 +46,7 @@ export const formatRecords = ({
         const formattedRecord: FormattedRecord = {
             id: stableId(datum),
             json: datum,
-            external_id: String(datum['id']),
+            external_id: String(datum['id']).replaceAll('\x00', ''),
```
TBonnin, line 50: can we add a comment to say why

TBonnin (approved):

packages/records/lib/helpers/format.ts
nalanj: Done in b4e250909

PR #3499, opened 2025-02-10
fix(records): wrong nextMerging cursor when last record was not updated

TBonnin:
Next merging cursor was wrongly set to the last record returned by the query even when the last record was unchanged (aka: same values and therefore same old updated_at). 
In this case this old record will be used for subsequent batch operations as the cursor and prevent more recent records to be updated.
Coupled with `track_deletes=true` it could lead to records be wrongly discarded and therefore deleted at the end of the sync

added integration tests


bodinsamuel (approved):
🥳

PR #3505, opened 2025-02-11
fix: wait for 10 successes to consider a runner healthy

TBonnin:
when a runner goes live we had a few cases where the fleet supervisor would successfully reach the /health endpoint and mark it as running but then a script execution would fail trying to reach the runner. My hypothesis is that Render network config can take a bit of time to propagate and even if a runner can be reached from one jobs instance (the one running fleet supervisor) there is not guarantee it can also be reached immediately by all jobs instances.
This commit requires consecutive successes when making sure the runner is healthy before it can be registered. It is making the registration slower but it is a one-off thing

### DIFF
--- a/packages/fleet/lib/utils/url.ts
+++ b/packages/fleet/lib/utils/url.ts
@@ -2,18 +2,25 @@ import type { Result } from '@nangohq/utils';
 import { Err, Ok } from '@nangohq/utils';
 import { setTimeout } from 'node:timers/promises';
 
-export async function waithUntilHealthy({ url, timeoutMs }: { url: string; timeoutMs: number }): Promise<Result<void>> {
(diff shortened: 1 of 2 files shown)


packages/fleet/lib/utils/url.ts
```diff
@@ -2,18 +2,25 @@ import type { Result } from '@nangohq/utils';
 import { Err, Ok } from '@nangohq/utils';
 import { setTimeout } from 'node:timers/promises';
 
-export async function waithUntilHealthy({ url, timeoutMs }: { url: string; timeoutMs: number }): Promise<Result<void>> {
+export async function waitUntilHealthy({ url, timeoutMs }: { url: string; timeoutMs: number }): Promise<Result<void>> {
```
TBonnin, line 5: unrelated but typo

bodinsamuel (approved):

PR #3507, opened 2025-02-11
fix: authenticate memory warning logs from runner to persist

TBonnin:
Making sure the correct authorization header is passed when making a call from runner to log memory usage warnings.

It was much easier to fix than I thought I wish I had looked into it earlier.

To test it locally you are gonna need to modify/remove the condition `if (memoryUsagePercentage > MEMORY_WARNING_PERCENTAGE_THRESHOLD)`

### DIFF
--- a/packages/utils/lib/environment/parse.ts
+++ b/packages/utils/lib/environment/parse.ts
@@ -56,6 +56,7 @@ export const ENVS = z.object({
     IDLE_MAX_DURATION_MS: z.coerce.number().default(0),
     RUNNER_NODE_ID: z.coerce.number().optional(),
     RUNNER_URL: z.string().url().optional(),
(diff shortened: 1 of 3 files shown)


packages/utils/lib/environment/parse.ts
```diff
@@ -56,6 +56,7 @@ export const ENVS = z.object({
     IDLE_MAX_DURATION_MS: z.coerce.number().default(0),
     RUNNER_NODE_ID: z.coerce.number().optional(),
     RUNNER_URL: z.string().url().optional(),
+    RUNNER_MEMORY_WARNING_THRESHOLD: z.coerce.number().optional().default(85),
```
TBonnin, line 59: increase the limit to 85 because I don't want to spam logs too much. We can tweak this value later

bodinsamuel (approved):

PR #3513, opened 2025-02-11
feat(sdk): Ignore _nango_metadata field in batchSave, batchUpdate, batchDelete

nalanj:
Previously we've allowed _nango_metadata to be submitted to our persist endpoints, which will save that data. Subsequent loading of those records will then overwrite whatever is saved.

## How I tested it

- Ran tests
- Ran syncs to make sure it's filtered


packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
```diff
@@ -275,15 +275,15 @@ export interface NangoProps {
     track_deletes?: boolean;
     attributes?: object | undefined;
     logMessages?:
-        | {
```
nalanj, line 278: I'm not totally sure why these keep re-formatting. It has happened to me a few times lately.

packages/runner-sdk/lib/sync.ts
```diff
@@ -55,4 +55,15 @@ export abstract class NangoSyncBase extends NangoActionBase {
...
     }
+
+    protected removeMetadata<T extends object>(results: T[]) {
+        return results.map((result) => {
+            if ('_nango_metadata' in result) {
+                const { _nango_metadata, ...rest } = result;
+                return rest;
+            }
+
+            return result;
```
bodinsamuel, line 65: It's not possible with `delete`? that might help to not reassign a new object

bodinsamuel (approved):

packages/runner-sdk/lib/sync.ts
nalanj: I think it is possible. When I read the delete code path we apply the records to the upsert, along with the delete, so there's a weird possibility of updating a record when soft deleting.

bodinsamuel: Not sure I get what you mean. I was just saying doing:
```ts
            delete result._nango_metadata;
            return result;
});
```

nalanj: Oh, gotcha. I was thinking we had an eslint rule against that.

I was just saying that deleteBatch actually updates records as well as soft deleting, so we still have to filter it out.

nalanj: See 2e20deeca

TBonnin (approved):

PR #3515, opened 2025-02-12
fix(logs): correctly identify internal logs

bodinsamuel:
## Changes 

- Fix `source` not being set to internal in the SDK
`source: user` should only be `nango.log()`


packages/shared/lib/services/proxy.service.ts
```diff
@@ -50,13 +50,20 @@ class ProxyService {
         const { connection, providerName } = internalConfig;
 
         if (!passedEndpoint && !baseUrlOverride) {
-            logs.push({ type: 'log', level: 'error', createdAt: new Date().toISOString(), message: 'Proxy: a API URL endpoint is missing.' });
+            logs.push({
+                type: 'log',
+                level: 'error',
+                source: 'internal',
+                createdAt: new Date().toISOString(),
+                message: 'Proxy: a API URL endpoint is missing.'
```
nalanj:
```suggestion
                message: 'Proxy: an API URL endpoint is missing.'
```

nalanj (approved):

TBonnin (approved):

PR #3517, opened 2025-02-12
chore: Add script to automate changelog generation for providers

nalanj:
Dumps markdown to show what providers were added when. Run with `npm run changelog:providers`

See https://linear.app/nango/issue/NAN-2619/changelog-automation-new-apis-and-pre-built-integrations


package.json
```diff
@@ -46,7 +46,8 @@
         "docs": "tsx scripts/docs-gen-snippets.ts && cd ./docs-v2 && npx mintlify dev --port 3033",
         "docs:generate": "tsx scripts/docs-gen-snippets.ts",
         "nango": "cd ./packages/server && npm run nango",
-        "dev:docker": "docker compose --file dev/docker-compose.dev.yaml up -d"
+        "dev:docker": "docker compose --file dev/docker-compose.dev.yaml up -d",
+        "changelog:providers": "tsx scripts/changelog/providers/changelog.ts"
```
bodinsamuel: script path is wrong

bodinsamuel (approved):
works great 👌🏻

package.json
nalanj: 🤦🏻

nalanj: Fixed in 8493deaf6

PR #3518, opened 2025-02-12
chore(dd): try to change some spampling

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2703/fine-tune-datadog-tracer

- Try to add a little bit of sampling on some resources 
Honestly, it's hard to pick a number so I just played it safe.

- Added missing persist-net 
- Removed dns persist that I wrongly set to true


nalanj:
@bodinsamuel Are we seeing DD have issues keeping up? I thought by default it tries to cap the root sample rate to around 10 per second.

bodinsamuel:
Sorry it's missing context, we are sending way too many traces for our plan so we are trying to optimize for the most important

TBonnin (approved):
let's see

PR #3519, opened 2025-02-12
fix: pause schedules when disabling integration

TBonnin:
when disabling an integration we currently only disable the sync config but we don't pause the schedules, which means we have schedules in STARTED state which are linked to disabled configs. The orchestrator is scheduling tasks for this schedules, even though the tasks fails quickly because the sync config is disable. 
Once this fix is merged, I will go over the existing schedules and paused every one of them that is associated to a disabled sync config. 

How to test:
1. Disable/enable integrations endpoints
2. check in db that the schedules state changes to PAUSED/STARTED

### DIFF
--- a/packages/shared/lib/services/sync/manager.service.ts
+++ b/packages/shared/lib/services/sync/manager.service.ts
@@ -8,7 +8,8 @@ import {
     getSyncsByProviderConfigAndSyncNames,
     getSyncByIdAndName,
     getSyncNamesByConnectionId,
-    softDeleteSync
+    softDeleteSync,
+    getSyncsBySyncConfigId
 } from './sync.service.js';
 import { errorNotificationService } from '../notification/error.service.js';
 import configService from '../config.service.js';
@@ -77,6 +78,12 @@ export class SyncManagerService {
             if (!syncConfig) {
                 continue;
             }
+
+            const existingSync = await getSyncByIdAndName(nangoConnectionId, syncConfig.sync_name);
+            if (existingSync) {
+                await orchestrator.unpauseSync({ syncId: existingSync.id, environmentId: nangoConnection.environment_id });
+                continue;
+            }
             const sync = await createSync(nangoConnectionId, syncConfig);
             if (sync) {
                 await orchestrator.scheduleSync({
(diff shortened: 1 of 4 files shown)


packages/shared/lib/services/sync/manager.service.ts
```diff
@@ -96,7 +102,7 @@ export class SyncManagerService {
         syncName: string,
         providerConfigKey: string,
         environmentId: number,
-        sync: IncomingFlowConfig,
+        flowConfig: IncomingFlowConfig,
```
TBonnin, line 106: unrelated but I though name was misleading

packages/shared/lib/services/sync/sync.service.ts
```diff
@@ -57,12 +57,6 @@ export const getById = async (id: string): Promise<Sync | null> => {
 };
 
 export const createSync = async (nangoConnectionId: number, syncConfig: DBSyncConfig): Promise<Sync | null> => {
-    const existingSync = await getSyncByIdAndName(nangoConnectionId, syncConfig.sync_name);
-
-    if (existingSync || !syncConfig.id) {
-        return null;
-    }
-
```
TBonnin, line 65: createSync is only used in the 2 cases above. Checking for an existing sync has been moved outside of the function

packages/shared/lib/services/sync/manager.service.ts
```diff
@@ -8,12 +8,12 @@ import {
...
     getSyncNamesByConnectionId,
-    softDeleteSync
+    softDeleteSync,
+    getSyncsBySyncConfigId
 } from './sync.service.js';
 import { errorNotificationService } from '../notification/error.service.js';
 import configService from '../config.service.js';
-import type { SyncWithConnectionId, ReportedSyncJobStatus, SyncCommand } from '../../models/Sync.js';
-import { SyncJobsType, SyncStatus } from '../../models/Sync.js';
+import type { SyncWithConnectionId, ReportedSyncJobStatus, SyncCommand, SyncJobsType, SyncStatus } from '../../models/Sync.js';
```
nalanj: Pretty sure all the build issues are from this being imported as a type.

nalanj (commented):
Had to clean up those imports to get it to build, but once I did it seemed to work well!

nalanj (approved):

PR #3521, opened 2025-02-13
feat: introduce sync variant

TBonnin:
Adding the 'variant' column to the syncs table with a default 'base'value and add variant parameters to sync related function so we can pass a different value later.
This commit doesn't change any behavior. There is still no way to create a variant and cardinality of sync_config:sync is still 1:1 

Next steps:
- modify existing public function/api (trigger, pause, status, ...) to accept a sync name + variant
- expose variant into the script
- modify runner-sdk to save variant records and modify getRecords to handle variants
- add new functions/api to be able to create a variant
- update docs

### DIFF
--- a/packages/database/lib/migrations/20250211192007_sync_variant.cjs
+++ b/packages/database/lib/migrations/20250211192007_sync_variant.cjs
@@ -1,0 +1,15 @@
+const TABLE = '_nango_syncs';
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.schema.raw(`
+        ALTER TABLE "${TABLE}"
+        ADD COLUMN IF NOT EXISTS "variant" varchar(255) NOT NULL DEFAULT 'base'
+    `);
+};
+
+exports.down = function () {
+    // do nothing
+};
--- a/packages/database/lib/migrations/20250211192007_sync_variant.cjs
+++ b/packages/database/lib/migrations/20250211192007_sync_variant.cjs
@@ -1,3 +1,5 @@
+exports.config = { transaction: false };
+
 const TABLE = '_nango_syncs';
 
 /**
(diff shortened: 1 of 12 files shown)


packages/server/lib/controllers/flow.controller.ts
```diff
@@ -58,7 +58,12 @@ class FlowController {
                 return;
             }
 
-            await syncManager.triggerIfConnectionsExist(preBuiltResponse.result, environment.id, logContextGetter, orchestrator);
+            await syncManager.triggerIfConnectionsExist({
+                flows: preBuiltResponse.result,
+                environmentId: environment.id,
+                logContextGetter,
+                orchestrator
+            });
```
TBonnin, line 66: just changing the parameters into an object for now. At some point,  I will have to pass the variant

packages/database/lib/migrations/20250211192007_sync_variant.cjs
```diff
@@ -0,0 +1,15 @@
+const TABLE = '_nango_syncs';
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.schema.raw(`
+        ALTER TABLE "${TABLE}"
+        ADD COLUMN IF NOT EXISTS "variant" varchar(255) NOT NULL DEFAULT 'base'
```
nalanj, line 11: Should the variant be unique against the identity for the sync? Seems like it's connection_id, name, and deleted, but I'm not 100% sure.

nalanj (approved):
Just the one migration comment - seems good. Ran it locally and was able to set up a new connection.

packages/database/lib/migrations/20250211192007_sync_variant.cjs
TBonnin: you are right it should be unique. Let me add it

TBonnin: added [2a4d7da](https://github.com/NangoHQ/nango/pull/3521/commits/2a4d7dafe7bd8fe259b85e4c7d494d098db94255)
I will remove the current unique constraint when I allow the creation of a sync variant

PR #3522, opened 2025-02-13
feat(sync variant): add sync variant to schedule and nangoProps

TBonnin:
Passing sync variant as part of schedule properties (adding support for variant to jobs and orchestrator along the way) so it can be exposed via NangoProps. For now we are doing nothing with it.

How to test:
- check that existing syncs can run successfully
- deploy a new sync or enable a pre-built one that you have never enabled. After running it check in the db that schedules and tasks now have a syncVariant field in their payload


packages/shared/lib/services/sync/manager.service.ts
```diff
@@ -40,13 +40,23 @@ export interface CreateSyncArgs {
...
+        connectionId,
+        syncVariant,
+        logContextGetter,
+        orchestrator
+    }: {
+        connectionId: number;
+        syncVariant: string;
+        logContextGetter: LogContextGetter;
+        orchestrator: Orchestrator;
+    }): Promise<void> {
```
TBonnin, line 59: unrelated but keeping it consistent with the other function `createSyncForConnections` with a `s`

nalanj (approved):

PR #3526, opened 2025-02-14
chore: Integrations changelog

nalanj:
This dumps markdown for Bastien to use for reporting our changelog for use cases. It reads flows.yaml over time. It seems like it's not very good before November of last year due to changes to the flows format, but it should be useful going forward.

Runnable with:

```
npm run changelog:integrations
```


bodinsamuel:
Fixed the parsing of the old endpoint format and typing

bodinsamuel (approved):

PR #3529, opened 2025-02-17
fix(ui): inline script configuration

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2441/surface-which-integration-endpoints-are-syncs-or-actions

- Inline script configuration
In the endpoint page (vs drawer before)

> It's easier to review with "Hide whitespace"

Screenshot 2025-02-17 at 14 56 13


packages/webapp/src/pages/Integrations/components/ScriptToggle.tsx
```diff
@@ -108,7 +108,16 @@ export const ScriptToggle: React.FC<{
...
+                    <div>
+                        <Switch
+                            name="script"
+                            checked={flow.enabled === true}
+                            onClick={(e) => {
+                                e.preventDefault();
+                                toggleSync();
+                            }}
+                        />
+                    </div>
```
bodinsamuel, line 120: Replaced with shadcn component

khaliqgant (approved):

PR #3532, opened 2025-02-18
docs: update basic auth configuration field names 1 of 3

viictoo:
Update Basic Auth configuration field names

[EXT-287](https://linear.app/nango/issue/EXT-287/update-basic-auth-configuration-field-names-1-of-3)


packages/providers/providers.yaml
```diff
@@ -1106,6 +1119,15 @@ booking-com:
             example: secure-supply|supply
             suffix: -xml.booking.com
             prefix: https://
+    credentials:
+        username:
+            type: string
+            title: Machine Account Username
```
bodinsamuel, line 1125: Machine account? Might be clearer with just Booking.com Username no?

bodinsamuel (approved):

packages/providers/providers.yaml
khaliqgant: Thanks for the review, think this is fine

PR #3537, opened 2025-02-18
fix(orchestrator): return 408 on timeout

bodinsamuel:
## Changes

- return 408 on timeout instead of 500
To reduce the number of identified error in the logs


khaliqgant (approved):

packages/orchestrator/lib/routes/v1/tasks/taskId/getOutput.ts
```diff
@@ -51,7 +51,7 @@ const handler = (scheduler: Scheduler, eventEmitter: EventEmitter) => {
             cleanupAndRespond((res) => res.status(200).json({ state: completedTask.state, output: completedTask.output }));
         };
         const timeout = setTimeout(() => {
-            cleanupAndRespond((res) => res.status(500).send({ error: { code: 'timeout', message: 'Long polling timeout' } }));
+            cleanupAndRespond((res) => res.status(408).send({ error: { code: 'timeout', message: 'Long polling timeout' } }));
```
TBonnin, line 54: That's fine, the status code is not used. The client logic relies on the `error` field in the content.
https://github.com/NangoHQ/nango/blob/2ec075ad327fe120897ae2b0e685abb4387693ad/packages/orchestrator/lib/clients/client.ts#L177-L185

TBonnin (approved):

PR #3538, opened 2025-02-18
feat(integrations): add support for amazon selling partner api

hassan254-prog:

- Add support for the Amazon Selling Partner API, I split the two, as they have different authorization parameters.


packages/providers/providers.yaml
```diff
@@ -306,6 +306,111 @@ amazon:
             example: com
             pattern: '^[a-z.]+$'
 
+amazon-selling-partner:
+    display_name: Amazon Selling Partner
+    categories:
+        - dev-tools
+        - e-commerce
+    auth_mode: OAUTH2
+    authorization_url: https://${connectionConfig.subdomain}.amazon.${connectionConfig.extension}/authorize/consent
```
bodinsamuel: It's not easier to just ask for the full domain and enforce the schema `subdomain.amazon.tld`?

packages/providers/providers.yaml
```diff
@@ -306,6 +306,111 @@ amazon:
...
+        base_url: https://sellingpartnerapi-${connectionConfig.region}.amazon.com
+    docs: https://docs.nango.dev/integrations/all/amazon-selling-partner
+    docs_connect: https://docs.nango.dev/integrations/all/amazon-selling-partner/connect
+    connection_config:
+        applicationId:
+            type: string
+            title: Application ID
+            description: The unique identifier for your Amazon Selling Partner application
+            example: 'amzn1.sellerapps.app.0bf296b5-36a6-4942-a13e-EXAMPLEfcd28'
+            pattern: '^[a-z0-9.-]+$'
```
bodinsamuel: this pattern does not validate the EXAMPLE in the example

bodinsamuel (approved):
Looks mostly good 👌🏻

packages/providers/providers.yaml
hassan254-prog: I have updated this with the latest commit, this hasn't been tested, talked to the community member and they are willing to test it on their end as they couldn't share the credentials.

packages/providers/providers.yaml
hassan254-prog: I have updated this and implemented the following pattern. @bodinsamuel, could you kindly confirm if this is the best solution? 

```
example: sellercentral.amazon.com
pattern: '^[a-z0-9.-]+\.amazon\.[a-z.]+$'
```

PR #3540, opened 2025-02-19
feat(proxy): full rewrite, controlled backoff, better logging, shared logic, etc.

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2736/new-retry-strategy-for-proxy
Fixes https://linear.app/nango/issue/NAN-2733/logs-are-stacked-during-http-retries-and-can-be-dropped-in-case-of
Fixes https://linear.app/nango/issue/NAN-2729/change-default-retry-strategy-update-docs
Contributes to https://linear.app/nango/issue/NAN-2671/investigate-rate-limit-and-observability

- **Rewrite `proxy.service.ts`**
Split that big file into 3 components: 
  - `request.ts` call the third-party, retry and log
  - `retry.ts` unit testable methods to determine retry strategy
  - `utils.ts` to generate proxy configuration

- **New retry method `retryFlexible`** 
The name is lame, but it's hard. The main goal of this method is to provide sane default (exponential backoff) with the ability to log the attempts, modify the wait time on the fly, and eventually skip.

## 🧪 Tests

Since it's a core part of the product I'll appreciate some actual local tests:
- Go to your dashboard
- Do some proxy calls with the API
- Execute syncs that do proxy calls
- Make sure to test with retries

**No retries**
Screenshot 2025-02-19 at 15 13 58

**With retries**
Screenshot 2025-02-19 at 15 14 17


packages/database/lib/index.ts
```diff
@@ -17,16 +17,16 @@ export class KnexDatabase {
...
                     directory: directory,
                     tableName: '_nango_auth_migrations',
                     schemaName: this.schema()
-                }),
+                });
+            },
             {
                 maxAttempts: 4,
-                delayMs: (attempt) => 500 * attempt,
-                retryIf: () => true
```
bodinsamuel, line 29: It's unclear why there was a retry here

packages/server/lib/controllers/proxy.controller.ts
```diff
@@ -229,34 +235,7 @@ class ProxyController {
...
-            request: {
-                method: config.method,
-                url: redactedURL,
-                headers: safeHeaders
-            },
-            response: {
-                code: responseStream.status,
-                headers: redactHeaders({ headers: responseStream.headers as Record<string, string> })
-            }
-        });
```
bodinsamuel, line 258: This part is now logged the same way whether it's api or sdk

packages/shared/lib/services/proxy/request.ts
```diff
@@ -0,0 +1,189 @@
...
+                        }
+
+                        return retry;
+                    }
+                }
+            );
+            return Ok(response);
+        } catch (err) {
+            return Err(err instanceof Error ? err : new ProxyError('unknown_error', '', err));
+        }
```
bodinsamuel, line 97: this method is the core of the PR

packages/utils/lib/retry.ts
```diff
@@ -33,6 +35,63 @@ export async function retry<T>(fn: () => T, { maxAttempts, delayMs, retryIf = ()
...
+    throw new Error('unreachable');
+}
+
+/**
+ * Get exponential backoff with a cap
+ * Base 2, minimum 1000ms
+ */
+export function getExponentialBackoff(attempt: number, maxDelay: number): number {
+    return Math.min(1000 * 2 ** attempt, maxDelay);
+}
```
bodinsamuel, line 93: @bastienbeurier With the new version, I have set up base 1000ms, exp 2, capped to 15minutes. (no cap on maximum retries rn that could break some existing syncs)
Do you want me to change it to something else?

bastienbeurier: This works IMO, but I tweak it to: 
- 3s initial delay
- 2x multiplier for exponential back-off
- 10min max delay
- no max retries

bodinsamuel: Okay I'll change to 3s and 10min. Note we will reach 10 minutes after the 7 retry `0s, 3s, 9s, 21s, 45s, 93s, 189s, 381s, 765s (8th capped), 1533s (9th capped), 3069s (10th capped), etc.`

bodinsamuel:
Tested on staging

khaliqgant (approved):
Code looks good from my initial high level read. I can take a deeper look and potentially test tomorrow but don’t want to block as I know this is high priority for a customer

bodinsamuel:
>Code looks good from my initial high level read. I can take a deeper look and potentially test tomorrow but don’t want to block as I know this is high priority for a customer

I prefer to be extra cautious than breaking stuff for everybody :D

packages/shared/lib/services/proxy/request.ts
```diff
@@ -0,0 +1,189 @@
...
+                    max: this.config.retries || 0,
+                    onError: async ({ err, nextWait, max, attempt }) => {
+                        const retry = getProxyRetryFromErr({ err, proxyConfig: this.config });
+                        if (retry.retry) {
+                            await this.logger({
+                                type: 'log',
+                                level: 'warn',
+                                source: 'internal',
+                                createdAt: new Date().toISOString(),
+                                message: `Retrying HTTP call (reason: ${retry.reason}). Waiting for ${retry.wait ? retry.wait : nextWait}ms [${attempt}/${max}]`
```
khaliqgant, line 78: Would be nice to know what status code triggered it

khaliqgant: I realize that it can be seen in the error log right before it but I still think it would be good to be very explicit if possible.

bodinsamuel: yeah agreed 👍🏻

khaliqgant:
Tested against a few real world scenarios. Looks good

image
image

bodinsamuel:
Thanks for taking the time @khaliqgant much appreciated 🙏🏻

packages/shared/lib/services/proxy/request.ts
bodinsamuel: fixed in [122eee1](https://github.com/NangoHQ/nango/pull/3540/commits/122eee1535c95a4f9668b27e8e0a79019aef81fb)

packages/database/lib/index.ts
TBonnin: isn't it to handle race condition when deploying server and multiple instance tries to run the migrations?

packages/shared/lib/services/proxy/retry.ts
```diff
@@ -0,0 +1,121 @@
...
+            } else if (customHeaderConf.remaining && err.response && err.response.headers[customHeaderConf.remaining] === '0') {
+                // Custom header in providers.yaml
+                isRetryable = true;
+                reason = 'provider_remaining';
+            }
+        }
+    }
+    if (!isRetryable) {
+        return { retry: false, reason: 'not_retryable' };
+    }
```
TBonnin, line 51: not sure why there is 3 `if (!isRetryable)` blocks here and not 
```
if (!isRetryable) {
    if (proxyConfig.retryOn?.includes(status)) {
        ...
    }
    if (customHeaderConf) {
        ....
    }
    return { retry: false, reason: 'not_retryable' };
}
```

packages/shared/lib/services/proxy/retry.ts
```diff
@@ -0,0 +1,121 @@
...
+ */
+export function getRetryFromHeader({
+    err,
+    type,
+    retryHeader
+}: {
+    err: AxiosError;
+    type: 'at' | 'after';
+    retryHeader: string;
+}): { found: false; reason: string } | { found: true; reason: string; wait: number } {
```
TBonnin, line 89: we need an `Option` type ;)

PR #3545, opened 2025-02-20
feat(integrations): add support for airtable pat

hassan254-prog:

- Add support for Airtable (PAT)


packages/providers/providers.yaml
```diff
@@ -241,6 +241,26 @@ airtable:
     webhook_routing_script: airtableWebhookRouting
     docs: https://docs.nango.dev/integrations/all/airtable
 
+airtable-pat:
+    display_name: Airtable (Personal Access Token)
+    categories:
+        - productivity
+    auth_mode: API_KEY
+    proxy:
```
hassan254-prog, line 249: The PAT is scoped hence the reason for omitting `proxy.verification`, although tested locally.

khaliqgant (approved):

PR #3546, opened 2025-02-20
feat(records): use read replicas when possible

bodinsamuel:
## Changes

- Introduce a way to connect to read replicas when configured
Depending on the system, it's not easy to have a load balancer directly in front of the DB cluster. I
In our case, this would be a big change, so to make it easier, we can just have 2 pools. Read is optional and will fallback to the regular pool if not specified.

Tested locally, this won't be testable on staging


khaliqgant (approved):
nice

PR #3550, opened 2025-02-21
fix(proxy): added a flag that when passed to proxy configuration will not include the auth headers

hassan254-prog:
- Added a flag `includeAuthentication` that when passed as false to proxy configuration, will not include the auth headers


bodinsamuel:
Can we have some context? ☺️

hassan254-prog:
> Can we have some context? ☺️

This issue occurs in the context of AWS S3. When a user provides a pre-signed URL that uses `X-Amz-*` query parameters for authentication while building proxies, AWS returns the error: **"Only one authentication mechanism is allowed."** This happens because, even when overridden in `ProxyConfiguration`, an empty string is still passed in the `Authorization` header.  

To fix this, users can now set the flag **`includeAuthentication: false`**, ensuring the `Authorization` header is completely omitted.

hassan254-prog:
closing in favor [of](https://github.com/NangoHQ/nango/pull/3564)

PR #3551, opened 2025-02-21
fix(proxy): upload file as binary

hassan254-prog:
- Add more binary types to enable file upload as binary


bodinsamuel (commented):
Missing a bit the context here:
What's the current behavior, and how do I test this? ☺️ I tried uploading a file to google-drive it's still not working (or at least it gives the wrong content-type)

hassan254-prog:
> Missing a bit the context here: What's the current behavior, and how do I test this? ☺️ I tried uploading a file to google-drive it's still not working (or at least it gives the wrong content-type)

The change enables the `bodyParser.raw` middleware to dynamically determine whether a request's `Content-Type` should be treated as binary, rather than statically limiting it to `application/octet-stream`. Previously, any other binary content types (e.g., `image/png`, `video/mp4`) were not recognized as binary and were instead processed using Express’s default handling, which could lead to incorrect parsing or data corruption.

This change was done in regards to zendesk, you can try uploading a file to zendesk, the response shows that the file was processed correctly.  
Screenshot 2025-02-25 at 11 33 42
Screenshot 2025-02-25 at 11 34 09
Screenshot 2025-02-25 at 11 34 30


@bodinsamuel , could you try the following request for google drive, 
Screenshot 2025-02-25 at 17 06 44

Screenshot 2025-02-25 at 17 04 11
Screenshot 2025-02-25 at 17 04 23

bodinsamuel:
I see thanks for the details. The big thing I never noticed is that the header `content-type` is not forwarded as-is, we should probably in that case but that's for another PR

packages/server/lib/routes.ts
```diff
@@ -124,6 +124,23 @@ const adminAuth: RequestHandler[] = [
...
+    'model/',
+    'message/',
+    'chemical/',
+    'x-world/',
+    'application/octet-stream'
+];
+const isBinaryContentType = (contentType: string | undefined): boolean => {
+    if (!contentType) return false;
+    return BINARY_CONTENT_TYPES.some((type) => contentType.startsWith(type));
+};
```
bodinsamuel: Can you just move that to server/utils? (wherever you want)

bodinsamuel (approved):

packages/server/lib/routes.ts
hassan254-prog: updated with the latest commit 🚀 .

PR #3553, opened 2025-02-21
feat(integrations): add support for check

dannylwe:


packages/providers/providers.yaml
```diff
@@ -1724,6 +1724,24 @@ chattermill:
             suffix: .chattermill.com
             prefix: https://
 
+check:
```
bodinsamuel: Can we name that Check HQ to avoid a name clash later?

packages/webapp/public/images/template-logos/check.svg
bodinsamuel, line 1: logo must be 62x62, with white background and black writing (cf: acuity or cal.com)

dannylwe: @bodinsamuel the logo is an svg, is it not resized on the frontend depending on user screensize?

bodinsamuel: it's just the default size, it's easier to compare logos and have a uniform format

packages/webapp/public/images/template-logos/checkhq.svg
bodinsamuel, line 1: still need a white background, a black writing

dannylwe: Hi, sorry about that. This has been updated to have black text with a white background.

bodinsamuel (approved):

khaliqgant (approved):

PR #3557, opened 2025-02-21
fix(proxy): remove token from payload, rewrite headers building

bodinsamuel:
## Changes

- Remove `token` from payload
I'm not sure why it was used, probably a legacy thing. It wasn't typed correctly and was unnecessary since we carry the credentials in the `connection`. It removes a big chunk of code.

- Rewrite header building 
It was confusing with a default case that was almost always wrong, that got override by custom headers.


packages/shared/lib/services/proxy/utils.ts
```diff
@@ -160,17 +122,13 @@ export function buildProxyURL(config: ApplicationConstructedProxyConfiguration)
         const connectionConfig = connection.connection_config;
         const splitApiBase = apiBase.split(/\s*\|\|\s*/);
 
-        if (!connectionConfig) {
```
bodinsamuel, line 163: This could not happen

TBonnin (approved):
@hassan254-prog should also have a look I think

hassan254-prog:
> @hassan254-prog should also have a look I think

@TBonnin , thanks 🙏 .

hassan254-prog (approved):

Issue #3560, opened 2025-02-22
Supplying slack user scopes via user_scope is not working


a9a4k:
As a workaound I supplied user_scope via authorization_params

```
      const connectSession = await createConnectSession.mutateAsync({
        integrationId: selectedIntegration.id,
      });
      new Nango({
        connectSessionToken: connectSession.token,
      })
        .auth(
          selectedIntegration.id,
          selectedIntegration.provider.id === 'slack'
            ? {
                authorization_params: {
                  user_scope: // <---- supplying user scopes like this works
                    'channels:history,channels:read,channels:write,channels:write.topic,chat:write',
                },
                user_scope: [ // <---- supplying user scopes like this does not works
                  'channels:history',
                  'channels:read',
                  'channels:write',
                  'channels:write.topic',
                  'chat:write',
                ],
              }
            : {},
        )
        .then((_result) => {
```

almeynman:
Created a PR fixing the issue https://github.com/NangoHQ/nango/pull/3562

PR #3562, opened 2025-02-22
fix: handle slack user token

a9a4k:
The solution is to flatten `authed_user` into the main object.

The problem is described in the issue https://github.com/NangoHQ/nango/issues/3560


packages/shared/lib/services/connection.service.ts
```diff
@@ -1042,10 +1042,18 @@ class ConnectionService {
     // Parses and arbitrary object (e.g. a server response or a user provided auth object) into AuthCredentials.
     // Throws if values are missing/missing the input is malformed.
     public parseRawCredentials(rawCredentials: object, authMode: AuthModeType, template?: ProviderOAuth2 | ProviderTwoStep): AuthCredentials {
-        const rawCreds = rawCredentials as Record<string, any>;
+        let rawCreds = rawCredentials as Record<string, any>;
 
         switch (authMode) {
             case 'OAUTH2': {
+                const isSlackUserToken = !rawCreds['access_token'] && !!rawCreds['authed_user'];
```
TBonnin, line 1049: I think it would make more sense to rely on the template name to identify this Slack specific case rather than relying on the shape of the token response.

PR #3564, opened 2025-02-24
feat(sdk): add uncontrolledFetch

bodinsamuel:
## Changes

- Add `uncontrolledFetch` 
To be able to just fetch some resources that are unrelated to connection or provider. It's using `undici` because if we want to get rid of axios it's probably better to not add more here.
Naming is debatable :D 

- Remove some leftover 

## Tests

```ts

  const res = await nango.uncontrolledFetch({
    url: new URL("https://httpstatuses.maor.io/500"),
  });

  await nango.log("Res", { body: await res.json() });
}
```


packages/runner-sdk/lib/action.ts
```diff
@@ -342,13 +343,36 @@ export abstract class NangoActionBase {
         const parsedInput = zodSchema.safeParse(input);
         if (!parsedInput.success) {
             for (const error of parsedInput.error.errors) {
-                await this.log(`Invalid input provided to create a user: ${error.message} at path ${error.path.join('.')}`, { level: 'error' });
+                await this.log(`Invalid input provided: ${error.message} at path ${error.path.join('.')}`, { level: 'error' });
```
khaliqgant, line 346: Thanks

khaliqgant (approved):
works for me, thanks!

PR #3565, opened 2025-02-24
fix(metadata-stub): override metadata if coming from getConnection

khaliqgant:
Noticed this small issue with stubbing metadata
`npx nango dryrun foo -m '{"some": "metadata"}'`
```
const metadata = await nango.getMetadata()
// {"some": "metadata"}
```

```
const connection = await nango.getConnection()
const { metadata } = connection
// null
```

This fixes it so that stubbed metadata will also show from the getConnection call


bodinsamuel (approved):
Indeed

PR #3566, opened 2025-02-24
fix(proxy): refresh connection between retries 

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2745/retry-strategy-should-call-getconnection-before-each-try

> credentials are expired between calls

- ProxyRequest now accepts a `getConnection` param called on every iteration
The main idea behind this PR is to re-inject the connection after TTL expires.
I had to rewrite part of the code to be able to dynamically build axios config (url + headers).

- Type: Split ProxyConfig and Connection
Passing a connection to this type was not useful and was kind of weird with the new behavior. Now we pass a base proxyconfig that tells what we want to query and we recompile the axios config based on that.

## Tests

- Call the proxy with the retry header
```
curl --request GET \
  --url http://localhost:3003/proxy/500 \
  --header 'Authorization: REDACTED' \
  --header 'Base-Url-Override: https://httpstatuses.maor.io' \
  --header 'Connection-Id: 025d9a38-f31c-4a64-b034-57a425761a90' \
  --header 'Retries: 10'
```

- Same for local sync
```

  await nango.log("started");
  await nango.proxy({
    baseUrlOverride: "https://httpstatuses.maor.io",
    endpoint: "/500",
    method: "GET",
    retries: 10
  });
}
```

**Very long retry with a periodic refresh**
Screenshot 2025-02-25 at 12 13 08


packages/runner/lib/sdk/sdk.ts
```diff
@@ -38,10 +38,10 @@ export class NangoActionRunner extends NangoActionBase {
...
 
-        if (!this.activityLogId) throw new Error('Parameter activityLogId is required when not in dryRun');
-        if (!this.environmentId) throw new Error('Parameter environmentId is required when not in dryRun');
-        if (!this.nangoConnectionId) throw new Error('Parameter nangoConnectionId is required when not in dryRun');
-        if (!this.syncConfig) throw new Error('Parameter syncConfig is required when not in dryRun');
+        if (!this.activityLogId) throw new Error('Parameter activityLogId is required');
+        if (!this.environmentId) throw new Error('Parameter environmentId is required');
+        if (!this.nangoConnectionId) throw new Error('Parameter nangoConnectionId is required');
+        if (!this.syncConfig) throw new Error('Parameter syncConfig is required');
     }
```
bodinsamuel, line 45: left over from rewrite

packages/server/lib/controllers/proxy.controller.ts
```diff
@@ -364,48 +340,20 @@ class ProxyController {
...
-                axiosConfig.data = data;
-            }
-            const proxy = new ProxyRequest({
-                logger: async (msg) => {
-                    await logCtx.log(msg);
-                },
-                proxyConfig: config
-            });
-
-            const responseStream = (await proxy.request({ axiosConfig })).unwrap();
```
bodinsamuel, line 404: All the removed logic has been abstracted in `getAxiosConfiguration`

packages/shared/lib/services/proxy/request.ts
```diff
@@ -4,71 +4,105 @@ import type { Result, RetryAttemptArgument } from '@nangohq/utils';
...
-        }
+    getProxyConfig?: () => MaybePromise<ApplicationConstructedProxyConfiguration>;
 
-        if (this.config.data) {
-            options.data = this.config.data;
-        }
+    /**
+     * Called on error, gives the ability to control the retry and wait time
+     */
+    onError?: Props['onError'];
```
bodinsamuel, line 36: not used yet

packages/shared/lib/services/proxy/request.ts
```diff
@@ -3,72 +3,88 @@ import type { AxiosResponse, AxiosRequestConfig } from 'axios';
...
+import { getAxiosConfiguration, ProxyError } from './utils.js';
 import { getProxyRetryFromErr } from './retry.js';
 
+interface Props {
+    proxyConfig: ApplicationConstructedProxyConfiguration;
+    logger: (msg: MessageRowInsert) => MaybePromise<void>;
+    onError?: (args: { err: unknown; max: number; attempt: number }) => { retry: boolean; reason: string; wait?: number };
+    getConnection: () => MaybePromise<ConnectionForProxy>;
+}
+
```
bodinsamuel, line 16: Rewrote this file to accept a getConnection param that's called on every iteration

packages/server/lib/controllers/proxy.controller.ts
```diff
@@ -6,36 +6,32 @@ import { Readable, Transform, PassThrough } from 'stream';
 import type { UrlWithParsedQuery } from 'url';
```
bodinsamuel, line 6: Removed a lot of 'functional method' that did nothing

packages/shared/lib/services/proxy/request.ts
```diff
@@ -97,22 +113,20 @@ export class ProxyRequest {
...
-        error: unknown;
-        axiosConfig: AxiosRequestConfig;
-        retryAttempt: RetryAttemptArgument;
-        start: Date;
-    }): Promise<void> {
-        const valuesToFilter = Object.values(this.config.connection.credentials);
-        const redactedURL = redactURL({ url: axiosConfig.url!, valuesToFilter });
+    /**
+     * For testing purpose only
+     * @private
```
TBonnin, line 117: How does this @private work? The method is still declared as public below

packages/shared/lib/services/proxy/utils.ts
```diff
@@ -26,23 +33,51 @@ export class ProxyError extends Error {
     }
 }
 
+const methodDataAllowed = ['POST', 'PUT', 'PATCH', 'DELETE'];
```
TBonnin, line 36: used only once. Not sure it requires its own global const

TBonnin (approved):

packages/shared/lib/services/proxy/request.ts
bodinsamuel: it's only for documentation purposes, I need it to be public so I can mock it

packages/shared/lib/services/proxy/utils.ts
bodinsamuel: true, but allocating an array at each call seems like a waste to me

PR #3567, opened 2025-02-24
feat: modify /sync/ API endpoints to support variant

TBonnin:
Part of https://linear.app/nango/issue/NAN-2418/sync-and-model-variants

Refactoring following API endpoints to accept sync name and variant, while still being backward compatible:
- POST /sync/trigger
- POST /sync/pause
- POST /sync/start
- PUT /sync/update-connection-frequency
- GET /sync/status

I haven't modified GET /sync/provider to accept a sync variant since all variants shared the same provider. Happy to discuss.

### DIFF
--- a/packages/server/lib/controllers/sync.controller.integration.test.ts
+++ b/packages/server/lib/controllers/sync.controller.integration.test.ts
@@ -1,0 +1,692 @@
+/* eslint-disable @typescript-eslint/unbound-method */
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import type { Request, Response, NextFunction } from 'express';
+import syncController from './sync.controller.js';
+import type { DBTeam, DBEnvironment, ConnectSession, DBUser, EndUser } from '@nangohq/types';
+import type { RequestLocals } from '../utils/express.js';
+import { syncManager } from '@nangohq/shared';
+
+const mockRunSyncCommand = vi.spyOn(syncManager, 'runSyncCommand').mockResolvedValue({
+    success: true,
+    response: true,
+    error: null
+});
+
+const createMockRequest = ({ body = {}, query = {}, params = {}, headers = {} as Record<string, any> } = {}): Request => {
+    return {
+        body,
+        params,
+        query,
+        get: (header: string) => headers[header]
+    } as Request;
+};
+const createMockResponse = (): Response<any, Required<RequestLocals>> => {
+    return {
+        status: vi.fn().mockReturnThis(),
+        send: vi.fn().mockReturnThis(),
+        locals: {
+            authType: 'secretKey',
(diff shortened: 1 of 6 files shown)


packages/shared/lib/services/sync/sync.service.ts
```diff
@@ -249,75 +249,57 @@ export const getSyncs = async (
     });
 };
```
TBonnin, line 251: below consolidating `getSyncsByProviderConfigKey` and `getSyncsByConnectionId` by adding a filter argument instead of having a bunch of different function doing more or less the same thing

packages/server/lib/controllers/sync.controller.ts
```diff
@@ -761,4 +764,34 @@ class SyncController {
...
+    if (!Array.isArray(syncs)) {
+        return Err('syncs must be an array');
+    }
+
+    const syncIdentifiers = syncs.map((sync) => {
+        if (typeof sync === 'string') {
+            if (sync.includes('::')) {
+                const [name, variant] = sync.split('::');
+                return { syncName: name, syncVariant: variant };
+            }
```
TBonnin, line 780: agreed with @bastienbeurier to support `sync_name::sync_variant` format for the GET `/sync/status` endpoint where syncs are passed as a query param

packages/server/lib/controllers/sync.controller.ts
```diff
@@ -105,7 +105,7 @@ class SyncController {
 
     public async trigger(req: Request, res: Response<any, Required<RequestLocals>>, next: NextFunction) {
```
TBonnin, line 106: in an ideal world I would have refactored all the endpoints to use the `Endpoint` type and zod validation, unfortunately because of a lack of time I've decided not to do it now

packages/server/lib/controllers/sync.controller.integration.test.ts
```diff
@@ -0,0 +1,692 @@
...
+            authType: 'secretKey',
+            account: { id: 0 } as DBTeam,
+            environment: { id: 1 } as DBEnvironment,
+            user: { id: 0 } as DBUser,
+            connectSession: { id: 0 } as ConnectSession,
+            endUser: { id: 0 } as EndUser
+        }
+    } as unknown as Response<any, Required<RequestLocals>>;
+};
+
```
TBonnin, line 37: tests are not exhaustive, I have mostly added tests for parameters validation but it is better than zero tests. :/

packages/server/lib/controllers/sync.controller.ts
bodinsamuel: wish I had done them sooner

bodinsamuel (approved):
Can't really test variant but everything still works 👌🏻

PR #3568, opened 2025-02-24
feat(node-client): add support for sync variants

TBonnin:
Part of https://linear.app/nango/issue/NAN-2418/sync-and-model-variants

Modify `triggerSync`, `startSync`, `pauseSync`, `syncStatus` and `updateSyncConnectionFrequency` in node client to support sync variant parameter

### DIFF
--- a/packages/node-client/lib/index.ts
+++ b/packages/node-client/lib/index.ts
@@ -563,16 +563,21 @@ export class Nango {
     /**
      * Triggers an additional, one-off execution of specified sync(s) for a given connection or all applicable connections if no connection is specified
      * @param providerConfigKey - The key identifying the provider configuration on Nango
-     * @param syncs - An optional array of sync names to trigger. If empty, all applicable syncs will be triggered
+     * @param syncs - An optional array of sync names or sync names/variants to trigger. If empty, all applicable syncs will be triggered
      * @param connectionId - An optional ID of the connection for which to trigger the syncs. If not provided, syncs will be triggered for all applicable connections
      * @param fullResync - An optional flag indicating whether to perform a full resynchronization. Default is false
      * @returns A promise that resolves when the sync trigger request is sent
      */
-    public async triggerSync(providerConfigKey: string, syncs?: string[], connectionId?: string, fullResync?: boolean): Promise<void> {
+    public async triggerSync(
+        providerConfigKey: string,
+        syncs?: (string | { name: string; variant: string })[],
+        connectionId?: string,
+        fullResync?: boolean
+    ): Promise<void> {
         const url = `${this.serverUrl}/sync/trigger`;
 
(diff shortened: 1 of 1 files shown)


packages/node-client/lib/index.ts
```diff
@@ -709,14 +731,14 @@ export class Nango {
...
-            sync,
+        const body = {
+            ...(typeof sync === 'string' ? { sync_name: sync } : { sync_name: sync.name, sync_variant: sync.variant }),
             provider_config_key: providerConfigKey,
             connection_id: connectionId,
             frequency
         };
 
-        const response = await this.http.put(url, { headers: this.enrichHeaders(), params });
+        const response = await this.http.put(url, body, { headers: this.enrichHeaders() });
```
TBonnin, line 744: the `updateSyncConnectionFrequency` function never worked  I think

packages/node-client/lib/index.ts
```diff
@@ -680,23 +702,23 @@ export class Nango {
...
-     * @param sync - The name of the sync to update
+     * @param sync - The name of the sync to update (or an object with name and variant properties)
      * @param connectionId - The ID of the connection for which to update the sync frequency
      * @param frequency - The new frequency to set for the sync, or null to revert to the default frequency
      * @returns A promise that resolves with the response data after updating the sync frequency
      */
     public async updateSyncConnectionFrequency(
         providerConfigKey: string,
-        sync: string,
+        sync: string | { name: string; variant: string },
```
bodinsamuel, line 715: maybe a type for the param could have created

packages/node-client/lib/index.ts
```diff
@@ -660,14 +669,27 @@ export class Nango {
...
         }
 
         const url = `${this.serverUrl}/sync/status`;
 
+        const getSyncFullName = (sync: string | { name: string; variant: string }) => {
+            if (typeof sync === 'string') {
+                return sync;
+            }
+            if (sync.variant) {
+                return `${sync.name}::${sync.variant}`;
```
bodinsamuel, line 686: Should we prevent "::" to be specific in sync.name?

bodinsamuel (approved):

packages/node-client/lib/index.ts
TBonnin: I couldn't find a good name so I went for the solution of not naming it :) do you have any idea?

packages/node-client/lib/index.ts
TBonnin: good idea. I will do that in another PR at the CLI compilation level

packages/node-client/lib/index.ts
bodinsamuel: not really 😓

PR #3569, opened 2025-02-24
feat: add support for variant when fetching records

TBonnin:
part of https://linear.app/nango/issue/NAN-2418/sync-and-model-variants

One can now specify a `variant=VARIANT` query parameter when fetching records via the API. 
Also accepting variant parameter in node-client `listRecords`

Note: not yet useful/testable since it is not possible to save records for a variant model. You can test that current behavior (without variant) is not affected though

### DIFF
--- a/packages/server/lib/controllers/records/getRecords.ts
+++ b/packages/server/lib/controllers/records/getRecords.ts
@@ -1,7 +1,7 @@
 import { z } from 'zod';
 import { asyncWrapper } from '../../utils/asyncWrapper.js';
 import { zodErrorToHTTP } from '@nangohq/utils';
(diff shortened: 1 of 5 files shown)


packages/server/lib/controllers/records/getRecords.ts
```diff
@@ -65,7 +66,7 @@ export const getPublicRecords = asyncWrapper<GetPublicRecords>(async (req, res)
 
     const result = await records.getRecords({
         connectionId: connection.id,
-        model: query.model,
+        model: query.variant && query.variant !== 'base' ? `${query.model}::${query.variant}` : query.model,
```
TBonnin, line 69: using the naming convention `model::variant` as the full model name. It will be enforced by the `runner-sdk` in following PR

bodinsamuel (approved):

PR #3573, opened 2025-02-25
feat(integrations): add support for openai admin keys

viictoo:
Add  Openai Admin


bodinsamuel (approved):
Feels a bit weird to have a provider for admin just because the verification doesn't fit but I guess we don't have a better mechanism in place

khaliqgant:
> Feels a bit weird to have a provider for admin just because the verification doesn't fit but I guess we don't have a better mechanism in place

Agreed. Advised this route instead of removing verification entirely which is a cop out IMO. More customers are starting to complain about lack of verification for certain integrations FWIW

PR #3582, opened 2025-02-26
feat: add support for sync variant in runner-sdk/cli

TBonnin:
Making runner-sdk and CLI sync variant aware.

You can't test variant yet unless you modify your test database (sync and schedule payload) but you can check that current behavior is not impacted.
Next PR: expose variant in Nango webhook payload

### DIFF
--- a/packages/cli/lib/services/dryrun.service.ts
+++ b/packages/cli/lib/services/dryrun.service.ts
@@ -19,7 +19,7 @@ import * as zod from 'zod';
 import * as unzipper from 'unzipper';
 import { Buffer } from 'buffer';
 import { serializeError } from 'serialize-error';
-import { ActionError, InvalidActionInputSDKError, InvalidActionOutputSDKError, SDKError, validateData } from '@nangohq/runner-sdk';
+import { ActionError, InvalidActionInputSDKError, InvalidActionOutputSDKError, SDKError, validateData, BASE_VARIANT } from '@nangohq/runner-sdk';
 import { NangoActionCLI, NangoSyncCLI } from './sdk.js';
 
 interface RunArgs extends GlobalOptions {
@@ -32,6 +32,7 @@ interface RunArgs extends GlobalOptions {
     optionalEnvironment?: string;
     optionalProviderConfigKey?: string;
     saveResponses?: boolean;
+    variant?: string;
 }
 
 export class DryRunService {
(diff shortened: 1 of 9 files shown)


packages/cli/lib/services/dryrun.service.ts
```diff
@@ -243,10 +258,8 @@ export class DryRunService {
             }
 
             if (options.saveResponses) {
-                const responseDirectoryPrefix = process.env['NANGO_MOCKS_RESPONSE_DIRECTORY'] ?? '';
-                const directoryName = `${responseDirectoryPrefix}${providerConfigKey}`;
-                responseSaver.ensureDirectoryExists(`${directoryName}/mocks/${syncName}`);
-                const filePath = `${directoryName}/mocks/${syncName}/input.json`;
+                responseSaver.ensureDirectoryExists(saveResponsesSyncDir);
+                const filePath = `${saveResponsesSyncDir}/input.json`;
```
TBonnin, line 262: @khaliqgant where are those `responseSaver` files being used. Do I need to modify something else? Thank you

packages/cli/lib/services/sdk.ts
```diff
@@ -263,4 +275,9 @@ export class NangoSyncCLI extends NangoSyncBase {
 
         return objects;
     }
+
+    public override async setMergingStrategy(_merging: { strategy: 'ignore_if_modified_after' | 'override' }, _model: string) {
+        // Not applicable to CLI
+        return Promise.resolve();
+    }
```
TBonnin, line 258: unrelated but I forgot to implement `setMergingStrategy` in CLI sdk

packages/runner-sdk/lib/sync.ts
```diff
@@ -17,6 +21,17 @@ export abstract class NangoSyncBase extends NangoActionBase {
...
+            this.variant = config.syncVariant;
+        }
+    }
+
+    public modelFullName(model: string) {
+        if (this.variant === BASE_VARIANT) {
+            return model;
+        }
+        return `${model}::${this.variant}`;
     }
```
bodinsamuel, line 35: Does it need to be inside the class? we can put helpers outside now

bodinsamuel (approved):

packages/runner-sdk/lib/sync.ts
TBonnin: it needs to access `this.variant`

PR #3585, opened 2025-02-26
feat(integrations): fixes for mip: add separate on prem and cloud provider

viictoo:
- Add Separate MIP providers for On premises and Cloud logins

**Notes**
This integration is untested thus the token path in the response has not been validated
MIP uses token auth with header Authorization: TOKEN instead of the default Authorization: Bearer TOKEN in Nango

[EXT-513](https://linear.app/nango/issue/EXT-513/add-support-for-mip)


khaliqgant (changes requested):
Let's call it `mip-on-premise`

khaliqgant:
Noting that there are two existing `mip` integrations existing in prod now, so I think I can just rename them so that the `mip` => `mip-cloud` rename can work

PR #3588, opened 2025-02-26
feat(ui): show sync variant on connections page

TBonnin:
https://linear.app/nango/issue/NAN-2726/[design]-sync-variant-feature-quick-one 

Show sync variant in a grey bubble next to the sync name 
Only show variant bubble if there is more than the base variant 
Long variant names are truncated. Full name can be seen in a tooltip

Screenshot 2025-02-26 at 15 13 38

### DIFF
--- a/packages/webapp/src/components/ui/label/Tag.tsx
+++ b/packages/webapp/src/components/ui/label/Tag.tsx
@@ -2,7 +2,7 @@ import { cn } from '../../../utils/utils';
 import type { VariantProps } from 'class-variance-authority';
 import { cva } from 'class-variance-authority';
 
-const variants = cva('', {
+const variantStyles = cva('', {
     variants: {
         variant: {
             success: 'bg-success-400 border-success-400 text-success-400',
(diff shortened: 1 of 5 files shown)


packages/webapp/src/pages/Connection/components/SyncRow.tsx
```diff
@@ -114,7 +127,17 @@ export const SyncRow: React.FC<{ sync: SyncResponse; connection: ApiConnectionFu
     return (
         <Table.Row className="text-white">
             <Table.Cell bordered>
-                <div className="w-36 max-w-3xl truncate">{sync.name}</div>
+                <div className="w-36 max-w-3xl truncate">
+                    {sync.name}
+                    {showSyncVariant && ' '}
```
bodinsamuel: You can remove that use flex wrapper
```tsx
<div className="flex gap-2">{...}</div>
```

packages/webapp/src/components/ui/label/Tag.tsx
```diff
@@ -2,30 +2,64 @@ import { cn } from '../../../utils/utils';
...
+    } & VariantProps<typeof variantStyles>
+> = ({ children, variant, textCase, size }) => {
     return (
-        <div className={cn('inline-flex px-2 pt-[1px] border-[0.5px] bg-opacity-30 rounded', variants({ variant }))}>
-            <div className={cn('uppercase text-[11px] leading-[17px]')}>{children}</div>
+        <div className={cn('inline-flex border-[0.5px] bg-opacity-30 rounded', variantStyles({ variant, size }))}>
+            <div className={cn(sizeTextStyles({ size }), textCaseStyles({ textCase }))}>{children}</div>
         </div>
     );
 };
```
bodinsamuel, line 53: Feedback on this component:
- sizes are usually `xs, s, sm, md, lg` with `md` being the default 
- sizeTextStyles can't just be applied in the variant styles? line-height should be inherited no?
- I'm not sure how relevant the casing are to the component, seems like an implementation detail but you can pass a className so you don't need this `textCaseStyles` and each usage can define this

TBonnin: addressed point 1 and 2. 
As mentioned earlier the text casing was already handled by the component. I just parametrized it even though I agree it might not be the responsibility of the component

bodinsamuel (approved):

PR #3589, opened 2025-02-26
feat: expose variant in nango sync webhook

TBonnin:
Expose sync variant in nango sync webhook.

How to test:
- Set webhook url in env settings
- run sync
- check the sync variant now appear in the webhook payload


packages/webhooks/lib/sync.ts
```diff
@@ -63,7 +65,7 @@ export const sendSync = async ({
             integration: { id: providerConfig.id!, name: providerConfig.unique_key, provider: providerConfig.provider },
             connection: { id: connection.id, name: connection.connection_id },
             syncConfig: { id: syncConfig.id, name: syncConfig.sync_name },
-            meta: { scriptVersion: syncConfig.version }
```
bodinsamuel, line 66: Should we put that in syncConfig and index it so we can filter logs with it?
I'm sure it would be useful in sync run too

bodinsamuel (approved):

packages/webhooks/lib/sync.ts
TBonnin: I didn't put it in sync config because variant is related to a sync instance not a sync config (aka: sync definition). 
You are right it would be useful to index and filter by variant. I won't do it in the PR though.

PR #3590, opened 2025-02-26
feat: limit valid characters in sync names

TBonnin:
I wanted to prevent using "::" in sync name in order to avoid conflict with sync name format `name::variant` and after checking the prod db with the following sql query:
```
WITH characters AS (
  SELECT DISTINCT unnest(string_to_array(name, NULL)) AS character
  FROM nango._nango_syncs
  WHERE deleted = FALSE
)
SELECT character
FROM characters
ORDER BY character;
```
I confirmed that nothing but alphanumeric + underscore + dash characters are being used in sync names.
Therfore I've decided to restrict sync name in the nango.yaml to only be characters in `a-zA-Z0-9-_`

How to test:
- add invalid character(s) in your test nango.yaml
- run `nango compile`
- check compilation error

### DIFF
--- a/packages/nango-yaml/lib/errors.ts
+++ b/packages/nango-yaml/lib/errors.ts
@@ -10,6 +10,16 @@ export class ParserError {
     }
 }
 
+export class ParserErrorInvalidSyncName extends ParserError {
+    constructor(options: { name: string; path: string[] }) {
+        super({
+            code: 'invalid_sync_name',
+            message: `Sync "${options.name}" contains invalid characters`,
+            path: options.path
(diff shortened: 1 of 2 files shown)


bodinsamuel (commented):
Actually after checking this is already enforced in `packages/cli/lib/nango.yaml.schema.v2.json:50` but it's not enforced on zod `packages/server/lib/controllers/sync/deploy/validation.ts:73`

TBonnin:
closing this PR. I will open a new one to enforce the validation of the API

PR #3591, opened 2025-02-27
feat(integrations): add support for lemlist

hassan254-prog:
##  Describe the problem and your solution
Add support for Lemlist


docs-v2/integrations/all/lemlist.mdx
```diff
@@ -0,0 +1,51 @@
+---
+title: Lemlist
```
TBonnin: They don't seem to capitalize lemlist on their homepage. Maybe we should do the same in our docs

TBonnin (approved):

docs-v2/integrations/all/lemlist.mdx
hassan254-prog: I agree 💪 , updated with the latest commit 🙏 .

PR #3592, opened 2025-02-27
chore: update popular APIs

bastienbeurier:


bodinsamuel (approved):
You are removing this category?

bastienbeurier:
> You are removing this category?

I'm just reducing the number of popular APIs from 70 to 18. It was a bit crowded with 70 popular ones. So we're keeping the category.

PR #3595, opened 2025-02-27
fix(logs): make drawer selectable

bodinsamuel:
## Changes

fixes https://linear.app/nango/issue/NAN-2788/make-drawer-selectable

- make drawer selectable
- bonus: do not display arrow near message if there is no more content to show


TBonnin (approved):
nice

PR #3597, opened 2025-02-27
fix(http): add env to change agent

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2802/try-again-http-agent-with-env-option

- Add ways to change http agent dynamically so I can finetune in prod
The last time I broke production so I will try to be more gradual in my approach


packages/utils/lib/axios.ts
```diff
@@ -37,7 +37,8 @@ if (hasHttpsProxy) {
 
 const config: AxiosRequestConfig = {
     httpAgent: httpAgent,
-    httpsAgent: httpsAgent
+    httpsAgent: httpsAgent,
+    ...(process.env['AXIOS_HTTP_AGENT'] ? JSON.parse(process.env['AXIOS_HTTP_AGENT']) : {})
```
TBonnin: is it gonna be temporary? 
Also we should probably ensure the json parsing is successful before to assign it

TBonnin (commented):
small comment to make sure we don't blow production with a bad env var

packages/utils/lib/axios.ts
bodinsamuel: it's probably going to be temporary until I find the right configuration, but maybe it's going to be one configuration per service so not sure. 
you are right I'll add a try/catch to be extra cautious 👌🏻

TBonnin (approved):

bodinsamuel:
Tested in staging, but the volume is too low to measure something

TBonnin (approved):
can we add a comment to explain why we might want to use the HTTP_AGENT_CONFIG
I am sure in a month I am gonna have forgotten

PR #3598, opened 2025-02-27
fix(api): GET /scripts/config

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2801/clean-and-document-scriptsconfig

- New format for `GET /scripts/config`
- Fix associated OpenAPI spec and documentation

- Fix some times surrounding this endpoint
I initially wanted to split action/syncs/on-events, to be able to properly type the return and OpenAPI, but turns out it was too much broken to fix rn. 
I "just" synced more types, moved some in `@types` and merged some. I also clarified what's coming from the CLI and what's internal. It's still far from perfect or even usable tbh


packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -892,7 +892,7 @@ async function compileDeployInfo({
                 runs,
                 active: true,
                 model_schema: model_schema as unknown as SyncModelSchema[],
-                input: typeof flow.input === 'string' ? flow.input : flow.input ? flow.input.name : null,
+                input: typeof flow.input === 'string' ? flow.input : null,
```
TBonnin, line 782: is this not a breaking change

TBonnin (approved):

packages/shared/lib/services/sync/config/deploy.service.ts
bodinsamuel: it was tagged as string | LegacySyncModelSchema but it's not true since a long time

PR #3599, opened 2025-02-27
fix(telemetry): remove unused logs

bodinsamuel:
## Changes 

Contributes to https://linear.app/nango/issue/NAN-2805/telemetry-remove-logs

- Remove unused logs
They were either unused or have metrics alternative


packages/shared/lib/clients/orchestrator.ts
```diff
@@ -218,22 +203,6 @@ export class Orchestrator {
...
-                    input: JSON.stringify(input),
-                    environmentId: String(connection.environment_id),
-                    connectionId: connection.connection_id,
-                    providerConfigKey: connection.provider_config_key,
-                    actionName,
-                    level: 'error'
-                },
-                `actionName:${actionName}`
-            );
-
```
TBonnin, line 236: I think monitors and health dashboard widgets depends on this ACTION_SUCCESS/ACTION_FAILURE logs

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -658,59 +588,16 @@ export async function deployPreBuilt({
...
-                syncName: configs.map((config) => config.name).join(', '),
-                accountId: String(account.id),
-                integration: configs.map((config) => config.provider).join(', '),
-                preBuilt: 'true',
-                is_public: isPublic ? 'true' : 'false',
-                level: 'error'
-            },
-            `deploy_type:${isPublic ? 'public.' : 'private.'}template`
-        );
-
```
TBonnin, line 713: the customers deployments widget in the health dashboard depends on those SYNC_DEPLOY_SUCCESS/SYNC_DEPLOY_FAILURE logs I think

bodinsamuel: good catch, we can either use span or add a metric, wdyt?

packages/shared/lib/clients/orchestrator.ts
bodinsamuel: can be replaced by nango.orch.action.failure and nango.orch.action.success

bodinsamuel: however I'm seeing small discrepancy between the two, not sure why especially weird since they are logged at the same place. maybe it's acceptable 
https://us3.datadoghq.com/metric/explorer?fromUser=false&start=1740744552139&end=1740748152139&paused=false#N4Ig7glgJg5gpgFxALlAGwIYE8D2BXJVEADxQEYAaELcqyKBAC1pEbghkcLIF8qo4AMwgA7CAgg4RKUAiwAHOChASAtnADOcAE4RNIKtrgBHPJoQaUAbVBGN8qVoD6gnNtUZCKiOq279VKY6epbINiAiGOrKQdpYZAYgUJ4YThr42gDGSsgg6gi6mZaBZnHKGniqyJEiMDgAdG6ZjPUYmRJS9YIYEGh4RsBwIgBuyPLaOFB47ZIiPK0aTpn4IggAFACUIHygkdG5sVgATInJCKnp-dnKaDgwxSBaGFnMRIfKQ8MQEyLqq2MTKYzKQAAnO92QbQ6IhcPT6Rm2VFEAmI+jCIAAVCAALpUGATPDyJwAIxoYVxIGWqnkBByoAwMHxcBgnlmymWeFWiMeCDcDJyrBwSB4FNc7jwmFC4TFqglGBipXiiNAMrlCp0x22FI0cjQdJA8gweoQCAFUDuS0lGggmUSaFEcCcckU5Rw9qgdodTnoTGUIjcHjQ3KgEHsmCwToUAuJz0sIp4fEe8ntCAAwlJhDAUCIJWgeEA

bodinsamuel: actually that might just be log times that differ

bodinsamuel: changed that widget to use metrics

packages/shared/lib/services/sync/config/deploy.service.ts
TBonnin: Let's add a span

TBonnin (approved):

packages/shared/lib/services/sync/config/deploy.service.ts
bodinsamuel: we can use the express span I guess

PR #3600, opened 2025-02-27
feat: add API endpoints and functions to create/delete sync variants

TBonnin:
Create API endpoints and node client functions to create and delete a sync variant

How to test:
```
// creation
curl --request POST --header 'Authorization: SECRET' \                                            
  --url http://localhost:3003/sync/SYNC_NAME/variant/NEW_VARIANT \
  --data '{
  "provider_config_key": "PROVIDER_CONFIG_KEY",
  "connection_id": "CONNECTION_ID"
}'

// deletion
curl --request DELETE --header 'Authorization: YOUR_SECRET' \                                            
  --url http://localhost:3003/sync/SYNC_NAME/variant/VARIANT \
  --data '{
  "provider_config_key": "PROVIDER_CONFIG_KEY",
  "connection_id": "CONNECTION_ID"
}'
```

### DIFF
--- a/packages/database/lib/migrations/20250227180909_sync_variant_unique.cjs
+++ b/packages/database/lib/migrations/20250227180909_sync_variant_unique.cjs
@@ -1,0 +1,20 @@
+exports.config = { transaction: false };
+
+const TABLE = '_nango_syncs';
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.schema.raw(`ALTER TABLE "${TABLE}" DROP CONSTRAINT IF EXISTS "_nango_syncs_name_nango_connection_id_deleted_at_unique";`);
+    await knex.schema.raw(`DROP INDEX CONCURRENTLY IF EXISTS "_nango_syncs_variant_name_nango_connection_id_deleted_at_unique"`);
+    await knex.schema.raw(`
+        CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS "_nango_syncs_variant_name_nango_connection"
+        ON "${TABLE}" ("variant", "name", "nango_connection_id")
+        WHERE deleted_at IS NULL;
+    `);
+};
+
+exports.down = function () {
+    // do nothing
+};
--- a/packages/node-client/lib/index.ts
+++ b/packages/node-client/lib/index.ts
@@ -762,6 +762,7 @@ export class Nango {
         const response = await this.http.post(url, body, { headers: this.enrichHeaders() });
         return response.data;
     }
+
     /**
      *
      * Delete an existing sync variant
@@ -778,7 +779,7 @@ export class Nango {
             },
             headers: this.enrichHeaders()
         });
-        return response.data as DeleteSyncVariant['Success'];
+        return response.data;
     }
 
     /**
--- a/packages/node-client/lib/index.ts
+++ b/packages/node-client/lib/index.ts
@@ -28,7 +28,9 @@ import type {
     SignatureCredentials,
     PostPublicConnectSessionsReconnect,
     GetPublicConnection,
-    NangoRecord
+    NangoRecord,
+    PostSyncVariant,
+    DeleteSyncVariant
 } from '@nangohq/types';
 import type {
     CreateConnectionOAuth1,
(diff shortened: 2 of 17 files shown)


packages/server/lib/controllers/sync.controller.ts
```diff
@@ -616,47 +615,6 @@ class SyncController {
         }
     }
 
-    public async deleteSync(req: Request, res: Response<any, Required<RequestLocals>>, next: NextFunction) {
```
TBonnin, line 619: as discussed earlier today, deleting the old delete sync endpoint

packages/shared/lib/services/sync/manager.service.ts
```diff
@@ -527,6 +527,7 @@ export class SyncManagerService {
             finishedAt: latestJob?.updated_at,
             nextScheduledSyncAt: schedule.nextDueDate,
             name: sync.name,
+            variant: sync.variant,
```
TBonnin, line 530: unrelated to the other commits but I  realized I forgot to expose the variant in `getSyncStatus` response

packages/shared/lib/services/sync/sync.service.ts
```diff
@@ -326,27 +326,6 @@ export const verifyOwnership = async (nangoConnectionId: number, environment_id:
...
-            [`${TABLE}.deleted`]: false
-        });
-
-    if (result.length === 0) {
-        return false;
-    }
-
-    return true;
-};
-
```
TBonnin, line 349: only used in legacy sync delete logic deleted above

packages/webapp/src/pages/Connection/components/SyncRow.tsx
```diff
@@ -132,7 +132,7 @@ export const SyncRow: React.FC<{ sync: SyncResponse; connection: ApiConnectionFu
                         {sync.name}
                         {showSyncVariant && (
                             <SimpleTooltip tooltipContent={sync.variant}>
-                                <Tag variant="gray1" textCase="lowercase" size="sm">
+                                <Tag variant="gray1" textCase="normal" size="sm">
```
TBonnin, line 135: we allow `a-zA-Z` so let's display the actual case the customer entered

packages/database/lib/migrations/20250227180909_sync_variant_unique.cjs
```diff
@@ -0,0 +1,20 @@
+exports.config = { transaction: false };
+
+const TABLE = '_nango_syncs';
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.schema.raw(`ALTER TABLE "${TABLE}" DROP CONSTRAINT IF EXISTS "_nango_syncs_name_nango_connection_id_deleted_at_unique";`);
+    await knex.schema.raw(`DROP INDEX CONCURRENTLY IF EXISTS "_nango_syncs_variant_name_nango_connection_id_deleted_at_unique"`);
```
TBonnin, line 10: The original unique constraint and the index I added when introducing the variant column are actually not working as expected. By default pg treats NULLs as distinct values. Since pg15 there is a `NULLS NOT DISTINCT` option that I could have used below but a partial index also does the job

packages/runner-sdk/models.d.ts
```diff
@@ -355,6 +355,7 @@ export declare class NangoAction {
     private logAPICall;
 }
 export declare class NangoSync extends NangoAction {
+    variant: string;
```
bodinsamuel, line 358: you added it to models.ts but not the actual runner code?

bodinsamuel, line 358: you added it to models.ts but not the actual runner code?

packages/node-client/lib/index.ts
```diff
@@ -746,6 +748,39 @@ export class Nango {
...
+    public async createSyncVariant(props: PostSyncVariant['Body'] & PostSyncVariant['Params']): Promise<PostSyncVariant['Success']> {
+        const url = `${this.serverUrl}/sync/${props.name}/variant/${props.variant}`;
+        const body = {
+            provider_config_key: props.provider_config_key,
+            connection_id: props.connection_id
+        };
+        const response = await this.http.post(url, body, { headers: this.enrichHeaders() });
+        return response.data;
+    }
+    /**
```
bodinsamuel, line 766:
```suggestion

    /**
```

packages/server/lib/controllers/sync/postSyncVariant.ts
```diff
@@ -0,0 +1,132 @@
...
+    if (!syncs) {
+        res.status(500).send({ error: { code: 'failed_sync_variant_creation' } });
+        return;
+    }
+
+    const maxSyncsPerConnection = 100; // arbitrary limit to prevent abuse/mistakes
+    if (syncs.length > maxSyncsPerConnection) {
+        res.status(400).send({ error: { code: 'resource_capped', message: `Maximum number of syncs per connection (${maxSyncsPerConnection}) reached` } });
+        return;
+    }
```
bodinsamuel, line 74: Is this limit something we changed recently?

packages/server/lib/controllers/sync/deleteSyncVariant.ts
```diff
@@ -0,0 +1,71 @@
...
+        return;
+    }
+
+    try {
+        await syncManager.softDeleteSync(sync.id, environment.id, orchestrator);
+    } catch {
+        res.status(500).send({ error: { code: 'failed_sync_variant_deletion' } });
+        return;
+    }
+    res.status(204).send();
```
bodinsamuel: I know we don't agree on this, but can we always send json? 
it's annoying on the HTTP client (customer side) to not receive JSON and conditionally parse JSON, and it makes hard to distinguish missing content vs no content 
```js
const res = nango.deleteSync();
if (!res) {
  // is it an error or is it expected?
}
```

bodinsamuel (approved):
Didn't had time to test it but looks good 👌🏻

packages/runner-sdk/models.d.ts
TBonnin: I had already added it to the runner-sdk. I think I somehow didn't commit the models.d.ts when I did so  https://github.com/NangoHQ/nango/blob/127f5155eb8914d6f78c6e3c8a12b311c0128242/packages/runner-sdk/lib/sync.ts#L9

packages/server/lib/controllers/sync/postSyncVariant.ts
TBonnin: it is something we are introducing here. Until now there were no limit but the only way to create a sync is to write yaml or click in the dashboard to enable a pre-built one, making it quite hard to create a lot of them.
Now it can be created programmatically we are setting a limit to prevent abuse and more importantly mistakes. I picked the limit arbitrarily but the goal is to prevent creating 100s of syncs by mistake

packages/server/lib/controllers/sync/deleteSyncVariant.ts
TBonnin: sure. let's send `{success: true}`

packages/server/lib/controllers/sync/postSyncVariant.ts
bodinsamuel: I see, can we make that an env var?

PR #3601, opened 2025-02-28
feat(cli): dynamically load extraneous dependency

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2807/investigate-soap-in-cli

- Dynamically load extraneous dependency
We don't want the CLI to end up installing the whole world. Until we figure out bundling, it's going to be this solution, so at least make it less impactful on install. 

- Added new command to run the project as headless
Got tired of running the UI when not needed, it's too heavy when debugging/using backend

## Tests

Testing is a bit annoying.
- `npm uninstall -w @nangohq/nango-runner soap` (desinstall in the project so it's not mistakenly loaded)
```ts
# soap.ts
import soap from "soap";

  console.log("soap", { soap });
  await soap.createClientAsync(
    `https://community.workday.com/sites/default/files/file-hosting/productionapi/top/vtop/top.wsdl`,
    {},
  );

  await nango.log("top");
}
```
- `npm run dwh`
- `node nango/packages/cli/dist/index.js dryrun soap 025d9a38-f31c-4a64-b034-57a425761a90`

When not installed locally it will yield this
Screenshot 2025-02-28 at 11 56 35


TBonnin:
>     * Added new command to run the project as headless
>       Got tired of running the UI when not needed, it's too heavy when debugging/using backend
🫶

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -541,7 +569,12 @@ export class DryRunService {
                                 return unzipper;
                             case 'soap':
                             case 'botbuilder':
-                                throw new Error(`Module '${moduleName}' not available in dry run. Please test the integration using the Nango dashboard`);
+                                if (!preloaded[moduleName]) {
+                                    throw new Error(
+                                        `Module '${moduleName}' needs to be installed on your side in a package.json. You can also test using the Nango dashboard`
```
TBonnin, line 574: > You can also test using the Nango dashboard
Should we be a bit more explicit and say "Deploy to Nango dev environment to test" or something similar

TBonnin (approved):

PR #3605, opened 2025-02-28
fix(db): allow using read-only replica

bodinsamuel:
## Changes

Contributes to https://linear.app/nango/issue/NAN-2816/use-read-replica-for-big-read-only-queries

- Allow to use read-only replica


bodinsamuel:
"Tested" on staging. We don't have read-replica but I set it to the main db

packages/database/lib/index.ts
```diff
@@ -9,10 +9,25 @@ const directory = path.join(projectRoot, 'packages/database/lib/migrations');
...
+        const readOnlyURL = process.env['NANGO_DB_READ_URL'];
+        if (readOnlyURL) {
+            (dbConfig.connection as knex.Knex.PgConnectionConfig).connectionString = readOnlyURL;
+            this.readOnly = knex({
+                ...dbConfig,
+                connection: {
+                    ...(dbConfig.connection as object),
+                    connectionString: readOnlyURL
+                }
+            });
```
TBonnin: can we avoid mutating the dbConfig?
```
  const readOnlyConfig = {
    ...dbConfig,
    connection: typeof dbConfig.connection === 'object' 
      ? { 
          ...dbConfig.connection,
          connectionString: readOnlyURL 
        }
      : readOnlyURL
  };
this.readOnly = knex(readOnlyConfig)
```

TBonnin (approved):

packages/database/lib/index.ts
bodinsamuel: I messed up anyway 🤦🏻

PR #3606, opened 2025-02-28
fix(fleet): do not fetch next page if no more nodes

TBonnin:
We were doing an unnecessary query to fetch nodes next page even when there were no more


bodinsamuel (approved):
indeed

PR #3613, opened 2025-03-03
fix(db): clean up models, use read-only for more calls

bodinsamuel:
## Changes

- Config controller
  - Only load connections if asked
  - `GET /api/v1/integrations/:providerConfigKey/connections` (getConnections) was unused
- Sync controller
  - `GET /sync/provider` (getSyncProvider) was unused
- Connection service
  - `getMetadata` was unused
  - Use count instead of listing every connections when checking for cap
- Slack
  -  Use environment when possible
- Jobs
  - Clean up unused
- Sync service
  - getLastSyncDate use read-only 
  - Clean up unused


TBonnin (approved):
🪓

PR #3615, opened 2025-03-03
fix: temporary disabling actions slack and bigquery reporting

TBonnin:


TBonnin:
moving the refactor of the script result handler to another branch since it is bigger than what I was anticipating. I will wait for your review @bodinsamuel 
in the meantime. I am gonna merge this branch since this is what is currently in prod

PR #3619, opened 2025-03-03
feat(cron): delete old jobs data 

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2827/cron-clean-up-jobs

- Cron to delete old jobs data
We agreed not having the latest job if it hasn't run for 90 days is an acceptable drawback for the moment


packages/shared/lib/services/sync/job.service.ts
```diff
@@ -193,3 +193,15 @@ export async function hardDeleteJobs({ syncId, limit }: { syncId: string; limit:
             sub.select('id').from('_nango_sync_jobs').where({ sync_id: syncId }).limit(limit);
         });
 }
+
+export async function deleteJobsByDate({ expiresAfterDays, limit }: { expiresAfterDays: number; limit: number }): Promise<number> {
```
TBonnin: `deleteJobsOlderThan` is a better name I think, conveying how the date is being used

packages/jobs/lib/crons/deleteOldJobsData.ts
```diff
@@ -0,0 +1,57 @@
...
+import { getLogger, metrics } from '@nangohq/utils';
+import tracer from 'dd-trace';
+import { setTimeout } from 'node:timers/promises';
+import { envs } from '../env.js';
+
+const logger = getLogger('Jobs.deleteOldsJobs');
+
+const limit = envs.CRON_DELETE_OLD_JOBS_LIMIT;
+const cronMinutes = 10;
+const expiresAfterDays = 90;
```
TBonnin: those 2 const could be env vars as well

TBonnin (approved):

PR #3623, opened 2025-03-04
feat: add sentry again

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2817/investigate-using-sentry

- Add sentry again
This time only for own errors, a bit intentionally + uncaughtError


bodinsamuel:
Currently running staging

packages/jobs/lib/app.ts
```diff
@@ -19,14 +19,18 @@ const logger = getLogger('Jobs');
...
     // not closing on purpose
 });
 
 process.on('uncaughtException', (err) => {
     logger.error('Received uncaughtException...', err);
+    report(err);
     // not closing on purpose
 });
 
+initSentry({ dsn: envs.SENTRY_DSN, applicationName: envs.NANGO_DB_APPLICATION_NAME, hash: envs.GIT_HASH });
```
TBonnin, line 32: what if SENTRY_DSN is not set and/or self-hosted setup?

TBonnin (approved):
should sentry be disabled for self-hosted setup?
Rest looks ok

packages/jobs/lib/app.ts
bodinsamuel: it's disabled if dsn is empty, errors.ts:26 👍🏻

PR #3624, opened 2025-03-04
fix(provider): fix mip

hassan254-prog:
Fix Mip token request `credentials` interpolation.


khaliqgant:
@hassan254-prog as we chatted about can you make a task to make `credentials` plural when used in `TWO_STEP`?

khaliqgant (approved):

PR #3625, opened 2025-03-05
fix(kvstore): move feature flags to dedicated package 

bodinsamuel:
## Changes

- Move feature flags to dedicated package
- Reuse Redis store
- Destroy the global instance

> Locking and various Rate Limit are still using an other Redis instance


packages/kvstore/lib/RedisStore.ts
```diff
@@ -1,24 +1,15 @@
...
-
-    public async connect(): Promise<void> {
-        return this.client.connect().then(() => {});
+    constructor(client: RedisClientType) {
+        this.client = client;
     }
 
     public async destroy(): Promise<void> {
-        await this.client.disconnect();
+        // DO nothing
```
TBonnin: why doing nothing?

packages/kvstore/lib/InMemoryStore.ts
```diff
@@ -55,4 +61,14 @@ export class InMemoryKVStore implements KVStore {
...
         return false;
     }
+
+    private clearExpired() {
+        for (const [key, value] of this.store) {
+            if (this.isExpired(value)) {
+                this.store.delete(key);
+            }
+        }
+        console.log(this.store.size);
```
TBonnin: console.log

packages/kvstore/lib/FeatureFlags.ts
```diff
@@ -1,37 +1,15 @@
...
-    }): Promise<boolean> {
-        if (!this.kvstore) {
-            return fallback;
-        }
-
+    async isSet(key: string, { distinctId = 'global', fallback = false }: { distinctId?: string; fallback?: boolean } = {}): Promise<boolean> {
         try {
-            const exists = await this.kvstore.exists(`flag:${key}:${distinctId}`);
-            return isExcludingFlag ? !exists : exists;
+            return await this.kvstore.exists(`flag:${key}:${distinctId}`);
```
TBonnin, line 12: why isExcludingFlag logic being removed?

packages/kvstore/lib/index.ts
```diff
@@ -1,17 +1,73 @@
...
+export async function getKVStore(): Promise<KVStore> {
+    if (kvstorePromise) {
+        return await kvstorePromise;
+    }
+
+    kvstorePromise = createKVStore();
+    return await kvstorePromise;
+}
+
+let ffClient: Promise<FeatureFlags> | undefined;
```
TBonnin: Not sure you need to store the promise. ffClient is always set or undefined and never in a pending state

packages/server/lib/utils/utils.ts
```diff
@@ -24,6 +25,8 @@ const BINARY_CONTENT_TYPES = [
     'application/octet-stream'
 ];
 
+export const ffClient = await getFeatureFlagsClient();
```
TBonnin: I would have kept the `featureFlags` name but up to you (or even a simple `ff`)

TBonnin (approved):
I made a few comments.

Unrelated to your PR but thinking about it I am not sure having featureFlags being part of kvstore is the best architecture. Ideally I would have a featureFlags package that is using KvStore one.

packages/kvstore/lib/RedisStore.ts
bodinsamuel: I'll add a comment 👌🏻

packages/kvstore/lib/InMemoryStore.ts
bodinsamuel: oops forgot to push this clean up

packages/kvstore/lib/index.ts
bodinsamuel: the class itself is never pending but waiting for the redis client yes

packages/kvstore/lib/FeatureFlags.ts
bodinsamuel: I found it really hard to understand vs just writing a flag name that makes sense. 
e.g: "webhooks fallback false with excluding", means disable this feature when it's set, whereas just setting a feature flag `disable-webhook` is straight forward. Let me know

TBonnin: the idea is to always think in term of a feature being enabled or not.
Without it you have something like `ff.isEnabled('disable-webhook')` which is also not straightforward to read (ie: is the flag to _disable_ webhook _enabled_?. Instead of `ff.isEnabled('webhook') aka is the webhook feature _enabled_) 
I don't feel strongly about it so feel free to remove it

packages/kvstore/lib/index.ts
TBonnin: yes but there is an `await getKVStore`. We never get to the assignement of ffClient unless the kvstore is ready. Proof we always assign a `Promise.resolve`

bodinsamuel: arf yes you are right

packages/kvstore/lib/FeatureFlags.ts
bodinsamuel: I have renamed it to isSet to avoid this confusion `ff.isSet('disable-webhook')` is "clear" = no double negative

packages/kvstore/lib/index.ts
bodinsamuel: fixed

PR #3629, opened 2025-03-05
fix: fail action when output is >10mb

TBonnin:
when output action is >10mb (max size supported by jobs and orchestrator) the request from runner to jobs to report the script result fails silently causing the task not to be updated and eventually being expired after 15mins
With this commit we ensure the task is errored asap. I took advantage of this change to refactor the runner logic calling jobs and persist API to make it more type-safe
As part of this change we are now also retrying the call to jobs PUT /tasks/ID endpoint

How to test:
- create an action that return a big result (ex: `return new Array(5_000_000).fill(0)`)
- run locally and check that runner register, heartbeat, idle and report memory warning correctly. You might want to tweak the different thresholds/intervals
```diff
diff --git a/packages/jobs/lib/runner/local.ts b/packages/jobs/lib/runner/local.ts
index 6043b061f..8c0823c47 100644
--- a/packages/jobs/lib/runner/local.ts
+++ b/packages/jobs/lib/runner/local.ts
@@ -33,7 +33,7 @@ export const localNodeProvider: NodeProvider = {
                     RUNNER_URL: `http://localhost:${port}`,
-                    IDLE_MAX_DURATION_MS: '0',
+                    IDLE_MAX_DURATION_MS: '10000',
                     PROVIDERS_URL: getProvidersUrl(),
diff --git a/packages/runner/lib/env.ts b/packages/runner/lib/env.ts
index a3a5e8048..17f9fd071 100644
--- a/packages/runner/lib/env.ts
+++ b/packages/runner/lib/env.ts
@@ -3,7 +3,7 @@ import { getJobsUrl, getPersistAPIUrl } from '@nangohq/shared';
 export const envs = parseEnvs(ENVS.required({ RUNNER_NODE_ID: true }));
-export const heartbeatIntervalMs = 30_000;
+export const heartbeatIntervalMs = 1_000;
 export const jobsServiceUrl = getJobsUrl();
diff --git a/packages/utils/lib/environment/parse.ts b/packages/utils/lib/environment/parse.ts
index 6d08c96f7..ca882eb3e 100644
--- a/packages/utils/lib/environment/parse.ts
+++ b/packages/utils/lib/environment/parse.ts
@@ -60,7 +60,7 @@ export const ENVS = z.object({
     RUNNER_URL: z.string().url().optional(),
-    RUNNER_MEMORY_WARNING_THRESHOLD: z.coerce.number().optional().default(85),
+    RUNNER_MEMORY_WARNING_THRESHOLD: z.coerce.number().optional().default(0),
```


packages/types/lib/persist/api.ts
```diff
@@ -1,5 +1,23 @@
+import type { ApiError, Endpoint } from '../api.js';
+import type { MessageRowInsert } from '../logs/messages.js';
 import type { MergingStrategy, NangoRecord } from '../record/api.js';
 
+interface LogBody {
+    activityLogId: string;
+    log: MessageRowInsert;
+}
+
+export type PostLog = Endpoint<{
```
bodinsamuel, line 10: Maybe we should prefix specific API to avoid conflict later, e.g: PostPersistLog or PersistPostLog, or else. wdyt?

packages/types/lib/jobs/api.ts
```diff
@@ -0,0 +1,53 @@
...
+        nodeId: number;
+    };
+    Body: {
+        url: string;
+    };
+    Error: ApiError<'register_failed'>;
+    Success: { status: 'ok' };
+}>;
+
+export type PostIdle = Endpoint<{
```
bodinsamuel, line 45: Same comment

packages/runner/lib/monitor.ts
```diff
@@ -79,17 +74,18 @@ export class RunnerMonitor {
...
                     activityLogId: activityLogId,
-                    level: 'warn',
-                    msg: `Memory usage is high: ${memoryUsagePercentage.toFixed(2)}% of the total available memory.`
-                })
+                    log: {
+                        type: 'log',
+                        level: 'warn',
+                        message: `Memory usage is high: ${memoryUsagePercentage.toFixed(2)}% of the total available memory.`,
+                        createdAt: new Date().toISOString()
+                    }
```
bodinsamuel, line 87: it was broken no?

packages/runner/lib/clients/http.ts
```diff
@@ -0,0 +1,16 @@
+import { logger } from '../logger.js';
+import type { RequestInfo } from 'undici/types/index.js';
+
+export async function httpFetch(resource: RequestInfo | URL, init?: RequestInit): Promise<Response> {
+    const res = await fetch(resource, init);
+    if (res.status >= 400) {
+        let content;
+        if (res.headers.get('content-type')?.includes('application/json')) {
```
bodinsamuel, line 8: 😬

packages/runner/lib/clients/jobs.ts
```diff
@@ -0,0 +1,96 @@
...
+        const res = await retryWithBackoff(async () => {
+            const resp = await httpFetch(`${this.baseUrl}/runners/${nodeId}/register`, {
+                method: 'POST',
+                headers: {
+                    'Content-Type': 'application/json'
+                },
+                body: JSON.stringify({ url })
+            });
+            if (!resp.ok) {
+                return Err(`PostRegister_failed`);
```
bodinsamuel, line 75: nit for consistency
```suggestion
                return Err(`postRegister_failed`);
```

bodinsamuel (approved):
Works well, much cleaner 🙏🏻

packages/runner/lib/clients/http.ts
```diff
@@ -0,0 +1,16 @@
+import { logger } from '../logger.js';
+import type { RequestInfo } from 'undici/types/index.js';
+
+export async function httpFetch(resource: RequestInfo | URL, init?: RequestInit): Promise<Response> {
+    const res = await fetch(resource, init);
```
bodinsamuel, line 5: can you use either Undici or axios? browser fetch is much slower and not compatible with everything. Didn't notice it's also used at other places

packages/types/lib/persist/api.ts
TBonnin: It is namespaced via the directory. I think it is ok to have the caller rename the import to avoid conflict if needed.

packages/runner/lib/clients/http.ts
TBonnin: this is node fetch which is powered by undici (hence the import above)

packages/runner/lib/monitor.ts
TBonnin: it was using the legacy log format which you just removed in https://github.com/NangoHQ/nango/commit/3dced1531cad838f13c7ddd14b3995290d392062#diff-871ef30263b732ac8701c666b7d42fcf2268407a048ae682633c42536a6a06fdL11-L19
so yes it was broken but not for a very long time

bodinsamuel: thanks 🙏🏻

packages/runner/lib/clients/http.ts
bodinsamuel: I know but I had read somewhere it wasn't exactly 1:1 because it needs to be compatible with browser fetch, cant' find this information. 
Also maybe we could think about using undici pipeline or stream to forward the output

PR #3631, opened 2025-03-06
fix(kvstore): move Locking

bodinsamuel:
## Changes 

Fixes https://linear.app/nango/issue/NAN-2837/redis-consolidate-one-pool

- Move Locking to KVStore

> I'm not going to move RateLimiter to kvstore since they are more specific and they might benefit having their own pool rn
> We'll see if we need that later


TBonnin (approved):
👌

PR #3632, opened 2025-03-06
fix(agent): better default for axios

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-941/backend-have-a-unified-fetcher-with-http-agent

- Better default for axios
Tested on all services in prod


packages/utils/lib/axios.ts
```diff
@@ -7,7 +7,12 @@ import type { Agent as HttpAgent } from 'node:http';
 import type { Agent as HttpsAgent } from 'node:https';
 import https from 'node:https';
 
-let agentOptions: http.AgentOptions = {};
+let agentOptions: http.AgentOptions = {
+    keepAlive: true,
+    timeout: 30000,
+    scheduling: 'fifo'
+};
```
TBonnin, line 14: I am curious: what are default defaults 😄 ? and why those value are better?

TBonnin (approved):

packages/utils/lib/axios.ts
bodinsamuel: keepAlive is not true, timeout is 60s, scheduling is lifo
https://nodejs.org/api/http.html#new-agentoptions

It's better first because it's not crashing prod like last time and it's working. I'm sure we could change some timeout or max but it's hard to finetune like this (especially with our current prod)

PR #3634, opened 2025-03-06
fix(es): do not wait some calls, change sharding

bodinsamuel:
## Changes

Contributes to https://linear.app/nango/issue/NAN-2843/do-not-wait-for-regular-log
Fixes https://linear.app/nango/issue/NAN-2844/reduce-number-of-shards

- Change sharding strategy and expose an env var
- Do not wait some calls
It's not necessary but we just want to test how it behaves rn


packages/logs/lib/es/schema.ts
```diff
@@ -145,7 +145,8 @@ export const indexMessages: estypes.IndicesCreateRequest = {
             'sort.field': ['createdAt', 'id'],
             'sort.order': ['desc', 'desc']
         },
-        number_of_shards: 10 // Made up number until we figured out a better strategy
+        // They are recommending 1 shard per 20gb-40gb
+        number_of_shards: envs.NANGO_LOGS_ES_SHARD_PER_DAY
```
TBonnin, line 149: what's the current value?

TBonnin (approved):

packages/logs/lib/es/schema.ts
bodinsamuel: 10 per index, so way too much. It will be 1 after this (only for new index)

PR #3635, opened 2025-03-06
fix(telemetry): remove logs

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2805/telemetry-remove-logs

- Remove logs
Sync failure and success had metrics already, token refresh was also covered, added metrics for general auth success


packages/runner/lib/exec.ts
```diff
@@ -47,7 +46,7 @@ export async function exec(
 
     const filename = `${nangoProps.syncConfig.sync_name}-${nangoProps.providerConfigKey}.js`;
 
-    return await tracer.trace<Promise<RunnerOutput>>(SpanTypes.RUNNER_EXEC, async (span) => {
+    return await tracer.trace<Promise<RunnerOutput>>('nango.runner.exec', async (span) => {
```
TBonnin, line 49: why this change?

TBonnin (approved):

packages/server/lib/controllers/auth/postApiKey.ts
```diff
@@ -230,6 +232,8 @@ export const postPublicApiKeyAuthorization = asyncWrapper<PostPublicApiKeyAuthor
             metadata: { providerConfigKey, connectionId }
         });
 
+        metrics.increment(metrics.Types.AUTH_FAILURE, 1, { auth_mode: 'API_KEY' });
+
```
TBonnin, line 236: Could we have the metrics in `AccessMiddleware` instead of having to repeat it for every single auth type

bodinsamuel: access middleware is for global authentication, but I see your point

packages/runner/lib/exec.ts
bodinsamuel: After removing log, there was only 2 spans defined in telemetry file, it seems we are not using this pattern so to be consistent I removed it

packages/server/lib/controllers/auth/postApiKey.ts
bodinsamuel: actually it's annoying because we don't store the auth_mode, going to keep it like this until better

PR #3642, opened 2025-03-07
chore(providers): add api verification endpoint (3 of 6)

viictoo:
This PR adds a verification endpoint to check the validity of the API proxy endpoint for various providers
All the verification endpoints introduced in this PR are tested.
Separate PR for untested endpoints in this project
#3643 

[EXT-283](https://linear.app/nango/issue/EXT-283/add-an-api-verification-endpoint-3-of-6)


bodinsamuel (commented):
Same comment about anthropic

viictoo:
> Same comment about anthropic

updated ✅

bodinsamuel (approved):

packages/providers/providers.yaml
```diff
@@ -3170,9 +3188,12 @@ fireflies:
         - productivity
     auth_mode: API_KEY
     proxy:
-        base_url: https://api.fireflies.ai
+        base_url: https://api.fireflies.ai/graphql
```
khaliqgant: @viictoo we should take care when editing the `base_url` as this could break out customers proxy calls and sync/action scripts. Is this change necessary?

viictoo: No. This change is not necessary. removed

khaliqgant (approved):

PR #3645, opened 2025-03-10
feat(integrations): add anthropic admin

viictoo:
add support for  anthropic admin provider


packages/providers/providers.yaml
```diff
@@ -494,6 +494,9 @@ anthropic:
             content-type: application/json
         retry:
             after: 'retry-after'
+        verification:
+            method: GET
+            endpoint: /v1/models
```
bodinsamuel: Contains other stuff from your other PR

bodinsamuel (approved):

packages/providers/providers.yaml
viictoo: all cleaned up now.

bodinsamuel (approved):

PR #3647, opened 2025-03-10
fix(otlp): manual span

bodinsamuel:
## Changes

Contributes to https://linear.app/nango/issue/NAN-2803/investigate-otlp-export

- Manually declare OTLP span
We need to dissociate span creation from logCtx, because getting the operation entirely from ES is not a viable solution. This fix only changes Span to be declared manually but does not fix the main issue. Just wanted to be sure it was doable and limit the scope of this PR.

- Fix nangoProps type
- Fix OTLP export broken by Sentry (skipOtel)


packages/types/lib/runner/sdk.ts
```diff
@@ -7,23 +7,26 @@ export interface NangoProps {
...
+    nangoConnectionId: number;
     syncJobId?: number | undefined;
     track_deletes?: boolean;
     attributes?: object | undefined;
     abortSignal?: AbortSignal;
     syncConfig: DBSyncConfig;
     runnerFlags: RunnerFlags;
+    /**
+     * @deprecated not used
+     */
```
TBonnin, line 29: is only debug deprecated?

packages/utils/lib/errors.ts
```diff
@@ -21,8 +21,7 @@ export function initSentry({ dsn, hash, applicationName }: { dsn: string | undef
     Sentry.init({
         dsn: dsn || '',
         sampleRate: 1,
-        tracesSampleRate: 0,
-        openTelemetryInstrumentations: [],
+        skipOpenTelemetrySetup: true,
```
TBonnin: can we add a comment to explain why we set skipOpenTelemetrySetup to true?

packages/jobs/lib/execution/sync.ts
```diff
@@ -92,6 +93,7 @@ export async function startSync(task: TaskSync, startScriptFn = startScript): Pr
                 meta: { scriptVersion: syncConfig.version }
             }
         );
+        logCtx.attachSpan(new OtlpSpan(logCtx.operation, startedAt));
```
TBonnin, line 96: I am not a huge fan of this pseudo independent `attachSpan` function that we need to sprinkle all over the codebase. 
Since it always depends on a `logContextGetter.create` why not making it a part of it. 
can `logContextGetter.create` take a `otlpSpan?: { startedAt?: Date } params`. When it is passed the `create` function can set `span = new OtlpSpan(...)`

bodinsamuel: `Since it always depends`, that's what I'm trying to partially remove actually, the problem with OTLP right now is that it assumes logCtx have the operation available at all times, which is not the case for syncs (and in the future in async actions).

When I need to recreate the span in `handleSuccess`, it's currently going through `logContextGetter.create` but it's only working because we are fetching the operation from ES. I want to get rid of that to optimize the code.

Right now it's not fully fixed as my PR desc since I'm still using logCtx.operation, I still need to figure out a good solution

packages/types/lib/runner/sdk.ts
bodinsamuel: yes (not sure I understand the question)

TBonnin: just confirming the comment only applies to the line below  and not all lines below.

packages/jobs/lib/execution/sync.ts
TBonnin: where do you intend to get the span `startedAt` if not rom the the log operation in ES?

TBonnin (approved):

PR #3651, opened 2025-03-10
fix(jobs): add index without deleted

bodinsamuel:
## Changes

Contributes to https://linear.app/nango/issue/NAN-2884/remove-soft-delete-from-jobs

- Add index without `deleted`
Soft delete is not used on this table, but we need a 3 three steps deployment to avoid issues. Next PR will remove usage


TBonnin (commented):
Given the size of the sync jobs table, I would split this PR, first create the new index and then do the drop in next PR. Since we are using CONCURRENTLY we don't know if new index will be ready when the old one is dropped

bodinsamuel:
Sorry it wasn't clear, I'm only deleting unused indexes unrelated to the one added before. The current code is still using the previous index with `where deleted`

TBonnin (approved):

PR #3652, opened 2025-03-10
fix(jobs): remove usage of deleted

bodinsamuel:
## Changes

Contributes to https://linear.app/nango/issue/NAN-2884/remove-soft-delete-from-jobs
Follow up of #3651 

- Remove usage of soft-delete


packages/shared/lib/models/Sync.ts
```diff
@@ -46,7 +46,7 @@ export interface Action extends TimestampsAndDeleted {
     name: string;
 }
 
-export interface Job extends TimestampsAndDeleted {
+export interface Job extends Timestamps {
```
bodinsamuel, line 49: Next PR I'll move that to types

packages/shared/lib/services/sync/job.service.ts
```diff
@@ -148,38 +148,35 @@ export const updateSyncJobResult = async (id: number, result: SyncResultByModel,
...
+        .select('*')
         .where({
             sync_id,
-            deleted: false,
             status: SyncStatus.RUNNING
         })
         .orderBy('created_at', 'desc')
-        .limit(1);
+        .limit(1)
+        .first();
```
TBonnin, line 159: aren't `limit(1)` and `first()` the same thing?

packages/shared/lib/services/sync/job.service.ts
```diff
@@ -76,7 +76,7 @@ export const getLatestSyncJob = async (sync_id: string): Promise<SyncJob | null>
 };
 
 export const getSyncJobByRunId = async (run_id: string): Promise<SyncJob | null> => {
-    const result = await schema().from<SyncJob>(SYNC_JOB_TABLE).where({ run_id, deleted: false }).first();
+    const result = await db.knex.from<SyncJob>(SYNC_JOB_TABLE).select('*').where({ run_id }).first();
```
TBonnin, line 79: have we already removed all the `deleted` sync_jobs. if not this query is potentiall gonna return a different row

bodinsamuel: we are not using this feature so it's not possible that it returns a different thing

packages/shared/lib/services/sync/job.service.ts
bodinsamuel: not exactly, limit is limiting the query, first is picking the first item in the array

packages/shared/lib/services/sync/job.service.ts
TBonnin: you mean it is never set to true? because `getSyncJobByRunId` is called

bodinsamuel: yes `deleted` is never set to true

packages/shared/lib/services/sync/job.service.ts
TBonnin: this is not `arrray.first()` but `queryBuilder.first()`
https://github.com/knex/knex/blob/9bd12999907436c2ef51f786df09a9a7e8931cca/lib/query/querybuilder.js#L1151

TBonnin (approved):

packages/shared/lib/services/sync/job.service.ts
bodinsamuel: yes it's still pick the first item in the array, but you are right it's also limiting
Screenshot 2025-03-12 at 09 43 57
Screenshot 2025-03-12 at 09 44 10

PR #3655, opened 2025-03-10
fix(es): prepare fields for duration and context

bodinsamuel:
## Changes

Contributes to https://linear.app/nango/issue/NAN-2778/http-logs-add-duration
Contributes to https://linear.app/nango/issue/NAN-2751/tag-http-calls-with-call-context

- Prepare fields for duration and context, endUser
Duration will help display the duration in the UI and compute graph
Context will help filtering HTTP calls depending on the context

### DIFF
--- a/packages/logs/lib/es/schema.ts
+++ b/packages/logs/lib/es/schema.ts
@@ -64,6 +64,7 @@ const props: Record<keyof MessageRow | keyof OperationRow, estypes.MappingProper
     type: { type: 'keyword' },
     level: { type: 'keyword' },
     state: { type: 'keyword' },
+    context: { type: 'keyword' },
 
     source: { type: 'keyword' },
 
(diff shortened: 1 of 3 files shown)


packages/logs/lib/es/schema.ts
```diff
@@ -96,7 +97,8 @@ const props: Record<keyof MessageRow | keyof OperationRow, estypes.MappingProper
     updatedAt: { type: 'date' },
     startedAt: { type: 'date' },
     expiresAt: { type: 'date' },
-    endedAt: { type: 'date' }
+    endedAt: { type: 'date' },
+    durationMs: { type: 'integer' }
```
TBonnin, line 103: cannot duration be calculated by doing `endedAt - startedAt`?

bodinsamuel: it's possible yes, but it's super slow at scale

bodinsamuel:
Can you re-review? I added endUserId and moved context in the right interface (🤦🏻 )

TBonnin (approved):

PR #3660, opened 2025-03-11
feat(cli-deploy): deploy scripts for a specific api

viictoo:
This PR introduces feature to optionally deploy scripts for a specific integration:
Usage:
`npx nango deploy dev --integration foobar`
or
`npx nango deploy dev -i foobar`

 - will only compile and deploy scripts for the integration `foobar`
 
 `npx nango deploy --integration foo --sync bars dev`
- will deploy the sync `bars` from the integration `foo` to the enviroment `dev`
- similarly a specific action from a specific integration can be deployed if both flags are provided ie
`npx nango deploy --integration foo --action bar dev` 

[ext-543](https://linear.app/nango/issue/EXT-543/deploy-scripts-for-a-specific-api)


packages/cli/lib/index.ts
```diff
@@ -176,6 +176,7 @@ program
     .option('-v, --version [version]', 'Optional: Set a version of this deployment to tag this integration with. Can be used for rollbacks.')
     .option('-s, --sync [syncName]', 'Optional deploy only this sync name.')
     .option('-a, --action [actionName]', 'Optional deploy only this action name.')
+    .option('--integration-id [integrationId]', 'Optional: Deploy all scripts related to a specific integration.')
```
TBonnin: should the option be `--integration` (with `-i` as a short version) what do you think @viictoo @khaliqgant @bastienbeurier @bodinsamuel

packages/cli/lib/services/compile.service.ts
```diff
@@ -275,57 +275,70 @@ export function listFilesToCompile({
...
+                printDebug(`Found ${rootFiles.length} TypeScript files in the root directory`);
+            }
+        }
+        const filteredIntegrations = providerConfigKey
+            ? parsed.integrations.filter((integration) => integration.providerConfigKey === providerConfigKey)
+            : parsed.integrations;
+
+        if (providerConfigKey && debug) {
+            if (filteredIntegrations.length > 0) {
+                printDebug(`Filtering compilation to only include files for provider config key: ${providerConfigKey}`);
```
TBonnin:
```suggestion
                printDebug(`Compiling file(s) for integration: ${providerConfigKey}`);
```

packages/cli/lib/services/compile.service.ts
```diff
@@ -275,57 +275,70 @@ export function listFilesToCompile({
...
+            } else {
+                printDebug(`Warning: No integration found with provider config key: ${providerConfigKey}`);
```
TBonnin:
```suggestion
                printDebug(`Warning: No '${providerConfigKey}' integration found`);
```
Let's not talk about provider config key since the option the user provide is actually integration. That might be confusing

packages/cli/lib/services/compile.service.ts
```diff
@@ -275,57 +275,70 @@ export function listFilesToCompile({
...
+            }
+        }
 
-            files = [...files, ...syncFiles, ...actionFiles, ...postFiles, ...onEventsFiles];
+        for (const integration of filteredIntegrations) {
+            const pcKey = integration.providerConfigKey;
```
TBonnin: not sure we need a variable. It is used only once

packages/cli/lib/services/deploy.service.ts
```diff
@@ -116,6 +115,9 @@ class DeployService {
         if (debug) {
             printDebug(`NANGO_HOSTPORT is set to ${process.env['NANGO_HOSTPORT']}.`);
             printDebug(`Environment is set to ${environment}`);
+            if (integrationId) {
+                printDebug(`Deploying only for provider config key: ${integrationId}`);
```
TBonnin:
```suggestion
                printDebug(`Deploying integration '${integrationId}' only`);
```

packages/cli/lib/services/deploy.service.ts
```diff
@@ -125,11 +127,19 @@ class DeployService {
         }
 
         const parser = parsing.value;
-        const singleDeployMode = Boolean(optionalSyncName || optionalActionName);
+
+        // Check if the integrationId exists in the parsed config
+        if (integrationId && !parser.parsed!.integrations.some((integration) => integration.providerConfigKey === integrationId)) {
```
TBonnin: can we replace the negation by `.every((integration) => integration.providerConfigKey !== integrationId))

packages/cli/lib/services/deploy.service.ts
```diff
@@ -125,11 +127,19 @@ class DeployService {
...
+            console.log(chalk.red(`Integration with ID "${integrationId}" not found in configuration.`));
+            process.exit(1);
+        }
+
+        const singleDeployMode = Boolean(optionalSyncName || optionalActionName || integrationId);
```
TBonnin: Should we make the single sync/action and integration option disjoint? Does it make sense to specify a integration if a single sync/action is specified?

packages/cli/lib/services/deploy.service.ts
```diff
@@ -316,6 +353,9 @@ class DeployService {
         if (debug) {
             printDebug(`NANGO_HOSTPORT is set to ${process.env['NANGO_HOSTPORT']}.`);
             printDebug(`Environment is set to ${environment}`);
+            if (integrationId) {
+                printDebug(`Deploying only for provider: ${integrationId}`);
```
TBonnin:
```suggestion
                printDebug(`Deploying for integration: '${integrationId}' only`);
```

packages/cli/lib/services/deploy.service.ts
```diff
@@ -325,14 +365,26 @@ class DeployService {
         }
 
         const parser = parsing.value;
+
+        if (integrationId && !parser.parsed!.integrations.some((integration) => integration.providerConfigKey === integrationId)) {
```
TBonnin: same comment about negation and some as above

packages/cli/lib/services/compile.service.ts
viictoo: Updated ✅

packages/cli/lib/index.ts
bastienbeurier: Indeed, I think `--integration` & `-i` are better.

packages/cli/lib/services/deploy.service.ts
viictoo: Either sync action or integration can be specified and both and are handled separately. 
Combined them here so as to reuse the singleDeployMode flag to prevent the system from treating other syncs/actions as "deleted" when you're only deploying a specific integration

packages/cli/lib/index.ts
viictoo: updated. ✅

packages/cli/lib/services/compile.service.ts
viictoo: updated✅

packages/cli/lib/services/compile.service.ts
viictoo: removed✅

packages/cli/lib/services/deploy.service.ts
viictoo: updated✅

packages/cli/lib/services/deploy.service.ts
viictoo: updated✅

packages/cli/lib/services/deploy.service.ts
viictoo: updated✅

packages/cli/lib/services/deploy.service.ts
viictoo: ✅

packages/cli/lib/services/deploy.service.ts
TBonnin: got it. Is the CLI going to error out if -i and -s/a are provided together?

packages/cli/lib/index.ts
TBonnin: @viictoo you added `-i` but kept `--integration-id` instead of `--integration`

packages/cli/lib/services/deploy.service.ts
viictoo: no error in the current setup. if both -i and -s are provided, current behaviour is that --integration-id takes precedence.
```if (singleDeployMode && !integrationIdMode) {```

packages/cli/lib/index.ts
viictoo: right! let me make that update as well

packages/cli/lib/services/deploy.service.ts
khaliqgant: I think the -s or -a flag should take precedence over the —integration flag

TBonnin: Make sense to have the most specific wins. Personally I would error to avoid any ambiguity but up to you

viictoo: added in check to error out when both --integration and -s/-a are provided to eradicate the ambiguity
[ if (integrationId && (optionalSyncName || optionalActionName)) Please use either one, but not both.](https://github.com/NangoHQ/nango/blob/58f245e90d16d7666fe4c3b30fde83262bcb29a1/packages/cli/lib/services/deploy.service.ts#L100)

khaliqgant:
Edge case question:

1. Deploy some syncs under a specific integration by passing in the `--integration foo` flag
2. Comment out one of those syncs in the nango.yaml
3. Deploy that same integration using the `--integration foo` flag.

What happens? Does it try and remove the one sync that was commented out? Or is no diff still not performed?

packages/cli/lib/services/deploy.service.ts
bodinsamuel: arg, actually that would be nice if we supported both at the same time when you have syncs with the same name across integration

khaliqgant: > arg, actually that would be nice if we supported both at the same time when you have syncs with the same name across integration

Yup, good point, agreed!

bodinsamuel (approved):
💪🏻

viictoo:
> Edge case question:
> 
>     1. Deploy some syncs under a specific integration by passing in the `--integration foo` flag
> 
>     2. Comment out one of those syncs in the nango.yaml
> 
>     3. Deploy that same integration using the `--integration foo` flag.
> 
> 
> What happens? Does it try and remove the one sync that was commented out? Or is no diff still not performed?

In this case, the already deployed sync is not removed when it is omitted from the second deployment.

packages/cli/lib/services/deploy.service.ts
viictoo: update:
when both `--integration` and sync/action `-s`/`-a` are specified, the specific action/sync will be deployed only if exists in the specified integration

TBonnin (approved):

PR #3661, opened 2025-03-11
feat(connect): store operation_id in DB

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2889/store-operation-id-in-db

We want to log events before we actually save a connection's credentials. To do that, we will create a new public endpoint that receives events from the Connect UI. We need to either pre-create an operation or create the operation at the first event.

- Prepare connect_sessions table to receive an operation_id


TBonnin (approved):
can you explain what the full flow would be? thank you

PR #3664, opened 2025-03-11
feat(integration): add missive support

miles-kt-inkeep:
# Describe your changes
Adding configuration for Missive API Key

# Checklist before requesting a review (skip if just adding/editing APIs & templates)
- I added tests, otherwise the reason is: no sync/actions yet
- I added observability, otherwise the reason is: no sync/actions yet
- I added analytics, otherwise the reason is: no sync/actions yet


bodinsamuel (approved):
Thanks 🙏🏻

PR #3666, opened 2025-03-11
feat(verification): make verification check more robust 

hassan254-prog:
The current credentials verification implementation has several limitations:

- Some providers return a `200 OK` status even when credentials are incorrect, including an error message in the response body (e.g., LastPass).
- Some providers require credentials to be passed in a different format (e.g., in the request body) compared to what the `proxyBuild` method currently supports, making it impossible to test credentials using `proxy.verification` (e.g., LastPass, Expensify).
- Some providers offer multiple key types, requiring separate provider configurations to accommodate verification for both.

This PR addresses these issues by introducing the ability to write custom verification scripts that run during credentials validation. These changes will allow:
- More control over how requests are constructed, making verification more robust.
- Support for multiple verification endpoints per provider. If any endpoint returns a `200 OK` response the function will return early and the credentials are considered valid.
- A validation check to ensure that we don't define both verification script and a `proxy.verification` method at the same time.


scripts/validation/providers/schema.json
```diff
@@ -459,6 +462,9 @@
                         "type": "string"
                     }
                 },
+                "verification_script": {
```
khaliqgant: Could potentially be unclear what this is referring to. Verification of what?

bodinsamuel: agree, could still be inside proxy or maybe credentials_verification_script

scripts/validation/providers/schema.json
```diff
@@ -258,14 +258,17 @@
...
                                     "type": "string",
                                     "enum": ["GET", "POST"]
                                 },
-                                "endpoint": {
-                                    "type": "string"
+                                "endpoints": {
+                                    "type": "array",
+                                    "items": {
+                                        "type": "string"
+                                    }
```
bodinsamuel, line 271: This change should be retro-compatible, plus it seems you are not using two endpoints ever?

packages/server/lib/hooks/connection/verificatiion-script.ts
```diff
@@ -0,0 +1,137 @@
+import type { Span } from 'dd-trace';
```
bodinsamuel: name of the file `verification-script.ts`

packages/server/lib/hooks/connection/verificatiion-script.ts
```diff
@@ -0,0 +1,137 @@
...
+    };
+
+    const logs: MessageRowInsert[] = [
+        {
+            type: 'log',
+            level: 'info',
+            message: 'Running automatic credentials verification via verification script',
+            createdAt: new Date().toISOString()
+        }
+    ];
```
bodinsamuel: this array is not used

packages/server/lib/hooks/connection/verificatiion-script.ts
```diff
@@ -0,0 +1,137 @@
...
+                return (await proxy.request()).unwrap();
+            }
+        };
+
+        await handler(internalNango);
+        span.setTag('nango.success', true);
+    } catch (err) {
+        const errorString = stringifyError(err);
+        span.setTag('nango.error', errorString);
+        throw err;
```
bodinsamuel:
```suggestion
    } catch (err) {
        span.setTag('error', err);
        throw err;
```

packages/server/lib/hooks/connection/verificatiion-script.ts
```diff
@@ -0,0 +1,137 @@
...
+                    getConnection: () => {
+                        return connection;
+                    }
+                });
+                return (await proxy.request()).unwrap();
+            }
+        };
+
+        await handler(internalNango);
+        span.setTag('nango.success', true);
```
bodinsamuel: we are not using that
```suggestion
```

packages/server/lib/hooks/connection/verificatiion-script.ts
```diff
@@ -0,0 +1,137 @@
...
+    };
+
+    const activeSpan = tracer.scope().active();
+
+    const span = tracer.startSpan('nango.server.hooks.verificationScript', {
+        childOf: activeSpan as Span,
+        tags: {
+            'nango.provider': providerName,
+            'nango.providerConfigKey': providerConfigKey,
+            'nango.connectionId': connectionId
```
bodinsamuel: The prefix is legacy you can remove
```suggestion
            'provider': providerName,
            'providerConfigKey': providerConfigKey,
            'connectionId': connectionId
```

packages/server/lib/hooks/connection/verificatiion-script.ts
```diff
@@ -0,0 +1,137 @@
...
+            return;
+        }
+
+        const handler = handlers[provider.verification_script];
+        if (!handler) {
+            return;
+        }
+
+        const internalNango: InternalNango = {
+            getCredentials: async () =>
```
bodinsamuel: Why make it a promise if it's a sync method?

packages/server/lib/refreshConnections.ts
```diff
@@ -5,7 +5,7 @@ import { logContextGetter } from '@nangohq/logs';
 import {
     connectionRefreshFailed as connectionRefreshFailedHook,
     connectionRefreshSuccess as connectionRefreshSuccessHook,
-    connectionTest as connectionTestHook
+    verifyConnectionBeforeCreation as connectionTestHook
```
bodinsamuel: the name is wrong then or it should be credentialsTest

packages/server/lib/hooks/connection/verification/lastpass-verification.ts
```diff
@@ -0,0 +1,41 @@
+import type { InternalNango as Nango } from '../verificatiion-script.js';
+import type { UserResponse } from '../../response-types/lastpass.js';
+
+export default async function execute(nango: Nango) {
+    const { credentials, providerConfigKey } = await nango.getCredentials();
+
+    if (!isValidCredentials(credentials)) {
+        throw new Error('Invalid credentials format');
+    }
```
bodinsamuel: I'm not entirely sure this can happen, when you call this method it's already been through zod and credentials formatting

bodinsamuel (commented):
Nice work, I haven't tested yet but few comments here and there 💪🏻 🚀

packages/server/lib/hooks/hooks.ts
```diff
@@ -72,6 +73,65 @@ export const connectionCreationStartCapCheck = async ({
     return false;
 };
 
+export async function verifyConnectionBeforeCreation({
```
TBonnin: The name refers to how the function is being used. Can we have a name that describe what it does instead?

packages/server/lib/hooks/connection/verificatiion-script.ts
TBonnin: I would even say `credentials-verification-script.ts`

packages/server/lib/hooks/connection/verificatiion-script.ts
```diff
@@ -0,0 +1,137 @@
...
+                const proxy = new ProxyRequest({
+                    logger: (msg) => {
+                        logs.push(msg);
+                    },
+                    proxyConfig,
+                    getConnection: () => {
+                        return connection;
+                    }
+                });
+                return (await proxy.request()).unwrap();
```
TBonnin: are we sure we want to unwrap? Unless we have a very good reason to unwra, its usage should be limited to tests

packages/server/lib/hooks/hooks.ts
```diff
@@ -212,12 +272,12 @@ export async function connectionTest({
 }): Promise<Result<{ logs: MessageRowInsert[]; tested: boolean }, NangoError>> {
     const providerVerification = provider?.proxy?.verification;
 
-    if (!providerVerification) {
+    if (!providerVerification || !Array.isArray(providerVerification.endpoints) || providerVerification.endpoints.length === 0) {
```
TBonnin:
```suggestion
    if (!providerVerification?.endpoints?.length) {
```

packages/server/lib/hooks/hooks.ts
```diff
@@ -247,58 +307,66 @@ export async function connectionTest({
...
+                type: 'log',
+                level: 'error',
+                message: `Error testing endpoint: ${endpoint},  ${stringifyError(err)}`,
+                createdAt: new Date().toISOString()
+            });
+        }
     }
+
+    const error = new NangoError('connection_test_failed', { logs });
+    span.setTag('nango.error', error);
```
TBonnin:
```suggestion
    span.setTag('error', error);
```

packages/server/lib/hooks/hooks.ts
```diff
@@ -247,58 +307,66 @@ export async function connectionTest({
...
 
-    const logs: MessageRowInsert[] = [
-        { type: 'log', level: 'info', message: `Running automatic credentials verification`, createdAt: new Date().toISOString() }
-    ];
-    try {
-        const proxyConfig = getProxyConfiguration({ externalConfig: configBody, internalConfig }).unwrap();
-        const proxy = new ProxyRequest({
-            logger: (msg) => {
-                logs.push(msg);
+    for (const endpoint of endpoints) {
```
TBonnin, line 318: I am curious: which provider has more than one endpoint to test for credentials

scripts/validation/providers/schema.json
hassan254-prog: thanks updated this with the latest commit to  `credentials_verification_script` 👍 .

packages/server/lib/hooks/connection/verificatiion-script.ts
hassan254-prog: makes it more cleaner, thanks 👍 .

packages/server/lib/hooks/connection/verificatiion-script.ts
hassan254-prog: removed this with the latest commit.

packages/server/lib/hooks/connection/verificatiion-script.ts
hassan254-prog: removed some probably left over ones with this [commit](https://github.com/NangoHQ/nango/pull/3666/commits/a2261a01df0d1ee9ad98930173d8928c6633891c) as well

packages/server/lib/hooks/connection/verificatiion-script.ts
hassan254-prog: aah yes there is no need updated this.

packages/server/lib/hooks/connection/verification/lastpass-verification.ts
hassan254-prog: It makes sense to use type assertion here to avoid TypeScript complaints, rather than a runtime type guard, since the credentials have already been validated through Zod and formatting.

packages/server/lib/hooks/hooks.ts
hassan254-prog: updated 👍

packages/server/lib/hooks/hooks.ts
hassan254-prog: I have opted for `testConnectionCredentials` instead, please let me know if this would be a correct naming convention for the function.

packages/server/lib/hooks/hooks.ts
hassan254-prog: @NangoHQ/engineers, we've been facing issues and limitations with some providers when adding verification. Certain providers offer different key types, such as basic and admin keys, similar to scoped keys, which can cause us to skip validation in some cases.  
To ensure proper validation, it would be beneficial to test against multiple endpoints and consider the credentials valid if at least one returns a success message.  
However, I'd like your input on whether we should proceed with this approach. If we implement it, should we move the multiple-endpoint testing logic to the `credentials-verification-script`, or would it be better to define an array of endpoints within the provider configurations?
It may not be ideal since we would need to make numerous API calls before creating the connection, but at least we would have verified the credentials.

This is for bitdefender, we could pick atleast one per API type;
[
Screenshot 2025-03-12 at 18 00 47
](url)

scripts/validation/providers/schema.json
hassan254-prog: I have left a [comment](https://github.com/NangoHQ/nango/pull/3666#discussion_r1991712579) that should also help answer this.

hassan254-prog:
> Nice work, I haven't tested yet but few comments here and there 💪🏻 🚀

Thanks 🙏 , there are test connections available within the shared dev account. You can retrieve the credentials to test both scripts included in this PR 😊

packages/server/lib/hooks/connection/verificatiion-script.ts
hassan254-prog: would it be better to handle errors explicitly rather than relying on `.unwrap()`?
```
const result = await proxy.request();
if (result.isErr()) {
    throw result.error;
}
return result.value; 
```

hassan254-prog: would it be better to handle errors explicitly rather than relying on `.unwrap()`?
```
const response = await proxy.request();
if (response.isErr()) {
    throw response.error;
}
return response.value; 
```

TBonnin: that's basically what unwrap does so if we are 100% happy to throw if it is an Err, using unwrap is ok

packages/server/lib/hooks/hooks.ts
TBonnin: I am fine with the array of endpoints in the config. I imagine 99% of the providers would have a single entry. right?

packages/server/lib/hooks/connection/verificatiion-script.ts
hassan254-prog: My thought was to make the error handling more explicit for readability, but I agree that the functionality remains the same 👍 .

packages/server/lib/hooks/hooks.ts
hassan254-prog: I would say more than 1% offer scoped api keys making it hard for us to verify with a single endpoint.

bodinsamuel: Not a fan of multi endpoints check but is there a way for us to know what credentials can match which endpoints? I guess not really, so even with a script, we will still end up querying multiple endpoints 😢 
So okay for multi-endpoints. imo we should move toward programmatic options and get rid of this kind of stuff in yaml

scripts/validation/providers/schema.json
bodinsamuel: I see, thanks 🙏🏻 
I'm still wondering about retro-compatibility (having endpoint and endpoints allowed for a while) but I can't think of a strong reason to do so, so let's proceed like this and we'll see 👍🏻

packages/server/lib/hooks/hooks.ts
```diff
@@ -72,6 +73,65 @@ export const connectionCreationStartCapCheck = async ({
...
+        {
+            type: 'log',
+            level: 'info',
+            message: 'Running automatic credentials verification via verification script',
+            createdAt: new Date().toISOString()
+        }
+    ];
+
+    try {
+        if (provider?.credentials_verification_script) {
```
bodinsamuel:
```suggestion
        if (provider.credentials_verification_script) {
```

packages/server/lib/hooks/hooks.ts
```diff
@@ -72,6 +73,65 @@ export const connectionCreationStartCapCheck = async ({
...
+            await executeVerificationScript(config, credentials, connectionId, connectionConfig);
+            return Ok({ logs, tested: true });
+        }
+
+        if (provider?.proxy?.verification) {
```
bodinsamuel:
```suggestion
        if (provider.proxy?.verification) {
```

packages/server/lib/hooks/connection/credentials-verification-script.ts
```diff
@@ -0,0 +1,124 @@
...
+import tracer from 'dd-trace';
+
+type VerificationScriptHandler = (internalNango: InternalNango) => Promise<void>;
+
+type VerificationScriptHandlersMap = Record<string, VerificationScriptHandler>;
+
+const handlers: VerificationScriptHandlersMap = verificationscriptHandlers as unknown as VerificationScriptHandlersMap;
+
+export interface InternalNango {
+    getCredentials: () => Promise<{
```
bodinsamuel: getCredentials doesn't seems to be a promise ever

bodinsamuel (approved):
Works well (at least for the endpoints check) 🙏🏻

TBonnin (approved):

PR #3669, opened 2025-03-12
feat(integrations): add support for readwise and readwise reader

hassan254-prog:

- Add support form readwise and readwise reader
- The main difference between the 2 is the api versioning. For readwise, they use `v2`, while for readwise reader they use `v3`


hassan254-prog:
Opened this new PR in favor [of](https://github.com/NangoHQ/nango/pull/3231).

bodinsamuel (approved):

PR #3672, opened 2025-03-12
fix(connections): fill new refresh columns

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2854/store-credential-expires-at

- Fill new refresh columns
Set credentials_expires_at when available and set success or failure when applicable. Columns are still not used


bodinsamuel:
Too much conflict

PR #3673, opened 2025-03-12
docs: fix real-time sync docs info

bastienbeurier:


docs-v2/guides/webhooks/webhooks-from-apis/real-time-syncing-with-webhooks.mdx
```diff
@@ -45,15 +45,6 @@ hubspot:
...
 ```
 
-Use `'*'` to list for any webhook subscription, e.g. 
-```yaml
-airtable:
-  syncs:
-    airtable-sync-name:
-	    webhook-subscriptions:
-			  - '*'
-```
```
bodinsamuel, line 55: We don't want that to be documented?

bodinsamuel (approved):

docs-v2/guides/webhooks/webhooks-from-apis/real-time-syncing-with-webhooks.mdx
bastienbeurier: I would like to, but it's not true :) It's an Airtable-specific notation that I thought was Nango-specific.

PR #3674, opened 2025-03-12
fix(connection): split refresh function

bodinsamuel:
## Changes 

- Move connection refresh method (and directly associated method) into it's own file
connection.service.ts is hard to parse and maintain, since I need to touch it a lot I figured it was time to do something. I tried to keep it simple, I didn't move everything and tried to keep the code almost similar.
There is a huge refactor pending this but not today

#### Notable changes:
- refreshOrTestCredentials now uses a switch case and call more function for readability
- refresh and test functions now always return the new connection, because we are going to modify more than just credentials, it was a footgun before


packages/shared/lib/services/connections/credentials/refresh.ts
```diff
@@ -0,0 +1,385 @@
...
+                metrics.increment(metrics.Types.REFRESH_CONNECTIONS_UNKNOWN);
+                res = Ok(props.connection);
+                break;
+            }
+            default: {
+                throw new Error('Unsupported credentials type');
+            }
+        }
+
+        // TODO: remove this
```
TBonnin, line 109: curious: why remove?

packages/shared/lib/services/connections/credentials/refresh.ts
```diff
@@ -0,0 +1,385 @@
...
+
+    return Ok(value.connection);
+}
+
+/**
+ * Try testing credentials,
+ * only relevant for some providers that have free-form credentials
+ */
+async function testCredentials(
+    { account, integration, environment, connection: oldConnection, logContextGetter, connectionTestHook, onRefreshFailed, onRefreshSuccess }: RefreshProps,
```
TBonnin, line 193: there doesn't seem to be any concept of old/new connection in this function. Can this variable be simply called `connection`?

TBonnin (approved):
thank you for doing this

packages/shared/lib/services/connections/credentials/refresh.ts
bodinsamuel: last_fetched is only used to refresh connection

packages/shared/lib/services/connections/credentials/refresh.ts
bodinsamuel: refreshRes actually returns a new connection, since some part of the previous was confusingly using both I feel like it's more future proof to have this name everywhere so never return an expired connection (or at least make that super obvious)

PR #3676, opened 2025-03-12
chore(persist): add accountId dimension to records metrics

TBonnin:
Adding accountId dimension so we can drill down those metrics per account.

Adding support for `Locals` in the express route types to make the account available in the endpoint handlers

### DIFF
--- a/packages/persist/lib/records.ts
+++ b/packages/persist/lib/records.ts
@@ -13,6 +13,7 @@ export const recordsPath = '/environment/:environmentId/connection/:nangoConnect
 
 export async function persistRecords({
     persistType,
+    accountId,
     environmentId,
     connectionId,
     providerConfigKey,
@@ -25,6 +26,7 @@ export async function persistRecords({
     merging = { strategy: 'override' }
 }: {
     persistType: PersistType;
+    accountId: number;
     environmentId: number;
     connectionId: string;
     providerConfigKey: string;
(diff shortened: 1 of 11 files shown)


packages/persist/lib/records.ts
```diff
@@ -124,8 +126,8 @@ export async function persistRecords({
         });
         await updateSyncJobResult(syncJobId, updatedResults, baseModel);
 
-        metrics.increment(metrics.Types.PERSIST_RECORDS_COUNT, records.length);
-        metrics.increment(metrics.Types.PERSIST_RECORDS_SIZE_IN_BYTES, recordsSizeInBytes);
+        metrics.increment(metrics.Types.PERSIST_RECORDS_COUNT, records.length, { accountId });
+        metrics.increment(metrics.Types.PERSIST_RECORDS_SIZE_IN_BYTES, recordsSizeInBytes, { accountId });
```
TBonnin, line 130: This is the metric change, everything else purpose is to make the account available in the endpoint handler logic

packages/persist/lib/routes/environment/environmentId/connection/connectionId/getCursor.ts
```diff
@@ -3,6 +3,7 @@ import { validateRequest } from '@nangohq/utils';
 import type { EndpointRequest, EndpointResponse, RouteHandler, Route } from '@nangohq/utils';
```
bodinsamuel, line 3: just realized since we are passing the secretKey `/environement/${id}` could be removed

bodinsamuel (approved):
TS is not happy but Looks good 🙏🏻

packages/persist/lib/routes/environment/environmentId/connection/connectionId/getCursor.ts
TBonnin: yes, it is nice for debugging to filter by `url` though

TBonnin:
> TS is not happy but Looks good 🙏🏻

forgot to stage one code change

PR #3681, opened 2025-03-12
fix(jobs): drop where deleted index

bodinsamuel:
## Changes 

Fixes https://linear.app/nango/issue/NAN-2884/remove-soft-delete-from-jobs

- Drop where deleted index
I'll drop the column in a couple of months maybe, to wait for enterprise to migrate

In the last 48hours those 2 index have now used
Screenshot 2025-03-13 at 00 17 35


TBonnin (approved):
not sure I follow the `waiting for enterprise to migrate` logic though

bodinsamuel:
> not sure I follow the waiting for enterprise to migrate logic though

Because stuff are created concurrently it's already kind of not a good idea to create many at the same time, but if I drop the column in between it might create some chocapics

PR #3684, opened 2025-03-13
feat(connect): telemetry

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2890/create-telemetryevent-api 
Fixes https://linear.app/nango/issue/NAN-2891/modify-connect-ui-to-call-telemetry-api

- `POST /connect/telemetry` 
Used by sendBeacon, as lightweight as possible. We send controlled event code so nobody can mess up with the messages.
Because sendBeacon is different than regular fetch it required a few tweaks regarding auth and cors

- UI calls this new endpoint

- Create operation when we create the session token
The only alternative is to use a lock when needed but annoying 

## 🧪 tests
- Trigger an auth with the connect ui
Screenshot 2025-03-13 at 16 23 20


packages/connect-ui/src/components/ErrorFallback.tsx
```diff
@@ -1,7 +1,9 @@
 import { IconX } from '@tabler/icons-react';
+import { useMount } from 'react-use';
 
 import { APIError } from '@/lib/api';
 import { triggerClose } from '@/lib/events';
+import { telemetry } from '@/lib/telemetry';
```
TBonnin, line 6: not sure telemetry is a good name. Since it is limited to send beacon event, why not just `beacon`

packages/types/lib/connect/api.ts
```diff
@@ -80,3 +80,29 @@ export type PostInternalConnectSessions = Endpoint<{
     Success: PostConnectSessions['Success'];
     Body: Pick<ConnectSessionInput, 'allowed_integrations' | 'end_user' | 'organization'>;
 }>;
+
+export type PostPublicConnectTelemetry = Endpoint<{
+    Method: 'POST';
+    Path: '/connect/telemetry';
```
TBonnin, line 86: would use `/connect/beacon`

TBonnin (approved):
looks good. I would not use the word `telemetry` but `beacon` instead

bodinsamuel:
> looks good. I would not use the word telemetry but beacon instead

It's a telemetry lib and endpoint, it's using sendBeacon as the tech behind, a beacon is the payload. I'm not sure I follow why it would be a better name.

PR #3686, opened 2025-03-13
chore: cron job to export connection count metric to dd

TBonnin:
also moving the refresh connection cron to `/crons/` folder

### DIFF
--- a/packages/server/lib/crons/exportMetrics.ts
+++ b/packages/server/lib/crons/exportMetrics.ts
@@ -1,0 +1,46 @@
+import * as cron from 'node-cron';
+import { errorManager, ErrorSourceEnum, connectionService } from '@nangohq/shared';
+import { stringifyError, getLogger, metrics } from '@nangohq/utils';
+import tracer from 'dd-trace';
+
+const logger = getLogger('Server');
+const cronName = '[exportUsageMetrics]';
+const cronMinutes = 5;
+
+export function exportUsageMetricsCron(): void {
+    cron.schedule(`*/${cronMinutes} * * * *`, () => {
+        (async () => {
+            try {
+                await exec();
+            } catch (err) {
+                const e = new Error('failed_to_export_metrics', {
+                    cause: err instanceof Error ? err.message : String(err)
+                });
+                errorManager.report(e, { source: ErrorSourceEnum.PLATFORM });
+            }
+        })().catch((err: unknown) => {
+            logger.error(`Failed to execute ${cronName} cron job`);
+            logger.error(err);
+        });
+    });
+}
+
+export async function exec(): Promise<void> {
+    await tracer.trace<Promise<void>>('nango.server.cron.exportUsageMetrics', async (span) => {
+        try {
+            logger.info(`${cronName} starting`);
+
+            const res = await connectionService.countMetric();
+            if (res.isErr()) {
+                throw res.error;
+            }
+            for (const { accountId, count } of res.value) {
+                metrics.gauge(metrics.Types.CONNECTIONS_COUNT, count, { accountId });
+            }
+            logger.info(`${cronName} ✅ done`);
+        } catch (err) {
+            logger.error(`${cronName} failed: ${stringifyError(err)}`);
+            span.setTag('error', err);
+        }
+    });
+}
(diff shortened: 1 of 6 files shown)


packages/server/lib/crons/exportMetrics.ts
```diff
@@ -0,0 +1,46 @@
+import * as cron from 'node-cron';
+import { errorManager, ErrorSourceEnum, connectionService } from '@nangohq/shared';
+import { stringifyError, getLogger, metrics } from '@nangohq/utils';
+import tracer from 'dd-trace';
+
+const logger = getLogger('Server');
+const cronName = '[exportUsageMetrics]';
+const cronMinutes = 5;
```
TBonnin: arbitrary, happy to discuss. it is doing a single query to the main db so it is definitely not a very heavy cron

packages/shared/lib/services/connection.service.ts
```diff
@@ -2049,6 +2049,24 @@ class ConnectionService {
...
+    async countMetric(): Promise<Result<{ accountId: number; count: number }[], NangoError>> {
+        const res = await db.knex
+            .withSchema('nango')
+            .from('_nango_connections')
+            .debug(true)
+            .join('_nango_environments', '_nango_connections.environment_id', '_nango_environments.id')
+            .count('_nango_connections.id as count')
+            .select<{ accountId: number; count: number }[]>('_nango_environments.account_id')
+            .where('_nango_connections.deleted_at', null)
+            .groupBy('_nango_environments.account_id');
```
TBonnin, line 2062: Query is currently <500ms in prod. The query plan could be improved but I am cautious about adding indexes for a single query that runs once every 5mins

packages/server/lib/crons/exportMetrics.ts
```diff
@@ -0,0 +1,46 @@
+import * as cron from 'node-cron';
+import { errorManager, ErrorSourceEnum, connectionService } from '@nangohq/shared';
+import { stringifyError, getLogger, metrics } from '@nangohq/utils';
+import tracer from 'dd-trace';
+
+const logger = getLogger('Server');
+const cronName = '[exportUsageMetrics]';
```
bodinsamuel: It's legacy from when we didn't have better. Using the logger name will automatically prepend all the logs
```suggestion
const logger = getLogger('Server.exportUsageMetrics');
```

packages/shared/lib/services/connection.service.ts
```diff
@@ -2049,6 +2049,24 @@ class ConnectionService {
             return { success, error, response: success ? (creds as OAuth2Credentials) : null };
         }
     }
+
+    // return the number of active connections per account
+    async countMetric(): Promise<Result<{ accountId: number; count: number }[], NangoError>> {
+        const res = await db.knex
+            .withSchema('nango')
```
bodinsamuel, line 2056: withSchema can't be used, but it's also not necessary
```suggestion
```

packages/server/lib/crons/exportMetrics.ts
```diff
@@ -0,0 +1,46 @@
+import * as cron from 'node-cron';
```
bodinsamuel, line 1: Not sure it belongs to server (or jobs tbh) but please mind all the pods will execute this cron job so it's going to be executed more than once

packages/server/lib/crons/exportMetrics.ts
bodinsamuel: can be a env var just in case

bodinsamuel (approved):

packages/shared/lib/services/connection.service.ts
TBonnin: why can we not used it?

packages/server/lib/crons/exportMetrics.ts
TBonnin: from a metrics point of view it doesn't matter since the value we are gonna be interested in is the last one.
Do you think it is worth locking anyway?

bodinsamuel: not sure I just wanted to point out, worst case all of them don't report the same value but you can use `max` I guess

packages/shared/lib/services/connection.service.ts
bodinsamuel: because we allow different schema process.env['NANGO_DB_SCHEMA']

PR #3687, opened 2025-03-13
fix(logs): put date in operation id

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2846/investigate-using-operation-id-with-the-date-inside-the-id

- Put date in operation id
This will allow us to get the indexName directly from the ID and also be able to compute duration directly from this date
e.g: `1741878251660_XQMgbkGG`
I have set 8 to avoid very long id, it should be more than enough to not have any collision since we have the date prepended https://zelark.github.io/nano-id-cc/


TBonnin (approved):
does that mean we don't need redis anymore to retrieve the index?

bodinsamuel:
> does that mean we don't need redis anymore to retrieve the index? 

That would be the goal yes. I need that in prod first, then test in > 48hours

PR #3688, opened 2025-03-13
chore(metrics): we also want connections with syncs/actions/webhooks count

TBonnin:


packages/utils/lib/telemetry/metrics.ts
```diff
@@ -78,7 +78,10 @@ export enum Types {
     GET_RECORDS_COUNT = 'nango.server.getRecords.count',
     GET_RECORDS_SIZE_IN_BYTES = 'nango.server.getRecords.sizeInBytes',
 
-    CONNECTIONS_COUNT = 'nango.server.connections.count'
```
TBonnin, line 81: doesn't have anything to do with server. it just happens to run  there (for now)

bodinsamuel: I think we should avoid prefix with the services in general 👌🏻

bodinsamuel (approved):

PR #3690, opened 2025-03-14
fix(cron): delete old data

bodinsamuel:
## Changes

Contributes to https://linear.app/nango/issue/NAN-2912/delete-old-data

- Move cron to server because it can be useful for self-hosted

- Delete expired private keys and connect sessions
It's missing more but trying to keep PR small


packages/keystore/lib/models/privatekeys.ts
```diff
@@ -152,3 +152,15 @@ function decryptValue(encryptedValue: Buffer): Result<string, PrivateKeyError> {
     }
     return Ok(encryption.decrypt(encrypted, iv, tag));
 }
+
+export async function deleteExpiredPrivateKeys(db: knex.Knex, { limit, deleteOlderThan }: { limit: number; deleteOlderThan: number }): Promise<number> {
```
TBonnin: `deleteOlderThan` could be `olderThan` since it is a parameter of a function called `delete...`

packages/server/lib/crons/deleteOldData.ts
```diff
@@ -0,0 +1,107 @@
...
+
+const cronMinutes = envs.CRON_DELETE_OLD_DATA_EVERY_MIN;
+
+const limit = envs.CRON_DELETE_OLD_JOBS_LIMIT;
+const deleteJobsOlderThan = envs.CRON_DELETE_OLD_JOBS_MAX_DAYS;
+
+const deleteConnectionSessionOlderThan = envs.CRON_DELETE_OLD_CONNECT_SESSION_MAX_DAYS;
+const deletePrivateKeysOlderThan = envs.CRON_DELETE_OLD_PRIVATE_KEYS_MAX_DAYS;
+
+export function deleteOldData(): void {
```
TBonnin, line 23: can we add a kill switch for every cron. ie: if  `cronMinutes == 0` nothing is started. Just in case

TBonnin (approved):

PR #3691, opened 2025-03-14
fix(logs): set auth operation as cancelled when expired

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2789/nango-connect-improve-logging

- Set auth operation as cancelled when expired
- Move cron jobs back to server, because self-hosted might use Elasticsearch


TBonnin (approved):
CRON ALL THE THINGS!!!!!

Should this one be in server as well? does being in jobs mean self-hosted don't have logs and logs cancellation

bodinsamuel:
> Should this one be in server as well? does being in jobs mean self-hosted don't have logs and logs cancellation

Self-hosted logs are disabled by default, but they can install ES, so if they do, they didn't have auto cancellation before. It wasn't a huge problem because I don't expect that scenario to happen often but let's keep it clean and do the right thing

PR #3709, opened 2025-03-14
chore: add metric for records total count per account

TBonnin:
Using the `records_count` table to aggregate the total number of records per account and sending the metric in the same cron job as the connection count.
I went for a single cron that submit the different metrics sequentially. wdyt?

### DIFF
--- a/packages/server/lib/crons/exportMetrics.ts
+++ b/packages/server/lib/crons/exportMetrics.ts
@@ -1,8 +1,9 @@
 import * as cron from 'node-cron';
-import { errorManager, ErrorSourceEnum, connectionService } from '@nangohq/shared';
+import { errorManager, ErrorSourceEnum, connectionService, environmentService } from '@nangohq/shared';
 import { stringifyError, getLogger, metrics } from '@nangohq/utils';
 import tracer from 'dd-trace';
 import { envs } from '../env.js';
+import { records } from '@nangohq/records';
 
 const logger = getLogger('Server.exportUsageMetrics');
 const cronMinutes = envs.CRON_EXPORT_USAGE_METRICS_MINUTES;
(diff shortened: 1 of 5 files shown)


packages/shared/lib/services/connection.service.ts
```diff
@@ -2054,7 +2054,7 @@ class ConnectionService {
     async countMetric(): Promise<
         Result<
             {
-                accountId: number;
+                account_id: number;
```
TBonnin: pg doesn't like camel case and transform field name to lowercase

bodinsamuel: it works with quotes

packages/shared/lib/services/environment.service.ts
```diff
@@ -224,6 +224,12 @@ class EnvironmentService {
         return encryptionManager.decryptEnvironment(result[0]);
     }
 
+    async getAll(): Promise<DBEnvironment[]> {
+        const result = await db.knex.select('*').from<DBEnvironment>(TABLE);
+        // NOTE: we are not decrypting the environments
+        return result || [];
+    }
```
bodinsamuel, line 230: not sure that scales very much, how about select only the id to reduce memory footprint?

packages/server/lib/crons/exportMetrics.ts
```diff
@@ -22,25 +23,69 @@ export function exportUsageMetricsCron(): void {
...
+            });
+            errorManager.report(e, { source: ErrorSourceEnum.PLATFORM });
+        }
+    });
+}
 
-            const res = await connectionService.countMetric();
-            if (res.isErr()) {
-                throw res.error;
+async function exportRecordsMetrics(): Promise<void> {
```
bodinsamuel, line 58: not sure this whole function will scale well but let see

bodinsamuel (approved):

PR #3710, opened 2025-03-14
chore: pass accountId to log context to use it in metrics

TBonnin:
we could also eventually add accountId and/or environmentId to add to the payload to make filtering easier

Tested in staging

### DIFF
--- a/packages/logs/lib/client.ts
+++ b/packages/logs/lib/client.ts
@@ -17,11 +17,13 @@ interface Options {
  */
 export class LogContextStateless {
     id: OperationRow['id'];
+    accountId: OperationRow['accountId'];
     dryRun: boolean;
     logToConsole: boolean;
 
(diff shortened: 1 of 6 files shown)


packages/logs/lib/client.ts
```diff
@@ -49,7 +51,7 @@ export class LogContextStateless {
             report(new Error('failed_to_insert_in_es', { cause: err }));
             return false;
         } finally {
-            metrics.duration(metrics.Types.LOGS_LOG, Date.now() - start);
+            metrics.duration(metrics.Types.LOGS_LOG, Date.now() - start, { accountId: this.accountId });
```
TBonnin, line 54: all of this to be able to add a accountId dimension

bodinsamuel (approved):
Looks good 👌🏻

PR #3712, opened 2025-03-14
fix(cron): delete all old data

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2912/delete-old-data

- Cron: delete all old data
  - delete soft deleted connections and associated data
  - delete soft deleted sync configs and associated data
  - delete soft deleted integrations and associated data
  - delete oauth sessions the same way
  - delete expired invitations

There are 3 db queries in the cron because there are super-specific and not filtered, I wasn't too sure it was a footgun putting them in the service. But let me know.

Everything is batched because it's almost exponential since each step can contain N children. Most of the time it will be too verbose and could have been batched but it's better than slow query.

- Remove cron in jobs
The only drawbacks is that we won't delete records straight away but this cron has always been tricky 

## 🧪 Tests 
- Set env var

```sh
CRON_DELETE_OLD_DATA_EVERY_MIN=1
CRON_DELETE_OLD_CONNECT_SESSION_MAX_DAYS=0
CRON_DELETE_OLD_PRIVATE_KEYS_MAX_DAYS=0
CRON_DELETE_OLD_OAUTH_SESSION_MAX_DAYS=0
CRON_DELETE_OLD_INVITATIONS_MAX_DAYS=0
CRON_DELETE_OLD_CONFIGS_MAX_DAYS=0
CRON_DELETE_OLD_SYNC_CONFIGS_MAX_DAYS=0
CRON_DELETE_OLD_CONNECTIONS_MAX_DAYS=0
```

- Delete one connection in the UI
Should delete one connection after 1 minute
- Delete an integration with a connection
Should delete one connection and an integration after 1 minute
- Delete an integration with a connection with records
Should delete records, one connection and an integration after 1 minute

**After deleting an integration with records**
Screenshot 2025-03-14 at 19 45 33
Screenshot 2025-03-14 at 19 45 41

> It's quite verbose so we can't miss it.


packages/server/lib/crons/deleteOldData.ts
```diff
@@ -50,62 +70,226 @@ export async function exec(): Promise<void> {
...
-            if (Date.now() - startTimestamp > ttlMs) {
-                logger.info(`Time limit reached, stopping`);
-                return;
-            }
-            await setTimeout(1000);
-        }
+        await batchDelete({
+            name: 'jobs',
+            startTimestamp,
+            ttlMs,
```
TBonnin: We are passing the same `ttlMs` to all batchDelete call but if the first one take 2mins the following one should have less time to execute. I would pass a `deadline` date instead?

packages/server/lib/crons/deleteOldData.ts
```diff
@@ -50,62 +70,226 @@ export async function exec(): Promise<void> {
...
+        if (deleted < limit) {
+            break;
+        }
+        if (Date.now() - startTimestamp > ttlMs) {
+            logger.info(`Time limit reached, stopping`);
+            return;
+        }
+        await setTimeout(1000);
+    }
+}
```
TBonnin, line 228: should `deleteFn` be passed a transaction? If the function executes multiples queries and one fails, we would end up with a partial deletion

packages/server/lib/crons/deleteOldData.ts
```diff
@@ -50,62 +70,226 @@ export async function exec(): Promise<void> {
...
 }
+
+async function batchDelete({
+    name,
+    deleteFn,
+    startTimestamp,
+    ttlMs
+}: {
+    name: string;
+    deleteFn: () => Promise<number>;
```
TBonnin: we have to be careful when passing deleteFn: if an entity is used to retrieve children, we must ensure that the entity is deleted last. Otherwise if the deletion is interrupted by a exception or expires we wouldn't be able to retrieve the childer again and end up with orphans.

packages/server/lib/crons/deleteOldData.ts
bodinsamuel: good point

packages/server/lib/crons/deleteOldData.ts
bodinsamuel: the problem with transaction is that it needs to keep everything in memory which is not always possible, degrading the whole instance performance.
If one fails, it will throw, bubble up and stop the cron, starting back from where it was next time

packages/server/lib/crons/deleteOldData.ts
bodinsamuel: this is what happened yes, we always delete at the end (unless mistaken)

TBonnin: this is what it is done indeed. just commenting that it could be easy to forget about this implicit assumption in the next batchDeletes we are going to add eventually. I don't think we can enforce anything unfortunately

TBonnin (approved):

bodinsamuel:
Tested on staging, it's working as intended 🤞🏻

PR #3713, opened 2025-03-14
chore: send metric with accountId as value and accountName dimension

TBonnin:
A little trick to be able to show a list in the usage dashboard linking the account id and the name. 
This metrics doesn't represent anything, it will just have a constant value that is the account id

Tested in staging
Screenshot 2025-03-14 at 16 28 31


bodinsamuel (commented):
I'm not sure about that. It's going to create 6000 metrics (= 1500$) just for a list

TBonnin:
> I'm not sure about that. It's going to create 6000 metrics (= 1500$) just for a list

$0.05 per metrics = $300. still just for a list. cc @bastienbeurier you decide if it is worth it

bodinsamuel:
My bad I thought it was 25$ but its 5€ per 100 👌🏻 I'm not sure it's super worth it but less of a pain indeed

bastienbeurier:
$300/month? Indeed, IMO not worth the price.

PR #3714, opened 2025-03-16
fix: use knex Transaction for all queries within a transaction

TBonnin:
We are not always using the Knex Transaction when making queries within a transaction, using more connections that necessary and potentially causing exhaustion of the db connections pool

### DIFF
--- a/packages/shared/lib/services/notification/slack.service.ts
+++ b/packages/shared/lib/services/notification/slack.service.ts
@@ -532,7 +532,7 @@ export class SlackService {
         provider: string;
     }): Promise<void> {
         const update = await db.knex.transaction(async (trx) => {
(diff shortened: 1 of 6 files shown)


packages/shared/lib/services/notification/slack.service.ts
```diff
@@ -532,7 +532,7 @@ export class SlackService {
         provider: string;
     }): Promise<void> {
         const update = await db.knex.transaction(async (trx) => {
-            const slackNotificationsEnabled = await environmentService.getSlackNotificationsEnabled(nangoConnection.environment_id);
+            const slackNotificationsEnabled = await environmentService.getSlackNotificationsEnabled(nangoConnection.environment_id, trx);
```
TBonnin, line 535: this one is problematic, it is on a very frequently executed code path (ie: every sync or action successful completion)

bodinsamuel (approved):

PR #3716, opened 2025-03-17
feat(integrations): add support for pandadoc api key

hassan254-prog:

- Add support for pandadoc api key


packages/webapp/public/images/template-logos/pandadoc-api-key.svg
```diff
@@ -0,0 +1 @@
+pandadoc.svg
```
bodinsamuel, line 1: Not from this PR but the background is just the ukrainian logo not part of the actual logo, if you have the motivation to remove it go for it 👍🏻

bodinsamuel (approved):

packages/webapp/public/images/template-logos/pandadoc-api-key.svg
hassan254-prog: seems like the one with the ukraine logo is their official logo, but confusing 🤔  at the same time since their website favicon  has a green background
Screenshot 2025-03-17 at 17 33 11

PR #3717, opened 2025-03-17
feat(hooks): reorder provider specific code under the provider instead of the type 

hassan254-prog:

- Reordered provider specific code under the provider instead of the type
- Added validation to ensure that `webhook_routing_script`, `post_connection_script`, and `credentials_verification_script` referenced in the provider configuration exist in their respective `index.ts`.


TBonnin (approved):

scripts/validation/providers/validate.ts
```diff
@@ -226,6 +230,39 @@ function validateProvider(providerKey: string, provider: Provider) {
...
+    if (provider.credentials_verification_script) {
+        if (!scriptsContent.includes(provider.credentials_verification_script)) {
+            console.error(
+                chalk.red('error'),
+                chalk.blue(providerKey),
+                `credentials_verification_script "${provider.credentials_verification_script}" not found in connection scripts index.ts`
+            );
+            error = true;
+        }
+    }
```
bodinsamuel, line 265: Thanks for going all the way 🚀

bodinsamuel (approved):
Nice. I think we will need to consolidate everything into shared or providers or another package at some point to have a single folder per provider, but that's a good step forward 💪🏻

PR #3718, opened 2025-03-17
fix(connections): fill refresh columns, fix row_to_json types

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2854/store-credential-expires-at-and-refresh

- Fill refresh columns
Store credentials expiration (when possible otherwise default value), store last success or failure, store attempts 

- Fix row_to_json types
Because I started using a date I realized the type was still slightly wrong since default select will parse date but not row_to_json.

## 🧪 Tests
- Run everything `npm run dev:watch:web`
- Go to any connection 
  - Check _nango_connections is backfilled for this connection
- Wait for the cron
  - `UPDATE _nango_connections SET last_fetched_at = NULL;` to make the cron trigger for all connections
  - Check _nango_connections is backfilled for all connections


packages/server/lib/formatters/connection.ts
```diff
@@ -4,6 +4,7 @@ import type {
     ApiPublicConnection,
     ApiPublicConnectionFull,
     DBConnection,
+    DBConnectionAsJSONRow,
```
TBonnin, line 7: would that make more sense to modify the queries so they actually return `DBConnection` instead of adding a slightly different type?

packages/shared/lib/services/connections/credentials/refresh.ts
```diff
@@ -111,9 +113,24 @@ export async function refreshOrTestCredentials(props: RefreshProps): Promise<Res
 
         if (res.isErr()) {
             span.setTag('error', res.error);
+            await connectionService.setRefreshFailure({ id: props.connection.id, currentAttempt: props.connection.refresh_attempts || 0 });
             return res;
         }
-        return Ok(res.value);
+
+        let newConnection = res.value;
+        // Backfill
```
TBonnin, line 121: not sure I understand the backfill comment

packages/shared/lib/services/connections/credentials/refresh.ts
```diff
@@ -300,7 +328,8 @@ async function refreshCredentialsIfNeeded({
 
         return {
             connection,
-            freshCredentials: shouldRefresh ? null : (connection.credentials as RefreshableCredentials)
+            shouldRefresh: shouldRefresh,
```
TBonnin:
```suggestion
            shouldRefresh,
```

packages/shared/lib/services/connections/credentials/refresh.ts
```diff
@@ -369,17 +409,25 @@ async function shouldRefreshCredentials(
     providerConfig: ProviderConfig,
     provider: RefreshableProvider,
     instantRefresh: boolean
-): Promise<boolean> {
-    const refreshCondition =
-        instantRefresh ||
-        (providerClient.shouldIntrospectToken(providerConfig.provider) && (await providerClient.introspectedTokenExpired(providerConfig, connection)));
+): Promise<{ should: boolean; reason: string }> {
+    if (providerClient.shouldIntrospectToken(providerConfig.provider) && (await providerClient.introspectedTokenExpired(providerConfig, connection))) {
```
TBonnin: did checking `instantRefresh` disappear from the condition?

packages/shared/lib/services/connections/utils.ts
```diff
@@ -0,0 +1,25 @@
...
+export const MAX_FAILED_REFRESH = 30;
+export const REFRESH_MARGIN_S = ms('15minutes') / 1000;
+
+export function getExpiresAtFromCredentials(credentials: AllAuthCredentials): Date | null {
+    if ('expires_at' in credentials && credentials['expires_at']) {
+        return credentials['expires_at'];
+    }
+
+    if (credentials.type === 'CUSTOM' || credentials.type === 'OAUTH1' || credentials.type === 'APP_STORE' || !credentials.type) {
+        return new Date(Date.now() + DEFAULT_UNAUTH_EXPIRES_AT_MS);
```
TBonnin: name of the constant is a bit confusing. What about mentioning infinite expiration or something similar?

packages/shared/lib/utils/encryption.manager.ts
```diff
@@ -79,17 +80,29 @@ export class EncryptionManager extends Encryption {
...
+            const parsed: DBConnectionDecrypted = {
+                ...connection,
+                credentials,
+                last_fetched_at: connection.last_fetched_at ? new Date(connection.last_fetched_at) : null,
+                credentials_expires_at: connection.credentials_expires_at ? new Date(connection.credentials_expires_at) : null,
+                last_refresh_success: connection.last_refresh_success ? new Date(connection.last_refresh_success) : null,
+                last_refresh_failure: connection.credentials_expires_at ? new Date(connection.credentials_expires_at) : null,
+                created_at: new Date(connection.created_at),
+                updated_at: new Date(connection.updated_at),
+                deleted_at: connection.deleted_at ? new Date(connection.deleted_at) : null
```
TBonnin, line 98: is that only required temporarily?

TBonnin (approved):
a few questions but nothing blocking

packages/server/lib/formatters/connection.ts
bodinsamuel: It's not possible, unfortunately, when we `await knex.select()` the query returns Date, when possible because the underlying `pg` package is parsing them, or a string.
Honestly what I have done in the past is to prevent this parsing but it's hard to fix the past because it creeps everywhere in the code.

packages/shared/lib/services/connections/credentials/refresh.ts
bodinsamuel: mmh yes indeed, and I messed up an other condition. Will add some test 👌🏻

packages/shared/lib/utils/encryption.manager.ts
bodinsamuel: unfortunately no, see your first question. Happy to get suggestion

packages/shared/lib/services/connections/credentials/refresh.ts
bodinsamuel: true, I'll add a better comment

PR #3719, opened 2025-03-17
fix(logs): compute index name from operation id

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2925/use-id-to-compute-index-name

- Compute index name from operation id
This finally gets rid of using Redis and es search to get back the operation. 
To limit the changes and still keep the library easy to use I had to create a new sub class that still contains the operation, but as soon as you get the operation back from another service it does not accept an operation so it's not possible to break this part.
Because of this, I had to pass the `accountId` manually where needed, as planned.

Hopefully, this should drastically improve perf, if not I do 100 push-up 🫡


packages/logs/lib/models/logContextGetter.ts
```diff
@@ -29,6 +29,7 @@ export const logContextGetter = {
 
         try {
             if (envs.NANGO_LOGS_ENABLED && !options?.dryRun) {
+                // TODO: remove this after full deploy
```
TBonnin, line 32: 🫶

packages/logs/lib/models/logContextGetter.ts
```diff
@@ -39,30 +40,19 @@ export const logContextGetter = {
...
-            }
-        } catch (err) {
-            report(new Error('failed_to_get_operation', { cause: err }), { id });
-        }
-
-        // If it failed, we create a fake operation for now
-        return new LogContext({ parentId: id, operation: { id: nanoid(), createdAt: new Date().toISOString() } as OperationRow }, { ...options, dryRun: true });
+    async get({ id, accountId }: { id: OperationRow['id']; accountId?: number | undefined }, options?: Options): Promise<LogContext> {
+        const split = id.split('_');
+        const createdAt = split[0] ? new Date(parseInt(split[0], 10)).toISOString() : new Date().toISOString(); // Fallback to default date
```
TBonnin, line 51: I don't know why we didn't think about it earlier 😝

TBonnin (approved):
👍

packages/logs/lib/models/logContextGetter.ts
bodinsamuel: yes and no, I didn't think about it at the time because I knew I could remove the ID and get better performance at indexing. So in my mind, it was about to get removed. However, the simplicity probably far outweighs the perf gain :D Will see 🤞🏻

bodinsamuel:
Didn't change anything :|

PR #3721, opened 2025-03-17
chore(eslint): import order as warning

bodinsamuel:
## Changes

- Enable `import/order` as warning
The goal is to provide consistency in our codebase, improve readability, and limit the noise when changing import.

This means all changed files will be automatically linted. Never too sure if it's the best strategy, but it's sometimes better than having one large commit that doesn't mean much.
The strategy is:
```
- side effect import `import ./tracer.js`
- built-in node package
- external npm package
- our own 
- local import 
- types
```
Import and object import are also sorted alphabetically

- Tested on a few file just for you to see


packages/cli/lib/index.ts
```diff
@@ -4,25 +4,27 @@
  * Copyright (c) 2024 Nango, all rights reserved.
  */
 
-import { Command } from 'commander';
 import fs from 'fs';
-import chalk from 'chalk';
-import figlet from 'figlet';
 import path from 'path';
+
+import { nangoConfigFile } from '@nangohq/nango-yaml';
```
TBonnin, line 10: I am not sure about the empty line. I'd prefer without it but I won't fight for it

TBonnin (commented):
👌

TBonnin (approved):

packages/cli/lib/index.ts
bodinsamuel: you mean no empty lines between groups or for this group in particular?
I'll merge as-is but we can talk about it if that becomes too annoying

TBonnin: I mean no empty lines between groups

PR #3722, opened 2025-03-17
chore: Slack notification sent via proxy instead of using an action

TBonnin:
https://linear.app/nango/issue/NAN-2903/stop-using-actions-for-slack-notifications

To test you need 
- a Slack integration setup in your local prod environment and the `NANGO_ADMIN_KEY` and `NANGO_ADMIN_UUID` env vars setup locally. 
- set the Slack alerts in your environment settings to connect a Slack channel
- run a failing sync/action to receive a notif in Slack


packages/server/lib/services/slack.ts
```diff
@@ -0,0 +1,8 @@
+import { getFeatureFlagsClient } from '@nangohq/kvstore';
+import { logContextGetter } from '@nangohq/logs';
+import { SlackService } from '@nangohq/shared';
+
+export const slackService = new SlackService({
+    logContextGetter: logContextGetter,
+    featureFlags: await getFeatureFlagsClient()
+});
```
TBonnin, line 8: a single SlackService instance for server instead of creating one in 3-4 different places

packages/jobs/lib/clients.ts
```diff
@@ -12,6 +13,6 @@ export const bigQueryClient = await BigQueryClient.createInstance({
 export const orchestratorClient = new OrchestratorClient({ baseUrl: getOrchestratorUrl() });
 
 export const slackService = new SlackService({
-    orchestrator: new Orchestrator(orchestratorClient),
-    logContextGetter: logContextGetter
+    logContextGetter: logContextGetter,
+    featureFlags: await getFeatureFlagsClient()
```
TBonnin, line 17: added a flag to disable sending notifications if needed. Faster than having to make a commit, build and deploy

packages/shared/lib/services/notification/slack.service.ts
```diff
@@ -114,53 +121,40 @@ export class SlackService {
...
+        account: DBTeam;
+        environment: DBEnvironment;
+        connection: ConnectionJobs;
+        name: string;
+        type: string;
+        originalActivityLogId: string;
+        provider: string;
+    }) {
+        if (await this.isDisabled()) {
+            console.log('>>>>>>>>>>>>>>>>>>>Slack notifications are disabled');
```
bodinsamuel: not sure if >>> is intentional :D
maybe it could be more useful for debugging to add more info, e.g:
```suggestion
            logger.info('Disabled, not sending failure for ${name} ${type}');
```

bodinsamuel (approved):
🔥

packages/shared/lib/services/notification/slack.service.ts
TBonnin: print debugging!!!!!!!!!!!! 🙈

PR #3725, opened 2025-03-18
fix(logs): remove usage of kvstore

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2926/remove-redis

- Remove usage of kvstore
- eslint: group parent and sibling
`../file.js` and `./file.js` are quite similar imports and it's one less line break


packages/logs/lib/models/logContextGetter.ts
```diff
@@ -1,10 +1,8 @@
-import { getKVStore } from '@nangohq/kvstore';
```
TBonnin, line 1: should it be also removed from package.json?

packages/logs/lib/models/logContextGetter.ts
```diff
@@ -33,9 +31,7 @@ export const logContextGetter = {
         try {
             if (envs.NANGO_LOGS_ENABLED && !options?.dryRun) {
                 // TODO: remove this after full deploy
-                const res = await createOperation(msg);
-                const store = await getKVStore();
-                await store.set(`es:operation:${msg.id}:indexName`, res.index, { ttlInMs: 5 * 60 * 1000 });
```
TBonnin, line 38: 🪓 👌

TBonnin (approved):

packages/logs/lib/models/logContextGetter.ts
bodinsamuel: true

PR #3729, opened 2025-03-18
fix(credentials): split JWT from connection

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2933/jwt-client

- Split JWT from connection.service.ts
Starting a new initiative to split auth-specific stuff into their own file to limit file size. I picked one randomly, bad luck it's currently not working as intended.

- Introduce `NangoInternalError`
It's the same superset as `NangoError` except it doesn't enforce the messaging so it's fake retro-compatible and we can remove some of them gradually without breaking anything.


packages/shared/lib/auth/jwt.ts
```diff
@@ -0,0 +1,138 @@
+import axios from 'axios';
```
bodinsamuel, line 1: This code was imported from connection.service.ts

packages/shared/lib/utils/error.ts
```diff
@@ -1,17 +1,33 @@
...
+    public payload: Record<string, unknown> = {};
+
+    constructor(type: string, options?: { cause?: unknown }) {
+        super(type, options);
+        this.type = type;
+    }
+
+    public setPayload(payload: any) {
+        this.payload = payload;
+    }
```
TBonnin, line 21: should the payload be a constructor param?

TBonnin (approved):

packages/shared/lib/utils/error.ts
bodinsamuel: rn I haven't found the need for it, it's just there for compatibility with NangoError. If we need to this could passed in the constructor yes

PR #3733, opened 2025-03-19
feat(integrations): add support for figma scim

hassan254-prog:

- Add support for figma scim


packages/providers/providers.yaml
```diff
@@ -3226,6 +3226,46 @@ figma:
...
+    display_name: Figma (SCIM)
+    categories:
+        - design
+        - productivity
+    auth_mode: API_KEY
+    proxy:
+        base_url: https://${connectionConfig.host}/scim
+        headers:
+            authorization: Bearer ${apiKey}
+        verification:
```
hassan254-prog, line 3264: @khaliqgant, this has been added untested. However, based on the documentation, the API key is not scoped, so we can enforce verification here.

bodinsamuel (approved):

PR #3734, opened 2025-03-19
fix(refresh): short-circuit if exhausted, trycatch hooks

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2877/shouldrefresh-stop-refreshing-based-on-max-tries

- Short-circuit if exhausted
We still return the connection but we won't attempt new refresh unless it's been manually forced

- try/catch hooks with no consequence if failing
Hooks can fail but they are not really part of the critical process imo, let me know if you disagree


packages/server/lib/hooks/hooks.ts
```diff
@@ -187,17 +190,21 @@ export const connectionRefreshSuccess = async ({
...
-        originalActivityLogId: null,
-        provider: config.provider
-    });
+        await slackService.removeFailingConnection({
+            connection,
+            name: connection.connection_id,
+            type: 'auth',
+            originalActivityLogId: null,
+            provider: config.provider
+        });
```
TBonnin, line 208: I agree that `auth.clear` and `slack.removeFailingConnection` are secondary. Should they have their own try catch though because here if auth.clear fails the slack notif will not be attempted

packages/server/lib/hooks/hooks.ts
```diff
@@ -219,37 +226,41 @@ export const connectionRefreshFailed = async ({
...
+            account,
+            environment,
+            connection,
+            name: connection.connection_id,
+            type: 'auth',
+            originalActivityLogId: logCtx.id,
+            provider: config.provider
+        });
+    } catch (err) {
+        report(new Error('refresh_failed_hook_failed', { cause: err }), { id: connection.id });
```
TBonnin, line 269: same question about the single try catch

packages/shared/lib/services/connections/credentials/refresh.ts
```diff
@@ -78,6 +78,14 @@ export async function refreshOrTestCredentials(props: RefreshProps): Promise<Res
...
         span.setTag('connectionId', props.connection.connection_id).setTag('authType', props.connection.credentials.type);
 
+        // TODO: remove this when cron is using other columns
+        await connectionService.updateLastFetched(props.connection.id);
+
+        // short-circuit if we know the refresh will fail
+        if (props.connection.refresh_exhausted && !props.instantRefresh) {
+            return Ok(props.connection);
+        }
+
```
TBonnin, line 89: would it make more sense to have this logic in `shouldRefreshCredentials`?

packages/server/lib/hooks/hooks.ts
bodinsamuel: true

packages/shared/lib/services/connections/credentials/refresh.ts
bodinsamuel: it's a global flag, shouldRefresh only applies to refreshable connections whereas this block refresh and test.

packages/server/lib/hooks/hooks.ts
bodinsamuel: changed

packages/server/lib/hooks/hooks.ts
bodinsamuel: also changed

TBonnin (approved):

PR #3735, opened 2025-03-19
feat(verification): add verification script for aws-iam

hassan254-prog:

- Add verification script for AWS-IAM
- Return the `connection` instead of only` credentials` in the `credentials-verification` .


packages/server/lib/hooks/connection/providers/aws-iam/credentials-verification.ts
```diff
@@ -0,0 +1,80 @@
...
+        const { credentials, provider_config_key, connection_config } = nango.getConnection();
+
+        const { username, password } = credentials as { username: string; password: string };
+
+        const requestParams: AWSIAMRequestParams = {
+            method: 'GET',
+            service: 'iam',
+            path: '/',
+            params: {
+                Action: 'ListUsers',
```
bodinsamuel: are we are sure all keys can access this?

packages/server/lib/hooks/connection/providers/aws-iam/credentials-verification.ts
```diff
@@ -0,0 +1,80 @@
...
+            requestParams.service,
+            requestParams.path,
+            queryParams,
+            username,
+            password,
+            connection_config['region']
+        );
+
+        await nango.proxy<ErrorResponse | { ListUsersResponse: ListUsersResponse }>({
+            baseUrlOverride: 'https://iam.amazonaws.com',
```
bodinsamuel: it seems to be the default value already

packages/server/lib/hooks/connection/providers/aws-iam/credentials-verification.ts
```diff
@@ -0,0 +1,80 @@
...
+
+function getAWSAuthHeader(
+    method: string,
+    service: string,
+    path: string,
+    querystring: string,
+    accessKeyId: string,
+    secretAccessKey: string,
+    region: string
+): AWSAuthHeader {
```
bodinsamuel: Can you make that an object? usually we try to make params an object when there are more than 3

bodinsamuel (approved):
Not tested but looks good

packages/server/lib/hooks/connection/providers/aws-iam/credentials-verification.ts
hassan254-prog: thanks 🎸

packages/server/lib/hooks/connection/providers/aws-iam/credentials-verification.ts
hassan254-prog: 🤦‍♂️

packages/server/lib/hooks/connection/providers/aws-iam/credentials-verification.ts
hassan254-prog: Opted for a safer endpoint, `GetCallerIdentity`, which verifies the identity of the token with the `STS` service. Tested this as well ✅.

PR #3736, opened 2025-03-19
fix(refresh): only update attempts once per day

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2948/increment-attempts-once-per-day-with-max-3

After discussing with Bastien we changed strategy. We increment only once per day, to avoid invalidating connections too soon when providers are down.

- Only update attempts once per day
- Mark as exhausted after 3 tries
- Fix bad c/c for `last_refresh_failure`


packages/shared/lib/services/connection.service.ts
```diff
@@ -480,15 +480,28 @@ class ConnectionService {
...
+
+        // Only increment once per day to avoid burst failed refresh invalidating a connection (e.g: provider being down)
+        if (
+            lastRefreshFailure &&
+            (lastRefreshFailure.getFullYear() < now.getFullYear() ||
+                lastRefreshFailure.getMonth() < now.getMonth() ||
+                lastRefreshFailure.getDate() < now.getDate())
+        ) {
+            attempt += 1;
+        }
```
TBonnin, line 500: Do i understand correctly that attempt will stay at 0 for the entire first day it is failing?
Does it also mean if it fail first at 23:59 and then again at 00:01, attempt will be 1?

packages/shared/lib/services/connections/utils.ts
```diff
@@ -6,7 +6,7 @@ export const DEFAULT_EXPIRES_AT_MS = ms('1day');
 export const DEFAULT_OAUTHCC_EXPIRES_AT_MS = ms('55minutes'); // This ensures we have an expiresAt value
 export const DEFAULT_BILL_EXPIRES_AT_MS = ms('35minutes'); //This ensures we have an expireAt value for Bill
 export const DEFAULT_INFINITE_EXPIRES_AT_MS = ms('99years');
-export const MAX_FAILED_REFRESH = 30;
+export const MAX_FAILED_REFRESH = 3;
```
TBonnin: see my comment above. If we want to ensure at least 3 days we should have 4 here to make sure 3 full days have elapsed. The first one might have been partial

TBonnin: I would also rename this env var and even the column in the db since it doesn't represent the number of failures anymore but the number of days it has been failing

packages/shared/lib/services/connection.service.ts
bodinsamuel: 1/ yes, but maybe it should be 1 you are right
2/ yes that is a limitation since we don't store all refresh date, but since it will stay N for the next 24 hours it's okay-ish imo

packages/shared/lib/services/connections/utils.ts
bodinsamuel: renaming a column used in prod is a bit dangerous tbh, but I'll rename the env var for sure 👌🏻

TBonnin (approved):

PR #3737, opened 2025-03-19
feat(verification): add verification method for ghost admin

hassan254-prog:

- Add verification method for ghost admin
- Remove docs section that mentioned how to obtain and use the token since we can now build proxy headers with authentication.

### DIFF
--- a/packages/providers/providers.yaml
+++ b/packages/providers/providers.yaml
@@ -3840,7 +3840,12 @@ ghost-admin:
         headers:
             accept: application/json
             accept-version: ${connectionConfig.version}
+            authorization: Ghost ${accessToken}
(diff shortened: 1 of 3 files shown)


packages/providers/providers.yaml
```diff
@@ -3840,7 +3840,12 @@ ghost-admin:
         headers:
             accept: application/json
             accept-version: ${connectionConfig.version}
+            authorization: Ghost ${accessToken}
         base_url: https://${connectionConfig.adminDomain}/ghost/api/admin/
+        verification:
```
hassan254-prog, line 3845: @bodinsamuel Maybe this will help, I tested the verification method, and it works with the `JWT` that Nango generates.

bodinsamuel (approved):

PR #3738, opened 2025-03-19
fix(logs): do not fallback on unknown operation id

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2973/fix-unknown-as-activitylogid-in-jobs

- do not fallback on unknown operation id
It was a problem because we were trying to parse an invalid `operationId`, plus we should just be aware that it can fail before having a log context

- make activityLogId required
Not sure why it was still optional, probably a left over


TBonnin (approved):
👍

PR #3750, opened 2025-03-19
feat(integrations): add support for Recruitee

albertocubeddu:
Even replacing the SVG with an existing integration i was not able to see the image: 
image

image
The integration seems to work, what is the best way to test it out from localhost? (e.g. calling the proxy?) 

Thanks in advance.


packages/providers/providers.yaml
```diff
@@ -7336,6 +7336,30 @@ recharge:
             pattern: '^[a-zA-Z0-9_]+$'
             doc_section: '#step-1-generating-your-api-token'
 
+recruitee:
+    display_name: Recruitee
+    categories:
+        - ats
+    auth_mode: API_KEY
+    proxy:
+        base_url: https://api.recruitee.com/c/${connectionConfig.company_id}
```
hassan254-prog, line 7345: would be good to have `company_id` as camelCase.

packages/providers/providers.yaml
```diff
@@ -7336,6 +7336,30 @@ recharge:
...
+        headers:
+            authorization: Bearer ${apiKey}
+        verification:
+            method: GET
+            endpoints:
+                - /candidates
+    docs: https://docs.recruitee.com
```
hassan254-prog, line 7352: The docs page should link to the internal Nango documentation. Could you also include the Connect-UI documentation? You can follow this example: [here](https://github.com/NangoHQ/nango/pull/3752). Would also be nice to include the `pattern`, `example` and `docs_section` for both `credentials` and `connection_config`

hassan254-prog (commented):
Thanks @albertocubeddu for your contribution, small comments inline but everything else looks great 🎸 , nice catch on adding the `proxy.verification` method. 
Just curious is the api key scoped?

hassan254-prog:
@albertocubeddu , after adding a new image, you will need to rebuild your webapp. 
One way to test if the API Key is correct, is by adding the `proxy.verification` method which you already have. This will send a proxy call to the endpoint specified with the apiKey you have supplied and incase the api key is incorrect, Nango will throw an incorrect credential error. The other way to test is by creating a custom integration which you can run against the connection and provider you just added. You can follow steps in [here](https://docs.nango.dev/guides/custom-integrations/overview) to achieve that.

PR #3751, opened 2025-03-20
docs: When there is a connect ui document, link to that for the setup guide

lordsarcastic:
1. Added a `docs_connect` property to integrations that have a setup guide but lacks the connect property
2. Implemented a pre-commit hook which automatically generates `docs` and `docs_connect` properties for integrations which already have a `connect.mdx` documentation file in them.

## How `2` works:
The hook works by retrieving all integrations in the `docs-v2/integrations/all` folder that has a `connect.mdx` file. Next, it performs a diff of integrations from the `providers.yaml` file which already have either `docs` or `docs_connect` properties to determine which needs to be updated. It then makes the updates and writes the result back to the `providers.yaml` file.

### Known issues
If an integration in the `providers.yaml` file has an alias, the hook will inadvertently copy properties from the aliased integration into the said integration. This is because the `loadProvidersYaml` function which loads providers also denormalizes aliases before it can be processed. However, this doesn't affect the integration since it is an explicit form of the `alias` feature of the `providers.yaml` file

Ticket link: https://linear.app/nango/issue/EXT-556/when-there-is-a-connect-ui-document-link-to-that-for-the-setup-guide


TBonnin (commented):
have you tested running Nango with the resulting `providers.yaml`. Some formatting is different and I wonder if it could cause problems

packages/providers/providers.yaml
```diff
@@ -114,13 +110,13 @@ adobe-umapi:
...
             description: ''
             automated: true
-
 adyen:
     display_name: Adyen
     categories:
         - payment
     auth_mode: OAUTH2
-    authorization_url: https://ca-${connectionConfig.environment}.adyen.com/ca/ca/oauth/connect.shtml
+    authorization_url: >-
```
TBonnin: not a huge fan of this formatting to be honest, especially when there is no new lines

bodinsamuel:
I'm not sure I understand the issue it's fixing here, is it just adding a `docs_connect` when you create the file first? 

We have `scripts/validation/providers/validate.ts` that make sure providers.yaml is correct, it could be done there without re-generating the file.

lordsarcastic:
> I'm not sure I understand the issue it's fixing here, is it just adding a `docs_connect` when you create the file first?
> 
> We have `scripts/validation/providers/validate.ts` that make sure providers.yaml is correct, it could be done there without re-generating the file.

Unlike the validate script which only runs validation, this ensures that we don't have to manually update the `docs` and `docs_connect` property which is the source of some tasks I've had to work on. This automates that process.

TBonnin (commented):
I am not a huge fan of mutating providers.yaml in-place and requiring an additional build step. 
I think the validation step is where we should verify the providers.yaml is correct.
Feel free to write your own automation to add the missing attributes  but I would keep the process simple and not commit and add an additional step

lordsarcastic:
So a post commit hook? or just a separate command that can be run on demand? @TBonnin

TBonnin:
My opinion would be no new command at all. How often do we need to modify providers.yaml?
Can we add a check in the `validate` script and keep adding the missing docs_connect in the providers.yaml manually. 
Just my opinion though. What do people think? cc @NangoHQ/engineers @NangoHQ/integration-engineers

khaliqgant:
> My opinion would be no new command at all. How often do we need to modify providers.yaml? Can we add a check in the `validate` script and keep adding the missing docs_connect in the providers.yaml manually. Just my opinion though. What do people think? cc @NangoHQ/engineers @NangoHQ/integration-engineers

Also agree that a check in the validate script would suffice here.

docs-v2/integrations/all/docusign/connect.mdx
bodinsamuel, line 1: leftover?

packages/providers/providers.yaml
```diff
@@ -58,41 +58,6 @@ accelo:
             suffix: .api.accelo.com
```
bodinsamuel: merge issue

bodinsamuel (commented):
Looks good, you have some leftover to clean 👍🏻

docs-v2/integrations/all/docusign/connect.mdx
bodinsamuel, line 1: Still not deleted

bodinsamuel (approved):
👌🏻

PR #3753, opened 2025-03-20
fix(CLI): nango.proxy should throw if request fails

TBonnin:
A dry-run execution is currently continuing if nango.proxy fails which is not consistant with the behavior when run by the runner

This behavior might have changed when we refactored the runner-sdk at the end of January.


khaliqgant (approved):
Nice, thanks!

PR #3754, opened 2025-03-20
fix: shouldRefreshCredentials to return false if credentials_expires_at is set

TBonnin:
in case of credentials without `expires_at` (ex: stripe) we fallback to the `credentials_expires_at` connection attribute to determine if a connection requires to be refreshed

### DIFF
--- a/packages/shared/lib/services/connections/credentials/refresh.ts
+++ b/packages/shared/lib/services/connections/credentials/refresh.ts
@@ -430,7 +430,8 @@ export async function shouldRefreshCredentials({
             return { should: false, reason: 'fresh_introspected_token' };
         }
 
-        if (credentials.expires_at && !isTokenExpired(credentials.expires_at, provider.token_expiration_buffer || REFRESH_MARGIN_S)) {
+        const expires_at = credentials.expires_at || connection.credentials_expires_at;
+        if (expires_at && !isTokenExpired(expires_at, provider.token_expiration_buffer || REFRESH_MARGIN_S)) {
             return { should: false, reason: 'fresh' };
         }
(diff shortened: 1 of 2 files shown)


packages/shared/lib/services/connections/credentials/refresh.ts
```diff
@@ -430,7 +430,8 @@ export async function shouldRefreshCredentials({
             return { should: false, reason: 'fresh_introspected_token' };
         }
 
-        if (credentials.expires_at && !isTokenExpired(credentials.expires_at, provider.token_expiration_buffer || REFRESH_MARGIN_S)) {
+        const expires_at = credentials.expires_at || connection.credentials_expires_at;
+        if (expires_at && !isTokenExpired(expires_at, provider.token_expiration_buffer || REFRESH_MARGIN_S)) {
             return { should: false, reason: 'fresh' };
         }
```
bodinsamuel: looking at the previous code the actual fix might be something like this:

```suggestion
        if (!credentials.expires_at) {
            return { should: false, reason: 'no_expires_at' };
        } else if (!isTokenExpired(expires_at, provider.token_expiration_buffer || REFRESH_MARGIN_S)) {
            return { should: false, reason: 'fresh' };
        }
```

https://github.com/NangoHQ/nango/pull/3718/files#diff-ba6c55003de91652744d8328507e85fd85ee4d59975394f9e28faccf2557ab19

TBonnin: Ok. I am gonna push this fix to bring back the previous behavior and unblock the customer. Thank you for your help

PR #3755, opened 2025-03-21
fix(connect-ui): do not detect closed auth-window

hassan254-prog:

- At times, the `modal.window` property is null making also `closed` flag of the modal to always be true for some providers i.e calendly, zendesk, this in turn makes it hard to detect closed auth window. 

- This can then cause double connection creation, since the `detectClosedAuthWindow` will always be true causing the user to "Try Again". This flag only works if the user has a pre-existing session with the provider and the 500ms delay doesn't `detectClosedAuthWindow`. But if the user is redirected to the login page, due to a non pre-existing session, the delay elapses, the error is thrown and once the user finishes signing in, the "Try Again" button is displayed and when they do so since they already have a session created, another connection will automatically be created.
- To avoid this, we can just have the user close the nango connect modal if they decide to also close the provider modal for now.


bodinsamuel:
You'll need to change doc and ts comments, its said to be true by default

TBonnin (approved):

PR #3756, opened 2025-03-21
fix(runner): do not exit on unhandledRejection

TBonnin:
we don't exit on uncaughtException but we do on unhandledRejection 
Making it consistant by only logging and not exiting the runner because one script throwing an exception/rejection should not crash the entire runner

### DIFF
--- a/packages/runner/lib/app.ts
+++ b/packages/runner/lib/app.ts
@@ -40,9 +40,8 @@ try {
     });
 
     process.on('unhandledRejection', (reason) => {
-        logger.error(`${id} Received unhandledRejection...`, reason);
-        process.exitCode = 1;
-        close();
+        logger.error(`${id} Received uncaughtException...`, reason);
+        // not closing on purpose
     });
 
     process.on('uncaughtException', (e) => {


packages/runner/lib/app.ts
```diff
@@ -40,9 +40,8 @@ try {
     });
 
     process.on('unhandledRejection', (reason) => {
-        logger.error(`${id} Received unhandledRejection...`, reason);
-        process.exitCode = 1;
-        close();
+        logger.error(`${id} Received uncaughtException...`, reason);
+        // not closing on purpose
```
TBonnin, line 44: doing the same thing as for `uncaughtException` below

khaliqgant (approved):

packages/runner/lib/app.ts
bodinsamuel: it's very unfortunate, we keep removing this close but usually it results in unstable code. 
In that very case could we catch those in `exec()` maybe in the vm code?

TBonnin: The problem is that misusing promises in scripts bypasses the exec try/catch and was killing the entire runner process. I would love to handle those promise rejections but I am not sure how at the moment

PR #3758, opened 2025-03-21
fix: check record is not null in removeMetadata

TBonnin:
One can possibly pass a null record to batchSave. (I need to come up with a 'ts types are lies meme')

It currently fails with a `cannot access _nango_metadata in null`. This commit make sure result is defined first.

Passing a null record will fail later anyway but the error message is going to be a bit more comprehensible (ie: `invalid_type: Expected object, received null`)

### DIFF
--- a/packages/runner-sdk/lib/sync.ts
+++ b/packages/runner-sdk/lib/sync.ts
@@ -77,7 +77,7 @@ export abstract class NangoSyncBase extends NangoActionBase {
 
     protected removeMetadata<T extends object>(results: T[]) {
         return results.map((result) => {
-            if ('_nango_metadata' in result) {
+            if (result && '_nango_metadata' in result) {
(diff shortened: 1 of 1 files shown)


packages/runner-sdk/lib/sync.ts
```diff
@@ -77,7 +77,7 @@ export abstract class NangoSyncBase extends NangoActionBase {
 
     protected removeMetadata<T extends object>(results: T[]) {
         return results.map((result) => {
```
bodinsamuel, line 82: results could also be "not an array"

bodinsamuel (approved):
Happy to have some asserts in place for the sdk

packages/runner-sdk/lib/sync.ts
TBonnin: good point

PR #3759, opened 2025-03-24
fix(types): use nango standard config from types

khaliqgant:
Fixes a small mismatch between the types. Uses the types package to consolidate.

### DIFF
--- a/packages/node-client/lib/types.ts
+++ b/packages/node-client/lib/types.ts
@@ -42,7 +42,9 @@ import type {
     PostConnectSessions,
     PostPublicConnectSessionsReconnect,
     GetPublicRecords,
-    UserProvidedProxyConfiguration
+    UserProvidedProxyConfiguration,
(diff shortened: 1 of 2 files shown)


packages/node-client/lib/types.ts
```diff
@@ -212,14 +212,6 @@ export interface UpdateSyncFrequencyResponse {
     frequency: string;
 }
 
-export interface StandardNangoConfig {
```
bodinsamuel, line 215: can you re-export StandardNangoConfig from `@nangohq/types` to avoid breaking change?
Same for NangoSyncConfig that can be deleted and re-exported

bodinsamuel (approved):

packages/node-client/lib/types.ts
khaliqgant: True, updated

bodinsamuel (approved):

PR #3761, opened 2025-03-24
fix(cli): deduplicate records when using batchSave in CLI

hassan254-prog:

- Deduplicate records when using `batchSave` in CLI 

- Add more regex patterns for shopify api key.


packages/cli/lib/services/sdk.ts
```diff
@@ -128,7 +128,21 @@ export class NangoSyncCLI extends NangoSyncBase {
...
+
+        for (let i = results.length - 1; i >= 0; i--) {
+            const record = results[i] as T & { id: string | number };
+            if (!seenIds.has(record.id)) {
+                seenIds.add(record.id);
+                deduplicatedResults.unshift(record);
+            } else {
+                console.warn(`batchSave detected duplicate records for ID: ${record.id}. Keeping the last occurrence.`);
+            }
+        }
```
bodinsamuel, line 143: why not just use a map? it will deduplicate automatically 
```ts
const records = new Map();
for ( const record in results) 
  records.set(record.id, record);

deduplicatedResults = Array.from(records.values);
```
but you can't log warnings, so as you wish

bodinsamuel (approved):

packages/cli/lib/services/sdk.ts
hassan254-prog: Using `map` would work for deduplication, but it will not preserve the order. It might be good to keep the order, especially when logging these results in the CLI, since the records could be ordered based on a specific query parameter.

PR #3762, opened 2025-03-24
fix(ui): show correct message for credentials error

bodinsamuel:
## Changes 

Fixes https://linear.app/nango/issue/NAN-2949/stop-refreshing-credentials-that-are-repeatedly-failing-beyond-the#comment-4e0b3541

- show correct message for credentials error
Dissociate test vs refresh

- Do not "leak" credentials_iv in private API
Not critical at all but it's cleaner

- Fix testing credentials would log twice the error

### DIFF
--- a/packages/server/lib/hooks/hooks.ts
+++ b/packages/server/lib/hooks/hooks.ts
@@ -1,43 +1,46 @@
-import type { Span } from 'dd-trace';
+import tracer from 'dd-trace';
+
 import {
+    AnalyticsTypes,
     CONNECTIONS_WITH_SCRIPTS_CAP_LIMIT,
     NangoError,
-    getSyncConfigsWithConnections,
+    ProxyRequest,
     analytics,
     errorNotificationService,
     externalWebhookService,
-    AnalyticsTypes,
-    syncManager,
     getProxyConfiguration,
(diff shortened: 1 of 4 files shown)


packages/server/lib/hooks/hooks.ts
```diff
@@ -353,15 +356,8 @@ export async function credentialsTest({
...
-            logs.push({ type: 'log', level: 'error', message: `Failed verification for endpoint: ${endpoint}`, createdAt: new Date().toISOString() });
-        } catch (err) {
-            logs.push({
-                type: 'log',
-                level: 'error',
-                message: `Error testing endpoint: ${endpoint},  ${stringifyError(err)}`,
-                createdAt: new Date().toISOString()
-            });
+        } catch {
+            // Already covered
```
TBonnin, line 360: not sure I follow

TBonnin (approved):

packages/server/lib/hooks/hooks.ts
bodinsamuel: historically the proxy calls were throwing the last error allowing us to log the last error, but we already log those errors, it's just throwing to allow the SDK to bubble up the request but it's useless in this case

PR #3764, opened 2025-03-24
fix(persist): only report the size and count of modified records

TBonnin:
So far we were counting all the records that were being batchSaved, even when the records was already stored in the db.

Note: modified = added + updated + deleted

### DIFF
--- a/packages/persist/lib/records.ts
+++ b/packages/persist/lib/records.ts
@@ -136,8 +136,9 @@ export async function persistRecords({
         metrics.increment(metrics.Types.PERSIST_RECORDS_SIZE_IN_BYTES, recordsSizeInBytes, { accountId });
 
         span.addTags({
-            'records.count': allModifiedKeys.size,
-            'records.sizeInBytes': recordsSizeInBytes
+            'records.in.count': records.length,
+            'records.modified.count': allModifiedKeys.size,
+            'records.modified.sizeInBytes': recordsSizeInBytes
         });
         span.finish();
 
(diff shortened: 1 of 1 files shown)


packages/persist/lib/records.ts
```diff
@@ -119,17 +117,30 @@ export async function persistRecords({
             void logCtx.error(`Found duplicate key '${nonUniqueKey}' for model ${baseModel}. The record was ignored.`);
         }
 
-        const total = summary.addedKeys.length + summary.updatedKeys.length + (summary.deletedKeys?.length || 0);
-        void logCtx.info(`Successfully batched ${total} record${total > 1 ? 's' : ''}`, {
+        await updateSyncJobResult(syncJobId, updatedResults, baseModel);
```
TBonnin, line 120: unrelated but make sense to move th log AFTER the job result is updated

packages/persist/lib/records.ts
```diff
@@ -119,17 +117,30 @@ export async function persistRecords({
...
+                return acc + Buffer.byteLength(JSON.stringify(record), 'utf8');
+            }
+            return acc;
+        }, 0);
+
+        metrics.increment(metrics.Types.PERSIST_RECORDS_COUNT, allModifiedKeys.size, { accountId });
         metrics.increment(metrics.Types.PERSIST_RECORDS_SIZE_IN_BYTES, recordsSizeInBytes, { accountId });
 
+        span.addTags({
+            'records.count': allModifiedKeys.size,
```
bodinsamuel: not sure I agree with this, or at least I would log both count but having 0 because nothing was modified is definitely not helping for debugging

bodinsamuel (approved):

packages/persist/lib/records.ts
TBonnin: Good idea. let's have 2 tags:`records.modified.count` and `records.in.count`?

PR #3765, opened 2025-03-24
fix(logs): buffer context, log oauth2 http calls

bodinsamuel:
## Changes

Contributes to https://linear.app/nango/issue/NAN-2975/investigate-failed-refresh

- Add log context buffer class
I got annoyed at buffering logs in an array and passing up the array. We now have a buffer class that acts like a regular logCtx so they are used without knowing if it's a buffer or not.

- Pass log buffer to oauth2 client
I was trying to debug something for a customer but we were not logging oauth2 calls. simple-oauth2 is hiding a lot of stuff, it's really annoying


khaliqgant:
> simple-oauth2 is hiding a lot of stuff, it's really annoying

I agree. I was tempted to rip it out more times than I can count but ended up not. Maybe as a start worth forking and making some tweaks? It is very much online with our offering

bodinsamuel:
> I agree. I was tempted to rip it out more times than I can count but ended up not. Maybe as a start worth forking and making some tweaks? It is very much online with our offering

We talked about that with Bastien a long time ago, but I took a long dive in the code and the whole library is probably 2-3 functions. The rest is plumbing and error handling.

packages/logs/lib/transport.ts
```diff
@@ -0,0 +1,56 @@
...
+    }
+}
+
+export class BufferTransport implements LogTransportAbstract {
+    buffer: MessageRowInsert[] = [];
+    log(msg: MessageRowInsert): boolean {
+        this.buffer.push(msg);
+        return true;
+    }
+}
```
TBonnin, line 74: 🫶

packages/shared/lib/services/connections/credentials/refresh.ts
```diff
@@ -182,6 +186,7 @@ async function refreshCredentials(
                 connection: { id: oldConnection.id, name: oldConnection.connection_id }
             }
         );
+        void flushLogsBuffer((logsBuffer.transport as BufferTransport).buffer, logCtx);
```
TBonnin: should flushing be abstracted and happen automatically on` logCtx.failed/success` if transport is Buffer. With maybe a `forceFlush` method in `LogContextStateless` to manually flush buffer if needed

TBonnin (commented):
> We now have a buffer class that acts like a regular logCtx so they used without knowing if it's a buffer or not

genius

packages/shared/lib/services/connections/credentials/refresh.ts
bodinsamuel: we would need to attach/pass the buffer to the actual logCtx that contains the proper transport.
Maybe we could have a `logCtx.merge(logCtx)` and the transport decide on what to do: 
- if buffer, it saves more buffer
- if es, it flushes everything
Wdyt?

TBonnin: I am not sure why we create two logCtx in this case. Can we not reuse the buffer one?

bodinsamuel: rn the buffer doesn't have the capability to create an operation, just for simplicity's sake, it's just buffering the logs but we could improve for sure

TBonnin: I see. a merge `logCtx.merge(logsBuffer)` could be a good idea indeed. this or passing a buffer to `logContextGetter.create`

bodinsamuel: Changed in [450e17a](https://github.com/NangoHQ/nango/pull/3765/commits/450e17a7c5a539bf0937cd5c377c214e3a93f359)

TBonnin (approved):

PR #3770, opened 2025-03-25
chore: sync total active time

TBonnin:
for more context: https://linear.app/nango/issue/NAN-2958/metric-sync-schedule-active-time

We want to track how long syncs are being active (aka: started) 
In order to get an approximation, we are pushing a metric per account for every sync being started. 
The value is equal to the frequency of the sync. By summing up this metric we should then be able to know the total active sync time for an account

### DIFF
--- a/packages/jobs/lib/execution/sync.ts
+++ b/packages/jobs/lib/execution/sync.ts
@@ -17,7 +17,9 @@ import {
     externalWebhookService,
     getApiUrl,
     getEndUserByConnectionId,
+    getFrequencyMs,
     getLastSyncDate,
+    getSync,
     getSyncConfigRaw,
     getSyncJobByRunId,
     setLastSyncDate,
@@ -154,6 +156,28 @@ export async function startSync(task: TaskSync, startScriptFn = startScript): Pr
 
         if (task.debug) {
             void logCtx.debug(`Last sync date is ${lastSyncDate?.toISOString()}`);
+        }
+
+        // tracking active schedule time
+        // by pushing the frequency of the sync as a metric
+        // that can be sumed up to get the total active
+        // time of the sync for a given period/account
+        const sync = await getSync({ connectionId: task.connection.id, name: task.syncName, variant: task.syncVariant });
+        const frequencyStr = sync?.frequency || syncConfig.runs;
+        if (frequencyStr) {
+            const frequencyMs = getFrequencyMs(frequencyStr);
+            if (frequencyMs.isErr()) {
+                logger.error(`Failed to get frequency for sync ${task.syncId}`, frequencyMs.error);
+            } else {
+                // we track 2 metrics
+                // high frequency if the frequency is less than 1 hour
+                // low frequency if the frequency is >= 1 hour
+                if (frequencyMs.value < 3_600_000) {
+                    metrics.gauge(metrics.Types.SYNC_HIGH_FREQUENCY_MS, frequencyMs.value, { accountId: team.id });
+                } else {
+                    metrics.gauge(metrics.Types.SYNC_LOW_FREQUENCY_MS, frequencyMs.value, { accountId: team.id });
+                }
+            }
         }
 
         metrics.increment(metrics.Types.SYNC_EXECUTION, 1, { accountId: team.id });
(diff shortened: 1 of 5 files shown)


packages/jobs/lib/execution/sync.ts
```diff
@@ -156,6 +158,28 @@ export async function startSync(task: TaskSync, startScriptFn = startScript): Pr
             void logCtx.debug(`Last sync date is ${lastSyncDate?.toISOString()}`);
         }
 
+        // tracking active schedule time
+        // by pushing the frequency of the sync as a metric
+        // that can be sumed up to get the total active
+        // time of the sync for a given period/account
+        const sync = await getSync({ connectionId: task.connection.id, name: task.syncName, variant: task.syncVariant });
```
TBonnin, line 165: not a huge fan of making an extra db query. if you have a better solution lmk

packages/shared/lib/services/sync/utils.ts
```diff
@@ -0,0 +1,31 @@
+import type { Result } from '@nangohq/utils';
```
TBonnin, line 1: not sure this is the best place for this new file but I don't think there is a good place for it. wdyt?

packages/jobs/lib/execution/sync.ts
bodinsamuel: can you make the orchestrator return the frequency in task instead?

packages/shared/lib/services/sync/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+    ]);
+    const interval = runsMap.get(runs) || runs.replace('every ', '');
+
+    const intervalMs = ms(interval as StringValue);
+    if (!intervalMs) {
+        const error = new NangoError('sync_interval_invalid');
+        return Err(error);
+    }
+
+    if (intervalMs < ms('30s')) {
```
bodinsamuel, line 25: not this PR but this line and the map could be constant instead of reassigning at each call

bodinsamuel (approved):
Not sure I understood everything from the task but wouldn't sending `endedAt - startedAt` be more accurate? 
If a run has a frequency of 1h but takes a second to run each time it's far from being accurately represented and the opposite is also true

TBonnin:
> Not sure I understood everything from the task but wouldn't sending `endedAt - startedAt` be more accurate? If a run has a frequency of 1h but takes a second to run each time it's far from being accurately represented and the opposite is also true

The metric is about knowing how long a sync has been started for, not how long it takes to execute. We could store events when syncs are started or paused and process those events to know how long syncs have been enabled for but it is a more complicated pipeline. Instead the idea was to increment a metric each time a sync is executed to get a good approximation.
Ex: I start a sync today with frequency = 1h. Every time it runs, we increment the metric by 1h. If we sum the metric after 10h it will give us 10h 🪄 (+/- one interval)

However I am now realizing that there is some big caveats: 
- manual execution.  when customers runs a scheduled sync manually it will double increment the metric
- sync that takes longer than the frequency. Ex: 1h interval but takes 3h -> we'll increment only by 1h every 3h :(

I am not sure this approach actually works @bastienbeurier

bastienbeurier:
> * manual execution.  when customers runs a scheduled sync manually it will double increment the metric

We should filter these out. It should be possible, no? 

> * sync that takes longer than the frequency. Ex: 1h interval but takes 3h -> we'll increment only by 1h every 3h :(

This is fine for me, since it's an edge case and lowers the bill if anything.

bastienbeurier:
> > * manual execution.  when customers runs a scheduled sync manually it will double increment the metric
> 
> We should filter these out. It should be possible, no?
> 
> > * sync that takes longer than the frequency. Ex: 1h interval but takes 3h -> we'll increment only by 1h every 3h :(
> 
> This is fine for me, since it's an edge case and lowers the bill if anything.

After thinking more about this, it's not good to have an approximation for billing. Let's chat more about how to handle this and if we can approximate this with another metric that simpler to track.

TBonnin:
> After thinking more about this, it's not good to have an approximation for billing. Let's chat more about how to handle this and if we can approximate this with another metric that simpler to track.

I have already mentioned it in Linear, metrics in datadog is ok to understand usage and detect abnormal behaviors. I don't think basing billing on datadog metrics is a good idea

TBonnin:
closing for now until we can come up with a good solution

PR #3771, opened 2025-03-25
fix: make Slack notifications aware of sync variants

TBonnin:
Slack notifications are currently not aware of sync variants, which means that all sync execution for the base sync and its variant fall into the same entry in the slack_notifications table. This could lead to bugs where one variant fails triggering the push of a 'Failed sync' notif followed by another variant that is successful, marking the initial Slack notifications as resolved even though it is a different variants.
Fixing the bug by appending the variant to the sync name if it is not the base one

### DIFF
--- a/packages/shared/lib/services/notification/slack.service.ts
+++ b/packages/shared/lib/services/notification/slack.service.ts
@@ -193,7 +193,7 @@ export class SlackService {
      *      2) Add an activity log entry for the notification to the admin account
      *
      */
-    async reportResolution(
+    private async reportResolution(
         connection: ConnectionJobs,
         name: string,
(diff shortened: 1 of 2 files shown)


packages/shared/lib/services/notification/slack.service.ts
```diff
@@ -193,7 +193,7 @@ export class SlackService {
      *      2) Add an activity log entry for the notification to the admin account
      *
      */
-    async reportResolution(
+    private async reportResolution(
```
TBonnin, line 196: unrelated but all those methods are only used internally so making them private

bodinsamuel (approved):

PR #3772, opened 2025-03-25
fix: connection is already deleted when executing pre-connection-deletion

TBonnin:
we currently execute the pre-connection-deletion on-event script asynchronously which means the connection is likely already deleted when the script execute.

This commit adds a way to execute a on-event script either asynchronously (for post-connection-creation) or synchronously (for pre-connection-deletion)
The drawback of waiting for the script to finish executing is that when deleting via the dasbhoard UI or the API, the user must now wait

### DIFF
--- a/packages/orchestrator/lib/clients/client.ts
+++ b/packages/orchestrator/lib/clients/client.ts
@@ -162,7 +162,7 @@ export class OrchestratorClient {
         }
     }
 
-    private async execute(props: ExecuteProps): Promise<ExecuteReturn> {
+    private async immediateAndWait(props: ExecuteProps): Promise<ExecuteReturn> {
         const scheduleProps = {
             retry: { count: 0, max: 0 },
             timeoutSettingsInSecs: { createdToStarted: 30, startedToCompleted: 30, heartbeat: 60 },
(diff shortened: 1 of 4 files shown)


packages/orchestrator/lib/clients/client.ts
```diff
@@ -162,7 +162,7 @@ export class OrchestratorClient {
         }
     }
 
-    private async execute(props: ExecuteProps): Promise<ExecuteReturn> {
+    private async immediateAndWait(props: ExecuteProps): Promise<ExecuteReturn> {
```
TBonnin, line 165: unrelated but `execute` was a bit vague. We now have an `recurring`, `immediate` and `immediateAndWait` functions

packages/orchestrator/lib/clients/client.ts
```diff
@@ -270,7 +270,8 @@ export class OrchestratorClient {
                 type: 'on-event' as const
             }
         };
-        const res = await this.immediate(schedulingProps);
+
+        const res: Result<any, ClientError> = props.async ? await this.immediate(schedulingProps) : await this.immediateAndWait(schedulingProps);
```
TBonnin, line 274: we actually don't care about the result success value. Using `any` to please the typescript god

bodinsamuel (approved):
Why would they wait? We can still do that in background no?

TBonnin:
> Why would they wait? We can still do that in background no?

because we cannot delete the connection until the script finish executing since it likely needs the connection

khaliqgant:
Thanks for squeezing this in Thomas!

PR #3773, opened 2025-03-25
feat(integrations): add support for active campaign

hassan254-prog:

- Add support for active campaign


packages/providers/providers.yaml
```diff
@@ -58,6 +58,41 @@ accelo:
...
+            api-token: ${apiKey}
+        base_url: https://${connectionConfig.apiUrl}/api
+        verification:
+            method: GET
+            endpoints:
+                - /3/users/me
+    docs: https://docs.nango.dev/integrations/all/active-campaign
+    docs_connect: https://docs.nango.dev/integrations/all/active-campaign/connect
+    connection_config:
+        apiUrl:
```
bodinsamuel: usually we call that hostname because it's not a full URL. Same applies for the description and title

packages/providers/providers.yaml
```diff
@@ -58,6 +58,41 @@ accelo:
...
+active-campaign:
+    display_name: ActiveCampaign
+    categories:
+        - marketing
+        - communication
+    auth_mode: API_KEY
+    proxy:
+        headers:
+            api-token: ${apiKey}
+        base_url: https://${connectionConfig.apiUrl}/api
```
bodinsamuel: is it possible to have a custom domain or just a subdomain? 
if it's just a subdomain I would rename that to subdomain and use `https://${subdomain}.activehosted.com`

packages/providers/providers.yaml
hassan254-prog: thanks 🙏 , updated with the latest commit.

packages/providers/providers.yaml
hassan254-prog: No it is not possible. From the documentation, it actually lists `api-us1.com` as the base URL, but it also advises that this may change. It's recommended to always use the full Base URL from the user's "Developer" tab instead of assuming a fixed one.

bodinsamuel: make sense, thanks

bodinsamuel (approved):

PR #3776, opened 2025-03-26
fix(api): split routes

bodinsamuel:
## Changes

Contributes to https://linear.app/nango/issue/NAN-2116/webapp-use-vite-instead-of-webhookcreate-react-app

- Split routes and use more localized middlewares
While trying to use Vite and remove webpack proxy I realised some of the API headers were incorrect because everything is defined at the root level. So for example the first cors headers defined are always the one returned even if there is a more precise one. 
To fix that, we need to use `router.use('/pa/th', subRouter);`. I decided it was time to split the routes into different files to achieve that cleanly, I should have done that day one imo.
After doing everything I realized `router.use('/')` for public API was still pushing everything at the root level so you still need some sorting but at least the rest is better scoped.

--- 
For example, right now the UI can't call `api.nango.dev` because the CORS are the ones from public API
Screenshot 2025-03-26 at 10 55 26


packages/server/lib/clients/auth.client.ts
```diff
@@ -22,7 +24,6 @@ const sessionStore = new KnexSessionStore({
 
 export function setupAuth(app: express.Router) {
     app.use(cookieParser());
-    app.use(express.static(path.join(dirname(), 'public')));
```
bodinsamuel, line 25: probably a left over from a long time ago

packages/server/lib/middleware/access.middleware.ts
```diff
@@ -456,7 +461,8 @@ async function fillLocalsFromSession(req: Request, res: Response<any, RequestLoc
 
         res.locals['user'] = user;
 
-        if (ignoreEnvPaths.includes(req.route.path)) {
+        const fullPath = path.join(req.baseUrl, req.route.path);
```
bodinsamuel, line 464: because we use subrouter and express is dumb we have to compute our own full path

bodinsamuel:
Tested in staging

TBonnin (approved):
I am gonna miss having to check only one file to get all the possible entrypoints of server. But it makes sense to split

I haven't checked all the routes. That's the kind of change I wish we had a way to do  partial rollout, blue/green or something that doesn't require to rollout everything at once :/

bodinsamuel:
> I am gonna miss having to check only one file to get all the possible entrypoints of server. But it makes sense to split

Yeah me too, at some point we'll probably split the services entirely, the ransom of glory

bodinsamuel:
> I haven't checked all the routes. That's the kind of change I wish we had a way to do partial rollout, blue/green or something that doesn't require to rollout everything at once :/

yep, or even fast rollback would be nice. That's why I want to move forward replacing server with docker image asap

PR #3777, opened 2025-03-26
fix(ui): use correct API hostname

bodinsamuel:
## Changes 

Contributes to https://linear.app/nango/issue/NAN-2116/webapp-use-vite-instead-of-webhookcreate-react-app

- Use the correct API hostname instead of relying on the proxy to forward app.nango.dev to the api
- Bonus stuff: missing package, import order matters in next version


bodinsamuel:
Tested on staging

TBonnin (approved):

PR #3779, opened 2025-03-26
fix(cli): reup validation error message

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2987/validation-fails-without-showing-specific-error-reasons

- Reup validation error message for syncs
Error was not properly handled since we had changed the constructor name. Plus I think it never worked for syncs since we are throwing anyway.

- Fix remove metadata not checking if record is an actual object
- Fix add color to log 
Used to be the case when we using logger in shared

### DIFF
--- a/packages/runner-sdk/lib/sync.ts
+++ b/packages/runner-sdk/lib/sync.ts
@@ -1,7 +1,8 @@
-import type { MaybePromise, NangoProps } from '@nangohq/types';
-import { validateData } from './dataValidation.js';
-import type { ValidateDataError } from './dataValidation.js';
 import { NangoActionBase } from './action.js';
+import { validateData } from './dataValidation.js';
+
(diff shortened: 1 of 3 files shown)


khaliqgant (approved):

packages/runner-sdk/lib/sync.ts
```diff
@@ -80,7 +81,7 @@ export abstract class NangoSyncBase extends NangoActionBase {
             return results;
         }
         return results.map((result) => {
-            if (result && '_nango_metadata' in result) {
+            if (result && typeof result === 'object' && '_nango_metadata' in result) {
```
TBonnin, line 84: weren't types supposed to prevent the need for excessive defensive programming? 🙃

TBonnin (approved):

PR #3781, opened 2025-03-26
chore: 2 metrics about ingested records

TBonnin:
- bringing back the old metrics counting the number of records ingested and their size 
- adding a new metric counting the number of modified records and their size

### DIFF
--- a/packages/persist/lib/records.ts
+++ b/packages/persist/lib/records.ts
@@ -125,20 +125,24 @@ export async function persistRecords({
             updatedResults
         });
 
-        const recordsSizeInBytes = recordsData.reduce((acc, record) => {
+        const recordsSizeInBytes = Buffer.byteLength(JSON.stringify(records), 'utf8');
+        const modifiedRecordsSizeInBytes = recordsData.reduce((acc, record) => {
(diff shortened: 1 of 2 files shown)


packages/persist/lib/records.ts
```diff
@@ -125,20 +125,24 @@ export async function persistRecords({
             updatedResults
         });
 
-        const recordsSizeInBytes = recordsData.reduce((acc, record) => {
+        const recordsSizeInBytes = Buffer.byteLength(JSON.stringify(records), 'utf8');
```
bodinsamuel, line 128: sugg: can we use req.contentLength or something like this to avoid the extensive stringify? if not, no worries

bodinsamuel (approved):
Thanks 🙏🏻

packages/persist/lib/records.ts
TBonnin: I quickly check and `stringify` represents very much nothing compared to the round trip between runner<->persist or saving to the db.
I am gonna keep it like this is for now. We can revisit if it ever becomes a bottleneck

bodinsamuel: alright makes sense

PR #3782, opened 2025-03-26
chore(api): refactor POST /sync/trigger

kaposke:
Before adding the new `sync_mode` property and deprecating `full_resync`, I'm re-writing the endpoint in the new way.

Pre-req for [NAN-2964](https://linear.app/nango/issue/NAN-2964/triggers-sync-clear-cache-from-apisdk)

Any feedback and suggestions on how to properly tests this is appreciated.


packages/types/lib/sync/api.ts
```diff
@@ -1,5 +1,22 @@
...
+        full_resync: boolean;
+        provider_config_key?: string;
+        connection_id?: string;
+    };
+    Headers: {
+        'Provider-Config-Key'?: string;
+        'Connection-Id'?: string;
+    };
+    Success: { success: boolean };
+    Error: ApiError<'invalid_query_params' | 'invalid_body' | 'invalid_sync' | 'invalid_headers' | 'missing_provider_config_key' | 'missing_connection_id'>;
```
bodinsamuel: We have default Error values, I think most of them are already covered.
See `ResDefaultErrors` in types/lib/api.ts

packages/server/lib/controllers/sync/postTrigger.ts
```diff
@@ -0,0 +1,102 @@
...
+        return;
+    }
+
+    const body = bodyValidation.safeParse(req.body);
+    if (!body.success) {
+        res.status(400).send({ error: { code: 'invalid_body', errors: zodErrorToHTTP(body.error) } });
+        return;
+    }
+
+    const headers = headersValidation.safeParse(req.headers);
```
bodinsamuel: You'll need to tweak because headers are case-insensitive but` req.headers` is not giving lowercase headers I think. See getRecords.ts for working implem

packages/types/lib/sync/api.ts
```diff
@@ -1,5 +1,22 @@
 import type { ApiError, Endpoint } from '../api.js';
 
+export type PostTrigger = Endpoint<{
```
bodinsamuel: Can you also add it to `PublicApiEndpoints` in api.endpoints.ts
```suggestion
export type PostPublicTrigger = Endpoint<{
```

packages/types/lib/sync/api.ts
```diff
@@ -1,5 +1,22 @@
...
+    Method: 'POST';
+    Path: '/sync/trigger';
+    Body: {
+        syncs: (string | { name: string; variant: string })[];
+        full_resync: boolean;
```
bodinsamuel: full_resync is optional

packages/server/lib/controllers/sync/postTrigger.ts
```diff
@@ -0,0 +1,102 @@
...
+    if (!headers.success) {
+        res.status(400).send({ error: { code: 'invalid_headers', errors: zodErrorToHTTP(headers.error) } });
+        return;
+    }
+
+    const { syncs, full_resync } = body.data;
```
bodinsamuel: can you reassign just to counter type? this is to prevent desync between zod and types
```suggestion
    const { syncs, full_resync }: PostTrigger["Body"] = body.data;
```

packages/server/lib/controllers/sync/postTrigger.integration.test.ts
```diff
@@ -0,0 +1,300 @@
...
+                    errors: [
+                        {
+                            code: 'invalid_type',
+                            message: 'Required',
+                            path: ['syncs']
+                        }
+                    ]
+                }
+            });
+        });
```
bodinsamuel, line 76: nit: to avoid too many tests, query params or body validation can be grouped into one big test. The goal is just to ensure we have validation and we catch most errors. Otherwise we should trust zod to work as expected

bodinsamuel (commented):
Amazing, congrats on your first big PR 💪🏻 
Not tested yet but looks good, mostly small comments 🚀

TBonnin (approved):
very good job

packages/server/lib/controllers/sync/postTrigger.ts
```diff
@@ -0,0 +1,108 @@
...
+
+    if (!success) {
+        errorManager.errResFromNangoErr(res, error);
+        return;
+    }
+
+    res.status(200).send({ success: true });
+});
+
+function normalizedSyncParams(syncs: (string | { name: string; variant: string })[]): { syncName: string; syncVariant: string }[] {
```
TBonnin: we are gonna need this function for other endpoints. We could already move it into a helpers.ts or something similar file in this folder

packages/types/lib/sync/api.ts
```diff
@@ -1,4 +1,21 @@
...
+        full_resync?: boolean | undefined;
+        provider_config_key?: string | undefined;
+        connection_id?: string | undefined;
+    };
+    Headers: {
+        'provider-config-key'?: string | undefined;
+        'connection-id'?: string | undefined;
+    };
+    Success: { success: boolean };
+    Error: ResDefaultErrors;
```
bodinsamuel: It's already the default value, so you don't even need to put that ☺️ 
Note that if you want custom code you can do `Error: ApiError<'missing_connection_id'>` and it will extend the default not override 👍🏻

bodinsamuel (approved):

PR #3785, opened 2025-03-26
feat(azure-devops): Implement support for Azure Devops

lordsarcastic:
This PR adds Nango proxy support for Azure Devops

Ticket: [EXT-573](https://linear.app/nango/issue/EXT-573/add-support-for-azure-devops)


packages/server/lib/hooks/connection/providers/azure-devops/credentials-verification.ts
```diff
@@ -0,0 +1,14 @@
+import type { InternalNango as Nango } from '../../credentials-verification-script.js';
+
+export default async function execute(nango: Nango) {
+    const { provider_config_key } = nango.getConnection();
+
+    const response = await nango.proxy({
+        endpoint: '/_apis/projects',
```
bodinsamuel, line 7: so it's static? can it just be an endpoint in that case? 👍🏻

packages/webapp/public/images/template-logos/azure-devops.svg
bodinsamuel, line 1: It must be 62x62

packages/server/lib/hooks/connection/providers/azure-devops/credentials-verification.ts
lordsarcastic: Not sure I understand what you mean

khaliqgant: This doesn't need to be an actual script but rather could live in the providers.yaml is a single line like this https://github.com/NangoHQ/nango/blob/master/packages/providers/providers.yaml#L11-L14

lordsarcastic: I understand now. The reason for the script is that the success response for the permissions check is 200 while the failure response is 202. This means that both success and failure are interpreted as success, hence the need for  a script

bodinsamuel: alright makes sense, a comment would be helpful for future reference in that case

khaliqgant (approved):

bodinsamuel (approved):

khaliqgant (approved):

PR #3786, opened 2025-03-26
fix(docs): Reported documentation issues

lordsarcastic:
A few integrations have some documentation issues: Zapier, Datadog, and Adobe

[EXT-578](https://linear.app/nango/issue/EXT-578/electric-reported-documentation-issues)


bodinsamuel (approved):

khaliqgant (changes requested):
keeper isn't updated. 

https://github.com/NangoHQ/nango/blob/master/packages/providers/providers.yaml#L5040C19-L5040C73 should be https://docs.nango.dev/integrations/all/keeper-scim/connect

(is missing the scim part)

khaliqgant (approved):

PR #3787, opened 2025-03-26
fix: fleet fallback db url

TBonnin:
avoiding issue in case an enterprise customer uses NANGO_DATABASE_URL instead of the multiple DB related env vars


bodinsamuel (approved):
ah yes indeed

PR #3792, opened 2025-03-27
fix(auth): oauth2_cc support authorization_params

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2996/support-authorization-params-in-oauth2-cc

- oauth2_cc support authorization_params
We were not forwarding the param like we are doing in oauth2. Also saving them in connectionConfig (unfortunately I wasn't really aware that it would be better in connectionConfig so I need to do it manually now)

- Introduce `loggedFetch` 
I noticed none of the auth have clear trace in the system so I tried to put some logging and realised it would be super repetitive. So it will help replace all of them with the same code.

---
Screenshot 2025-03-27 at 15 12 48
Screenshot 2025-03-27 at 15 11 58


packages/shared/lib/utils/http.ts
```diff
@@ -0,0 +1,92 @@
...
+                meta: { body }
+            });
+        } else {
+            void options.logCtx.http(`${props.method} ${url.href}`, {
+                request: requestLog,
+                response: { code: res.status, headers: redactHeaders({ headers: res.headers }) },
+                context: options.context,
+                createdAt
+            });
+        }
```
bodinsamuel, line 71: Note there are two kinds of errors, critical one like Network error and HTTP error, I chose to still return Ok on HTTP Error because it's easier to consume. Let me know if you have suggestion

khaliqgant (approved):

packages/shared/lib/utils/http.ts
```diff
@@ -0,0 +1,92 @@
...
+        headers?: Record<string, string> | undefined;
+        body?: string | null;
+    },
+    options: {
+        // TODO: maybe put that in logCtx
+        valuesToFilter: string[];
+        logCtx: LogContextStateless;
+        context: MessageRow['context'];
+    }
+): Promise<Result<{ res: Response; body: TBody }>> {
```
TBonnin, line 28: what about a function with this signature: 
```
export function withHttpLogging({
    logCtx,
    context,
    valuesToFilter
}: {
    logCtx: LogContextStateless;
    context: MessageRow['context'];
    valuesToFilter: string[];
}): { fetch: (url: URL | string, init?: RequestInit) => Promise<Response> } {
    return {
        fetch: async (url, init) => {
            // ...
            const response = await fetch(url, init);
            // ...
            return response;
        }
    };
}

```

so you can use it exactly like fetch with just some prefixing:
```
const response = await withHttpLogging({logCtx, context: 'auth', ...}).fetch(url, options)
```

TBonnin (approved):

packages/shared/lib/utils/http.ts
bodinsamuel: Is it just about the signature shape? Because I still need to try catch fetch and decode the body to log it so the response would be the same. If we agree on this, sure I can do that. Looks better

TBonnin: yes the body of the function is the same except it is returning the fetch response so can be used as a fetch drop-in replacement.

bodinsamuel: I don't think it's possible to do a drop-in replacement, native fetch returns .text()/.json() but they can only be used once otherwise it throws. And also in a drop-in replacement it would throw on network error which we don't want (unless I'm rethrowing but that would means a try/catch just for doing nothing)

Do you have a suggestion?

TBonnin: good point. Fine with returning a `Result` then. 
I just feel the name doesn't fully reflect the behavior of the function. It should be called `fetchThatLogsAndUnwrapBodyAndNotThrowOnNetworkError` 😆 

I trust you, go for the signature you prefer

bodinsamuel: I'll go like this the, since there is no consensus on naming and can be changed later since it's internal

PR #3794, opened 2025-03-27
fix(api): POST /logout new format

bodinsamuel:
## Changes

- POST /logout new format


packages/types/lib/account/api.ts
```diff
@@ -93,6 +93,15 @@ export type PostSignin = Endpoint<{
     };
 }>;
 
+export type PostLogout = Endpoint<{
+    Method: 'POST';
+    Path: '/api/v1/account/logout';
+    Error: ApiError<'email_not_verified'> | ApiError<'unauthorized'>;
+    Success: {
+        user: ApiUser;
```
kaposke: I think it's actually an empty body?

TBonnin (approved):
curious about @kaposke comment but looks good otherwise

packages/types/lib/account/api.ts
bodinsamuel: bad c/c thanks

PR #3795, opened 2025-03-27
feat(api): add sync_mode param to /sync/trigger

kaposke:
We're adding a new `sync_mode` parameter to `/sync/trigger`, and deprecating `full_resync` while maintaining backwards compatibility.

Docs and SDK to be done in follow-up PRs.

[NAN-2964](https://linear.app/nango/issue/NAN-2964/triggers-sync-clear-cache-from-apisdk)


packages/server/lib/controllers/sync/postTrigger.ts
```diff
@@ -18,11 +18,26 @@ const bodyValidation = z
                 errorMap: () => ({ message: 'Each sync must be either a string or a { name: string, variant: string } object' })
             })
         ),
+        sync_mode: z
+            .string()
+            .toLowerCase()
+            .pipe(z.enum(['incremental', 'full_refresh', 'full_refresh_and_clear_cache']))
+            .optional(),
```
bodinsamuel: I don't think we care about supporting uppercase/different casing, on the contrary, this allows people to set invalid values and get away with it

packages/types/lib/sync/index.ts
```diff
@@ -0,0 +1 @@
+export type SyncMode = 'incremental' | 'full_refresh' | 'full_refresh_and_clear_cache';
```
bodinsamuel: Can you make it more explicitly tied to the API body? SyncMode is vague and could be mistakenly reused ☺️

packages/types/lib/sync/api.ts
```diff
@@ -1,13 +1,16 @@
...
 export type PostPublicTrigger = Endpoint<{
     Method: 'POST';
     Path: '/sync/trigger';
     Body: {
         syncs: (string | { name: string; variant: string })[];
-        full_resync?: boolean | undefined;
+        sync_mode?: SyncMode | undefined;
         provider_config_key?: string | undefined;
         connection_id?: string | undefined;
+        // Deprecated in favor of sync_mode
```
bodinsamuel:
```suggestion
        // @deprecrated in favor of sync_mode
```

bodinsamuel (commented):
Works well 🙏🏻 
Can you update `spec.yaml` to add the new prop and mark the other as deprecated please

packages/types/lib/sync/index.ts
kaposke: Renamed to `SyncTriggerMode`

packages/server/lib/controllers/sync/postTrigger.integration.test.ts
```diff
@@ -65,6 +67,12 @@ describe(`POST ${endpoint}`, () => {
                             message: 'Each sync must be either a string or a { name: string, variant: string } object',
                             path: ['syncs', 2]
                         },
+                        {
+                            code: 'invalid_enum_value',
+                            message:
+                                "Invalid enum value. Expected 'incremental' | 'full_refresh' | 'full_refresh_and_clear_cache', received 'invalid-sync-mode'",
```
TBonnin, line 73: not sure the `received 'invalid-sync-mode'` part adds much value

TBonnin (approved):

packages/server/lib/controllers/sync/postTrigger.integration.test.ts
kaposke: Yeah, but it's zod's default. Should probably just go with it?

packages/types/lib/sync/index.ts
TBonnin: I think Samuel means using the `PostPublicTrigger['Body']['sync_mode']` type directly instead of declaring a new type that can be used outside of the PostTrigger context

packages/server/lib/controllers/sync/postTrigger.integration.test.ts
TBonnin: I don't feel strongly. I just think that an error message saying `error: invalid value, expected A, received WHATEVER_I_SENT` would make sense but `error: invalid value, expected A, received invalid` is not better than simply responding with `error:  invalid value, expected A`

packages/types/lib/sync/index.ts
kaposke: The same type is going to be used in the SDK, so I think it should be it's own thing indeed?

kaposke: After private convo, I've removed the type entirely and I now use `PostPublicTrigger['Body']['sync_mode']` instead.

bodinsamuel: I didn't mean that, but yes it can work too. Finding the right balance is tricky because users might want to use those types too.

bodinsamuel (approved):

PR #3797, opened 2025-03-27
fix: abort all scripts when timing out

TBonnin:
We currently only abort syncs when they timed out, which means that action, webhook or on-event scripts keep executing in the runner even though the task is timed out
This commit adds support for aborting any script

### DIFF
--- a/packages/jobs/lib/execution/operations/abort.ts
+++ b/packages/jobs/lib/execution/operations/abort.ts
@@ -1,8 +1,25 @@
 import type { Result } from '@nangohq/utils';
 import { Err, Ok } from '@nangohq/utils';
 import { getRunner } from '../../runner/runner.js';
+import { environmentService } from '@nangohq/shared';
+import { logger } from '../../logger.js';
+import type { TaskAbort } from '@nangohq/nango-orchestrator';
 
-export async function abortScript({ taskId, teamId }: { taskId: string; teamId: number }): Promise<Result<void>> {
+export async function abortTask(task: TaskAbort): Promise<Result<void>> {
+    const accountAndEnv = await environmentService.getAccountAndEnvironment({ environmentId: task.connection.environment_id });
+    if (!accountAndEnv) {
+        throw new Error(`Account and environment not found`);
+    }
+    const { account: team } = accountAndEnv;
+
+    const abortedScript = await abortTaskWithId({ taskId: task.abortedTask.id, teamId: team.id });
+    if (abortedScript.isErr()) {
+        logger.error(`failed to abort script for task ${task.abortedTask.id}`, abortedScript.error);
+    }
+    return abortedScript;
+}
+
+export async function abortTaskWithId({ taskId, teamId }: { taskId: string; teamId: number }): Promise<Result<void>> {
     const runner = await getRunner(teamId);
     if (runner.isErr()) {
         return Err(runner.error);
--- a/packages/jobs/lib/execution/operations/abort.ts
+++ b/packages/jobs/lib/execution/operations/abort.ts
@@ -8,7 +8,7 @@ import type { TaskAbort } from '@nangohq/nango-orchestrator';
 export async function abortTask(task: TaskAbort): Promise<Result<void>> {
     const accountAndEnv = await environmentService.getAccountAndEnvironment({ environmentId: task.connection.environment_id });
     if (!accountAndEnv) {
-        throw new Error(`Account and environment not found`);
+        return Err(`Account and environment not found`);
     }
     const { account: team } = accountAndEnv;
 
--- a/packages/jobs/lib/processor/handler.ts
+++ b/packages/jobs/lib/processor/handler.ts
@@ -6,6 +6,7 @@ import { startSync, abortSync } from '../execution/sync.js';
 import { startAction } from '../execution/action.js';
 import { startWebhook } from '../execution/webhook.js';
 import { startOnEvent } from '../execution/onEvent.js';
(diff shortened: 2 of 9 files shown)


packages/orchestrator/lib/clients/types.ts
```diff
@@ -76,6 +77,26 @@ interface TaskCommon extends TaskCommonFields {
     isAction(this: OrchestratorTask): this is TaskAction;
     isOnEvent(this: OrchestratorTask): this is TaskOnEvent;
     isSyncAbort(this: OrchestratorTask): this is TaskSyncAbort;
+    isAbort(this: OrchestratorTask): this is TaskAbort;
+}
+export interface TaskAbort extends TaskCommon, AbortArgs {}
```
TBonnin, line 82: I could have gone with a abort type per type of script (ie: TaskSyncAbort, TaskActionAbort, TaskWebhookAbort, etc...) or a single type with optional field to deal with TaskSyncAbort which is a bit different. 
I decided to go for a middle ground solution with TaskSyncAbort and a generic TaskAbort for everything else

packages/orchestrator/lib/clients/validate.ts
```diff
@@ -178,9 +187,22 @@ export function validateTask(task: Task): Result<OrchestratorTask> {
             })
         );
     }
-    return Err(
-        `Cannot validate task ${JSON.stringify(task)}: ${stringifyError(sync.error || action.error || webhook.error || onEvent.error || syncAbort.error)}`
```
TBonnin, line 182: returning the error is useless since if no parsing was successful it would always return the first sync.error

packages/runner/lib/server.ts
```diff
@@ -67,9 +67,7 @@ function startProcedure() {
                 }, heartbeatIntervalMs);
                 try {
                     const abortController = new AbortController();
-                    if (nangoProps.scriptType == 'sync' && nangoProps.activityLogId) {
```
TBonnin, line 70: not sure why the activityLogId condition was there. Maybe some legacy code

packages/jobs/lib/processor/handler.ts
```diff
@@ -48,5 +49,13 @@ export async function handler(task: OrchestratorTask): Promise<Result<void>> {
             return res;
         });
     }
+    if (task.isAbort()) {
+        const span = tracer.startSpan('jobs.handler.abort');
+        return await tracer.scope().activate(span, async () => {
+            const res = await abortTask(task);
+            span.finish();
+            return res;
+        });
```
bodinsamuel, line 58: nb: we could use await tracer.trace for simplicity

packages/jobs/lib/execution/operations/abort.ts
```diff
@@ -1,8 +1,25 @@
...
 import { getRunner } from '../../runner/runner.js';
+import { environmentService } from '@nangohq/shared';
+import { logger } from '../../logger.js';
+import type { TaskAbort } from '@nangohq/nango-orchestrator';
 
-export async function abortScript({ taskId, teamId }: { taskId: string; teamId: number }): Promise<Result<void>> {
+export async function abortTask(task: TaskAbort): Promise<Result<void>> {
+    const accountAndEnv = await environmentService.getAccountAndEnvironment({ environmentId: task.connection.environment_id });
+    if (!accountAndEnv) {
+        throw new Error(`Account and environment not found`);
```
bodinsamuel: why throw if you can return Err?

packages/jobs/lib/execution/operations/abort.ts
```diff
@@ -1,8 +1,25 @@
...
+    }
+    const { account: team } = accountAndEnv;
+
+    const abortedScript = await abortTaskWithId({ taskId: task.abortedTask.id, teamId: team.id });
+    if (abortedScript.isErr()) {
+        logger.error(`failed to abort script for task ${task.abortedTask.id}`, abortedScript.error);
```
bodinsamuel, line 17: why not Err?

packages/orchestrator/lib/abort.ts
```diff
@@ -3,40 +3,43 @@ import { validateTask } from './clients/validate.js';
...
+            : payload,
+        groupKey: aborted.value.groupKey,
+        retryMax: 0,
+        retryCount: 0,
+        createdToStartedTimeoutSecs: 60,
+        startedToCompletedTimeoutSecs: 60,
+        heartbeatTimeoutSecs: 60
+    });
+    if (abortTask.isErr()) {
+        logger.error(`Failed to create abort task: ${stringifyError(abortTask.error)}`);
```
bodinsamuel: sugg you can also do this
```suggestion
        logger.error(`Failed to create abort task`, abortTask.error);
```

bodinsamuel (approved):

packages/jobs/lib/execution/operations/abort.ts
TBonnin: good point. originally the code came from execution/sync.ts where error where handled locally via a try catch but here it doesn't make any sense. Thanks for catching

packages/jobs/lib/execution/operations/abort.ts
TBonnin: the result is returned below. this is just adding an extra logging. You could argue it is not 100% necessary.

packages/jobs/lib/processor/handler.ts
TBonnin: yes. I'll do it next time I will touch the handler since we'll have to do it for all types of tasks

PR #3800, opened 2025-03-28
feat(integrations): add support for github pat

hassan254-prog:

- Add support for Github PAT


khaliqgant (approved):

packages/providers/providers.yaml
```diff
@@ -3895,6 +3895,37 @@ github-app-oauth:
...
+    auth_mode: API_KEY
+    proxy:
+        base_url: https://api.github.com
+        retry:
+            at: 'x-ratelimit-reset'
+            remaining: 'x-ratelimit-remaining'
+            error_code: 403
+        headers:
+            authorization: Bearer ${apiKey}
+            x-github-api-version: '2022-11-28'
```
bodinsamuel: Can be this override by connectionConfig? I feel like it's going to bite us quickly

bodinsamuel (approved):

packages/providers/providers.yaml
hassan254-prog: @bodinsamuel Yes, I agree 💯! There are so many ways to do it. Should we make it `optional` or `automated` with a fallback `2022-11-28` so that our customer customers don’t have to fill it out themselves?

bodinsamuel: as long as it's doable at some point I don't think we need to do it now, unless you are motivated :D 
If so, automated seems correct or just hidden maybe I don't know, they can specify that in the connect session default connectionConfig

hassan254-prog: I guess it is better we sort it now than have issues with it later down the line.

packages/shared/lib/utils/utils.ts
```diff
@@ -289,9 +289,12 @@ export function mapProxyBaseUrlInterpolationFormat(baseUrl: string | undefined):
...
     if (str.includes('${')) {
         if (str.includes('||')) {
-            const parts = str.split('||');
-            const firstPart = parts[0] ? interpolateStringFromObject(parts[0].trim(), replacers) : undefined;
+            const parts = str.split('||').map((part) => part.trim());
+            let firstPart = parts[0] ? interpolateStringFromObject(parts[0].trim(), replacers) : undefined;
             const secondPart = parts[1] ? interpolateStringFromObject(parts[1].trim(), replacers) : undefined;
+            if (firstPart && firstPart.includes('${')) {
+                firstPart = undefined;
+            }
```
TBonnin, line 297: do we need to interpolate the fallback if first value is interpolated?
```
            const parts = str.split('||').map((part) => part.trim());
            cont left = parts[0] ? interpolateStringFromObject(parts[0].trim(), replacers) : undefined;
            // if(left && left !=== parts[0]) { // left was interpolated
                return left
            }
            return  interpolateStringFromObject(parts[1].trim(), replacers) : undefined;
```

hassan254-prog: I agree 💯 , there is no need of interpolating the fallback if we already got a value from the left side. updated with the latest [commit](https://github.com/NangoHQ/nango/pull/3800/commits/e8ebf2cc6f30eb30159fcd2fbc4c8d9e5e9d118a)

TBonnin (approved):

PR #3801, opened 2025-03-28
feat: nango to respond/forward webhook immediately

TBonnin:
This commit is changing the behavior of 3rd party webhooks. 
Nango will responds to the `/webhook` request immediately and not wait for the webhook script to execute. The webhook is also forwarded immediately

### DIFF
--- a/packages/jobs/lib/execution/sync.ts
+++ b/packages/jobs/lib/execution/sync.ts
@@ -803,4 +803,5 @@ async function onFailure({
     }
 
     metrics.increment(metrics.Types.SYNC_FAILURE);
+    metrics.duration(metrics.Types.SYNC_TRACK_RUNTIME, Date.now() - startedAt.getTime(), team ? { accountId: team.id } : {});
 }
--- a/packages/server/lib/controllers/webhook/environmentUuid/postWebhook.ts
+++ b/packages/server/lib/controllers/webhook/environmentUuid/postWebhook.ts
@@ -66,19 +66,9 @@ export const postWebhook = asyncWrapper<PostPublicWebhook>(async (req, res) => {
 
             metrics.increment(metrics.Types.WEBHOOK_INCOMING_RECEIVED);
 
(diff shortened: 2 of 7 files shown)


packages/jobs/lib/execution/sync.ts
```diff
@@ -803,4 +803,5 @@ async function onFailure({
     }
 
     metrics.increment(metrics.Types.SYNC_FAILURE);
+    metrics.duration(metrics.Types.SYNC_TRACK_RUNTIME, Date.now() - startedAt.getTime(), team ? { accountId: team.id } : {});
```
TBonnin, line 806: unrelated but we were not tracking the duration of failed syncs

packages/server/lib/controllers/webhook/environmentUuid/postWebhook.ts
```diff
@@ -66,19 +66,9 @@ export const postWebhook = asyncWrapper<PostPublicWebhook>(async (req, res) => {
...
 
-            metrics.duration(metrics.Types.WEBHOOK_TRACK_RUNTIME, totalRunTime, { accountId: account.id });
-
-            if (!responsePayload) {
-                res.status(200).send();
-                return;
-            }
-
-            res.status(200).send(responsePayload);
+            res.status(200).send();
```
TBonnin, line 71: I went for responding with a 200 with no body. wdyt?

packages/jobs/lib/execution/sync.ts
bodinsamuel: damn, good catch. Maybe we should have a dimension for failed vs success (we are no longer counting ahah)

bodinsamuel (approved):
Yes 🔥

packages/jobs/lib/execution/sync.ts
TBonnin: > Maybe we should have a dimension for failed vs success
Maybe...😏

PR #3805, opened 2025-03-28
feat(node-client): support sync_mode in triggerSync

kaposke:
Replacing `full_resync` with `syncMode` in the node client while maintaining backwards compatibility.

[NAN-2964](https://linear.app/nango/issue/NAN-2964/triggers-sync-clear-cache-from-apisdk)

### DIFF
--- a/packages/node-client/lib/index.ts
+++ b/packages/node-client/lib/index.ts
@@ -1,37 +1,10 @@
 import crypto from 'node:crypto';
-import type { AxiosInstance, AxiosRequestConfig, AxiosResponse, InternalAxiosRequestConfig } from 'axios';
-import axios from 'axios';
 import https from 'node:https';
 
-import type {
-    ApiKeyCredentials,
-    AppCredentials,
-    OAuth1Token,
-    AppStoreCredentials,
-    BasicApiCredentials,
-    CredentialsCommon,
-    CustomCredentials,
-    OAuth2ClientCredentials,
-    TbaCredentials,
-    TableauCredentials,
-    UnauthCredentials,
-    BillCredentials,
-    GetPublicProviders,
-    GetPublicProvider,
-    GetPublicListIntegrations,
-    GetPublicListIntegrationsLegacy,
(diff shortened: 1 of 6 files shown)


packages/node-client/lib/index.ts
```diff
@@ -570,26 +573,32 @@ export class Nango {
...
         if (syncs && !Array.isArray(syncs)) {
             throw new Error('Syncs must be an array. If it is a single sync, please wrap it in an array.');
         }
 
+        if (typeof syncMode === 'boolean') {
+            syncMode = syncMode ? 'full_refresh' : 'incremental';
+        }
+
+        syncMode ??= 'incremental';
+
```
TBonnin, line 596: I wish ts had pattern matching.

packages/server/lib/controllers/sync/postTrigger.ts
```diff
@@ -74,10 +74,6 @@ export const postPublicTrigger = asyncWrapper<PostPublicTrigger>(async (req, res
     }
 
     const connectionId: string | undefined = body.connection_id || headers['connection-id'];
-    if (!connectionId) {
-        res.status(400).send({ error: { code: 'missing_connection_id', message: 'Missing connection_id. Provide it in the body or headers.' } });
-        return;
-    }
```
TBonnin, line 80: why removing this?

TBonnin (approved):

packages/server/lib/controllers/sync/postTrigger.ts
kaposke: Sorry, I was about to add a comment. I realized it's actually optional. The original endpoint that I revamped didn't have this validation, but it had a bang (`connectionId!`) operator when passing it down, so I thought it was forgotten. But it's actually optional in the SDK, and there's code to deal with it's absence down the line.

packages/node-client/lib/index.ts
kaposke: I wish TS was C#

PR #3806, opened 2025-03-28
fix: do not block request if rate limit cannot be computed

TBonnin:
As experienced recently, a redis outage would cause all requests made to Nango api and web to fail.
This commit changes the behavior to not block requests if rate limit cannot be computed, which will allow requests to dashboard and api to continue to be served even if redis isn't reachable 
Caveat: if Redis is down we don't have any rate-limit protection but it is better imho than to be fully down. wdyt?

### DIFF
--- a/packages/server/lib/middleware/ratelimit.middleware.ts
+++ b/packages/server/lib/middleware/ratelimit.middleware.ts
@@ -62,7 +62,8 @@ export const rateLimiterMiddleware = (req: Request, res: Response, next: NextFun
             }
 
             logger.error('Failed to compute rate limit', { error: err });
-            res.status(500).send({ error: { code: 'server_error', message: 'Failed to compute rate limit' } });
+            // If we can't get the rate limit (ex: redis is unreachable), we should not block the request
+            next();
         });
 };
 


packages/server/lib/middleware/ratelimit.middleware.ts
```diff
@@ -62,7 +62,8 @@ export const rateLimiterMiddleware = (req: Request, res: Response, next: NextFun
             }
 
             logger.error('Failed to compute rate limit', { error: err });
-            res.status(500).send({ error: { code: 'server_error', message: 'Failed to compute rate limit' } });
+            // If we can't get the rate limit (ex: redis is unreachable), we should not block the request
```
bodinsamuel, line 65: we could use `report()` now in that case instead of swallowing the error

bodinsamuel (approved):

packages/server/lib/middleware/ratelimit.middleware.ts
TBonnin: we'd be flooding the error reporting though. I am for now making the assumption that if Redis is unreachable we will get alerted somehow via another channel

PR #3803, opened 2025-03-28
feat(plan): trial 

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2981/show-banner-in-ui
Figma https://www.figma.com/design/aR5Pjn4JHJWsYxJ3s4iD3D/Free-Trial?node-id=64-1276&t=9lU4gkM197KZudiB-0 

- Added: trial banner in UI
- Added: API to extend trial
- Added: tooltip next to script toggles
- Added: Send email 3 days before trial expire 
- Changed: expired trial do not have the ability to enable scripts anymore
- Changed: increase free plan to 50 (checked with @bastienbeurier)
- Changed: pause all scripts after trial is over

- Email at the end of the trial
- Small leftover for is_capped 

## 🧪 Tests

- Launch your local dev `npm run dev:watch:web`
- Modify your plan in DB, set "free" and some date from the trial
- Go to integrations

Screenshot 2025-04-03 at 17 11 11
Screenshot 2025-04-03 at 17 11 31

PR #3807, opened 2025-03-31
feat(plans): table migration

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2980/modify-db-to-store-trial-information
Split of #3803

- Migration: create plans table and types


packages/database/lib/migrations/20250325122817_plans.cjs
```diff
@@ -0,0 +1,31 @@
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(
+        `CREATE TABLE "plans" (
```
TBonnin, line 6: I am not a huge fan of the name `plans`. For me plans represents a bundle of features, where here we are storing the features individually. Why not `features`?

packages/types/lib/plans/db.ts
```diff
@@ -0,0 +1,38 @@
...
+    trial_end_at: Date | null;
+    trial_extension_count: number;
+    trial_end_notified_at: Date | null;
+
+    /**
+     * Limit the number of connections with active scripts
+     * Set to null to remove limit
+     * @default 3
+     */
+    max_connection_with_scripts: number | null;
```
TBonnin: I would be consistent in the naming and follow what's done with the trial features aka `FEATURE_ATTRIBUTE`
```
connection_with_scripts_max
environments_max
...
```

packages/types/lib/plans/db.ts
```diff
@@ -0,0 +1,38 @@
...
+     * @default 2
+     */
+    max_environments: number | null;
+
+    /**
+     * Limit the minimum frequency of a sync
+     * Not used yet
+     * @default 86400
+     */
+    min_sync_frequency: number;
```
TBonnin: when storing duration as number I like to include unit in the name so one doesn't have to guess the unit:
```
sync_frequency_secs_min
```

packages/types/lib/plans/db.ts
```diff
@@ -0,0 +1,38 @@
...
+     * Limit the number of environments that can be created
+     * Set to null to remove limit
+     * @default 2
+     */
+    max_environments: number | null;
+
+    /**
+     * Limit the minimum frequency of a sync
+     * Not used yet
+     * @default 86400
```
TBonnin, line 28: is it 1 day?

packages/types/lib/plans/db.ts
```diff
@@ -0,0 +1,38 @@
...
+     */
+    min_sync_frequency: number;
+
+    /**
+     * Enable or disabled sync variant
+     * @default false
+     */
+    has_sync_variant: boolean;
+}
```
TBonnin, line 43: can we add the `otel` feature already? Thank you 😁

packages/types/lib/plans/db.ts
```diff
@@ -0,0 +1,38 @@
+import type { Timestamps } from '../db';
+
+export interface DBPlan extends Timestamps {
+    id: number;
+    account_id: number;
+    name: string;
+    trial_start_at: Date | null;
+    trial_end_at: Date | null;
+    trial_extension_count: number;
```
TBonnin, line 9: cannot it be null?

packages/types/lib/plans/db.ts
```diff
@@ -0,0 +1,38 @@
...
+     * @default 3
+     */
+    max_connection_with_scripts: number | null;
+
+    /**
+     * Limit the number of environments that can be created
+     * Set to null to remove limit
+     * @default 2
+     */
+    max_environments: number | null;
```
TBonnin: why null if it has a default value?

packages/database/lib/migrations/20250325122817_plans.cjs
bodinsamuel: We are already discussing having stripe in the coming weeks; storing the plan dates will fit nicely in this table. Do you have an alternative suggestion?

packages/types/lib/plans/db.ts
bodinsamuel: 👍🏻

packages/types/lib/plans/db.ts
bodinsamuel: 1 day for free tier that's the number I got from the latest iteration (unless mistaken), it will definitely change for sure

packages/types/lib/plans/db.ts
bodinsamuel: As stated in the comment can be set to null to remove limitation

packages/types/lib/plans/db.ts
bodinsamuel: changed

packages/types/lib/plans/db.ts
bodinsamuel: changed

packages/types/lib/plans/db.ts
bodinsamuel: added

packages/types/lib/plans/db.ts
bodinsamuel: changed to be not null

packages/database/lib/migrations/20250325122817_plans.cjs
TBonnin: I haven't part of the discussion regarding stripe. How is this table gonna look like then?

packages/types/lib/plans/db.ts
TBonnin: up to you but I will personally make it not null and the code set a high enough value. Even if from a commercial point of view it is "unlimited" we should probably always have a ceiling value, just to make sure there is no accidental addition of 1000s of environments. For now  accidents are unlikely because adding an environment is done via the UI but if one day we allow to do it programmatically accidents are gonna happen 😁

TBonnin (approved):
I am not fully up-to-date regarding the plan ;) to add stripe data into this table. But I don't have any major concerns about this PR. 🚢

packages/database/lib/migrations/20250325122817_plans.cjs
bodinsamuel: Not sure yet but from my experience we just need a few fields
```
stripeCustomerId      
stripeSubscriptionId     
stripePriceId           
stripeCurrentPeriodStart 
stripeCurrentPeriodEnd          
```

packages/types/lib/plans/db.ts
bodinsamuel: true

PR #3810, opened 2025-03-31
feat(pagination): add pagination callback function 

hassan254-prog:

- Add a pagination callback function. This will enable users keep track by logging the response during pagination. 

The following configuration will log responses during pagination:
```
const config: ProxyConfiguration = {
    endpoint: '/crm/v3/objects/contacts',
    params: {
        properties: properties.join(',')
    },
    paginate: {
        type: 'cursor',
        cursor_path_in_response: 'paging.next.after',
        limit_name_in_request: 'limit',
        cursor_name_in_request: 'after',
        response_path: 'results',
        limit: 100,
        on_page: async ({ nextPageParam, response }) => {
            await nango.log(`Next page cursor = ${nextPageParam}`);
            // or log the total number of records if the provider offers this here as well
            await nango.log(`Fetched ${response.data.tags.length} records`);
          },
    },
    retries: 10
};

### CODE REVIEW

File: packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap [hunk 1]
```diff
@@ -62,6 +62,13 @@ interface Pagination {
     response_path?: string;
     limit_name_in_request: string;
     in_body?: boolean;
+    on_page?: (
+        paginationState: {
+            nextPageParam?: string | number | undefined;
+            response: AxiosResponse;
+            pageIndex: number;
```
TBonnin: I am not sure about the `pageIndex` name. It could be misleading if api uses a non-0 based indexing for their page or if customer doesn't start pagination at first page. what about `pageCount` or `paginationCount` which make it clearer it is the number of time pagination occurs

TBonnin (commented):
> This will enable users keep track by logging the response during pagination.

I am curious to understand why customers want to log during pagination?

packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
TBonnin: also we don't absolutely need the pageIndex attribute. It is very easy for customers to track their own counter

bodinsamuel: or `iteration`

bodinsamuel (approved):
~Also curious about the use case, otherwise looks good~ Nevermind it's in the ticket

packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
hassan254-prog: @TBonnin , I agree, the `response` and `nextPageParam` would be sufficient. Customers can easily track their own counter within the scripts. The feature will greatly benefit customers who would like to get a sense of how much data they have imported vs how much more is still left, if the provider has this in the response. That way they can log and track the progress. Updated this with the latest [commit](https://github.com/NangoHQ/nango/pull/3810/commits/a5bb8abb8f11f17c80d133d5cadc70e89bd3d41b)

TBonnin (approved):

PR #3811, opened 2025-03-31
fix(cli): compile only scripts in the nango yaml config

hassan254-prog:
- Compile only scripts in the nango yaml config. This will enable users comment out the integration in the nango yaml while still preserving the scripts themselves in the `nango-integration` folder
- Fixed a minor bug where the compilation was using an outdated YAML cache while compiling the ts files.

Run `npx nango dev`, then comment any integration while still preserving the scripts in the `nango-integration` folder, the compiler will log the following example message;
```
Skipping compilation: No configuration found for list-databases
```


packages/cli/lib/services/compile.service.ts
```diff
@@ -180,23 +200,32 @@ function compileImportedFile({
...
     const providerConfiguration = getProviderConfigurationFromPath({ filePath: file.inputPath, parsed });
     if (!providerConfiguration) {
         return null;
     }
 
-    const syncConfig = [...providerConfiguration.syncs, ...providerConfiguration.actions].find((sync) => sync.name === file.baseName);
-    const type = syncConfig?.type || 'sync';
+    const config = [...providerConfiguration.syncs, ...providerConfiguration.actions].find((config) => config.name === file.baseName);
+
+    if (!config) {
```
bodinsamuel, line 233: it's wrongly skipping on-events scripts

packages/cli/lib/services/compile.service.ts
```diff
@@ -180,23 +200,32 @@ function compileImportedFile({
...
+        console.log(chalk.red(`Skipping compilation: No configuration found for ${file.baseName}`));
```
bodinsamuel: can you output the full path, it's hard to tell where those files are rn

packages/cli/lib/services/compile.service.ts
```diff
@@ -58,28 +71,36 @@ export async function compileAllFiles({
...
         }
     }
 
-    return success;
+    if (debug && compilationErrors.length > 0) {
+        printDebug('Compilation errors summary:');
+        compilationErrors.forEach((error) => printDebug(`- ${error}`));
+    }
+
+    return allSuccess;
```
bodinsamuel, line 102: can you take the opportunity to add a log when everything compiled, rn if there is an error and we fix it using `nango dev` it's not saying anything which is confusing 👍🏻

bodinsamuel (commented):
Works well, minor comments 👌🏻

packages/cli/lib/services/compile.service.ts
hassan254-prog: thanks, updated with the latest [commit](https://github.com/NangoHQ/nango/pull/3811/commits/13b5bdace4057b60dec82530d23934ca59a28cc4) 🙏

packages/cli/lib/services/compile.service.ts
hassan254-prog: @bodinsamuel, please let me know if the added logging is sufficient.

packages/cli/lib/services/compile.service.ts
```diff
@@ -180,23 +204,40 @@ function compileImportedFile({
...
+    }
+
     const providerConfiguration = getProviderConfigurationFromPath({ filePath: file.inputPath, parsed });
     if (!providerConfiguration) {
         return null;
     }
 
-    const syncConfig = [...providerConfiguration.syncs, ...providerConfiguration.actions].find((sync) => sync.name === file.baseName);
-    const type = syncConfig?.type || 'sync';
+    const syncsAndActions = [...providerConfiguration.syncs, ...providerConfiguration.actions];
```
TBonnin: you don't need this variable and pass both below: 
```
const config = [...providerConfiguration.syncs, ...providerConfiguration.actions, ...onEventScripts]...
```

TBonnin (approved):

packages/cli/lib/services/compile.service.ts
bodinsamuel: yes thanks a lot

bodinsamuel (approved):

PR #3814, opened 2025-03-31
feat(webapp): remember last environment

hassan254-prog:

- Remember the last environment when one visits the root path.


packages/webapp/src/components/AutoRedirect.tsx
```diff
@@ -0,0 +1,26 @@
...
+import storage, { LocalStorageKeys } from '../utils/local-storage';
+
+export const AutoRedirect: FC = () => {
+    const navigate = useNavigate();
+    const location = useLocation();
+    const env = useStore((state) => state.env);
+
+    useEffect(() => {
+        if (env) {
+            storage.setItem(LocalStorageKeys.LastEnvironment, env);
```
bodinsamuel: useLocalStorage would be nice (not that this class is bad but it's one less thing to maintain)

packages/webapp/src/components/AutoRedirect.tsx
```diff
@@ -0,0 +1,26 @@
+import type { FC } from 'react';
```
bodinsamuel: Not sure why you added this instead of using Root, it's a bit less meaningful. Also you didn't deleted Root and you forgot to keep /getting-started redirect if dev

bodinsamuel (commented):
Thanks, it's been long since I wanted to do that

packages/webapp/src/components/AutoRedirect.tsx
hassan254-prog: yes 🤦, thanks @bodinsamuel! I've updated this with the latest [commit](https://github.com/NangoHQ/nango/pull/3814/commits/a14da965122d37a6e8df72b582d3a6ba0b2ff9e5)

bodinsamuel (commented):
Much simpler thanks :D 
One remaining problem is the initial redirect to Getting Started, it should always redirect to this page if the onboarding is not done and env === dev

bodinsamuel (approved):

PR #3815, opened 2025-03-31
chore: webhook can execute for 1 hour (15mins previously)

TBonnin:
to accomodate a customer who need longer running webhooks.

This commit also adds heartbeat for webhooks (since there are now running for longer) and actions (not strictly necessary but would help failing faster)


bodinsamuel (approved):
😮‍💨

PR #3819, opened 2025-04-01
chore: add billed records metrics

TBonnin:
Latest iteration of the records billing is as follow: The same record is accounted for billing only once during a calendar month, which means:
- inserted records are billed
- updated records are billed if they were not inserted/updated in the same calendar month
- deleted records are billed if they were not inserted/updated in the same calendar month
- undeleted records are billed if they were not inserted/updated in the same calendar month

### DIFF
--- a/packages/persist/lib/records.ts
+++ b/packages/persist/lib/records.ts
@@ -133,6 +133,7 @@ export async function persistRecords({
             return acc;
         }, 0);
 
+        metrics.increment(metrics.Types.BILLED_RECORDS_COUNT, new Set(summary.billedKeys).size, { accountId });
         metrics.increment(metrics.Types.PERSIST_RECORDS_COUNT, records.length, { accountId });
         metrics.increment(metrics.Types.PERSIST_RECORDS_SIZE_IN_BYTES, recordsSizeInBytes, { accountId });
         metrics.increment(metrics.Types.PERSIST_RECORDS_MODIFIED_COUNT, allModifiedKeys.size, { accountId });
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -405,13 +405,12 @@ export async function upsert({
                 // A record is billed only once per month. ie:
                 // - If a record is inserted, it is billed
                 // - If a record is updated, it is billed if it has not been billed yet during the current month
-                // - If a record is undeleted, it is billed if it has not been billed yet during the current month
-                // - If a record is deleted, it is billed if it has not been billed yet during the current month
+                // - If a record is undeleted, it is not billed
+                // - If a record is deleted, it is not billed
 
                 if (softDelete) {
                     const deleted = res.filter((r) => r.status === 'deleted');
                     summary.deletedKeys?.push(...deleted.map((r) => r.external_id));
-                    summary.billedKeys.push(...billable(deleted).map((r) => r.external_id));
                 } else {
                     const undeletedRes = res.filter((r) => r.status === 'undeleted');
                     const changedRes = res.filter((r) => r.status === 'changed');
@@ -420,7 +419,7 @@ export async function upsert({
                     const undeletedKeys = undeletedRes.map((r) => r.external_id);
                     const addedKeys = insertedKeys.concat(undeletedKeys);
                     const updatedKeys = changedRes.map((r) => r.external_id);
-                    const billableKeys = [...insertedKeys, ...billable([...changedRes, ...undeletedRes]).map((r) => r.external_id)];
+                    const billableKeys = [...insertedKeys, ...billable(changedRes).map((r) => r.external_id)];
 
                     summary.addedKeys.push(...addedKeys);
                     summary.updatedKeys.push(...updatedKeys);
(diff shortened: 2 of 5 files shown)


packages/records/lib/models/records.ts
```diff
@@ -375,21 +391,36 @@ export async function upsert({
...
-                const deleted = updatedRes.filter((r) => r.status === 'deleted').map((r) => r.external_id);
-                const updated = updatedRes.filter((r) => r.status === 'changed').map((r) => r.external_id);
+                const res = await withRetry(query);
+
+                // Billing:
+                // A record is billed only once per month. ie:
+                // - If a record is inserted, it is billed
+                // - If a record is updated, it is billed if it has not been billed yet during the current month
+                // - If a record is undeleted, it is billed if it has not been billed yet during the current month
+                // - If a record is deleted, it is billed if it has not been billed yet during the current month
```
TBonnin: cc @bastienbeurier to make sure I correctly understood the requirements

packages/persist/lib/records.ts
```diff
@@ -133,6 +133,7 @@ export async function persistRecords({
             return acc;
         }, 0);
 
+        metrics.increment(metrics.Types.BILLED_RECORDS_COUNT, new Set(summary.billedKeys).size, { accountId });
```
TBonnin, line 136: @bastienbeurier do we still need the records count/size and modified records count/size for billing and platform usage?
If not we could remove the one we don't need to lower our datadog bill

packages/records/lib/models/records.ts
bastienbeurier: Thanks for pinging me on the details of this. This is important!

 I would only bill for:

> inserted records are billed
> updated records are billed if they were not inserted/updated in the same calendar month

I would not bill for:

> deleted records are billed if they were not inserted/updated in the same calendar month

I know we have costs when records are deleted. But I think it would be misleading if a customer deleted a connection and was billed for all records for that connection. There would basically be some kind of exit tax for Nango, which I don't think is a good pattern.

> undeleted records are billed if they were not inserted/updated in the same calendar month

This looks like unmodified records to me, so I would not bill for this.

packages/persist/lib/records.ts
bastienbeurier: Good point. 

Can we just keep:
- `Monthly modified records`
- `Records in (size in bytes)`

 and remove: 
- `Records out (size in bytes)`
- `Records in (count)`
- `Records out (count)`
- `Records Modified (size in bytes)`
- `Records Modified (count)`
- `Records total count (snapshot)`

bodinsamuel: Happy to keep all the metrics except "Records total count (snapshot)" without the account_id dimension, that should only be 6 metrics that can be helpful to debug

packages/records/lib/models/records.ts
```diff
@@ -545,6 +577,17 @@ export async function update({
...
+                    const firstUpdateThisMonth = updated.filter((r) => {
+                        const firstDayOfMonth = dayjs().startOf('month').toDate();
+                        const oldRecord = oldRecords.find((old) => old.external_id === r.external_id);
+                        if (!oldRecord) {
+                            return false;
+                        }
+                        const previousLastModifiedAt = dayjs(oldRecord.updated_at);
+                        return previousLastModifiedAt.isBefore(firstDayOfMonth);
+                    });
+                    billedKeys.push(...firstUpdateThisMonth.map((r) => r.external_id));
```
bodinsamuel: why not do a regular for loop that pushes directly to billedKeys? you are iterating twice over the array rn

bodinsamuel (approved):

packages/records/lib/models/records.ts
TBonnin: > I know we have costs when records are deleted. But I think it would be misleading if a customer deleted a connection and was billed for all records for that connection. There would basically be some kind of exit tax for Nango, which I don't think is a good pattern.

this logic is only happening on batchSave/batchUpdate, not when deleting a connection

> This looks like unmodified records to me, so I would not bill for this.
Functionally speaking undeleting a record corresponds to an insertion, I don't see why we wouldn't bill it the same way

bastienbeurier: Still I would focus on inserting/edition!

TBonnin: I don't understand why undelete shouldn't be treating as an insertion and deletion as an update but up to you

packages/persist/lib/records.ts
TBonnin: keeping the metrics as is for now. Let's revisit after DD next bill

packages/records/lib/models/records.ts
bastienbeurier: Ah sorry, I think I misinterpreted what "undeleted" means. I thought it designated the records that were untouched (not deleted nor edited). But if "undeleted" means taking a record from a deleted state to a normal state, then yes, this should be counted as an insertion and billed.

packages/persist/lib/records.ts
bastienbeurier: I can already tell you the bill is bad :D

PR #3822, opened 2025-04-01
feat(plan): create free plan for new account

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-3011/create-plan-on-account-creation
Split of #3803

- Create free plan for new account

## 🧪 Tests
- Set `FLAG_PLAN_ENABLED` to true in your .env
- Create a new account
- It should have created a row in table plans
- It's not used yet


packages/shared/lib/services/plans/plans.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export async function getPlan(db: Knex, { accountId }: { accountId: number }): Promise<DBPlan | null> {
+    const res = await db.from<DBPlan>('plans').select<DBPlan>('*').where('account_id', accountId).first();
+    return res || null;
+}
+
+export async function createPlan(
+    db: Knex,
+    { account_id, ...rest }: Pick<DBPlan, 'account_id' | 'name'> & Partial<Omit<DBPlan, 'account_id' | 'created_at' | 'updated_at'>>
+): Promise<DBPlan> {
```
TBonnin: I would have all public functions to return a `Result` and have a try catch so one doesn't "forget" to handle errors. I know that's not how all the model services have been written but for new ones we should aim for better error handling

packages/shared/lib/services/plans/plans.ts
```diff
@@ -0,0 +1,50 @@
...
+}
+
+export async function updatePlan(db: Knex, { id, ...data }: Pick<DBPlan, 'id'> & Partial<Omit<DBPlan, 'id'>>) {
+    await db
+        .from<DBPlan>('plans')
+        .where('id', id)
+        .update({ ...data, updated_at: new Date() });
+}
+
+export async function getTrialCloseToFinish(db: Knex, { inDays }: { inDays: number }): Promise<DBPlan[]> {
```
TBonnin: nit: suggestion for the function name: `function getTrialsApproachingExpiration(db: Knex, { daysLeft }: { daysLeft: number })`

packages/shared/lib/services/account.service.ts
```diff
@@ -84,15 +88,19 @@ class AccountService {
...
 
-        if (result[0]?.id) {
-            await environmentService.createDefaultEnvironments(result[0].id);
+        if (!result[0]) {
+            return null;
+        }
 
-            return result[0];
+        await environmentService.createDefaultEnvironments(result[0].id);
+        if (flagHasPlan) {
```
TBonnin, line 99: why the flag since the plans table is not being used yet

packages/shared/lib/services/account.service.ts
```diff
@@ -84,15 +88,19 @@ class AccountService {
...
+            await createPlan(db.knex, { account_id: result[0].id, name: 'free' });
```
TBonnin: should we log something if this fails?

packages/shared/lib/services/account.service.ts
bodinsamuel: it's for enterprise and self-hosted
but the plans table will be used on cloud after this PR

packages/shared/lib/services/account.service.ts
bodinsamuel: reported

packages/shared/lib/services/plans/plans.ts
bodinsamuel: changed

packages/shared/lib/services/plans/plans.ts
bodinsamuel: changed

packages/shared/lib/services/plans/plans.ts
```diff
@@ -0,0 +1,71 @@
...
+                trial_end_at: new Date(Date.now() + TRIAL_DURATION),
+                ...rest,
+                created_at: new Date(),
+                updated_at: new Date(),
+                account_id
+            })
+            .onConflict('account_id')
+            .ignore()
+            .returning('*');
+        return Ok(res[0]!);
```
TBonnin, line 38: you can do `const [res] = ...` above and have a more pleasing to the eye `return(Ok(res))` 😁

TBonnin (approved):

PR #3825, opened 2025-04-01
feat(plan): replace is_capped

bodinsamuel:
## Changes

- Replace usage of `is_capped` by plan flags
I replaced almost everything except rate limit. There are also two spots where I need to hardcode the plan name which is not good, I should probably add a boolean for those features (e.g: create environment)

- Update all integration tests that were relying on the default account that does not have a plan


packages/server/lib/controllers/connection.controller.ts
```diff
@@ -158,11 +160,12 @@ class ConnectionController {
 
             const providerName = integration.provider;
 
-            if (account.is_capped && provider_config_key) {
+            if (plan && plan.connection_with_scripts_max && provider_config_key) {
```
TBonnin: why not `plan?.connection_with_scripts_max` btw?

packages/server/lib/controllers/v1/flows/preBuilt/postDeploy.ts
```diff
@@ -45,8 +48,13 @@ export const postPreBuiltDeploy = asyncWrapper<PostPreBuiltDeploy>(async (req, r
...
 
-    if (account.is_capped) {
-        const isCapped = await connectionService.shouldCapUsage({ providerConfigKey: body.providerConfigKey, environmentId, type: 'deploy' });
+    if (plan && plan.connection_with_scripts_max) {
+        const isCapped = await connectionService.shouldCapUsage({
+            providerConfigKey: body.providerConfigKey,
+            environmentId,
+            type: 'deploy',
+            limit: plan.connection_with_scripts_max
+        });
```
TBonnin, line 57: what about passing the plan to `shouldCapUsage` and let it have all the logic in there

packages/server/lib/middleware/access.middleware.ts
```diff
@@ -97,16 +110,28 @@ export class AccessMiddleware {
...
         if (!result) {
             return Err('unknown_user_account');
         }
-        return Ok(result);
+
+        let plan: DBPlan | null = null;
+        if (flagHasPlan) {
+            const planRes = await getPlan(db.knex, { accountId: result.account.id });
+            if (planRes.isErr()) {
+                return Err('invalid_plan');
```
TBonnin: would `plan_not_found` be a better error name?

packages/server/lib/middleware/ratelimit.middleware.ts
```diff
@@ -84,9 +84,11 @@ function getPointsToConsume(req: Request, res: Response<any, RequestLocals>): nu
     if (specialPaths.includes(fullPath)) {
         // limiting to 6 requests per period to avoid brute force attacks
         return Math.floor(rateLimiter.points / 6);
-    } else if (!res.locals.account || res.locals.account.is_capped) {
+    } else if (!res.locals.account || (res.locals.plan && res.locals.plan.name === 'free')) {
+        // TODO: maybe store this in plans
```
TBonnin, line 88: +1 to store in the plan

packages/utils/lib/telemetry/tracer.ts
```diff
@@ -1,10 +1,12 @@
-import type { DBEnvironment, DBTeam } from '@nangohq/types';
 import tracer from 'dd-trace';
 
+import type { DBEnvironment, DBTeam } from '@nangohq/types';
+
 export function tagTraceUser({ account, environment }: { account: DBTeam; environment: DBEnvironment }) {
     tracer.setUser({
         id: String(account.id),
         environmentId: String(environment.id),
+        // TODO: use plan
```
TBonnin, line 9: is it too complicated to add in this PR?

TBonnin (approved):
small comments but lg othewise

bodinsamuel:
note to me: I forgot otlp

packages/utils/lib/telemetry/tracer.ts
bodinsamuel: The PR is already big and we don't have the plan everywhere it's used (jobs, persist), I'll do it just after,

packages/server/lib/controllers/connection.controller.ts
bodinsamuel: Changed

packages/server/lib/controllers/v1/flows/preBuilt/postDeploy.ts
bodinsamuel: Changed

packages/server/lib/middleware/access.middleware.ts
bodinsamuel: Changed

packages/server/lib/middleware/ratelimit.middleware.ts
bodinsamuel: I'll do that in an other PR too

bodinsamuel:
Tested on staging

PR #3826, opened 2025-04-01
docs: update .env.example

steliosmavro:
### Problem

While trying to run Nango locally, I ran into a few issues due to missing or misconfigured environment variables in the .env.example file. These included:

- Missing AWS credentials, which caused errors when trying to enable and deploy syncs (e.g., file upload errors).

- Missing NANGO_PUBLIC_CONNECT_URL, which led to validation errors on startup.

- MAILGUN_API_KEY being set to an empty string caused unexpected behavior (e.g., unauthorized API errors instead of falling back to console logging).

### Fixes

- Added placeholders for `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`

- Set `AWS_REGION=us-west-2` as the default used internally

- Set a default `NANGO_PUBLIC_CONNECT_URL=http://localhost:3009` for local dev

- Commented out `MAILGUN_API_KEY` to avoid authorization issues when unset

### Suggestion

An additional code-level improvement that could prevent similar issues: 

Instead of:
```
process.env['MAILGUN_API_KEY'] === undefined
```

Use:

```
!process.env['MAILGUN_API_KEY']
```

It would also handle the case where the env variable is defined but empty.


.env.example
```diff
@@ -108,7 +108,12 @@ ORCHESTRATOR_SERVICE_URL="http://localhost:3008"
...
-NANGO_PUBLIC_CONNECT_URL=
+NANGO_PUBLIC_CONNECT_URL=http://localhost:3009
 # Only relevant for self-hosted
 NANGO_CONNECT_UI_PORT=3009
 FLAG_SERVE_CONNECT_UI=true
+
+# ---- AWS
+# AWS_ACCESS_KEY_ID=<ACCESS-KEY-ID>
+# AWS_SECRET_ACCESS_KEY=<SECRET-ACCESS-KEY>
+AWS_REGION=us-west-2
```
bodinsamuel: I would remove this because it's a us centric default that might not fit everybody and not strictly necessary to have this env (unless I'm mistaken)
```suggestion
# AWS_REGION=us-west-2
```

bodinsamuel (commented):
Thanks 🙏🏻

.env.example
steliosmavro: You’re absolutely right. I tested it and it’s not necessary. Removed!

bodinsamuel (approved):

PR #3828, opened 2025-04-01
feat(api): add NANGO_ADMIN_PROD_ENV env var

kaposke:
The environment name will soon be editable, so I'm making this value editable via env var to make it clearer.

Part of [NAN-2685](https://linear.app/nango/issue/NAN-2685/editdelete-environments-in-the-ui)


packages/server/lib/controllers/connection.controller.ts
```diff
@@ -30,7 +32,7 @@ class ConnectionController {
 
             const integration_key = process.env['NANGO_SLACK_INTEGRATION_KEY'] || 'slack';
             const nangoAdminUUID = NANGO_ADMIN_UUID;
-            const env = 'prod';
+            const env = process.env['NANGO_ADMIN_PROD_ENV'] || 'prod';
```
TBonnin, line 35: we now prefer to use the validated and type safed `parsedEnvs`. Example: https://github.com/NangoHQ/nango/blob/5ea7299ab539532e1f5195b0f98d94cf96e3b618/packages/server/lib/server.ts#L33

bodinsamuel: yes, and that way you can put a default value too

packages/shared/lib/services/notification/slack.service.ts
```diff
@@ -85,7 +87,7 @@ export class SlackService {
 
     private integrationKey = process.env['NANGO_SLACK_INTEGRATION_KEY'] || 'slack';
     private nangoAdminUUID = process.env['NANGO_ADMIN_UUID'];
-    private env = 'prod';
+    private env = process.env['NANGO_ADMIN_PROD_ENV'] || 'prod';
```
bodinsamuel, line 90: nb in shared you wont be able to use parsed, but you can refactor the class to accept the env in the constructor

bodinsamuel (approved):
Looks good, just changing to envs and it will be fine

kaposke:
Don't think this is relevant anymore.

PR #3831, opened 2025-04-01
feat(webhooks): webhook env var injected primary URLs

kishenpateldotwork:
Problem:
We are self-hosting Nango and as part of the [public key deprecation](https://docs.nango.dev/guides/api-authorization/public-key-deprecation) we have set up webhooks but we do not want to have to [go into each customer instance to define the primary webhook URL](https://docs.nango.dev/guides/webhooks/webhooks-from-nango#configuring-nango-webhooks) and instead we would like to just configure it via environment variables (similar to how we can override the secret keys via env vars NANGO_SECRET_KEY_PROD and NANGO_PUBLIC_KEY_PROD).

Solution:
Within the existing external-webhook.service, obtain the environment from the database using the id, so we can use the name to, check if a related env var is defined for using as the primary url.

Testing:
Simply set env var for a named environment (e.g. NANGO_WEBHOOK_PRIMARY_URL_PROD=https://example_env_prod.com/webook_from_nango_prod) and then see if the UI display it and that any webhooks requests are sent to it accordingly.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -112,3 +112,7 @@ NANGO_PUBLIC_CONNECT_URL=
 # Only relevant for self-hosted
 NANGO_CONNECT_UI_PORT=3009
 FLAG_SERVE_CONNECT_UI=true
+
+# Override webhook primary URL for a specific environment
+# NANGO_WEBHOOK_PRIMARY_URL_PROD=https://example_env_prod.com/webook_from_nango_prod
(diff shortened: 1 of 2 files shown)


bodinsamuel:
Hi, Thanks for the PR. 
We don't really want to go this way, it's quite challenging to maintain this kind of env injection (we already have this legacy `NANGO_SECRET_KEY_` that is quite annoying). As stated during our discussion we'll eventually go with a public API to let you update env settings on the fly. 
Hope you'll understand, sorry about the trouble

PR #3833, opened 2025-04-02
fix(api): GET /api/v1/integrations new format

bodinsamuel:
## Changes

Contributes to https://linear.app/nango/issue/NAN-2109/finish-api-migrations-to-new-endpoint-format-and-validation

- GET /api/v1/integrations new format
Simplified the endpoint by pre-fetching active syncs per integration instead of doing N queries
Also changed a bit the output to avoid mixing the actual integration and additional metadata.

### DIFF
--- a/packages/webapp/package.json
+++ b/packages/webapp/package.json
@@ -27,7 +27,6 @@
         "@mantine/core": "7.12.1",
         "@mantine/prism": "5.10.5",
         "@nangohq/frontend": "0.58.1",
-        "@nangohq/server": "file:../server",
         "@nangohq/types": "file:../types",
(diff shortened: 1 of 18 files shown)


packages/webapp/package.json
```diff
@@ -27,7 +27,6 @@
         "@mantine/core": "7.12.1",
         "@mantine/prism": "5.10.5",
         "@nangohq/frontend": "0.58.1",
-        "@nangohq/server": "file:../server",
```
bodinsamuel, line 30: :feels-good:

TBonnin: 🫶

TBonnin (approved):

PR #3834, opened 2025-04-02
fix(test): fix failing tests for windows

hassan254-prog:

- This PR fix failing tests on Windows specifically. The issue occurred because files were only compiled in the first test (`should package correctly`), causing the second test (`should filter by integrationId when specified`) to fail due to missing compiled files. This was specific to Windows due to differences in file system handling. By compiling the files once, both tests can use the compiled files, resolving the issue.


TBonnin (approved):

hassan254-prog:
Maybe @viictoo, you could help test this as well? ☺️

bodinsamuel (approved):
Thanks a lot 🙏🏻

viictoo (commented):
Nicely done! 🙌
tested locally and all tests pass.
💪

viictoo (approved):

PR #3835, opened 2025-04-02
feat(es): meta search

bodinsamuel:
## Changes

Contributes to https://linear.app/nango/issue/NAN-2698/full-text-search-includes-payloads
Contributes to https://linear.app/nango/issue/NAN-2699/search-logs-by-record-id

- Introduce meta_search
Elasticsearch has this feature that allows copying fields to another field. This is more or less the same as doing that manually except it's not "stored", so it's only using storage for the index.
Using a global field allows us to search in all relevant fields with less impact than searching into all fields separately. 
I don't know how indexing and storage will be impacted, unfortunately.

- Changed the indexing of `meta` to allow storing some fields
This is mostly to store RecordIDs when indexing records, but we could need more.

## 🧪 Tests

This feature is not really used right now, I just want to test the storage and indexing in prod.
However it's been tested in staging 

**Example of what this field can contain**
Screenshot 2025-04-07 at 17 25 18


TBonnin (commented):
cannot really test but I trust you 😀 
I am interested to see the impact this has on ES storage

TBonnin (approved):
Screenshot 2025-04-08 at 10 49 48
Did anything change since yesterday?

bodinsamuel:
> Did anything change since yesterday?

No, I think you just commented, not approved

PR #3837, opened 2025-04-02
fix(scriptConfig): don't use disallowed character for the function name

khaliqgant:
The name is just the sync or action to allow direct calling of the sync/action

### DIFF
--- a/packages/server/lib/controllers/scripts/config/getScriptsConfig.ts
+++ b/packages/server/lib/controllers/scripts/config/getScriptsConfig.ts
@@ -59,7 +59,7 @@ export function transformToOpenAIFunctions(configs: StandardNangoConfig[]): Open
         // Process syncs - they don't have parameters
         for (const sync of config.syncs) {
             functions.push({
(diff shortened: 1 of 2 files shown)


bodinsamuel (approved):
Without the provider there could be some name conflict, you cant use dash or something else?

khaliqgant:
> Without the provider there could be some name conflict, you cant use dash or something else?

It doesn’t work fully. There will definitely be conflicts but with the description there will be differentiation. Also if someone goes this route of passing functions directly to an agent they’ll need to take care to give specific names to the action. This is something we’ll document

bodinsamuel:
> It doesn’t work fully. There will definitely be conflicts but with the description there will be differentiation. Also if someone goes this route of passing functions directly to an agent they’ll need to take care to give specific names to the action. This is something we’ll document

A human can differentiate, it's going to be hard for a program to do it automatically. 
just adding something like this could help
```
  "nango-integration-id": "hubspot",
```

PR #3838, opened 2025-04-02
fix(docker): cleanup

bodinsamuel:
## Changes 

- Deleted: server docker image
- Changed: self-hosted docker image is using the main docker image as a source 
The main benefit is that it compiles in just a few secs now
- Changed: self-hosted docker image is tagged with the version 
This means the version is rolling until we release another one
- Added: error reporting in UI 
Just because I realised it was missing from the migration to vite


.github/workflows/deploy.yaml
```diff
@@ -32,13 +32,7 @@ jobs:
         steps:
             - name: Checkout
               uses: actions/checkout@v4
-            - uses: docker/login-action@v3
-              with:
-                  username: ${{ secrets.DOCKER_USERNAME }}
-                  password: ${{ secrets.DOCKER_PASSWORD }}
-            - name: Push tag
-              run: |
-                  docker buildx imagetools create nangohq/nango-server:${{ github.sha }} --tag nangohq/nango-server:${{ inputs.stage }}
```
TBonnin, line 41: we should probably cleanup old images from docker hub.

Dockerfile.self_hosted
```diff
@@ -0,0 +1,12 @@
+
+ARG BASE_IMAGE_HASH
+FROM nangohq/nango:${BASE_IMAGE_HASH}
```
TBonnin, line 3: There is a bit of a implicit dependency but I guess it is fine since we only ever build images via the github action above.

scripts/build_docker_self_hosted.sh
```diff
@@ -0,0 +1,45 @@
...
+echo ""
+echo -e "Building self-hosted nangohq/nango-server:hosted-$GIT_HASH"
+
+VERSION=$(node -p "require('../package.json').version")
+
+docker buildx build \
+  --platform linux/amd64 \
+  --build-arg BASE_IMAGE_HASH="$GIT_HASH" \
+  --cache-from type=gha \
+  --cache-to type=gha,mode=max \
```
TBonnin, line 30: curious about those caching options

TBonnin (approved):
haven't tested but looks good

.github/workflows/deploy.yaml
bodinsamuel: what do you mean? they might still be used

scripts/build_docker_self_hosted.sh
bodinsamuel: it's the github actions cache, but this shit is barely working anyway

Dockerfile.self_hosted
bodinsamuel: I wouldn't call it implicit if it's the first line but yes it's no longer a standalone image

PR #3843, opened 2025-04-03
fix: display git hash in api and UI

bodinsamuel:
## Changes 

- Display git hash in API and UI
For convenience, since we don't increment the version at each commit. It's a bit complex to be sure which version is deployed without that


packages/server/lib/controllers/v1/getEnvJs.ts
```diff
@@ -1,19 +1,22 @@
...
         apiUrl: baseUrl,
         publicUrl: basePublicUrl,
         connectUrl: connectUrl,
-        publicSentryKey: process.env['PUBLIC_SENTRY_KEY'] || '',
-        publicPosthogKey: process.env['PUBLIC_POSTHOG_KEY'] || '',
-        publicPosthogHost: process.env['PUBLIC_POSTHOG_HOST'] || '',
-        publicLogoDevKey: process.env['PUBLIC_LOGODEV_KEY'] || '',
-        publicKoalaApiUrl: process.env['PUBLIC_KOALA_API_URL'] || '',
-        publicKoalaCdnUrl: process.env['PUBLIC_KOALA_CDN_URL'] || '',
+        gitHash: envs.GIT_HASH,
```
TBonnin, line 13: how is the env var set? would it make more sense to build a `version.json` static file that server could serve?

bodinsamuel: it's set in the docker image.

getEnvJs.ts is already a "static file" kind of, I don't see a big need to compile a file at build time but yes it's an alternative

TBonnin: can we have a comment stating when GIT_HASH is being set?

TBonnin (approved):

packages/server/lib/controllers/v1/getEnvJs.ts
bodinsamuel: It's already written in the parse file
https://github.com/NangoHQ/nango/blob/08dd308e52546014b0313575dd0e129a79c4a794/packages/utils/lib/environment/parse.ts#L15-L16

PR #3846, opened 2025-04-03
fix(connect-ui): show unique_key suffix when integration names are du…

steliosmavro:
## Issue

In Slack, a user reported that when having multiple integrations for the same provider (e.g., multiple Wildix PBXs), it's unclear which one to select from the Connect UI.
Only the provider name is shown, making it impossible to distinguish between them visually [Slack thread](https://nango-community.slack.com/archives/C04ABR352H0/p1743696112486929).

## Fix (Option 1 — this PR)

- When a `display_name` appears more than once, the UI appends the `unique_key` in parentheses after a dash. 
	- Example: `Dropbox - (dropbox-tmwr)`
- Non-duplicate integrations are left unchanged for visual cleanliness

## Alternative (Option 2 — separate PR)

There is a second PR available that applies a simpler approach: it **replaces** the `display_name` entirely with the `unique_key` for _all_ integrations, not just duplicates.

Feel free to choose whichever behaviour you prefer. Happy to adapt based on your preference!


bastienbeurier:
This works as a temporary fix (given it only shows when there's a duplicate). @bodinsamuel @TBonnin  What do you think?

TBonnin:
> This works as a temporary fix (given it only shows when there's a duplicate). @bodinsamuel @TBonnin What do you think?

I personally prefer this approach but I don't feel strongly either way

bodinsamuel (approved):
Thanks @SteliosMav 
Modified the code to fit the style but otherwise looks. Will release by the end of the day 💪🏻

PR #3848, opened 2025-04-04
fix(plan): use plan for tracer

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-3020/update-tracer-with-plan
Contributes to https://linear.app/nango/issue/NAN-3019/store-ratelimit-configuration-in-plan

- Use plan for tracer
It's only partially used, in jobs we don't really need it for the moment.

- Added new column to store api rate limit
I have a big question here, rate-limit-flexible does not allow points to be changed dynamically. What we currently do is increase/decrease the consumption based on some condition but it's not really precise. I have looked for ways to do it natively but there is no way. What I have done in the past is to create one class per customers, but not sure with our scale it would fit.

```ts
const limiters = new Map();
function getLimiterForCustomer(customerId, points, duration = 60) {
  if (!limiters.has(customerId)) {
    const limiter = new RateLimiterMemory({ points, duration });
    limiters.set(customerId, limiter);
  }

  return limiters.get(customerId);
}
```
@TBonnin


packages/persist/lib/middleware/auth.middleware.ts
```diff
@@ -28,18 +30,27 @@ export const authMiddleware = (req: Request, res: Response<any, AuthLocals>, nex
...
-            next();
-        })
-        .catch((err: unknown) => {
-            res.status(401).json({ error: `Unauthorized: ${stringifyError(err)}` });
-        });
+            plan = resPlan.value;
+        }
+
+        res.locals['account'] = accountAndEnv.account;
+        res.locals['environment'] = accountAndEnv.environment;
```
TBonnin, line 51: should the plan be saved in `locals` here?

TBonnin (commented):
> * I have a big question here, rate-limit-flexible does not allow points to be changed dynamically. What we currently do is increase/decrease the consumption based on some condition but it's not really precise. I have looked for ways to do it natively but there is no way. What I have done in the past is to create one class per customers, but not sure with our scale it would fit.

Instead of a limiter per customer a middle ground approach would be to have rate limit buckets, storing the bucket in the plan (ex: rate_limit_bucket = xs, s, m, l, xl, etc...) and then have small/fixed amount of limiters (ie: one per bucket).

```
const limiters: Record<RateLimitBucket, RateLimiter> = {
  xs: new RateLimiter({ points: 600, ... }),
  s: new RateLimiter({ points: 1200, ... },
  m: new RateLimiter({ points: 2400, ... },
  l: new RateLimiter({ points: 4800, ... },
  xl: new RateLimiter({ points: 9600, ... },
  ...
};

...
const limiter = limiters[plan.rateLimitBucket]
limiter.consume(...)
```

bodinsamuel:
> Instead of a limiter per customer a middle ground approach would be to have rate limit buckets, storing the bucket in the plan (ex: rate_limit_bucket = xs, s, m, l, xl, etc...) and then have small/fixed amount of limiters (ie: one per bucket).

Indeed could work, I think I'll do a mix of both. Stay with points so we can still have specific rate limit but reuse based on points.

```ts
const defPoint = process.env['DEFAULT_RATE_LIMIT_PER_MIN'];
const limiters = new Map({
  [defPoint]:  new RateLimiterMemory({ points: defPoint, duration });
});
function getLimiterForCustomer(customerId, points, duration = 60) {
  if (!limiters.has(points)) {
    const limiter = new RateLimiterMemory({ points, duration });
    limiters.set(points, limiter);
  }

  return limiters.get(customerId);
}
```
For the next PR 🚀

TBonnin:
> > Instead of a limiter per customer a middle ground approach would be to have rate limit buckets, storing the bucket in the plan (ex: rate_limit_bucket = xs, s, m, l, xl, etc...) and then have small/fixed amount of limiters (ie: one per bucket).
> 
> Indeed could work, I think I'll do a mix of both. Stay with points so we can still have specific rate limit but reuse based on points.

I would prefer to have a static and known-in-advance number of limiters to be honest. Your solution would be of course very fine at the beginning but without a hard dependency between what's in the db and the code the chance to end up with a large amount of limiters without even realizing is at term quite high.
Unless you want to enforce the bucketing at the db level and allow only 1000s as a value in the db

bodinsamuel:
Fair, changed

packages/database/lib/migrations/20250404142927_plans_rate_limit.cjs
```diff
@@ -2,12 +2,12 @@
  * @param {import('knex').Knex} knex
  */
 exports.up = async function (knex) {
-    await knex.raw(`ALTER TABLE "plans" ADD COLUMN "api_rate_limit_points" int4 NOT NULL DEFAULT 1000`);
+    await knex.raw(`ALTER TABLE "plans" ADD COLUMN "api_rate_limit_size" varchar(4) NOT NULL DEFAULT 'm'`);
```
TBonnin, line 5: I would go for default = `xs` which would be the lowest limit you get when not paying, and then go up for different level of paying customers

packages/utils/lib/telemetry/tracer.ts
```diff
@@ -1,12 +1,12 @@
...
+import type { DBEnvironment, DBPlan, DBTeam } from '@nangohq/types';
 
-export function tagTraceUser({ account, environment }: { account: DBTeam; environment: DBEnvironment }) {
+export function tagTraceUser({ account, environment, plan }: { account: DBTeam; environment: DBEnvironment; plan?: DBPlan | null }) {
     tracer.setUser({
         id: String(account.id),
         environmentId: String(environment.id),
-        // TODO: use plan
-        paying: account.is_capped ? 'free' : 'paying'
+        paying: plan?.name !== 'free' ? 'paying' : 'free',
```
TBonnin, line 9: we can have this condition in DD. not sure we need to send a extra tag

TBonnin (approved):
one comment about the default size but ltgm

packages/database/lib/migrations/20250404142927_plans_rate_limit.cjs
bodinsamuel: setting xs means we don't have a way to lower some users, we go to 99xl so I think we are good for paying

packages/utils/lib/telemetry/tracer.ts
bodinsamuel: we are not paying per dimension, it's easier to have this boolean than to rewrite the condition N times imo

packages/database/lib/migrations/20250404142927_plans_rate_limit.cjs
TBonnin: up to you but we could do the same with`2xs, 3xs, etc...`

PR #3849, opened 2025-04-04
feat: rename and delete environments

kaposke:
### Description
 
This PR adds the ability to rename and delete environments. The `prod` environment is a special case, where you can't do either.

Specific changes:
- /environment-settings
  - Added environment name field (On prod environment, editing is blocked with an informative tooltip)
  - Added delete button (doesn't show up in prod env)
    image
  - Created `DestructiveActionModal`, which requires a confirmation keyword for it's action to occur. Used it for deleting environments:
    image

- Environment Picker
  - Block state for `Create environment` button based on `environments_max` feature flag.
  - Different tooltip message for scale plan.
    image
  - Removed message from create environment dialog that said only the prod environment is billed.

### Ticket

[NAN-2576](https://linear.app/nango/issue/NAN-2576/locking-in-scripts-part-i)

### Testing

Please checkout into this branch and verify that all the changes described above are true. Create new environments, delete them, reach the limits.


packages/database/lib/migrations/20250402205913_cascade_environment_dependencies.cjs
```diff
@@ -0,0 +1,31 @@
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`ALTER TABLE "_nango_environment_variables" DROP CONSTRAINT "_nango_environment_variables_environment_id_foreign",
```
TBonnin, line 5: I don't see a `_nango_environment_variables_environment_id_foreign` currently on `_nango_environment_variables` table

packages/database/lib/migrations/20250402205913_cascade_environment_dependencies.cjs
```diff
@@ -0,0 +1,31 @@
...
+        FOREIGN KEY ("environment_id") 
+        REFERENCES "_nango_environments" ("id") 
+        ON DELETE CASCADE`);
+
+    await knex.raw(`ALTER TABLE "_nango_configs" DROP CONSTRAINT "_nango_configs_environment_id_foreign",
+        ADD CONSTRAINT "_nango_configs_environment_id_foreign" 
+        FOREIGN KEY ("environment_id") 
+        REFERENCES "_nango_environments" ("id") 
+        ON DELETE CASCADE`);
+};
```
TBonnin, line 16: since those constraints didn't exist before. there is the chance the data in the prod db are "dirty" and doesn't allow this constraints to be added, therefore failing the migration.

packages/server/lib/controllers/v1/environment/deleteEnvironment.ts
```diff
@@ -0,0 +1,29 @@
...
+export const deleteEnvironment = asyncWrapper<DeleteEnvironment>(async (req, res) => {
+    const emptyQuery = requireEmptyQuery(req, { withEnv: true });
+    if (emptyQuery) {
+        res.status(400).send({ error: { code: 'invalid_query_params', errors: zodErrorToHTTP(emptyQuery.error) } });
+        return;
+    }
+
+    const { environment } = res.locals;
+
+    if (environment.name === PROD_ENVIRONMENT_NAME) {
```
TBonnin, line 17: I would have preferred to add a column to the environment table called `isDeletable` rather than relying on the name. I wouldn't be surprise if one day customers ask for environment protection

packages/server/lib/controllers/v1/environment/patchEnvironment.ts
```diff
@@ -36,9 +38,18 @@ export const patchEnvironment = asyncWrapper<PatchEnvironment>(async (req, res)
     }
 
     const body: PatchEnvironment['Body'] = valBody.data;
-    const { environment, plan } = res.locals;
+    const { account, environment, plan } = res.locals;
 
     const data: Partial<DBEnvironment> = {};
+    if (typeof body.name !== 'undefined') {
```
TBonnin: can we simply `if(!body.name)` and enforce a min length of 1 anyway for the env name?

packages/shared/lib/services/environment.service.ts
```diff
@@ -479,6 +483,11 @@ class EnvironmentService {
 
         return globalCallbackUrl;
     }
+
+    async deleteEnvironment(id: number): Promise<boolean> {
+        const result = await db.knex.from<DBEnvironment>(TABLE).where({ id }).del();
```
TBonnin: this can throw. I know no other functions in this file is handling exception but I think this is a pattern we should move away from. I would recommend we write a line in the sand and always handle exception at locally as possible. At least can we agree that db-related functions should deal with knex exception and return a `Promise<Result<...>>`? wdyt? @kaposke @bodinsamuel

packages/webapp/src/components/CreateEnvironmentButton.tsx
```diff
@@ -0,0 +1,103 @@
...
+    };
+
+    const isMaxEnvironmentsReached = meta?.environments && meta?.plan && meta?.environments.length >= meta?.plan?.environments_max;
+    let tooltipContent: React.ReactNode = null;
+    if (isMaxEnvironmentsReached) {
+        tooltipContent = (
+            <span>
+                Max number of environments reached.{' '}
+                {meta?.plan?.name === 'scale' ? (
+                    <>Contact us to add more</>
```
TBonnin: `Contact Nango to add more`. no?

packages/webapp/src/components/CreateEnvironmentButton.tsx
```diff
@@ -0,0 +1,103 @@
...
+                ) : (
+                    <>
+                        <Link to="https://app.withsurface.com/s/cm1zve3340001l503sm0xtvo1" className="underline">
```
TBonnin, line 58: @bastienbeurier url is not branded? it looks very phishy

packages/webapp/src/pages/Environment/Settings/DeleteAlert.tsx
```diff
@@ -0,0 +1,34 @@
...
+            <IconTrash stroke={1} size={18} />
+            <span>Delete environment</span>
+        </Button>
+    );
+
+    return (
+        <DestructiveActionModal
+            title="Proceed carefully!"
+            description="This action is destructive & irreversible. It will delete all API credentials, connection metadata, synced records & various configurations linked to this environment."
+            inputLabel="To confirm, type your current environment's name below:"
```
TBonnin: should the label include the name of the environment? ex: `To confirm, type your current environment's name (_the_env_name_) below:`

packages/webapp/src/pages/Environment/Settings/Main.tsx
```diff
@@ -0,0 +1,41 @@
...
+            </Link>
+            <div className="px-8 flex flex-col gap-10 w-3/5">
+                <EditableInput
+                    title="Environment Name"
+                    name="environmentName"
+                    originalValue={env}
+                    apiCall={(name) => apiPatchEnvironment(env, { name })}
+                    onSuccess={async (newName) => {
+                        // We have to start by changing the url, otherwise PrivateRoute will the env revert it based on it.
+                        window.history.replaceState({}, '', `/${newName}/environment-settings`);
```
TBonnin: I have noticed a flickering when saving the new name. It doesn't happen all the time though. I wonder if it is because of this

TBonnin (commented):
Thank you for this PR. Tested it locally. It worked well, I sometimes notice a flickering when renaming a env. 
I'll let Samuel fully approve the frontend stuff but lgtm otherwise

packages/webapp/src/components/CreateEnvironmentButton.tsx
kaposke: Went by the design, but fine with either.

packages/webapp/src/pages/Environment/Settings/DeleteAlert.tsx
kaposke: Also faithful to the design, but yeah, showing it there does make sense

packages/server/lib/controllers/v1/environment/patchEnvironment.integration.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+    it('should not allow renaming to an existing environment name', async () => {
+        const { env, account } = await seeders.seedAccountEnvAndUser();
+        await environmentService.createEnvironment(account.id, 'existing');
+
+        // Try to rename to 'test' which already exists from the seeder
+        const res = await api.fetch(endpoint, {
+            method: 'PATCH',
+            // @ts-expect-error query params are required
+            query: { env: 'test' },
```
TBonnin: `query: { env: env.name}` so we don't have to worry if the seed env name is ever modified

packages/webapp/src/pages/Environment/Settings/Main.tsx
kaposke: Yeah, this part took me a long time to even get working because we have a useEffect in the `PrivateRoute` that does many things. I avoided changing that because I don't know all the side effects it could cause or the reasons for some things there. Maybe @bodinsamuel can shed some light on this.

packages/shared/lib/services/environment.service.ts
kaposke: Makes sense to me

packages/server/lib/controllers/v1/environment/patchEnvironment.ts
kaposke: Good call

packages/server/lib/controllers/v1/environment/deleteEnvironment.ts
kaposke: We also considered flagging "production ready" or something similar. Ended up going with the easiest while the product definition isn't 100% clear. Fine with adding the column though

packages/database/lib/migrations/20250402205913_cascade_environment_dependencies.cjs
kaposke: I do see a foreign key in prod, and in my local after seeding as well. That seems to be the default name schema given to fks. 

This is the prod replica:
image

packages/server/lib/controllers/v1/environment/patchEnvironment.ts
kaposke: Actually, it's validated by zod already. But simplified the condition

packages/shared/lib/services/environment.service.ts
bodinsamuel: I don't like it, but yes let's be consistent 👍🏻

packages/webapp/src/store.ts
```diff
@@ -18,8 +20,8 @@ interface State {
 }
 
 export const useStore = create<State>((set, get) => ({
-    env: 'dev',
-    envs: [{ name: 'dev' }, { name: 'prod' }],
+    env: PROD_ENVIRONMENT_NAME,
+    envs: [{ name: PROD_ENVIRONMENT_NAME }],
```
bodinsamuel: I think this breaks the default behavior "Go to dev and show getting started".
We might want to hold longer until we have loaded /meta

packages/webapp/src/pages/Environment/Settings/Show.tsx
```diff
@@ -36,6 +48,26 @@ export const EnvironmentSettings: React.FC = () => {
         element.scrollIntoView({ behavior: 'smooth' });
     }, [environmentAndAccount]);
 
+    const handleDelete = async () => {
+        const { res } = await apiDeleteEnvironment(env);
+        if (res.status >= 200 && res.status < 300) {
+            setShowDeleteAlert(false);
+            // We have to start by changing the url, otherwise PrivateRoute will the env revert it based on it.
+            window.history.replaceState({}, '', `/${PROD_ENVIRONMENT_NAME}/environment-settings`);
```
bodinsamuel: You can use `const navigate = useNavigate()` that will be more "native"

packages/webapp/src/pages/Environment/Settings/Main.tsx
bodinsamuel: The flickering is coming from the fact that we are actually loading a new env from the API so the loading is triggered again while we fetch the new payload. 
Since we conflate id and display name it's hardly avoidable `packages/webapp/src/pages/Environment/Settings/Show.tsx:71`.

packages/webapp/src/pages/Environment/Settings/DeleteAlert.tsx
```diff
@@ -0,0 +1,34 @@
+import { IconTrash } from '@tabler/icons-react';
```
bodinsamuel: nit: Technically it's a DeleteButton

packages/webapp/src/constants.ts
```diff
@@ -0,0 +1 @@
+export const PROD_ENVIRONMENT_NAME = 'prod';
```
bodinsamuel, line 1: it's not compatible with your other PR #3828 
use getEnvJs to forward this dynamic env var

packages/shared/lib/services/environment.service.ts
```diff
@@ -479,6 +495,19 @@ class EnvironmentService {
...
+            const result = await db.knex.from<DBEnvironment>(TABLE).where({ id }).del();
+            if (result !== 0) {
+                return Ok(undefined);
+            }
+        } catch {
+            // Empty catch block
+        }
+        // Error if no environment was found OR exception was thrown
+        return Err(new EnvironmentError({ code: 'deletion_failed', message: 'Failed to delete environment' }));
+    }
```
bodinsamuel: We actually want to differentiate the two for error reporting. 

```suggestion
        try {
            const result = await db.knex.from<DBEnvironment>(TABLE).where({ id }).del();
            if (result !== 0) {
                return Ok(undefined);
            }
             // Error if no environment was found OR exception was thrown
             return Err(new Error('blabla'));
        } catch (err) {
            return Err(new Error('blabla', { cause: err }), 
        }
       
    }
```

Also we don't want to introduce `EnvironmentError` rn. I appreciate the initiative, and it's probably going to look similar to that at some point we are actually trying to have global revamp of the error and it's a bit too soon. Using generic Error should be fine

packages/server/lib/controllers/v1/meta/getMeta.ts
```diff
@@ -10,12 +13,13 @@ export const getMeta = asyncWrapper<GetMeta>(async (req, res) => {
...
     }
 
-    const sessionUser = res.locals['user'];
+    const { user: sessionUser, plan } = res.locals;
 
     const environments = await environmentService.getEnvironmentsByAccountId(sessionUser.account_id);
     const onboarding = await getOnboarding(sessionUser.id);
     res.status(200).send({
         data: {
+            plan: plan ? planToApi(plan) : null,
```
bodinsamuel, line 22: it's a bit counter-intuitive but the plan is returned in another endpoint
https://github.com/NangoHQ/nango/blob/master/packages/server/lib/controllers/v1/environment/getEnvironment.ts
This endpoint should disappear at some point (hopefully)

packages/server/lib/middleware/access.middleware.ts
```diff
@@ -505,6 +505,17 @@ async function fillLocalsFromSession(req: Request, res: Response<any, RequestLoc
...
+        let plan: DBPlan | null = null;
+        if (flagHasPlan) {
+            const planRes = await getPlan(db.knex, { accountId: user.account_id });
+            if (planRes.isErr()) {
+                res.status(401).send({ error: { code: 'plan_not_found' } });
+                return;
+            }
+            plan = planRes.value;
+        }
+        res.locals['plan'] = plan;
```
bodinsamuel, line 517: What does it fixes?

packages/server/lib/controllers/v1/environment/patchEnvironment.ts
```diff
@@ -36,9 +38,18 @@ export const patchEnvironment = asyncWrapper<PatchEnvironment>(async (req, res)
...
 
     const body: PatchEnvironment['Body'] = valBody.data;
-    const { environment, plan } = res.locals;
+    const { account, environment, plan } = res.locals;
 
     const data: Partial<DBEnvironment> = {};
+    if (body.name) {
+        const existingEnvironment = await environmentService.getByEnvironmentName(account.id, body.name);
+        if (existingEnvironment && existingEnvironment.id !== environment.id) {
+            res.status(400).send({ error: { code: 'conflict', message: 'An environment with this name already exists' } });
```
bodinsamuel: code should be `409`

packages/database/lib/migrations/20250402205913_cascade_environment_dependencies.cjs
bodinsamuel: it's the same constraint, he is just adding on delete cascade

packages/server/lib/controllers/v1/environment/deleteEnvironment.ts
```diff
@@ -0,0 +1,29 @@
...
+        res.status(400).send({ error: { code: 'cannot_delete_prod_environment', message: 'Cannot delete prod environment' } });
+        return;
+    }
+
+    const deleted = await environmentService.deleteEnvironment(environment.id);
+    if (!deleted) {
```
bodinsamuel: wrong condition since you have changed to Result

bodinsamuel (commented):
Works well for the rename 💪🏻 
The delete doesn't work because of Result but also because the cascade is not complete, it's probably missing from a few tables unfortunately (e.g: connections).

I would also maybe push back a little on using direct delete and not soft delete. It's annoying to manage soft delete but a customer deleting the wrong environment will bite us very quickly. Plus Cascade only works very small amount of data so it's better to manage the cascading manually (see cron job).
@bastienbeurier 

NB: you could have split the PR (and you still can) to avoid blocking this feature

packages/webapp/src/pages/Environment/Settings/DeleteAlert.tsx
```diff
@@ -0,0 +1,34 @@
...
+            description="This action is destructive & irreversible. It will delete all API credentials, connection metadata, synced records & various configurations linked to this environment."
+            inputLabel={`To confirm, type your current environment's (${environmentName}) name below:`}
+            confirmationKeyword={environmentName}
+            confirmButtonText="Delete Environment"
+            trigger={trigger}
+            onConfirm={onDelete}
+            open={open}
+            onOpenChange={onOpenChange}
+        />
+    );
```
bodinsamuel: Need to handle long name
Screenshot 2025-04-07 at 10 49 13

packages/webapp/src/components/CreateEnvironmentButton.tsx
kaposke: (Updated)

packages/webapp/src/pages/Environment/Settings/DeleteAlert.tsx
kaposke: (Added)

packages/server/lib/controllers/v1/meta/getMeta.ts
kaposke: The plan being returned in `getEnvironment` is weirder to me

packages/server/lib/middleware/access.middleware.ts
kaposke: Plan was not added for the paths ignored in https://github.com/NangoHQ/nango/blob/aa711b3851f341720862f4a4ffcbe4ae98df6057/packages/server/lib/middleware/access.middleware.ts#L520
Which includes `getMeta`

packages/webapp/src/constants.ts
kaposke: I'll retro-fit that PR later. I was adding that env variable assuming `prod` would be renamable. Since we changed that, that env variable becomes unnecessary, or a nice to have.

packages/webapp/src/pages/Environment/Settings/Show.tsx
kaposke: Oh yeah. The reason why I went with window.history doesn't exist anymore, so good call.

kaposke:
In sights of complications regarding deleting environments and soft-deletes, I'll be dividing this work into smaller PRs.

PR #3850, opened 2025-04-04
fix(cli): remove global store to allow tests to reuse code

bodinsamuel:
## Changes

- remove global store to allow tests to reuse code


TBonnin (approved):
thank you

PR #3858, opened 2025-04-07
feat(ui, api): auto-add missing scopes on flow enable

steliosmavro:
## 🐞 Issue

In the [Quickstart](https://docs.nango.dev/getting-started/quickstart#run-the-sample-app), the required scopes are only mentioned when setting up the Slack OAuth app, but **not** when creating the integration or enabling endpoints.

As a result, if you follow the Quickstart exactly as written, you’ll run into the following error when trying to create a new connection with that integration:

```
Invalid permissions requested  
No scopes requested
```

## 🛠️ Fix 

Instead of mentioning the scopes in the docs when setting up the integration or enabling endpoints, have the endpoints automatically add their corresponding scopes to the integration.

Since scopes are required when adding a connection, and the necessary scopes are already known based on the enabled endpoints, it would make sense to automatically include them when setting up the integration. This would reduce manual configuration and make the process more intuitive.

When enabling, any missing required scopes should be added to the integration.
When disabling, scopes should not be removed, since the user might have added them manually for other endpoints that require the same scopes.

### 🧪  Testing Instructions

1. **Start with an integration** that does **not** have any scopes, or remove them manually.
2. Go to the **Endpoints** page and **enable an endpoint** that requires a specific scope.
   - ✅ The required scope should automatically be added to the integration's scopes.
3. Now, **disable the same endpoint**.
   - ✅ The scope should **not** be removed from the integration.
4. Re-enable the same endpoint.
   - ✅ The scope should **not** be added again since it already exists once.

### DIFF
--- a/packages/server/lib/controllers/v1/integrations/providerConfigKey/patchIntegration.ts
+++ b/packages/server/lib/controllers/v1/integrations/providerConfigKey/patchIntegration.ts
@@ -13,7 +13,8 @@ const privateKey = z.string().startsWith('-----BEGIN RSA PRIVATE KEY----').endsW
 const validationBody = z
     .object({
         integrationId: providerConfigKeySchema.optional(),
-        webhookSecret: z.string().min(0).max(255).optional()
+        webhookSecret: z.string().min(0).max(255).optional(),
+        scopes: z.string().optional()
     })
     .strict()
     .or(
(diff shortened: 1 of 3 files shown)


khaliqgant:
Thanks for the contribution and sorry for the delay here. Unfortunately this doesn't fully solve the problem and is something we'll look to improve internally in combination with input from the product team.

Closing this PR.

PR #3860, opened 2025-04-07
fix(fleet): add more Render server definition

bodinsamuel:
## Changes

- add more Render server definition

### DIFF
--- a/packages/jobs/lib/runner/render.ts
+++ b/packages/jobs/lib/runner/render.ts
@@ -161,10 +161,10 @@ function getPlan(node: Node): 'starter' | 'standard' | 'pro' | 'pro_plus' | 'pro
     if (node.cpuMilli >= 4000 && node.memoryMb >= 8000) {
         return 'pro_plus';
     }
-    if (node.cpuMilli > 2000 || node.memoryMb > 4000) {
+    if (node.cpuMilli > 2000 || node.memoryMb >= 4000) {
         return 'pro';
     }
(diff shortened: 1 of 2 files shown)


packages/jobs/lib/runner/render.ts
```diff
@@ -147,8 +151,17 @@ async function withRateLimitHandling<T>(rateLimitGroup: 'create' | 'delete' | 'r
...
+    if (node.cpuMilli >= 4000 && node.memoryMb >= 16000) {
+        return 'pro_max';
+    }
+    if (node.cpuMilli >= 4000 && node.memoryMb >= 8000) {
+        return 'pro_plus';
+    }
+    if (node.cpuMilli > 2000 || node.memoryMb > 4000) {
         return 'pro';
     }
     if (node.cpuMilli > 1000 || node.memoryMb > 1024) {
```
TBonnin: we should probably be consistant and have `node.cpuMilli >= 1000 || node.memoryMb >= 2000` here and `node.cpuMilli >= 2000 || node.memoryMb >= 4000` above

TBonnin (approved):

packages/jobs/lib/runner/render.ts
bodinsamuel: imo we should be exact but yeah I can change

TBonnin: the thing is next provider like for example AWS is not gonna have the exact same instance size but we are gonna want to match roughly

PR #3862, opened 2025-04-07
feat(jwt): use dynamic credentials to build the jwt

hassan254-prog:

- Add support for Snowflake (JWT)
- Add  a `data` field for `proxy.verification` to enable us test `POST` methods that require body to be present.
- Improve `JWT` auth mode to use dynamic credentials when generating a `JWT`


packages/frontend/lib/types.ts
```diff
@@ -75,14 +75,11 @@ export interface TableauCredentials {
...
-        | {
-              id: string;
-              secret: string;
-          }
-        | string; // Colon-separated string for Ghost Admin: 'id:secret'
+    type?: 'JWT';
+    privateKey?: {
+        id: string;
+        secret: string;
+    };
```
bodinsamuel: is it missing `[key: string]: string`?

packages/server/lib/controllers/auth/postJwt.ts
```diff
@@ -31,20 +31,7 @@ import type { LogContext } from '@nangohq/logs';
...
-            z.object({
-                id: z.string(),
-                secret: z.string()
-            }),
-            z.string()
-        ])
-    })
-    .strict();
-
+const bodyValidation = z.object({}).catchall(z.any()).strict();
```
bodinsamuel, line 34: I feel like privateKey is still a standard prop that could be validated

packages/types/lib/auth/http.api.ts
```diff
@@ -109,16 +109,7 @@ export type PostPublicTableauAuthorization = Endpoint<{
...
-        privateKeyId?: string;
-        issuerId?: string;
-        privateKey:
-            | {
-                  id: string;
-                  secret: string;
-              }
-            | string;
-    };
+    Body: Record<string, any>;
```
bodinsamuel, line 112: Same here, there is some inconsistencies with the type. Could at least be like this no?
```ts
{ 
[key: string]: string;
privateKey?: ...
}
```

packages/types/lib/auth/api.ts
```diff
@@ -160,14 +160,7 @@ export interface TableauCredentials extends CredentialsCommon {
...
     type: AuthModes['Jwt'];
-    privateKeyId?: string;
-    issuerId?: string;
-    privateKey:
-        | {
-              id: string;
-              secret: string;
-          }
-        | string; // Colon-separated string for Ghost Admin: 'id:secret'
+    [key: string]: any;
```
bodinsamuel, line 163: same

packages/connect-ui/src/views/Go.tsx
```diff
@@ -44,15 +44,7 @@ const formSchema: Record<AuthModeType, z.AnyZodObject> = {
...
-        privateKeyId: z.string().optional(),
-        issuerId: z.string().optional(),
-        privateKey: z.union([
-            z.object({
-                id: z.string(),
-                secret: z.string()
-            }),
-            z.string()
-        ])
+        // JWT is custom every time
```
bodinsamuel, line 47: same

packages/connect-ui/src/views/Go.tsx
```diff
@@ -236,9 +228,18 @@ export const Go: React.FC = () => {
                         detectClosedAuthWindow
                     });
                 } else {
+                    let credentials = values['credentials'];
+                    // Handle Ghost Admin API key transformation
+                    if (integration?.provider.includes('ghost-admin') && credentials?.privateKey) {
```
bodinsamuel: Not super fan of having this logic inside the frontend. I would prefer this to be in the backend if possible

packages/shared/lib/auth/jwt.ts
```diff
@@ -7,66 +7,70 @@ import { AuthCredentialsError } from '../utils/error.js';
...
+        payload['exp'] = now + provider.token.expires_in_ms / 1000;
 
-    try {
-        const token = signJWT({
-            payload,
-            secretOrPrivateKey: Buffer.from(privateKey, 'hex'),
-            options: { algorithm: provider.token.headers.alg, header }
-        });
-        const expiresAt = new Date(Date.now() + provider.token.expires_in_ms);
+        const header = Object.entries(provider.token.headers).reduce<Record<string, any>>((acc, [key, value]) => {
```
bodinsamuel: should be plural no?

packages/server/lib/utils/utils.ts
```diff
@@ -189,12 +190,20 @@ export function parseCredentialsParamsFromTemplate(provider: ProviderTwoStep): s
         return matches;
     };
 
-    if (provider.token_params || provider.token_headers) {
-        const tokenParamsMatches = provider.token_params ? extractCredentialParams(provider.token_params) : [];
-        return [...new Set(tokenParamsMatches.map(cleanParamName))]; // Remove duplicates
+    const matches: string[] = [];
+
+    if ('token_params' in provider || 'token_headers' in provider) {
```
bodinsamuel: feels like it could be two ifs

bodinsamuel (commented):
Not tested yet, just a batch of comments to be sure I understand well

packages/server/lib/utils/utils.ts
hassan254-prog: updated with the latest commit 👍 .

packages/shared/lib/auth/jwt.ts
hassan254-prog: Thanks 🤦, it's good to keep it singular, just like `JWT` protocols, updated in latest commit as well.

packages/connect-ui/src/views/Go.tsx
hassan254-prog: I have moved it to the backend, that way when we remove `CreateLegacy.tsx`, we can have that `apikey` split logic in one location.

packages/frontend/lib/types.ts
hassan254-prog: I have updated it to `[key: string]: any` for consistency. At least that way, it is fully dynamically built, the same will apply to the rest.

packages/shared/lib/auth/jwt.ts
```diff
@@ -7,66 +7,78 @@ import { AuthCredentialsError } from '../utils/error.js';
...
+
+            if (typeof strippedValue === 'object' && strippedValue !== null) {
+                acc[key] = interpolateObject(strippedValue, dynamicCredentials);
+            } else if (typeof strippedValue === 'string') {
+                acc[key] = interpolateString(strippedValue, dynamicCredentials);
+            } else {
+                acc[key] = strippedValue;
+            }
+            return acc;
+        }, {});
```
TBonnin: if don't think you need to use reduce. what about:
```
  const payload = {};
  for (const [key, value] of Object.entries(provider.token.payload)) {
    const strippedValue = stripCredential(value);
    if (strippedValue === null) {
      payload[key] = null;
    } else if (typeof strippedValue === 'object') {
      payload[key] = interpolateObject(strippedValue, dynamicCredentials);
    } else if (typeof strippedValue === 'string') {
      payload[key] = interpolateString(strippedValue, dynamicCredentials);
    } else {
      payload[key] = strippedValue;
    }
  }
```

packages/shared/lib/auth/jwt.ts
```diff
@@ -7,66 +7,78 @@ import { AuthCredentialsError } from '../utils/error.js';
...
-        ...provider.token.payload,
-        iat: now,
-        exp: now + provider.token.expires_in_ms / 1000
-    };
-    const header = {
-        ...provider.token.headers,
-        alg: provider.token.headers.alg,
-        kid: privateKeyId
-    };
+        const header = Object.entries(provider.token.header).reduce<Record<string, any>>((acc, [key, value]) => {
```
TBonnin: same comment about reduce

packages/shared/lib/auth/jwt.ts
```diff
@@ -7,66 +7,78 @@ import { AuthCredentialsError } from '../utils/error.js';
...
+                ? signJWT({
+                      payload,
+                      secretOrPrivateKey: formatPrivateKey(interpolatedSigningKey),
+                      options: { algorithm: provider.token.header.alg, header }
+                  })
+                : signJWT({
+                      payload,
+                      secretOrPrivateKey: Buffer.from(interpolatedSigningKey, 'hex'),
+                      options: { algorithm: provider.token.header.alg, header }
+                  });
```
TBonnin: what about:
```
const pKey = provider.signature.protocol === 'RSA' ? formatPrivateKey(interpolatedSigningKey) : Buffer.from(interpolatedSigningKey, 'hex')
const token = signJWT({
   payload,
   pKey,
   options: { algorithm: provider.token.header.alg, header }
})
```

packages/webapp/src/pages/Connection/Authorization.tsx
```diff
@@ -198,22 +198,33 @@ export const Authorization: React.FC<AuthorizationProps> = ({ connection, errorL
...
+                                    <SecretInput disabled defaultValue={`${privateKey.id}:${privateKey.secret}`} copy={true} />
                                 </div>
                             );
-                        })}
+                        }
+                        return Object.keys(connection.credentials)
+                            .filter((key) => !['type', 'token', 'expires_at', 'raw'].includes(key))
+                            .map((key) => {
+                                const value = (connection.credentials as Record<string, string>)[key];
+                                return (
```
TBonnin: would it be possible to have a single chain to map the key and return div definition. Something like:
```ts
return Object.keys(connection.credentials).flatMap((key) => {
   let value: string
   let label: string
   // Handle privateKey if it's an object (id and secret), backwards compatibility with ghost-admin
    const privateKey = (connection.credentials as Record<string, any>)['privateKey'];
     if (key === 'privateKey' && ...) {
         value = `${privateKey.id}:${privateKey.secret}`;
         label = `PRIVATE_KEY`
     } else if(!['type', 'token', 'expires_at', 'raw'].includes(key)) {
         value = (connection.credentials as Record<string, string>)[key]
         label = key.replace(/_/g, ' ').replace(/\b\w/g, (char) => char.toUpperCase())
     } else {
         return []
     }
     return [(
<div className="flex flex-col" key={key}>
<span className="text-gray-400 text-xs uppercase mb-1">${label}</span>
<SecretInput disabled defaultValue={value} copy={true} />
</div>)]
})
                            .
```

hassan254-prog: updated with the [latest](https://github.com/NangoHQ/nango/pull/3862/commits/5a8a47a349d17e8c3cd7673491408e7deddb7934) 🙏 .

packages/shared/lib/auth/jwt.ts
hassan254-prog: updated with the [latest](https://github.com/NangoHQ/nango/pull/3862/commits/5a8a47a349d17e8c3cd7673491408e7deddb7934) 🙏 .

bodinsamuel (approved):
Looks okay, I don't have the mean to test 👌🏻

PR #3863, opened 2025-04-07
feat: nango locks in scripts

TBonnin:
Adding functions to the nango runner-sdk to manage locks.
```
await nango.tryAcquireLock({ key: "abc", ttlMs: 60_000 })
```
It makes it possible for customers to handle custom locking logic in their script, like skipping a webhook execution if one is already running.
I decided to make the locking a runner-sdk implementation details and be ok with a bit more duplication of code rather than having the implementation as part of the runner-sdk abstract class (and then have the state not fully self-contained). Locks state is imho at the runner-level and not script-level.

Caveat: for now the locks are stored in memory (by the runner). During deployment of runners, there might be 2 runners up at the same time and potentially the same lock could be acquired twice. In a second steps we will store the locks in a shared datastore to ensure 100% unique locks

# How to test
You can have 2 syncs that both tries to acquire the same lock and wait to ensure overlap
```
    const lock = await nango.tryAcquireLock({
        key: 'MY_KEY',
        ttlMs: 60_000,
    });
    if (!lock) {
        await nango.log('could not acquire lock. skipping...');
        return
    }
    await nango.log('acquired lock');

    const waitSecs = 10;
    for (let i = 0; i < waitSecs; i += 1) {
        await new Promise((resolve) => setTimeout(resolve, 1000));
        await nango.log(`waiting for ${waitSecs - i} seconds`);
    }
```

### DIFF
--- a/packages/cli/lib/services/sdk.ts
+++ b/packages/cli/lib/services/sdk.ts
@@ -96,6 +96,23 @@ export class NangoActionCLI extends NangoActionBase {
         this.log(`This has no effect but on a remote Nango instance would start a schedule`);
         return Promise.resolve();
     }
+
+    // eslint-disable-next-line @typescript-eslint/require-await
+    public override async tryAcquireLock(_props: { key: string; ttlMs: number }): Promise<boolean> {
+        // Not applicable to CLI
+        return true;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/require-await
+    public override async releaseLock(_props: { key: string }): Promise<boolean> {
+        // Not applicable to CLI
+        return true;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/require-await
+    public override async releaseAllLocks(): Promise<void> {
+        // Not applicable to CLI
+    }
 }
 
 export class NangoSyncCLI extends NangoSyncBase {
@@ -287,6 +304,23 @@ export class NangoSyncCLI extends NangoSyncBase {
     public override async setMergingStrategy(_merging: { strategy: 'ignore_if_modified_after' | 'override' }, _model: string) {
         // Not applicable to CLI
         return Promise.resolve();
+    }
+
+    // eslint-disable-next-line @typescript-eslint/require-await
+    public override async tryAcquireLock(_props: { key: string; ttlMs: number }): Promise<boolean> {
+        // Not applicable to CLI
+        return true;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/require-await
+    public override async releaseLock(_props: { key: string }): Promise<boolean> {
+        // Not applicable to CLI
+        return true;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/require-await
+    public override async releaseAllLocks(): Promise<void> {
+        // Not applicable to CLI
     }
 }
 
--- a/packages/cli/lib/services/sdk.ts
+++ b/packages/cli/lib/services/sdk.ts
@@ -146,6 +146,9 @@ export class NangoSyncCLI extends NangoSyncBase {
     log = NangoActionCLI['prototype']['log'];
     triggerSync = NangoActionCLI['prototype']['triggerSync'];
     startSync = NangoActionCLI['prototype']['startSync'];
+    tryAcquireLock = NangoActionCLI['prototype']['tryAcquireLock'];
+    releaseLock = NangoActionCLI['prototype']['releaseLock'];
+    releaseAllLocks = NangoActionCLI['prototype']['releaseAllLocks'];
 
     public batchSave<T extends object>(results: T[], model: string) {
         if (!results || results.length === 0) {
@@ -304,23 +307,6 @@ export class NangoSyncCLI extends NangoSyncBase {
     public override async setMergingStrategy(_merging: { strategy: 'ignore_if_modified_after' | 'override' }, _model: string) {
         // Not applicable to CLI
         return Promise.resolve();
-    }
-
-    // eslint-disable-next-line @typescript-eslint/require-await
-    public override async tryAcquireLock(_props: { key: string; ttlMs: number }): Promise<boolean> {
-        // Not applicable to CLI
-        return true;
-    }
-
-    // eslint-disable-next-line @typescript-eslint/require-await
-    public override async releaseLock(_props: { key: string }): Promise<boolean> {
-        // Not applicable to CLI
-        return true;
-    }
-
-    // eslint-disable-next-line @typescript-eslint/require-await
-    public override async releaseAllLocks(): Promise<void> {
-        // Not applicable to CLI
     }
 }
 
--- a/packages/runner-sdk/lib/action.ts
+++ b/packages/runner-sdk/lib/action.ts
@@ -48,7 +48,7 @@ export type ProxyConfiguration = Omit<UserProvidedProxyConfiguration, 'files' | 
 export abstract class NangoActionBase {
     abstract nango: Nango;
     private attributes = {};
-    activityLogId?: string | undefined;
+    activityLogId: string;
     syncId?: string;
     nangoConnectionId?: number;
     environmentId: number;
@@ -72,10 +72,7 @@ export abstract class NangoActionBase {
         this.environmentId = config.environmentId;
         this.providerConfigKey = config.providerConfigKey;
         this.runnerFlags = config.runnerFlags;
-
-        if (config.activityLogId) {
-            this.activityLogId = config.activityLogId;
-        }
+        this.activityLogId = config.activityLogId;
 
         if (config.syncId) {
             this.syncId = config.syncId;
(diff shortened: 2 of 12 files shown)


packages/runner-sdk/lib/action.ts
```diff
@@ -72,10 +72,7 @@ export abstract class NangoActionBase {
         this.environmentId = config.environmentId;
         this.providerConfigKey = config.providerConfigKey;
         this.runnerFlags = config.runnerFlags;
-
-        if (config.activityLogId) {
-            this.activityLogId = config.activityLogId;
-        }
+        this.activityLogId = config.activityLogId;
```
TBonnin, line 75: config activityLogId is of type `string` already. I think it is a leftover from a long time ago

packages/runner/lib/sdk/locks.ts
```diff
@@ -0,0 +1,85 @@
...
+
+interface Lock {
+    key: string;
+    owner: string;
+    expiresAt: Date;
+}
+
+export class Locks {
+    private store = new Map<string, Lock>();
+    private mutex = new Map<string, boolean>();
```
TBonnin, line 13: locks for the locks 😬

packages/runner/lib/sdk/locks.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    public async tryAcquireLock({ owner, key, ttlMs }: { owner: string; key: string; ttlMs: number }): Promise<Result<boolean>> {
```
TBonnin, line 15: those functions doesn't require async at the moment but in v2 the store will be a db so we will need it then.

packages/runner-sdk/models.d.ts
```diff
@@ -358,6 +358,9 @@ export declare class NangoAction {
      * Only use that method when you want to access resources that are unrelated to the current connection/provider.
      */
     uncontrolledFetch(options: { url: URL; method?: HTTP_METHOD; headers?: Record<string, string> | undefined; body?: string | null }): Promise<Response>;
+    tryAcquireLock(props: { key: string; ttlMs: number }): Promise<boolean>;
+    releaseLock(props: { key: string }): Promise<boolean>;
+    releaseAllLocks(): Promise<void>;
```
TBonnin: we don't need to expose `releaseAllLock` but why not. wdyt?

packages/cli/lib/services/sdk.ts
```diff
@@ -96,6 +96,23 @@ export class NangoActionCLI extends NangoActionBase {
         this.log(`This has no effect but on a remote Nango instance would start a schedule`);
         return Promise.resolve();
     }
+
+    // eslint-disable-next-line @typescript-eslint/require-await
+    public override async tryAcquireLock(_props: { key: string; ttlMs: number }): Promise<boolean> {
+        // Not applicable to CLI
+        return true;
```
bodinsamuel, line 103: I think it's the right choice, it's not possible to have two scripts running with the CLI

packages/cli/lib/services/sdk.ts
```diff
@@ -288,6 +305,23 @@ export class NangoSyncCLI extends NangoSyncBase {
...
+    // eslint-disable-next-line @typescript-eslint/require-await
+    public override async releaseLock(_props: { key: string }): Promise<boolean> {
+        // Not applicable to CLI
+        return true;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/require-await
+    public override async releaseAllLocks(): Promise<void> {
+        // Not applicable to CLI
+    }
```
bodinsamuel: if you don't want code repetition you can use the fake extend, e.g:
`startSync = NangoActionCLI['prototype']['startSync'];`

packages/runner-sdk/lib/action.ts
```diff
@@ -385,4 +382,19 @@ export abstract class NangoActionBase {
 
         return await fetch(options.url, props);
     }
+
+    /**
+     * Try to acquire a lock for a given key.
+     * The lock is acquired if the key does not exist or if it exists but is expired.
+     * The lock is valid for the entire execution of the script and will be released automatically when the script ends.
```
bodinsamuel: I would add `or when you call "releaseLock"`

packages/runner/lib/sdk/sdk.ts
```diff
@@ -460,6 +478,54 @@ export class NangoSyncRunner extends NangoSyncBase {
...
+    public override async releaseLock(props: { key: string }): Promise<boolean> {
+        return this.locking.releaseLock(props);
+    }
+
+    public override async releaseAllLocks(): Promise<void> {
+        return this.locking.releaseAllLocks();
+    }
+}
+
+class Locking {
```
bodinsamuel, line 495: nit I feel like this class is not bringing much

packages/runner/lib/sdk/locks.ts
```diff
@@ -0,0 +1,85 @@
...
+            return Ok(false);
+        } finally {
+            this.mutex.delete(key);
+        }
+    }
+
+    public async releaseAllLocks({ owner }: { owner: string }): Promise<Result<void>> {
+        for (const [key, lock] of this.store.entries()) {
+            if (lock.owner === owner) {
+                this.releaseLock({ owner, key });
```
bodinsamuel:
```suggestion
                void this.releaseLock({ owner, key });
```

packages/runner/lib/sdk/locks.ts
bodinsamuel: I'm not sure I get why it's necessary? I don't think it's possible to have a race condition with the current code no?

packages/runner/lib/exec.ts
```diff
@@ -277,6 +285,7 @@ export async function exec(
                 };
             }
         } finally {
+            await nango.releaseAllLocks().catch((err: unknown) => logger.warning('Failed to release all locks', { reason: err }));
```
bodinsamuel: await and catch is confusing imo, can you use a try/catch?

packages/runner-sdk/models.d.ts
bodinsamuel: it might lead people to incorrectly use it. imo I initially thought it would release all locks vs it releases all locks of this scripts

packages/runner/lib/sdk/locks.ts
```diff
@@ -0,0 +1,85 @@
...
+    private mutex = new Map<string, boolean>();
+
+    public async tryAcquireLock({ owner, key, ttlMs }: { owner: string; key: string; ttlMs: number }): Promise<Result<boolean>> {
+        if (!owner || owner.length === 0 || owner.length > 255) {
+            return Err('Invalid lock owner (must be between 1 and 255 characters)');
+        }
+        if (!key || key.length === 0 || key.length > 255) {
+            return Err('Invalid lock key (must be between 1 and 255 characters)');
+        }
+        if (ttlMs <= 0) {
```
bodinsamuel, line 22: I'm wondering if it's a foot gun to require a TTL? 
Customers will need to add something but they rarely know how long their syncs are running. If we consider that 99.9% of locks will go through `releaseAllLock` then I would not require a TTL or at least set a default to 24H

bodinsamuel (approved):
Tested locally, worked well 💪🏻 
Just a few comments

packages/runner/lib/sdk/sdk.ts
TBonnin: It is helping a bit with duplication of code but the main reason was to encapsulate `owner`. Instead of using `activityLogId` multiple times in the body of the functions, it forcing setting the owner once when creating the instance of the class

packages/runner/lib/exec.ts
TBonnin: sure. I thought for logging only it was simple enough

packages/runner/lib/sdk/locks.ts
TBonnin: I agree with you. I also think it is good that the user think twice about this feature and since it is a feature I thought it won't be a bad idea to require the ttl. Also I am not sure what would be a good default: 24h for syncs is fine but not for webhooks or actions. Let's see if I can set a default value based on the type of script

bodinsamuel: mmh true good point. Default based on the script could work

packages/runner/lib/sdk/locks.ts
TBonnin: the race condition is just below, between the `get` and `set` (if I was using pg, this read and write would be wrapped in a SELECT FOR UPDATE  transaction)
```
const existing = this.store.get(key);
if (...) {
...
}
this.store.set(key, { key, owner, expiresAt: new Date(now.getTime() + ttlMs) });
```

packages/runner-sdk/models.d.ts
khaliqgant: People will use and abuse anything we expose. I wouldn’t expose it personally but if we did would give it a clearer name

packages/runner/lib/sdk/locks.ts
TBonnin: even based on the type of script is hard to come up with a good default. It depends very much on what the script is doing. I am gonna stick with it and keep the ttl required. The ttl value is a tradeoff the customer has to make

packages/runner/lib/sdk/locks.ts
bodinsamuel: I don't think it can happen right now, the code is synchronous, so the whole function is executed at once but maybe later with async indeed

PR #3864, opened 2025-04-07
feat: rename environments

kaposke:
### Description
This PR contains cherry-picked files (with tweaks) from #3849, excluding anything related to deleting environments.

- Added new `Main` section to environment settings with `EditableInput` for environment name.
image

- Prod environment shouldn't be renamable with informative tooltip 
image

- Tweaked `Create Environment` button be blocked and show different tooltip based on plan, when maximum amount of environments is reached
image

- Removed message from create environment dialog that said only the prod environment is billed.

### Ticket
[NAN-2685](https://linear.app/nango/issue/NAN-2685/editdelete-environments-in-the-ui)

### Testing instructions
Checkout to branch and test mentioned changes in UI.

### DIFF
--- a/packages/server/lib/controllers/v1/meta/getMeta.ts
+++ b/packages/server/lib/controllers/v1/meta/getMeta.ts
@@ -1,6 +1,9 @@
-import { asyncWrapper } from '../../../utils/asyncWrapper.js';
-import { baseUrl, NANGO_VERSION, requireEmptyQuery, zodErrorToHTTP } from '@nangohq/utils';
 import { environmentService, getOnboarding } from '@nangohq/shared';
+import { NANGO_VERSION, baseUrl, requireEmptyQuery, zodErrorToHTTP } from '@nangohq/utils';
+
+import { planToApi } from '../../../formatters/plan.js';
+import { asyncWrapper } from '../../../utils/asyncWrapper.js';
+
 import type { GetMeta } from '@nangohq/types';
 
 export const getMeta = asyncWrapper<GetMeta>(async (req, res) => {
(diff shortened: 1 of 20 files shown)


packages/server/lib/middleware/access.middleware.ts
```diff
@@ -503,7 +503,25 @@ async function fillLocalsFromSession(req: Request, res: Response<any, RequestLoc
...
+            }
+            plan = planRes.value;
+        }
+
         res.locals['user'] = user;
+        res.locals['account'] = account;
+        res.locals['plan'] = plan;
 
         const fullPath = path.join(req.baseUrl, req.route.path);
         if (ignoreEnvPaths.includes(fullPath)) {
```
kaposke, line 527: Moved fetching of account and plan to before this early exit. Account and plan are independent of any environment, and not having them in specific endpoints seems inconsistent. In my case, I needed it for `postEnvironment`, `patchEnvironment` and `getMeta`.

Also, `user` is not present in tests because we generally use `env.token`. We should probably make the behavior consistent across tests and real auth.

packages/server/lib/controllers/v1/meta/getMeta.ts
```diff
@@ -10,12 +13,13 @@ export const getMeta = asyncWrapper<GetMeta>(async (req, res) => {
...
     }
 
-    const sessionUser = res.locals['user'];
+    const { user: sessionUser, plan } = res.locals;
 
     const environments = await environmentService.getEnvironmentsByAccountId(sessionUser.account_id);
     const onboarding = await getOnboarding(sessionUser.id);
     res.status(200).send({
         data: {
+            plan: plan ? planToApi(plan) : null,
```
bodinsamuel: Still pushing back on this

bodinsamuel (approved):
Same feedback, otherwise LGTM 💪🏻

TBonnin (approved):
same concern as @bodinsamuel about using the /meta endpoint. Probably also because I don't like this endpoint all together. Can we use the `/current` environment endpoint instead to get the plan? Not a blocker at this stage, we can merge as is and follow up with a cleanup PR

PR #3867, opened 2025-04-08
fix(frontend): better modal handling

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-3060/investigate-safari-blocking-auth-popup

- Better modal handling
To bypass the popup blocker Safari requires the modal to be open in the same function, not in a async mode. So I refactored the code around that

- Move AuthorizationModal into its own file
Also removed the status which was mostly useless

## 🧪 Tests
 
I would love some tests since it's browser/behavior specific.

- Go to dashboard
- Use connect ui or legacy ui
- Connect to an oauth integration


packages/frontend/lib/index.ts
```diff
@@ -88,7 +89,6 @@ export default class Nango {
         }
 
         this.hostBaseUrl = config.host.replace(/\/+$/, ''); // Remove trailing slash.
-        this.status = AuthorizationStatus.IDLE;
```
TBonnin, line 91: I am pretty sure `status` was here for a good reason. Why is it mostly useless now?

bodinsamuel: I'm not sure why it was useful tbh, it's some kind of locking mechanism, I feel like there is no benefit in erroring vs just reopening a popup.
Maybe @khaliqgant knows best?

khaliqgant: I don't recall any specific implementation detail as this state tracking goes back to the Pizzly days: https://github.com/triggerdotdev/Pizzly/blob/master/packages/frontend/lib/index.ts#L80

If you're now tracking this in a different way then would go forth in removing. Otherwise I would be reluctant to remove it unless we clearly see the cases where it was used previously are no longer valid / needed.

bodinsamuel: I really don't think it's useful now. If I trust the code it was to prevent further message from the websocket after it's done, but with this PR I'm actually closing the connection at the end so I don't see any problem removing it

khaliqgant: Cool. Works for me

TBonnin: I have the feeling it was useful, something to do with the the window closing detection feature but I cannot remember any details unfortunately, I haven't touched this code for more than a year.
Your call though. If I were you I would keep it but I trust your judgment anyway.

TBonnin (approved):

packages/frontend/lib/index.ts
bodinsamuel: I see, it's a problem if we have feeling but we don't know, I can't even properly put it back because I don't know what to fix :D 
I'll see in prod 🤞🏻

PR #3868, opened 2025-04-08
fix(rateLimit): use dynamic size

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-3019/store-ratelimit-configuration-in-plan

- Use dynamic size from the plan
It's still a bit arbitrary but at least it has some flexibility. Ideally, we reduce the current default value to something like `100` so we can protect ourselves from free accounts and still have good margins for paying customers.


packages/server/lib/middleware/ratelimit.middleware.ts
```diff
@@ -4,67 +4,95 @@ import { RateLimiterMemory, RateLimiterRedis, RateLimiterRes } from 'rate-limite
...
 
-const rateLimiter = await (async () => {
+const defaultLimit = parseInt(process.env['DEFAULT_RATE_LIMIT_PER_MIN'] || '0') || 3500;
+const rateLimiterSize: Record<DBPlan['api_rate_limit_size'], number> = {
+    s: defaultLimit / 2,
+    m: defaultLimit,
+    l: defaultLimit * 5,
+    xl: defaultLimit * 10,
+    '2xl': defaultLimit * 50,
+    '3xl': defaultLimit * 100
```
TBonnin, line 23: is the plan to quickly lower the default limit? Otherwise those factors are huge.

packages/server/lib/middleware/ratelimit.middleware.ts
```diff
@@ -4,67 +4,95 @@ import { RateLimiterMemory, RateLimiterRedis, RateLimiterRes } from 'rate-limite
...
+};
+const limiter = new Map<DBPlan['api_rate_limit_size'], RateLimiterAbstract>();
```
TBonnin: I would use plural `limiters` since it is a map of several limiters

packages/server/lib/middleware/ratelimit.middleware.ts
```diff
@@ -79,14 +107,14 @@ function getKey(req: Request, res: Response<any, RequestLocals>): string {
...
     const fullPath = path.join(req.baseUrl, req.route.path);
-    if (specialPaths.includes(fullPath)) {
+
+    if (specialPaths.some((p) => fullPath.startsWith(p))) {
         // limiting to 6 requests per period to avoid brute force attacks
-        return Math.floor(rateLimiter.points / 6);
-    } else if (!res.locals.account || (res.locals.plan && res.locals.plan.name === 'free')) {
-        // TODO: maybe store this in plans
-        // Free account get way less requests
+        return Math.floor(maxPoints / 10);
```
TBonnin: was `6` now `10`. What's the reason?

TBonnin (approved):
like it

packages/server/lib/middleware/ratelimit.middleware.ts
bodinsamuel: yes, I think I'll go for 100 for default, should be more than enough for free user

packages/server/lib/middleware/ratelimit.middleware.ts
bodinsamuel: indeed

packages/server/lib/middleware/ratelimit.middleware.ts
bodinsamuel: ah leftover I was just testing

PR #3871, opened 2025-04-09
fix(integration-deploy): fix install of a public integration to install the latest version

hassan254-prog:
- Fix install of a public integration(sync or action) to install the latest version from the available public flows.


packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -442,7 +443,14 @@ export async function deployPreBuilt({
...
 
-        const version = bumpedVersion || '0.0.1';
+        // Get the latest version from the flows
+        const availablePublicFlows = flowService.getAllAvailableFlowsAsStandardConfig();
+        const [template] = availablePublicFlows.filter((flow: { providerConfigKey: string }) => flow.providerConfigKey === config.provider);
+        const templateFlows: NangoSyncConfig[] = [...(template?.actions || []), ...(template?.syncs || [])];
+        const flow = templateFlows.find((f) => f.name === sync_name);
+        const latestVersion = flow?.version;
+
+        const version = latestVersion || bumpedVersion || '1.0.0';
```
bodinsamuel: not 0.0.1?

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -442,7 +443,14 @@ export async function deployPreBuilt({
             }
         }
 
-        const version = bumpedVersion || '0.0.1';
+        // Get the latest version from the flows
+        const availablePublicFlows = flowService.getAllAvailableFlowsAsStandardConfig();
+        const [template] = availablePublicFlows.filter((flow: { providerConfigKey: string }) => flow.providerConfigKey === config.provider);
+        const templateFlows: NangoSyncConfig[] = [...(template?.actions || []), ...(template?.syncs || [])];
+        const flow = templateFlows.find((f) => f.name === sync_name);
```
bodinsamuel: The flow is already fetched in postDeploy.ts can you just pass it as params OR add version to `PreBuiltFlowConfig`

bodinsamuel (commented):
Thanks, minor comment that should simplify the PR

packages/shared/lib/services/sync/config/deploy.service.ts
hassan254-prog: makes it simpler thanks 🙏 , updated with latest [commit](https://github.com/NangoHQ/nango/pull/3871/commits/f531c263ec63b1073c05c2d024f968514b768f80)

TBonnin (approved):

bodinsamuel (approved):

PR #3874, opened 2025-04-09
fix: runner-sdk triggerSync signature

TBonnin:
triggerSync signature was not up-to-date in `packages/runner-sdk/models.d.ts`:
- missing variant
- missing syncMode recent refactor

### DIFF
--- a/packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
+++ b/packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
@@ -43,7 +43,7 @@ exports[`buildModelTs > should return empty (with sdk) 1`] = `
 
 import type { Nango } from '@nangohq/node';
 import type { AxiosInstance, AxiosInterceptorManager, AxiosRequestConfig, AxiosResponse, AxiosError } from 'axios';
-import type { ApiEndUser, DBSyncConfig, DBTeam, GetPublicIntegration, HTTP_METHOD, RunnerFlags } from '@nangohq/types';
+import type { ApiEndUser, DBSyncConfig, DBTeam, GetPublicIntegration, HTTP_METHOD, RunnerFlags, PostPublicTrigger } from '@nangohq/types';
 import type { ZodSchema, SafeParseSuccess } from 'zod';
 
 export declare const oldLevelToNewLevel: {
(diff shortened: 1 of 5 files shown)


packages/runner-sdk/lib/action.ts
```diff
@@ -355,7 +356,7 @@ export abstract class NangoActionBase {
         providerConfigKey: string,
         connectionId: string,
         sync: string | { name: string; variant: string },
```
TBonnin, line 358: is there any reason why the runner-sdk `triggerSync` function doesn't have the same signature as the node-client one? here we can only trigger a single sync while, with the node-client, we can pass an array of syncs.
@bodinsamuel @khaliqgant @bastienbeurier 

This PR is not addressing this problem but I am curious

bodinsamuel: Not that I am aware of, I think I was barely here when it got introduced

packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
```diff
@@ -415,7 +415,12 @@ export declare class NangoAction {
     paginate<T = any>(config: ProxyConfiguration): AsyncGenerator<T[], undefined, void>;
     triggerAction<In = unknown, Out = object>(providerConfigKey: string, connectionId: string, actionName: string, input?: In): Promise<Out>;
     zodValidateInput<T = any, Z = any>({ zodSchema, input }: { zodSchema: ZodSchema<Z>; input: T }): Promise<SafeParseSuccess<Z>>;
-    triggerSync(providerConfigKey: string, connectionId: string, syncName: string, fullResync?: boolean): Promise<void | string>;
+    triggerSync(
+        providerConfigKey: string,
+        connectionId: string,
+        sync: string | { name: string; variant: string },
+        syncMode?: PostPublicTrigger['Body']['sync_mode'] | boolean
```
bodinsamuel, line 422: You sure its not going to break some scripts?

bodinsamuel (approved):

packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
TBonnin: it is backward compatible since it allows the old type `string` for sync and `boolean` for syncMode

PR #3876, opened 2025-04-10
fix(rateLimit): increase limit for providers.json

bodinsamuel:
## Changs

- increase limit for providers.json
Since all runners hit this endpoint with just 2-4 ips, it spammed the API. It should be better like this, but ultimately it's not enough

- Add more granular increment for rate limit size
The jump was a bit too big

### DIFF
--- a/packages/server/lib/middleware/ratelimit.middleware.ts
+++ b/packages/server/lib/middleware/ratelimit.middleware.ts
@@ -19,8 +19,8 @@ const rateLimiterSize: Record<DBPlan['api_rate_limit_size'], number> = {
     m: defaultLimit,
     l: defaultLimit * 5,
     xl: defaultLimit * 10,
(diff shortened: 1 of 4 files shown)


packages/server/lib/middleware/ratelimit.middleware.ts
```diff
@@ -117,6 +118,8 @@ function getPointsToConsume(req: Request, res: Response<any, RequestLocals>, max
     if (specialPaths.some((p) => fullPath.startsWith(p))) {
         // limiting to 6 requests per period to avoid brute force attacks
         return Math.floor(maxPoints / 6);
+    } else if (fullPath === '/providers.json') {
+        return 1;
```
TBonnin, line 123: maybe a comment to explain why /providers.json get a higher limit

TBonnin (approved):

PR #3877, opened 2025-04-10
feat(logs): search in global field, index batchSave

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2698/full-text-search-includes-payloads
Fixes https://linear.app/nango/issue/NAN-2699/search-logs-by-record-id

- Search in global field
It's still not searching everything but a selected number of fields, however, it's already better since we now have error messages, status codes and such. And we can add more.

- Index `batchSave`
Index this payload outside meta so we can store records id in the global field. This will probably log a lot of data in prod.

## 🧪 Tests

- Do that

https://github.com/user-attachments/assets/d6526b6e-4666-46d7-a782-4a1d2d7a06b9

### DIFF
--- a/packages/persist/lib/records.ts
+++ b/packages/persist/lib/records.ts
@@ -127,7 +127,7 @@ export async function persistRecords({
             `Successfully batched ${allModifiedKeys.size} record${allModifiedKeys.size > 1 ? 's' : ''} for model ${baseModel}`,
             { persistType },
             {
-                batchSave: {
+                persistResults: {
                     model: baseModel,
                     added: summary.addedKeys.length,
                     updated: summary.updatedKeys.length,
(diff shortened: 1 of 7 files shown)


packages/persist/lib/records.ts
```diff
@@ -123,19 +123,22 @@ export async function persistRecords({
...
-                //addedKeys: summary.addedKeys,
-                updated: summary.updatedKeys.length,
-                //updatedKeys: summary.updatedKeys,
-                deleted: summary.deletedKeys?.length || 0
-                //deleteKeys: summary.deletedKeys || []
+        void logCtx.info(
+            `Successfully batched ${allModifiedKeys.size} record${allModifiedKeys.size > 1 ? 's' : ''} for model ${baseModel}`,
+            { persistType },
+            {
+                batchSave: {
```
TBonnin: name might be misleading here since it can be the result of batchSave or batchUpdate or batchDelete. What about persistResult or something similar

TBonnin (approved):
question about the `batchSave` naming but lg otherwise

PR #3881, opened 2025-04-10
feat(webapp, api): allow updating sync frequency for a specific conne…

steliosmavro:
## Issue

Currently, there's no way for users to edit the sync frequency of a connection directly from the dashboard UI. The only available methods are via the public API or Node SDK, which adds unnecessary friction compared to making the change directly in the interface. This limitation was highlighted in [Slack](https://nango-community.slack.com/archives/C04ABR352H0/p1744242419038669) by a user trying to modify the frequency for a Salesforce connection.

## Fix

Adds support for editing the sync frequency of individual connections directly from the dashboard UI. Users can now click a pencil icon next to the frequency cell to open a dialog, enter a new frequency, validate it, and update it via the API.

Implements the `syncController.updateFrequencyForConnection` controller under the hood.

Includes:
- Frequency edit UI with validation, enter key handler, and dynamic button state
- Integration with new `syncController.updateFrequencyForConnection` controller
- SWR refresh with proper loading & error handling

## Testing

1. Create an integration (if you don’t have one already).
2. Create a connection tied to that integration.
3. Navigate to the **Connections** page.
4. In the **Frequency** column, you should see a ✏️ pencil icon.
5. Clicking the icon opens a modal with an input for the new frequency.
6. Input is validated using a regex (`every <number> <unit>` format).
7. The **Save** button is disabled unless the frequency is valid _and_ different from the current one.
8. Hitting **Enter** inside the input or clicking **Save** should update the frequency and close the modal.

## Preview

Screenshot 2025-04-10 at 11 13 05 PM
Screenshot 2025-04-10 at 11 12 35 PM

---

This PR introduces the ability for users to update the sync frequency for specific connections directly from the dashboard UI, addressing feedback that previously only allowed this via API or SDK. The update includes front-end modal interactions, validation logic, and back-end endpoint additions to securely support frequency changes on a per-connection basis without editing all connection frequencies at once.

**Key Changes:**
• Adds a pencil/edit icon to the sync frequency column on the Connections dashboard, which opens a modal for frequency input.
• Implements front-end validation for frequency input with a shared regex and improved user feedback (error handling, disabled save if invalid).
• Introduces apiUpdateSyncConnectionFrequency in webapp, and a unified frequencyRegex utility for input validation.
• Extends the /api/v1/sync/update-connection-frequency API with controller, route, and types for updating frequency on a per-connection basis.
• Ensures SWR cache invalidation for sync data after successful update.
• Refactors related UI and moves frequency regex to a utils/validation file for DRY validation across components.

**Affected Areas:**
• Dashboard UI (SyncRow in Connection page)
• API routes and type declarations (server/routes.private, types/lib/sync/api.ts)
• Custom hooks for connection management (useConnections.tsx)
• Validation utility (utils/validation.ts)

### DIFF
--- a/packages/types/lib/sync/api.ts
+++ b/packages/types/lib/sync/api.ts
@@ -36,6 +36,21 @@ export type PostSyncVariant = Endpoint<{
     Success: { id: string; name: string; variant: string };
 }>;
 
+export type UpdateSyncConnectionFrequency = Endpoint<{
+    Method: 'PUT';
+    Path: '/api/v1/sync/update-connection-frequency';
+    Querystring: { env: string };
+    Body: {
+        connection_id: string;
+        provider_config_key: string;
+        name: string;
+        variant: string;
+        frequency: string | null;
+    };
+    Error: ApiError<'unknown_provider_config'> | ApiError<'unknown_connection'>;
+    Success: { data: { frequency: string } };
+}>;
+
 export type DeleteSyncVariant = Endpoint<{
     Method: 'DELETE';
     Path: '/sync/:name/variant/:variant';
--- a/packages/types/lib/sync/api.ts
+++ b/packages/types/lib/sync/api.ts
@@ -36,7 +36,7 @@ export type PostSyncVariant = Endpoint<{
     Success: { id: string; name: string; variant: string };
 }>;
 
-export type UpdateSyncConnectionFrequency = Endpoint<{
+export type UpdateConnectionFrequency = Endpoint<{
     Method: 'PUT';
     Path: '/api/v1/sync/update-connection-frequency';
     Querystring: { env: string };
--- a/packages/webapp/src/pages/Connection/components/SyncRow.tsx
+++ b/packages/webapp/src/pages/Connection/components/SyncRow.tsx
@@ -22,12 +22,15 @@ import { Button, ButtonLink } from '../../../components/ui/button/Button';
 import { Popover, PopoverTrigger } from '../../../components/ui/Popover';
 import { PopoverContent } from '@radix-ui/react-popover';
 import { SimpleTooltip } from '../../../components/SimpleTooltip';
-import { IconClockPause, IconClockPlay, IconRefresh, IconX } from '@tabler/icons-react';
+import { IconClockPause, IconClockPlay, IconRefresh, IconX, IconPencil } from '@tabler/icons-react';
 import { useToast } from '../../../hooks/useToast';
 import { mutate } from 'swr';
 import { Select, SelectItem, SelectContent, SelectTrigger, SelectValue } from '../../../components/ui/Select';
 import { Checkbox } from '../../../components/ui/Checkbox';
 import { apiRunSyncCommand } from '../../../hooks/useSyncs';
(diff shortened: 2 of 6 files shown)


packages/types/lib/sync/api.ts
```diff
@@ -36,6 +36,21 @@ export type PostSyncVariant = Endpoint<{
     Success: { id: string; name: string; variant: string };
 }>;
 
+export type UpdateSyncConnectionFrequency = Endpoint<{
```
TBonnin: I would have the name of the consistent with the endpoint path. wdyt? 
```suggestion
export type UpdateConnectionFrequency = Endpoint<{
```

packages/webapp/src/pages/Connection/components/SyncRow.tsx
```diff
@@ -124,6 +134,52 @@ export const SyncRow: React.FC<{ sync: SyncResponse; connection: ApiConnectionFu
         resetLoaders();
     };
 
+    const onFrequencyChange = (value: string) => {
+        setSelectedFrequency(value);
+
+        const reg = value.match(frequencyRegex);
+        if (!reg || !reg.groups) {
+            setFrequencyError('Format should be "every (number) (unit)", e.g: "every 1 hour", "every 20 days"');
```
TBonnin, line 142: I think the regex allows to skip the `every`

packages/webapp/src/pages/Connection/components/SyncRow.tsx
```diff
@@ -124,6 +134,52 @@ export const SyncRow: React.FC<{ sync: SyncResponse; connection: ApiConnectionFu
...
+            return;
+        }
+
+        const unit = reg.groups['unit'];
+        const amount = parseInt(reg.groups['amount'], 10);
+
+        if (unit.startsWith('s') && (amount < 30 || !amount)) {
+            setFrequencyError('The minimum frequency is 30 seconds');
```
TBonnin, line 150: fyi the min value depends on the customer plan.

steliosmavro:
Hey @TBonnin I implemented the changes you requested.

Just for context the regex was originally written by Samuel. I didn’t write it from scratch. I just abstracted it so it could be reused where I needed it. That said, I updated it exactly as you suggested and ran a bunch of test cases to verify. You were totally right, it works as expected now 🙌

Regarding the minimum frequency check: I based it on another part of the app (see attached screenshot), where it was already implemented like this. If we want to make it dynamic based on the customer’s plan, I’d be more than happy to help, but I assume that would involve a bit more work on the backend to expose the relevant data to the frontend.

Screenshot 2025-06-06 at 5 00 49 PM

TBonnin:
> Hey @TBonnin I implemented the changes you requested.

Thank you very much. The code doesn't seem to compile though.

> Regarding the minimum frequency check: I based it on another part of the app (see attached screenshot), where it was already implemented like this. If we want to make it dynamic based on the customer’s plan, I’d be more than happy to help, but I assume that would involve a bit more work on the backend to expose the relevant data to the frontend.

We can keep it as it is for now.

PR #3883, opened 2025-04-10
fix: clearing cache of variant syncs deletes records of the base sync

TBonnin:
- [fix: clearing cache of variant sync deletes records of the base sync](https://github.com/NangoHQ/nango/commit/b34204b76f58f63015a5d3f26fb3a97b47a2fa73) and therefore not clearing the cache for the variant sync.
- [fix: all rows of variant syncs shows records count of base sync](https://github.com/NangoHQ/nango/commit/dbf55c494893410ac807e99d4150f3a6f96fdced)

# How to test
1. create variant syncs that are saving records
2. Trigger full sync with cache clearing
3. Check the number of deleted records is correct in the logs and that sync is saving again the correct amount of records

### DIFF
--- a/packages/server/lib/controllers/sync.controller.ts
+++ b/packages/server/lib/controllers/sync.controller.ts
@@ -77,11 +77,15 @@ class SyncController {
 
     private async addRecordCount(syncs: (Sync & { models: string[] })[], connectionId: number, environmentId: number) {
         const byModel = await recordsService.getRecordCountsByModel({ connectionId, environmentId });
-
         if (byModel.isOk()) {
             return syncs.map((sync) => ({
                 ...sync,
-                record_count: Object.fromEntries(sync.models.map((model) => [model, byModel.value[model]?.count ?? 0]))
+                record_count: Object.fromEntries(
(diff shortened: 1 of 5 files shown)


packages/server/lib/controllers/sync.controller.ts
```diff
@@ -77,11 +77,15 @@ class SyncController {
...
         if (byModel.isOk()) {
             return syncs.map((sync) => ({
                 ...sync,
-                record_count: Object.fromEntries(sync.models.map((model) => [model, byModel.value[model]?.count ?? 0]))
+                record_count: Object.fromEntries(
+                    sync.models.map((model) => {
+                        const modelFullName = sync.variant === 'base' ? model : `${model}::${sync.variant}`;
+                        return [model, byModel.value[modelFullName]?.count ?? 0];
+                    })
+                )
```
TBonnin, line 88: unrelated to the clearing cache issue but while testing I realized the records count on the connections page was showing the base sync count for all variants :(

bodinsamuel (approved):

PR #3884, opened 2025-04-10
fix(runner-sdk): no need for a mutex

TBonnin:
you were right @bodinsamuel 
Mutex is not necessary when acquiring/releasing lock. No race condition can happen because of nodejs concurrency model.
I think I had a mutex because in my original implementation there were an async operation between the get and set (making it possible for the execution to yield) but that's not the case now

### DIFF
--- a/packages/runner/lib/sdk/locks.ts
+++ b/packages/runner/lib/sdk/locks.ts
@@ -10,7 +10,6 @@ interface Lock {
 
 export class Locks {
     private store = new Map<string, Lock>();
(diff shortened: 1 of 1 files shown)


bodinsamuel (approved):
I was okay keeping it because we will have async soon no?

TBonnin:
> I was okay keeping it because we will have async soon no?

1. soon might not be so soon :/
2. if we use pg as the backend, the locking will be done with a `SELECT FOR UPDATE`

removing it for now and we will reintroduce whatever we need when we need it

PR #3885, opened 2025-04-10
fix: do not increment/decrement DD metric if value = 0

TBonnin:
saving incrementing metric by 0 :)


bodinsamuel (approved):
What does it fix?

TBonnin:
> What does it fix?

nothing is broken but we probably keep some metrics/dimensions "alive" by incrementing them by 0. (for example a sync that keep running without ever saving any new records like all my tests syncs ;) ).  Not sure it is gonna be a significant amount though but not incrementing by 0 isn't going to affect the metric value anyway

PR #3887, opened 2025-04-11
fix(integrations): workday require user to add tenant and hostname as connectionConfigs

hassan254-prog:

- Update the `base_url` to require users to provide the `Tenant` and `Hostname` as `connectionConfig` values, since both are mandatory. Safe to update this as there are no real connections in the db against workday.
- Update the connect-ui docs to to clearly explain how to create an ISU(Integration System User) account, which is required for Basic Authentication.


khaliqgant:
Verified in the database that it won't break any users setup of Workday in Nango cloud.

bodinsamuel (approved):

TBonnin (approved):

PR #3888, opened 2025-04-11
fix(ui): logs fix history, virtualization, data loading and caching

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2995/improve-log-navigation

Big changes regarding logs interface. And it's not done yet

- **Fix history navigation**
I coded something badly regarding query params. I knew at the time by experience that when you need a two-way sync between page and query params you are in trouble. I decided to rewrite this logic and found `nuqs` which basically does that plus better parsing.

- **Add virtualization for operations list**
When loading a lot of rows in a UI you usually quickly hit a bottleneck and/or crash, mostly because of DOM, not memory. Virtualization keeps the DOM short and the rest in memory.
It now supports thousand of operations and loads them much faster

- **Fix data loading and caching**
When I did this page we only had SWR which doesn't support POST. I immediately hated Vercel a little bit more but decided to bite the bullet and recode my own data fetching. Building a UI is hard and it's easy to get lost in `useEffect` and such. The code was definitely messy and not always working. 
So I decided to stop pretending SWR is good and introduce tanstack react-query which is ultimately a better solution for data fetching in every way. 
I replaced my code with `useInfiniteQuery` which does 90% of what I wanted, except one annoying thing about interval refresh refreshing every page.
On top of that, we have better caching and un-caching

## 🧪 Tests

- Stress the UI

https://github.com/user-attachments/assets/c5d21aa8-c12e-4ec6-a928-fa04040b7cbe

- You can use autocannon to push more operations
```sh
npx autocannon -a 10000 http://localhost:3003/action/trigger \
-H"Authorization: REDAC" \
--header 'Connection-Id: 765b3a78-11ed-402b-8385-09708519ffad'\
--header 'Content-Type: application/json'\
-m POST -b '{"action_name": "returnBase","input": { "foobar": "foobar"}}'

### CODE REVIEW

File: packages/webapp/src/pages/Logs/components/SearchAllOperations.tsx [hunk 1]
```diff
@@ -0,0 +1,309 @@
...
+            // Slide the window automatically when live
+            // We do it only at query time so the URL stays the same
+            if (isLive) {
+                const tmp = slidePeriod({ from: period[0], to: period[1] });
+                periodCopy = { from: tmp.from!.toISOString(), to: tmp.to!.toISOString() };
+            } else {
+                periodCopy = { from: new Date(period[0]).toISOString(), to: new Date(period[1]).toISOString() };
+            }
+
+            // await wait(1000);
```
TBonnin: 🙈

packages/webapp/src/pages/Logs/components/SearchAllOperations.tsx
```diff
@@ -0,0 +1,309 @@
...
+    const totalFetched = flatData.length;
+    const fetchMoreOnBottomReached = useCallback(
+        (containerRefElement?: HTMLDivElement | null) => {
+            if (!containerRefElement) {
+                return;
+            }
+
+            const { scrollHeight, scrollTop, clientHeight } = containerRefElement;
+            // once the user has scrolled within 200px of the bottom of the table, fetch more data if we can
+            if (scrollHeight - scrollTop - clientHeight < 200 && !isFetching && totalFetched < totalOperations) {
```
TBonnin, line 169: should the threshold depend on the size of the window? also do we care about mobile?

packages/webapp/src/pages/Logs/components/SearchAllOperations.tsx
```diff
@@ -0,0 +1,309 @@
...
+    return (
+        <>
+            <div className="flex justify-between items-center">
+                <h2 className="text-3xl font-semibold text-white mb-2 flex gap-4 items-center">Logs {(isLoading || isFetching) && <Spinner size={1} />}</h2>
+                <div className="text-white text-xs">
+                    {totalHumanReadable} {totalOperations > 1 ? 'logs' : 'log'} found
+                </div>
+            </div>
+            <div className="flex gap-2 justify-between mb-4">
+                <div className="w-full">{/* <Input before={<MagnifyingGlassIcon className="w-5 h-5" />} placeholder="Search operations..." /> */}</div>
```
TBonnin: 🙈

packages/webapp/src/pages/Logs/components/SearchAllOperations.tsx
```diff
@@ -0,0 +1,309 @@
...
+    table: ReactTable<SearchOperationsData>;
+    tableContainerRef: React.RefObject<HTMLDivElement>;
+    onSelectOperation: Props['onSelectOperation'];
+}> = ({ table, tableContainerRef, onSelectOperation }) => {
+    const { rows } = table.getRowModel();
+
+    // Important: Keep the row virtualizer in the lowest component possible to avoid unnecessary re-renders.
+    const rowVirtualizer = useVirtualizer<HTMLDivElement, HTMLTableRowElement>({
+        count: rows.length,
+        estimateSize: () => 41, // estimate row height for accurate scrollbar dragging
```
TBonnin: curious to know how this value is estimated

packages/webapp/src/pages/Logs/constants.tsx
```diff
@@ -42,24 +37,27 @@ export const columns: ColumnDef<SearchOperationsData>[] = [
...
         cell: ({ row }) => {
             return <div className="truncate font-code text-s">{row.original.syncConfigName || '-'}</div>;
         }
     },
     {
         accessorKey: 'connectionId',
         header: 'Connection',
-        minSize: 0,
-        size: 0,
+        // minSize: 290,
```
TBonnin: 🙈

TBonnin (approved):
tested locally. It is fast 
giphy

packages/webapp/src/utils/table.ts
```diff
@@ -0,0 +1,63 @@
...
+        else totalAvailableWidth -= getSize(column.size, column.maxSize, column.minSize);
+    });
+
+    const sizing: Record<string, number> = {};
+
+    columns.forEach((header) => {
+        const column = header.column.columnDef;
+        if (column.meta?.isGrow) {
+            let calculatedSize = 100;
+            calculatedSize = Math.floor(totalAvailableWidth / totalIsGrow);
```
TBonnin: can we ever be dividing by 0 here?

packages/webapp/src/utils/table.ts
```diff
@@ -0,0 +1,63 @@
...
+            const size = getSize(calculatedSize, column.maxSize, column.minSize);
+            column.size = size;
```
TBonnin, line 56: function is called calculate but we are mutating the column size here. is that ok?

packages/webapp/src/utils/table.ts
```diff
@@ -0,0 +1,63 @@
...
+                }
+
+                const size = getSize(calculatedSize, column.maxSize, column.minSize);
+
+                column.size = size;
+            }
+        }
+
+        if (column.meta?.isGrow) totalIsGrow += 1;
+        else totalAvailableWidth -= getSize(column.size, column.maxSize, column.minSize);
```
TBonnin, line 45: can totalAvailableWidth ever become negative?

packages/webapp/src/utils/table.ts
```diff
@@ -0,0 +1,63 @@
...
+ *
+ * @param {Header<DataType, unknown>[]} columns - An array of column headers. Each header contains
+ *   metadata about the column, including size, constraints, and growth behavior.
+ * @param {number} totalWidth - The total width available for the table, including padding and margins.
+ *
+ * @returns {Record<string, number>} An object mapping column IDs to their calculated sizes.
+ */
+export const calculateTableSizing = (columns: Header<any, unknown>[], totalWidth: number): Record<string, number> => {
+    let totalAvailableWidth = totalWidth;
+    let totalIsGrow = 0;
```
TBonnin: I don't find the naming super clear. what about `growingColumnsCount` or something similar?

packages/webapp/src/pages/Logs/components/SearchAllOperations.tsx
bodinsamuel: it's 200px below the screen so it should be good for any screen size I think

packages/webapp/src/pages/Logs/components/SearchAllOperations.tsx
bodinsamuel: it's the actual height of the rows, I'll remove the comment saying it's the estimation :D

packages/webapp/src/utils/table.ts
bodinsamuel: I don't think so. 
nb: this whole function was copied from their github

packages/webapp/src/utils/table.ts
bodinsamuel: yes that's on purpose unfortunately

packages/webapp/src/utils/table.ts
bodinsamuel: hopefully not :D

packages/webapp/src/utils/table.ts
bodinsamuel: changing that

PR #3891, opened 2025-04-11
feat(db): add soft delete columns to environment

kaposke:

### DIFF
--- a/packages/database/lib/migrations/20250409232749_add_soft_delete_to_environment.cjs
+++ b/packages/database/lib/migrations/20250409232749_add_soft_delete_to_environment.cjs
@@ -1,0 +1,13 @@
+exports.up = function (knex) {
+    return knex.schema.alterTable('_nango_environments', function (table) {
+        table.boolean('deleted').defaultTo(false);
+        table.dateTime('deleted_at').defaultTo(null);
+    });
+};
+
+exports.down = function (knex) {
+    return knex.schema.alterTable('_nango_environments', function (table) {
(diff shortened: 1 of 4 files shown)


packages/database/lib/migrations/20250409232749_add_soft_delete_to_environment.cjs
```diff
@@ -0,0 +1,13 @@
+exports.up = function (knex) {
+    return knex.schema.alterTable('_nango_environments', function (table) {
+        table.boolean('deleted').defaultTo(false);
+        table.dateTime('deleted_at').defaultTo(null);
+    });
```
bodinsamuel, line 5: We'll talk about index after, but you will need a few

packages/types/lib/environment/db.ts
```diff
@@ -46,6 +46,9 @@ export interface DBEnvironment extends Timestamps {
 
     webhook_receive_url: string | null;
     otlp_settings: { endpoint: string; headers: Record<string, string> } | null;
+
+    deleted: boolean;
+    deleted_at?: Date | null;
```
bodinsamuel: can't be undefined
```suggestion
    deleted_at: Date | null;
```

Alternatively you can change the extends to use `TimestampsAndDeletedCorrect`

bodinsamuel (approved):

packages/types/lib/environment/db.ts
kaposke: Good to know

PR #3898, opened 2025-04-14
fix(gong-oauth): make api_base_url_for_customer field not required

hassan254-prog:

- Make field not requrired to allow for Nango to grab the `api_base_url_for_customer` from the token response.


bodinsamuel (commented):
Automated means there is a pre-connection script, which we don't have.
Are you sure that's the fix? 
also it's missing the context so it's harder to understand the fix

hassan254-prog:
> Automated means there is a pre-connection script, which we don't have. Are you sure that's the fix? also it's missing the context so it's harder to understand the fix

When a user creates a new connection in Nango, the `token_response_metadata` will grab and store the `api_base_url_for_customer` from the token response, this will then be used to build the `base_url` for proxy calls.

hassan254-prog:
> Automated means there is a pre-connection script, which we don't have. Are you sure that's the fix? also it's missing the context so it's harder to understand the fix

Sorry for missing full context here, when a user creates a new connection in Nango, the `token_response_metadata` action will grab and store the `api_base_url_for_customer` from the token response, this will then be used to build the `base_url` for proxy calls, eliminating the need for users to have to input this value.

bodinsamuel (approved):

bodinsamuel:
> Sorry for missing full context here, when a user creates a new connection in Nango, the token_response_metadata action will grab and store the api_base_url_for_customer from the token response, this will then be used to build the base_url for proxy calls, eliminating the need for users to have to input this value.

Aah true 👌🏻

PR #3899, opened 2025-04-14
feat: delete environments

kaposke:
## Description
This PR adds the ability to (soft) delete environments, from frontend to backend.

### Frontend
A `Delete environment` button was added to the environment-settings (doesn't show up for the prod environment).
image

A `DestructiveActionModal` component requires the environment name to be typed in to confirm deletion.
image

### Backend
The environment is soft-deleted along with:
- It's Provider/Integration configs (nango_config) - Ps: we need to consolidate on a name for these
- Syncs - and schedules are deleted
- Sync Configs
- Connections

Hard deletion of old soft-deleted environments (and it's dependents) will be added in a following PR.

## Testing
Please checkout to branch and test deleting environments in the UI.
Also, help me verify that anything that wastes computation or would impact billing is stopped / soft-deleted by my changes.


packages/shared/lib/services/environment.service.ts
```diff
@@ -483,6 +475,22 @@ class EnvironmentService {
...
+        const configs = await configService.listProviderConfigs(environmentId);
+        for (const config of configs) {
+            // This handles deleting connections and syncs down the line
+            await configService.deleteProviderConfig({
+                id: config.id!,
+                environmentId,
+                providerConfigKey: config.unique_key,
+                orchestrator
+            });
+        }
```
kaposke: @bodinsamuel @TBonnin I believe `configService.deleteProviderConfig` does everything I was talking about: soft deletes relevant dependencies and stops schedules from running. Can you verify that that's true?

packages/shared/lib/services/environment.service.ts
```diff
@@ -229,8 +217,12 @@ class EnvironmentService {
...
 
     async getAll(): Promise<{ environmentId: number; accountId: number }[]> {
-        const result = await db.knex.select('id as environmentId', 'account_id as accountId').from<{ environmentId: number; accountId: number }[]>(TABLE);
-        return result || [];
+        const result = await db.knex
+            .select('id as environmentId', 'account_id as accountId', 'deleted')
+            .from<{ environmentId: number; accountId: number; deleted: boolean }>(TABLE)
+            .where({ deleted: false });
+
+        return result.map(({ environmentId, accountId }) => ({ environmentId, accountId }));
```
TBonnin: I cannot find any usage of this function. We can get rid of it then.

packages/shared/lib/services/environment.service.ts
TBonnin: connections, sync configs, syncs and schedules are soft deleted so it looks good to me

packages/webapp/src/pages/Environment/Settings/Show.tsx
```diff
@@ -65,6 +96,9 @@ export const EnvironmentSettings: React.FC = () => {
 
             <div className="flex justify-between mb-8 items-center">
                 <h2 className="flex text-left text-3xl font-semibold tracking-tight text-white">Environment Settings</h2>
+                {env !== PROD_ENVIRONMENT_NAME && (
```
TBonnin, line 99: @bastienbeurier should we keep the button but disable it with a tooltip to explain why it is not deletable. Otherwise customers might wonder why the button is not here

TBonnin (approved):
👍 Comments are not blocking. Thank you for doing this

packages/shared/lib/services/environment.service.ts
```diff
@@ -483,6 +475,22 @@ class EnvironmentService {
...
+        // TODO: Ideally we would soft delete everything down the tree in a transaction
+        const configs = await configService.listProviderConfigs(environmentId);
+        for (const config of configs) {
+            // This handles deleting connections and syncs down the line
+            await configService.deleteProviderConfig({
+                id: config.id!,
+                environmentId,
+                providerConfigKey: config.unique_key,
+                orchestrator
+            });
```
bodinsamuel, line 479: The called method needs a refactor but that's for later

packages/shared/lib/services/environment.service.ts
```diff
@@ -483,6 +475,22 @@ class EnvironmentService {
 
         return globalCallbackUrl;
     }
+
+    async softDelete({ environmentId, orchestrator }: { environmentId: number; orchestrator: Orchestrator }): Promise<void> {
+        await db.knex.from<DBEnvironment>(TABLE).where({ id: environmentId, deleted: false }).update({ deleted: true, deleted_at: new Date() });
+
+        // TODO: Ideally we would soft delete everything down the tree in a transaction
```
bodinsamuel: We won't do it in a transaction ever so you can remove this part imo. A more efficient strategy is to add joins everywhere unfortunately

packages/server/lib/controllers/v1/environment/deleteEnvironment.integration.test.ts
```diff
@@ -0,0 +1,204 @@
...
+        }
+
+        const providerConfig = await configService.createProviderConfig(
+            {
+                environment_id: testEnv.id,
+                unique_key: providerName,
+                provider: providerName
+            },
+            provider
+        );
```
bodinsamuel: createConfigSeed()

packages/server/lib/controllers/v1/environment/deleteEnvironment.integration.test.ts
```diff
@@ -0,0 +1,204 @@
...
+                    enabled: true,
+                    models: [],
+                    auto_start: true,
+                    deleted: false,
+                    nango_config_id: providerConfig!.id!,
+                    version: '1'
+                })
+                .returning('*')
+        )[0];
+        expect(syncConfig).not.toBeNull();
```
bodinsamuel: createSyncSeeds() (after creating connection)

packages/shared/lib/services/environment.service.ts
```diff
@@ -483,6 +475,22 @@ class EnvironmentService {
 
         return globalCallbackUrl;
     }
+
+    async softDelete({ environmentId, orchestrator }: { environmentId: number; orchestrator: Orchestrator }): Promise<void> {
+        await db.knex.from<DBEnvironment>(TABLE).where({ id: environmentId, deleted: false }).update({ deleted: true, deleted_at: new Date() });
```
bodinsamuel: update at the end

bodinsamuel (commented):
Not tested yet, but looks good.
Can you create a following linear issues to add at least those indexes:
- account_id, name where deleted = false
- id where deleted = false 
- secret_key_hashed where deleted = false
- public_key where deleted = false

packages/shared/lib/services/environment.service.ts
kaposke: what do you mean?

packages/shared/lib/services/environment.service.ts
kaposke: That sounds unfeasible for the lifetime of the project. Will remove the TODO though.

packages/webapp/src/pages/Environment/Settings/Show.tsx
kaposke: I'l merge this for now and add that button in a followup PR along with a couple UI additions I have inline

packages/shared/lib/services/environment.service.ts
bodinsamuel: you should update the env after the soft delete not before, so you get a change to try again if that fails

TBonnin (approved):

PR #3900, opened 2025-04-14
feat: calculate billable connections metric

TBonnin:
A billable connection is currently any connection that has not been deleted. 
Connections are prorated if created or deleted during the month

It is not yet exported to the billing platform yet. We are currently evaluating 3rd party solutions


packages/server/lib/crons/usage.ts
```diff
@@ -0,0 +1,88 @@
+import tracer from 'dd-trace';
```
TBonnin, line 1: I went for reusing the existing usage cron. I renamed it to `usage` and moved functions around but it is basically just adding a extra function to calculate billable connections

packages/server/lib/crons/usage.ts
```diff
@@ -0,0 +1,88 @@
...
+                metrics.gauge(metrics.Types.RECORDS_TOTAL_COUNT, Number(countRes.value.count));
+            } catch (err) {
+                span.setTag('error', err);
+                report(new Error('cron_failed_to_export_records_metrics', { cause: err }));
+            }
+        });
+    }
+};
+
+const billing = {
```
TBonnin, line 73: below is the added function

packages/server/lib/crons/usage.ts
```diff
@@ -0,0 +1,88 @@
...
+import { connectionService } from '@nangohq/shared';
+import { getLogger, metrics, report, flagHasUsage } from '@nangohq/utils';
+
+import { envs } from '../env.js';
+
+const logger = getLogger('cron.exportUsage');
+const cronMinutes = envs.CRON_EXPORT_USAGE_MINUTES;
+
+export function exportUsageCron(): void {
+    if (!flagHasUsage || cronMinutes <= 0) {
```
TBonnin, line 14: added a flag to disable usage reporting by default (so nothing happen on Enterprise setup).

packages/shared/lib/services/connection.service.ts
```diff
@@ -1742,6 +1742,90 @@ class ConnectionService {
...
+                month: number;
+            }[],
+            NangoError
+        >
+    > {
+        // Note:
+        // a billable connection is a connection that is not deleted and has not been deleted during the month
+        // connections are pro-rated based on the number of seconds they were existing in the month
+
+        const targetDate = new Date(Date.UTC(referenceDate.getUTCFullYear(), referenceDate.getUTCMonth(), referenceDate.getUTCDate(), 0, 0, 0, 0));
```
TBonnin, line 1760: went with setting billing start/end of month to be UTC. I guess at some point we are gonna want to be account specific but I don't think we have an account timezone setting yet and unless there is any legal requirement using UTC is simpler

packages/server/lib/crons/usage.ts
```diff
@@ -0,0 +1,88 @@
...
+            }
+        });
+    }
+};
+
+const billing = {
+    exportBillableConnections: async (): Promise<void> => {
+        await tracer.trace<Promise<void>>('nango.cron.exportUsage.billing.connections', async (span) => {
+            try {
+                const res = await connectionService.billableConnections(new Date());
```
TBonnin, line 77: because of the nature of cron jobs, we might need to run 2 queries to handle the switch from one month to another. We'll see if we need this when we export the data to the billing platform

packages/shared/lib/services/connection.service.ts
```diff
@@ -1742,6 +1742,90 @@ class ConnectionService {
...
+                    db.readOnly.raw(`FLOOR(SUM(connection_weight)) as count`),
+                    db.readOnly.raw(`${year} as year`),
+                    db.readOnly.raw(`${month} as month`)
+                )
+                    .from('prorated')
+                    .groupBy('account_id');
+            })
+            .select('*')
+            .from('totals')
+            .where('count', '>', 0);
```
TBonnin, line 1821: FYI: query is currently taking ~500ms against main db replica

packages/shared/lib/services/connection.service.ts
```diff
@@ -1742,6 +1742,90 @@ class ConnectionService {
...
+                    'e.account_id',
+                    'c.id as connection_id',
+                    'c.created_at',
+                    db.readOnly.raw(`COALESCE(c.deleted_at, (SELECT month_end FROM month_info) + INTERVAL '1 day') AS effective_end_date`),
+                    db.readOnly.raw(`(SELECT month_start FROM month_info) AS month_start`),
+                    db.readOnly.raw(`(SELECT month_end FROM month_info) AS month_end`),
+                    db.readOnly.raw(`(SELECT total_seconds_in_month FROM month_info) AS total_seconds_in_month`)
+                )
+                    .from('nango._nango_connections as c')
+                    .join('nango._nango_environments as e', 'c.environment_id', 'e.id')
```
TBonnin, line 1786: does ALL the connections for ALL environments count for billing? @bastienbeurier

bastienbeurier: Yes!

packages/server/lib/crons/usage.ts
```diff
@@ -0,0 +1,88 @@
...
+}
+
+const observability = {
+    exportConnectionsMetrics: async (): Promise<void> => {
+        await tracer.trace<Promise<void>>('nango.cron.exportUsage.observability.connections', async (span) => {
+            try {
+                const connRes = await connectionService.countMetric();
+                if (connRes.isErr()) {
+                    throw connRes.error;
+                }
```
bodinsamuel, line 44: if you always throw you could just unwrap but that's just me :D

packages/shared/lib/services/connection.service.ts
bodinsamuel: just me: it's definitely a case where the query builder is definitely less readable and impossible to debug without logging it first

packages/shared/lib/services/connection.service.ts
bodinsamuel: I thought dev env was not billed?

bodinsamuel (approved):
not tested but looks good

packages/server/lib/crons/usage.ts
TBonnin: indeed but I like to follow the principle to never `unwrap` in prod code. (only tests or one-off script)

PR #3901, opened 2025-04-14
feat(ui): improve operation search, perf, caching

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2993/improve-sub-logs-when-theres-many
Fixes https://linear.app/nango/issue/NAN-3089/improve-operation-logs-navigation-and-perf

- Improve operation search, perf, caching
Same PR as the previous one #3888 The diff is that I had already coded the cursorBefore so it was a bit easier. But not really since there is a search bar that's annoying to handle.

- Renamed some files
- Changed some icon

## 🧪 Tests
- Go mess with the UI

https://github.com/user-attachments/assets/1b728977-4567-4eb1-a484-3624d24d424a


packages/webapp/src/pages/Logs/Operation/Show.tsx
```diff
@@ -28,12 +29,14 @@ export const ShowOperation: React.FC<{ operationId: string }> = ({ operationId }
...
     }, [operation]);
+
     const createdAt = useMemo(() => {
         return operation?.createdAt ? formatDateToLogFormat(operation?.createdAt) : 'n/a';
     }, [operation?.createdAt]);
 
     const isLive = useMemo(() => {
-        return !operation || operation.state === 'waiting' || operation.state === 'running';
+        // We keep refreshing N seconds after end to catch logs that could be indexed after the operation is done
+        return !operation || !operation.endedAt || new Date(operation.endedAt).getTime() > Date.now() - 160_000;
```
TBonnin, line 39: why 160 btw?

packages/webapp/src/pages/Logs/Operation/components/Logs.tsx
```diff
@@ -0,0 +1,304 @@
...
+                    cursorAfter: pageParam && 'after' in pageParam ? pageParam.after : undefined,
+
+                    // Some cache bug issue I haven't solved yet
+                    // Lib is sometimes storing an old pageParam, when we re-open the Drawer it's incorrect in the past
+                    // So we discard any cursor that are too old unless we search
+                    cursorBefore:
+                        pageParam &&
+                        'before' in pageParam &&
+                        pageParam.before &&
+                        (JSON.parse(atob(pageParam.before))[0] > Date.now() - 30_000 || debouncedSearch)
```
TBonnin, line 65: is it linked to the staleTime below?

TBonnin (approved):
tested locally. works well

packages/webapp/src/pages/Logs/Operation/Show.tsx
bodinsamuel: The comment is explaining why, and for the value it's completely arbitrary :D

packages/webapp/src/pages/Logs/Operation/components/Logs.tsx
bodinsamuel: yes and no, staleTime makes react-query refetch stuff but it will re-use the param anyway, so when I refetch with the same params after 30s I'm specifically removing the cursor

PR #3904, opened 2025-04-15
fix(integrations): use the correct base url

hassan254-prog:

- Update the API `base-url` to use the correct url instead of the token [Introspect](https://developers.attio.com/reference/post_oauth-introspect) `base-url`.


hassan254-prog:
@bodinsamuel , is it possible to check if we have users using this so that we can update without breaking for other users ☺️

bodinsamuel (approved):

PR #3906, opened 2025-04-15
feat(logs): global search

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-3039/top-level-log-search

- Logs global search

You will see some comments in the PR, but basically, the main challenge is that it's a **two-step process**. First, we do a regular search on operations with the filters (if any), and then we search inside the `operationIds` that were found.

**Doing post-filtering** instead of pre-filtering saves a lot of headaches for pagination and security, but it's not perfect. 
Since the pagination is working (we have a valid cursor) but the limit might be incorrect (we asked for 20 and got less) the UI will ask for more due to the infinite scroll.
To prevent excessive HTTP calls, I have added **a manual mode**. As soon as you type, we stop the live refresh and disable infinite scroll. It will still try to load as much as it can when we possibly, but the initial load is always manual.

I'm actually surprised the PR is that small, hopefully it's not bringing too much bugs 🙏🏻 

https://github.com/user-attachments/assets/a1362512-986f-4373-8255-3645144e938e


TBonnin (approved):
tested locally. worked well.
The total logs found (top right corner) is not matching the logs being showed. Showing the total amount when the search is actually filtering rows. Is that on purpose to always show the total, regardless of the search

I would also expect the search to "follow me" when opening an operation panel. If I search for something and then click on an operation that has lots of logs. I must search again to only see the specific log entries that matches what I am looking for

bodinsamuel:
>The total logs found (top right corner) do not match the. Showing the total amount when the search is actually filtering rows. Is that on purpose to always show the total, regardless of the search

Not really on purpose but it's because I don't have a way to have actual accurate numbers. The only way to fix all issues regarding count and pagination would be to:
- do a first search, getting all operations that match the filters (I mean all that could be 10K)
- Search in all of those `operationsIds`, getting a list of operations IDs that can match the search
- do a third search to get the first 20 operations that match both

Not sure if worth it tbh

>I would also expect the search to "follow me" when opening an operation panel. If I search for something and then click on an operation that has lots of logs. I must search again to only see the specific log entries that matches what I am looking for

Yes good feedback, I'm planning on doing that just after this PR ☺️

PR #3908, opened 2025-04-15
feat(cli): add message when all files compile successfully in `nango dev`

kaposke:
### Description
We currently have no feedback for when all files compile successfully, leading to a confusing DX when the last thing you see is the last error you got.

This PR adds this:
image
> Feel free to suggest another message

To achieve this, I had to keep track of failed files and tweak how things were done. I believe I've fixed a couple bugs as a tangent. I'll add comments to the relevant sections with more details.

### Testing
Please checkout to the branch and run `nango dev` in a nango project. Break and fix typescript files and the config file, in mixed order, to see if the DX is looking good. The `No compilation errors.` message should only show up when everything is looking good, both with TS files as well as `nango.yaml`.


packages/cli/lib/__snapshots__/sync.unit.cli-test.ts.snap
```diff
@@ -92,83 +92,3 @@ models:
         __extends: GithubIssue
 "
 `;
-
-exports[`generate function tests > should init the expected files in the nango-integrations directory 1`] = `
```
kaposke: These were obsolete apparently

packages/cli/lib/index.ts
```diff
@@ -164,11 +165,7 @@ program
         const fullPath = process.cwd();
         await verificationService.necessaryFilesExist({ fullPath, autoConfirm, debug, checkDist: false });
 
-        if (compileInterfaces) {
```
kaposke, line 171: From what I've gathered, this flag skipped watching the config file here, but it was still watched in `tscWatch`. I needed to track it there anyway, so I've moved the flag to `tscWatch`.

packages/cli/lib/cli.ts
```diff
@@ -129,14 +133,19 @@ export function generate({ fullPath, debug = false }: { fullPath: string; debug?
     }
 }
 
-export function tscWatch({ fullPath, debug = false }: { fullPath: string; debug?: boolean }) {
-    const tsconfig = fs.readFileSync(path.resolve(getNangoRootPath(), 'tsconfig.dev.json'), 'utf8');
-    const parsed = loadYamlAndGenerate({ fullPath, debug });
-    if (!parsed) {
```
kaposke, line 133: This meant that if you started `nango dev` with a broken config, it would never watch for ts files during that run.

packages/cli/lib/cli.ts
```diff
@@ -129,14 +133,19 @@ export function generate({ fullPath, debug = false }: { fullPath: string; debug?
...
+        console.log(chalk.green('No compilation errors.'));
     }
+}
 
-    const watchPath = ['./**/*.ts', `./${nangoConfigFile}`];
+export function tscWatch({ fullPath, debug = false, watchConfigFile }: { fullPath: string; debug?: boolean; watchConfigFile: boolean }) {
+    const tsconfig = fs.readFileSync(path.resolve(getNangoRootPath(), 'tsconfig.dev.json'), 'utf8');
+
+    const watchPath = ['./**/*.ts'];
+    if (watchConfigFile) {
```
kaposke, line 142: Here's the moved condition from the command function

packages/cli/lib/cli.ts
```diff
@@ -146,7 +155,7 @@ export function tscWatch({ fullPath, debug = false }: { fullPath: string; debug?
         ignoreInitial: false,
         ignored: (filePath: string) => {
             const relativePath = path.relative(__dirname, filePath);
-            return relativePath.includes('node_modules') || path.basename(filePath) === TYPES_FILE_NAME;
+            return relativePath.includes('node_modules') || path.basename(filePath) === TYPES_FILE_NAME || relativePath.includes('.nango');
```
kaposke, line 154: `.nango/schema.ts` was updated frequently, causing events to fire multiple times. Since it's auto-generated, should it be fine? I'm taking feedback on this.

packages/cli/lib/cli.ts
```diff
@@ -159,15 +168,65 @@ export function tscWatch({ fullPath, debug = false }: { fullPath: string; debug?
...
+    // First parsing of the config file
+    let parsed: NangoYamlParsed | null = loadYamlAndGenerate({ fullPath, debug });
+
+    const failedFiles = new Set<string>();
+
+    watcher.on('add', (filePath: string) => {
+        if (filePath === nangoConfigFile || !parsed) {
+            return;
+        }
+        void compileSingleFile({
```
kaposke: Changed everything to `.then` syntax to clear warnings.

packages/cli/lib/cli.ts
```diff
@@ -129,14 +133,19 @@ export function generate({ fullPath, debug = false }: { fullPath: string; debug?
...
-export function tscWatch({ fullPath, debug = false }: { fullPath: string; debug?: boolean }) {
-    const tsconfig = fs.readFileSync(path.resolve(getNangoRootPath(), 'tsconfig.dev.json'), 'utf8');
-    const parsed = loadYamlAndGenerate({ fullPath, debug });
-    if (!parsed) {
-        return;
+function showCompilationMessage(failedFiles: Set<string>) {
+    if (failedFiles.size === 0) {
+        console.log(chalk.green('No compilation errors.'));
     }
+}
```
TBonnin, line 136: not sure you need a function for that. Inlining the if would be good enough imho

packages/cli/lib/cli.ts
TBonnin: .nango files are generated from the nango.yaml so It should be fine not watching them

packages/cli/lib/cli.ts
TBonnin: @bodinsamuel is not gonna like that. 💥 😂  
More seriously, which warnings are we taking about?

kaposke: The callback shouldn't return a promise
```
Promise returned in function argument where a void return was expected.eslint[@typescript-eslint/no-misused-promises](https://typescript-eslint.io/rules/no-misused-promises)
```

packages/cli/lib/cli.ts
kaposke: If that's a NIT, I prefer having it. I generally like to avoid duplicating user-facing messages. It should be updatable in a single place when needed.

packages/cli/lib/cli.ts
TBonnin: would this not work?
```        
         const success = await compileSingleFile({
            fullPath,
            file: getFileToCompile({ fullPath, filePath }),
            tsconfig,
            parsed,
            debug
        });
        if (success) {
            failedFiles.delete(filePath);
        } else {
            failedFiles.add(filePath);
        }
        showCompilationMessage(failedFiles);
```

kaposke: Works, but with a warning because you have to make the function async, which makes it return a promise.

bodinsamuel: I think the "proper fix" is to wrap it in a function. I'm not a promise nazi it's just for consistency and clarifies that the callback will not await results which could lead to race condition.
e.g:
```ts
setTimeout(() => {
    async function fetchMyAPI() {
      let response = await fetch('api/data')
    }

    void fetchMyAPI();
});
```

package.json
```diff
@@ -43,6 +43,7 @@
         "test:unit": "vitest",
         "test:integration": "vitest --config ./vite.integration.config.ts",
         "test:cli": "vitest --config ./vite.cli.config.ts",
+        "test:cli:update-snapshots": "vitest --config ./vite.cli.config.ts --update",
```
bodinsamuel, line 46: note you can just press `u` when running tests to update snapshots

bodinsamuel (commented):
Seems to work, just one comment
- When I use dev locally it parse the nango.yaml once per file and every time I save a script too, which is super noisy when you have warning (and also probably not necessary)

Unrelated
- compile.service.ts:145 it's missing chalk.red, if you can take the opportunity to fix it that would be nice 🙏🏻

bodinsamuel (commented):
I still have one nango.yaml parsing per file or per save. 
and no success message at the end


https://github.com/user-attachments/assets/c6473a3f-9e03-4a66-8d91-328fb8b40720

kaposke:
> I still have one nango.yaml parsing per file or per save. and no success message at the end
> 
>  Screen.Recording.2025-04-17.at.10.55.42.mov

Oh thanks for the video. I'll try to reproduce it.

bodinsamuel:
You get no message when you have an external lib, put some helpers outside your sync folder, and import them in your sync. It's a common pattern so it should be handled

kaposke:
> You get no message when you have an external lib, put some helpers outside your sync folder, and import them in your sync. It's a common pattern so it should be handled

Thanks. I was able to reproduce it and fix it. The problem was that non sync/action files are not compiled on change. They are only compiled when the sync/action is changed, and they get bundled together in the output JS. That actually raises another problem, which is beyond the scope of this PR. Raised [NAN-3152](https://linear.app/nango/issue/NAN-3152/nango-dev-doesnt-recompile-files-when-a-dependency-file-is-updated)

Anyways, ignored files were being treated as failures. By changing that the message now works. However, if you break the dependency file, it will only show compilation errors when you change it's dependent file.

bodinsamuel:
Makes sense, thanks for investigating. I think the compilation is unnecessarily complex right now. We had plan in (not moving much unfortunately) to rewrite how we handle scripts and compilation that would remove the need for all of this

bodinsamuel (approved):

PR #3919, opened 2025-04-16
fix(webapp): apply default scopes for oauth2

hassan254-prog:

- Fix webapp to apply default scopes for oauth2


TBonnin (approved):

packages/webapp/src/pages/Integrations/providerConfigKey/Settings/components/OAuth.tsx
```diff
@@ -24,12 +24,17 @@ export const SettingsOAuth: React.FC<{ data: GetIntegration['Success']['data'];
     const [loading, setLoading] = useState(false);
     const [clientId, setClientId] = useState(integration.oauth_client_id || '');
     const [clientSecret, setClientSecret] = useState(integration.oauth_client_secret || '');
-    const [scopes, setScopes] = useState(integration.oauth_scopes || '');
+    const [scopes, setScopes] = useState(template.default_scopes || '');
```
bodinsamuel: that's not good you are just removing the scopes here and default_scopes is a static string so it will always override.
it should be done in the backend when we create the integration

hassan254-prog: This is now long overdue, sorry for that @bodinsamuel , updated with the latest [commit](https://github.com/NangoHQ/nango/pull/3919/commits/406dcc79ce7598047d3f1fffc4fcc9e7e9ca68b2)

packages/shared/lib/services/config.service.ts
```diff
@@ -82,7 +82,10 @@ class ConfigService {
     async createProviderConfig(config: DBCreateIntegration, provider: Provider): Promise<IntegrationConfig | null> {
         const configToInsert = config.oauth_client_secret ? encryptionManager.encryptProviderConfig(config as ProviderConfig) : config;
         configToInsert.missing_fields = this.validateProviderConfig(provider.auth_mode, config as ProviderConfig);
-
+        if (provider.default_scopes?.length) {
+            configToInsert.oauth_scopes = provider.default_scopes.join(',');
+        }
+        console.log('this is the config to insert', configToInsert);
```
bodinsamuel:
```suggestion
        if (!configToInsert.oauth_scopes && provider.default_scopes?.length) {
            configToInsert.oauth_scopes = provider.default_scopes.join(',');
        }
```

hassan254-prog: 🤦 🔧

bodinsamuel: you forgot the check on the `if`. Customer can pass their own scopes so we don't want to override that

hassan254-prog: Thank you Samuel for spotting this, I would have missed it. 🙏, fixed with new [commit](https://github.com/NangoHQ/nango/pull/3919/commits/9506c15c396dad73553af1e1c55b42b708fdcb6c)

bodinsamuel (approved):

PR #3920, opened 2025-04-17
feat: track nango.proxy() egress/ingress

TBonnin:
We now keep track of the total size of requests/responses data made in scripts to 3rd party provider via nango.proxy. The 2 stats, `proxy_egress_bytes` and `proxy_ingress_bytes`, are then sent back alongside the script output/error. This commit only logs the stats for now, next step will be exporting them to our billing platform

How to test:
 - run any script that is making a use of `nango.proxy()` or related method.
 - check the stats numbers in the logs

### DIFF
--- a/packages/jobs/lib/routes/tasks/putTask.ts
+++ b/packages/jobs/lib/routes/tasks/putTask.ts
@@ -89,10 +89,12 @@ const validate = validateRequest<PutTask>({
                 output: jsonSchema.default(null),
                 stats: z
                     .object({
-                        proxy_egress_bytes: z.number().default(0),
-                        proxy_ingress_bytes: z.number().default(0)
+                        proxy_success_egress_bytes: z.number().default(0),
+                        proxy_success_ingress_bytes: z.number().default(0),
+                        proxy_failure_egress_bytes: z.number().default(0),
+                        proxy_failure_ingress_bytes: z.number().default(0)
                     })
-                    .default({ proxy_egress_bytes: 0, proxy_ingress_bytes: 0 })
+                    .default({ proxy_success_egress_bytes: 0, proxy_success_ingress_bytes: 0, proxy_failure_egress_bytes: 0, proxy_failure_ingress_bytes: 0 })
             })
             .parse(data),
     parseParams: (data) => z.object({ taskId: z.string().uuid() }).strict().parse(data)
--- a/packages/jobs/lib/routes/tasks/putTask.ts
+++ b/packages/jobs/lib/routes/tasks/putTask.ts
@@ -5,6 +5,7 @@ import type { EndpointRequest, EndpointResponse, RouteHandler } from '@nangohq/u
 import { handleError, handleSuccess } from '../../execution/operations/handler.js';
 import type { JsonValue } from 'type-fest';
 import { operationIdRegex } from '@nangohq/logs';
+import { logger } from '../../logger.js';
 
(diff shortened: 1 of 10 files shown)


packages/jobs/lib/routes/tasks/putTask.ts
```diff
@@ -103,6 +110,9 @@ const handler = async (req: EndpointRequest<PutTask>, res: EndpointResponse<PutT
     } else {
         await handleSuccess({ taskId, nangoProps, output: output || null });
     }
+
+    logger.info(`Stats for task ${taskId}`, stats); // TODO: do something with stats
```
TBonnin, line 116: just logging for now. Once billing platform decision is made. We'll push the stats to it.

packages/runner/lib/sdk/sdk.ts
```diff
@@ -80,9 +91,17 @@ export class NangoActionRunner extends NangoActionBase {
...
             }
         });
-        const response = (await proxy.request()).unwrap();
 
-        return response;
+        const res = await proxy.request();
+
+        this.runnerStats.proxy_egress_bytes += sizeInBytes(proxy.config.data);
+        if (res.isOk()) {
+            this.runnerStats.proxy_ingress_bytes += Number(res.value.headers['content-length']) || sizeInBytes(res.value.data);
```
TBonnin: using `content-length` header if exist, otherwise fallback to estimated size

packages/jobs/lib/routes/tasks/putTask.ts
```diff
@@ -85,15 +86,21 @@ const validate = validateRequest<PutTask>({
...
                     })
                     .optional(),
-                output: jsonSchema.default(null)
+                output: jsonSchema.default(null),
+                stats: z
+                    .object({
+                        proxy_egress_bytes: z.number().default(0),
+                        proxy_ingress_bytes: z.number().default(0)
+                    })
+                    .default({ proxy_egress_bytes: 0, proxy_ingress_bytes: 0 })
```
bodinsamuel: should be optional during transition no?

packages/runner/lib/sdk/sdk.ts
```diff
@@ -80,9 +91,17 @@ export class NangoActionRunner extends NangoActionBase {
                 return connection;
             }
         });
-        const response = (await proxy.request()).unwrap();
 
-        return response;
+        const res = await proxy.request();
```
bodinsamuel, line 95: You are only getting the last request here, you should plug inside the method or add a callback

bodinsamuel (approved):
Few comments but otherwise good 👍🏻

packages/jobs/lib/routes/tasks/putTask.ts
TBonnin: that's what default does by setting a value if not defined

packages/runner/lib/sdk/sdk.ts
TBonnin: if by last request you mean last retry attempt, yes, that was on purpose. I don't think we should account for failed attempt. wdyt @bastienbeurier ?
After a discussing with Bastien, I am also gonna track stats for proxy success and failure separately, so they are tracked separately and failure might not be billed

TBonnin:
we want to track the `/proxy` endpoint as well. Moving to draft for now

PR #3921, opened 2025-04-17
fix(ui): disable cache when live

bodinsamuel:
## Changes

- Disable cache when live
When you have cache you have more problems. Tried to be smart, but it wasn't always working, especially when discarding the cursor it would at some point refresh the wrong page. Changed to disable cache when we are live (top of the page or in a running script). And use manual refetch for operations too


packages/webapp/src/pages/Logs/Operation/components/Logs.tsx
```diff
@@ -58,13 +58,7 @@ export const Logs: React.FC<{ operationId: string; isLive: boolean }> = ({ opera
...
                     // Lib is sometimes storing an old pageParam, when we re-open the Drawer it's incorrect in the past
                     // So we discard any cursor that are too old unless we search
-                    cursorBefore:
-                        pageParam &&
-                        'before' in pageParam &&
-                        pageParam.before &&
-                        (JSON.parse(atob(pageParam.before))[0] > Date.now() - 30_000 || debouncedSearch)
-                            ? pageParam.before
-                            : undefined
+                    cursorBefore: pageParam && 'before' in pageParam && pageParam.before ? pageParam.before : undefined
```
TBonnin: is that not the same as `cursorBefore: pageParam && 'before' in pageParam ? pageParam.before : undefined`

TBonnin (approved):

packages/webapp/src/pages/Logs/Operation/components/Logs.tsx
bodinsamuel: true

PR #3922, opened 2025-04-17
feat: event tracking v2

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-3073/posthog-cleanup

- Refactor analytics to be just internal event tracking
  - Removed call to DB in favor of explicit parameters when possible  
  - Remove most events that were not deemed useful
- Remove telemetry from self-hosted
Not useful right now


packages/server/lib/controllers/sync.controller.ts
```diff
@@ -434,7 +435,7 @@ class SyncController {
         try {
             const { account, environment } = res.locals;
 
-            const { schedule_id, command, nango_connection_id, sync_id, sync_name, sync_variant, provider, delete_records } = req.body;
```
TBonnin, line 437: can we stop passing those params in https://github.com/NangoHQ/nango/blob/7caf79bc921b195a3eddd1b66c7ac79b305bf4aa/packages/webapp/src/hooks/useSyncs.tsx#L20

packages/shared/lib/utils/eventTracking.ts
```diff
@@ -0,0 +1,79 @@
+import { PostHog } from 'posthog-node';
```
TBonnin: I am not a huge fan of the naming `eventTracking`. Soon we are going to have billing events and I am sure more sort of events in the future.

I am also not super clear of the value of PostHog. @bastienbeurier How is it being used and what does PostHog provides that we cannot have without it?

TBonnin (approved):
I have a question about why we use PostHog at all. But otherwise nice PR 👍

packages/shared/lib/utils/eventTracking.ts
bodinsamuel: > I am not a huge fan of the naming eventTracking. Soon we are going to have billing events and I am sure more sort of events in the future.

What would you propose? Analytics was also incorrect, maybe internalEventTracking? Billing events will be called billing events, so I'm not shocked.

> I am also not super clear of the value of PostHog. @bastienbeurier How is it being used and what does PostHog provides that we cannot have without it?

I'll let Bastien complete, imo those are product events; with it we have dashboards, insights, Slack alerts, trends, etc. that allow us to have a better sense of what's happening for some critical part of the product. We could not achieve the same with DD because we are missing dimensions and/or sampling, and simply not made for that.
By removing most events it's clear that it's less useful I agree

bastienbeurier: I think it's a common argument from engs that DD can replace product analytics. Initially, it's somewhat true. But rapidly product analytics are used for a different set of events and surfacing (e.g. with Slack notifications).

packages/server/lib/controllers/sync.controller.ts
bodinsamuel: good point, I'll change that when migrating the endpoint

packages/shared/lib/utils/eventTracking.ts
TBonnin: got it. `productEvent`, `productTracking` or something similar then.

TBonnin (approved):

packages/shared/lib/utils/eventTracking.ts
bodinsamuel: let's go for productTracking

PR #3924, opened 2025-04-17
fix(cli): correctly resolve file with fake extension

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-3090/cli-issue-when-detecting-lib

- CLI: correctly resolve file with fake extension
It was buggy when importing stuff like `schema.zod`. It was resolved to `schema.ts` instead of `schema.zod.ts`.


packages/cli/lib/services/compile.service.ts
```diff
@@ -170,26 +170,33 @@ function compileImportedFile({
 
     for (const importedFile of importedFiles) {
         const importedFilePath = path.resolve(path.dirname(filePath), importedFile);
-        const importedFilePathWithoutExtension = path.join(path.dirname(importedFilePath), path.basename(importedFilePath, path.extname(importedFilePath)));
```
TBonnin, line 173: I don't follow where the bug is. `path.basename('a/b/c/schema.zod.ts', path.extname('ts'))) `should give `a/b/c/schema.zod` no?

bodinsamuel: It's because extension-less imports are valid in Typescript, so when you autocomplete, it often imports this `schema.zod` and since we are less smart, we think .zod is the extension.

TBonnin: got it

TBonnin (approved):

PR #3926, opened 2025-04-17
fix(auth): split Tableau from connection

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2937/tableau-client

- Split Tableau from connection.service.ts
I have no way to test this. Maybe @hassan254-prog has some credentials?


TBonnin (approved):
👍

PR #3927, opened 2025-04-17
feat: billing with Lago

TBonnin:
## Changes

- Send usage events to Lago
I kept the original implementation from Thomas that allows multiple clients, which can be handy if we ever want to do some integration tests.

- Only send metrics for paying accounts
Lago does not reject the events, but we just we to start small

## 🧪 Tests

- Go to Lago
- Create a customer and assign a plan with the external id = account_id
- Modify .env
```
LAGO_API_KEY=SHH
FLAG_USAGE_ENABLED=true
```
- Trigger an action
- ??
- Profit


bodinsamuel:
Tested the query plan
Before

Screenshot 2025-04-22 at 17 45 13

After
Screenshot 2025-04-22 at 17 03 19

packages/server/lib/crons/usage.ts
```diff
@@ -74,11 +77,17 @@ const billing = {
...
                 if (res.isErr()) {
                     throw res.error;
                 }
-                logger.info(`Exporting ${res.value.length} billable connections`); //TODO: export to billing platform
+
+                const events = res.value.map<BillingMetric>(({ accountId, count }) => {
+                    return { type: 'billable_connections', value: count, properties: { accountId, timestamp: now } };
+                });
+
+                await usageBilling.sendAll(events);
```
kaposke, line 90: do we want to filter our the free plan here as well?

kaposke (commented):
Can I get an invite to lago?

packages/billing/lib/billing.ts
```diff
@@ -0,0 +1,54 @@
...
+
+        try {
+            await this.client.ingest(events);
+            return Ok(undefined);
+        } catch (err: unknown) {
+            const e = new Error(`Failed to send billing event`, { cause: err });
+            report(e);
+            return Err(e);
+        }
+    }
```
TBonnin, line 75: I would retry certain errors here like network ones for instance? wdyt?

packages/server/lib/controllers/sync.controller.ts
```diff
@@ -228,11 +229,16 @@ class SyncController {
                 logCtx
             });
 
+            // Note: all executed actions are billed
```
TBonnin: this comment is not up-to-date anymore. Only successful actions are billed

packages/shared/lib/services/connection.service.ts
```diff
@@ -1585,15 +1592,11 @@ class ConnectionService {
...
-        // Note:
-        // a billable connection is a connection that is not deleted and has not been deleted during the month
-        // connections are pro-rated based on the number of seconds they were existing in the month
-
         const targetDate = new Date(Date.UTC(referenceDate.getUTCFullYear(), referenceDate.getUTCMonth(), referenceDate.getUTCDate(), 0, 0, 0, 0));
         const year = referenceDate.getUTCFullYear();
         const month = referenceDate.getUTCMonth() + 1; // js months are 0-based
 
-        const res = await db.readOnly
+        const q = db.readOnly
```
TBonnin: we can await the query directly. I think it is a leftover of some early-on debugging

TBonnin (commented):
made a few tiny comments. lgtm otherwise

TBonnin:
> made a few tiny comments. lgtm otherwise

I cannot approved my own PR @bodinsamuel so feel free to merge once you have addressed the comments

packages/shared/lib/services/connection.service.ts
bodinsamuel: I don't really mind having it, sometimes its handy but yeah let's change

packages/billing/lib/billing.ts
bodinsamuel: Added

bodinsamuel (approved):

PR #3928, opened 2025-04-17
feat: delete old environments (and refactor delete old data)

kaposke:
### Description
This adds environments to our `deleteOldData` cron job. It goes very in depth on deleting anything under an environment. To achieve so, some existing parts of `deleteOldEnvironments` was refactored / modified. It should go down the whole graph of dependencies under environments. Here's how the dependency graph looks, meaning that deleting a node means also deleting all it's child nodes:
```
_nango_environments
├── _nango_external_webhooks
├── _nango_slack_notifications
├── _nango_environment_variables
├── end_users
│   └── _nango_connections
│       ├── connect_sessions
│       │   └── _nango_oauth_sessions
│       └── _nango_syncs
│           ├── _nango_sync_jobs
│           └── _nango_active_logs
└── _nango_configs
    └── on_event_scripts
        ├── _nango_connections
        │   ├── connect_sessions
        │   │   └── _nango_oauth_sessions
        │   └── _nango_syncs
        │       ├── _nango_sync_jobs
        │       └── _nango_active_logs
        └── _nango_sync_configs
            ├── _nango_syncs
            │   ├── _nango_sync_jobs
            │   └── _nango_active_logs
            └── _nango_sync_endpoints
```
> Hope this is all correct, looking at this is kind of mind-bending with all the crossed dependencies.

I'll add some comments throughout with some more details

### Testing
I have not yet tested this anywhere. It's very unlikely that I got it all right out of the gate. I'm taking suggestions on how to test it.


packages/server/lib/crons/utils/batchDelete.ts
```diff
@@ -0,0 +1,31 @@
...
+        if (deleted) {
+            logger.info(`Deleted ${deleted} ${name}`);
+        }
+        if (deleted < limit) {
+            break;
+        }
+        if (Date.now() > deadline.getTime()) {
+            logger.info(`Time limit reached, stopping`);
+            // This will propagate up nested batchDeletes and block any subsequent hardDeletes
+            throw new Error('time_limit_reached');
```
kaposke: This is fundamental. If some nested `batchDelete` just returns, it means the hardDeletes happen without all it's dependencies being deleted. This error propagates up and avoids it.

packages/server/lib/crons/utils/deleteSyncData.ts
```diff
@@ -0,0 +1,59 @@
...
+    });
+
+    await batchDelete({
+        ...opts,
+        name: 'active_logs < sync',
+        deleteFn: async () => {
+            const activeLogs = await db.knex.from<ActiveLog>('_nango_active_logs').where({ sync_id: sync.id }).limit(limit).delete();
+            return activeLogs;
+        }
+    });
```
kaposke: Lmk if any of these shouldn't be deleted, but they do have FKs to syncs

kaposke: Oh I just realized this one has `ON DELETE cascade`. I'll double check FKs when I'm back to remove anything unnecessary.

bodinsamuel: It should be deleted and the FK still have the same problems has any other FK so I think the code is correct like this

packages/server/lib/crons/utils/batchDelete.ts
bodinsamuel: Thanks

packages/server/lib/crons/utils/deleteConnectSessionData.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    await batchDelete({
+        ...opts,
+        name: 'oauth_sessions < connect_session',
+        deleteFn: async () => {
+            const oauthSessionsDeletedCount = await db.knex
+                .from<OAuthSession>('_nango_oauth_sessions')
+                .where({ connectSessionId: connectSession.id })
+                .limit(opts.limit)
+                .delete();
```
bodinsamuel: Delete limit doesn't work unfortunately, good news you can remove this batchDelete, it's already taken care of with `oauthSessionService.deleteExpiredSessions`

packages/server/lib/crons/utils/deleteConnectionData.ts
```diff
@@ -0,0 +1,47 @@
...
+    await batchDelete({
+        ...opts,
+        name: 'connect_sessions < connection',
+        deleteFn: async () => {
+            const connectSessions = await db.knex.from<DBConnectSession>('connect_sessions').where({ connection_id: connection.id }).limit(opts.limit);
+
+            for (const connectSession of connectSessions) {
+                await deleteConnectSessionData(connectSession, opts);
+            }
+
```
bodinsamuel: You can change to a batch delete directly instead of selecting

packages/server/lib/crons/utils/deleteEnvironmentData.ts
```diff
@@ -0,0 +1,106 @@
...
+    await batchDelete({
+        ...opts,
+        name: 'end_users < environment',
+        deleteFn: async () => {
+            const endUsers = await db.knex.from<DBEndUser>('end_users').where({ environment_id: environment.id }).limit(opts.limit);
+
+            for (const endUser of endUsers) {
+                await deleteEndUserData(endUser, opts);
+            }
+
```
bodinsamuel: You can also directly batch delete since connections will be deleted in an other loop

packages/server/lib/crons/utils/deleteEndUserData.ts
```diff
@@ -0,0 +1,26 @@
...
+            const connections = await db.knex.from<DBConnection>('_nango_connections').where({ end_user_id: endUser.id }).limit(opts.limit);
+
+            for (const connection of connections) {
+                await deleteConnectionData(connection, opts);
+            }
+
+            return connections.length;
+        }
+    });
+}
```
bodinsamuel: This whole file can be removed (see deleteEnv)

packages/server/lib/crons/utils/deleteProviderConfigData.ts
```diff
@@ -0,0 +1,50 @@
...
+
+    await batchDelete({
+        ...opts,
+        name: 'on_event_scripts < providerConfig',
+        deleteFn: async () => {
+            const onEventScriptsDeletedCount = await db.knex
+                .from<DBOnEventScript>('on_event_scripts')
+                .where({ config_id: providerConfig.id! })
+                .limit(opts.limit)
+                .delete();
```
bodinsamuel: Delete limit doesn't work

packages/server/lib/crons/utils/deleteEnvironmentData.ts
```diff
@@ -0,0 +1,106 @@
...
+        deleteFn: async () => {
+            const externalWebhooksDeletedCount = await db.knex
+                .from<DBExternalWebhook>('_nango_external_webhooks')
+                .where({ environment_id: environment.id })
+                .limit(opts.limit)
+                .delete();
+
+            return externalWebhooksDeletedCount;
+        }
+    });
```
bodinsamuel: there is only row per env for this table

packages/server/lib/crons/utils/deleteEnvironmentData.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    await batchDelete({
+        ...opts,
+        name: 'slack_notifications < environment',
+        deleteFn: async () => {
+            const slackNotifications = await db.knex
+                .from<DBSlackNotification>('_nango_slack_notifications')
+                .where({ environment_id: environment.id })
+                .limit(opts.limit)
+                .delete();
```
bodinsamuel: Delete limit doesn't work

packages/server/lib/crons/utils/deleteEnvironmentData.ts
```diff
@@ -0,0 +1,106 @@
...
+    logger.info('Deleting all environment variables in environment...', environment.id, environment.name);
+
+    await batchDelete({
+        ...opts,
+        name: 'environment_variables < environment',
+        deleteFn: async () => {
+            const environmentVariablesDeletedCount = await db.knex
+                .from<DBEnvironmentVariable>('_nango_environment_variables')
+                .where({ environment_id: environment.id })
+                .limit(opts.limit)
```
bodinsamuel: delete limit doesn't work, but since there will be limited number of rows you can skip the batch delete and just delete all in one query

packages/server/lib/crons/utils/deleteProviderConfigData.ts
```diff
@@ -0,0 +1,50 @@
...
+import { configService } from '@nangohq/shared';
+
+import { batchDelete } from './batchDelete.js';
+import { deleteSyncConfigData } from './deleteSyncConfigData.js';
+
+import type { BatchDeleteSharedOptions } from './batchDelete.js';
+import type { Config } from '@nangohq/shared';
+import type { DBOnEventScript, DBSyncConfig } from '@nangohq/types';
+
+export async function deleteProviderConfigData(providerConfig: Config, opts: BatchDeleteSharedOptions) {
```
bodinsamuel: you should use IntegrationConfig (just so you know types imported from shared are deprecated)

packages/server/lib/crons/deleteOldData.ts
```diff
@@ -83,76 +77,55 @@ export async function exec(): Promise<void> {
...
-
-                    // Delete connections
-                    await batchDelete({
-                        name: 'connections < integration',
-                        deadline,
-                        deleteFn: async () => await connectionService.hardDeleteByIntegration({ limit, integrationId: integration.id! })
-                    });
-
-                    await configService.hardDelete(integration.id!);
+                    await deleteProviderConfigData(integration as Config, opts);
```
bodinsamuel:
```suggestion
                    await deleteProviderConfigData(integration, opts);
```

bodinsamuel (commented):
Not tested yet, looks good overall, just a few comments here and there 🙏🏻

packages/server/lib/crons/utils/batchDelete.ts
```diff
@@ -0,0 +1,31 @@
+import { setTimeout } from 'node:timers/promises';
```
TBonnin: I'll rename the `crons/utils` subfolder to `crons/delete`.

packages/server/lib/crons/utils/deleteConnectSessionData.ts
TBonnin: more info about DELETE not having a LIMIT clause in pg: https://www.postgresql.org/docs/current/sql-delete.html
> ...
> While there is no LIMIT clause for DELETE, it is possible to get a similar effect using the same method described in the documentation of UPDATE:
```
WITH delete_batch AS (
  SELECT l.ctid FROM user_logs AS l
    WHERE l.status = 'archived'
    ORDER BY l.creation_date
    FOR UPDATE
    LIMIT 10000
)
DELETE FROM user_logs AS dl
  USING delete_batch AS del
  WHERE dl.ctid = del.ctid;
```

packages/server/lib/crons/utils/deleteProviderConfigData.ts
kaposke: Haha that's good to know

packages/server/lib/crons/utils/deleteEnvironmentData.ts
kaposke: Where else are connections deleted? You mean the soft-deleted ones in the base cron?

There's a conflict between having our delete function do what you'd think they would (delete all it's dependencies), and things being deleted elsewhere, meaning we trusts other parts of the cron job to clean things properly. I was going for redundancy in most cases, but I'll remove redundancy where performance could be significantly affected (removed for oauth sessions and connect sessions, for instance).

packages/server/lib/crons/utils/deleteProviderConfigData.ts
kaposke: Removed limit entirely since `on_event_scripts` shouldn't represent significant numbers.

packages/server/lib/crons/delete/deleteProviderConfigData.ts
```diff
@@ -0,0 +1,45 @@
...
+        ...opts,
+        name: 'on_event_scripts < providerConfig',
+        deleteFn: async () => {
+            const onEventScriptsDeletedCount = await db.knex.from<DBOnEventScript>('on_event_scripts').where({ config_id: providerConfig.id! }).delete();
+
+            return onEventScriptsDeletedCount;
+        }
+    });
+
+    await configService.hardDelete(providerConfig.id);
```
bodinsamuel, line 59: You forgot to delete connections data before delete the provider

packages/server/lib/crons/utils/deleteEnvironmentData.ts
bodinsamuel: Sorry I was not very clear. end_users table is not an entry point to connections.
They should be considered independent, because legacy customers are not using this system and it's optional for new customers.

So you can batch delete them without selecting (e.g: `DELETE WHERE IN`)
And delete connections independently

bodinsamuel (commented):
Sorry this back and forth is a bit annoying, maybe we should have made it testable.
It's currently not working, for providers and env, it's missing the deletion of connections

packages/server/lib/crons/delete/deleteProviderConfigData.ts
kaposke: moved connection deletion to this function, it makes more sense now. Thanks

packages/server/lib/crons/utils/deleteEnvironmentData.ts
kaposke: Moved to isolated delete under `deleteEnvironment`.

packages/server/lib/crons/utils/deleteConnectSessionData.ts
kaposke: That whereIn syntax makes sense now, thanks!

packages/server/lib/crons/delete/deleteProviderConfigData.ts
```diff
@@ -15,6 +16,20 @@ export async function deleteProviderConfigData(providerConfig: IntegrationConfig
...
+        deleteFn: async () => {
+            const connections = await db.knex.from<DBConnection>('_nango_connections').where({ config_id: providerConfig.id! }).limit(opts.limit);
+
+            for (const connection of connections) {
+                await deleteConnectionData(connection, opts);
+            }
+
+            return connections.length;
+        }
+    });
```
bodinsamuel: Can you just delete connections after the next block, otherwise the FK will still try to delete N syncs

bodinsamuel (approved):
Working great 🚀

PR #3929, opened 2025-04-18
feat(persist): log unchanged records

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2916/log-unchanged-records

- Log unchanged records 
In #3496 I tried to do that, but also changing the webhook body, which created some debate and potential breaking changes. Took a simpler approach this time by just modifying the log. We need this for transparency but also to be able to really find records by ID. I expect this change to have a **bigger impact on log storage** but we'll see

(*don't mind the singular it's fixed*)
Screenshot 2025-04-18 at 11 31 55
Inside the log msg
Screenshot 2025-04-18 at 11 32 08


packages/types/lib/logs/messages.ts
```diff
@@ -126,6 +126,8 @@ export interface MessageRow {
               addedKeys: string[];
               updated: number;
               updatedKeys: string[];
+              unchanged: number;
+              unchangedKeys: string[];
```
TBonnin, line 130: my guess is that a lot of records are actually unchanged so, as you mentioned, we might end up logging way more data. should we start by only adding the unchanged count to the logs, assess the impact once we have the numbers and see what would be the impact before we start adding the unchangedKeys to the logs?

TBonnin (approved):
one question about the logs but lgtm otherwise

PR #3934, opened 2025-04-21
feat(webhooks): add in webhook support for Xero within Nango

viictoo:
Add webhook support for Xero

Requirements:
## Handle Intent to Receive Validation:

Xero sends a series of HTTPS POST requests with an empty events payload to validate a new endpoint.
- Respond with 2xx for the correctly signed payloads.
- Respond with 401 Unauthorized for the incorrectly signed payloads.

## Verify the `x-xero-signature` header using HMAC-SHA256 with the provided webhook signing key.
Compare the computed hash (base64-encoded) with the signature in the header.
Respond with:
2xx for valid signatures.
401 Unauthorized for invalid signatures.

Changes (Summary):
- Added webhook routing script for  Xero provider.
- Updated webhook response handling to allow returning of appropriate status codes.
- eslint[import/order](https://github.com/import-js/eslint-plugin-import/blob/v2.31.0/docs/rules/order.md) fixes

[ext-602](https://linear.app/nango/issue/EXT-602/add-in-webhook-support-for-xero-within-nango)

Testing instructions 
### 1. Webhook Registration
1. Get webhook secret from Xero developer portal
2. Configure in Nango integration settings
3. Register Nango webhook URL in Xero

### 2. Validation Testing
- Verify "Intent to receive" validation:
  - Should return 200 for valid signatures
  - Should return 401 for invalid signatures
- Test webhook signature validation:
  - Valid signatures → 200 response
  - Invalid/missing signatures → 401 response

### 3. Event Processing
- Trigger events in Xero


packages/server/lib/controllers/webhook/environmentUuid/postWebhook.ts
```diff
@@ -72,7 +74,14 @@ export const postWebhook = asyncWrapper<PostPublicWebhook>(async (req, res) => {
                 res.status(200).send();
                 return;
             }
-            res.status(200).send(response);
+
+            const typedResponse = response as { statusCode?: number; acknowledgementResponse?: unknown };
```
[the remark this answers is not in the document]
TBonnin: I would take the opportunity to refactor `routeWebhook` return type to not be unknown so we don't have to force any type casting.

packages/server/lib/webhook/types.ts
```diff
@@ -16,6 +16,7 @@ export type WebhookResponse =
           acknowledgementResponse?: unknown;
           parsedBody?: unknown;
           connectionIds?: string[] | undefined;
+          statusCode?: number;
```
TBonnin: I know you didn't start it but `WebhookResponse` should be an union type that clearly define the possible responses instead of having all fields to be optional and allow responses that should not be representable: 
For example:
```
type WebhookResponsOnly = { response: string | Record<string, any> | null; statusCode: number }
export type WebhookResponse =
| WebhookResponsOnly
| WebhookResponsOnly & { connectionIds: string[];  }
| WebhookResponsOnly & { toForward: unknown; connectionIds: string[];}
```
forcing routing logic for all providers to define the response and status code of the response explicitely and potentially returned other data like connection ids or parsed body

bodinsamuel: Actually, I think we should just return Result, + in that case, I'm not even sure the status code matters that much, we just need to reject.

bodinsamuel (commented):
Looks good 👍🏻 
I think returning statusCode to convey "error" is a mistake that was committed a lot in this code base. Refactoring everything can be too big of a task but I would prefer to:
- Return an Error
- Throw an Error
- Use `Result`

Error should be typed using NangoInternalError, e.g: WebhookRoutingError that contains known code

packages/server/lib/webhook/types.ts
viictoo: For Xero, we  have to return the status code in in "Intent to receive"  checks https://developer.xero.com/documentation/guides/webhooks/configuring-your-server/#intent-to-receive

packages/server/lib/webhook/types.ts
```diff
@@ -31,3 +34,10 @@ export interface AirtableWebhookReference {
     };
     timestamp: string;
 }
+
+export interface RouteWebhookResponse {
+    response: string | Record<string, any> | null;
+    statusCode: number;
+    connectionIds?: string[];
+    toForward?: unknown;
+}
```
TBonnin: do we need this type? Can we use `WebhookResponse` from above?

packages/server/lib/controllers/webhook/environmentUuid/postWebhook.ts
```diff
@@ -75,12 +75,10 @@ export const postWebhook = asyncWrapper<PostPublicWebhook>(async (req, res) => {
...
 
-            const typedResponse = response as { statusCode?: number; acknowledgementResponse?: unknown };
-
-            if (typedResponse.statusCode) {
-                res.status(typedResponse.statusCode).send(typedResponse.acknowledgementResponse || {});
+            if (response.statusCode) {
+                res.status(response.statusCode).send(response.response || {});
             } else {
-                res.status(200).send(typedResponse.acknowledgementResponse || {});
+                res.status(200).send(response.response || {});
```
TBonnin: isn't statusCode and response content always defined?

packages/server/lib/webhook/types.ts
```diff
@@ -11,14 +11,17 @@ export type WebhookHandler<T = any> = (
     logContextGetter: LogContextGetter
 ) => Promise<WebhookResponse>;
 
+// Base response type that all webhook responses must include
+export interface WebhookResponseOnly {
+    response: string | Record<string, any> | null;
```
TBonnin: I am tempted to call it `content` instead of response. wdyt?

packages/server/lib/webhook/webhook.manager.ts
```diff
@@ -32,20 +32,20 @@ export async function routeWebhook({
     body: any;
     rawBody: string;
     logContextGetter: LogContextGetter;
-}): Promise<unknown> {
+}): Promise<RouteWebhookResponse | null> {
```
TBonnin: would only return `WebhookResponse`. if needed adding a case to the union type that is clearly describing this scenario

packages/server/lib/webhook/webhook.manager.ts
```diff
@@ -57,8 +57,8 @@ export async function routeWebhook({
         return null;
     });
 
-    const webhookBodyToForward = res?.parsedBody || body;
-    const connectionIds = res?.connectionIds || [];
+    const webhookBodyToForward = res && 'toForward' in res ? res.toForward : body;
+    const connectionIds = res && 'connectionIds' in res ? (res.connectionIds as string[]) : [];
```
TBonnin: is the as string[] casting necessary?

packages/server/lib/webhook/types.ts
TBonnin: @bodinsamuel the challenge is there is no clear cut of what's a ok/err result. Each provider has its own rules and defining exactly what should be returned would be more explicit imho

packages/server/lib/webhook/types.ts
viictoo: Was not needed, made the update to use `WebhookResponse` type

packages/server/lib/controllers/webhook/environmentUuid/postWebhook.ts
viictoo: correct. Updated

packages/server/lib/webhook/types.ts
viictoo: Yes. I agree with that as response is already used

packages/server/lib/webhook/webhook.manager.ts
viictoo: Updated as well

packages/server/lib/webhook/webhook.manager.ts
viictoo: Improved the types to handle this more appropriately without  the assertion

packages/server/lib/webhook/types.ts
```diff
@@ -11,13 +11,25 @@ export type WebhookHandler<T = any> = (
...
+
+export interface WebhookResponseWithConnectionIds extends WebhookResponseOnly {
+    connectionIds: string[];
+}
+
+export interface WebhookResponseWithForward extends WebhookResponseWithConnectionIds {
+    toForward: unknown;
+}
+
+export interface WebhookResponseNoAction {
```
TBonnin: should this be called `WebhookResponseNoContent`?

packages/shared/lib/utils/error.ts
```diff
@@ -489,6 +491,21 @@ export class NangoError extends NangoInternalError {
...
+            case 'missing_signature':
+                this.status = 401;
+                this.message = 'Missing webhook signature';
+                break;
+
+            case 'invalid_payload':
+                this.status = 400;
+                this.message = 'Invalid webhook payload';
+                break;
+
```
TBonnin, line 506: should those error be prefixed with `webhook_...`?

packages/server/lib/webhook/linear-webhook-routing.ts
```diff
@@ -27,22 +29,27 @@ const route: WebhookHandler = async (nango, integration, headers, body, rawBody,
     const signature = headers['linear-signature'];
     if (!signature) {
         logger.error('missing signature', { configId: integration.id });
-        return;
+        throw new WebhookRoutingError('missing_signature');
```
TBonnin: I would prefer if the webhook handlers return a Result instead of throwing. wdyt?

packages/server/lib/webhook/types.ts
viictoo: Yes. Updated, that would be more accurate.

packages/shared/lib/utils/error.ts
viictoo: Added prefix and extended to capture all webhook errors

packages/server/lib/webhook/linear-webhook-routing.ts
viictoo: Updated. Used the Result pattern already defined in the codebase to return error (if error) instead of throwing errors in the webhook routing handlers.

packages/server/lib/webhook/webhook.manager.ts
```diff
@@ -29,48 +33,83 @@ export async function routeWebhook({
...
+    }
 
-    const webhookSettings = await externalWebhookService.get(environment.id);
+    const res = result.value;
+    if (!res) {
+        return {
+            content: null,
+            statusCode: 204
+        };
+    }
```
TBonnin: can `res` ever be undefined or null here?

viictoo: removed as it is redundant at this point with all the checks and new typescript types

TBonnin (approved):

bodinsamuel (approved):
Looks good, thanks 🙏🏻

PR #3935, opened 2025-04-22
feat(api): enforce max connections

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-3124/enfore-max-1000-connections-for-free-users
Fixes https://linear.app/nango/issue/NAN-3022/connect-ui-handle-max-connection

- Enforce max connections when creating new one
Changed `getSyncConfigsWithConnections` that actually fetched all connections (sic) with a query that only count total connection and connection with scripts
```sql
SELECT
	"_nango_configs"."unique_key" AS "providerConfigKey",
	COUNT(
		DISTINCT CASE
			WHEN _nango_sync_configs.active = TRUE
			AND _nango_sync_configs.deleted = FALSE THEN _nango_connections.id
		END
	) AS connectionsWithScripts,
	COUNT(DISTINCT _nango_connections.id) AS total
FROM
	"_nango_connections"
	INNER JOIN "_nango_configs" ON "_nango_connections"."provider_config_key" = "_nango_configs"."unique_key"
	LEFT JOIN "_nango_sync_configs" ON "_nango_sync_configs"."nango_config_id" = "_nango_configs"."id"
WHERE
	"_nango_connections"."environment_id" = 18
	AND "_nango_connections"."deleted" = FALSE
	AND "_nango_configs"."deleted" = FALSE
GROUP BY
	ROLLUP ("_nango_configs"."unique_key");
```

- Handle this error in the Connect UI

## Tests

- Change your plan `connections_max = 1`
- Launch web
- Try to add a connection


packages/shared/lib/services/sync/config/config.service.ts
```diff
@@ -487,43 +487,33 @@ export async function getSyncConfigsWithConnectionsByEnvironmentId(environment_i
...
-        });
-
-    return result;
+        .from('_nango_connections')
+        .join('_nango_configs', '_nango_connections.provider_config_key', '_nango_configs.unique_key')
+        .leftJoin('_nango_sync_configs', `_nango_sync_configs.nango_config_id`, '_nango_configs.id')
+        .where('_nango_connections.environment_id', environmentId)
+        .andWhere('_nango_connections.deleted', false)
+        .andWhere('_nango_configs.deleted', false)
+        .groupByRaw('ROLLUP("_nango_configs"."unique_key")');
```
TBonnin, line 509: I am not sure what the ROLLUP with a single column does? Can you explain?

TBonnin (approved):

packages/shared/lib/services/sync/config/config.service.ts
bodinsamuel: Rollup will summarize the column. The query count per unique_key, and then the last row `unique_key` will be null and the count will be the total. I could do it by hand tbh, it's just one less loop.
Screenshot 2025-04-23 at 17 11 10

PR #3938, opened 2025-04-22
feat(integrations): add support for workday oauth

hassan254-prog:

- Add support for workday oauth


packages/shared/lib/clients/provider.client.ts
```diff
@@ -24,6 +25,7 @@ class ProviderClient {
             case 'tiktok-accounts':
             case 'stripe-app':
             case 'stripe-app-sandbox':
+            case 'workday-oauth':
```
hassan254-prog, line 28: Workday does not allow `client_credentials` to be included in the request body; they prefer it to be passed only in the headers as a base64 encoded string. To accommodate this, we needed to use `shouldUseProviderClient`.

packages/providers/providers.yaml
```diff
@@ -10022,6 +10022,53 @@ workday:
...
+    token_params:
+        grant_type: authorization_code
+    refresh_params:
+        grant_type: refresh_token
+    proxy:
+        base_url: https://${connectionConfig.tokenHostname}/ccx/api
+    docs: https://docs.nango.dev/integrations/all/workday-oauth
+    docs_connect: https://docs.nango.dev/integrations/all/workday-oauth/connect
+    connection_config:
+        authorizationHostname:
```
bodinsamuel: can we use domain please? It's more accurate and user friendly

packages/shared/lib/utils/utils.ts
```diff
@@ -367,3 +367,10 @@ export function getConnectionMetadataFromTokenResponse(params: any, provider: Pr
```
bodinsamuel, line 365: Not related to your pr but this whole file needs a cleanup

packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -985,8 +986,16 @@ class OAuthController {
 
             const tokenUrl = typeof provider.token_url === 'string' ? provider.token_url : (provider.token_url?.['OAUTH2'] as string);
 
+            const interpolatedTokenUrl = makeUrl(tokenUrl, session.connectionConfig, provider.token_url_skip_encode);
+
             if (providerClientManager.shouldUseProviderClient(session.provider)) {
-                rawCredentials = await providerClientManager.getToken(config, tokenUrl, authorizationCode, session.callbackUrl, session.codeVerifier);
+                rawCredentials = await providerClientManager.getToken(
+                    config,
+                    interpolatedTokenUrl.origin + interpolatedTokenUrl.pathname,
```
bodinsamuel: interpolatedTokenUrl.href is not working?

bodinsamuel (approved):
Looks okay 👌🏻

packages/shared/lib/utils/utils.ts
hassan254-prog: I agree 💯

packages/providers/providers.yaml
hassan254-prog: They are different, which is why we have both `authorizationDomain` and `tokenDomain`. I've updated this in the latest commit.

PR #3941, opened 2025-04-22
feat(integrations): add support for docuware

hassan254-prog:

- Add support for Docuware

    

---

## Summary by mrge
Added DocuWare integration with support for username and password authentication, including setup guides, documentation, and UI assets.

- **Dependencies**
  - Updated provider schema and connection logic to support multi-step authentication flows and DocuWare-specific config.


hassan254-prog:
Another advanced use case for the `TWO_STEP` flow 🥳. In this scenario, Nango will make three separate requests:

1. `/IdentityServiceInfo` – This endpoint returns the **IdentityServiceUrl**.
2. `/.well-known/openid-configuration` – Nango will append this path to the **IdentityServiceUrl** obtained in step 1 to retrieve the `token_endpoint`.
3. `token_endpoint` – Using the `token_endpoint` from step 2, Nango will request the **access_token **and **expire_at** values using the users credentials (Username and Password).

packages/providers/providers.yaml
```diff
@@ -2886,6 +2886,60 @@ docusign-sandbox:
...
+        token_expiration: expires_in
+        token_expiration_strategy: expireIn
+    docs: https://docs.nango.dev/integrations/all/docuware
+    docs_connect: https://docs.nango.dev/integrations/all/docuware/connect
+    connection_config:
+        hostName:
+            type: string
+            title: Host Name
+            description: The host name to your DocuWare account
+            pattern: ^https?:\/\/.*
```
bodinsamuel: can you change to not include https:// and add prefix instead?

packages/providers/providers.yaml
```diff
@@ -2886,6 +2886,60 @@ docusign-sandbox:
...
+    proxy:
+        base_url: ${connectionConfig.hostName}/DocuWare/Platform
+    token_response:
+        token: access_token
+        token_expiration: expires_in
+        token_expiration_strategy: expireIn
+    docs: https://docs.nango.dev/integrations/all/docuware
+    docs_connect: https://docs.nango.dev/integrations/all/docuware/connect
+    connection_config:
+        hostName:
```
bodinsamuel: name it `domain`

packages/providers/providers.yaml
```diff
@@ -7632,7 +7686,7 @@ redtail-crm-sandbox:
         - crm
     auth_mode: TWO_STEP
     token_url: https://review.crm.redtailtechnology.com/api/public/v1/authentication
-    token_request_method: GET
+    request_method: GET
```
bodinsamuel: I'm not sure about this rename, it's too vague and can apply to many different things. Why did you rename?

packages/providers/providers.yaml
hassan254-prog: updated in latest commit 👍 .

packages/providers/providers.yaml
hassan254-prog: mmmh, DocuWare can be [self hosted](https://start.docuware.com/cloud-on-premises-hybrid-deployment), which means we can have domains with `http://`, should we still change to rely on the `prefix` instead?

packages/providers/providers.yaml
hassan254-prog: Within the `additional_steps` , we can have requests that are not necessarily meant to request a token, which naming convention would you prefer we use?

bodinsamuel: it should be specified on each step imo with just method

hassan254-prog: `TWO_STEP` is dynamic in nature, and it might be hard to know the specific request at each step. I think we can just revert it to `token_request_method` for both the `additional_steps` and the main request. What do you think?

packages/providers/providers.yaml
bodinsamuel: I don't think http should be allowed anyway, let's change and see if it blocks someone

packages/providers/providers.yaml
bodinsamuel: yes let's revert, it's clearer like this

packages/providers/providers.yaml
hassan254-prog: thanks, updated with the latest [commit](https://github.com/NangoHQ/nango/pull/3941/commits/95320df12c58ff5b0e26cf49fd00c9fadc6c0fff) 🙏

packages/providers/providers.yaml
hassan254-prog: thanks, updated 👍

bodinsamuel (approved):

PR #3945, opened 2025-04-23
fix(auth): split Apple App Store from connection

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2941/app-store-client

- Split Apple App Store from connection.service.ts
No logic change. Just removing some unrelated code from the model
I'm sure this one could be rebranded JWT but there are a few users in prod. We can assess actual refactoring later


packages/server/lib/controllers/auth/postAppStore.ts
```diff
@@ -139,38 +147,28 @@ export const postPublicAppStoreAuthorization = asyncWrapper<PostPublicAppStoreAu
...
-                    auth_mode: 'APP_STORE',
-                    error: {
-                        type: 'credential_fetch_failure',
-                        description: `Error during App store credentials auth: ${error?.message}`
-                    },
-                    operation: 'unknown'
-                },
-                account,
-                config
-            );
```
bodinsamuel, line 164: Note I intentionally removed this because we are not doing that for all others. We probably should however, there are so many early returns we need to find a better solution

TBonnin (approved):
> I'm sure this one could be rebranded JWT
💯

khaliqgant:
>   I'm sure this one could be rebranded JWT but there are a few users in prod. We can assess actual refactoring later

@bodinsamuel 
Are there actual connections against this provider? I remember checking and there were just empty provide configs with nothing attached to it.

bodinsamuel:
> Are there actual connections against this provider? I remember checking and there were just empty provide configs with nothing attached to it.

Most probably yes. I think tableau, app store, bill are mostly empty and could be rewritten but I think it will be easier after this clean up. We also need to think of the self-hosted user unfortunately

PR #3946, opened 2025-04-23
fix(auth): split Github App from connection 

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-2935/app-client

- Split Github App from connection.service.ts
No logic change. Just removing some unrelated code from the model
This one is a bit all over the place, unfortunately

    

---

## Summary by mrge
Moved Github App authentication logic out of connection.service.ts into a new githubApp.ts file for better separation. No logic changes were made.


packages/server/lib/controllers/appAuth.controller.ts
```diff
@@ -125,36 +125,19 @@ class AppAuthController {
...
-
-            if (!success || !credentials) {
-                void logCtx.error('Error during app token retrieval call', { error });
+            const credentialsRes = await githubAppClient.createCredentials({ provider: provider as ProviderGithubApp, integration: config, connectionConfig });
+            if (credentialsRes.isErr()) {
+                report(credentialsRes.error);
+                void logCtx.error('Error during Github App credentials creation', { error: credentialsRes.error });
                 await logCtx.failed();
-
-                void connectionCreationFailedHook(
```
bodinsamuel, line 134: Same as apple store refactor, it's probably good that we have that but it's inconsistent across the endpoints and even in this endpoint

TBonnin (approved):

PR #3948, opened 2025-04-23
fix(deploy): cli should not pass empty string for actions' run

bodinsamuel:
## Changes

- Accept empty strings in runs while the fix is being deployed
- Fix the CLI
    

---

## Summary by mrge
The CLI no longer sends an empty string for the "runs" field during deploy actions, preventing validation errors with older CLI versions.

- **Bug Fixes**
  - Accepts both empty strings and null for "runs" to support smooth deployment while the fix rolls out.


packages/cli/lib/services/deploy.service.ts
```diff
@@ -578,7 +580,7 @@ class DeployService {
                         providerConfigKey,
                         models: action.output || [],
                         version: version || action.version,
-                        runs: '',
+                        runs: null,
```
bodinsamuel, line 583: Probably a leftover from before I fixed the types

kaposke (approved):

PR #3949, opened 2025-04-23
feat(api): enforce max connections (again) 

bodinsamuel:
## Changes

See #3935 

- Fixed the SQL query
We were only filtering nango_configs.unique_key, which is, as the name doesn't suggest, not very unique. So it was actually joining random configs, can't believe I got caught by this one again after all this time.
The query plan is sub ms again

```sql
GroupAggregate  (cost=523.11..525.88 rows=139 width=28) (actual time=0.452..0.459 rows=2 loops=1)
  Group Key: _nango_configs.unique_key
  Group Key: ()
  ->  Sort  (cost=523.11..523.46 rows=138 width=18) (actual time=0.408..0.414 rows=89 loops=1)
        Sort Key: _nango_configs.unique_key
        Sort Method: quicksort  Memory: 29kB
        ->  Hash Right Join  (cost=447.54..518.21 rows=138 width=18) (actual time=0.380..0.391 rows=89 loops=1)
              Hash Cond: (_nango_sync_configs.nango_config_id = _nango_configs.id)
              ->  Index Scan using _nango_sync_configs_environment_id_index on _nango_sync_configs  (cost=0.42..70.80 rows=75 width=6) (actual time=0.015..0.015 rows=0 loops=1)
                    Index Cond: (environment_id = 10591)
              ->  Hash  (cost=445.40..445.40 rows=138 width=20) (actual time=0.361..0.362 rows=89 loops=1)
                    Buckets: 1024  Batches: 1  Memory Usage: 13kB
                    ->  Nested Loop  (cost=5.25..445.40 rows=138 width=20) (actual time=0.046..0.349 rows=89 loops=1)
                          ->  Bitmap Heap Scan on _nango_connections  (cost=4.96..163.82 rows=143 width=8) (actual time=0.035..0.279 rows=89 loops=1)
                                Recheck Cond: ((environment_id = 10591) AND (NOT deleted))
                                Heap Blocks: exact=96
                                ->  Bitmap Index Scan on idx_connections_envid_connectionid_provider_where_deleted  (cost=0.00..4.92 rows=143 width=0) (actual time=0.021..0.022 rows=97 loops=1)
                                      Index Cond: (environment_id = 10591)
                          ->  Memoize  (cost=0.30..2.05 rows=1 width=16) (actual time=0.000..0.000 rows=1 loops=89)
                                Cache Key: _nango_connections.config_id
                                Cache Mode: logical
                                Hits: 88  Misses: 1  Evictions: 0  Overflows: 0  Memory Usage: 1kB
                                ->  Index Scan using _nango_configs_pkey on _nango_configs  (cost=0.29..2.04 rows=1 width=16) (actual time=0.008..0.008 rows=1 loops=1)
                                      Index Cond: (id = _nango_connections.config_id)
                                      Filter: (NOT deleted)
Planning Time: 0.386 ms
Execution Time: 0.507 ms
```

    

---

## Summary by mrge
Fixed and improved the logic for enforcing maximum allowed connections per plan, ensuring accurate limits and clear error messages when caps are reached.

- **Bug Fixes**
  - Corrected SQL query to count connections reliably by provider config.
  - Updated error handling to show a clear message when the connection cap is hit.
  - Improved checks for both total connections and connections with scripts enabled.


packages/shared/lib/services/sync/config/config.service.ts
```diff
@@ -487,43 +487,39 @@ export async function getSyncConfigsWithConnectionsByEnvironmentId(environment_i
...
-            '_nango_configs.environment_id': environment_id,
-            '_nango_configs.unique_key': providerConfigKey,
-            active: true,
-            enabled: true,
-            '_nango_configs.deleted': false,
-            [`${TABLE}.deleted`]: false
-        });
+        .from('_nango_connections')
+        .join('_nango_configs', function () {
+            this.on('_nango_configs.id', '_nango_connections.config_id').andOnVal('_nango_configs.deleted', false);
```
bodinsamuel, line 505: this is the main change

bodinsamuel: Previous PR https://github.com/NangoHQ/nango/pull/3935/files#diff-aa1d59c80f805a69cc66a3362df92019a5b1c280c078fba7fdc20d98fa2e1560

TBonnin: I am personally not a huge fan of filtering conditions used as join conditions (also done below for the environment). Somehow my brain find it hard to parse.
Does it affect the query plan if they are simply `where` conditions?

TBonnin (approved):

packages/shared/lib/services/sync/config/config.service.ts
bodinsamuel: I have the opposite view ahah I found it weird that stuff are not grouped. 
I think most of the time it's doesn't affect the plan, especially if the join is inner, but if it's a left join I think there are some cases where it can be different or even incorrect

PR #3950, opened 2025-04-24
feat(api): GET /plans

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-3154/api-expose-plans

- GET /plans
To build the billing page later 

    

---

## Summary by mrge
Added a new GET /plans API endpoint to fetch available plans, and added a React hook to access plans data in the webapp.

- **New Features**
  - Exposed GET /plans endpoint with environment support.
  - Added useApiGetPlans React hook for fetching plans in the frontend.


packages/types/lib/plans/http.api.ts
```diff
@@ -23,3 +23,12 @@ export interface PlanDefinition {
     hidden?: boolean;
     flags: Omit<Partial<DBPlan>, 'id' | 'account_id'>;
 }
+
+export type GetPlans = Endpoint<{
+    Method: 'GET';
+    Path: '/api/v1/plans';
+    Querystring: { env: string };
+    Success: {
+        data: PlanDefinition[];
```
TBonnin, line 32: do we need the data field. Can we return the list directly?

packages/webapp/src/utils/api.tsx
```diff
@@ -174,3 +174,11 @@ export function useGetHmacAPI(env: string) {
         }
     };
 }
+
+export class APIError extends Error {
+    details;
+    constructor(details: { res: Response; json: Record<string, any> | ApiError<any> }) {
+        super('err');
```
TBonnin: not sure where you are going with this error subclass. Should the default message be more explicit, even something like `api_error`?

TBonnin (approved):
question about the error class but rest looks good

packages/webapp/src/utils/api.tsx
bodinsamuel: The APIError class is only there to transport the payload, we need to throw something so that the library knows the request failed, and throwing json is not "valid". I'll change the error code to something more explicit

e.g:
```ts
const { error } = usePlan();

if (error) {
  if (error.json.code === 'feature_disabled') {
    return <div>Upgrade</div>
  }
}
```

packages/types/lib/plans/http.api.ts
bodinsamuel: It's mostly for consistency. My experience tells me that if we allow ourselves to do that, then at some point we will say "there is only one object when we get one resource, so let's return it directly," and this object might have an `error` prop, which makes error detection error-prone

PR #3955, opened 2025-04-24
fix: wrong webhook duration in logs

kaposke:
The external webhook executions were made asyncronous, meaning they where scheduled for immediate execution, but we stopped waiting for their results. The acknowledgment of success however, was not moved. This PR fixes that.

    

---

## Summary by mrge
Fixed incorrect webhook duration logs by moving success and failure logging to after webhook execution completes. This ensures logs now reflect the actual runtime of each webhook.


packages/jobs/lib/execution/webhook.ts
```diff
@@ -168,6 +168,8 @@ export async function startWebhook(task: TaskWebhook): Promise<Result<void>> {
 }
 
 export async function handleWebhookSuccess({ taskId, nangoProps }: { taskId: string; nangoProps: NangoProps }): Promise<void> {
+    const logCtx = logContextGetter.get({ id: String(nangoProps.activityLogId), accountId: nangoProps.team.id });
```
TBonnin: I think the `String()` is a leftover from a time the `activityLogId` could be a string or a number. Now it is exclusively a string so no need for it anymore

TBonnin (approved):

bodinsamuel (approved):

PR #3958, opened 2025-04-25
fix(persist): adapt wording for batchDelete, log missing error

bodinsamuel:
## Changes

- Adapt wording and lock name for batchDelete
- Notice we were not logging error which was confusing 
They only appeared in the operation payload which I never check

@TBonnin I found a "bug" in the batchDelete that took me a while to understand. If you batchDelete a record that never existed in the first place, it will create it (which is debatable), but because it actually creates it, the computed stats are also wrong. Didn't find an easy solution

    

---

## Summary by mrge
Updated log messages for batchDelete to use clearer wording and lock names, and added missing error logging for better debugging.


TBonnin (approved):

TBonnin:
>I found a "bug" in the batchDelete that took me a while to understand. If you batchDelete a record that never existed in the first place, it will create it (which is debatable), but because it actually creates it, the computed stats are also wrong. Didn't find an easy solution

Assuming we don't want to fix the debatable behavior, is the fix for the stats simply addding a extra condition 
```
WHEN existing.external_id IS NULL AND upsert.deleted_at IS NOT NULL THEN 'inserted'
``` 

here: https://github.com/NangoHQ/nango/blob/8d2ba82c2eedd3170ed55009a9418829016b2787/packages/records/lib/models/records.ts#L385 so the record is not considered `inserted`

bodinsamuel:
Thanks for the suggestion, I'll note that in Linear to fix later 👌🏻

PR #3960, opened 2025-04-25
refactor(runner): run scripts in worker thread

TBonnin:
The objective is to isolate each script execution to a worker thread and monitor the memory consumption of each thread.

For now I have kept the memory warning mechanism we had before, only adding the count of concurrent scripts running and the memory consumption of current script. I have discussed with @bastienbeurier and we would like to also add the max memory consumed to the script operation payload, the same way we have run time. I'll add that in next PR. This one is already big enough

The main complication is due to the Locks mechanism that was until now stored in a Map but it cannot be shared across threads unfortunately. I have rewritten the Locks class to be backed up by a SharedArrayBuffer that can be shared across threads
    
Will do more testing in staging on Monday

---

## Summary by mrge
Scripts now run in isolated worker threads, allowing each script’s memory usage to be tracked and managed separately.

- **Refactors**
  - Moved script execution to worker threads for better isolation and monitoring.
  - Rewrote the Locks mechanism to use SharedArrayBuffer for cross-thread locking.
  - Updated monitoring to track memory usage and concurrent scripts per worker.


packages/runner/lib/env.ts
```diff
@@ -1,9 +1,7 @@
 import { ENVS, parseEnvs } from '@nangohq/utils';
 import { getJobsUrl, getPersistAPIUrl } from '@nangohq/shared';
 
-export const envs = parseEnvs(ENVS.required({ RUNNER_NODE_ID: true }));
```
TBonnin, line 4: not fully related but I have been wanting to do that for a long time. It should not required for enterprise setup

packages/runner/lib/sdk/locks.ts
```diff
@@ -2,62 +2,294 @@
 import type { Result } from '@nangohq/utils';
```
TBonnin, line 2: I wish I didn't have to write the code below but since Locks must be shared across worker threads we must store the locks state in a SharedArrayBuffer.

It is a lot of code (got helped by Gemini) and handling race conditions is tricky. Remind me of my college years writing some C :/
I don't expect you to fully review every single line but go through it at least to get the gist of it (Got Gemini to help me with extra comments)

packages/runner/lib/state.ts
```diff
@@ -1 +0,0 @@
-export const abortControllers = new Map<string, AbortController>();
```
TBonnin, line 1: instead of keeping 2 pieces of state, here and in `Monitor`, I decided `Monitor` will be the place where worker references will be kept

packages/runner/lib/sdk/locks.ts
```diff
@@ -2,62 +2,336 @@
...
 import { Err, Ok } from '@nangohq/utils';
 
-interface Lock {
-    key: string;
-    owner: string;
-    expiresAt: Date;
-}
+const MAX_LOCKS = 256;
```
TBonnin, line 5: this implementation only allows for 256 different locks. I didn't want to deal with dynamic allocation.
Given how niche of a feature it is I think it is a acceptable limitation

packages/runner/lib/sdk/locks.ts
```diff
@@ -2,62 +2,336 @@
...
+
+    private getStateIndex(slotIndex: number): number {
+        return slotIndex * (BYTES_PER_LOCK / 4) + STATE_OFFSET / 4;
+    }
+    private getExpiryIndex(slotIndex: number): number {
+        return slotIndex * (BYTES_PER_LOCK / 8) + EXPIRY_OFFSET / 8;
+    }
+
+    private findSlotByKey(keyBytes: Uint8Array): number {
+        for (let i = 0; i < MAX_LOCKS; i++) {
```
TBonnin, line 140: looking for a lock is O(n). 
Going for hashing and collisions handling would be more complicated and given the limited max number of locks (256) I think this is a acceptable tradeoff. let me know if you disagree

packages/runner/lib/sdk/locks.ts
```diff
@@ -2,62 +2,420 @@
...
 
-interface Lock {
-    key: string;
-    owner: string;
-    expiresAt: Date;
-}
+const MAX_LOCKS = 256;
+const STATE_FREE = 0;
+const STATE_LOCKED = 1;
+const MAX_STRING_LEN = 255; // Max length for key and owner strings
```
TBonnin, line 8: owner and key strings are fixed size which is a bit of wasted space, especially if most of the lock keys used are short but once again I didn't want to deal with arbitrary length

packages/runner/lib/sdk/locks.ts
bodinsamuel: yeah I can't help for this file 😓

packages/runner/lib/server.ts
```diff
@@ -60,31 +57,28 @@ function startProcedure() {
...
+            worker.on('error', (err) => {
+                logger.error(`Task ${taskId} failed`, { err });
+            });
+            worker.on('exit', (exitCode) => {
+                if (exitCode) {
+                    logger.error(`Task ${taskId} exited with code ${exitCode}`);
                 }
+                monitor.untrack(worker);
             });
+            monitor.track(worker);
```
bodinsamuel: Can you track it just before start, just to be sure?

packages/runner/lib/workers/worker.integration.test.ts
```diff
@@ -0,0 +1,140 @@
+import { expect, describe, it, vi, beforeAll, afterAll } from 'vitest';
+import type { DBSyncConfig, NangoProps } from '@nangohq/types';
+import { RunnerWorker } from './worker.js';
+import express from 'express';
+import type { Server } from 'node:http';
+import getPort from 'get-port';
+import { Locks } from '../sdk/locks.js';
```
bodinsamuel, line 7: Not sure what is your setup but it's not auto-fixing the warning for the import ordering

packages/runner/lib/workers/worker.ts
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    async run(): Promise<void> {
+        const heartbeat = setInterval(async () => {
+            try {
+                await this.jobsClient.postHeartbeat({ taskId: this.taskId });
+            } catch (err) {
+                this.logger.error('Heartbeat failed', { err });
+            }
+        }, this.heartbeatIntervalMs);
```
bodinsamuel: should this be a setTimeout to avoid overlapping calls if jobs is unresponsive?

packages/runner/lib/workers/worker.ts
```diff
@@ -0,0 +1,153 @@
...
+            await this.jobsClient.putTask({
+                taskId: this.taskId,
+                nangoProps: this.execOpts.nangoProps,
+                ...(error ? { error } : { output: output as any })
+            });
+            this.logger.info(`Completed task: ${this.taskId}`);
+        } finally {
+            clearInterval(heartbeat);
+            clearInterval(memoryUsage);
+            process.exit(0);
```
bodinsamuel: can you run on nextTick/few ms later, sometimes logger and other stuff are not done yet

bodinsamuel (approved):
Tested everything I could: syntax error, locks, validation, logs, batch save, proxy, abort, etc. All good 💪🏻 🔥

packages/runner/lib/workers/worker.ts
TBonnin: ~will do~

actually it is cleaner to call `parent.close`

packages/runner/lib/workers/worker.ts
TBonnin: good point

packages/runner/lib/server.ts
TBonnin: yes

TBonnin:
worker threads is too much overhead for the way we are executing script. Loading all the dependencies required for exec.ts is requiring 100mb+ when the worker thread is started. There is still a significant overhead passed the startup time, meaning that creating a static pool a workers would still consume more memory (in addition to add a lot of complexity to manage the worker pool, the queuing, etc…).
For now we abandon the idea of running scripts in worker threads

PR #3963, opened 2025-04-28
feat(jira): update Jira integration with multiple domain support

viictoo:
# Allow specifying Atlassian site during connection creation

## Situation
When connecting to Jira/Confluence via OAuth, Atlassian's API returns all sites a user has access to through the accessible resources endpoint (/oauth/token/accessible-resources). A single access token can be valid for multiple Atlassian sites (e.g., nango-hq.atlassian.net and nango-test.atlassian.net) and have different permission scopes across these sites.
Currently, our post-connection script automatically selects the first available site from this list, which may not be the site the user intends to connect to. This can lead to confusion when users have access to multiple Atlassian instances but want to work with a specific one.
This behavior is documented in the [Atlassian Community](https://community.developer.atlassian.com/t/urgent-api-accessible-resources-endpoint-returns-sites-resources-that-are-not-permitted-by-the-user/66899/8).

## Task
Implement the ability for users to specify which Atlassian site they want to connect to during connection creation, while maintaining backward compatibility for existing connections.

## Action
1. Updated the post-connection script to:
   - Continue fetching accessible sites via `/oauth/token/accessible-resources`
   - Check if baseUrl exists in connection config
   - If baseUrl exists, find the matching site from accessible resources
   - If no match found, silently fall back to first site (maintaining backward compatibility)
   - Store the selected site's cloudId, baseUrl, and accountId in connection config

2. Updated documentation to:
   - Explain the two connection options (domain-specific vs auto-select)
   - Clarify the multi-site token behavior
   - Document the connection process steps
   - Update API gotchas with new connection behavior

## Result
- Users can now specify which Attlassian site to connect to by providing the domain during connection creation
- Existing connections continue to work with the first-site behavior
- Clear documentation on how to handle multi-site connections

## Testing
- [x] Test connection creation with domain specified
- [x] Test connection creation without domain (legacy behavior)
- [x] Test with token having access to multiple sites

## Migration
No migration needed. This is a backward-compatible change:
- Existing connections will continue to work as before
- New connections can optionally specify a domain

[EXT-661](https://linear.app/nango/issue/EXT-661/fix-atlassian-issue)
    

---

## Summary by mrge
Added support for connecting to a specific Jira domain when creating a connection, improving multi-site handling and user control.

- **New Features**
  - Users can now specify a Jira domain during connection setup.
  - The integration selects the correct site based on the provided domain or defaults to the first available site for backward compatibility.
  - Updated documentation to explain domain selection and multi-site behavior.


packages/providers/providers.yaml
```diff
@@ -5124,14 +5136,14 @@ jira-basic:
...
-        base_url: https://${connectionConfig.subdomain}.atlassian.net
+        base_url: https://${connectionConfig.baseUrl}.atlassian.net
         verification:
             method: GET
             endpoints:
                 - /rest/api/3/myself
     docs: https://docs.nango.dev/integrations/all/jira-basic
     connection_config:
-        subdomain:
+        baseUrl:
```
bodinsamuel: you can't change this it will break existing install

packages/providers/providers.yaml
```diff
@@ -5114,6 +5116,16 @@ jira:
...
+    connection_config:
+        baseUrl:
+            type: string
+            title: Domain
+            optional: true
+            description: The domain of your Attlassian account
+            pattern: '^[a-zA-Z0-9.-]+$'
+            example: subdomain
+            suffix: .atlassian.net
+            prefix: https://
```
bodinsamuel, line 5261: it should be named subdomain, as it's not a full url

viictoo: renamed to subdomain for the UI but preserved as baseUrl in the connection config as that is what is expected in the postConnection script

packages/providers/providers.yaml
viictoo: removed this edit. unrelated to this PR. Thanks

packages/server/lib/hooks/connection/providers/jira/post-connection.ts
```diff
@@ -12,20 +15,42 @@ export default async function execute(nango: Nango) {
...
     }
 
-    const { accountId } = accountResponse.data;
-
-    await nango.updateConnectionConfig({ cloudId, baseUrl, accountId });
+    await nango.updateConnectionConfig({
+        cloudId: site.id,
+        baseUrl: site.url,
+        accountId: accountResponse.data.accountId
+    });
```
TBonnin, line 47: can avoid the repetition like this:
```
const accountId = accountResponse?.data?.accountId
await nango.updateConnectionConfig({
        cloudId: site.id,
        baseUrl: site.url,
        accountId: ...(accountId ? { accountId } : {})
    });
```

TBonnin (approved):

packages/server/lib/hooks/connection/providers/jira/post-connection.ts
viictoo: 🙌 
Updated ✅

PR #3984, opened 2025-05-01
chore(fleet): expose a overrideNodeConfig function

TBonnin:
the function is to be used in an internal endpoint to change the spec of a given runner instance.
    

---

## Summary by mrge
Added a new overrideNodeConfig function to let internal endpoints update or reset the config for a specific runner instance.

- **Refactors**
  - Replaced create and update methods with a single upsert method for node config overrides.
  - Updated tests and error codes to use the new upsert logic.


packages/fleet/lib/fleet.ts
```diff
@@ -155,4 +155,26 @@ export class Fleet {
...
+            defaultConfig.cpuMilli === override.cpuMilli &&
+            defaultConfig.memoryMb === override.memoryMb &&
+            defaultConfig.storageMb === override.storageMb;
+
+        if (isDefault) {
+            return nodeConfigOverrides.remove(this.dbClient.db, override.routingId);
+        }
+
+        return nodeConfigOverrides.upsert(this.dbClient.db, override);
+    }
```
bodinsamuel, line 181: should this be in a transaction? fleet could modify or access this in between calls no?

bodinsamuel (approved):

packages/fleet/lib/fleet.ts
TBonnin: indeed. concurrent calls to this function could lead to race condition. Let me add a `FOR UPDATE` and transaction

PR #3986, opened 2025-05-01
chore(scheduler): add groups table

TBonnin:
Adding a groups table in scheduler in order to 
- store the max concurrency value for each group 
- ensure each group is only processed by one processor at a time. (since the algo is going to be 1. check how many slots are available for a group 2. select candidate tasks for this group limit slots, we are gonna have some locking at the group level)

    

---

## Summary by mrge
Added a groups table to the scheduler to store max concurrency for each group and ensure only one processor handles a group at a time.

### DIFF
--- a/packages/scheduler/lib/db/migrations/20250501173800_create_groups_table.ts
+++ b/packages/scheduler/lib/db/migrations/20250501173800_create_groups_table.ts
@@ -1,0 +1,17 @@
+import type { Knex } from 'knex';
+import { GROUPS_TABLE } from '../../models/tasks.js';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.raw(`
+            CREATE TABLE IF NOT EXISTS ${GROUPS_TABLE} (
+                key varchar(255) PRIMARY KEY,
+                max_concurrency integer NOT NULL,
+                created_at timestamp with time zone NOT NULL,
+                last_modified_at timestamp with time zone NOT NULL,
+                deleted_at timestamp with time zone NULL,
+                last_task_added_at timestamp with time zone NULL
+            );
+        `);
+}
+
+export async function down(): Promise<void> {}
--- a/packages/scheduler/lib/db/migrations/20250501173800_create_groups_table.ts
+++ b/packages/scheduler/lib/db/migrations/20250501173800_create_groups_table.ts
@@ -1,17 +1,36 @@
 import type { Knex } from 'knex';
-import { GROUPS_TABLE } from '../../models/tasks.js';
+import { GROUPS_TABLE } from '../../models/groups.js';
 
 export async function up(knex: Knex): Promise<void> {
-    await knex.raw(`
+    await knex.transaction(async (trx) => {
+        await trx.raw(`
             CREATE TABLE IF NOT EXISTS ${GROUPS_TABLE} (
                 key varchar(255) PRIMARY KEY,
                 max_concurrency integer NOT NULL,
                 created_at timestamp with time zone NOT NULL,
-                last_modified_at timestamp with time zone NOT NULL,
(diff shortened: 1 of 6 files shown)


packages/scheduler/lib/db/migrations/20250501173800_create_groups_table.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.raw(`
+            CREATE TABLE ${GROUPS_TABLE} (
+                key varchar(255) PRIMARY KEY,
+                max_concurrency integer NOT NULL,
+                created_at timestamp with time zone NOT NULL,
+                last_modified_at timestamp with time zone NOT NULL,
+                deleted_at timestamp with time zone NULL,
+                last_task_added_at timestamp with time zone NULL
```
TBonnin, line 13: useful to cleanup groups that have not been used recently

packages/scheduler/lib/db/migrations/20250501173800_create_groups_table.ts
```diff
@@ -0,0 +1,17 @@
+import type { Knex } from 'knex';
+import { GROUPS_TABLE } from '../../models/tasks.js';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.raw(`
+            CREATE TABLE ${GROUPS_TABLE} (
+                key varchar(255) PRIMARY KEY,
```
TBonnin, line 8: ideally I would like the PK here to be referenced as the `group_key` column in the `tasks` table but adding a FK constraint on `tasks` table would lock it. So as a result there will be no FK enforcement by pg. If someone has an idea about how to do it without any live impact on the `tasks` table in production, let me know.

packages/scheduler/lib/db/migrations/20250501173800_create_groups_table.ts
```diff
@@ -0,0 +1,17 @@
...
+                max_concurrency integer NOT NULL,
+                created_at timestamp with time zone NOT NULL,
+                last_modified_at timestamp with time zone NOT NULL,
```
bodinsamuel: why not just updated_at?

bodinsamuel (approved):
Do you already have an implementation or it's an educated guess?

packages/scheduler/lib/db/migrations/20250501173800_create_groups_table.ts
TBonnin: updated_at is indeed better

TBonnin:
> Do you already have an implementation or it's an educated guess?

I don't have a full implementation but the query is gonna look in concept similar to:
```sql
WITH candidates AS (
   SELECT 
      id, 
      group_key,
      RANK () OVER (PARTITION BY group_key ORDER BY starts_at ASC) rank
      max_concurrency
   FROM tasks 
   JOIN groups ON tasks.group_key = groups.key
   WHERE state = 'CREATED'
   FOR UPDATE SKIP LOCKED
),
WITH running AS (
   SELECT 
      count(id), 
      max_concurrency,
      group_key 
   FROM tasks 
   WHERE state = 'STARTED' 
   AND key IN (SELECT distinct group_key FROM candidates)
   GROUP BY group_key
)
SELECT id, ...
FROM candidates
JOIN running ON candidates.group_key = running.group_key
WHERE rank + count <= max_concurrency
```

TBonnin:
@bodinsamuel I've added the groups model and inserting the existing groups in the migration. Can I have another review? Thank you

I am planning to keep things simple and upsert the group every time a task is created. 
It means the max_concurrency will be the one from the last task but the plan is to keep the max_concurrency stable anyway. It avoids having to manage tasks and groups independently.

bodinsamuel (approved):

PR #3987, opened 2025-05-01
feat(scheduler): dequeue on group key pattern

TBonnin:
In order to support script execution max concurrency, we are gonna need to have dynamic groups/queues in the scheduler.
This commit allows to dequeue tasks based on a group_key pattern. 
Keeping things simple for now and only allowing wildcard `*`

query plan in prod before:
```
Limit  (cost=2280.57..2285.57 rows=400 width=22) (actual time=0.466..0.479 rows=7 loops=1)
  ->  LockRows  (cost=2280.57..2298.02 rows=1396 width=22) (actual time=0.465..0.477 rows=7 loops=1)
        ->  Sort  (cost=2280.57..2284.06 rows=1396 width=22) (actual time=0.392..0.392 rows=7 loops=1)
              Sort Key: id
              Sort Method: quicksort  Memory: 25kB
              ->  Index Scan using idx_tasks_group_key_created on tasks  (cost=0.38..2213.26 rows=1396 width=22) (actual time=0.378..0.383 rows=7 loops=1)
                    Index Cond: ((group_key)::text = 'sync'::text)
                    Filter: ((state = 'CREATED'::nango_scheduler.task_states) AND (starts_after <= CURRENT_TIMESTAMP))
Planning Time: 0.492 ms
Execution Time: 0.512 ms
```

query plan in prod after:
```
Limit  (cost=2595.85..2600.85 rows=400 width=22) (actual time=0.776..0.777 rows=0 loops=1)
  ->  LockRows  (cost=2595.85..2613.30 rows=1396 width=22) (actual time=0.775..0.776 rows=0 loops=1)
        ->  Sort  (cost=2595.85..2599.34 rows=1396 width=22) (actual time=0.774..0.775 rows=0 loops=1)
              Sort Key: id
              Sort Method: quicksort  Memory: 25kB
              ->  Index Scan using idx_tasks_group_key_created on tasks  (cost=0.38..2528.53 rows=1396 width=22) (actual time=0.754..0.754 rows=0 loops=1)
                    Filter: (((group_key)::text ~~ 'sync%'::text) AND (state = 'CREATED'::nango_scheduler.task_states) AND (starts_after <= CURRENT_TIMESTAMP))
Planning Time: 0.181 ms
Execution Time: 0.804 ms
```

*How to test*
- *No behavior has been changed so just ensure all types of task are being processed.*

### DIFF
--- a/packages/jobs/lib/processor/processor.ts
+++ b/packages/jobs/lib/processor/processor.ts
@@ -17,28 +17,28 @@ export class Processor {
         try {
             const syncWorker = new ProcessorWorker({
                 orchestratorUrl: this.orchestratorServiceUrl,
-                groupKey: 'sync',
+                groupKey: 'sync*',
                 maxConcurrency: 200
             });
             syncWorker.start();
 
             const actionWorker = new ProcessorWorker({
                 orchestratorUrl: this.orchestratorServiceUrl,
-                groupKey: 'action',
+                groupKey: 'action*',
                 maxConcurrency: 200
             });
             actionWorker.start();
 
             const webhookWorker = new ProcessorWorker({
                 orchestratorUrl: this.orchestratorServiceUrl,
-                groupKey: 'webhook',
+                groupKey: 'webhook*',
                 maxConcurrency: 50
             });
             webhookWorker.start();
 
             const onEventWorker = new ProcessorWorker({
                 orchestratorUrl: this.orchestratorServiceUrl,
-                groupKey: 'on-event',
+                groupKey: 'on-event*',
                 maxConcurrency: 50
             });
             onEventWorker.start();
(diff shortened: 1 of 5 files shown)


bodinsamuel (approved):
Not sure I get the objective but let see :D

TBonnin:
> Not sure I get the objective but let see :D

Sorry I didn't explain properly.
We currently have tasks grouped by type (sync, actions, etc..) and have one processor per type.

In order to introduce max concurrency we are gonna need to differentiate tasks that belongs to the same concurrency group. To achieve this we could:
1. make task's `group_key` dynamic, ex: `sync:account:1` and allow the processor to dequeue multiple groups at once using a pattern (ex: `sync:*`)
2. make task's `group_key` dynamic and have a dynamic number of processor, each one dequeuing a single group.
3. Split the concept of `group` in two: one for concurrency and another one for dequeueing

I went for option 1 because it easy to implement (`LIKE` vs `=`) and still performant (SQL query is still fast enough and the number of processors stays low)

Does that make sense?

PR #3988, opened 2025-05-02
feat(jamf): Add support for Jamf in Nango

lordsarcastic:
- Introduced two new integrations for Jamf Pro: Bearer Token and Client Credentials.
- Added corresponding documentation files for each integration, detailing access requirements, setup guides, useful links, and API gotchas.
- Updated providers.yaml to register both integrations with their respective authentication modes and configuration details.

Ticket: [EXT-663](https://linear.app/nango/issue/EXT-663/add-support-for-jamf-in-nango)

**Additional note**: While Jamf Basic authentication provides an endpoint to invalidate an existing token and create a new one, Nango does not yet support that process for `TWO_STEP` auth mode. Instead, we rely on using the same endpoint for generating a token, to generate a new one. This isn't a concern since each token expires in 20 minutes

    

---

## Summary by mrge
Added support for Jamf Pro in Nango with two new integrations: Bearer Token and Client Credentials authentication.

- **New Features**
  - Registered both Jamf Pro authentication modes in providers.yaml with full config.
  - Added documentation for each integration, including setup info and useful links.


packages/providers/providers.yaml
```diff
@@ -5221,6 +5221,66 @@ intuit:
...
+            type: string
+            title: Jamf Pro URL
+            description: Your Jamf Pro instance URL (e.g. https://yourinstance.jamfcloud.com)
+            pattern: '^[a-z0-9_-]+$'
+            example: yourinstance.jamfcloud.com
+            prefix: https://
+            order: 1
+
+jamf-basic:
+    display_name: Jamf Pro (Bearer Token)
```
hassan254-prog: Might name this Basic Auth

packages/providers/providers.yaml
```diff
@@ -5221,6 +5221,66 @@ intuit:
...
+    categories:
+        - other
+    auth_mode: TWO_STEP
+    proxy:
+        base_url: ${connectionConfig.baseUrl}
+        headers:
+            authorization: Bearer ${token}
+    token_url: ${connectionConfig.baseUrl}/api/v1/auth/token
+    token_request_method: POST
```
hassan254-prog: The method is `POST` by default, you can omit it here and also within the validation scripts.

packages/providers/providers.yaml
```diff
@@ -5221,6 +5221,66 @@ intuit:
...
+    token_headers:
+        authorization: Basic ${base64(${credentials.username}:${credentials.password})}
+    token_response:
+        token: response.token
```
hassan254-prog: The response received is: 

```
{
  "token": "eyJhbGciOiJIUzUxMiJ9.eyJhdXRoZW50aWNhdGVkLWFwcCI6IkdFTkVSSUMiLCJhdXRoZW50aWNhdGlvbi10eXBlIjoiSlNTIiwiZ3JvdXBzIjpbXSwic3ViamVjdC10eXBlIjoiSlNTX1VTRVJfSUQiLCJ0b2tlbi11dWlkIjoiNzc0YWY3MGYtYWQ0Yy00N2QzLTk2MzktZjEwMjBhMTIwYzExIiwibGRhcC1zZXJ2ZXItaWQiOi0xLCJzdWIiOiIxIiwiZXhwIjoxNTM5NjE5MzQ4fQ.0t7sgYyIyA7kTTmrM8tMGE7fnXcJ1ZzQODAJp0pzg92-cBMQS0Cv8S9oWjkJD7VJS-CHA1dOppr0G_2dCPOfng",
  "expires": "2020-04-21T21:09:31.626Z"
}
```
so the correct path would be 

```
token_response:
        token: token
        token_expiration: expires
 ```

hassan254-prog (commented):
- Does Jamf offer self-hosting? If not, the URL should include the `https://` prefix. Also, Jamf uses the endpoint `https://yourserver.jamfcloud.com/api/v1/oauth/token`, wouldn't `https://{instance}.jamfcloud.com/api/v1/oauth/token` work, consider the other urls within the PR.
- Might be good to have a symbolic link for the logo instead.

packages/providers/providers.yaml
lordsarcastic: It is actually 'GET'. I had to change the schema.json to allow for POST method

hassan254-prog: I meant, the method within the code is `POST` by [default](https://github.com/NangoHQ/nango/blob/26a13fd510b62d2eca02080ec6fbae954aed903a/packages/shared/lib/services/connection.service.ts#L1129), you can omit this line and the one in validation as well.

hassan254-prog (approved):

PR #3990, opened 2025-05-02
feat(webapp): jump to specific time in logs

kaposke:
### Description
In the log filters, we currently allow selecting custom date ranges but we give no control over specific TIME ranges. This PR reworks the DatePicker (maybe not the best name anymore?) to allow for custom inputs, with heavy inspiration on DataDog. See [Figma design](https://www.figma.com/design/a4t7gaUATMqUcWedowFPFd/Design-Sytem---Nango?node-id=4516-34222&p=f&t=pPuIjoDVaG3A0M37-0).

I took some liberties over the design and functionality. The custom input always comes pre-filled with the current time range based on the selected preset. This way serves as both an example and a starting point for editing:
image

I've also free-styled errors:
image

### Testing
Please checkout to this branch and try to break it. Any feedback is welcome.

    

---

## Summary by mrge
Added support for jumping to a specific time range in logs by allowing custom date and time input in the DatePicker.

- **New Features**
  - Users can now enter a custom date and time range to filter logs.
  - Improved error handling and input validation for custom ranges.
  - Updated UI to show preset time ranges with short labels.

### DIFF
--- a/packages/webapp/src/pages/Logs/components/DatePicker.tsx
+++ b/packages/webapp/src/pages/Logs/components/DatePicker.tsx
@@ -1,4 +1,3 @@
-import { CalendarIcon } from '@radix-ui/react-icons';
 import { IconCalendar, IconCheck } from '@tabler/icons-react';
 import { format, subDays } from 'date-fns';
 import { useEffect, useMemo, useState } from 'react';
@@ -115,7 +114,7 @@ export const DatePicker: React.FC<{
                     size={'sm'}
                     className={cn('flex-grow truncate text-text-light-gray', period && selectedPreset?.name !== 'last24h' && 'text-white')}
                 >
-                    <CalendarIcon />
+                    <IconCalendar size={18} />
                     {display} {isLive && '(live)'}
                 </Button>
             </PopoverTrigger>
@@ -131,7 +130,7 @@ export const DatePicker: React.FC<{
                             )}
                         >
                             <div className="w-12 flex-shrink-0 flex justify-center items-center h-full p-1 bg-grayscale-5 rounded-md font-medium">
-                                <IconCalendar className="w-[18px] h-[18px]" />
+                                <IconCalendar size={18} />
                             </div>
                             <form onSubmit={onSubmitCustomRange} className="w-full">
                                 <input
--- a/packages/webapp/src/pages/Logs/components/DatePicker.tsx
+++ b/packages/webapp/src/pages/Logs/components/DatePicker.tsx
@@ -119,9 +119,10 @@ export const DatePicker: React.FC<{
                     {display} {isLive && '(live)'}
                 </Button>
             </PopoverTrigger>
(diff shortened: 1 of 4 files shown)


TBonnin:
I'll test it after lunch but I have a first comment. Can we show the timezone (even if read-only)? I already imagine the support request coming with a screenshot and me trying to guess what it means in UTC time

kaposke:
> I'll test it after lunch but I have a first comment. Can we show the timezone (even if read-only)? I already imagine the support request coming with a screenshot and me trying to guess what it means in UTC time

Great idea. Wyt?
image

TBonnin (commented):
works well. 
Without autocomplete, it is quite hard to match the exact format, especially if starting to write the entire date range from scratch. It is better than not having the feature though but I feel like we are quickly gonna need an easier way to input the dates

packages/webapp/src/utils/logs.tsx
```diff
@@ -90,14 +95,39 @@ export function getPresetRange(preset: PresetNames): DateRange {
...
     }
 
     return null;
 }
+
+export function parseDateRange(input: string, dateTimeFormat: string): { dateRange: DateRange | null; error: string | null } {
+    // Validate input format
+    const dateFormatRegex = /^[A-Z][a-z]{2} \d{1,2}, \d{1,2}:\d{2} - [A-Z][a-z]{2} \d{1,2}, \d{1,2}:\d{2}$/;
+    if (!dateFormatRegex.test(input)) {
+        return { dateRange: null, error: 'Invalid format' };
```
TBonnin: should the valid format be part of the error message.

kaposke: It could yeah. It's also in the place holder, and the autofilled value when you open the popover though.

TBonnin: Screenshot 2025-05-05 at 09 41 36

Yes but when I get the error and I need to fix it I cannot see the valid format to understand what's wrong with what I entered

kaposke: Good point. Added!

packages/webapp/src/pages/Logs/components/DatePicker.tsx
```diff
@@ -113,30 +119,51 @@ export const DatePicker: React.FC<{
...
+                        <span className="text-xs text-grayscale-8 self-end">UTC{format(new Date(), 'XXX')}</span>
+                        <div
+                            className={cn(
+                                'h-10 flex items-center bg-grayscale-3 rounded-md',
+                                !selectedPreset && 'border border-grayscale-8',
+                                rangeInputErrorMessage && 'border border-alert-red'
+                            )}
+                        >
+                            <div className="w-12 flex-shrink-0 flex justify-center items-center h-full p-1 bg-grayscale-5 rounded-md font-medium">
+                                <IconCalendar className="w-[18px] h-[18px]" />
```
bodinsamuel: It's recommended to use `size={18}`

bodinsamuel (approved):
Works well 🙏🏻

PR #3995, opened 2025-05-05
fix(github-app-oauth): refresh user credentials when refreshing github app oauth connection

hassan254-prog:

- Allow for refreshing of user credentials when refreshing github app oauth connection

    

---

## Summary by mrge
Updated the GitHub App OAuth flow to refresh user credentials when refreshing the app connection. This ensures both app and user tokens stay up to date.

- **Bug Fixes**
  - Refreshed user credentials are now saved alongside app credentials during OAuth refresh.
  - Updated credential handling to support combined app and user tokens.

### DIFF
--- a/packages/shared/lib/models/Auth.ts
+++ b/packages/shared/lib/models/Auth.ts
@@ -100,11 +100,6 @@ export interface OAuth2Credentials extends CredentialsCommon {
         client_secret?: string;
     };
 }
-export interface CombinedOauth2AppCredentials extends CredentialsCommon {
-    type: 'CUSTOM';
-    app: AppCredentials;
-    user: OAuth2Credentials | null;
-}
 export interface OAuth2ClientCredentials extends CredentialsCommon {
     type: 'OAUTH2_CC';
     token: string;
--- a/packages/shared/lib/models/Auth.ts
+++ b/packages/shared/lib/models/Auth.ts
@@ -100,7 +100,11 @@ export interface OAuth2Credentials extends CredentialsCommon {
         client_secret?: string;
     };
 }
-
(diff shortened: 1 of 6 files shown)


packages/shared/lib/services/connection.service.ts
```diff
@@ -988,6 +989,50 @@ class ConnectionService {
...
+                logCtx
+            }),
+            githubAppClient.createCredentials({
+                integration: config,
+                provider: provider as ProviderGithubApp,
+                connectionConfig: connection.connection_config
+            })
+        ]);
+
+        if (!appResult.isOk()) {
```
TBonnin: let's use `isErr` instead of `!isOk`

packages/shared/lib/models/Auth.ts
```diff
@@ -100,7 +100,11 @@ export interface OAuth2Credentials extends CredentialsCommon {
         client_secret?: string;
     };
 }
-
+export interface CombinedOauth2AppCredentials extends CredentialsCommon {
```
TBonnin: why not use the type defined in the `types` package?

bodinsamuel: same question

bodinsamuel (approved):
Not tested, but looks good

packages/shared/lib/services/connection.service.ts
hassan254-prog: thanks, updated with the latest commit 🙏 .

packages/shared/lib/models/Auth.ts
hassan254-prog: updated with the latest commit 🙏 .

bodinsamuel (approved):

PR #3997, opened 2025-05-05
chore(scheduler): update groups max concurrency

TBonnin:
Depends on https://github.com/NangoHQ/nango/pull/3986

Set the max concurrency for a task group when running immediate tasks (actions/webhooks) or scheduling new schedules (sync). Max concurrency for existing groups is currenlty always 0 (aka: no max concurrency)

How to test:
- Not much to test. Just make sure running scripts is working as before

    

---

## Summary by mrge
Set the max concurrency for task groups when creating immediate tasks or new schedules. This allows each group to control how many tasks run at the same time.

- **Refactors**
  - Updated scheduler logic to set group max concurrency from input props.
  - Cleaned up group deletion for unused groups.

---

**Task Group Concurrency Management in Scheduler**

This PR enhances the scheduler system by adding max concurrency control for task groups. It allows limiting how many tasks can run in parallel within a group when creating immediate tasks (actions/webhooks) or new schedules (syncs). This provides better resource management and system stability by preventing too many parallel tasks from overwhelming resources.

**Key Changes:**
• Added max concurrency settings to task groups to control parallel execution
• Updated scheduler to set group concurrency limits from input properties
• Modified ``API`` endpoints to accept group configuration with max concurrency
• Implemented cleanup for unused groups to prevent database bloat
• Added backward compatibility for existing code

**Affected Areas:**
• Scheduler core functionality
• Orchestrator ``API`` endpoints
• Task group management
• Database models and schemas

**Potential Impact:**

**Functionality**: Adds ability to control concurrent task execution per group, improving resource management

**Performance**: May improve system stability by preventing too many parallel tasks from overwhelming resources, but adds additional DB queries

**Security**: No direct security implications

**Scalability**: Enhances scalability by allowing fine-grained control over task execution throughput

**Review Focus:**
• Database performance impact of additional group queries
• Potential circular dependencies in type definitions
• Backward compatibility preprocessing logic
• Feature flag implementation for action-related group updates

<details>

• Verify task scheduling and execution with various concurrency settings
• Test behavior when concurrency limits are reached
• Confirm backward compatibility with existing clients
• Check group cleanup functionality works properly

</details>

<details>

**packages/scheduler/lib/models/groups.ts**: Well-designed selective updating of max_concurrency only when provided

**packages/orchestrator/lib/routes/v1/postImmediate.ts**: Good backward compatibility handling through preprocessing

**packages/scheduler/lib/types.ts**: Potential circular dependency issue with ScheduleProps redefinition

</details>

<details>

**Feature Rollout**:
• Feature flag for controlling action-related group updates

**Maintainability**:
• Regular cleanup of unused groups to prevent ``DB`` bloat

**Backward Compatibility**:
• Schema preprocessing to handle old request formats

</details>

<details>

• Additional database queries for group management could impact performance under high load
• Circular dependency in type definitions may cause compilation issues
• Feature flag dependency adds complexity to action-related group updates

</details>

---

### DIFF
--- a/packages/orchestrator/lib/routes/v1/postImmediate.ts
+++ b/packages/orchestrator/lib/routes/v1/postImmediate.ts
@@ -18,6 +18,7 @@ export type PostImmediate = Endpoint<{
         group: {
             key: string;
             maxConcurrency: number;
+            updateFlag?: boolean;
         };
         retry: {
             count: number;
@@ -63,7 +64,8 @@ const validate = validateRequest<PostImmediate>({
                 name: z.string().min(1),
                 group: z.object({
                     key: z.string().min(1),
-                    maxConcurrency: z.coerce.number()
+                    maxConcurrency: z.coerce.number(),
+                    updateFlag: z.boolean().default(false)
                 }),
                 retry: z.object({
                     count: z.number().int(),
@@ -97,6 +99,7 @@ const handler = (scheduler: Scheduler) => {
             payload: req.body.args,
             groupKey: req.body.group.key,
             groupKeyMaxConcurrency: req.body.group.maxConcurrency,
+            groupUpdateFlag: req.body.group.updateFlag,
             retryMax: req.body.retry.max,
             retryCount: req.body.retry.count,
             createdToStartedTimeoutSecs: req.body.timeoutSettingsInSecs.createdToStarted,
--- a/packages/orchestrator/lib/routes/v1/postImmediate.ts
+++ b/packages/orchestrator/lib/routes/v1/postImmediate.ts
@@ -15,7 +15,10 @@ export type PostImmediate = Endpoint<{
     Path: typeof path;
     Body: {
         name: string;
-        groupKey: string;
+        group: {
+            key: string;
+            maxConcurrency: number;
+        };
         retry: {
             count: number;
             max: number;
@@ -55,10 +58,13 @@ const validate = validateRequest<PostImmediate>({
             }
             return z.never();
         }
-        return z
+        const schema = z
             .object({
                 name: z.string().min(1),
-                groupKey: z.string().min(1),
+                group: z.object({
+                    key: z.string().min(1),
+                    maxConcurrency: z.coerce.number()
+                }),
                 retry: z.object({
                     count: z.number().int(),
                     max: z.number().int()
@@ -70,7 +76,16 @@ const validate = validateRequest<PostImmediate>({
                 }),
                 args: argsSchema(data)
             })
-            .strict()
+            .strict();
+        return z
+            .preprocess((o) => {
+                // for backwards compatibility
+                if (o && typeof o === 'object' && 'groupKey' in o) {
+                    const { groupKey, ...rest } = o;
+                    return { ...rest, group: { key: groupKey, maxConcurrency: 0 } };
+                }
+                return o;
+            }, schema)
             .parse(data);
     }
 });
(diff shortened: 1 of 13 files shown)


packages/scheduler/lib/scheduler.ts
```diff
@@ -170,16 +173,26 @@ export class Scheduler {
...
-                    startsAfter: new Date(),
+                    startsAfter: now,
                     scheduleId: null
                 };
+
+                const group = await groups.upsert(trx, {
+                    key: props.groupKey,
+                    maxConcurrency: props.groupKeyMaxConcurrency ?? 0,
+                    lastTaskAddedAt: now
+                });
```
TBonnin: I will keep an eye on how much load this extra query adds to the orch db since it executes on every actions

packages/scheduler/lib/models/groups.ts
```diff
@@ -49,9 +53,9 @@ export async function upsert(db: knex.Knex, group: Omit<Group, 'createdAt' | 'up
             .insert(toInsert)
             .onConflict('key')
             .merge({
-                max_concurrency: group.maxConcurrency,
                 last_task_added_at: group.lastTaskAddedAt,
-                deleted_at: null
+                deleted_at: null,
+                ...(group.maxConcurrency ? { max_concurrency: group.maxConcurrency } : {})
```
TBonnin, line 61: only update max concurrency if it is provided

packages/orchestrator/lib/routes/v1/postImmediate.ts
```diff
@@ -70,7 +76,16 @@ const validate = validateRequest<PostImmediate>({
...
+            .strict();
+        return z
+            .preprocess((o) => {
+                // for backwards compatibility
+                if (o && typeof o === 'object' && 'groupKey' in o) {
+                    const { groupKey, ...rest } = o;
+                    return { ...rest, group: { key: groupKey, maxConcurrency: 0 } };
+                }
+                return o;
+            }, schema)
```
TBonnin, line 90: I will remove the preprocess once this change is deployed to all services

packages/scheduler/lib/scheduler.ts
bodinsamuel: I'm a bit surprised it's defined at execution time. I would have expect to only happen once at deploy time, no?

bodinsamuel (approved):
Since I was a bit "out" last week can you confirm if I understand well?
The new group table is the source of truth, you have to query it first with rank over to know can actually execute. So you always query everything from this table and then filter out the tasks. Because of this it's not infinitely scalable, so you made the rows temporary instead of permanent config, right?

packages/scheduler/lib/scheduler.ts
TBonnin: that's the downside of letting customers choose at runtime if the action will be sync or async, which means that we don't know at deploy time in which group it is going to end up. I thought about doing it a deploy time and create 2 groups for each action (one for sync execution and one for async). Besides potentially creating a lot of unused groups, the problem with this option is that we wouldn't be able to delete any group ever, since once deleted only a deploy can add the group again.

TBonnin: I want to see how the db behaves with this extra query but if it needs optimization it would be a very good use case for a in-memory transient cache to prevent 99% of the queries

TBonnin:
> The new group table is the source of truth, you have to query it first with rank over to know can actually execute. So you always query everything from this table and then filter out the tasks. Because of this it's not infinitely scalable, so you made the rows temporary instead of permanent config, right?

The group table holds the max concurrency value but the main purpose of the table is to be able to lock against something stable to avoid race condition since the algo is the typical 2-steps read then write:
1. read how many tasks must run and how many are already running.
2. start to fill the available slots

Without locking the following race condition will occur (imagine group is the same and max concurrency is 1):
1. task A is created 
2. processor X starts dequeuing: read created/started tasks => 1 task waiting, 0 running
3. task B is created
4. processor Y  starts dequeuing: read created/started tasks => 1 task waiting (taskB, task A is locked and therefore not visible by Y), 0 running still
5. processor X starts task A
6. processor Y starts task B
=> 2 tasks are running concurrently despite max concurrency of 1

with a lock on the group, processor Y won't be able to process task B until processor X is finished (aka: a group is only processed by one processor at a time). 
In the following PR I am gonna modify the dequeue query to implement this logic. Maybe it will make more sense then

PR #4001, opened 2025-05-06
chore: update the DryRunService to improve error messages 

viictoo:
- Added logic to differentiate between variants within the same integration and scripts across different integrations, prompting users to specify the correct variant or integration ID as needed.

[EXT-610](https://linear.app/nango/issue/EXT-610/improve-dryrun-error-messaging)

**Open Issues**
Will require validation and additional testing for variants

---

This PR enhances error messaging in the CLI's DryRunService by differentiating between script variants within the same integration versus scripts across different integrations. When a script name conflict is detected, users now receive more specific guidance on whether to use '--variant' or '--integration-id' flags.

### DIFF
--- a/packages/cli/lib/services/dryrun.service.ts
+++ b/packages/cli/lib/services/dryrun.service.ts
@@ -157,15 +157,30 @@ export class DryRunService {
             }
 
             // Priority for syncs and actions
+            let foundInProviderConfigKey: string | undefined;
             for (const script of [...integration.syncs, ...integration.actions]) {
                 if (script.name !== syncName) {
                     continue;
                 }
                 if (scriptInfo) {
-                    console.log(chalk.red(`Multiple integrations contain a script named "${syncName}". Please use "--integration-id"`));
+                    // TODO: Confirm: If exists another script with the same name in the same integration, it's a variant issue
+                    if (foundInProviderConfigKey === providerConfigKey) {
+                        console.log(
+                            chalk.red(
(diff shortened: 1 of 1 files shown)


packages/cli/lib/services/dryrun.service.ts
```diff
@@ -157,15 +157,30 @@ export class DryRunService {
...
+                                `Multiple variants of script "${syncName}" found in integration "${providerConfigKey}". Please specify which variant to use with "--variant"`
+                            )
+                        );
+                    } else {
+                        console.log(
+                            chalk.red(
+                                `Script "${syncName}" exists in multiple integrations (${foundInProviderConfigKey}, ${providerConfigKey}). Please specify which integration to use with "--integration-id"`
+                            )
+                        );
+                    }
```
TBonnin, line 179: I am not sure I  follow the logic. I don't think you can detect if variants exist unless querying the API

viictoo: Yeah, wasn't too sure about this one.
if we cant distinguish between the two should it just be an update to the existing error message to include variant?
```
Multiple integrations contain a script named "${syncName}". Please use "--integration-id or --variant to specify"
```

TBonnin: Maybe I don't really understand what the problem is with variants. Based on the code above it looks like if a variant is not specified the CLI default to the base one so I am not sure I understand the issue

viictoo: Let me dig into this further get what the root cause of this request was

PR #4007, opened 2025-05-07
feat(integrations): whitelist a node js package (dinero.js)

hassan254-prog:

- White list `dinero.js`

---

This PR adds support for the dinero.js library (a JavaScript library for working with monetary values) by adding it to the allowlist of packages that can be imported within the Nango integration environment. The PR includes both the main dinero.js package and the @dinero.js/currencies package, along with their TypeScript type definitions.

**Key Changes:**
• Added 'dinero.js' and '@dinero.js/currencies' to the allowed imports list
• Updated package dependencies in CLI and runner packages
• Added import handlers in dryrun.service.ts and exec.ts

**Affected Areas:**
• package-lock.json
• packages/cli/lib/services/dryrun.service.ts
• packages/runner/lib/exec.ts
• packages/cli/package.json
• packages/runner/package.json
• packages/cli/lib/services/compile.service.ts


TBonnin:
curious about why it is needed? Also the package doesn't seem to be maintained anymore

hassan254-prog:
> curious about why it is needed? Also the package doesn't seem to be maintained anymore

A customer has asked us to handle the conversion of monetary values to cents based on the given currency. I noticed that as well, but I couldn’t find a better package.

TBonnin:
> A customer has asked us to handle the conversion of monetary values to cents based on the given currency

Can you share more details about what's needed? A code example would be great

hassan254-prog:
> > A customer has asked us to handle the conversion of monetary values to cents based on the given currency
> 
> Can you share more details about what's needed? A code example would be great

This is a sample snippet from when I was playing around with it:
```
import dinero from 'dinero.js';
import currencies from '@dinero.js/currencies';

const { JPY } = currencies;
const fromMajorToDinero = (amount, currency) => {
  const exponent = currency.exponent;
  const amountInSubunits = Math.round(amount * 10 ** exponent);
  return dinero({ amount: amountInSubunits, currency });
};

const usdAmount = fromMajorToDinero(123.45, JPY);
console.log(usdAmount.getAmount());
```
The customer wants us to handle the conversion on our end and give the value in cents, having a standard way (value *100) without considering the currency type might not work with certain currencies.

bodinsamuel:
> curious about why it is needed? Also the package doesn't seem to be maintained anymore

I know the maintainer, I will ask

hassan254-prog:
@TBonnin, @bodinsamuel, sorry, but we've decided to bundle this with the integration instead of whitelisting it in Nango.

PR #4015, opened 2025-05-08
fix(scheduler): allow skipping group update if locked

TBonnin:
Adding an option to groups.upsert to allow skipping the update if the group is being modified by another transaction. When triggering action, the upsert is mostly used to set the last_task_added_at. Because it is happenning a lot in production this query is waiting on other transaction. This commit allow to skip the update since the last_task_added_at doesn't have to be 100% accurate. It is currently only used to cleanup unused groups

---

This PR adds a skipLocked option to the groups.upsert function to prevent transactions from waiting when another transaction is modifying the same group. The immediate scheduler now uses this option when updating last_task_added_at timestamps, avoiding unnecessary blocks since this value only needs to be approximate for group cleanup purposes.

### DIFF
--- a/packages/scheduler/lib/models/groups.ts
+++ b/packages/scheduler/lib/models/groups.ts
@@ -39,6 +39,9 @@ export async function upsert(
         key: string;
         lastTaskAddedAt: Date | null;
         maxConcurrency?: number | undefined;
+    },
+    opts?: {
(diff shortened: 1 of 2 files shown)


kaposke (approved):
Looks good as long as skipping locked is actually ok.

PR #4022, opened 2025-05-09
chore(scheduler): decrease frequency of monitor worker

TBonnin:
Expiring tasks isn't time critical. ie: it is ok if a task is not considered expired for an extra second.
Since expiring tasks is the most expensive SQL query, we are lowering its frequency to lower the pressure on the orchestrator db 
This commit also adds env vars for all the scheduler worker tick intervals so we can tweak the without having to make a code change

---

This PR reduces the scheduler's monitor worker frequency from 100ms to 1000ms (1s) to lower database pressure. The monitor worker handles task expiration checks which involves expensive SQL queries but isn't time-critical. The PR also adds environment variables for all scheduler worker tick intervals to enable runtime adjustments without code changes.

**Key Changes:**
• Decreased monitor worker frequency from 100ms to 1000ms
• Added environment variables for scheduler worker tick intervals
• Updated tests to account for the new timing configuration

**Affected Areas:**
• Scheduler monitoring system
• Task expiration checks
• Worker configuration

### DIFF
--- a/packages/utils/lib/environment/parse.ts
+++ b/packages/utils/lib/environment/parse.ts
@@ -62,6 +62,9 @@ export const ENVS = z.object({
     ORCHESTRATOR_DATABASE_URL: z.string().url().optional(),
     ORCHESTRATOR_DATABASE_SCHEMA: z.string().optional().default('nango_scheduler'),
     ORCHESTRATOR_DB_POOL_MAX: z.coerce.number().optional().default(50),
+    ORCHESTRATOR_MONITOR_TICK_INTERVAL_MS: z.coerce.number().optional().default(1000),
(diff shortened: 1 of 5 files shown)


packages/utils/lib/environment/parse.ts
```diff
@@ -62,6 +62,9 @@ export const ENVS = z.object({
     ORCHESTRATOR_DATABASE_URL: z.string().url().optional(),
     ORCHESTRATOR_DATABASE_SCHEMA: z.string().optional().default('nango_scheduler'),
     ORCHESTRATOR_DB_POOL_MAX: z.coerce.number().optional().default(50),
+    ORCHESTRATOR_MONITOR_TICK_INTERVAL_MS: z.coerce.number().optional().default(1000),
```
TBonnin, line 65: this is the only value that is changed. from 100ms to 1s

bodinsamuel (approved):

PR #4025, opened 2025-05-09
fix(slack): add an alternate path for access_token if undefined

hassan254-prog:

- Add an alternate path for `access_token` interpolation if not found at top-level, Nango's default for Oauth2.

---

This PR adds support for retrieving access tokens from alternative paths in OAuth2 responses. It specifically addresses an issue with Slack's implementation, where access tokens are sometimes returned in a nested structure at 'authed_user.access_token' rather than at the top level.

**Key Changes:**
• Added 'alternate_access_token_response_path' property to ProviderOAuth2 interface
• Modified parseRawCredentials to check alternate token paths when the primary path returns undefined
• Updated Slack provider configuration to use the new alternate path

**Affected Areas:**
• OAuth2 authentication flow
• Connection credential parsing service
• Slack provider configuration


packages/providers/providers.yaml
```diff
@@ -8989,6 +8989,7 @@ slack:
     auth_mode: OAUTH2
     authorization_url: https://slack.com/oauth/v2/authorize
     scope_separator: ','
+    alternate_access_token_response_path: authed_user.access_token
```
bodinsamuel, line 9049: sugg: why not `token_response_access_token` to follow `token_response_metadata?

bodinsamuel (approved):

packages/providers/providers.yaml
hassan254-prog: I believe `token_response_access_token` could be somehow misleading because this is more of an alternative path rather than than the primary path that is currently used by Nango.

khaliqgant (approved):

PR #4028, opened 2025-05-12
fix(auth): can't detect popup with security headers

bodinsamuel:
## Changes

- Auth frontend: we cant' reliably detect popup
e.g: linear is using `Cross-Origin-Opener-Policy: same-origin`. I think it's a mistake on their end but it's breaking auth flow.

---

This PR fixes an issue where the auth popup couldn't be reliably detected when third-party services use strict security headers like `Cross-Origin-Opener-Policy: same-origin`. The fix changes how the popup window's closed state is checked.

### DIFF
--- a/packages/connect-ui/src/views/Go.tsx
+++ b/packages/connect-ui/src/views/Go.tsx
@@ -212,7 +212,7 @@ export const Go: React.FC = () => {
             }
 
             telemetry('click:connect');
-            setLoading(true);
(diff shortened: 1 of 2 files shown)


TBonnin (commented):
the popup closing detection is probably one of the biggest time sink in the history of Nango 😛

TBonnin (approved):

Issue #1371, opened 2023-12-04
Add prompt=consent for Salesforce OAuth flow


rguldener:
Currently, we don't set prompt=consent for the Salesforce OAuth login. This makes it impossible for a user to connect two separate Salesforce instances to the same OAuth app, as it auto-approves and redirects if access was previously granted.
[See Salesforce docs](https://help.salesforce.com/s/articleView?id=sf.remoteaccess_oauth_web_server_flow.htm&type=5#:~:text=Specifies%20how%20the%20authorization%20server%20prompts%20the%20user%20for%20reauthentication%20and%20reapproval.%20Salesforce%20supports%20these%20values.).

Desired behavior:
- User should always have to confirm the connection
- User should always be able to pick the account to connect (if the user is logged into multiple Salesforce instances)

We should add `prompt=consent` to the authorization step and verify it now asks for consent every time.
We may also need to add `login` (together with `consent`), depending on how Salesforce handles the implementation.

Feel free to ask me if you have questions about the desired behavior!

Issue #1040, opened 2023-09-21
Auth params: override from `nango.auth`


bastienbeurier:
Follows this resolved issue: [Ask consent on Microsoft re-auth ](https://github.com/NangoHQ/nango/issues/951)

From the [community](https://nango-community.slack.com/archives/C04ABR352H0/p1695231629754309?thread_ts=1693415564.566349&cid=C04ABR352H0): 

> The plot thickens. Just got off the phone with a customer’s IT, they did admin consent in Azure, but their users were still continuing to be prompted for admin consent when attempting OAUTH as a non-admin user.
> Digging deeper, apparently, setting prompt=consent is frowned upon for O365. There’s actually documentation on this, buried inside Azure Portal. I attached the full doc as a PDF below, but a couple quotes:
> Note: Configuring an application with prompt=consent is not a best practice for developers. Azure AD might be receiving prompt=consent in authentication attempts, prompting an admin consent loop. Ask the developer to remove the prompt=consent property.
> The Zoom application has a known issue where users are prompted with an admin consent request while signing in to the application, even when users have all permissions assigned in Azure. If you disable the Office 365 users can consent to enterprise applications accessing company data on their behalf option on the Zoom admin portal, then users won’t be prompted for admin consent.
> So maybe Zoom defaults to prompt=consent too, but then has a admin configuration that removes it for orgs that have certain Azure configurations? I’d be down with defaulting to prompt=consent, but being able to flag prompt=consent off, depending on my client’s Azure configuration?

Issue #951, opened 2023-09-01
Ask consent on Microsoft re-auth


bastienbeurier:
From the [community](https://nango-community.slack.com/archives/C04ABR352H0/p1693415564566349): 

> Sorry if this question has been asked before: we recently had a user connect with O365, but remove all but Read scope, which is insufficient for our features. If a user doesn’t select the necessary Oauth scopes for an integration (in this case, Office 365 / teams), it doesn’t seem like they get another chance - re-connecting skips the scope selection wizard. Is there a way to force the connection setup flow to re-prompt for scope selection / REQUIRE that all scopes are approved? looks like this was answered in the thread immediately above - not possible, for very good security reasons, so the best we can do is handle the outcome gracefully

PR #4040, opened 2025-05-13
fix(microsoft): remove prompt=consent

khaliqgant:
Remove prompt for microsoft as it isn't required since permission will be prompted if consent hasn't been previously granted. [ref1](https://learn.microsoft.com/en-us/entra/identity-platform/application-consent-experience), [ref2](https://learn.microsoft.com/en-us/entra/identity-platform/howto-convert-app-to-be-multi-tenant)

## Background
This parameter was added with https://github.com/NangoHQ/nango/pull/952 which was due to #951, however in that scenario we should recommend our users to explicitly pass `authorization_params: { prompt: 'consent'}` like we recommended [here](https://github.com/NangoHQ/nango/issues/1040)

There is also a related #1371 and fix in #1374 however, upon testing Microsoft does allow you to select an account if there are multiple without the prompt param.

---

This PR removes the default `prompt=consent` parameter from Microsoft OAuth providers and adds `.default` to the default scopes. Microsoft's OAuth flow automatically prompts for permission when needed, making the explicit consent prompt unnecessary by default. Users can still force consent when needed by explicitly setting the parameter.

**Key Changes:**
• Removed `prompt=consent` from default authorization parameters in providers.yaml
• Added `.default` to the default scopes
• Created a shared Microsoft API gotchas snippet to centralize documentation
• Updated docs to show how to explicitly force consent when needed

**Affected Areas:**
• Microsoft OAuth provider configurations
• Documentation for multiple Microsoft integration providers

### DIFF
--- a/docs-v2/integrations/all/microsoft-ads.mdx
+++ b/docs-v2/integrations/all/microsoft-ads.mdx
@@ -6,6 +6,7 @@ sidebarTitle: Microsoft Ads
 import Overview from "/snippets/overview.mdx"
 import PreBuiltTooling from "/snippets/generated/microsoft-ads/PreBuiltTooling.mdx"
 import PreBuiltUseCases from "/snippets/generated/microsoft-ads/PreBuiltUseCases.mdx"
+import APIGotchas from "/snippets/microsoft-api-gotchas.mdx"
 
 <Overview />
 <PreBuiltTooling />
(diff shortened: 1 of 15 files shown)


bodinsamuel (approved):

khaliqgant:
Thanks for the quick review @bodinsamuel. I hadn't requested reviews yet because I was still updating documentation – which is now updated!

bodinsamuel (approved):

PR #4044, opened 2025-05-14
feat: adding GET /action/:id endpoint

TBonnin:
GET /action/:id allows to retrieve the output of an action `id` is tasks retryKey

if action (including retries) isn't completed yet, the endpoint returns a 404: `{ error: 'not_found', message: '...' }`
if action (or retries) is successful, the endpoint returns the same response as if it was trigger synchronously
if action failed, it returns a 500 or 424, the same as if it has been triggerred synchronously:
```
{ "error":
   {
      "message":"An internal error occurred during the script execution",
      "code":"script_internal_error",
      "payload":{
         "message":"...", ...
      }
   }
}
```

# How to test
The endpoint to trigger the action async and return the `id` doesn't exist yet. That's my next task.
To test locally:
1. trigger an action
2. get the  `retryKey` from the db
3. call the endpoint with the retryKey as the `id`

---

This PR adds a new endpoint `GET /action/:id` that allows retrieving the output of an asynchronous action using a task's retry key as the ID parameter. The endpoint follows the same error handling patterns as synchronous actions, returning appropriate status codes based on the action's completion state.

**Key Changes:**
• Added new GET /action/:id endpoint for retrieving action outputs
• Extended OrchestratorClient to support getOutput for actions by retry key
• Enhanced scheduler to support searching tasks by retryKey and ownerKey
• Added retryKey to PostImmediate response

**Affected Areas:**
• Orchestrator client and routes
• Server controllers and routes
• Task search and retry functionality

### DIFF
--- a/packages/server/lib/controllers/action/getAsyncActionResult.ts
+++ b/packages/server/lib/controllers/action/getAsyncActionResult.ts
@@ -1,0 +1,38 @@
+import { z } from 'zod';
+import type { GetAsyncActionResult } from '@nangohq/types';
+import { asyncWrapper } from '../../utils/asyncWrapper.js';
+import { getOrchestrator } from '../../utils/utils.js';
+import { zodErrorToHTTP } from '@nangohq/utils';
+import { errorManager } from '@nangohq/shared';
+
+const orchestrator = getOrchestrator();
+
+const paramValidation = z
+    .object({
+        id: z.string().uuid()
+    })
+    .strict();
+
+export const getAsyncActionResult = asyncWrapper<GetAsyncActionResult>(async (req, res) => {
+    const paramValue = paramValidation.safeParse(req.params);
+    if (!paramValue.success) {
+        res.status(400).send({ error: { code: 'invalid_uri_params', errors: zodErrorToHTTP(paramValue.error) } });
+        return;
+    }
+
+    const { environment } = res.locals;
+    const retryKey = paramValue.data.id;
+    const result = await orchestrator.getActionOutput({ retryKey, environmentId: environment.id });
+
+    if (result.isErr()) {
+        errorManager.errResFromNangoErr(res, result.error);
+        return;
+    }
+
+    if (result.value === null) {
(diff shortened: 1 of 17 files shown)


packages/types/lib/action/api.ts
```diff
@@ -0,0 +1,10 @@
+import type { Endpoint } from '../api';
+
+export type GetAsyncActionResult = Endpoint<{
+    Method: 'GET';
+    Path: `/action/:id`;
+    Params: {
+        id: string;
+    };
+    Success: Record<string, any>;
```
bodinsamuel, line 12: Can you just add a comment that says it's not true and why ☺️

packages/server/lib/controllers/action/getAsyncActionResult.ts
```diff
@@ -0,0 +1,38 @@
...
+        errorManager.errResFromNangoErr(res, result.error);
+        return;
+    }
+
+    if (result.value === null) {
+        res.status(404).json({ error: { code: 'not_found', message: `No action '${retryKey}' found` } });
+        return;
+    }
+
+    res.status(200).json(result.value as GetAsyncActionResult['Success']);
```
bodinsamuel, line 37: I still think it would have been a great time to fix this and not a big breaking change because it's a new client method anyway.
The opportunity to always return 200 no matter the action status, and eventually says that it errored in the payload

bodinsamuel (approved):

packages/server/lib/controllers/action/getAsyncActionResult.ts
TBonnin: I am "not a fan" (read: I really dislike) how script error handling is currently done. 
On one hand I think it would be better not to report script errors the same way as internal errors. 
I am not against doing what you propose but it would make the response inconsistent with what we currently return for actions. I don't think we can have sync and async actions not return the same response and we cannot break current actions response neither. It seems to me the only "acceptable" option is to carry on with the legacy behavior, even if it is not ideal.

PR #4048, opened 2025-05-15
feat: trigger async actions

TBonnin:
Modifying trigger action endpoint to support async actions.

For now the only thing that differentiate async actions in the logs is the `x-async: true` header in the operation payload. We might want to show something a bit more visual maybe. cc @bastienbeurier 

How to test:
1. Trigger async action (you can also use the custom endpoint instead of `/action/trigger`)
```
curl --request POST \
  --url http://localhost:3003/action/trigger \
  --header 'Authorization: Bearer ...' \
  --header 'Connection-Id: ...' \
  --header 'X-Async: true' \
  --header 'X-Max-Retries: 1' # optional
  --data '{
     "action_name": "...", "input": { ... }
   }'
```
2. fetch result 
```
curl --request GET -v \
    --url http://localhost:3003/action/ID_FROM_PREVIOUS_CALL \
    --header 'Authorization: Bearer a...' \
    --header 'Content-Type: application/json'
```
3. Check the logs in the dashboard 

---

This PR enhances the action trigger system to support asynchronous execution with configurable retry capability. Actions can now be triggered with the `X-Async: true` header to enable fire-and-forget execution, allowing clients to proceed without waiting for completion while still being able to check results later via a status endpoint.

**Key Changes:**
• Added async action execution flow with X-Async header option
• Implemented retry capability with X-Max-Retries header (0-5)
• Created endpoint to check async action status
• Moved logging and metrics to action execution handler for better async support

**Affected Areas:**
• Orchestrator client interface
• Action execution flow
• HTTP API endpoints
• Task state tracking and logging


packages/jobs/lib/execution/action.ts
```diff
@@ -129,7 +129,18 @@ export async function startAction(task: TaskAction): Promise<Result<void>> {
...
+    const task = await setTaskSuccess({ taskId, output });
+
+    const logCtx = logContextGetter.get({ id: nangoProps.activityLogId, accountId: nangoProps.team.id });
+
+    void logCtx.info(`The action was successfully run${formatAttempts(task)}`, {
+        action: nangoProps.syncConfig.sync_name,
+        connection: nangoProps.connectionId,
+        integration: nangoProps.providerConfigKey
+    });
+    void logCtx.success();
```
TBonnin, line 141: move this logging from the sync controller trigger endpoint handler since for async actions a response doesn't mean the action has been run.

packages/jobs/lib/execution/action.ts
```diff
@@ -129,7 +129,18 @@ export async function startAction(task: TaskAction): Promise<Result<void>> {
...
+
+    metrics.increment(metrics.Types.ACTION_SUCCESS);
```
TBonnin, line 143: same for the metrics. I only kept the action duration in the controller since it measure the full action call duration

packages/orchestrator/lib/clients/types.ts
```diff
@@ -70,6 +71,7 @@ interface TaskCommonFields {
     groupKey: string;
     state: TaskState;
     attempt: number;
+    attemptMax: number;
```
TBonnin, line 74: exposing the attemptMax so we can show it in the logs and decide to fail the log operation if last attempt

packages/orchestrator/lib/routes/v1/postImmediate.ts
```diff
@@ -110,7 +110,7 @@ const handler = (scheduler: Scheduler) => {
             res.status(500).json({ error: { code: 'immediate_failed', message: task.error.message } });
             return;
         }
-        res.status(200).json({ taskId: task.value.id, retryKey: task.value.retryKey });
+        res.status(200).json({ taskId: task.value.id, retryKey: task.value.retryKey! });
```
TBonnin, line 113: a freshly created task always has a retryKey. It is only null for tasks created before the introduction of retryKey

packages/server/lib/controllers/sync.controller.ts
```diff
@@ -180,6 +183,13 @@ class SyncController {
                 return;
             }
 
+            if (async && (retryMax < 0 || retryMax > 5)) {
```
TBonnin, line 186: arbitrary number. I am happy to change to something else

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -193,27 +214,6 @@ export class Orchestrator {
...
-            errorManager.report(err, {
-                source: ErrorSourceEnum.PLATFORM,
-                operation: LogActionEnum.SYNC_CLIENT,
-                environmentId: connection.environment_id,
-                metadata: {
-                    actionName,
-                    connectionDetails: JSON.stringify(connection),
-                    input
-                }
-            });
```
TBonnin, line 214: I think this is redundant. we already set the span error tag below and attach meaningful tags at the top of the function

bodinsamuel (approved):
Not tested, looks okay.
Retry could have been done separately from this imo

packages/server/lib/controllers/sync.controller.ts
kaposke: sounds good

TBonnin:
> Retry could have been done separately from this imo

indeed. I got carried away

kaposke (approved):
Looks good and works as expected from my tests 👍

PR #4049, opened 2025-05-15
fix: fix processor flacky test

TBonnin:
This tasks processor flacky test has been bothering me for a loooong time. Finally managed to reproduce and fix it.

The test relied on the processor's internal queue being empty to determine if all tasks were processed, which is flawed. The queue only tracks dequeued tasks, not whether all pending tasks have been processed.

This fix implements an explicit count to verify the handler has been called the expected number of times before completing the test. 
A alternative fix would be to start the processor after all the tasks have been added so all the tasks are dequeued at once but it is less representative of real-world usage.

---

This PR fixes a flaky test in the tasks processor by replacing an unreliable queue size check with an explicit counter. The test now accurately counts when handler functions are executed rather than relying on the processor's internal queue state.

### DIFF
--- a/packages/orchestrator/lib/clients/processor.integration.test.ts
+++ b/packages/orchestrator/lib/clients/processor.integration.test.ts
@@ -67,20 +67,23 @@ describe('OrchestratorProcessor', () => {
 });
 
 async function processN(handler: (task: OrchestratorTask) => Promise<Result<void>>, groupKey: string, n: number) {
+    let processCount = 0;
     const processor = new OrchestratorProcessor({
-        handler,
+        handler: (task: OrchestratorTask) => {
+            processCount++;
+            return handler(task);
+        },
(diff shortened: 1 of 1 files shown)


packages/orchestrator/lib/clients/processor.integration.test.ts
```diff
@@ -67,20 +67,23 @@ describe('OrchestratorProcessor', () => {
...
     }
-    // Wait so the processor can process all tasks
-    while (processor.queueSize() > 0) {
+    // wait for all tasks to be processed
+    while (processCount < n) {
         await setTimeout(100);
     }
-    await setTimeout(500);
-
+    processor.stop();
```
TBonnin, line 86: not linked to the fix but it is cleaner to stop the processor at the end

kaposke (approved):
Awesome!

PR #4053, opened 2025-05-15
chore: update Facebook API endpoint from v15 (deprecated) to v22 (latest)

wub:
The Facebook provider is referencing [v15](https://developers.facebook.com/docs/graph-api/changelog/version15.0/), which was deprecated in November 2024.

This updates the auth call to the latest version - [v22](https://developers.facebook.com/docs/graph-api/changelog/version22.0).

> Released September 15, 2022 | Available until November 20, 2024

---

This PR updates the Facebook provider's OAuth endpoints from the deprecated API version v15 to the latest version v22 within the providers configuration file. The change ensures continued compatibility as v15 is no longer supported.


bodinsamuel:
@khaliqgant do you know the implication of doing that? ☺️

khaliqgant:
> @khaliqgant do you know the implication of doing that? ☺️

I think this should be ok as it is _just_ the auth and token URL. We should test that the auth call still works however. @hassan254-prog think you could quickly verify this in the next couple of days?

hassan254-prog:
> > @khaliqgant do you know the implication of doing that? ☺️
> 
> I think this should be ok as it is _just_ the auth and token URL. We should test that the auth call still works however. @hassan254-prog think you could quickly verify this in the next couple of days?

Tested locally and works as expected 👍 .

hassan254-prog (approved):

bodinsamuel (approved):

PR #4054, opened 2025-05-15
feat: run async action sequentially per environment

TBonnin:
async actions are not inline so they don't have to be run as quickly as possible. We are therefore running them sequentially per environment to minimize load and concurrency issues like rate limits

---

PR limits async action concurrency to 1 per environment by introducing an environment-scoped group key (`action:environment:<id>`) with `maxConcurrency: 1`. It also relaxes the `createdToStarted` timeout for those async actions from 30 s to 24 h, allowing them to wait in the queue longer before expiring.

**Key Changes:**
• executeActionAsync(): timeoutSettingsInSecs.createdToStarted changed from 30 s to 24 h (+ inline comment).
• shared orchestrator client: group key updated to `action:environment:<envId>` and `maxConcurrency` set to 1 (was 0/unlimited) so only one async action runs at a time per environment.

**Affected Areas:**
• packages/orchestrator/lib/clients/client.ts (timeout defaults)
• packages/shared/lib/clients/orchestrator.ts (task grouping & concurrency)

### DIFF
--- a/packages/orchestrator/lib/clients/client.ts
+++ b/packages/orchestrator/lib/clients/client.ts
@@ -239,7 +239,7 @@ export class OrchestratorClient {
             ...rest,
             retry: { count: props.retry?.count || 0, max: props.retry?.max || 0 },
             timeoutSettingsInSecs: {
-                createdToStarted: 30,
+                createdToStarted: 24 * 60 * 60, // async action must starts within 24h after being created
                 startedToCompleted: 15 * 60,
                 heartbeat: 5 * 60
             },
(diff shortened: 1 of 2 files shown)


packages/orchestrator/lib/clients/client.ts
```diff
@@ -239,7 +239,7 @@ export class OrchestratorClient {
             ...rest,
             retry: { count: props.retry?.count || 0, max: props.retry?.max || 0 },
             timeoutSettingsInSecs: {
-                createdToStarted: 30,
+                createdToStarted: 24 * 60 * 60, // async action must starts within 24h after being created
```
TBonnin, line 242: I choose 24h as the arbitrary limit for an async action to expire if it is not processed. wdyt?

bodinsamuel (approved):

PR #4057, opened 2025-05-16
feat(frontend): match sdk error types casing

hassan254-prog:

- Match sdk error types casing 

---

This PR standardizes all frontend SDK authentication error type string literals from camelCase to snake_case for consistency with industry conventions and future maintainability. Updates include type definitions, error construction and handling, UI components, and documentation to exclusively use the new casing across both code and documentation.

**Key Changes:**
• Changed AuthErrorType union members from camelCase (e.g., 'missingAuthToken') to snake_case (e.g., 'missing_auth_token') in packages/frontend/lib/types.ts
• Updated all error constructions and handlers in packages/frontend/lib/index.ts to use new snake_case values
• Modified Connect UI logic in packages/connect-ui/src/views/Go.tsx to recognize the new error types
• Rewrote error references in the frontend SDK documentation (docs-v2/reference/sdks/frontend.mdx) to align with new casing

**Affected Areas:**
• TypeScript error typing and handling (packages/frontend/lib/types.ts, packages/frontend/lib/index.ts)
• Frontend UI logic for error handling (packages/connect-ui/src/views/Go.tsx)
• SDK documentation (docs-v2/reference/sdks/frontend.mdx)


bodinsamuel (commented):
unfortunately this would be a breaking changes for a lot of customers

khaliqgant:
> unfortunately this would be a breaking changes for a lot of customers

Indeed. We could do our "normal" deprecation pattern.

1. Make these additions only
2. Remove the ones we're replacing from any documentation
3. Add a deprecation notice to the ones that we will no longer support
4. Set a deprecation date

hassan254-prog:
@khaliqgant  and  @bodinsamuel , I suppose there's no way to deprecate individual string literals within a union type for `AuthErrorType`. To take a cleaner approach, I thought it best to deprecate `AuthError` entirely and replace it with a new `AuthErrorV2`, which now uses enums for better future-proofing. If needed, we can deprecate individual auth error types within the enum moving forward.

packages/frontend/lib/index.ts
```diff
@@ -57,6 +61,20 @@ export class AuthError extends Error {
     }
 }
 
+export class AuthErrorV2 extends Error {
+    legacy?: AuthError;
+    type: AuthErrorTypeV2;
+
+    constructor(message: string, type: AuthErrorTypeV2, legacyType?: AuthErrorType) {
```
TBonnin: a simpler API would be in my opinion to no pass the legacyType and internally map the AuthErrorTypeV2 to the legacy one (since the legacy one are not gonna evolve anymore). wdyt?

bodinsamuel (commented):
Deprecated doesn't raise any alarm in typescript so I wouldn't use that as a safety net. Not entirely sure what does `AuthErrorV2` bring since it's still not retro-compatible and still breaking.

If we really want to achieve this, I only see two options:
- opt-in for new errors and have the two system running together
- announce breaking change and just go with it

I would just go for later if that's an accepted risk. So maybe your first option was fine but I'm still not sure why we bother tbh

---

Tegarding enum I would try not to use them more, since even Microsoft acknowledge they are not good. The only annoying part is `@deprecated` but you can use a regular object or constant [playground](https://www.typescriptlang.org/play/?ssl=10&ssc=28&pln=4&pc=1#code/FAjGHsDsGcBcAICG8C88Dkj0G5gHoAqA+AAQBMBTABwCcLRFYKz4i9gIYEAjVDbnMFgBPKhXgAFGuCoAVUeLTwRY8ADMk8AD7KF6+N1wcoceADdEAGwCuFRAC5J0uQr6JcnUxZsVujqTLyYnyGQA)

hassan254-prog:
> Deprecated doesn't raise any alarm in typescript so I wouldn't use that as a safety net. Not entirely sure what does `AuthErrorV2` bring since it's still not retro-compatible and still breaking.
> 
> If we really want to achieve this, I only see two options:
> 
> * opt-in for new errors and have the two system running together
> * announce breaking change and just go with it
> 
> I would just go for later if that's an accepted risk. So maybe your first option was fine but I'm still not sure why we bother tbh
> 
> Tegarding enum I would try not to use them more, since even Microsoft acknowledge they are not good. The only annoying part is `@deprecated` but you can use a regular object or constant [playground](https://www.typescriptlang.org/play/?ssl=10&ssc=28&pln=4&pc=1#code/FAjGHsDsGcBcAICG8C88Dkj0G5gHoAqA+AAQBMBTABwCcLRFYKz4i9gIYEAjVDbnMFgBPKhXgAFGuCoAVUeLTwRY8ADMk8AD7KF6+N1wcoceADdEAGwCuFRAC5J0uQr6JcnUxZsVujqTLyYnyGQA)

`AuthErrorV2` introduces the correct casing convention (`snake_case`) that we’ll be using moving forward. It includes support for future deprecation scenarios. At the same time, we’re maintaining support for the original `AuthError` to avoid breaking changes for existing customers and to give them time to migrate to `AuthErrorV2`. The plan is to eventually remove `AuthError`, but new users will be guided to use `AuthErrorV2` via the documentation. This is my first time introducing such a breaking change for users in Nango, I will need proper guidance ☺️ .

bodinsamuel:
I understand the new class it's just that it's a breaking change, so there is no need to introduce a new layer if it's also breaking. 
What will happen if we merge:
- People staying before this version = no problem
- People upgrading = they have to use AuthErrorV2 because AuthError is no longer thrown

If you want to upgrade without breaking change you need to do the opposite, by no changing type and introducing a new type
```ts
class AuthError {
 type,
 typeV2
}
```

But if you do that, that nobody will upgrade since no TS errors will be raised. So that's why I'm saying, maybe the only way is just to create a breaking change and be done with it.

hassan254-prog:
> I understand the new class it's just that it's a breaking change, so there is no need to introduce a new layer if it's also breaking. What will happen if we merge:
> 
> * People staying before this version = no problem
> * People upgrading = they have to use AuthErrorV2 because AuthError is no longer thrown
> 
> If you want to upgrade without breaking change you need to do the opposite, by no changing type and introducing a new type
> 
> ```ts
> class AuthError {
>  type,
>  typeV2
> }
> ```
> 
> But if you do that, that nobody will upgrade since no TS errors will be raised. So that's why I'm saying, maybe the only way is just to create a breaking change and be done with it.

Aaah, that makes perfect sense, thanks! I think going for it with the breaking changes is the only way. I've reverted to the initial commit of the PR.

TBonnin:
what's the plan to release this breaking change?

hassan254-prog:
> what's the plan to release this breaking change?

Announce the breaking change with the newer version of the frontend SDK once it is merged and the version is bumped.

TBonnin (approved):

hassan254-prog:
@bodinsamuel , could you please help release this 🙏 .

PR #4066, opened 2025-05-19
fix(email): sanitize html

bodinsamuel:
## Changes

- Sanitize user provided content in the email
NB: the npm packages seems kindof weird but it has +25M downloads :D

---

Adds the ‘he’ library and wraps all user-supplied strings (name, account name, etc.) in outgoing emails with he.encode(). package.json and lock file updated accordingly.


TBonnin (commented):
I know html sanitization isn't something that required constant update but the package is 7 years old :( 
What about https://www.npmjs.com/package/sanitize-html?

bodinsamuel:
> I know html sanitization isn't something that required constant update but the package is 7 years old :(
> What about https://www.npmjs.com/package/sanitize-html?

I'm not that worried, it has zero prod dependency. The other package has prod dependencies and much more features that we will not use but that can causes bug. I'm sure at some point will move to a more modern stack like react-email or something like this. Until then, I feel like it's not a liability imo

TBonnin:
> it has zero prod dependency. 

I don't know what you mean by zero prod dependency. Anyway, your call 😁

TBonnin (approved):

bodinsamuel:
> I don't know what you mean by zero prod dependency. Anyway, your call 😁

I mean it has no dependency that would be a liability later and no know security issues. So the fact that it has no update in 7years is fine in that regard

PR #4067, opened 2025-05-19
fix(save-responses): fix metadata and header parsing

khaliqgant:
1. Don't add `base` if there is no variant as it breaks how we pull in the data
2. Don't overwrite stubbed metadata if we pass in a stub

---

Small patch to response-saving logic. It stops auto-creating the "base" variant folder, adds "retry-on" to header filters, and ensures stubbed metadata supplied by dry-run is not overwritten.


packages/cli/lib/services/response-saver.service.ts
```diff
@@ -88,18 +92,20 @@ export function onAxiosRequestFulfilled({
...
+            });
+        }
 
         return response;
     }
 
     const requestIdentity = computeConfigIdentity(response.config);
 
-    const syncSubDir = syncVariant ? `${syncName}/${syncVariant}` : syncName;
+    const syncSubDir = syncVariant && syncVariant !== BASE_VARIANT ? `${syncName}/${syncVariant}` : syncName;
```
TBonnin, line 108: does that mean we cannot snapshot the responses in variant sync? Maybe it is not a problem

khaliqgant: I haven't tested it actually so I'm not sure we can or not. We should support it if it doesn't work but I think we should cross that path later.

TBonnin: up to you. maybe add a TODO to mention it'd be a limitation and variant are not supported

TBonnin (approved):

PR #4068, opened 2025-05-19
fix: mcp sdk installation

kaposke:
---

Removes the SDK from the monorepo root and pins @modelcontextprotocol/sdk at v1.11.4 inside packages/server, resolving install issues and aligning with the repo’s version-pinning convention.

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -23,7 +23,7 @@
         "npm": ">=6.14.11"
     },
     "dependencies": {
-        "@modelcontextprotocol/sdk": "^1.11.4",
+        "@modelcontextprotocol/sdk": "1.11.4",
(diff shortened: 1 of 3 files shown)


packages/server/package.json
```diff
@@ -23,7 +23,7 @@
         "npm": ">=6.14.11"
     },
     "dependencies": {
-        "@modelcontextprotocol/sdk": "^1.11.2",
+        "@modelcontextprotocol/sdk": "^1.11.4",
```
TBonnin: as you can see in the reset of the file we tend to pin to a specific version

TBonnin (approved):

bodinsamuel (approved):

PR #4081, opened 2025-05-21
fix(sdk): better handle 401 false positive

bodinsamuel:
## Changes

- better handle 401 false positive
Compare credentials, if they are not different it's not going to retry more

Screenshot 2025-05-21 at 10 26 37

---

This pull request updates the SDK's proxy request logic to better manage repeated 401 Unauthorized responses. By comparing credentials before and after cache clearing, the system now halts redundant retry attempts when credentials remain unchanged, reducing false positives for authentication errors.


TBonnin (commented):
logic is pretty hard to follow with those hook functions and mutable variables.
If it was me I would actually force refresh the connection in the onError and check there if credentials are identical. Maybe adding a force options param to this.getConnection so it can use the same code while ensuring it is not using the cached connection

bodinsamuel:
> logic is pretty hard to follow with those hook functions and mutable variables.

Agreed it's not the best, but onError happens before the wait, so credentials could be refreshed during the wait, creating another issue. Alternatively, doing the check on getConnection is not possible because it's happening after waiting and is also not made to kill the proxy (unless you throw which is annoying).
If you want to, feel free to suggest a better solution. For the ref, this is what happens:

- 1
  - getConnection  -> refresh or cache
  - onError.              -> get 401, clean cache
    - wait or stop      -> wait for N seconds
- 2
  - getConnection. -> refresh
  - onError               -> get 2n 401, stops

TBonnin (commented):
let's fix it but I am sure when I am gonna re-read this code in 3 months I am gonna be super confused 😁

TBonnin (approved):

PR #4085, opened 2025-05-21
fix(save-responses): don't remove content type if not the default application/json & remove dupes

khaliqgant:
- if we're explicitly passing in the content type include that in the response saver
- Dedupe headers since they could be repeated with the `nango-proxy` prefix
- Some API's allow limit to be a string as well

 

---

This PR improves the response-saver's handling of headers and corrects typing for pagination limits across several core packages. The response-saver now only removes the 'content-type' header if its value is exactly 'application/json' and properly deduplicates headers, including those prefixed with 'nango-proxy-'. Additionally, the pagination 'limit' parameter now supports both string and number types in all related type declarations and logic, ensuring reliable pagination regardless of the provided type.

**Key Changes:**
• Updates response-saver.service.ts to deduplicate headers and retain explicit non-default content-types.
• Allows 'limit' in pagination configuration to be a string or number in types (types, models.d.ts, api.ts, snapshot).
• Adjusts runner-sdk pagination logic (paginate.service.ts) to convert 'limit' to a number where needed, preventing type-related paging errors.

**Affected Areas:**
• CLI response saver service (header filtering/deduplication)
• Pagination typing and logic in runner-sdk/types/models
• Snapshot compilation


packages/cli/lib/services/response-saver.service.ts
```diff
@@ -166,9 +165,20 @@ function computeConfigIdentity(config: AxiosRequestConfig): ConfigIdentity {
...
-            .filter(([key]) => !FILTER_HEADERS.includes(key.toLowerCase()));
+            .filter(([key]) => !FILTER_HEADERS.includes(key.toLowerCase()))
+            // Filter out content-type if value is exactly application/json (case-insensitive)
+            .filter(([key, value]) => key.toLowerCase() !== 'content-type' || String(value).toLowerCase() !== 'application/json')
+            .filter(([key]) => {
+                const lowerKey = key.toLowerCase();
+                if (seen.has(lowerKey)) return false;
+                seen.add(lowerKey);
+                return true;
+            });
```
TBonnin, line 186: I feel like this could be simplified
```
const filteredHeaders = Object.entries(config.headers)
  .map<[string, string]>(([key, value]) => [
    key.toLowerCase().startsWith('nango-proxy-') ? key.slice(12) : key,
    String(value)
  ])
  .filter(([key, value]) => {
    const lowerKey = key.toLowerCase();
    
    // Skip if already seen
    if (seen.has(lowerKey)) return false;
    seen.add(lowerKey);
    
    // Skip filtered
    if (FILTER_HEADERS.includes(lowerKey)) return false;
    
    // Skip application/json
    if (lowerKey === 'content-type' && value.toLowerCase() === 'application/json') return false;
    
    return true;
  });
```

TBonnin (approved):

packages/cli/lib/services/response-saver.service.ts
khaliqgant: yeah, agreed, thanks for the suggestion. Will update and merge

bodinsamuel (approved):

PR #4087, opened 2025-05-21
chore(billing): add dimensions to billing events 

TBonnin:
Adding properties like environmentId, connectionId, etc... to billing events.
Even though those properties are not being used for billable metrics
they might be used in the future and/or to display more granular data to
customers
Note: for billable_connections, since we are already sending the
aggregated value, we cannot easily split it to add dimensions. I
therefore haven't added any dimension to it

---

This PR enhances billing event tracking by including additional dimensional properties (such as environmentId, providerConfigKey, connectionId, syncId, model, and actionName) in the data attached to 'billable_actions' and 'monthly_active_records' events. These properties do not affect current billing but are intended to support future use cases and more granular analytics for customers. The 'billable_connections' billing event remains unchanged due to aggregation constraints.

**Key Changes:**
• Augmented 'billable_actions' event properties to include environmentId, providerConfigKey, connectionId, and actionName in sync.controller.ts.
• Expanded 'monthly_active_records' event payload with environmentId, providerConfigKey, connectionId, syncId, and model in records.ts.
• Broadened BillingIngestEvent and BillingMetric type definitions in billing/types.ts to support richer property types.

**Affected Areas:**
• packages/server/lib/controllers/sync.controller.ts
• packages/persist/lib/records.ts
• packages/types/lib/billing/types.ts

### DIFF
--- a/packages/persist/lib/records.ts
+++ b/packages/persist/lib/records.ts
@@ -157,7 +157,7 @@ export async function persistRecords({
         const mar = new Set(summary.billedKeys).size;
 
         if (plan) {
-            billing.add('monthly_active_records', mar, { accountId });
+            billing.add('monthly_active_records', mar, { accountId, environmentId, providerConfigKey, connectionId, syncId, model });
         }
 
         metrics.increment(metrics.Types.BILLED_RECORDS_COUNT, mar, { accountId });
--- a/packages/server/lib/controllers/sync.controller.ts
+++ b/packages/server/lib/controllers/sync.controller.ts
@@ -249,7 +249,14 @@ class SyncController {
                 }
 
                 if (plan) {
-                    billing.add('billable_actions', 1, { accountId: account.id, idempotencyKey: logCtx.id });
+                    billing.add('billable_actions', 1, {
+                        accountId: account.id,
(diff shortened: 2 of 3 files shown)


TBonnin:
@bastienbeurier For billable connections since we do aggregate ourselves, if we want to add more dimensions like environmentnor provider we would need to break the events down and re-aggregate them in Orb. I have decided not to add dimensions to billable connections events then to keep the metric calculation simple in Orb. Let me know if it is an absolute requirement to add more dimensions to billable connections and I will find a solution

bastienbeurier:
> @bastienbeurier For billable connections since we do aggregate ourselves, if we want to add more dimensions like environmentnor provider we would need to break the events down and re-aggregate them in Orb. I have decided not to add dimensions to billable connections events then to keep the metric calculation simple in Orb. Let me know if it is an absolute requirement to add more dimensions to billable connections and I will find a solution

Connections are more easily observability in the Nango UI, so IMO less important to slice and dice by environment, provider, etc. This is more important for Actions & MAR, later networking.

bodinsamuel (approved):

PR #4088, opened 2025-05-21
fix(server): fix nested query params parsing

hassan254-prog:

- After upgrading Express, nested query parameters (e.g., authorization_params[display_ui]=always) were no longer parsed correctly, they became flattened, causing issues during OAuth and other parameter i.e credentials.
- This PR explicitly sets a custom query parser using the `qs` library to restore correct parsing behavior for deeply nested query structures.

---

This PR addresses an issue with nested query parameters parsing that arose after an Express upgrade, which caused parameters like authorization_params[display_ui]=always to be flattened incorrectly. The fix explicitly sets Express's query parser to 'extended' to restore handling of deeply nested query structures, such as those required for OAuth.

### DIFF
--- a/packages/server/lib/server.ts
+++ b/packages/server/lib/server.ts
@@ -6,6 +6,7 @@ import http from 'node:http';
 import express from 'express';
 import * as cron from 'node-cron';
 import { WebSocketServer } from 'ws';
+import qs from 'qs';
 
 import db, { KnexDatabase } from '@nangohq/database';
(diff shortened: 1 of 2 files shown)


bodinsamuel (commented):
Thanks, I believe the fix is this.set('query parser', 'extended');
https://github.com/czaarek99/express/commit/7bca2737d207189e6df65d1fee6fc735d370114e

khaliqgant (approved):

PR #4091, opened 2025-05-22
fix(api): tmp fix for rewriting express query

bodinsamuel:
##

- API: /v1 tmp fix for rewriting express query
We can't write to req.query but TS doesn't seems to catch this one.
Ideally we would do better, but it's a legacy endpoint, will rewrite later

---

This PR applies a temporary fix to the /v1 API handler for rewriting Express's `req.query` object, enabling the assignment of the 'model' value. This resolves an issue where direct assignment to `req.query` is not allowed by Express, bypassing the limitation with `Object.defineProperty`.


khaliqgant (approved):
thanks!

PR #4092, opened 2025-05-22
fix(scheduler): max concurrency race condition

TBonnin:
fixing the race condition, using pg advisory lock to ensure max one single dequeue per groupKey and correct max concurrency handling.
The groups table is now unnecessary since it sole purpose was to have something to lock against. I will followup with PRs to remove the group table and denormalize the max concurrency value into the tasks table. Will save a few queries and table maintenance.

---

This PR addresses a race condition in the scheduler's dequeue logic related to max concurrency per group. It introduces PostgreSQL advisory locks (pg_advisory_xact_lock) keyed by groupKey hash to serialize dequeues and properly enforce max concurrency semantics. The change also refactors transaction handling in task dequeuing, removes unnecessary updates to the groups table, and hardens the integration tests for group uniqueness and error handling.

**Key Changes:**
• Introduced use of pg_advisory_xact_lock on groupKey in scheduler dequeue logic (tasks.ts)
• Wrapped dequeue logic in a transaction to ensure advisory lock consistency
• Removed group table update on dequeue, as it's now redundant
• Strengthened integration tests to use random groupKeys (nanoid) and added group/task creation error checks
• Reduced test durations for improved efficiency

**Affected Areas:**
• packages/scheduler/lib/models/tasks.ts (task dequeue logic, group locking)
• packages/scheduler/lib/models/tasks.integration.test.ts (integration tests for scheduler/task handling)

### DIFF
--- a/packages/scheduler/lib/models/tasks.integration.test.ts
+++ b/packages/scheduler/lib/models/tasks.integration.test.ts
@@ -22,7 +22,7 @@ describe('Task', () => {
         const props = {
             name: 'Test Task',
             payload: { foo: 'bar' },
-            groupKey: 'groupA',
+            groupKey: nanoid(),
             retryMax: 3,
             retryCount: 1,
             startsAfter: new Date(),
@@ -84,7 +84,7 @@ describe('Task', () => {
         }
     });
     it('should be dequeued', async () => {
-        const t0 = await createTask(db, { groupKey: 'A' });
+        const t0 = await createTask(db, { groupKey: nanoid() });
         const t1 = await createTask(db);
         const t2 = await createTask(db, { groupKey: t1.groupKey });
         await createTask(db, { groupKey: t0.groupKey });
@@ -123,7 +123,7 @@ describe('Task', () => {
         expect(dequeued).toHaveLength(0);
     });
     it('should be dequeued according to group max concurrency ', async () => {
-        const groupKey = 'A';
+        const groupKey = nanoid();
         const groupMaxConcurrency = 2;
         const t0 = await createTask(db, { groupKey, groupMaxConcurrency });
         const t1 = await createTask(db, { groupKey, groupMaxConcurrency });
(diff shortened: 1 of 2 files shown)


packages/scheduler/lib/models/tasks.integration.test.ts
```diff
@@ -22,7 +22,7 @@ describe('Task', () => {
         const props = {
             name: 'Test Task',
             payload: { foo: 'bar' },
-            groupKey: 'groupA',
+            groupKey: nanoid(),
```
TBonnin, line 25: unrelated but using random id to prevent any collision

packages/scheduler/lib/models/tasks.integration.test.ts
```diff
@@ -168,7 +168,7 @@ describe('Task', () => {
             dequeuePromises.push(tasks.dequeue(db, { groupKey, limit: 100 }).then((d) => d.unwrap()));
         }, 1);
 
-        await new Promise((resolve) => void setTimeout(resolve, 2000));
+        await new Promise((resolve) => void setTimeout(resolve, 200));
```
TBonnin, line 171: reducing the duration of the test. Dequeuing 100s of tasks was taking a few seconds

packages/scheduler/lib/models/tasks.ts
```diff
@@ -258,75 +258,72 @@ export async function dequeue(db: knex.Knex, { groupKey, limit }: { groupKey: st
...
-            .with('running', (qb) => {
-                qb.select(db.raw('count(id) as running_count'), 'group_key')
-                    .from(TASKS_TABLE)
-                    .where('state', 'STARTED')
-                    .whereIn('group_key', function () {
-                        this.distinct('group_key').from('candidates');
+        const tasks = await db.transaction(async (trx) => {
+            // Acquire a lock to prevent concurrent dequeueing of the same group
+            // in order to ensure max concurrency is respected
+            await trx.raw(`SELECT pg_advisory_xact_lock(?) as "lock_dequeue_${groupKey}"`, [stringToHash(groupKey)]);
```
TBonnin, line 264: this is the fix!!!!

bodinsamuel (approved):

PR #4093, opened 2025-05-22
chore: add group max concurrency to tasks table

TBonnin:
The groups table is going to be removed since its sole purpose was to be able to lock at the group level in `dequeue` (we are now using a advisory lock). 
In this commit max concurrency is denormalized and stored in tasks table. We are assuming max concurrency is stable for a given group key.

It would allow us to save a few queries that were updating the groups tables as well as saving a join in dequeue. We won't have to cleanup the groups table neither.
I will stop using the groups table in next PR and will remove the table eventually.

---

**Denormalize group maxConcurrency from groups table to tasks table**

This PR denormalizes the 'max concurrency' property, previously managed in the groups table, by introducing a new groupMaxConcurrency column directly in the tasks table and updating all relevant logic, types, and tests to reference this field. The change is a staged part of phasing out the now-redundant groups table, as concurrency management has shifted to advisory locks. Code, validation, and test fixtures are refactored for groupMaxConcurrency, and a migration adds the new field to the database.

**Key Changes:**
• Added `groupMaxConcurrency` column to tasks table via `PostgreSQL` migration with a default value of 0.
• Refactored type definitions and migration logic, removing all `groupKeyMaxConcurrency` usages and introducing `groupMaxConcurrency` everywhere needed.
• Simplified and removed code in the scheduler (recurring tasks, ``API`` handlers) that previously required groups table upserts or referenced group-level concurrency.
• Updated all relevant orchestrator/client typings, validation schemas, and test fixtures to track `groupMaxConcurrency` on tasks.
• Integration and unit tests updated accordingly.

**Affected Areas:**
• Database schema (tasks table, migration)
• Scheduler logic (task creation, recurring schedules)
• ``API`` endpoints for task scheduling
• Typescript types (task and schedule-related types)
• Validation layers in orchestrator and scheduler
• Unit and integration tests in scheduler and orchestrator

**Potential Impact:**

**Functionality**: Tasks now directly retain max concurrency data, which can change per task. The groups table is being phased out, so code referencing it will need further cleanup. Dequeue and concurrency checks can be made efficiently on the task row itself.

**Performance**: Migration is fast (defaulted column). Fewer joins and write queries when scheduling and dequeueing tasks, so dequeue and task persistence should be faster.

**Security**: No direct security implications.

**Scalability**: Improves scalability of dequeue and task scheduling (fewer joins, no lock contention on groups table, max concurrency available per-task).

**Review Focus:**
• Migration safety (column add with default ``INT``, no locking or full-table scan in `PostgreSQL` 11+)
• Comprehensive update to types and validation (zod schemas, usage in orchestrator)
• Detection of any lingering uses of `groupKeyMaxConcurrency` or references to the groups table for concurrency
• Test coverage for new structure and backwards compatibility in ``API`` integrations

<details>

• Regression testing across all task scheduling (immediate and recurring)
• Validation that max concurrency logic is applied per task, including concurrent dequeues
• Database migration tested for both clean installs and upgrades
• Check backward compatibility for tasks created before migration (default to 0 value)

</details>

<details>

**packages/scheduler/lib/db/migrations/20250522160900_tasks_add_group_max_concurrency.ts**: Migration is forward-only (down is a no-op), safe for production as per PostgreSQL docs for defaulted columns.

**packages/scheduler/lib/scheduler.ts**: Logic and parameter passing updated clearly; recurring schedule upsert simplification is correct.

**packages/orchestrator/lib/clients/types.ts**: All groupMaxConcurrency usage is updated.

**packages/orchestrator/lib/clients/validate.ts**: zod schemas and validation integrate new field, default is handled.

**packages/scheduler/lib/types.ts**: Correct refactor, old groupKeyMaxConcurrency logic eliminated.

</details>

<details>

**Database Migration**:
• Non-blocking additive migration
• `PostgreSQL` behavior checked for defaulted columns

**Type Safety**:
• Typescript types in sync across modules
• Validation schemas updated

**Test Coverage**:
• Integration and unit tests updated for new schema

**Incremental Refactoring**:
• Staged cleanup of deprecated tables and fields

</details>

<details>

• If the migration is run on PostgreSQL <11, ALTER TABLE ... ADD COLUMN with default could take longer; but docs clarify 11+ is safe.
• Down migration does not drop the column; rollback leaves the column in place.
• If max concurrency needs to change dynamically for a groupKey, further changes would be required; code assumes per-task stability.
• Code outside of scheduler/orchestrator (including legacy or untested consumers) still using groups table may break when it is ultimately dropped.

</details>

---


packages/scheduler/lib/db/migrations/20250522160900_tasks_add_group_max_concurrency.ts
```diff
@@ -0,0 +1,12 @@
+import type { Knex } from 'knex';
+import { TASKS_TABLE } from '../../models/tasks.js';
+
+export const config = {
+    transaction: false
+};
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.raw(`ALTER TABLE ${TASKS_TABLE} ADD COLUMN IF NOT EXISTS group_max_concurrency INT NOT NULL DEFAULT 0`);
```
bodinsamuel, line 9: You sure adding a default column is going to be instant in prod?

bodinsamuel (approved):
Just to be sure, validation wont break in case of progressive deploy? Also if there is any issue with rollback it goes the other way; In jobs for example

packages/scheduler/lib/db/migrations/20250522160900_tasks_add_group_max_concurrency.ts
TBonnin: > Just to be sure, validation wont break in case of progressive deploy? Also if there is any issue with rollback it goes the other way; In jobs for example

Have tested the partial deployment in staging without any issue. The max concurrency value was already set to 0 if not existing, that was to handle this very same question when introduced in the first place. The value is not used anywhere yet and when zod validated it is assigned a 0 default value if absent

TBonnin: > You sure adding a default column is going to be instant in prod?

from https://www.postgresql.org/docs/current/ddl-alter.html#DDL-ALTER-ADDING-A-COLUMN
_From PostgreSQL 11, adding a column with a constant default value no longer means that each row of the table needs to be updated when the ALTER TABLE statement is executed. Instead, the default value will be returned the next time the row is accessed, and applied when the table is rewritten, making the ALTER TABLE very fast even on large tables._

kaposke (approved):

PR #4097, opened 2025-05-22
docs: add warning about getMetadata caching in sync

TBonnin:
---

This PR updates the customer configuration guide to include a warning about the 60-second caching behavior when using `nango.getMetadata()` within a sync. The note clarifies that metadata changes during a running sync may not be immediately reflected and will only be visible after the cache expires or in the next sync execution.


khaliqgant:
Good call to update this and call this out. Important to note! Thanks 🙏

bodinsamuel (approved):

PR #4101, opened 2025-05-23
fix(express): correct querystring parsing in persist, support querystring in v1

bodinsamuel:
## Changes

Another express upgrade followup

- Correct querystring parsing in persist
Get records by ids can pass an array of string, it passed the test because it's only passing one. Always a bad idea to rely on array in query params

- Fix support querystring in v1
Broken by originalUrl actually containing the query params, so it was no longer matching endpoints when query params were set.

---

This PR addresses querystring parsing issues introduced by a previous Express upgrade, ensuring that array-type query parameters are correctly parsed in the 'persist' package and that the v1 API routes on the server handle URLs with query parameters properly. The update sets the Express query parser to 'extended' in persist, and revises v1 endpoint routing logic to extract paths from URLs cleanly regardless of query presence.

**Key Changes:**
• Set Express query parser to 'extended' in packages/persist/lib/server.ts to properly handle array query parameters.
• Refactored URL/path extraction for v1 endpoints in packages/server/lib/controllers/sync.controller.ts to use URL parsing and baseUrl, ensuring endpoint matching works when query parameters are present.
• Minor import update to bring 'baseUrl' into scope for path resolution.

**Affected Areas:**
• Persist express server initialization and query parsing
• v1 endpoint routing in sync.controller.ts


packages/persist/lib/server.ts
```diff
@@ -21,6 +21,8 @@ const maxSizeJsonRecords = '100mb';
 
 export const server = express();
 
+server.set('query parser', 'extended');
```
TBonnin, line 24: is this necessary for the other apps?

TBonnin (commented):
the test has 2 externalIds being queried and it was still passing https://github.com/NangoHQ/nango/blob/master/packages/persist/lib/server.integration.test.ts#L387

bodinsamuel:
> the test has 2 externalIds being queried and it was still passing https://github.com/NangoHQ/nango/blob/master/packages/persist/lib/server.integration.test.ts#L387

indeed I misread, I think the big diff is that the test is using node-fetch and the sdk is using axios, so maybe the query params are not stringified the same way

packages/persist/lib/server.ts
bodinsamuel: I don't think so, we mostly use body for jobs and runner

TBonnin (approved):

PR #4103, opened 2025-05-23
fix: race condition in processor tests

TBonnin:
Fixing a a flacky processor integration test.

Incrementing processCount before actually executing the processor handler could cause the processN function to return before the tasks were updated and therefore failing the tests.

Instead of checking for some proxy variables to assess if processing is done, we are now waiting for the expected state and timeout if it is never reached. 

---

This PR refactors the processor integration test to eliminate a race condition caused by incrementing a counter before the processor handler completed. The test now waits for tasks to reach their expected states with a timeout, resulting in more reliable and deterministic behavior.

### DIFF
--- a/packages/orchestrator/lib/clients/processor.integration.test.ts
+++ b/packages/orchestrator/lib/clients/processor.integration.test.ts
@@ -1,4 +1,5 @@
-import { expect, describe, it, beforeAll, afterAll, vi } from 'vitest';
+import { describe, it, beforeAll, afterAll, vi } from 'vitest';
+import type { Task } from '@nangohq/scheduler';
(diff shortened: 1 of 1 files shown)


packages/orchestrator/lib/clients/processor.integration.test.ts
```diff
@@ -37,52 +38,61 @@ describe('OrchestratorProcessor', () => {
 
     afterAll(async () => {
         scheduler.stop();
+        await setTimeout(100); // wait for the scheduler to stop
```
TBonnin, line 41: unrelated. but sometimes db will be cleared before the scheduler is actually stopped

bodinsamuel (approved):

PR #4106, opened 2025-05-26
feat(syncConfig): store sdk version along syncConfigs, onevents

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-3311/store-sdk-version-in-db

- Store sdk version along syncConfigs, onEvents
To be able to know which sdk the scripts are using, and also determines if it's a yaml or not. This will help refine some logic of the code without having to load the actual code. Also in general it will give us the ability to know what's the lowest version we can support.

NB: it's only types and default value, it's not used yet

---

This PR adds support for persisting the SDK version (sdk_version) used during deployments of sync configurations and on-event scripts. These changes introduce new database columns, update relevant TypeScript types, validation schemas, service logic, API endpoints, and test/seed logic to accept and store the SDK version as a first-class, optional property. The version value is currently only captured, stored, and exposed, but not yet actively used for any runtime logic or compatibility decisions.

**Key Changes:**
• Database migration adds `sdk_version` columns to `_nango_sync_configs` and `on_event_scripts` tables.
• TypeScript types/interfaces for sync configs and on-event scripts extended to include `sdk_version`.
• API types, endpoints (`/sync/deploy`, `/sync/deploy/confirmation`, `/sync/deploy/internal`), and validation updated to support and validate the `sdkVersion` property (with regex constraint).
• Service layers (deployment, on-event script service) refactored to propagate and persist the new field.
• Seeder utilities, unit and integration tests updated to set and check `sdkVersion`.
• SDK version is defaulted to `null` if unset; no behavioral change yet.

**Affected Areas:**
• Sync config and on-event script types, validation and service layers
• All sync deployment API routes and controllers
• Seeders and test suites across CLI, server, jobs, and types packages

### DIFF
--- a/packages/database/lib/migrations/20250519153421_sync_config_sdk_version.cjs
+++ b/packages/database/lib/migrations/20250519153421_sync_config_sdk_version.cjs
@@ -1,0 +1,9 @@
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`ALTER TABLE "_nango_sync_configs" ADD COLUMN "sdk_version" varchar(25)`);
+    await knex.raw(`ALTER TABLE "on_event_scripts" ADD COLUMN "sdk_version" varchar(25)`);
+};
+
+exports.down = function () {};
--- a/packages/cli/lib/services/dryrun.service.ts
+++ b/packages/cli/lib/services/dryrun.service.ts
@@ -341,6 +341,7 @@ export class DryRunService {
                 is_public: false,
                 metadata: {},
                 pre_built: false,
+                sdk_version: null,
                 sync_type: lastSyncDate ? 'incremental' : 'full',
                 version: '0.0.1'
             };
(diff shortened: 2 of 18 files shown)


packages/database/lib/migrations/20250519153421_sync_config_sdk_version.cjs
```diff
@@ -0,0 +1,9 @@
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`ALTER TABLE "_nango_sync_configs" ADD COLUMN "sdk_version" varchar(25)`);
+    await knex.raw(`ALTER TABLE "on_event_scripts" ADD COLUMN "sdk_version" varchar(25)`);
+};
+
+exports.down = function () {};
```
kaposke, line 9: Do we really not care about the down statement?

kaposke (approved):

packages/database/lib/migrations/20250519153421_sync_config_sdk_version.cjs
bodinsamuel: yes, we don't even a way to call them in the codebase. Thomas and I agreed it was useless based on our experience of never ever executing this way. Only forward is easier to understand and makes the expectation clearer

PR #4113, opened 2025-05-26
feat: download scripts from UI

kaposke:
### Description
Adds download button to endpoint page. Uses existing `downloadFlow` function.

image

### How to test
You can try it locally, but ideally try it in staging. Unless overwritten, this should be on staging right now.

---

This PR implements a 'Download Script' button in the UI on the endpoint details page, enabling users to download endpoint scripts (as ZIP files) via the browser. The frontend leverages the existing API, adding client-side download handling and improved ZIP filename logic (handling null versions and prefixing version numbers with 'v'). Backend changes update the download flow to resolve public script templates using the `provider` field (rather than `providerConfigKey`), which corrects alignment between frontend and backend expectations, and removes the obsolete `public_route` field from the shared FlowDownloadBody type.

**Key Changes:**
• UI: Added 'Download Script' button to endpoint page (One.tsx) with full client-side download logic.
• Frontend now queries the backend to fetch ZIP file and triggers a browser download, with ZIP filename updated to handle 'null' versions and added 'v' prefix.
• Backend (flow.controller.ts): Download handler for public templates now uses `provider` instead of `providerConfigKey` to match data expectations for public flows.
• Shared types: FlowDownloadBody interface cleaned of obsolete `public_route` property, usage updated across client/server.

**Affected Areas:**
• Webapp endpoint details UI
• Backend flow controller (download logic)
• Shared type definitions for flow download


bodinsamuel (approved):
Works okay 👍🏻 
Would have love to see downloadFlow migrated to new format

packages/server/lib/controllers/flow.controller.ts
```diff
@@ -93,7 +93,7 @@ class FlowController {
             }
 
             if (!id && is_public) {
-                await remoteFileService.zipAndSendPublicFiles(res, name, accountId, environmentId, body.public_route as string, flowType);
+                await remoteFileService.zipAndSendPublicFiles(res, name, accountId, environmentId, providerConfigKey, flowType);
```
khaliqgant: Should this be `provider` instead of `providerConfigKey`?

A user can have a microsoft-teams public template active but called their integration `microsoft-abcd`. The public template woudn't be found using the `providerConfigKey` of `microsoft-abcd`

kaposke: That's the kind of thing I'm still confused about, thanks.

packages/webapp/src/pages/Integrations/providerConfigKey/Endpoints/components/One.tsx
```diff
@@ -122,13 +130,57 @@ export const EndpointOne: React.FC<{ integration: GetIntegration['Success']['dat
...
+
+            if (!response.ok) {
+                throw new Error('Network response was not ok');
+            }
+
+            const blob = await response.blob();
+            const url = window.URL.createObjectURL(blob);
+            const a = document.createElement('a');
+            a.href = url;
+            a.download = `${provider}_${name}_${version}.zip`;
```
TBonnin: version can be null

Screenshot 2025-05-28 at 09 04 42

packages/shared/lib/models/Flow.ts
```diff
@@ -3,7 +3,6 @@ export interface FlowDownloadBody {
     name: string;
     provider: string;
     is_public: boolean;
-    public_route?: string;
```
TBonnin, line 6: what was the point of having public_route before?

packages/webapp/src/pages/Integrations/providerConfigKey/Endpoints/components/One.tsx
TBonnin: I would also add `v` prefix before the version. wdyt?

TBonnin (approved):

packages/server/lib/controllers/flow.controller.ts
bodinsamuel: But then it would not work for regular scripts, no?
Shouldn't it be `syncConfig.template ? provider : providerConfigKey` ?

packages/webapp/src/pages/Integrations/providerConfigKey/Endpoints/components/One.tsx
kaposke: Makes sense yeah. And sorry, didn't know it could be null

packages/shared/lib/models/Flow.ts
kaposke: It beats me, but I confirmed it's just supposed to be `provider`

packages/server/lib/controllers/flow.controller.ts
khaliqgant: I missed the comments here after I made mine. This method specifically is just for public templates

isPublic

packages/shared/lib/models/Flow.ts
khaliqgant: It was how it was named coming in from the fronted but then poorly named/confusing within the context of the backend

PR #4118, opened 2025-05-27
fix(cli): prepare zero yaml 

bodinsamuel:
## Changes

- (cli): prepare zero yaml 
Just block stuff that are not ready to be used with zero yaml

---

This PR introduces formal support and early validation for 'zero-yaml' projects within the Nango CLI, where projects have a `.nango` marker and an `index.ts`, but no `nango.yaml`. Core CLI commands now run a centralized pre-check to block commands incompatible with zero-yaml project layouts, centralizing this logic in `verificationService`. As part of this refactor, CLI option typings have been unified for consistency, and debug logging has been made more systematic. User-facing error handling and help messaging have also seen framework-wide improvements, increasing clarity for both valid and invalid project states.

**Key Changes:**
• Added `verificationService.preCheck` and `verificationService.ensureNangoYaml` for project structure/type validation (classic yaml vs. zero-yaml)
• Most CLI commands now perform an early pre-check and block execution if run in an unsupported (zero-yaml) project context
• Refactored and enforced CLI option typings (now using `GlobalOptions`); reduced ad-hoc debug flag logic and standardized `printDebug` usage
• Restructured imports and improved modularity across CLI entrypoint, utilities, and services
• Enhanced error and debug outputs, providing earlier and clearer user feedback on project validity

**Affected Areas:**
• CLI entrypoint (index.ts)
• verificationService (services/verification.service.ts)
• CLI utility and option parsing (utils.ts)

### DIFF
--- a/packages/cli/lib/index.ts
+++ b/packages/cli/lib/index.ts
@@ -111,7 +111,7 @@ program
     .action(async function (this: Command) {
         const { debug } = this.opts<GlobalOptions>();
         const fullPath = process.cwd();
-        const precheck = await verificationService.ensureNangoV1({ fullPath, debug });
+        const precheck = await verificationService.ensureNangoYaml({ fullPath, debug });
         if (!precheck) {
             return;
         }
@@ -149,7 +149,7 @@ program
         const { autoConfirm, debug, e: environment, integrationId, validation, saveResponses } = this.opts();
         const fullPath = process.cwd();
 
-        const precheck = await verificationService.ensureNangoV1({ fullPath, debug });
+        const precheck = await verificationService.ensureNangoYaml({ fullPath, debug });
         if (!precheck) {
             return;
         }
@@ -178,7 +178,7 @@ program
         const { compileInterfaces, autoConfirm, debug } = this.opts();
         const fullPath = process.cwd();
 
-        const precheck = await verificationService.ensureNangoV1({ fullPath, debug });
+        const precheck = await verificationService.ensureNangoYaml({ fullPath, debug });
         if (!precheck) {
             return;
         }
@@ -202,7 +202,7 @@ program
         const options = this.opts<DeployOptions>();
         const { debug } = options;
         const fullPath = process.cwd();
-        const precheck = await verificationService.ensureNangoV1({ fullPath, debug });
+        const precheck = await verificationService.ensureNangoYaml({ fullPath, debug });
         if (!precheck) {
             return;
         }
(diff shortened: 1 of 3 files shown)


packages/cli/lib/index.ts
```diff
@@ -25,20 +25,18 @@ import { directoryMigration, endpointMigration, v1toV2Migration } from './servic
...
-        cmd.option('--debug', 'Run cli in debug mode, outputting verbose logs.');
+        cmd.option('--debug', 'Run cli in debug mode, outputting verbose logs.', false);
         cmd.hook('preAction', async function (this: Command, actionCommand: Command) {
             const { debug } = actionCommand.opts();
-            if (debug) {
-                printDebug('Debug mode enabled');
-                if (fs.existsSync('.env')) {
-                    printDebug('.env file detected and loaded');
-                }
+            printDebug('Debug mode enabled', debug);
```
TBonnin, line 37: is it on purpose to always show the debug mode option?

packages/cli/lib/services/verification.service.ts
```diff
@@ -120,6 +123,48 @@ class VerificationService {
...
+        return {
+            isNango,
+            folderName: path.basename(fullPath),
+            hasNangoYaml,
+            hasIndexTs,
+            isZeroYaml
+        };
+    }
+
+    public async ensureNangoV1({ fullPath, debug }: { fullPath: string; debug: boolean }) {
```
TBonnin: what does the v1 means here. Is it linked to yaml v1/v2?

packages/cli/lib/utils.ts
```diff
@@ -36,8 +39,10 @@ if (parsedHostport.slice(-1) === '/') {
 
 export const hostport = parsedHostport;
 
-export function printDebug(message: string) {
-    console.log(chalk.gray(message));
+export function printDebug(message: string, debug?: boolean) {
+    if (debug === true || debug === undefined) {
```
TBonnin, line 43: undefined debug is equivalent to true?

TBonnin (approved):

packages/cli/lib/services/verification.service.ts
bodinsamuel: yeah, maybe I could have named it ensureNangoYaml

packages/cli/lib/utils.ts
bodinsamuel: it's never undefined unless I haven't yet migrated the use of this function, e.g:
```ts
printDebug('foobar', true | false);
// vs old way
if (debug) {
  printDebug('foobar');
}
```

packages/cli/lib/index.ts
bodinsamuel: cf the function, I'm passing the debug so that we can remove the hundreds of if in the code

PR #4019, opened 2025-05-09
feat(integrations): Add support for Dropbox sign

lordsarcastic:
CONVERTED TO DRAFT SINCE NANGO DOESN'T YET SUPPORT COMPLETE REMOVAL OF THE SCOPE PARAM FROM THE URL WHICH CAUSES AN ERROR WHEN OBTAINING AUTHORIZATION.

This PR adds support for the Dropbox Sign OAuth integration, including documentation for setup and usage.

- Ticket: [EXT-696](https://linear.app/nango/issue/EXT-696/add-dropbox-sign-as-an-oauth-provider)

---

This PR introduces Dropbox Sign as a new OAuth2 provider within the Nango platform. It adds the necessary provider configuration (endpoints, OAuth parameters, and retry logic) in `providers.yaml`, comprehensive documentation for setup and usage, and the provider's assets (including an SVG logo) to the documentation/UI. The documentation includes setup details, API requirements, and notes about production app approval, ensuring users have clear integration guidance.

**Key Changes:**
• Added Dropbox Sign (HelloSign) OAuth2 configuration to `providers.yaml` with endpoints, proxy base URL, and retry mechanisms.
• Created provider setup and usage documentation under `docs-v2/integrations/all/dropbox-sign.mdx`, with embedded tooling and use-case snippets.
• Registered new documentation entry for Dropbox Sign in the docs navigation via `docs-v2/mint.json`.
• Added Dropbox Sign SVG logo asset for UI representation.

**Affected Areas:**
• OAuth provider configuration (`providers.yaml`)
• Integration documentation (various `.mdx` files, `mint.json`)
• Webapp image assets

*This summary was automatically generated by @propel-code-bot*

PR #4124, opened 2025-05-28
feat(authorization): skip undefined/empty query parameter

hassan254-prog:

- Some providers like [Dropbox sign OAuth](https://github.com/NangoHQ/nango/pull/4019), do not allow empty query parameters, such as `scope` which we automatically inject.
- This PR introduces a flag `authorization_url_skip_undefined` that can be set to `true` for providers where we want to omit these undefined  parameters from the authorization URL.

---

This PR introduces a new boolean configuration flag, `authorization_url_skip_empty`, to provider definitions, enabling selective omission of query parameters with empty or undefined values from OAuth authorization URLs. The update addresses cases where some providers (e.g., Dropbox Sign) reject URLs containing such parameters (like `scope`) when they're empty, enhancing compatibility with strict provider requirements.

**Key Changes:**
• Added `authorization_url_skip_empty?: boolean` to provider type definitions (`provider.ts`)
• Extended provider JSON schema (`schema.json`) to include the new flag
• Updated OAuth controller logic (`oauth.controller.ts`) to remove empty string query parameters from the authorization URL when the flag is enabled

**Affected Areas:**
• OAuth controller logic
• Provider configuration schema validation
• Provider TypeScript interface definitions


bodinsamuel (approved):
Is there any risk to always remove them?

hassan254-prog:
> Is there any risk to always remove them?

I was thinking about that, but I’m concerned it might be a breaking change for some. What do you think, should we omit empty query params entirely for all?

bodinsamuel:
> I was thinking about that, but I’m concerned it might be a breaking change for some. What do you think, should we omit empty query params entirely for all?

I guess it could be 😞  lets go with that then 👍🏻

TBonnin (approved):
can we always skip the empty params? regarless of the provider and then avoid having to add yet another option to provider.yaml?

bodinsamuel:
> can we always skip the empty params? regarless of the provider and then avoid having to add yet another option to provider.yaml?

asked the same but in retrospect I'm sure some are relying on this unfortunately

PR #4128, opened 2025-05-28
fix(slack): ts is not always present

bodinsamuel:
## Changes

- Slack `ts` is not always present
We have had constant base failure when delivering slack messages, contrary to what the documentation tells, it seems `ts` is not always present however I haven't dug further to check when that happen. This PR makes sure it's not failing the full operation

---

This PR updates the Slack notification service to gracefully handle cases where the 'ts' (timestamp) field is absent from Slack API responses, addressing a recurring cause of failures. Code paths that previously assumed 'ts' was always present now check both the top-level and nested 'message' object for this value, and update logs and downstream processing accordingly. Additional minor fixes were made around external config typing.

**Key Changes:**
• Modified PostSlackMessageResponse interface to make 'ts' optional.
• Updated all usages (logging, DB updates) to safely access 'ts' from either response.ts or response.message.ts.
• Improved import ordering and removed unused imports for clarity.
• Standardized proxy config method typing for Slack channel actions.

**Affected Areas:**
• packages/shared/lib/services/notification/slack.service.ts


packages/shared/lib/services/notification/slack.service.ts
```diff
@@ -40,7 +42,7 @@ interface NotificationPayload {
 interface PostSlackMessageResponse {
     ok: boolean;
     channel: string;
-    ts: string;
```
TBonnin, line 43: can we keep it and set it ourselves in sendSlackNotification if not present. Then the rest of the code doesn't have to change and we can ensure it is always present

TBonnin (approved):

packages/shared/lib/services/notification/slack.service.ts
bodinsamuel: I'm not sure to get what you mean? If it's a new msg then we can guess this value

TBonnin: I mean can we do `res.value.ts ||= res.value.message.ts;` in sendSlackNotification so we don't have to repeat  `const ts = res.value.ts || res.value.message.ts;` on L185 below

bodinsamuel: not a fan of modifying the response object, I need to investigate more why it's not returning correctly I'm just trying to fix the thousands of errors spamming regularly

PR #4134, opened 2025-05-28
fix(server): generate mcp input schema

kaposke:
## Description
Turns out the `models_json_schema` column only contains output models. I've looked into adding input models to it and to the `models` column as well, but I'm uncertain of the consequences, and any existing action would only be updated on redeploy.

Instead, we generate a json schema on-demand based on the `models_schema` column, which contains every model, but in the `NangoModel` format instead of json schema.

Unit tests generated by AI (probably the best use of AI ever).

## How to test
Use this curl to invoke `tools/list` and check the tool's input json schema.
```
curl -X POST http://localhost:3003/mcp \
-H "Authorization: Bearer <your-token>" \
-H "connection-id: <cid>" \
-H "provider-config-key: <provider-config-key>" \
-H "Accept: application/json, text/event-stream" \
-H "Content-Type: application/json" -d '{
    "jsonrpc": "2.0",
    "method": "tools/list",
    "id": 1
  }'
```
> The required replacements are the token, connection-id and provider-config-key

Deeper test:
you can use our sample https://github.com/NangoHQ/nango-mcp-client-example to actually interact with the MCP server via a chat client.

---

**Dynamic JSON Schema Generation for MCP Input Models via NangoModel**

This PR refactors input schema generation for MCP tools by introducing on-demand conversion of NangoModel definitions to JSON Schema, replacing reliance on statically stored output-only schemas. It centralizes the schema conversion logic in the shared utils package, ensures explicit handling of recursion and union/model/array types, excludes legacy schema formats from MCP exposure, and adds a comprehensive, edge-case-covering unit test suite for schema conversion.

**Key Changes:**
• Introduced `nangoModelToJsonSchema` utility for converting `NangoModel` to ``JSON`` Schema with recursion and type-safety checks.
• Refactored ``MCP`` server action registration to use runtime schema generation via the new utility, skipping unsupported legacy model schemas.
• Moved conversion logic to utils package for reuse and updated exports accordingly.
• Extensively expanded unit test coverage for the conversion utility, covering primitives, arrays, unions, deep nesting, error modes, and circular references.
• Explicitly handle `date` types in schema conversion.

**Affected Areas:**
• ``MCP`` server controller (`actionToTool` logic, tool schema generation)
• Utils/shared libraries (json-schema.ts, index.ts)
• Unit testing of schema conversion logic

**Potential Impact:**

**Functionality**: Provides accurate, up-to-date input schemas for MCP tools using NangoModel definitions. Tools/actions using legacy model schemas are intentionally omitted until redeployed.

**Performance**: Adds minor runtime overhead for schema conversion, with explicit recursion protection to avoid stack overflows; suitable for most use-cases.

**Security**: No direct new risks; improved input validation through stronger schemas.

**Scalability**: Error handling and circular reference checks ensure that pathological or very deep model graphs fail safely instead of overloading the system. Recursive schemas remain unsupported due to technical limitations of the MCP protocol.

**Review Focus:**
• Correctness and coverage of `nangoModelToJsonSchema` implementation (especially recursion and union/array handling).
• Depth and validity of new test cases.
• Proper exclusion of legacy model schemas in ``MCP`` controller.
• Appropriateness of utility code extraction to utils for future reuse.

<details>

• Run the expanded unit test suite in packages/utils/lib/json-schema.unit.test.ts.
• Test ``MCP`` /tools/list endpoint and verify output schemas for diverse action models.
• Regress with deep nesting, union types, arrays, and malformed/circular models.
• Try sample client against ``MCP`` server as described in the ``PR`` body.

</details>

<details>

**packages/utils/lib/json-schema.ts**: Highly structured, with clear separation of concerns, early error returns, explicit recursion guards, and full support for error scenarios. Follows functional paradigm by returning Result type.

**packages/utils/lib/json-schema.unit.test.ts**: Extensive and systematic, covers a wide variety of normal and edge cases-including deep nesting, unions, optionals, errors, circular references; validates all major code paths.

**packages/server/lib/controllers/mcp/server.ts**: Improved modularity; schema logic is simplified and robust, skips unsupported legacy formats, leverages the new shared utility. Naming improvements suggested in review were implemented.

**packages/utils/lib/index.ts**: Properly re-exports new shared utility.

</details>

<details>

**Defensive Programming**:
• Recursion/circular reference detection with error messages.
• Early error returns for unexpected model/field shapes.

**Testing**:
• Broad and deep unit test coverage for both typical and edge/failure cases.

**Code Organization**:
• General-purpose logic extracted and re-exported via utils for future reuse.

</details>

<details>

• Legacy schemas for actions prior to adoption of NangoModel remain unavailable to MCP until redeployment; this is intentionally documented and aligned with team consensus.
• Schemas with direct or indirect circular references are rejected-true recursion cannot be represented due to protocol/MCP schema limitations.
• If a model graph is excessively deep, stack guard will prevent infinite loops but could still impose small runtime cost on conversion.
• Unknown or exotic NangoModel value types default to string records, which may lose specificity.

</details>

---


packages/server/lib/controllers/mcp/server.ts
```diff
@@ -73,11 +82,7 @@ function actionToTool(action: DBSyncConfig): Tool | null {
...
         name: action.sync_name,
-        inputSchema: {
-            type: 'object',
-            properties: inputSchema.properties,
-            required: inputSchema.required
-        },
+        inputSchema: inputSchema as Tool['inputSchema'],
         description
     };
 }
```
TBonnin, line 94: I think you can simplify above an do an early return if inputModel is not defined
```
if(!inputModel || isLegacyModelSchema(inputModel)) {
   return null;
}
return {
   name: action.sync_name,
   inputSchema: nangoModelToJsonSchema(...) as Tool['inputSchema'],
   description: action.metadata.description || action.sync_name;
}
```

packages/server/lib/controllers/mcp/server.ts
```diff
@@ -58,11 +58,20 @@ async function getActionsForProvider(environment: DBEnvironment, providerConfig:
...
 function actionToTool(action: DBSyncConfig): Tool | null {
-    const inputSchema =
-        action.input && action.models_json_schema?.definitions && action.models_json_schema?.definitions?.[action.input]
-            ? (action.models_json_schema.definitions[action.input] as JSONSchema7)
-            : ({ type: 'object' } as JSONSchema7);
+    const inputModelName = action.input;
+    const inputModel = action.model_schema?.find((m) => m.name === inputModelName);
+
+    if (inputModel && isLegacyModelSchema(inputModel)) {
+        // We don't support legacy model schemas for MCP. Redeploying the action should update the model schema.
```
TBonnin, line 69: why don't we support legacy model schema?

packages/server/lib/controllers/mcp/utils/json-schema.ts
```diff
@@ -0,0 +1,84 @@
+import type { NangoModel, NangoModelField } from '@nangohq/types';
+import type { JSONSchema7 } from 'json-schema';
+
```
TBonnin: functions that can fail should return a `Result` to force explicit error handling

packages/server/lib/controllers/mcp/utils/json-schema.ts
```diff
@@ -0,0 +1,84 @@
+import type { NangoModel, NangoModelField } from '@nangohq/types';
```
TBonnin: I would like to see this file in utils or in a shareable package. It is not specific to mcp and I wouldn't be surprise if we can reuse this nango schema to json schema logic somewhere else

packages/server/lib/controllers/mcp/utils/json-schema.ts
```diff
@@ -0,0 +1,84 @@
...
+        };
+    } else if (field.value === 'boolean') {
+        return {
+            type: 'boolean'
+        };
+    }
+
+    return {
+        type: 'string'
+    };
```
TBonnin: is `date` a type that should be explicitely handled?

packages/server/lib/controllers/mcp/utils/json-schema.ts
```diff
@@ -0,0 +1,84 @@
+import type { NangoModel, NangoModelField } from '@nangohq/types';
+import type { JSONSchema7 } from 'json-schema';
+
+export function nangoModelToJsonSchema(model: NangoModel, models_schema: NangoModel[]): JSONSchema7 {
+    return nangoModelToJsonSchemaInternal(model, models_schema, new Set());
+}
+
+function nangoModelToJsonSchemaInternal(model: NangoModel, models_schema: NangoModel[], modelStack: Set<string>): JSONSchema7 {
+    if (modelStack.has(model.name)) {
+        throw new Error(`Circular reference detected: ${Array.from(modelStack).join(' -> ')} -> ${model.name}`);
```
TBonnin: is it a big limitation not to support circular reference? I guess we can start without it and add later when needed

packages/server/lib/controllers/mcp/utils/json-schema.ts
```diff
@@ -0,0 +1,84 @@
+import type { NangoModel, NangoModelField } from '@nangohq/types';
+import type { JSONSchema7 } from 'json-schema';
+
+export function nangoModelToJsonSchema(model: NangoModel, models_schema: NangoModel[]): JSONSchema7 {
+    return nangoModelToJsonSchemaInternal(model, models_schema, new Set());
+}
+
+function nangoModelToJsonSchemaInternal(model: NangoModel, models_schema: NangoModel[], modelStack: Set<string>): JSONSchema7 {
```
TBonnin: would call `modelStack` something like `visitedModels` or something similar to make it clearer what it is

TBonnin (commented):
This question is probably very much out of scope but should `model_schema` stored as jsonschema in the db instead of using our homegrown `NangoModel` format?

packages/server/lib/controllers/mcp/utils/json-schema.unit.test.ts
```diff
@@ -0,0 +1,622 @@
+import { describe, expect, it } from 'vitest';
+
+import { nangoModelToJsonSchema } from './json-schema.js';
+
+import type { NangoModel, NangoModelField } from '@nangohq/types';
+
+describe('nangoModelToJsonSchema', () => {
```
TBonnin: happy to see a pretty comprehensive tests suite

packages/server/lib/controllers/mcp/server.ts
kaposke: I don't think it's worth it. It represents less than a thousand active actions, involves me studying and understanding the format and writing more conversion logic, and it is replaced by the new schema in new deploys.

packages/server/lib/controllers/mcp/utils/json-schema.ts
kaposke: Agreed, do you think the utils package is a nice fit?

packages/server/lib/controllers/mcp/utils/json-schema.ts
kaposke: Good question. I'm working out of examples and unaware of all the possibilities. Let me check

packages/server/lib/controllers/mcp/utils/json-schema.ts
kaposke: Actually MCP doesn't support nested schemas at all, which I think is stupid. We are supporting it because it's easy enough, but circular references require, well... references, and references are not supported here (there's nowhere explicit to put other models to reference)

packages/server/lib/controllers/mcp/utils/json-schema.unit.test.ts
kaposke: Cursor generated. I find it does very well with this sort unit of tests...

kaposke:
> This question is probably very much out of scope but should `model_schema` stored as jsonschema in the db instead of using our homegrown `NangoModel` format?

Tbh I kept asking myself the reason why we're inventing new specifications instead of using json schema. And yes, I totally think we should.

packages/server/lib/controllers/mcp/utils/json-schema.ts
kaposke: Apparently `Record<string, any>` is also a thing...

packages/server/lib/controllers/mcp/server.ts
bodinsamuel: just to be clear both of them are legacy:
- legacy is > 1year ago (limited syntax and parsing)
- legacy < 1 year (better syntax but still Nango syntax)

bodinsamuel (commented):
Thanks for investigating.
I would prefer if we would get rid of the old syntax, it's a legacy thing that we are actively removing with the zero yaml.

What we saw is that jsonSchema should be okay for regular scripts, it's just broken for templates. You can see in deploy.service.ts:747 we correctly filters the models for regular scripts, this logic needs to be ported to line 492.
With this you won't need to maintain this large piece of code and actually fix the bug. With the only requirement of redeploying scripts which is imo a completely fine requirement


So to answer both of you
>This question is probably very much out of scope but should model_schema stored as jsonschema in the db instead of using our homegrown NangoModel format?
> Tbh I kept asking myself the reason why we're inventing new specifications instead of using json schema. And yes, I totally think we should.

We are, model_schema is just a legacy syntax, column models_json_schema should be the source of truth and if broken, it should be fixed

packages/server/lib/controllers/mcp/server.ts
kaposke: Right.
To clarify 100%, in this case the model_schema is:
```ts
model_schema: NangoModel[] | LegacySyncModelSchema[] | null;
```
and I'm only supporting `NangoModel`.

kaposke:
> Thanks for investigating. I would prefer if we would get rid of the old syntax, it's a legacy thing that we are actively removing with the zero yaml.
> 
> What we saw is that jsonSchema should be okay for regular scripts, it's just broken for templates. You can see in deploy.service.ts:747 we correctly filters the models for regular scripts, this logic needs to be ported to line 492. With this you won't need to maintain this large piece of code and actually fix the bug. With the only requirement of redeploying scripts which is imo a completely fine requirement
> 
> So to answer both of you
> 
> > This question is probably very much out of scope but should model_schema stored as jsonschema in the db instead of using our homegrown NangoModel format?
> > Tbh I kept asking myself the reason why we're inventing new specifications instead of using json schema. And yes, I totally think we should.
> 
> We are, model_schema is just a legacy syntax, column models_json_schema should be the source of truth and if broken, it should be fixed

The fact that only output models are saved to that new json_schema column tells me that `model_schema` is used in places right now. Yes, we can deprecate it but definitely goes way beyond the scope of getting MCP shipped. I say we raise this but let me merge this to unblock MCP.

bodinsamuel:
> The fact that only output models are saved to that new json_schema column tells me that model_schema is used in places right now. Yes, we can deprecate it but definitely goes way beyond the scope of getting MCP shipped. I say we raise this but let me merge this to unblock MCP.

Forget about deprecating or doing more than this PR, I'm saying using model_schema content is a liability, you should not use it. 

> only output models are saved to that new json_schema column

As I was saying, unless mistaken, It's only broken for template which just needs a fix. 
Notice that we are also using `model_schema` not output. Ideally in a near future, we also stop using model_schema here but we don't want to new code using it.
https://github.com/NangoHQ/nango/blob/401208e4f1e93faa0719813139e754d599fa824c/packages/shared/lib/services/sync/config/deploy.service.ts#L746-L773

kaposke:
Closing this as we decided to go for fixing the root cause on #4142

PR #4135, opened 2025-05-29
fix: add missing index for sync endpoint table

TBonnin:
We are joining on sync_config_id here:
https://github.com/NangoHQ/nango/blob/a093baae68861bce97b7ede53074eb9ab710c9da/packages/shared/lib/services/sync/config/config.service.ts#L707 but there were no index

 

---

This PR adds a database migration that creates a btree index on the sync_config_id column of the _nango_sync_endpoints table, which is used in join operations but previously lacked an index. The migration utilizes CREATE INDEX CONCURRENTLY to avoid locking the table during index creation.

### DIFF
--- a/packages/database/lib/migrations/20250529002200_sync_endpoint_sync_config_id_index.cjs
+++ b/packages/database/lib/migrations/20250529002200_sync_endpoint_sync_config_id_index.cjs
@@ -1,0 +1,9 @@
+exports.config = { transaction: false };
+
+exports.up = async function (knex) {
+    await knex.schema.raw(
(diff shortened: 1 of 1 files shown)


bodinsamuel (approved):

packages/database/lib/migrations/20250529002200_sync_endpoint_sync_config_id_index.cjs
```diff
@@ -0,0 +1,9 @@
+exports.config = { transaction: false };
+
+exports.up = async function (knex) {
+    await knex.schema.raw(
+        'CREATE INDEX CONCURRENTLY IF NOT EXISTS "idx_nango_sync_endpoints_sync_config_id" ON nango._nango_sync_endpoints USING BTREE ("sync_config_id")'
```
bodinsamuel: Remove nango namespace

bodinsamuel (approved):

PR #4139, opened 2025-05-30
docs: Improved pg pooler docs and fixed formatting of env variables

paradoxloop:
Added a documentation note that Nango doesn't play well with postgres database poolers and added missing sh tags to some of the env display panels.

---

This PR updates the self-hosting configuration documentation to explicitly warn users that Nango does not support Postgres connection poolers with `pool_mode=transaction`, providing a clear note on the limitation and guidance for connection setup. Additionally, it standardizes and corrects the formatting of environment variable code blocks by adding 'sh' (shell) language tags for improved readability and consistency.


bodinsamuel:
Hi,
> Added a documentation note that Nango doesn't play well with postgres database poolers and added missing sh tags to some of the env display panels.

Where did you get that? we did test a while back and it worked

paradoxloop:
Nango self hosting doesn't work with Neon or DigitalOcean postgres poolers. I explored solutions with @TBonnin and @bastienbeurier by email and it seemed turning off the poolers was the only solution.

paradoxloop:
@TBonnin helpfully found the sections of neon and pgpooler [docs](https://neon.com/docs/connect/connection-pooling) that explain the problem:

**Avoid using SET statements over a pooled connection**
_Due to the transaction mode limitation described above, users often encounter issues when running SET statements over a pooled connection. For example, if you set the Postgres search_path session variable using a SET search_path statement over a pooled connection, the setting is only valid for the duration of the transaction. As a result, a session variable like search_path will not remain set for subsequent transactions.
This particular search_path issue often shows up as a relation does not exist error. To avoid this error, you can:
    Use a direct connection string when you need to set the search path and have it persist across multiple transactions.
    Explicitly specify the schema in your queries so that you don’t need to set the search path.
    Use an ALTER ROLE your_role_name SET search_path TO <schema1>, <schema2>, <schema3>; command to set a persistent search path for the role executing queries. See the ALTER ROLE
For the official list of limitations, refer to the "SQL feature map for pooling modes" section in the [pgbouncer.org](http://pgbouncer.org/) Features documentation._

**Connection pooling with schema migration tools**
_We recommend using a direct (non-pooled) connection string when performing migrations using Object Relational Mappers (ORMs) and similar schema migration tools. With the exception of recent versions of Prisma ORM, which support using a pooled connection string with Neon, using a pooled connection string for migrations is likely not supported or prone to errors. Before attempting to perform migrations over a pooled connection string, please refer to your tool's documentation to determine if pooled connections are supported._

TBonnin:
the issue is not the pooler but the transaction mode. Nango doesn't play well with pool_mode=transaction, which is what some of the pg providers like neon are using. We experimented with other pool modes in the past without any issue. Could you modify your PR to reflect the problem comes from pool_mode=transaction? Thank you

paradoxloop:
@TBonnin of course! Just trying to be helpful to the next person who might get stuck with this.

docs-v2/guides/self-hosting/free-self-hosting/configuration.mdx
```diff
@@ -72,6 +68,12 @@ NANGO_DATABASE_URL=postgresql://user:password@host:port/dbname
 ```
 Special characters in the NANGO_DATABASE_URL should be URL encoded.
 
+<Note>
+    For external databases using poolers, Nango does not play well with `pool_mode=transaction`
+     which some providers (e.g. Neon and DigitalOcean) use in their poolers. Use a direct connection or change the pooler mode.
```
TBonnin:
```suggestion
Nango is incompatible with connection poolers using `pool_mode=transaction`. Please use a direct database connection or configure the pooler to use a different mode.
```

bodinsamuel (approved):

PR #4146, opened 2025-06-03
fix: more descriptive errors for disabled sync configs

rossmcewan:
This PR changes how disabled sync configs are handled and reported throughout the codebase. Instead of treating disabled configs as 'not found,' it now explicitly checks the config's 'enabled' status after retrieving records and returns a 'disabled' error where appropriate; this applies to syncs, actions, and webhooks, affecting both API responses and internal task execution layers.

**Key Changes:**
• In sync controller and orchestrator, shifted enabled check to after config lookup and emit specific error for disabled configs instead of 'not found.'
• Applied explicit disabled checks to execution paths for syncs, actions, and webhooks in the jobs package, throwing descriptive errors as needed.
• Removed 'enabled: true' direct query filter in some places to allow fetching disabled configs for accurate error reporting.
• Standardized error response codes to explicitly indicate the 'disabled' status (HTTP 400 with error code 'disabled').

**Affected Areas:**
• Server sync controller (API responses)
• Orchestrator scheduling logic
• Jobs execution for syncs, actions, webhooks
• Shared sync config service (query adjustment)

---

This PR refines how disabled sync, action, and webhook configurations are handled across API endpoints and internal execution layers. Rather than returning a generic 'not found' error when such resources are disabled, the system now explicitly distinguishes disabled resources, emitting a standardized 'disabled_resource' error code and clear message. The implementation shifts the enabled-status check out of database queries and into the response logic, ensuring more descriptive diagnostics and consistency between the API and backend jobs. Types and error responses are updated to standardize this explicit signaling.

**Key Changes:**
• Removed 'enabled: true' filter from some config queries to fetch both enabled and disabled configs for accurate error reporting.
• Controllers, orchestrator logic, and jobs for syncs/actions/webhooks now perform an explicit check on the 'enabled' status after lookup, returning or throwing specific 'disabled_resource' errors where appropriate.
• Standardized error responses with the 'disabled_resource' code (HTTP 404) and descriptive messages in API and job results.
• Updated the API error typeset to include 'disabled_resource'.
• Refactored type usage and error handling to propagate and surface the new error code and messaging.

**Affected Areas:**
• API sync controller (sync.controller.ts)
• Jobs package execution logic (sync.ts, action.ts, webhook.ts)
• Shared orchestrator client (orchestrator.ts)
• Shared sync config service (config.service.ts)
• API/Type definitions (types/lib/api.ts)


packages/shared/lib/services/sync/config/config.service.ts
```diff
@@ -759,7 +759,6 @@ export async function getSyncConfigRaw(opts: { environmentId: number; config_id:
             sync_name: opts.name,
             nango_config_id: opts.config_id,
             active: true,
-            enabled: true,
```
bodinsamuel, line 762: is it a leftover from testing?

packages/server/lib/controllers/sync.controller.ts
```diff
@@ -213,6 +213,11 @@ class SyncController {
                 return;
             }
 
+            if (!syncConfig.enabled) {
+                res.status(400).json({ error: { code: 'disabled' } });
```
bodinsamuel: This is an old endpoint, so typing is not really enforced. You'll need to add this new code to types/src/api.ts 
also it might not be the most explicit, can be confused with a disabled endpoint. Maybe `disabled_resource` and/or a more explicit message with the code

TBonnin: Agreed with the name.
I would also return a 404 in this case for 2 reasons:
- it is still expressing that the sync is missing
- it won't be a breaking change

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -686,6 +686,12 @@ export class Orchestrator {
                 name: syncName,
                 isAction: false
             });
+            if (!syncConfig) {
+                throw new Error(`Sync config not found: ${sync.id}`);
```
TBonnin:
```suggestion
                throw new Error(`Sync not found: ${sync.id}`);
```

TBonnin (approved):
small comments about the status code and error name but lgtm otherwise. Once addressed let me know if you need any help for deploying

packages/shared/lib/services/sync/config/config.service.ts
rossmcewan: What do you mean? The enabled check has to be removed.

PR #4147, opened 2025-06-03
fix: adjustable log payload height

rossmcewan:
This PR makes the payload section on the Operation Log details page vertically resizable, allowing users to adjust its height for better usability when viewing large payloads. The fixed, scrollable area is replaced with a container supporting min, default, and max height constraints and 'resize-y' styling.


bodinsamuel (approved):
Nice, one little nitpick you can actually increase the size more than the height of the window which make you lose the pointer maybe a maxHeight can fix that

rossmcewan:
> Nice, one little nitpick you can actually increase the size more than the height of the window which make you lose the pointer maybe a maxHeight can fix that

Hmm. There is a maxHeight on the div already. Will take a look to see why it's not working.

PR #4159, opened 2025-06-04
fix(deploy): refactor template deploy

bodinsamuel:
## Changes

Contributes to https://linear.app/nango/issue/NAN-3309/support-for-old-and-new-integration-template
Contributes to https://linear.app/nango/issue/NAN-3253/put-from-local
Contributes to https://linear.app/nango/issue/NAN-3252/put-from-s3
Contributes to https://linear.app/nango/issue/NAN-3251/load-from-s3

- Split and refactor template deploy
Put it on its own file, only upload one file to simplify the code, handled more errors. 

- Removed `POST /admin/flow/deploy/pre-built`
Discussed with khaliq it's not used, and it's making the codebase much more complexe than intended, mixing templates and private integration (integration that we have developed for customers that they can't access)

- Removed associated CLI command
- Changed some file interface for clarity 

---

This PR is a substantial refactor of how template deployment is structured in the codebase. The endpoint and CLI logic for deploying admin/pre-built flows have been removed, and template deployment has been isolated into a new `deployTemplate` service. File upload/copy operations have been simplified to consistently operate on one file at a time with explicit parameter objects, local/S3 handling is clarified, and multiple legacy or ambiguous logic branches (e.g., mixing private integrations and templates, the admin flow deploy endpoint) are eliminated. Types and file interfaces are reworked for clarity and maintainability, and numerous files are adjusted to support the new API. The `remoteFileService` and `localFileService` APIs are now uniform, redundant or confusing controller logic is stripped, and affected code paths are updated to use the new deploy flow.

**Key Changes:**
• Isolated template deployment into a new `packages/shared/lib/services/deploy/template.ts` file and `deployTemplate` API.
• Removed the `POST /admin/flow/deploy/pre-built` endpoint and corresponding CLI (`admin:deploy`) command.
• All template deploy/copy/upload logic is now handled in one place; file operations consistently take structured parameter objects.
• Clarified file service APIs for both remote (S3) and local scenarios, including changes to destination path conventions.
• Refactored remote and local file service internals, type signatures, and interface to improve explicitness and correctness.
• Simplified deploy logic for templates: now only single-file upload/copy for each artifact (nangoYaml, JS, TS, etc.), better error handling during file operations.
• Removed legacy code and ambiguity between public templates and private integrations.
• Unified and adjusted type definitions for template deployment, splitting or renaming types for better intent.
• Replaced old controller/service routes, CLI code, and validation logic to match the new structure.
• Added a utility file for database update logic related to activation/inactivation of sync configs.

**Affected Areas:**
• Template deployment logic and endpoints
• File upload/copy services: `remoteFileService`, `localFileService`
• Deploy-related CLI commands and service logic
• Server controller logic for flows, deployment, and admin endpoints
• Backend types related to template/config deployment
• Public routing and access middleware (removal of admin deploys)


packages/shared/lib/services/file/remote.service.ts
```diff
@@ -73,19 +76,30 @@ class RemoteFileService {
...
+    async copy({
+        sourcePath,
+        destinationPath,
+        destinationLocalPath
+    }: {
+        sourcePath: string;
+        destinationPath: string;
+        /**
+         * sic
+         * Destination when not uploading to S3, it should be similar to destinationPath but no
```
kaposke: Not sure I understand what this meant

packages/shared/lib/services/deploy/template.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    // fetch the json schema so we have type checking
+    const jsonSchemaString = await remoteFileService.getPublicTemplateJsonSchemaFile(publicRoute);
+    if (!jsonSchemaString) {
+        void logCtx.error('There was an error getting the json schema');
+        await logCtx.failed();
+        return Err(new NangoError('source_copy_error'));
+    }
+
+    const modelsNames = template.models.map((m) => m.name);
```
kaposke, line 108: I was going to say:
```ts
const modelsNames = [...template.models.map((m) => m.name), template.input?.name].filter(Boolean) as string[];
```

But I tested it and... oh boy, `template.models` in this case DOES already have input models. The inconsistencies are driving me crazy. This means that `modelsNames` also contains the input, which is correct for getting all relevant json schemas, but should not be attributed to the `models` column. The `models` column should really be named `outputs`, as it's mapped to `returns` later (which is also an inconsistency).

packages/shared/lib/services/deploy/template.ts
```diff
@@ -0,0 +1,238 @@
...
+
+        if (template.type === 'sync') {
+            const syncs = await getSyncsByProviderConfigKey({
+                environmentId: environment.id,
+                providerConfigKey: deployInfo.integrationId,
+                filter: [{ syncName: template.name, syncVariant: 'base' }]
+            });
+            for (const sync of syncs) {
+                const interval = sync.frequency || template.runs || 'every 30d';
+                const res = await orchestrator.updateSyncFrequency({
```
TBonnin: remind me why the frequency must be updated?

packages/shared/lib/services/deploy/utils.ts
```diff
@@ -0,0 +1,28 @@
...
+                    AND old_config.nango_config_id = active_config.nango_config_id
+                    AND old_config.environment_id = active_config.environment_id
+                WHERE old_config.id = ?
+                    AND active_config.active = true
+            )
+            WHERE sync_config_id = ?`,
+            [oldSyncConfigId, oldSyncConfigId]
+        );
+    });
+}
```
TBonnin, line 28: I would personally prefer to have the sql in the same place, instead of another utils file. I know config.service is a mess but at least I know where to look for in the mess

packages/shared/lib/services/file/remote.service.ts
```diff
@@ -73,19 +76,30 @@ class RemoteFileService {
...
+         */
+        destinationLocalPath: string;
```
TBonnin, line 101: can we just infer the local path by taking the last component of the destinationPath, so we don't have to pass a second param?

packages/types/lib/deploy/incomingFlow.ts
```diff
@@ -32,10 +32,28 @@ export interface LegacySyncModelSchema {
 }
 
 // TODO: split into action | sync type
```
TBonnin: is the TODO still relevant?

kaposke (commented):
Oh boy, I'll have many conflicts with this 😅

TBonnin (approved):

packages/shared/lib/services/deploy/utils.ts
bodinsamuel: yes me too, I just wasn't sure were since it's touching _nango_sync_configs and syncs

packages/shared/lib/services/deploy/template.ts
bodinsamuel: I ported the code, so this one I can't really answer. I wonder if it's necessary indeed.
However I notice a bug in the previous logic, template.name could clash with local script name `getSyncAndActionConfigByParams`, I'll probably need to add a 4th param

packages/shared/lib/services/deploy/template.ts
bodinsamuel: yes, mostly because template and scripts are one of the first code that was written. Unfortunately we are still dealing with this

kaposke: Just remember to not save these in `models` before you merge.

packages/shared/lib/services/file/remote.service.ts
bodinsamuel: it's most of the time, not the same path. In s3 we upload by account, env, config. Locally we upload in a nango-integrations folder. I prefer to keep the logic here agnostic imo

packages/shared/lib/services/file/remote.service.ts
bodinsamuel: see comment but this function has mixed responsibility unfortunately so we use it on our own machine instead of copying to s3 it copies to nango-integrations, hence different path

packages/types/lib/deploy/incomingFlow.ts
bodinsamuel: right :D

packages/shared/lib/services/deploy/template.ts
bodinsamuel: I removed it, I think it was only to handle private integration not template

bodinsamuel:
Tested on staging

PR #4165, opened 2025-06-04
chore(docs): Update webhooks-from-nango.mdx to describe working with `modifiedAfter`

CharlieDigital:
Updated documentation to mention how to work with `modifiedAfter`

 

The documentation for the Nango webhook payloads does not describe how to correctly work with `modifiedAfter`; in fact, this value represents the start of the *current* sync and should be stored for the next sync event

(N/A; docs only)

(N/A; docs only)

---

This PR updates the Nango webhook documentation to clarify the meaning and usage of the `modifiedAfter` field in sync webhook payloads. It introduces an explanatory note on how to use `modifiedAfter` as a sync bookmark, ensuring consumers of the webhook understand that this value should be stored and used for reliable incremental syncing.


docs-v2/guides/webhooks/webhooks-from-nango.mdx
```diff
@@ -120,6 +120,8 @@ Payload received following a successful sync execution:
 }
 ```
 
+The `modifiedAfter` is an ISO-8601 format string (e.g. `2025-05-21T18:52:49.838Z`) that represents the execution time of the Nango sync.  When the webook target application receives this payload, it should store this value against the `connectionId` as a "bookmark".  On the next sync webhook event, use the stored bookmark to retrieve the changed records and store the new `modifiedAfter` as the next bookmark.  This mechanism ensures that no deltas are missed.
```
bodinsamuel:
```suggestion
The `modifiedAfter` is an ISO-8601 format string (e.g. `2025-05-21T18:52:49.838Z`) that represents the start time of the last sync.  When the webhook target application receives this payload, it should store this value against the `connectionId` as a "bookmark".  On the next sync webhook event, use the stored bookmark to retrieve the changed records and store the new `modifiedAfter` as the next bookmark.  This mechanism ensures that no deltas are missed.
```

bodinsamuel (approved):
Look good to me, I'm not entirely sure it's mandatory to store it, maybe I would suggest instead. cc @bastienbeurier

CharlieDigital:
Yeah, it may not be "mandatory", but it seems like if the system does not store it, there is not accurate way to then get the correct set of deltas, no?

I think the thing that I needed from the docs was that this value is pretty much necessary to process the incremental syncs correctly.  Without storing it, there's the possibility of dropping messages in a time gap or repeat retrieval.

bastienbeurier (approved):

PR #4175, opened 2025-06-06
fix(retry-logic): handle both seconds and milliseconds in retry-at logic

hassan254-prog:

- Handle `retry-at` logic when provider sends this back in ms instead of seconds, e.g [exactonline](https://support.exactonline.com/community/s/knowledge-base#All-All-DNO-Simulation-gen-apilimits)

---

This PR updates the retry logic to support 'retry-at' HTTP headers that can be sent as either seconds or milliseconds since epoch. The code now detects the correct unit and normalizes the value to maintain accurate retry timing, and it includes a new unit test to verify correct handling of millisecond-based headers.

**Key Changes:**
• Enhanced parseRetryValue to distinguish and handle both millisecond and second formats in 'retry-at' headers
• Added a unit test covering retry-at logic with millisecond epoch values

**Affected Areas:**
• packages/shared/lib/services/proxy/retry.ts
• packages/shared/lib/services/proxy/retry.unit.test.ts


packages/shared/lib/services/proxy/retry.ts
```diff
@@ -181,10 +181,9 @@ function parseRetryValue({
     if (type === 'at') {
         const currentEpochTime = Math.floor(Date.now() / 1000);
         let retryAtEpoch: number;
-
         const numericValue = Number(rawValue);
         if (!isNaN(numericValue)) {
-            retryAtEpoch = numericValue;
+            retryAtEpoch = numericValue > currentEpochTime + 60 * 60 * 24 * 365 ? Math.floor(numericValue / 1000) : numericValue;
```
bodinsamuel: Mmh not sure it's bulletproof (not sure if any good way to do it).
I have done this in the past =
new Date(timestampInSeconds).getYear() will = 1970 even with a very large number of seconds
while
new Date(timestampInMs) will be >= current year

You could do
```sh
const isMs = new Date(numericValue).getYear() <= 1971;
```

bodinsamuel (approved):

packages/shared/lib/services/proxy/retry.ts
hassan254-prog: I tested it, and it works 😊 thanks.

bodinsamuel (approved):

PR #4181, opened 2025-06-10
feat(utils): add NangoModel to json-schema mapping

kaposke:
### Description
I'm pulling the logic from the closed (not merged) [PR](https://github.com/NangoHQ/nango/pull/4134) I did for MCP to convert `NangoModel` to json-schema. I've split the json-schema utilities just to make the files smaller.

I plan to use it for properly displaying snippets for non-deployed prebuilt flows, which come from `flows.yaml` so theres no json-schema anywhere for them. It's a temporary solution while we consider the next steps for `flows.yaml`. It means I can't completely remove `NangoModel` because I need it to parsed from `flows.yaml`, but it's an ok compromise for now and allows me to put and end to this chapter.

I'm adding these utilities in this isolated PR to not bloat the full migration PR.

---

This PR introduces a new `nangoModelsToJsonSchema` function within the `nango-yaml` package, enabling conversion of arrays of NangoModel objects into JSON Schema (JSONSchema7) format. The utility robustly supports primitive types, arrays, union types, nested and circular model references, and optional fields. The implementation also includes comprehensive unit test coverage using Vitest with snapshot testing, verifying correctness across a wide range of schema scenarios-including primitives, unions, complex arrays, empty input, empty models, and recursive structures.

**Key Changes:**
• Implemented `nangoModelsToJsonSchema` function to output JSON Schema with definitions per model.
• Added support for primitive, model, array, and union types as well as nested and circular references using .
• Extracted and used a static primitive type map for maintainability.
• Created extensive unit tests with Vitest, including snapshot files, to validate all conversion logic and edge cases.

**Affected Areas:**
• packages/nango-yaml/lib/json-schema.ts
• packages/nango-yaml/lib/json-schema.unit.test.ts
• packages/nango-yaml/lib/__snapshots__/json-schema.unit.test.ts.snap


packages/utils/lib/json-schema/nango-model-to-json-schema.ts
```diff
@@ -0,0 +1,102 @@
...
+    try {
+        return Ok(nangoModelToJsonSchemaInternal(model, models_schema, new Set()));
+    } catch (err) {
+        return Err(err as Error);
+    }
+}
+
+function nangoModelToJsonSchemaInternal(model: NangoModel, models_schema: NangoModel[], visitedModels: Set<string>): JSONSchema7 {
+    if (visitedModels.has(model.name)) {
+        throw new Error(`Circular reference detected: ${Array.from(visitedModels).join(' -> ')} -> ${model.name}`);
```
TBonnin: can we comment why circular ref are not supported so we remember in 3 months :wink:

packages/utils/lib/json-schema/nango-model-to-json-schema.ts
```diff
@@ -0,0 +1,102 @@
...
+            format: 'date-time'
+        };
+    }
+
+    return {
+        type: 'object',
+        additionalProperties: {
+            type: 'string'
+        }
+    };
```
TBonnin:
```suggestion
    const typeMap = {
        number: { type: 'number' },
        boolean: { type: 'boolean' },
        string: { type: 'string' },
        date: { type: 'integer', format: 'date-time' }
    };

    return typeMap[field.value] || {
        type: 'object',
        additionalProperties: {
            type: 'string'
        }
    };
```

TBonnin (commented):
should this logic live in `nangoYaml` package where NangoModel is coming from?

packages/utils/lib/json-schema/nango-model-to-json-schema.ts
bodinsamuel: if you go this way, you can store the static object outside this function

packages/utils/lib/json-schema/nango-model-to-json-schema.ts
bodinsamuel: circular is supported by using #ref no?

bodinsamuel (approved):

packages/utils/lib/json-schema/nango-model-to-json-schema.ts
kaposke: Yeah, my bad for the oversight. This was MCP specific, we should support it fully now. 
I'll: 
1. Modify it to map references
2. Move it to the `nangoYaml` package.
3. Simplify some of the logic as you suggested below (I had a similar typemap somewhere, must have been lost in my iterations)

PR #4184, opened 2025-06-11
fix(cli): zeroyaml feedback and test

bodinsamuel:
## Changes

Few feedback from for real life test

- Handle string that needs quote in models
- Ignore migrating known file like models.ts or schema.ts to avoid errors
- Remove generic from getMetadata
- Allow passing custom interface for batchX 
- Missing return type for getRecordsById

---

This PR delivers several real-world-driven enhancements to the ZeroYaml migration logic and TypeScript type definitions in the runner SDK. Key updates include robust model property key handling during code generation, the addition of reserved word and identifier validation, stricter file migration management, more flexible batch method typings, and enhanced compile utilities and error messaging. Related test files and snapshots are updated to cover these adjustments and ensure reliability.

**Key Changes:**
• Model property names in generated code now use string literals when not valid JS identifiers or reserved words (via new isValidIdentifier helper and reserved word Set).
• Migration now ignores 'models.ts' and '.nango/schema.ts' files to prevent known errors.
• batchSave, batchDelete, and batchUpdate methods in NangoSyncBase support custom interfaces (RawModel), increasing SDK flexibility; getRecordsByIds return type clarified.
• getMetadata generic is removed for signature consistency.
• Error messaging improved for missing entry points during compile.
• Tests and snapshots updated for stricter type expectations and edge-case model key handling.

**Affected Areas:**
• cli/lib/migrations/toZeroYaml.ts (migration core logic, key quoting, file ignore, helper function additions)
• runner-sdk/lib/sync.ts (batch method typing changes)
• runner-sdk/lib/types.ts (RawModel type introduced)
• runner-sdk/lib/sync.unit.test.ts (type assertion tests updated)
• cli/lib/zeroYaml/compile.ts (compile error message clarity)
• migrations unit test and snapshots


packages/cli/lib/migrations/toZeroYaml.ts
```diff
@@ -1146,6 +1157,83 @@ function isImportSpecifier(s: unknown): s is ImportSpecifier {
...
 
+// Helper to check if a string is a valid JavaScript identifier
+function isValidIdentifier(name: string): boolean {
+    if (!name || typeof name !== 'string') {
+        return false;
+    }
+    // Must start with letter, underscore, or dollar sign
+    if (!/^[a-zA-Z_$]/.test(name)) {
+        return false;
+    }
```
TBonnin: isn't the regex below also testing that first char is valid?

packages/cli/lib/migrations/toZeroYaml.ts
```diff
@@ -1146,6 +1157,83 @@ function isImportSpecifier(s: unknown): s is ImportSpecifier {
...
+        'int',
+        'long',
+        'native',
+        'short',
+        'synchronized',
+        'throws',
+        'transient',
+        'volatile'
+    ];
+    return !reservedWords.includes(name);
```
TBonnin: nit: use a `Set` and `has` for some "constant" lookup

TBonnin (approved):

packages/cli/lib/migrations/toZeroYaml.ts
bodinsamuel: right

packages/cli/lib/migrations/toZeroYaml.ts
bodinsamuel: I let AI slop fool me

PR #4187, opened 2025-06-12
fix(proxy): fix build proxy headers

hassan254-prog:

- We have a couple of providers that require the `user-agent` header to be passed in a specific format. A good example is Rippling, which has a policy prohibiting third-party tools like Nango from making direct requests to their API. To comply with this, we need to obfuscate all outbound identifiers.
Instead of requiring users to manually set this header within their scripts, we can pull it from the `providers.yaml` configuration. This ensures the correct header is applied to all requests, including those made by syncs, actions, and proxy calls to the provider.

---

This PR modifies the proxy header handling logic to ensure that the 'user-agent' header, as defined in each provider's configuration (specifically providers.yaml), is always respected, even when other custom headers are set by user scripts. The change enforces that special headers like 'user-agent' are not overridden by direct script-supplied headers, addressing provider-specific requirements such as Rippling's API policy.

**Key Changes:**
• Updated buildProxyHeaders function to prioritize 'user-agent' from providers.yaml, preventing it from being overridden by user-supplied config.headers.
• Added 'user-agent: Node.js/20.12.2 (linux; x64)' to Rippling's proxy headers in providers.yaml.

**Affected Areas:**
• Proxy header construction logic ('buildProxyHeaders' in utils.ts)
• Provider configuration for Rippling (providers.yaml)


packages/shared/lib/services/proxy/utils.ts
```diff
@@ -318,7 +318,15 @@ export function buildProxyHeaders({
...
     if (config.headers) {
-        // Headers set in scripts should override the default ones
+        // Headers set in scripts should override the default ones except for special headers like 'user-agent'
+        const specialHeaders = ['user-agent'];
+        for (const key of specialHeaders) {
+            const lowerKey = key.toLowerCase() as Lowercase<string>;
+            if (headers[lowerKey]) {
+                config.headers[key] = headers[lowerKey];
+            }
+        }
```
bodinsamuel:
```suggestion
        const specialHeaders: Lowercase<string>[] = ['user-agent'];
        for (const key of specialHeaders) {
            if (headers[key]) {
                config.headers[key] = headers[key];
            }
        }
```

packages/shared/lib/services/proxy/utils.ts
```diff
@@ -318,7 +318,15 @@ export function buildProxyHeaders({
     }
 
     if (config.headers) {
-        // Headers set in scripts should override the default ones
+        // Headers set in scripts should override the default ones except for special headers like 'user-agent'
+        const specialHeaders = ['user-agent'];
```
bodinsamuel: This can be stored outside this function

bodinsamuel (approved):

packages/shared/lib/services/proxy/utils.ts
hassan254-prog: yes, I have updated that with the latest [commit](https://github.com/NangoHQ/nango/pull/4187/commits/f69b094a7fc8465192a2cece24541bed2a46ea90) 🙏 .

packages/shared/lib/services/proxy/utils.ts
```diff
@@ -43,6 +43,7 @@ export class ProxyError extends Error {
 }
 
 const methodDataAllowed = ['POST', 'PUT', 'PATCH', 'DELETE'];
+const specialHeaders: Lowercase<string>[] = ['user-agent'];
```
hassan254-prog: @bodinsamuel, just a quick check, are there any outbound identifiers besides `user-agent`? I haven't seen any others.

bodinsamuel: not sure

hassan254-prog: I couldn't identify any others from the request either ☺️, maybe @TBonnin could confirm. 
Screenshot 2025-06-12 at 16 16 53

TBonnin: not sure I understand the question 😬

TBonnin: `providedHeaders` might be a slightly better name than `special`

TBonnin (approved):

packages/shared/lib/services/proxy/utils.ts
hassan254-prog: Rippling has a policy that prohibits third-party tools like Nango from making direct requests to their API. To comply with this, we need to obfuscate all outbound identifiers that might reveal the requests are coming from Nango. This isn’t a technical restriction—it doesn't prevent requests from going through at the moment—but to be safe, it's a good idea to obfuscate any identifiers, such as the user-agent, that could expose our identity. I was asking whether there are any other such identifiers we should be aware of?  I guess the ip but there is no way for us to hide that 😄 .

bodinsamuel: I think we just need to try and see

PR #4188, opened 2025-06-12
fix(file): correctly zip from s3

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-3255/fix-zip-s3

- Correctly zip zeroyaml script
- Ensure a flow do exists before initiating a download
- Fixes integrationName -> scriptName
That's my bad

---

This PR addresses a bug where integration files (scripts and YAML) were inconsistently zipped and downloaded from S3. The changes unify the zipping logic by introducing a flexible array-based approach for collecting files to zip, switch all parameters from 'integrationName' to 'scriptName' for improved naming consistency, and ensure existence of a flow before serving downloads for public templates.

**Key Changes:**
• Refactored remote and local zipping logic to use a generic array-of-files approach, improving extensibility and consistency for both S3 and local sources.
• Replaced all references of 'integrationName' with 'scriptName' throughout the affected services and controllers.
• In controller, added existence check for the flow before download in public flow cases.
• Error handling was improved during file lookup and zipping, with appropriate status codes and error responses on failure.

**Affected Areas:**
• Remote file service (S3 zip logic)
• Local file service (local zip logic)
• Flow controller (download endpoint logic)

### DIFF
--- a/packages/shared/lib/services/file/remote.service.ts
+++ b/packages/shared/lib/services/file/remote.service.ts
@@ -195,83 +195,81 @@ class RemoteFileService {
 
     async zipAndSendPublicFiles({
         res,
-        integrationName,
+        scriptName,
         providerPath,
         flowType
     }: {
         res: Response;
-        integrationName: string;
+        scriptName: string;
         providerPath: string;
         flowType: string;
     }): Promise<void> {
+        // TODO: handle zero yaml here
+
+        const files: { name: string; content: Readable }[] = [];
         const { success, error, response: nangoYaml } = await this.getStream(`${this.publicRoute}/${providerPath}/${nangoConfigFile}`);
         if (!success || nangoYaml === null) {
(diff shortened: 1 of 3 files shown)


packages/shared/lib/services/file/remote.service.ts
```diff
@@ -195,83 +195,81 @@ class RemoteFileService {
...
         providerPath,
         flowType
     }: {
         res: Response;
-        integrationName: string;
+        scriptName: string;
         providerPath: string;
         flowType: string;
     }): Promise<void> {
+        // TODO: handle zero yaml here
```
bodinsamuel, line 207: for template we don't know yet if they are zero or not. Not a big issue rn since there aren't one deployed

bodinsamuel:
Tested in staging

TBonnin (approved):

PR #4192, opened 2025-06-12
feat(billing): don't bill first month of MAR for connection

kaposke:
### Describe
We want to stop billing for MAR on the first month of any given connection, to stop scaring customers with huge initial sync charges https://linear.app/nango/issue/NAN-3410/exclude-first-month-of-monthly-active-records-from-billing

There are multiple ways of doing this. I went with:
<strike>Query connection age on persist and provide it as data to the records package who ultimately decides what's billable.</strike>
- Renamed billedKeys in the `records` package to `activatedKeys`.
- On `persistRecords`, we only bill `activatedKeys` if the connection is older than 30 days

---

This PR fundamentally changes the billing logic for Monthly Active Records (MAR) to prevent billing during the first 30 days of a connection, addressing concerns about large charges on initial sync. To achieve this, the full connection object (including `created_at`) is now propagated through all relevant persistence, upsert, update, and test flows, allowing all logic related to record billing to conditionally suppress charges for new connections. Code across the records, persistence, jobs execution, server API controller, and related tests are updated or refactored to consistently pass connection objects, check connection age, and ensure this new policy is respected. Comprehensive test coverage was added to verify billing is skipped for new connections and still occurs for older ones.

**Key Changes:**
• MAR billing logic updated to skip billing if connection is under 30 days old.
• Refactored `persistRecords`, `upsert`, and `update` to require and use the full connection object instead of just connectionId.
• Pulled connection data (including creation date) via shared `connectionService` at persistence time.
• Removed redundant per-record or per-call passing of connectionId and replaced with the connection object throughout code and tests.
• Documentation and inline comments explain new MAR billing exclusion rules.
• Extended and refactored tests to cover new/old connections and their billing, with utilities updated to support the changed interfaces.

**Affected Areas:**
• Billing logic for records (MAR calculation)
• Record persistence, upsert, and update logic (records/models/records.ts, persist/records.ts)
• API controller and job execution layers interacting with records
• All related integration and model-layer tests
• Test utilities and seeding logic


packages/persist/lib/records.ts
```diff
@@ -98,6 +121,16 @@ export async function persistRecords({
...
     }
 
+    // Get connection creation date for billing logic
+    const connectionResult = await db.knex.from('_nango_connections').select<{ created_at: Date }>('created_at').where('id', nangoConnectionId).first();
+    if (!connectionResult) {
+        const err = new Error(`Connection ${nangoConnectionId} not found`);
+        void logCtx.error('Connection not found', { error: err, persistType });
+        span.setTag('error', err).finish();
+        return Err(err);
+    }
```
TBonnin: I would prefer that all the db logic lives with the rest of the db queries for connections in shared connection.service

packages/records/lib/models/records.ts
```diff
@@ -301,14 +301,16 @@ export async function upsert({
     environmentId,
     model,
     softDelete = false,
-    merging = { strategy: 'override' }
+    merging = { strategy: 'override' },
+    connectionCreatedAt
```
TBonnin: instead of passing `connectionId` and `connectionCreatedAt`, let's just pass the entire connection

packages/records/lib/models/records.ts
```diff
@@ -419,7 +421,10 @@ export async function upsert({
                     const undeletedKeys = undeletedRes.map((r) => r.external_id);
                     const addedKeys = insertedKeys.concat(undeletedKeys);
                     const updatedKeys = changedRes.map((r) => r.external_id);
-                    const billableKeys = [...insertedKeys, ...billable(changedRes).map((r) => r.external_id)];
+
+                    // Check if connection is younger than 30 days - if so, skip all billing
+                    const shouldSkipBilling = dayjs().utc().diff(dayjs(connectionCreatedAt).utc(), 'days') < 30;
```
TBonnin: do we need to convert to utc? isn't dayjs dealing with tz automatically. 
also does connectionCreatedAt need to be wrapped? isn't `dayjs.diff` properly handling `Date`?

packages/records/lib/models/records.ts
TBonnin: another option would be to keep connectionId and pass an additional `skipBilling` parameter

packages/persist/lib/records.ts
TBonnin: that also sucks we have to make an extra db query for each batch :(
I guess it is super cachable if we ever encounter perf issues

packages/records/lib/models/records.ts
kaposke: Yeah I don't think we do. This is AI slop I've missed

packages/persist/lib/records.ts
```diff
@@ -98,6 +121,16 @@ export async function persistRecords({
         return Err(err);
     }
 
+    // Get connection creation date for billing logic
+    const connectionResult = await db.knex.from('_nango_connections').select<{ created_at: Date }>('created_at').where('id', nangoConnectionId).first();
```
bodinsamuel: We can take the opportunity to fix this:
.where(deleted, false)

bodinsamuel (approved):
Looks good 👍🏻 

Not you but this introduce complexity in the billing and more ways to game it too, a free tier would be so much simpler to understand imo @bastienbeurier

bastienbeurier:
> Looks good 👍🏻
> 
> Not you but this introduce complexity in the billing and more ways to game it too, a free tier would be so much simpler to understand imo @bastienbeurier

The problem of the free tier is that the amount of data in the 1st sync varies drastically. So there could still be some scary month-1 bills with the free tier. Also we know that Fivetran does this, so there's a mature precedent.

bastienbeurier:
@kaposke just checking that:
- this is performant enough to support our syncing scale
- this plays well with Thomas' aggregation logic (I don't see why not)
- we'll be able to see in Datadog billed MAR AND total MAR separately

packages/persist/lib/records.ts
kaposke: Yeah, I agree it sucks to do an extra DB query, but I guess there's not much we can do? We could take more connection info from the payload, I guess? Does the runner have the full connection info?

packages/records/lib/models/records.ts
kaposke: Changed behavior (updated PR description)

packages/persist/lib/records.ts
```diff
@@ -59,6 +59,16 @@ export async function persistRecords({
...
+    const logCtx = logContextGetter.getStateLess({ id: String(activityLogId), accountId });
+
+    const connection = await connectionService.getConnectionById(nangoConnectionId);
+    if (!connection) {
+        const err = new Error(`Connection ${nangoConnectionId} not found`);
+        void logCtx.error('Connection not found', { error: err, persistType });
+        span.setTag('error', err).finish();
+        return Err(err);
+    }
+
```
TBonnin, line 71: getting the connection is purely internal logic. I don't think we should expose that to the user, or even fail.

packages/persist/lib/records.ts
```diff
@@ -154,13 +164,18 @@ export async function persistRecords({
             return acc;
         }, 0);
 
-        const mar = new Set(summary.billedKeys).size;
+        const thirtyDaysAgo = new Date();
+        thirtyDaysAgo.setDate(thirtyDaysAgo.getDate() - 30);
+        const isConnectionOlderThan30Days = new Date(connection.created_at) < thirtyDaysAgo;
+        const shouldBillMonthlyActiveRecords = isConnectionOlderThan30Days;
```
TBonnin, line 170: why not a single `const`?

packages/records/lib/types.ts
```diff
@@ -58,7 +58,10 @@ export interface UpsertSummary {
     updatedKeys: string[];
     deletedKeys?: string[];
     unchangedKeys: string[];
-    billedKeys: string[];
+    /**
+     * MAR - Monthly Active Records - Carries keys that have been added/modified from the start of the month
+     */
+    activatedKeys: string[];
```
TBonnin, line 64: what about `activeKeys` since we talk about monthly **active** records

packages/records/lib/models/records.ts
```diff
@@ -37,18 +37,18 @@ interface UpsertResult {
     status: 'inserted' | 'changed' | 'undeleted' | 'deleted' | 'unchanged';
 }
 
-function isBillable(record: { last_modified_at: string | Date; previous_last_modified_at: string | Date }): boolean {
+function isInactiveThisMonth(record: { last_modified_at: string | Date; previous_last_modified_at: string | Date }): boolean {
```
TBonnin, line 40: I would keep the `isBillable` name. It is still describing what make a record billable (ie: touched this month). It happens that we skip the billable records when connection is too young but that doesn't change what make a record billable or not.
Or at least let's call it `isActive`

packages/persist/lib/records.ts
kaposke: I sometimes like to do this sort of thing in favor of readability. One can immediately understand this without comments.

packages/records/lib/types.ts
kaposke: Could be an interesting metric.

kaposke: This is the list of records that got activated by this particular update operation. So `activeKeys` could be misleading. I'm trying to be very explicit in all my naming.

packages/records/lib/models/records.ts
kaposke: I'm against going back to `isBillable`. The `records` package doesn't really understand what is billed down the line. Furthermore, we now have metrics for total MAR and Billed MAR. This distinction is made in the `persist` package. For all intents and purposes, the `records` package is providing data on what are `new MAR`, or `activated` records (records that got activated by this upsert).

As for `isActive`, that's not what the function checks for. I went with `isInactiveThisMonth` because it's very literal on what it does.

I believe it all reads nice and makes whats happening very explicit:
```ts
const activatedKeys = [...insertedKeys, ...getInactiveThisMonth(changedRes).map((r) => r.external_id)];
```

While before this I had to really take my time to understand what was happening.

packages/persist/lib/records.ts
kaposke: Is it? If a sync is running for a non-existing connection, it should probably fail, right?

TBonnin: yes but it will be caught way earlier so I wouldn't bother with customer-facing logs for it in persist

packages/records/lib/models/records.ts
TBonnin: I would find `wasInactiveThisMonth` to be more explicit then. up to you though.

TBonnin (approved):

bodinsamuel (approved):

packages/records/lib/models/records.ts
kaposke: It uses `previous_modified_at`, which is the date before the update at hand, to see if it was NOT modified this month (except for right now at the occurring update).
So you could read it as:
The keys that are activated are:
- the new inserted keys
- the keys that were modified that were previously inactive this month.

PR #4193, opened 2025-06-12
refactor(scheduler): migrate from workers to in-process execution

TBonnin:
Running scheduling, cleaning and expiring logic in worker threads brings very little benefit. Any failure should be treated as critical and should cause the orchestrator to exit
This commit is refactoring scheduler, removing worker threads which add more complexity (messages passing vs direct function calls, multiple db connection pools, ...) in favor of simple promises.

---

This PR performs a significant refactor of the background task scheduling and orchestration system. It migrates all scheduling, cleaning, and expiring logic from Node.js worker threads to an in-process, promise-driven daemon model managed by AbortController, eliminating complexity due to message passing, additional DB connection pools, and worker bootstrapping. The Scheduler API and its integration into the orchestrator app have been overhauled, introducing a unified, synchronous in-process management flow with tighter error handling, improved testability, and simplified lifecycle management. Extensive updates are made across scheduler infrastructure, daemon implementation, environment configuration, integration tests, and orchestrator startup/shutdown logic.

**Key Changes:**
• Fully removes worker-thread code: worker classes, monitor/scheduling/cleanup workers, and all boot scripts.
• Introduces SchedulerDaemon base class and concrete ExpiringDaemon, SchedulingDaemon, and CleaningDaemon, all running in-process.
• Refactors Scheduler to use the new daemons, injecting direct DB and error/callback handlers, with lifecycle managed by AbortController.
• All background work now runs within the main process; any error in a background daemon causes orchestrator shutdown.
• Removes message-based state propagation between threads; task state transitions now use direct callbacks.
• Refactors and extends tests for scheduler, orchestrator, events, helper utilities, and daemon class coverage.
• Updates environment variable handling: replaces MONITOR_TICK and CLEANUP_TICK with new EXPIRING_TICK and CLEANING_TICK settings.
• Cleans up and unifies code style, reduces indirection, and clarifies lifecycle control in orchestrator/app.ts.
• Significant updates in integration tests and helpers for the new direct-invocation daemon model.

**Affected Areas:**
• packages/scheduler/lib/scheduler.ts (core Scheduler logic)
• packages/scheduler/lib/daemons/ (new daemons and abstract base)
• packages/orchestrator/lib/app.ts (app entrypoint, error handling, shutdown logic)
• all affected test suites (daemon/integration, scheduler, orchestrator, helpers)
• environment config parsing and validation
• removal of workers/ and corresponding scripts


packages/scheduler/lib/daemons/cleaning/cleaning.daemon.ts
```diff
@@ -1,29 +1,18 @@
-import type { MessagePort } from 'node:worker_threads';
 import * as tasks from '../../models/tasks.js';
 import * as schedules from '../../models/schedules.js';
 import type knex from 'knex';
 import { logger } from '../../utils/logger.js';
-import { SchedulerWorker, SchedulerWorkerChild } from '../worker.js';
+import { SchedulerDaemon } from '../daemon.js';
```
TBonnin, line 5: went for `daemon` name because I didn't want to use worker to make it clear they are not worker threads. Other name like `processor`, `jobs`, `tasks` are already used in the codebase for other concepts. `daemon` express the idea of background processing so I went for it

packages/scheduler/lib/daemons/expiring/expiring.daemon.ts
```diff
@@ -0,0 +1,46 @@
...
+        if (expired.isErr()) {
+            logger.error(`Error expiring tasks: ${stringifyError(expired.error)}`);
+        } else {
+            if (expired.value.length > 0) {
+                for (const task of expired.value) {
+                    this.onTaskStateChange(task);
+                }
+                logger.info(`Expired tasks: ${JSON.stringify(expired.value.map((t) => t.id))}`);
+            }
+        }
```
kaposke: Guard clauses ftw
```suggestion
    async run(): Promise<void> {
        const expired = await tasks.expiresIfTimeout(this.db);
        if (expired.isErr()) {
            logger.error(`Error expiring tasks: ${stringifyError(expired.error)}`);
            return;
        }
        if (expired.value.length > 0) {
            for (const task of expired.value) {
                this.onTaskStateChange(task);
            }
            logger.info(`Expired tasks: ${JSON.stringify(expired.value.map((t) => t.id))}`);
        }
      }
```

kaposke (approved):
I wasn't familiar with this, but oh man is it a great cleanup.

bodinsamuel (approved):
Works well, I wonder if it's going to be better local performance.

PR #4199, opened 2025-06-16
fix(pagination): fix link pagination type

hassan254-prog:

- Don’t break the loop early on an empty response. Some providers may return an empty response but still include a next link. We should only break the pagination loop after all pages have been processed, specifically, when there is no next link.

---

This PR updates the link pagination logic so the loop no longer breaks early on empty responses if a 'next' link is present. Now, pagination continues until both the response data is empty and there is no next page link, ensuring that all available pages are processed as intended.


TBonnin (approved):
> Some providers may return an empty response but still include a next link

🤷 
also I want names 😉 

Should we check if nextLink is not the same as current url in order to avoid infinite loop?

hassan254-prog:
> > Some providers may return an empty response but still include a next link
> 
> 🤷 also I want names 😉
> 
> Should we check if nextLink is not the same as current url in order to avoid infinite loop?

Found this use case in Microsoft Graph API, specifically with OneDrive. Yes, I've updated this with the [latest](https://github.com/NangoHQ/nango/pull/4199/commits/05dbe58092826189eefe3cccfe69fc97539c0d03) commit 🙏 .

TBonnin (approved):

PR #4201, opened 2025-06-16
chore(orchestrator): tasks events via pg LISTEN/NOTIFY

TBonnin:
in order to make orchestrator multi-instances, task event notifications cannot be only emitted/received locally.
This commit is using postgres LISTEN/NOTIFY mechanism as a way to broadcast the task events (task created, completed, ...) to all instances of the orchestrator

---

This PR replaces local-only EventEmitter-based task event handling in the orchestrator with a new PgEventEmitter system that leverages PostgreSQL LISTEN/NOTIFY for broadcasting task events (e.g., task created, started, completed) across multiple orchestrator instances. It introduces the PgEventEmitter class, modifies TaskEventsHandler to use this abstraction, updates event naming conventions, adapts API route handlers and test/integration code, and adds an extensive integration test suite to validate event delivery, error conditions (including payload limitations), and distributed behaviors. Changes include constructor and invocation signature updates for event handlers, improved connection lifecycle management, and enforcement of best practices around PostgreSQL NOTIFY's 8KB payload limit.

**Key Changes:**
• Replaced Node.js EventEmitter-based event propagation with PgEventEmitter using PostgreSQL LISTEN/NOTIFY for inter-instance broadcast.
• Introduced new PgEventEmitter class to manage notification emission, subscription, connection errors, reconnection, and payload sizing.
• Refactored TaskEventsHandler to extend PgEventEmitter and updated callbacks to only emit lightweight event payloads (e.g., taskId) due to NOTIFY payload limits.
• Updated all usage points, server/app bootstrap, and API route handlers (e.g., getOutput, postDequeue) to use new event abstractions and event naming.
• Enhanced integration and unit tests with scenarios covering connection lifecycle, event parsing, payload too large, and multiple orchestrator instances.
• Modified test harnesses and helper classes to work with new TaskEventsHandler constructor and behavior.
• Ensured proper disconnect/release of PgEventEmitter on orchestrator shutdown to avoid PG connection leaks.

**Affected Areas:**
• TaskEventsHandler and events.ts module
• Orchestrator service/app bootstrap logic (app.ts)
• API endpoints/routes relying on task event notifications (getOutput, postDequeue, etc.)
• Integration and unit tests (including helpers, processor, client, and API tests)
• TestOrchestratorService and other test harnesses

### DIFF
--- a/packages/orchestrator/lib/routes/v1/tasks/taskId/getOutput.ts
+++ b/packages/orchestrator/lib/routes/v1/tasks/taskId/getOutput.ts
@@ -7,6 +7,7 @@ import type { ApiError, Endpoint } from '@nangohq/types';
 import type { EndpointRequest, EndpointResponse, Route, RouteHandler } from '@nangohq/utils';
 import type { EventEmitter } from 'node:events';
 import type { JsonValue } from 'type-fest';
+import { taskEvents } from '../../../../events.js';
 
 type GetOutput = Endpoint<{
     Method: typeof method;
@@ -37,7 +38,7 @@ const validate = validateRequest<GetOutput>({
 const handler = (scheduler: Scheduler, eventEmitter: EventEmitter) => {
     return async (_req: EndpointRequest, res: EndpointResponse<GetOutput>) => {
         const longPollingTimeoutMs = res.locals.parsedQuery.longPolling || 120_000;
-        const eventId = `task:completed:${res.locals.parsedParams.taskId}`;
+        const eventId = taskEvents.taskCompleted(res.locals.parsedParams.taskId);
         const cleanupAndRespond = (respond: (res: EndpointResponse<GetOutput>) => void) => {
             if (timeout) {
                 clearTimeout(timeout);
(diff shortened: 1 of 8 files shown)


packages/orchestrator/lib/routes/v1/tasks/taskId/getOutput.ts
```diff
@@ -49,8 +50,13 @@ const handler = (scheduler: Scheduler, eventEmitter: EventEmitter) => {
...
         };
-        const onCompletion = (completedTask: Task) => {
-            cleanupAndRespond((res) => res.status(200).json({ state: completedTask.state, output: completedTask.output }));
+        const onCompletion = async (taskId: Task['id']) => {
+            const completedTask = await scheduler.get({ taskId });
+            if (completedTask.isErr()) {
+                cleanupAndRespond((res) => res.status(404).json({ error: { code: 'task_not_found', message: completedTask.error.message } }));
+                return;
+            }
+            cleanupAndRespond((res) => res.status(200).json({ state: completedTask.value.state, output: completedTask.value.output }));
```
TBonnin, line 59: here is the drawback of using pg. Before we could pass the entire task as the argument of the event when everything was happening locally, but pg has a 8kb limit for the payload of NOTIFY. We therefore only passing the taskId and retrieve the state and output by querying the db

bodinsamuel (approved):
Looks good, one big question about LISTEN/NOTIFY it's a fan out process so if you have 2 orch both will receive the events, might have missed something but not sure it's handled here

TBonnin:
> Looks good, one big question about LISTEN/NOTIFY it's a fan out process so if you have 2 orch both will receive the events, might have missed something but not sure it's handled here

you are right but it is on purpose. For example taking the 2 events that currently exists:
- `task:completed:12345` -> events is broadcasted to all instances via pg and then re-emitted locally on all instances. If there is a local listener for this event it will be handled and the output of the task will be returned. On the other end, if an instance has no listener for this specific event it will be dropped
- `task:created:groupA` -> events is broadcasted to all instances that get notified there are some tasks ready to be picked up for this group. Each instances, if in the middle of a long polling dequeue, will try again to dequeue tasks. It is possible that one instance will dequeue all tasks and the other instances will dequeue nothing and waste a call but it is still better than dequeuing in a very tight loop.

PR #4203, opened 2025-06-17
fix(api): create billing customer on account creation

bodinsamuel:
## Changes

Contributes to https://linear.app/nango/issue/NAN-3444/app-show-usage-for-free-plans

- Create a customer on Orb as soon as the user register in Nango
It's handy for self-serve billing, and we'll be handy to show usage for free user too.

## Tests

- `FLAG_USAGE_ENABLED=true`
- Add `ORB_API_KEY`
- Create a new account
- Go to orb > customers > all customers, you should see yourself

---

This PR ensures that a billing customer is created or updated (upserted) in Orb as soon as a user or team is registered in the app, covering both direct and SSO/managed onboarding. The billing abstraction is expanded to provide an idempotent 'upsertCustomer' method, and both standard and SSO onboarding flows now invoke this to synchronize customer state in Orb. This foundational change supports future usage tracking and billing for all accounts, including free plans, and incorporates robust error handling to avoid blocking account creation if external billing APIs fail.

**Key Changes:**
• Introduced 'upsertCustomer' method to Billing abstraction and Orb client for idempotent customer creation or update.
• Updated account signup (standard and managed) flows to call billing.upsertCustomer upon new team creation.
• Refactored type and interface definitions in billing/types to support new billing flows.
• Added error logging/reporting for billing customer creation failures, which no longer block signup.
• Improved code organization in billing and controller files.
• Minor script consistency fix in .husky/pre-commit.

**Affected Areas:**
• Account signup (standard and SSO managed controllers)
• Billing service and Orb integration
• Billing client and type definitions
• .husky/pre-commit script


packages/billing/lib/clients/orb.ts
```diff
@@ -28,6 +28,20 @@ export class OrbClient implements BillingClient {
         }
     }
 
+    async createCustomer(team: DBTeam, user: DBUser): Promise<Result<BillingCustomer>> {
+        try {
+            const customer = await this.orbSDK.customers.create({
+                external_customer_id: String(team.id),
+                currency: 'USD',
+                name: team.name,
```
TBonnin, line 50: should the name in orb be updated when an user modifies it in Nango?

packages/server/lib/controllers/v1/account/managed/getCallback.ts
```diff
@@ -71,49 +74,55 @@ export const getManagedCallback = asyncWrapper<GetManagedCallback>(async (req, r
...
             res.status(500).send({ error: { code: 'error_creating_user', message: 'There was a problem creating the user. Please reach out to support.' } });
             return;
         }
+
+        if (isNewTeam) {
+            const resCreate = await billing.createCustomer(account, user);
+            if (resCreate.isErr()) {
+                report(resCreate.error);
+            }
+        }
```
TBonnin, line 125: would it be possible to make createCustomer idempotent (aka: no op if one with same accountId already exists in Orb) and call it in user service when the user is verified? This way we only create a customer in orb if user is verified and it is all happening in a single file

bodinsamuel: I'm not sure we care about user being verified or not tbh and it's not related to userService. 
But I'll add idem potency that will simplify the current code

packages/billing/lib/clients/orb.ts
bodinsamuel: yes it's planned for later

kaposke (approved):
If it fails we lose the timing and have to handle non-existent accounts later when trying to submit the first metric. Should we go this path and check for the error to be "non-existing customer" when sending metrics?

bodinsamuel:
> If it fails we lose the timing and have to handle non-existent accounts later when trying to submit the first metric. Should we go this path and check for the error to be "non-existing customer" when sending metrics?

Orb will not fail if we send events for non-existing customers and it would be inefficient to check at each event. That's where a job queue would be interesting because we don't want this to block the account creation but we still want this to eventually execute

packages/server/lib/controllers/v1/account/managed/getCallback.ts
TBonnin: if we don't wait for verification it means we are gonna have "ghost" customers in Orb, who have not verified their email and are not actual users of Nango. Not the end of the world but would be cleaner imho to only create a customer when an account is actually "active"

PR #4206, opened 2025-06-17
chore(persist): add tracing in auth middleware

TBonnin:
first step to narrow down where time is being spent in persist auth middleware

---

This PR introduces Datadog tracing to the persist service's authentication middleware. It wraps the key async operations-environment lookup and plan retrieval-in tracer.trace calls, enabling better insight into where time is being spent during middleware execution.

### DIFF
--- a/packages/persist/lib/middleware/auth.middleware.ts
+++ b/packages/persist/lib/middleware/auth.middleware.ts
@@ -1,6 +1,7 @@
 import db from '@nangohq/database';
 import { environmentService, getPlan } from '@nangohq/shared';
 import { flagHasPlan, stringifyError, tagTraceUser } from '@nangohq/utils';
+import tracer from 'dd-trace';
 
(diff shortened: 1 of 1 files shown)


bodinsamuel (approved):
We can add that but I fear it's the only thing that's already tracked and we won't learn much 
cf: the 2 queries in the trace
Screenshot 2025-06-17 at 20 35 51

Since both are back to back and yet we see a big hole between them I think the problem is elsewhere.

TBonnin:
> We can add that but I fear it's the only thing that's already tracked and we won't learn much cf: the 2 queries in the trace Screenshot 2025-06-17 at 20 35 51
> 
> Since both are back to back and yet we see a big hole between them I think the problem is elsewhere.

the first function (account and environment) is also doing some hashing before the db query and some decrypting after. The new trace will confirm if any of those operation is slow. 
I am not expecting to learn much from the getPlan trace since the function is basically only the db query but I have added it just in case it reveal some gap before

PR #4207, opened 2025-06-17
chore(orch): more tracing and prevent multiple concurrent cleaning/expiring

TBonnin:
add locking around the cleaning/expiring logic to ensure only one instance is executing at a time.
This prevent:
- executing expensive db queries more than necessary
- double deletion or expiration of tasks
Also add tracing around cleaning and expiring logic Simplify tracing logic in scheduling
Also add a trace in scheduling so we can have a metric/alert that doesn't include instances that couldn't acquire the lock and are actually not scheduling tasks

 

---

This PR introduces PostgreSQL advisory locks to the cleaning, expiring, and scheduling daemons to ensure only one instance can execute critical logic at a time, preventing race conditions such as double deletion/expiration and unnecessary database load under high-availability deployments. Additionally, it simplifies and expands tracing across these daemons for improved observability, enabling more accurate metrics and alerting, and cleans up redundant trace logic to better encapsulate task scheduling behavior.

**Key Changes:**
• Added PostgreSQL advisory locks (pg_try_advisory_xact_lock) to cleaning, expiring, and scheduling daemons.
• Conditional execution of cleaning, expiring, or scheduling logic only when a lock is obtained; otherwise, instances wait before retrying.
• Refactored tracing: replaced manual span management with tracer.trace, applied consistent nested tracing for key operations.
• Removed redundant console logging and streamlined error handling in scheduling query logic.
• Updated the SchedulerDaemon base class to add tracing at the run-cycle level.

**Affected Areas:**
• scheduler/daemon.ts (Core daemon logic & tracing)
• scheduler/daemons/cleaning/cleaning.daemon.ts
• scheduler/daemons/expiring/expiring.daemon.ts
• scheduler/daemons/scheduling/scheduling.daemon.ts
• scheduler/daemons/scheduling/scheduling.ts (dueSchedules query logic)

### DIFF
--- a/packages/scheduler/lib/daemons/daemon.ts
+++ b/packages/scheduler/lib/daemons/daemon.ts
@@ -1,6 +1,7 @@
 import { logger } from '../utils/logger.js';
 import { setTimeout } from 'node:timers/promises';
 import type knex from 'knex';
+import tracer from 'dd-trace';
 
 export abstract class SchedulerDaemon {
     private name: string;
@@ -40,7 +41,14 @@ export abstract class SchedulerDaemon {
             }
             this.status = 'running';
             while (!this.abortSignal.aborted) {
-                await this.run();
+                await tracer.trace(`scheduler.${this.name.toLowerCase()}.run`, async (span) => {
+                    try {
+                        await this.run();
+                    } catch (err) {
+                        span?.setTag('error', err);
+                        throw err;
+                    }
+                });
                 await setTimeout(this.tickIntervalMs);
             }
         } catch (err) {
--- a/packages/scheduler/lib/daemons/daemon.ts
+++ b/packages/scheduler/lib/daemons/daemon.ts
@@ -41,13 +41,8 @@ export abstract class SchedulerDaemon {
             }
             this.status = 'running';
             while (!this.abortSignal.aborted) {
-                await tracer.trace(`scheduler.${this.name.toLowerCase()}.run`, async (span) => {
-                    try {
-                        await this.run();
-                    } catch (err) {
-                        span?.setTag('error', err);
-                        throw err;
-                    }
+                await tracer.trace(`scheduler.${this.name.toLowerCase()}.run`, async () => {
+                    await this.run();
                 });
                 await setTimeout(this.tickIntervalMs);
             }
--- a/packages/scheduler/lib/daemons/scheduling/scheduling.daemon.ts
+++ b/packages/scheduler/lib/daemons/scheduling/scheduling.daemon.ts
@@ -36,70 +36,62 @@ export class SchedulingDaemon extends SchedulerDaemon {
 
     async run(): Promise<void> {
         return this.db.transaction(async (trx) => {
-            const span = tracer.startSpan('scheduler.scheduling.schedule');
             try {
                 // Try to acquire a lock to prevent multiple instances from scheduling at the same time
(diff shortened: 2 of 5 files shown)


packages/scheduler/lib/daemons/daemon.ts
```diff
@@ -40,7 +41,14 @@ export abstract class SchedulerDaemon {
...
             while (!this.abortSignal.aborted) {
-                await this.run();
+                await tracer.trace(`scheduler.${this.name.toLowerCase()}.run`, async (span) => {
+                    try {
+                        await this.run();
+                    } catch (err) {
+                        span?.setTag('error', err);
+                        throw err;
+                    }
+                });
```
TBonnin, line 46: we are currently not tracing the expiring and cleaning daemon. I think that would be good to add them to the dashboard and maybe even have alerts

packages/scheduler/lib/daemons/scheduling/scheduling.daemon.ts
```diff
@@ -36,70 +36,62 @@ export class SchedulingDaemon extends SchedulerDaemon {
 
     async run(): Promise<void> {
         return this.db.transaction(async (trx) => {
-            const span = tracer.startSpan('scheduler.scheduling.schedule');
```
TBonnin, line 39: this trace is superseded by the one in the parent class daemon.ts

packages/scheduler/lib/daemons/scheduling/scheduling.daemon.ts
```diff
@@ -36,70 +36,62 @@ export class SchedulingDaemon extends SchedulerDaemon {
...
                 const lockGranted = res?.rows.length > 0 ? res.rows[0]!.lock_schedule : false;
-                lockSpan.finish();
 
                 if (lockGranted) {
-                    const dueSchedulesSpan = tracer.startSpan('scheduler.scheduling.due_schedules', { childOf: span });
-                    const getDueSchedules = await tracer.scope().activate(dueSchedulesSpan, async () => {
-                        return dueSchedules(trx);
-                    });
-                    dueSchedulesSpan.finish();
+                    await tracer.trace('scheduler.scheduling.schedule_tasks', async () => {
```
TBonnin, line 47: the indentation makes it hard to review but this diff is:
- simplifying the tracing logc by using `tracer.trace` instead of starting the spans manually
- add a new span that encompass the due schedules query and the task creations. Currently we have a dashboard widget on scheduler.scheduling.schedule above but when more than one instance (ex: during deployment) the metric account for the instance that couldn't acquire the lock and is waiting, skewing the metric which doesn't then represent the actual total scheduling duration

packages/scheduler/lib/daemons/daemon.ts
```diff
@@ -40,7 +41,14 @@ export abstract class SchedulerDaemon {
             }
             this.status = 'running';
             while (!this.abortSignal.aborted) {
-                await this.run();
+                await tracer.trace(`scheduler.${this.name.toLowerCase()}.run`, async (span) => {
+                    try {
+                        await this.run();
+                    } catch (err) {
+                        span?.setTag('error', err);
+                        throw err;
```
bodinsamuel: I never remember if it's necessary, I feel like it should be done by default by the wrapper

bodinsamuel (approved):

packages/scheduler/lib/daemons/daemon.ts
TBonnin: no, you are right. this is done by `trace`

PR #4210, opened 2025-06-18
feat(release): managed release process for self-hosted images

rossmcewan:
---

This PR introduces a comprehensive, automated release workflow for self-hosted Docker images using GitHub Actions, with semantic versioning tied to the application version and deployment stability. The process uses manifest-driven versioning and release history tracking, in-cluster deployment tests using kind/Helm, and enforces manual approval for breaking changes or failed deployments, including logic for major version incrementing.

**Key Changes:**
• Created .github/workflows/managed-release.yaml with a multi-stage CI pipeline for managed releases.
• Implemented scripts/manage-managed-release.mjs to handle manifest updates, semantic versioning logic, and GitHub comparison URL generation.
• Added managed-manifest.json for tracking latest and historical releases, including commit hash, release date, and diff URLs.
• Improved scripts/package.json/package-lock.json with reorganization and semver as a dependency.
• Integrated automated in-cluster testing of new images with kind and Helm before finalizing image tag and release.
• Introduced manual approval step and major version bumping if deployment or tests fail.
• Automated branch and PR creation for manifest/version changes.

**Affected Areas:**
• .github/workflows/managed-release.yaml
• scripts/manage-managed-release.mjs
• managed-manifest.json
• scripts/package.json
• package-lock.json


scripts/manage-managed-release.mjs
```diff
@@ -0,0 +1,86 @@
...
+
+function getChangesSinceLastRelease(lastCommitHash) {
+    try {
+        const changes = execSync(`git log ${lastCommitHash}..HEAD --pretty=format:"%s"`).toString().split('\n');
+        return changes.filter((change) => change.trim() !== '');
+    } catch (err) {
+        console.error('Error getting changes:', err);
+        return [];
+    }
+}
```
bodinsamuel, line 44: I would question the value of copying the commit titles like this without anymore context, we could just link to the github list of commits

.github/workflows/managed-release.yaml
```diff
@@ -0,0 +1,374 @@
...
+            - name: Checkout code
+              uses: actions/checkout@v4
+              with:
+                  ref: ${{ needs.manage-versions.outputs.commit_hash }}
+                  fetch-depth: 1
+                  token: ${{ secrets.GITHUB_TOKEN }}
+
+            - name: Bump major version if needed
+              id: bump-major
+              if: needs.test-deployment.outputs.status == 'failure'
```
bodinsamuel, line 303: it's not really clear why faillure === major?

.github/workflows/managed-release.yaml
```diff
@@ -0,0 +1,374 @@
...
+              if: ${{ env.ACT }}
+              run: |
+                  echo "Running in act - would perform Docker operations:"
+                  echo "Would pull: nangohq/nango:${{ needs.manage-versions.outputs.commit_hash }}"
+                  if [ "${{ needs.test-deployment.outputs.status }}" = "failure" ]; then
+                    echo "Would tag as: nangohq/nango:managed-${{ steps.bump-major.outputs.image_version }}-${{ steps.bump-major.outputs.app_version }}-${{ needs.manage-versions.outputs.commit_hash }}"
+                  else
+                    echo "Would tag as: nangohq/nango:managed-${{ needs.manage-versions.outputs.image_version }}-${{ needs.manage-versions.outputs.app_version }}-${{ needs.manage-versions.outputs.commit_hash }}"
+                  fi
+                  echo "Would tag as: nangohq/nango:managed-latest"
```
bodinsamuel: leftover?

.github/workflows/managed-release.yaml
```diff
@@ -0,0 +1,374 @@
...
+                    --base main
+
+            - name: Log PR creation (act)
+              if: ${{ env.ACT }}
+              run: |
+                  echo "Running in act - would create PR:"
+                  echo "Would create branch: version-update-$(date +%Y%m%d-%H%M%S)"
+                  echo "Would commit changes to managed-manifest.json"
+                  echo "Would create PR with title: Update version in manifest"
+                  echo "Would create PR with body: This PR updates the version in the manifest. Please review and merge to ensure the next release uses the correct version."
```
bodinsamuel, line 372: For both git and docker, would it be possible to not split act vs not act
It's the best way to have inconsistencies. For example here, the only thing you don't want to do is pushing the commit and create the PR. Same for docker, putting the push in a if should be enough

.github/workflows/managed-release.yaml
```diff
@@ -0,0 +1,374 @@
...
+            - name: Install dependencies
+              run: |
+                  echo "installing dependencies"
+                  npm install -w scripts --ignore-scripts
+
+            - name: Run version script
+              id: version-script
+              run: |
+                  echo "running version script"
+                  node scripts/manage-managed-release.mjs ${{ inputs.commit_hash }}
```
bodinsamuel, line 49: I don't get why we call this here and then call major bump later? maybe it's related to my other question

rossmcewan: In a none-breaking deployment there is no need to do a major bump as there aren't any breaking changes. The default is to increment the appropriate version segment based on the changes in the deployable application. This would only bump minors and patches which represent none-breaking changes. 

After we deploy to the kind cluster we check if it deploys and runs ok. At the moment this is considered "good enough" to assume that no breaking changes were introduced. 

If the cluster fails to deploy and start then a manual approval is required. The idea behind this approval is that a human in the loop can determine if it is a breaking change and requires additional deployment steps that should be captured in the release notes. If the changes are approved then the major version is bumped to indicate this status. If not, the deployment process exits and can be run again (e.g. if it was related to something else that isn't linked to this release).

.github/workflows/managed-release.yaml
rossmcewan: The idea is that the workflow can only get here under two conditions:
- The deployment was successful
- The deployment was not successful (i.e. status is failed) but was approved for release. If this is the case, then a major version bump is required.

.github/workflows/managed-release.yaml
rossmcewan: I'm unsure whether to remove or not as they are extremely valuable for local testing. The condition that is evaluated (`if: ${{ env.ACT }}`) is only true when using [act](https://nektosact.com/) to run locally.

.github/workflows/managed-release.yaml
rossmcewan: Yeah but its a lot of logging that isn't really necessary when not in a local environment. When act is true, I really didn't want any side-effects so thought the best way to guarantee this was to not mix the step. Also very easy to remove if we decide we don't want the logging.

scripts/manage-managed-release.mjs
rossmcewan: Fair point. Can just put a link there.

.github/workflows/managed-release.yaml
bodinsamuel: I specifically meant this line "echo "Would tag as: nangohq/nango:managed-latest"" you are never tagging latest

.github/workflows/managed-release.yaml
bodinsamuel: It's not that I don't want, on the contrary it's just going to be outdated pretty soon but we'll see

.github/workflows/managed-release.yaml
bodinsamuel: could benefits from this comment in the code

.github/workflows/managed-release.yaml
bodinsamuel: I see, could benefits from the comments in the code too 👍🏻

bodinsamuel (approved):

.github/workflows/managed-release.yaml
rossmcewan: Ah. Yes, then you're right 🙈 I originally was tagging latest but removed it.

.github/workflows/managed-release.yaml
```diff
@@ -0,0 +1,378 @@
...
+
+            - name: Retag and push final image
+              if: ${{ !env.ACT }}
+              run: |
+                  echo "Pulling image..."
+                  docker pull nangohq/nango:${{ needs.manage-versions.outputs.commit_hash }}
+
+                  if [ "${{ needs.test-deployment.outputs.status }}" = "failure" ]; then
+                      echo "Test deployment failed, using bumped major version"
+                      docker tag nangohq/nango:${{ needs.manage-versions.outputs.commit_hash }} nangohq/nango:managed-${{ steps.bump-major.outputs.image_version }}-${{ steps.bump-major.outputs.app_version }}-${{ needs.manage-versions.outputs.commit_hash }}
```
TBonnin, line 339: first tag (nangohq/nango:COMMIT) is already there. no?

.github/workflows/managed-release.yaml
```diff
@@ -0,0 +1,378 @@
...
+              with:
+                  version: 'v0.20.0'
+                  install_only: true
+
+            - name: Install Helm
+              uses: azure/setup-helm@v3
+              with:
+                  version: 'v3.12.3'
+
+            - name: Deploy to sandbox
```
TBonnin, line 70: I would add a timeout on this one. just to make sure it doesn't run for ages

.github/workflows/managed-release.yaml
```diff
@@ -0,0 +1,378 @@
...
+                  apiVersion: v1
+                  kind: Secret
+                  metadata:
+                    name: nango-secrets
+                    namespace: nango
+                  type: Opaque
+                  data:
+                    postgres-password: bmFuZ28xMjM=
+                    encryption-key: bUJVcm5WQ2xUdmJkTXlwMjNwWTRkNXJiaE56VENZN1JWWjY5OE9zVm9raz0=
+                    mailgun-api-key: eW91ci1tYWlsZ3VuLWtleS1oZXJl
```
TBonnin, line 90: should those be secrets?

.github/workflows/managed-release.yaml
```diff
@@ -0,0 +1,378 @@
...
+                    --set shared.namespace=nango \
+                    --set shared.tag=${{ needs.manage-versions.outputs.commit_hash }} \ 
+                    --set shared.encryptionEnabled=true \
+                    --set shared.NANGO_SERVER_URL="http://localhost:3003" \
+                    --set shared.NANGO_CALLBACK_URL="http://localhost:3003/oauth/callback" \
+                    --set shared.NANGO_LOGS_ENABLED=false \
+                    --set server.useLoadBalancer=false \
+                    --set elasticsearch.enabled=false \
+                    --set postgresql.auth.postgresPassword=nango123 \
+                    --wait=false &
```
TBonnin, line 118: why wait=false and then sleeping 10s?

PR #4219, opened 2025-06-19
fix(auth): handle csp in auth callback html

bodinsamuel:
## Changes

Customer reported ```callback:16 Refused to execute inline event handler because it violates the following Content Security Policy directive: "script-src-attr 'none'". Either the 'unsafe-inline' keyword, a hash ('sha256-...'), or a nonce ('nonce-...') is required to enable inline execution. Note that hashes do not apply to event handlers, style attributes and javascript: navigations unless the 'unsafe-hashes' keyword is present.```

- Use click event instead of onClick to handle csp directive

---

This PR updates the authentication callback HTML to address a customer-reported Content Security Policy (CSP) violation. It removes the inline onClick handler from the close button and replaces it with a programmatically attached click event via JavaScript, ensuring compliance with strict CSP settings.


packages/server/lib/utils/html.ts
```diff
@@ -34,6 +34,14 @@ Nango OAuth flow callback. Read more about how to use it at: https://github.com/
         window.open('', '_self').close();
       }
 
+      // Add event listener to close button
+      document.addEventListener('DOMContentLoaded', function() {
+        const closeButton = document.getElementById('closeButton');
+        if (closeButton) {
+          closeButton.addEventListener('click', closeWindow);
```
TBonnin, line 41: can closeWindow fail? if yes does it matter?

TBonnin (approved):

packages/server/lib/utils/html.ts
bodinsamuel: it's already failing, but it doesn't matter it can only fail if there are some csp issues so I can't work anyway

PR #4224, opened 2025-06-20
fix: commit hash added to success path tag

rossmcewan:
Discrepancy between what was executed locally and the actual image tag that was pushed. In a standard successful scenario, the image tag did not include the commit hash. Changed to always append the commit hash to the docker tag.

---

This PR updates the managed-release GitHub Actions workflow to ensure that Docker image tags in both success and failure release paths always include the commit hash. This change addresses discrepancies where images built locally differed from those pushed due to missing commit hashes in the tag, improving traceability and tag consistency.


bodinsamuel (approved):
(told you so :D)

PR #4228, opened 2025-06-20
fix(plans): new types, store customer_id at creation

bodinsamuel:
## Changes

- new types after migration
- Store orb customer_id at creation so we can reuse it and match internal team from orb webhooks

---

This PR extends the DBPlan type to support additional billing fields for Stripe and Orb, following a recent migration, and updates the plan creation logic to store orb_customer_id at the time of account creation. These changes enable future integration with Orb webhooks for mapping internal teams and standardize the plan management API, seeder, and type definitions accordingly. Updates are made to both the user sign-up and managed account callback flows to ensure orb_customer_id is set when applicable.

**Key Changes:**
• Expanded DBPlan interface with new Stripe and Orb billing fields: stripe_customer_id, stripe_payment_id, orb_customer_id, orb_subscription_id, orb_future_plan, orb_future_plan_at
• Modified plan seeder and formatter to handle new fields
• Added updatePlanByTeam() service to update plans using account_id
• Controllers (signup and managed account) now update orb_customer_id at plan creation when flagHasUsage is set
• Imported and used flagHasUsage utility to control conditional Orb logic

**Affected Areas:**
• DBPlan type definitions (packages/types/lib/plans/db.ts)
• Plan seeder and test data (packages/shared/lib/seeders/plan.seeder.ts)
• Plan formatter for API output (packages/server/lib/formatters/plan.ts)
• Plan services (packages/shared/lib/services/plans/plans.ts)
• Controllers: signup, managed account callback


packages/shared/lib/services/plans/plans.ts
```diff
@@ -51,6 +51,21 @@ export async function updatePlan(db: Knex, { id, ...data }: Pick<DBPlan, 'id'> &
     }
 }
 
+export async function updatePlanByTeam(
+    db: Knex,
+    { account_id, ...data }: Pick<DBPlan, 'account_id'> & Partial<Omit<DBPlan, 'id' | 'account_id'>>
```
TBonnin, line 56: is `account_id` required in the `Omit` since it is in `Pick`?

packages/shared/lib/services/plans/plans.ts
```diff
@@ -51,6 +51,21 @@ export async function updatePlan(db: Knex, { id, ...data }: Pick<DBPlan, 'id'> &
...
+): Promise<Result<boolean>> {
+    try {
+        await db
+            .from<DBPlan>('plans')
+            .where('account_id', account_id)
+            .update({ ...data, updated_at: new Date() });
```
TBonnin: can we use `db.fn.now()` so pg CURRENT_TIMESTAMP is being used?

TBonnin (approved):

packages/shared/lib/services/plans/plans.ts
bodinsamuel: yes, it translates to `{ account_id: string } & EverythingElseIsOptional<>`
I need the pick to avoid having account_id being set has optional too

PR #4230, opened 2025-06-20
fix(records): paginate markPreviousGenerationRecordsAsDeleted

TBonnin:
query to mark records as deleted when track_deletes is enabled can timeout if the number of records is big.
With this commit we now paginate the update query.
Also adding tests for `markPreviousGenerationRecordsAsDeleted`

---

This PR refactors the `markPreviousGenerationRecordsAsDeleted` function to process large record deletions in batches, mitigating database timeouts when marking previous generations as deleted while `track_deletes` is enabled. It introduces a configurable `batchSize` to the method and adds comprehensive integration tests to verify correct behavior, including batching, scope, soft-delete handling, and record counts.

**Key Changes:**
• Refactored `markPreviousGenerationRecordsAsDeleted` to paginate updates in batches (default 5000)
• Added `batchSize` parameter to the method
• Accumulates and returns deleted `external_id`s across batches
• Extensive integration tests for batching, scoping, soft-deletes, and record count updates

**Affected Areas:**
• packages/records/lib/models/records.ts (core batch logic, method signature change)
• packages/records/lib/models/records.integration.test.ts (new/expanded test suite)

### DIFF
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -673,34 +673,50 @@ export async function markPreviousGenerationRecordsAsDeleted({
     connectionId,
     model,
     syncId,
-    generation
+    generation,
+    batchSize = 5000
 }: {
     connectionId: number;
     model: string;
     syncId: string;
     generation: number;
+    batchSize?: number;
 }): Promise<string[]> {
     const now = db.fn.now(6);
-    let res: string[] = [];
     return db.transaction(async (trx) => {
-        res = (await trx
-            .from<FormattedRecord>(RECORDS_TABLE)
-            .where({
-                connection_id: connectionId,
-                model,
-                sync_id: syncId,
-                deleted_at: null
-            })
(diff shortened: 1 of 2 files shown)


packages/records/lib/models/records.ts
```diff
@@ -673,34 +673,50 @@ export async function markPreviousGenerationRecordsAsDeleted({
...
+                            sync_id: syncId,
+                            deleted_at: null
+                        })
+                        .where('sync_job_id', '<', generation)
+                        .limit(batchSize);
+                })
+                .update({
+                    deleted_at: now,
+                    updated_at: now,
+                    sync_job_id: generation
```
kaposke, line 707: Why do we also update the generation?

kaposke (approved):

packages/records/lib/models/records.ts
TBonnin: because the records are now updated/marked as deleted by this generation

PR #4234, opened 2025-06-23
fix(plans): add auto_idle flag

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-3456/make-auto-idling-a-feature-flag

- add auto_idle flag
Not used yet but we'll replace most of the trial flag

---

This PR introduces an 'auto_idle' boolean flag to the 'plans' table, adds it to the DBPlan type definition, and updates the plan seeder to include this new field. The flag is added with a default value of true and is not yet used in application logic, but will eventually replace most uses of the 'trial' flag for plan-based machine idling.


kaposke (approved):
What will the flag do? In which case does the machine idle?

TBonnin (approved):

bodinsamuel:
> What will the flag do? In which case does the machine idle?

After N days, the machine will be idled by default no matter the usage

PR #4236, opened 2025-06-23
fix(db): remove some schema leftover

bodinsamuel:
## Changes

- remove some schema leftover
Schema is configurable so it should never be hardcoded

---

This PR removes hardcoded 'nango.' schema prefixes from SQL queries in packages/shared/lib/services/deploy/utils.ts to align with the application's support for configurable schemas. Table references are now schema-agnostic, preventing issues in environments where the default schema is not 'nango'.


kaposke (approved):

TBonnin (approved):
I think this was a mistake not to have a fixed db schema but at least it is consistent

bodinsamuel:
> I think this was a mistake not to have a fixed db schema but at least it is consistent

I agree but can't break it

PR #4239, opened 2025-06-23
fix(runner): implement MaxLifetimeAgent for persist client socket management

TBonnin:
Requests from runner to `persist` (records, logs, ...) are not evenly balanced between `persist` instances since a single execution will use a single connection (`keepalive=true`) that is always routed to the same persist instance. (Render confirmed they are using connection-based routing)
This is an attempt at more evenly-balanced requests from runners to persist by adding a custom HTTPS agent that enforces maximum socket lifetime to prevent a single script requests to always be routed to the same persist instance, potentially overwhelming it.
Configurable via RUNNER_PERSIST_MAX_SOCKET_MAX_LIFETIME_MS environment variable (defaults to 30s).

---

This PR introduces a custom HTTPS agent (MaxLifetimeAgent) for the runner's persist client, which enforces a configurable maximum socket lifetime to address uneven load balancing across persist service instances when connection-based routing is used (e.g., Render). The maximum socket lifetime is controlled via a new environment variable, RUNNER_PERSIST_MAX_SOCKET_MAX_LIFETIME_MS (default 30s), with updates to the environment schema and integration test config for validation and coverage.

**Key Changes:**
• Introduced MaxLifetimeAgent, a custom subclass of node:https.Agent that limits socket lifetime.
• PersistClient now uses MaxLifetimeAgent for outgoing HTTPS requests to persist.
• Added RUNNER_PERSIST_MAX_SOCKET_MAX_LIFETIME_MS to environment variable schema with a default of 30,000 ms.
• Integration test configuration updated to include RUNNER_NODE_ID.

**Affected Areas:**
• packages/runner/lib/sdk/persist.ts (PersistClient, HTTP agent behavior)
• packages/utils/lib/environment/parse.ts (environment validation/config)
• vite.integration.config.ts (test runner env config)

### DIFF
--- a/packages/runner/lib/sdk/persist.ts
+++ b/packages/runner/lib/sdk/persist.ts
@@ -62,7 +62,7 @@ export class MaxLifetimeAgent extends Agent {
         }
 
         // https://nodejs.org/docs/latest/api/http.html#agentkeepsocketalivesocket
-        // @ts-expect-error - @types/node does not define kkeepSocketAlive
+        // @ts-expect-error - @types/node does not define keepSocketAlive
         return super.keepSocketAlive(socket);
     }
 
--- a/packages/runner/lib/sdk/persist.ts
+++ b/packages/runner/lib/sdk/persist.ts
@@ -1,5 +1,6 @@
-import https from 'node:https';
-
+import type { RequestOptions } from 'node:https';
+import { Agent } from 'node:https';
+import type { Socket } from 'net';
 import axios, { AxiosError, isAxiosError } from 'axios';
 
 import { getUserAgent } from '@nangohq/node';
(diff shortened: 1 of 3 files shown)


packages/utils/lib/environment/parse.ts
```diff
@@ -81,6 +81,7 @@ export const ENVS = z.object({
     RUNNER_NODE_ID: z.coerce.number().optional(),
     RUNNER_URL: z.string().url().optional(),
     RUNNER_MEMORY_WARNING_THRESHOLD: z.coerce.number().optional().default(85),
+    RUNNER_PERSIST_MAX_SOCKET_MAX_LIFETIME_MS: z.coerce.number().optional().default(30_000),
```
TBonnin, line 84: I pick 30s quite arbitrarily. I am open to suggestion

kaposke: Sounds good to me

kaposke (approved):

packages/runner/lib/sdk/persist.ts
```diff
@@ -19,6 +20,57 @@ import type {
...
+
+        if (birthTime) {
+            const age = Date.now() - birthTime;
+            if (age >= this.maxSocketLifetime) {
+                return false;
+            }
+        }
+
+        // https://nodejs.org/docs/latest/api/http.html#agentkeepsocketalivesocket
+        // @ts-expect-error - @types/node does not define kkeepSocketAlive
```
bodinsamuel:
```suggestion
        // @ts-expect-error - @types/node does not define keepSocketAlive
```

bodinsamuel (approved):

PR #4243, opened 2025-06-24
fix: do not send external facing connection id inside billing events

TBonnin:
Customers used to be able to set the connectionId and therefore it can contain customer data.
This commit ensures we attach the internal connection id (the auto-incremented int) to billing events

---

This PR replaces the external, user-controllable `connectionId` with the internal auto-incremented numeric connection ID (`nangoConnectionId` / `connection.id`) across persistence flows and billing event emission. This unifies usage of identifiers to prevent exposure of customer data and ensures that internal IDs are consistently propagated within billing events and record APIs, with appropriate updates to types, APIs, endpoints, and log metadata.

**Key Changes:**
• Replaced all usage of external-facing `connectionId` with internal numeric connection ID (`nangoConnectionId`/`connection.id`) in record persistence and billing logic.
• Refactored and clarified parameter usage in the persist records API and handler files to exclusively use internal connection ID.
• Removed trust in user-supplied `connectionId` from API inputs, extracting internal ID from router parameters.
• Updated sync controller to emit billing events with the internal connection ID in the payload.
• Unified and updated types and logger metadata to reflect the new internal ID usage.

**Affected Areas:**
• Record persistence logic (packages/persist/lib/records.ts)
• Record API endpoints (POST/PUT/DELETE for records)
• Sync controller billing event code (packages/server/lib/controllers/sync.controller.ts)
• API type definitions and parameter parsing in routes

### DIFF
--- a/packages/persist/lib/records.ts
+++ b/packages/persist/lib/records.ts
@@ -21,7 +21,6 @@ export async function persistRecords({
     connectionId,
     plan,
     providerConfigKey,
-    nangoConnectionId,
     syncId,
     syncJobId,
     model,
@@ -32,10 +31,9 @@ export async function persistRecords({
     persistType: PersistType;
     accountId: number;
     environmentId: number;
-    connectionId: string;
+    connectionId: number;
     plan: DBPlan | null;
     providerConfigKey: string;
-    nangoConnectionId: number;
     syncId: string;
     syncJobId: number;
     model: string;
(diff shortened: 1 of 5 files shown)


packages/persist/lib/records.ts
```diff
@@ -32,10 +31,9 @@ export async function persistRecords({
     persistType: PersistType;
     accountId: number;
     environmentId: number;
-    connectionId: string;
+    connectionId: number;
```
TBonnin, line 34: the external connectionId was only used for internal error reporting. Removing it to prevent confusion

bodinsamuel (approved):

PR #4252, opened 2025-06-25
fix(figma): Fix Figma SCIM integration credential verification failure

yksolanki9:
Problem:
The Figma SCIM integration was failing during credential verification with an "invalid credentials" error. The root cause was an inconsistency in how connection configuration properties were being referenced.

Cause:
- Figma verification endpoint  uses `connectionConfig.tenantId` to reference tenantId
- In the `buildProxyURL` method, while `connectionConfig` was being properly replaced with `connection_config` for the base_url, this replacement was not being applied to the endpoint string
- This mismatch caused the verification process to fail

Solution:
- Updated the `buildProxyURL` method to use `mapProxyBaseUrlInterpolationFormat` which replaces `connectionConfig` with `connection_config` in the string
- This fix ensures that Figma SCIM integration credentials can be successfully verified and added.

---

This PR addresses a credential verification failure in the Figma SCIM integration caused by inconsistent variable referencing between the base URL and endpoint when building proxy URLs. The fix introduces a new helper, interpolateProxyUrlParts, to normalize and interpolate both URL parts consistently, ensuring references like `connectionConfig.tenantId` are mapped to `connection_config.tenantId` throughout. The change ensures that credentials can be verified and integrations set up successfully.

**Key Changes:**
• Introduced interpolateProxyUrlParts helper to unify variable substitution for both base URLs and endpoints in proxy URLs.
• Refactored buildProxyURL to apply consistent interpolation logic, concatenating the formatted base and endpoint before performing value replacement.
• Replaced the older mapProxyBaseUrlInterpolationFormat with the new utility and updated relevant imports.
• Removed the now-obsolete mapProxyBaseUrlInterpolationFormat function.

**Affected Areas:**
• Proxy URL generation logic (packages/shared/lib/services/proxy/utils.ts)
• Utility functions for proxy URL formatting and interpolation (packages/shared/lib/utils/utils.ts)


yksolanki9:
Hi @hassan254-prog, can you please have a look at this pr. Thanks!

packages/shared/lib/services/proxy/utils.ts
```diff
@@ -173,7 +173,7 @@ export function buildProxyURL({ config, connection }: { config: ApplicationConst
     const endpoint = apiEndpoint.startsWith('/') ? apiEndpoint.slice(1) : apiEndpoint;
 
     const fullEndpoint = interpolateIfNeeded(
-        `${mapProxyBaseUrlInterpolationFormat(base)}${endpoint ? '/' : ''}${endpoint}`,
+        `${mapProxyBaseUrlInterpolationFormat(base)}${endpoint ? '/' : ''}${endpoint ? mapProxyBaseUrlInterpolationFormat(endpoint) : ''}`,
```
hassan254-prog: This might be a good opportunity to improve this for readability, could this work

```
const normalizedBase = apiBase?.endsWith('/') ? apiBase.slice(0, -1) : apiBase;
const normalizedEndpoint = apiEndpoint.startsWith('/') ? apiEndpoint.slice(1) : apiEndpoint;

const baseFormatted = interpolateProxyUrlParts(normalizedBase);
const endpointFormatted = normalizedEndpoint ? interpolateProxyUrlParts(normalizedEndpoint) : '';
const combinedUrl = [baseFormatted, endpointFormatted].filter(Boolean).join('/');
const fullEndpoint = interpolateIfNeeded(
        combinedUrl,
        connectionCopyWithParsedConnectionConfig(connection) as unknown as Record<string, string>
 );
```

Then improve the helper function to correctly handle the reference, since we're now interpolating both parts;
```
export function interpolateProxyUrlParts(proxyUrlPart: string | undefined): string | undefined {
    // Maps the format that is used in providers.yaml (inherited from oauth), to the format of the Connection model.
    return proxyUrlPart ? proxyUrlPart.replace(/connectionConfig/g, 'connection_config') : proxyUrlPart;
}
```

hassan254-prog (commented):
Thanks for your contribution, it looks great 🔥. I left a few minor comments to improve the function's readability.

hassan254-prog (approved):

PR #4265, opened 2025-06-26
fix: improve LegacyModel to Json Schema conversion

kaposke:
### Description
I looked at dozens of real world legacy schemas to find anything I've missed with my first implementation.
Thankfully so, because I've found a few things I didn't know about.

This adds support for:
- missing primitives.
- inline objects (defined as js objects...)
- object hierarchy defined by dots in the name (Eg. `person.name`, `person.age`)

> Not too much effort on making code incredible, since all of this is intended for a single migration and should be deleted as soon as we fully migrate flows.yaml

---

This PR significantly expands and corrects the legacySyncModelToJsonSchema utility to better handle a wide range of real-world legacy schema cases. The updates add support for previously missing primitive types (such as char, varchar, int, float, bool, true/false, undefined, any, object, and array), robust handling of inline objects (where types can be JS objects or arrays of such), and nested property handling using dot notation to build object hierarchies. The default behavior for unknown or non-existent references is changed to emit a const string type. Test coverage has been greatly expanded and adjusted to reflect these new and fixed behaviors.

**Key Changes:**
• Extended primitive type mapping to handle many previously unsupported types (char, varchar, int, float, bool, true/false, undefined, object, any, array)
• Added support for fields with inline object or array type definitions (JS object/array as `type` field)
• Introduced parsing of nested fields using dot notation to correctly construct object hierarchies
• Changed handling of references to missing models: now emits a string const rather than a dangling $ref
• Improved and expanded unit tests and snapshots to cover the new scenarios and edge cases
• Introduced util functions (objectToFields, mergeObjects) for recursive and nested structure handling

**Affected Areas:**
• packages/utils/lib/jsonSchema/legacySyncModelToJsonSchema.ts
• packages/utils/lib/jsonSchema/legacySyncModelToJsonSchema.unit.test.ts
• packages/utils/lib/jsonSchema/__snapshots__/legacySyncModelToJsonSchema.unit.test.ts.snap

### DIFF
--- a/packages/utils/lib/jsonSchema/legacySyncModelToJsonSchema.unit.test.ts
+++ b/packages/utils/lib/jsonSchema/legacySyncModelToJsonSchema.unit.test.ts
@@ -11,13 +11,27 @@ describe('legacySyncModelsToJsonSchema', () => {
                 name: 'AllPrimitiveTypes',
                 fields: [
                     { name: 'stringField', type: 'string' },
+                    { name: 'charField', type: 'char' },
+                    { name: 'varcharField', type: 'varchar' },
                     { name: 'numberField', type: 'number' },
+                    { name: 'floatField', type: 'float' },
+                    { name: 'intField', type: 'int' },
+                    { name: 'integerField', type: 'integer' },
                     { name: 'booleanField', type: 'boolean' },
+                    { name: 'boolField', type: 'bool' },
+                    { name: 'trueField', type: 'true' },
+                    { name: 'falseField', type: 'false' },
                     { name: 'dateField', type: 'date' },
+                    { name: 'undefinedField', type: 'undefined' },
+                    { name: 'nullField', type: 'null' },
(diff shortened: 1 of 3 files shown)


TBonnin (approved):

packages/utils/lib/jsonSchema/legacySyncModelToJsonSchema.unit.test.ts
```diff
@@ -123,4 +153,265 @@ describe('legacySyncModelsToJsonSchema', () => {
...
+                                mimeType: 'string'
+                            }
+                        },
+                        {
+                            name: 'eventType',
+                            type: 'string'
+                        }
+                    ]
+                } as LegacySyncModelSchema
+            ];
```
bodinsamuel, line 411: I feel like you can remove 90% of this test

bodinsamuel (approved):

PR #4266, opened 2025-06-27
chore: add last_execution_at column to connection table

TBonnin:
we want to add a new metric to automate the billing of some of the legacy scale plan. The metric is "number of connections with at least a script execution during the period". There is currently no source of truth in the db regarding script executions (actions for example are not tracked). In order to avoid sending an event for every single execution, this commit is adding a last_execution_at column to the connection table that will be updated every time a script execution starts

 

---

This PR introduces a new 'last_execution_at' timestamptz column to the '_nango_connections' table to record the latest script execution per connection. An index is also created on this column to support efficient querying and future billing automation for legacy scale plans.

### DIFF
--- a/packages/database/lib/migrations/20250626163600_connection_last_execution_at.cjs
+++ b/packages/database/lib/migrations/20250626163600_connection_last_execution_at.cjs
@@ -1,0 +1,16 @@
+exports.config = { transaction: false };
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`ALTER TABLE "_nango_connections" ADD COLUMN IF NOT EXISTS "last_execution_at" timestamptz;`);
+    await knex.raw(
+        `CREATE INDEX CONCURRENTLY IF NOT EXISTS "idx_nango_connections_last_execution_at" ON "_nango_connections" ("last_execution_at") WHERE "last_execution_at" IS NOT NULL;`
+    );
+};
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.down = function () {};


packages/database/lib/migrations/20250626163600_connection_last_execution_at.cjs
```diff
@@ -0,0 +1,16 @@
+exports.config = { transaction: false };
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`ALTER TABLE "_nango_connections" ADD COLUMN IF NOT EXISTS "last_execution_at" timestamptz;`);
```
TBonnin, line 7: went for adding a column to the connections table. We could also have a dedicated table for it. wdyt?

bodinsamuel: we had this discussion multiple times, last time I didn't do it because I was going to end up joining the table all the time anyway. Maybe this one can be in a different table since it's never going to get queried (or not, could be valuable actually but just for the API)

packages/database/lib/migrations/20250626163600_connection_last_execution_at.cjs
```diff
@@ -0,0 +1,16 @@
...
+    await knex.raw(
+        `CREATE INDEX CONCURRENTLY IF NOT EXISTS "idx_nango_connections_last_execution_at" ON "_nango_connections" ("last_execution_at") WHERE "last_execution_at" IS NOT NULL;`
```
bodinsamuel, line 9: you probably need deleted = false too

bodinsamuel (approved):

packages/database/lib/migrations/20250626163600_connection_last_execution_at.cjs
TBonnin: that's why I hesitated. It is sort of self contained for now so having an extra table wouldn't be a bad idea. 
However this billing metric is simple at the moment and isn't affected by the connection creation date or the fact it is deleted or not for example. I am worried that if we introduced a similar billing metrics we might have to join with the connections table anyway. I am gonna go with the added column for now

packages/database/lib/migrations/20250626163600_connection_last_execution_at.cjs
TBonnin: `deleted` is actually not a criteria. Connections that were deleted during the current month should be accounted for.

PR #4267, opened 2025-06-27
chore: add 'count of active connections' billing metric

TBonnin:
- track connection last_execution_at every time a script starts
- count all connections with last_execution_at in current month
- send metric in cron

---

This PR introduces a new 'active connections' metric to support legacy billing requirements. Each time a script runs, the associated connection's last_execution_at timestamp is updated; a cron job now emits a metric counting all connections per account that have executed any script in the current calendar month (restricted to prod environments). The changes involve backend service logic, database schema migration (adding and indexing last_execution_at), cron job updates, and the types defining available billing metrics.

**Key Changes:**
• Schema change: adds last_execution_at column (with index) to _nango_connections
• Service: adds trackExecution (set timestamp) and billableActiveConnections (query by last_execution_at in month)
• Script execution: updates to call trackExecution when a script starts
• Cron: new billing.exportActiveConnections sends 'billable_active_connections' metric events
• Expanded billing metric types and ingest event definitions

**Affected Areas:**
• Connection service implementation
• Job execution flow (script start ops)
• Billing/usage cron logic
• Billing types and metrics

### DIFF
--- a/packages/server/lib/crons/usage.ts
+++ b/packages/server/lib/crons/usage.ts
@@ -32,6 +32,7 @@ export async function exec(): Promise<void> {
     await tracer.trace<Promise<void>>('nango.cron.exportUsage', async () => {
         logger.info(`Starting`);
         await billing.exportBillableConnections();
+        await billing.exportActiveConnections();
         await observability.exportConnectionsMetrics();
         await observability.exportRecordsMetrics();
         logger.info(`✅ done`);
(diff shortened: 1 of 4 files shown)


packages/server/lib/crons/usage.ts
```diff
@@ -32,6 +32,7 @@ export async function exec(): Promise<void> {
     await tracer.trace<Promise<void>>('nango.cron.exportUsage', async () => {
         logger.info(`Starting`);
         await billing.exportBillableConnections();
+        await billing.exportLegacyActiveConnections();
```
kaposke: Can't it just be `exportActiveConnections`? Legacy gives the impression of "needs to be refactored".

I've read the discussion in the issue, this is essentially a 4th metric that we can or not charge for in Orb. I think it's fine as `active connections` instead of `legacy active connections`, since the other metric is just a total number of connections, right?.

kaposke (approved):

PR #4268, opened 2025-06-27
fix: migration to fix json schema in public flows

kaposke:
### Describe
A migration to update the `models_json_schema` for all our public flows.

It uses either the `legacySyncModelsToJsonSchema` or `nangoModelsToJsonSchema` I have been developing to convert the `model_schema`, which is expected to be the trustworthy source of truth.

We do it for EVERY public flow, because their json schema was incorrectly saved until recently.

I add a tag to the resulting schema to show that it was generated by this migration, in case we need to apply it again with some tweak, we have something to filter for.

---

This PR implements a database migration that recalculates and updates the `models_json_schema` for all public, non-deleted sync configs in `_nango_sync_configs`. It ensures the JSON schema is derived from the trusted `model_schema` using either the legacy or new schema conversion logic, adds version tagging for traceability, improves schema conversion robustness for legacy edge cases (such as empty arrays), and enhances unit test coverage to reflect real-world data anomalies.

<details>

• Introduces migration (20250618140753_fix_json_schemas.cjs) to update `models_json_schema` for all public, non-deleted sync configs.
• Migrated schemas receive a `` marker for identification.
• Handles empty/invalid `model_schema` by setting `models_json_schema` to null and logging errors.
• Extends `legacySyncModelsToJsonSchema` logic to support empty array types, with matching unit tests/snapshots.
• Adds JSDoc type annotations and ensures migration script consistency.

</details>

<details>

• _nango_sync_configs (models_json_schema field)
• Database migration scripts
• Schema conversion utilities (legacy & new)
• Test suite and snapshots

</details>


packages/database/lib/migrations/20250618140753_fix_json_schemas.cjs
```diff
@@ -0,0 +1,47 @@
...
+            await knex.update({ models_json_schema: null }).from('_nango_sync_configs').where({ id: syncConfig.id });
+            continue;
+        }
+
+        const nangoModels = Array.isArray(syncConfig.model_schema) ? syncConfig.model_schema : [syncConfig.model_schema];
+        const newJsonSchema = isLegacyModelSchema(nangoModels) ? legacySyncModelsToJsonSchema(nangoModels) : nangoModelsToJsonSchema(nangoModels);
+
+        // Add a marker to the JSON schema to indicate that it was migrated.
+        // In case we ever need to fix something about this migration.
+        newJsonSchema['$nango_migration'] = '20250618140753';
```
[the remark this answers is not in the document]
kaposke: There's no path in which they return null. But adding a try-catch makes sense in case we get an invalid object from `model_schema`, which is possible.

TBonnin: I'd call the tag `version`

TBonnin (approved):
how long it is expected to take in production. There is no timeout for migration query I think so it should be a problem if it takes a bit of time though

kaposke:
> how long it is expected to take in production. There is no timeout for migration query I think so it should be a problem if it takes a bit of time though

There are a bit over 7100 flows to update. But it's all done in separate update queries, so ~7100 update queries. Should I wrap it in a transaction?

@TBonnin

packages/database/lib/migrations/20250618140753_fix_json_schemas.cjs
```diff
@@ -0,0 +1,47 @@
+const { legacySyncModelsToJsonSchema, nangoModelsToJsonSchema } = require('@nangohq/utils');
+
+exports.up = async function (knex) {
```
bodinsamuel, line 6: There is an implicit transaction that you'll need to disable, I think, otherwise it will never finish.

bodinsamuel, line 6: can you add the types like all recent migrations please

packages/database/lib/migrations/20250618140753_fix_json_schemas.cjs
```diff
@@ -0,0 +1,47 @@
...
+    for (const syncConfig of publicSyncConfigs) {
+        /**
+         * Set to null if there are no models.
+         */
+        if (
+            syncConfig.model_schema === null ||
+            Object.keys(syncConfig.model_schema).length === 0 ||
+            (Array.isArray(syncConfig.model_schema) && syncConfig.model_schema.length === 0)
+        ) {
+            await knex.update({ models_json_schema: null }).from('_nango_sync_configs').where({ id: syncConfig.id });
```
bodinsamuel, line 21: It's already null by default, Is there anything that needs fixing here?

bodinsamuel (approved):

TBonnin:
> > how long it is expected to take in production. There is no timeout for migration query I think so it should be a problem if it takes a bit of time though
> 
> There are a bit over 7100 flows to update. But it's all done in separate update queries, so ~7100 update queries. Should I wrap it in a transaction?
> 
> @TBonnin

Is the migration idempotent? if not, a transaction would be needed

packages/database/lib/migrations/20250618140753_fix_json_schemas.cjs
kaposke: Just making sure it's consistent as we're assuming these had bad logic.

PR #4270, opened 2025-06-27
feat: show in-app usage for free plans

kaposke:
## Description
- Created a free plan in orb (both test and prod). 
- On the usage endpoint, we backfill orb customers and orb subscriptions based on the presence of `orb_customer_id` and `orb_subscription_id`. 
- We create a subscription to the free plan by default (I'm assuming that if there's no subscription in orb, then it has to be a free user, right?) *1
- We set the starting date of the subscription to the account creation date, so that any event ingested in the past is associated with the metrics.
- Allow user from any plan to see usage. *1

*1: Thinking about this now that I'm writing this, are there paying customers who are not on orb? How does this play with self-hosted and @TBonnin work on self-hosted billing?

---

This PR introduces significant changes to the billing infrastructure by ensuring all user accounts, including those on free plans, are fully tracked and backfilled as customers and subscriptions in Orb. It accomplishes this by introducing new generic billing utilities for customer/subscription linkage, updating onboarding/usage flows to invoke these, restructuring plan definitions for clear Orb mapping, and allowing in-app usage metrics viewing for all users regardless of plan. The billing client interface is extended for improved provider abstraction, with improved error handling and subscription backdating limited to Orb's 95-day restriction.

<details>

• Added utility functions (`linkBillingCustomer` and `linkBillingFreeSubscription`) to create/backfill Orb customers/subscriptions and sync DB state.
• Refactored plan definitions to individual plan objects (with `orbId`) and reorganized `plansList` composition.
• Updated onboarding (signup, managed callback) and usage endpoint flows to invoke new billing utilities and ensure Orb records are present (backfilling when necessary for legacy/new accounts).
• Relaxed usage metrics gating: all users (regardless of paid/free plan) can see in-app usage.
• Extended billing client/types: added `createSubscription` method and corresponding Orb client logic for backdated (max 95 days) subscription creation.
• Added Orb configuration to `.env.example`.
• Improved error messaging and interface boundary to reduce Orb-specific leakage into main codebase.

</details>

<details>

• Billing utilities (`packages/server/lib/utils/billing.ts`)
• Plan definitions (`packages/shared/lib/services/plans/definitions.ts`)
• User/account onboarding (`signup.ts`, `getCallback.ts`)
• Usage metrics endpoint (`getUsage.ts`)
• Billing client abstraction and Orb adapter
• .env.example

</details>


packages/types/lib/billing/types.ts
```diff
@@ -8,6 +8,7 @@ export interface BillingClient {
     linkStripeToCustomer(teamId: number, customerId: string): Promise<Result<void>>;
     getCustomer: (accountId: number) => Promise<Result<BillingCustomer>>;
     getSubscription: (accountId: number) => Promise<Result<BillingSubscription | null>>;
+    createSubscription: (team: DBTeam, planExternalId: string) => Promise<Result<BillingSubscription>>;
```
kaposke, line 11: This might be getting too Orb specific

packages/server/lib/controllers/v1/account/managed/getCallback.ts
```diff
@@ -118,14 +118,9 @@ export const getManagedCallback = asyncWrapper<GetManagedCallback>(async (req, r
...
-            if (resCreate.isErr()) {
-                report(resCreate.error);
-            } else {
-                const resUpdate = await updatePlanByTeam(db.knex, { account_id: account.id, orb_customer_id: resCreate.value.id });
-                if (resUpdate.isErr()) {
-                    report(resUpdate.error);
-                }
+            const res = await linkOrbCustomer(account, user);
+            if (res.isErr()) {
+                report(res.error);
```
bodinsamuel, line 123: can you update signup.ts too please?

packages/billing/lib/clients/orb.ts
```diff
@@ -77,6 +77,19 @@ export class OrbClient implements BillingClient {
         }
     }
 
+    async createSubscription(team: DBTeam, planExternalId: string): Promise<Result<BillingSubscription>> {
+        try {
+            const subscription = await this.orbSDK.subscriptions.create({
+                external_customer_id: String(team.id),
+                external_plan_id: planExternalId,
+                start_date: team.created_at.toISOString()
```
bodinsamuel: Does this modify the invoice start date or just the subscription start date? fyi invoice should always begin at the start of the month

packages/server/lib/utils/linkOrb.ts
```diff
@@ -0,0 +1,40 @@
+import { billing } from '@nangohq/billing';
```
bodinsamuel: linkOrb.ts seems really specific, just orb.ts would be fine imo

packages/server/lib/utils/linkOrb.ts
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    return Ok(resUpsert.value);
+}
+
+/**
+ * Creates an Orb subscription (free) and links it to the account plan
+ */
+export async function linkOrbFreeSubscription(account: DBTeam): Promise<Result<BillingSubscription>> {
+    const resCreate = await billing.createSubscription(account, 'free');
```
bodinsamuel: Can you use plans definitions `orbId` please

bodinsamuel (approved):

packages/billing/lib/clients/orb.ts
kaposke: I have to double check actually

packages/server/lib/utils/linkOrb.ts
kaposke: Oh nice, didn't know we had that.

packages/billing/lib/clients/orb.ts
kaposke: Checked. Invoicing is aligned with the start of the months.
While testing though, I've found out that the limit for backdating the start_date is 95 days, so I've added capping.

bodinsamuel (approved):

.env.example
```diff
@@ -109,8 +109,4 @@ FLAG_SERVE_CONNECT_UI=true
 
 # ---- AWS
 # AWS_ACCESS_KEY_ID=<ACCESS-KEY-ID>
-# AWS_SECRET_ACCESS_KEY=<SECRET-ACCESS-KEY>
-
-# ---- Orb
-# ORB_API_KEY=<ORB-API-KEY>
-# ORB_WEBHOOKS_SECRET=<ORB-WEBHOOKS-SECRET>
```
bodinsamuel: actually that was a good idea, I forgot to add them

kaposke: I was unsure about one thing: is this example for us, or for self-hosted? If it's for us, we're missing a bunch. If it's for  self-hosted, we do without.

bodinsamuel: it's for us, self-hosted don't use the repo directly, just the docker-compose.yaml. It's definitely missing a lot
edit: actually we recommend cloning but whatever it's better to document those

packages/server/lib/controllers/v1/account/signup.ts
```diff
@@ -111,13 +111,13 @@ export const signup = asyncWrapper<PostSignup>(async (req, res) => {
     }
 
     if (!token && flagHasUsage) {
-        const resCreate = await billing.upsertCustomer(account, user);
-        if (resCreate.isErr()) {
-            report(resCreate.error);
+        const linkOrbCustomerRes = await linkOrbCustomer(account, user);
```
TBonnin: one of the goal of the `billing` package is to encapsulate all the billing logic so orb stuff don't leak into the app. Can we keep it this way, renaming `utils/orb.ts` to `utils/billing.ts`. Ideally using a generic enough interface (ie: `linkBillingCustomer`( even though it might not be easy to fully abstract orb in the implementation.

packages/server/lib/controllers/v1/plans/usage/getUsage.ts
```diff
@@ -12,12 +13,30 @@ export const getUsage = asyncWrapper<GetUsage>(async (req, res) => {
...
+    if (!plan) {
         res.status(400).send({ error: { code: 'feature_disabled' } });
         return;
     }
 
+    // Backfill orb customer
+    if (!plan.orb_customer_id) {
+        const linkOrbCustomerRes = await linkOrbCustomer(account, user);
+        if (linkOrbCustomerRes.isErr()) {
+            res.status(500).send({ error: { code: 'server_error', message: 'Failed to link Orb customer' } });
```
TBonnin:
```suggestion
            res.status(500).send({ error: { code: 'server_error', message: 'Failed to link billing customer' } });
```
No need to mention `orb` imho

packages/server/lib/controllers/v1/plans/usage/getUsage.ts
```diff
@@ -12,12 +13,30 @@ export const getUsage = asyncWrapper<GetUsage>(async (req, res) => {
...
+            return;
+        }
+    }
+
+    // Backfill orb subscription (free by default)
+    if (!plan.orb_subscription_id) {
+        const linkOrbSubscriptionRes = await linkOrbFreeSubscription(account);
+        if (linkOrbSubscriptionRes.isErr()) {
+            res.status(500).send({ error: { code: 'server_error', message: 'Failed to link Orb subscription' } });
```
TBonnin:
```suggestion
            res.status(500).send({ error: { code: 'server_error', message: 'Failed to link billing subscription' } });
```

packages/server/lib/utils/orb.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (resUpsert.isErr()) {
+        return Err(resUpsert.error);
+    }
+
+    const resUpdate = await updatePlanByTeam(db.knex, { account_id: account.id, orb_customer_id: resUpsert.value.id });
+    if (resUpdate.isErr()) {
+        return Err(resUpdate.error);
+    }
+
+    return Ok(resUpsert.value);
```
TBonnin: can you not `return updatePlanByTeam(....)`

packages/server/lib/utils/orb.ts
```diff
@@ -0,0 +1,40 @@
...
+    if (resCreate.isErr()) {
+        return Err(resCreate.error);
+    }
+
+    const resUpdate = await updatePlanByTeam(db.knex, { account_id: account.id, orb_subscription_id: resCreate.value.id });
+    if (resUpdate.isErr()) {
+        return Err(resUpdate.error);
+    }
+
+    return Ok(resCreate.value);
```
TBonnin: same here about returning `updatePlanByTeam(...)`

packages/shared/lib/services/plans/definitions.ts
```diff
@@ -1,138 +1,153 @@
...
+    description: 'Congrats, you are an insider.',
+    canUpgrade: false,
+    canDowngrade: false,
+    cta: 'Contact Us',
+    hidden: true,
+    orbId: 'internal',
+    flags: {
+        api_rate_limit_size: 'l',
+        connection_with_scripts_max: null,
+        environments_max: 3,
```
TBonnin: I know you copied it but can we have way more than 3?

packages/shared/lib/services/plans/definitions.ts
```diff
@@ -1,138 +1,153 @@
...
+        has_otel: false,
+        has_sync_variants: false,
```
TBonnin: true, please! :smile:

packages/shared/lib/services/plans/definitions.ts
```diff
@@ -1,138 +1,153 @@
...
+        name: 'yc',
```
TBonnin, line 98:
```suggestion
        name: 'internal',
```

packages/types/lib/billing/types.ts
TBonnin: let's hope that if we ever switch to another billing provider, they also have a concept of plan and use a string as `id` ;)

packages/server/lib/utils/orb.ts
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Creates an Orb customer (if it doesn't exist) and links it to the account plan
+ */
+export async function linkOrbCustomer(account: DBTeam, user: DBUser): Promise<Result<BillingCustomer>> {
+    const resUpsert = await billing.upsertCustomer(account, user);
+    if (resUpsert.isErr()) {
+        return Err(resUpsert.error);
+    }
+
+    const resUpdate = await updatePlanByTeam(db.knex, { account_id: account.id, orb_customer_id: resUpsert.value.id });
```
TBonnin: those 2 queries shows that maybe our data model is not ideal. Is the `orb_customer_id` required in the plan table. Maybe we should have a self-contained billing table that links account id and orb id. Would make sense since we are planning to have a dedicated `billing` service

packages/server/lib/utils/orb.ts
kaposke: Notice we return `resUpsert.value` and not `resUpdate.value`

packages/shared/lib/services/plans/definitions.ts
kaposke: there's another one for internal, are they supposed to be the same?

packages/types/lib/billing/types.ts
kaposke: Yeah, unfortunately I don't believe such migration will be flawless 1:1. But the interface you made should save us lots of time still.

packages/server/lib/utils/orb.ts
kaposke: I agree, but would like to keep it out of the scope for this one

PR #4277, opened 2025-07-02
fix(providers): Move Figma SCIM tenantId from endpoint to base URL for proper SCIM API calls

yksolanki9:
Background:
- PR #4252 fixed credential verification by properly interpolating connectionConfig references in endpoints. However, this approach only ensured `tenantId` was available during verification, not for subsequent API calls.

Correct Solution:
- Moved `tenantId` from individual endpoint paths to the base URL configuration `https://${connectionConfig.host}/scim/v2/${connectionConfig.tenantId}`. 
- This change ensures that the tenant ID is automatically included in all SCIM API calls.

---

This PR updates the Figma SCIM provider configuration by moving the tenantId interpolation from individual endpoint paths into the base URL in providers.yaml. This ensures tenantId is consistently available for all SCIM API calls, streamlining endpoint definition and fixing coverage beyond credential verification.


packages/providers/providers.yaml
```diff
@@ -3694,13 +3694,13 @@ figma-scim:
         - productivity
     auth_mode: API_KEY
     proxy:
-        base_url: https://${connectionConfig.host}/scim
+        base_url: https://${connectionConfig.host}/scim/v2/${connectionConfig.tenantId}
```
hassan254-prog, line 3791: This means the API version number would need to be included in the `baseURL`, which we strongly advise against. Instead, we typically recommend placing the API version in the `endpoint` within the integration itself. You can use the `getConnection` method to dynamically populate this in the integration. cc @khaliqgant

khaliqgant: Since the tenant id needs to be there and this is SCIM I think this is fine in this case

yksolanki9: I agree, making two api calls is redundant. 

A better solution might be to add the base url instead of the tenant id in connection config. This way, if and when the api version changes, only the connection needs to be updated and different connections can be on different versions?

hassan254-prog: To add to what @khaliqgant mentioned, I think that's okay, you can keep this as is: `base_url: https://${connectionConfig.host}/scim/v2/${connectionConfig.tenantId}`. Then revert the logic in `buildProxyURL` to allow for other providers that may use it later on.

hassan254-prog: Hey @yksolanki9, just bumping this 🙂.

yksolanki9: Hey @hassan254-prog, have made the changes. Please have a look

hassan254-prog: looks good 👍, @khaliqgant before merging could you please verify we don't have any active connections present.

hassan254-prog (approved):

packages/providers/providers.yaml
khaliqgant: Confirmed, thanks both!

PR #4305, opened 2025-07-09
fix(proxy): retry only on specific error codes for some providers

hassan254-prog:

- Some providers appear to always return a retry header regardless of the status code. This causes Nango to retry requests even when it should return an error immediately. A good example is Exact Online: when a request fails with status 500, it still includes an x-ratelimit-reset header, which always seems to reset at midnight. As a result, we end up retrying this status instead of failing fast.

- This PR also improves backward compatibility for providers that rely on status code handling in the code.

---

This PR overhauls the Nango proxy retry mechanism to allow precise provider-level configuration of retryable status codes via arrays of exact codes, wildcard patterns (e.g., '5xx'), or numeric ranges in providers.yaml. The implementation refactors the core retry logic, introduces a status code matcher, updates types and schema validation, amends multiple key provider YAML definitions (notably GitHub, Google Calendar, Linear, Exact Online), and significantly expands unit test coverage for all new behaviors. Changes are backward-compatible and resolve problems where certain providers (e.g., Exact Online) returned retry headers inappropriately, resulting in excessive retries on non-retryable errors.

<details>

• Proxy retry logic refactored to support provider-specific, array-based error_code rules with support for patterns like '5xx' and ranges ('500-502').
• Addition of a matchesStatusCode helper to match status codes to new flexible rules.
• providers.yaml updated for multiple providers (GitHub, Google Calendar, Linear, Exact Online, etc.) to specify new error_code arrays.
• RetryHeaderConfig type and provider schema updated to accept string[] error_code.
• Improved JSON schema disallows spaces in error_code and enforces array entries.
• Expanded and more comprehensive tests for all retry scenarios including edge cases.
• Preserves backward compatibility for providers without explicit error_code configuration.

</details>

<details>

• Proxy retry handling (packages/shared/lib/services/proxy/retry.ts)
• Provider configuration and schema (providers.yaml and scripts/validation/providers/schema.json)
• Typescript type definitions (packages/types/lib/proxy/api.ts)
• Unit tests (packages/shared/lib/services/proxy/retry.unit.test.ts)

</details>


packages/shared/lib/services/proxy/retry.ts
```diff
@@ -214,3 +218,31 @@ function parseRetryValue({
...
+                case '<=':
+                    return status <= numeric;
+            }
+        }
+    } else if (/^\d+$/.test(trimmed)) {
+        return status === Number(trimmed);
+    }
+
+    return false;
+}
```
TBonnin, line 245: can we simplify how we model range of error codes? 
The simplest would be `5xx`. If we want slightly more fine grained ranges we could model it as range like `500-502`. The implementation of this function would be slightly simplified and not relying on a regex

TBonnin (commented):
I found this kind of logic is gonna be a game of cat and mouse, trying to deal with all the quirks of external apis. But that something we already had in the codebase anyway.
Made a comment about how to define range of error code though

packages/shared/lib/services/proxy/retry.ts
bodinsamuel: I agree

hassan254-prog: Simplifies the logic, thanks, updated with the latest [commit](https://github.com/NangoHQ/nango/pull/4305/commits/ad856669976ca47b81a40b8f33e54a5e9d0acbe5).

packages/shared/lib/services/proxy/retry.ts
```diff
@@ -214,3 +218,30 @@ function parseRetryValue({
 
     return { found: false, reason: `unknown_type:${type}` };
 }
+function matchesStatusCode(status: number, rule: string): boolean {
+    if (!rule) return false;
+
+    const trimmed = rule.trim();
```
bodinsamuel: I think we can just disallow space in the JSON schema that validates providers.yaml

packages/shared/lib/services/proxy/retry.ts
```diff
@@ -214,3 +218,30 @@ function parseRetryValue({
...
+
+    // Handle exact match (e.g., "403")
+    if (/^\d{3}$/.test(trimmed)) {
```
bodinsamuel: !isNaN(Number(rule))
might be faster

bodinsamuel (approved):

packages/shared/lib/services/proxy/retry.ts
hassan254-prog: thanks, makes it easier, updated with the latest [commit](https://github.com/NangoHQ/nango/pull/4305/commits/521a67b1969ad35bc0d5922b74bbdf56a40b5cce).

bodinsamuel (approved):

PR #4315, opened 2025-07-11
fix(jira-post-connection): [nan-3592] wrap potential failing call in a try catch

khaliqgant:
Wrap potentially failing calls in a try, catch. I thought about checking explicitly for scopes but this seemed like a potential cat/mouse game in case granular or classic scopes change.

---

This PR enhances the reliability of the Jira post-connection logic by wrapping API calls that require specific permissions or scopes in a try/catch block. It introduces logging for failed calls and prevents the entire connection process from breaking due to insufficient scopes, without explicitly checking for scope presence.

### DIFF
--- a/packages/cli/fixtures/zero/valid/github/actions/createIssue.ts
+++ b/packages/cli/fixtures/zero/valid/github/actions/createIssue.ts
@@ -34,4 +34,3 @@ export default createAction({
         });
     }
 });
-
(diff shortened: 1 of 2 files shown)


TBonnin (approved):

khaliqgant:
As noted in Slack we should be very conservative in what these post connection scripts are doing and shouldn't use calls that require certain scopes. The script should be light weight and 100% fail safe cc/ @hassan254-prog

TBonnin:
> As noted in Slack we should be very conservative in what these post connection scripts are doing and shouldn't use calls that require certain scopes. The script should be light weight and 100% fail safe cc/ @hassan254-prog

what is not 100% clear to me is the executing of post-connection logic is itself wrapped into a try/catch and logged, so I don't fully understand why a failure would bubble up

PR #4320, opened 2025-07-11
fix(github-app): handle update without auth code

khaliqgant:
If a github app has already been installed the OAuth flow isn't kicked off for some reason. In this case we don't need the auth user since there is no approval flow since the app has already been installed. So on an update short circuit the flow to allow the connection to be created

---

This PR updates the OAuth flow for GitHub App integrations to support cases where the app has already been installed and an OAuth code is not present during an 'update' action. It introduces logic in the OAuth controller to conditionally bypass the need for an authorization code on such updates, and adjusts connection service response values to support proper handling.

<details>

• Added a conditional path in oauth.controller.ts to allow updates without an authorization code if authMode is 'CUSTOM' and setupAction is 'update'.
• Creates the connection using available installation/app IDs without requiring a new OAuth token when updating an already installed GitHub app.
• Changed the return type of getAppCredentialsAndFinishConnection in connection.service.ts to return the upserted connection result, aligning with the needs of the new callback flow.

</details>

<details>

• OAuth2 callback handling in packages/server/lib/controllers/oauth.controller.ts
• App credentials connection flow in packages/shared/lib/services/connection.service.ts

</details>

### DIFF
--- a/packages/server/lib/controllers/oauth.controller.ts
+++ b/packages/server/lib/controllers/oauth.controller.ts
@@ -909,6 +909,81 @@ class OAuthController {
         const callbackMetadata = getConnectionMetadataFromCallbackRequest(req.query, provider);
 
         const installationId = req.query['installation_id'] as string | undefined;
+        const authMode = session.authMode;
+        const setupAction = req.query['setup_action'] as string | undefined;
+
+        if (!authorizationCode && authMode === 'CUSTOM' && setupAction === 'update') {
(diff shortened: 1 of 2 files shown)


bodinsamuel (approved):

khaliqgant:
https://stackoverflow.com/questions/68347504/github-apps-authorize-request-on-applications-doesnt-return-code-parameter

PR #4325, opened 2025-07-15
fix(runner): prevent parallel run and zombie tasks

bodinsamuel:
## Changes

- Prevent zombie tasks by allowing runner to auto-kill it own tasks
Passing down heartbeatTimeoutSecs so we know when to cut the task

- Prevent unwanted parallel runs for syncs 
Just checking if we have the same syncId running should be enough (but maybe I forgot something) 

---

This PR revamps runner task execution and heartbeat management to prevent both unintentional parallel execution of syncs and 'zombie' tasks that hang without proper status reporting. Core changes include propagation of a new `heartbeatTimeoutSecs` property throughout task orchestration and runner code, aggressive enforcement of this value to auto-abort tasks missing heartbeats, and new logic to ensure sync jobs with the same syncId do not run concurrently at the runner level. Extensive type updates, schema validation changes, CLI/test support, and job property propagation ensure all workflow entrypoints adhere to these constraints. This elevates reliability, resource efficiency, and correct concurrency management in job execution.

<details>

• Introduced and propagated heartbeatTimeoutSecs throughout orchestrator, jobs, runner, types, schemas, and test/CLI infrastructure.
• Runner implements auto-abort/self-kill for tasks missing heartbeats past timeout, preventing zombie/stale processes.
• Added parallelism guard to runner (monitor.hasConflictingSync), blocking concurrent sync jobs with the same syncId.
• Validation schemas (Zod) and type definitions updated for new property.
• All job types (sync, webhook, action, on-event, abort) updated to accept and respect heartbeatTimeoutSecs.
• CLI and test mocks updated to supply heartbeatTimeoutSecs to NangoProps.
• Tasks and orchestration code updated to pass down and validate heartbeatTimeoutSecs.
• Test code updated to reflect new required/optional fields.

</details>

<details>

• Runner execution/server code (packages/runner/lib/server.ts)
• Runner concurrency monitor (packages/runner/lib/monitor.ts)
• Task type definitions and orchestrator validation (packages/orchestrator/lib/clients/types.ts, validate.ts)
• Job execution handlers (packages/jobs/lib/execution/*)
• Job/runner APIs and route handlers (packages/jobs/lib/routes/tasks/putTask.ts)
• CLI and test infrastructure
• Type definitions across orchestrator/jobs/runner/types

</details>


packages/jobs/lib/routes/tasks/putTask.ts
```diff
@@ -24,6 +24,7 @@ const nangoPropsSchema = z
             id: z.number(),
             name: z.string().min(1)
         }),
+        heartbeatTimeoutSecs: z.number().optional(),
```
bodinsamuel, line 27: optional during migration

rossmcewan (approved):

packages/runner/lib/monitor.ts
```diff
@@ -114,6 +114,20 @@ export class RunnerMonitor {
             this.checkIdle(nextTimeout);
         }, timeoutMs);
     }
+
+    hasConflictingSync(newTask: NangoProps): boolean {
+        if (newTask.scriptType !== 'sync') {
+            return false;
+        }
+
+        for (const task of this.tracked.values()) {
```
TBonnin, line 123: I would have modified `tracked` to use syncId as the map key and fail `track()` if key already exists

packages/types/lib/runner/sdk.ts
```diff
@@ -30,6 +30,7 @@ export interface NangoProps {
     debug: boolean;
     startedAt: Date;
     endUser: { id: number; endUserId: string | null; orgId: string | null } | null;
+    heartbeatTimeoutSecs?: number | undefined;
```
TBonnin, line 33: I am not a huge fan of leaking task scheduling info into the nango props and add scheduling related logic into the runner. Especially since we have not confirmed with a real example that this scenario where heartbeat and abort are both not happening

bodinsamuel: Locally, you can have a zombie task; you just need to kill the orchestrator (or make it unavailable). If you combine that with an unreliable network, even without repro, I wouldn't trust two back-to-back HTTP calls to never fail.
You have a better strategy?

packages/runner/lib/monitor.ts
bodinsamuel: True, maybe it's a bit more explicit like this imo but I can change

packages/types/lib/runner/sdk.ts
TBonnin: I don't have a better strategy and I agree that in theory it could happen but I don't think the scenario this PR is fixing is what happen today with our customer.

PR #4326, opened 2025-07-15
fix: heartbeat ms -> sec, set last_heartbeat_at on started

bodinsamuel:
## Changes

- Fix using value as ms instead of secs
It was in the name but still made the mistake ^^' 

- Set `last_heartbeat_at` when we start a task
This can prevent expiring task that took too long to schedule

- Renamed type DbTask to DBTask because a variable was already named like this and it was actually not working as intended

---

This PR standardizes heartbeat timeout handling by ensuring all references and logic across the scheduler, orchestrator, runner, and jobs packages use seconds instead of milliseconds. It also sets the `last_heartbeat_at` timestamp when tasks transition to the 'STARTED' state, updates the runner to prevent parallel execution of tasks with the same sync ID, and globally renames the `DbTask` type/interface to `DBTask`. Corresponding updates are made to validation logic and tests.

<details>

• All heartbeat timeout logic and type fields now use and expect seconds rather than milliseconds (`heartbeatTimeoutSecs`).
• Explicitly sets `last_heartbeat_at` when transitioning tasks to 'STARTED' to prevent premature expiry.
• Prevents runner from executing new tasks for a sync ID already in progress, reducing possible race conditions.
• Renames `DbTask` to `DBTask` throughout code and tests for consistency.
• Updates type usage, DB code, and tests to align with new naming and heartbeat timeout semantics.

</details>

<details>

• scheduler task model and database logic (`models/tasks.ts`)
• runner logic and heartbeat management (`runner/lib/server.ts`)
• scheduler test coverage (`scheduling.integration.test.ts`)

</details>


rossmcewan (approved):

TBonnin (commented):
https://github.com/NangoHQ/nango/pull/4326/files#diff-d024d64a76fc8583777783ea2786f70cdd25364a55160884e64c474be587e503R320 looks good.

As I commented on the other PR adding heartbeat logic into the runner is imho unecessary

bodinsamuel:
> As I commented on the other PR adding heartbeat logic into the runner is imho unecessary

We can discuss a better strategy when we have time, for sure, but I feel like my point is still standing. HTTP is not a reliable API, and with the current code, the transaction does not roll back on failing HTTP calls, + there is no retry. 

We can rework `onExpiring` to truly await a successful reply, with the risk of keeping a very long transaction and all that implies for the DB/memory, or only get the expired task, send the cancellation events, and let the runner bubble up the success instead.

PR #4346, opened 2025-07-21
fix(billing): feedback

bodinsamuel:
## Changes

- Use `stripe_charge` instead of `stripe_invoice` 
This allow Orb to do the full invoicing and collection

- Apply changes directly instead of waiting for webhooks
I wanted to have limited touch points that would change the plans but Orb reliability is not helping

- Replicate billing address from payment method to customer in Stripe so that it's available in Orb too

---

This PR refactors the billing integration by switching from `stripe_invoice` to `stripe_charge` for Orb, enabling full invoicing and auto-collection through Orb. It also changes the plan upgrade flow so that plan changes are applied immediately after stripe payment success (bypassing unreliable Orb webhooks), and synchronizes billing address and name from Stripe payment methods to customer records, ensuring accurate customer data in Orb. Type signatures and interfaces are updated to reflect these changes, and the frontend now routes non-upgradeable plan CTAs to a demo page.

<details>

• Orb integration now uses `stripe_charge` with `auto_collection: true` for customer management.
• Plan changes are applied directly after Stripe payment success rather than waiting for Orb webhooks.
• Billing address and name from payment method updates are replicated to Stripe customer records.
• Updated `applyPendingChanges` in the billing client to return a subscription object instead of void; updated type signatures accordingly.
• Frontend PlanCard CTA for non-changeable plans now links to the demo page.
• Backend logic for syncing and updating plans is duplicated to handle Orb webhook latency.

</details>

<details>

• Backend: Stripe webhook handler (postWebhooks.ts)
• Billing client integration with Orb (orb.ts)
• Type interfaces for billing (billing/types.ts)
• Frontend: Billing plan selection UI (PlanCard.tsx)

</details>


packages/server/lib/controllers/v1/stripe/postWebhooks.ts
```diff
@@ -79,6 +79,25 @@ async function handleWebhook(event: Stripe.Event): Promise<Result<void>> {
             return Ok(undefined);
         }
 
+        // payment method was updated through our UI
+        // we replicate to the customer to keep it in sync and to be able to generate invoices
+        // It might be incorrect in some cases, but it's better than nothing
```
kaposke, line 84: In what case can this be incorrect?

packages/server/lib/controllers/v1/stripe/postWebhooks.ts
```diff
@@ -141,7 +160,42 @@ async function handleWebhook(event: Stripe.Event): Promise<Result<void>> {
                 return Err(resApply.error);
             }
 
-            return Ok(undefined);
+            // This operation is also done in orb/postWebhooks
+            // But their webhook system is so slow that we need to duplicate the logic here
```
kaposke, line 164: Is it not extractable into a function? I understand if it has enough differences to not be worth it.

kaposke (approved):

packages/server/lib/controllers/v1/stripe/postWebhooks.ts
bodinsamuel: when the card billing address is not the customer actual address for example

packages/server/lib/controllers/v1/stripe/postWebhooks.ts
bodinsamuel: yeah it's slightly different, mostly because we don't receive the same info, duplication is more flexible in that case

PR #4351, opened 2025-07-21
fix(cli): replace depreciated endpoint

hassan254-prog:

-  Replace usage of a deprecated endpoint with the new current version

---

This PR replaces usage of the deprecated `/config/{providerConfigKey}` endpoint with the updated `/providers/{providerConfigKey}` endpoint in the CLI codebase. The change affects two files to ensure the CLI interacts with the correct API version.

### DIFF
--- a/packages/cli/lib/utils.ts
+++ b/packages/cli/lib/utils.ts
@@ -201,7 +201,7 @@ export async function getConnection(
 }
 
 export async function getConfig(providerConfigKey: string, debug = false) {
-    const url = process.env['NANGO_HOSTPORT'] + `/config/${providerConfigKey}`;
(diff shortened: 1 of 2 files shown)


bodinsamuel (approved):

packages/cli/lib/utils.ts
```diff
@@ -201,7 +201,7 @@ export async function getConnection(
 }
 
 export async function getConfig(providerConfigKey: string, debug = false) {
-    const url = process.env['NANGO_HOSTPORT'] + `/config/${providerConfigKey}`;
```
bodinsamuel, line 204: actually it should be /integrations no?

PR #4383, opened 2025-07-25
chore(orchestrator): add next execution date and task state to schedules

TBonnin:
The goal is to speed up the scheduling query by removing the join with the tasks table and the inline calculation of the due date. This PR:
- duplicates the task state into the schedules table in order to avoid a join in the scheduling query
- Also calculate and store the schedule next execution date when a task completes instead of doing it at scheduling time

The cost is an extra db query to update a schedule when a task completes and the added complexity to keep the state of tasks in sync but the benefits of removing the join and the inlined due date calculation will greatly speed up the scheduling query

This PR doesn't use the new columns for the scheduling of syncs yet. In next PR I would modify the query as such:
```
SELECT
   *
FROM
   "nango_scheduler"."schedules"
WHERE
   "state" = 'STARTED'
   AND "last_scheduled_task_state" NOT IN ('CREATED', 'STARTED')
   AND "next_execution_at" <= CURRENT_TIMESTAMP
   FOR UPDATE SKIP LOCKED
```

---

This PR introduces a major architectural change by denormalizing key task metadata (last scheduled task state and next execution date) directly into the schedules table, eliminating the need for expensive joins and inline due-date calculation during scheduling queries. This is implemented via a complex DB migration, extensive model and API adjustments, new batch update methods, and wide-ranging changes to daemons and main orchestrator logic to keep schedule state in sync. The new fields are written in all relevant flows (creation, state transitions, expiring), but are not yet actively used to drive the main scheduling query-adoption for that is staged for a future PR. Comprehensive tests cover both the new data flow and the legacy behavior to ensure correctness.

<details>

• Database migration: Adds last_scheduled_task_state and next_execution_at columns to schedules, backfills data, and creates optimized indices for fast scheduling queries.
• Model/API changes: Schedule types, DB conversion routines, API contracts, and validation are updated to handle new fields.
• New batch operations: Implements setLastScheduledTask and updateLastScheduledTaskState for atomic, efficient schedule updates after batch task operations.
• Refactored daemons/services: Scheduling and expiring daemons now use batch updates to set schedule fields after task processing; main scheduler logic updated to maintain consistency.
• Tests: Extensive test suite updates and new assertions/matchers verify that schedule/task fields are synchronized as expected throughout the task lifecycle.
• Auxiliary changes: Test matchers are extended for date tolerances; type definitions updated for the new model contracts.

</details>

<details>

• Database schema and migrations (schedules table)
• Task-schedule relationship logic
• Scheduling and expiring daemons, main orchestrator logic
• All API contracts and validation layers around schedules
• Test suites for schedule/task integration

</details>

### DIFF
--- a/packages/scheduler/lib/db/migrations/20250724115100_schedule_next_execution_at.ts
+++ b/packages/scheduler/lib/db/migrations/20250724115100_schedule_next_execution_at.ts
@@ -31,9 +31,9 @@ export async function up(knex: Knex): Promise<void> {
             last_scheduled_task_state = ts.last_task_state,
             next_execution_at = (
                 CASE
-                    WHEN ts.last_task_state = 'SUCCEEDED'
-                    THEN s.starts_at + (ceiling(extract(EPOCH FROM (now() - s.starts_at)) / extract(EPOCH FROM s.frequency)) * s.frequency)
-                    ELSE s.starts_at + (floor(extract(EPOCH FROM (now() - s.starts_at)) / extract(EPOCH FROM s.frequency)) * s.frequency)
+                    WHEN ts.last_task_state IS NULL
+                    THEN s.starts_at + (floor(extract(EPOCH FROM (now() - s.starts_at)) / extract(EPOCH FROM s.frequency)) * s.frequency)
+                    ELSE s.starts_at + (ceiling(extract(EPOCH FROM (now() - s.starts_at)) / extract(EPOCH FROM s.frequency)) * s.frequency)
                 END
             )
         FROM task_states ts
--- a/packages/scheduler/lib/db/migrations/20250724115100_schedule_next_execution_at.ts
+++ b/packages/scheduler/lib/db/migrations/20250724115100_schedule_next_execution_at.ts
@@ -1,0 +1,60 @@
+import type { Knex } from 'knex';
+
+import { SCHEDULES_TABLE } from '../../models/schedules.js';
+import { TASKS_TABLE } from '../../models/tasks.js';
+
+export const config = {
+    transaction: false
+};
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.raw(
+        `ALTER TABLE ${SCHEDULES_TABLE}
+            ADD COLUMN IF NOT EXISTS last_scheduled_task_state task_states DEFAULT NULL,
+            ADD COLUMN IF NOT EXISTS next_execution_at TIMESTAMP WITH TIME ZONE DEFAULT NULL;`
+    );
+
+    // Update existing schedules
+    // last_scheduled_task_state is read from tasks table
+    // if task exists and is SUCCEEDED next_execution_at is next due date calculatd using ceiling()
+    // if no task exists or task is not SUCCEEDED, next_execution_at is last due date calculated using floor()
+    await knex.raw(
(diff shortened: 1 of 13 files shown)


packages/scheduler/lib/db/migrations/20250724115100_schedule_next_execution_at.ts
```diff
@@ -0,0 +1,60 @@
...
+            next_execution_at = (
+                CASE
+                    WHEN ts.last_task_state = 'SUCCEEDED'
+                    THEN s.starts_at + (ceiling(extract(EPOCH FROM (now() - s.starts_at)) / extract(EPOCH FROM s.frequency)) * s.frequency)
+                    ELSE s.starts_at + (floor(extract(EPOCH FROM (now() - s.starts_at)) / extract(EPOCH FROM s.frequency)) * s.frequency)
+                END
+            )
+        FROM task_states ts
+        WHERE s.id = ts.schedule_id;`
+    );
```
TBonnin, line 41: schedules table in prod is small (around 70k entries). This query should be relatively fast

packages/scheduler/lib/models/schedules.ts
```diff
@@ -192,6 +207,60 @@ export async function update(
     }
 }
 
+export async function setLastScheduledTask(db: knex.Knex, updates: { id: string; taskId: string; taskState: TaskState }[]): Promise<Result<Schedule[]>> {
```
TBonnin, line 210: The two new functions accept an array as input in order to do a single query when scheduling/expiring a batch of tasks

packages/scheduler/lib/db/migrations/20250724115100_schedule_next_execution_at.ts
```diff
@@ -0,0 +1,60 @@
...
+        `CREATE INDEX CONCURRENTLY IF NOT EXISTS "idx_schedules_scheduling"
+            ON ${SCHEDULES_TABLE} (state, last_scheduled_task_state, next_execution_at)
+            WHERE state = 'STARTED';`
+    );
+    // index on last_scheduled_task_id to update last_scheduled_task_state
+    await knex.raw(
+        `CREATE INDEX CONCURRENTLY IF NOT EXISTS "idx_schedules_last_scheduled_task_id"
+            ON ${SCHEDULES_TABLE} (last_scheduled_task_id)
+            WHERE last_scheduled_task_id IS NOT NULL;`
+    );
```
bodinsamuel, line 57: are you sure you want to create the column and update the values at the same time? 
also (unless mistaken) there will be a little delay between this migration and the new scheduler take over so a few tasks will not have the values

bodinsamuel (approved):

packages/scheduler/lib/db/migrations/20250724115100_schedule_next_execution_at.ts
TBonnin: The new columns are not read yet. Only written until I modified the scheduling query. I am happy to split the migration to add/update the column in its own PR

TBonnin: split the db migration to https://github.com/NangoHQ/nango/pull/4397

PR #4389, opened 2025-07-28
feat(providers): add migrations for pre-provisioning provider credentials

hassan254-prog:

- This PR introduces the initial database migrations and types required to support pre-provisioning of provider credentials.

---

This PR implements the foundational support for pre-provisioning provider credentials by introducing a new `providers_shared_credentials` table with a unique constraint, JSON credential storage, and proper timestamping. It extends the `_nango_configs` integration configs table and related TypeScript types with an optional, indexed `shared_credentials_id` foreign key, ensuring integration configs can now reference shared credentials. The changes are thoroughly propagated through the service layer, controller logic, seeder utilities, and all relevant test suites, defaulting the new linkage field to `null` (not yet exposed via outward APIs).

<details>

• Adds `providers_shared_credentials` table with unique `name`, JSON credentials, indexed PK, and enforced constraints.
• Extends `_nango_configs` with nullable, FK-indexed `shared_credentials_id` for referencing shared credentials.
• Propagates new shared credentials field to all relevant TypeScript types (IntegrationConfig, Provider), DB models, and interfaces.
• Updates service logic, seeder helpers, and endpoints to set and handle `shared_credentials_id` (defaults to null).
• Amends and validates all affected test suites to ensure the field is present in integration config objects.
• Adjusts API DTOs to explicitly omit `shared_credentials_id`, keeping internal field hidden from public responses.

</details>

<details>

• Integration config models and types (shared, provider, DB)
• Integration config service layer and controllers
• Seeder utilities
• Persist, webhook, and config service tests
• API response shaping/types

</details>


packages/database/lib/migrations/20250728113758_add_user_defined_to_configs.cjs
```diff
@@ -0,0 +1,14 @@
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`
+        ALTER TABLE _nango_configs
+        ADD COLUMN user_defined BOOLEAN NOT NULL DEFAULT TRUE
```
bodinsamuel: Took the time to think. I guess we will query the table `providers_shared_credentials` and copy the credentials to the new row `_nango_configs`? 
I have two propositions: 
- we could either have a column named `shared_credentials_id` (int nullable) that can point to the actual credentials that were used.
- changed the name to `using_shared_credentials` (boolean false) for explicitness

It's not a big change, but `user_defined` is too blurry as we don't know what is not `user_defined` when reading at it.
Let me know what you think ☺️

bodinsamuel (approved):
Looks good, one open comment 👍🏻

packages/database/lib/migrations/20250728113758_add_user_defined_to_configs.cjs
hassan254-prog: Both options sound good, especially option one. It would also make it easier for us to rotate existing integration keys. I spoke with Bastien, and the current flow is to rotate the keys after integrations have been created. Maybe we can improve this by going with option one and slightly adjusting the flow, to allow for key rotation this way. wdyt?

packages/database/lib/migrations/20250728111830_create_providers_shared_credentials_table.cjs
```diff
@@ -0,0 +1,16 @@
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    return knex.schema.createTable('providers_shared_credentials', function (table) {
+        table.increments('id').primary();
+        table.string('name').notNullable().unique();
+        table.json('credentials').notNullable();
+        table.timestamps(true, true);
+    });
```
TBonnin: I personally prefer raw sql instead of knex abstraction because one must understand what `increments` and `timestamps` translate to anyway. raw sql saves some cognitive load imho

packages/database/lib/migrations/20250728113758_add_user_defined_to_configs.cjs
TBonnin: let's also consider the access pattern. Storing a reference (option 1) requires a join. Depending on the size of the table and if how fast/often retrieving the credentials must happen it might make sense to go for option 2.

hassan254-prog: @TBonnin the spec has changed, we now need to rotate both existing and new integrations. In my opinion, using a foreign key might be a better approach, but I’m curious: do you have a better way of handling this?

TBonnin: in this case we need a reference to the shared credentials. The question about access pattern is still valid though: do we copy the shared credentials which means the rotation is more involved and requires updating the _nango_configs table but accessing the credentials doesn't require a join. If credentials are not copied, the rotation would be simpler but fetching the credentials requires a join if shared credentials. I don't know about the access pattern so you call

hassan254-prog: I’ve decided to implement an indexed foreign key to reference shared credentials. This approach enables centralized management, making credential rotation more streamlined. It also minimizes data duplication and reduces the operational complexity involved in updating multiple integrations during rotation events. Given that the `shared_credentials` table is expected to remain relatively small, the additional overhead of a join seems minimal in this context.

packages/database/lib/migrations/20250728111830_create_providers_shared_credentials_table.cjs
```diff
@@ -0,0 +1,23 @@
...
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`
+        CREATE TABLE IF NOT EXISTS ${SHARED_CREDENTIALS_TABLE} (
+            id SERIAL PRIMARY KEY,
+            name VARCHAR(255) NOT NULL UNIQUE,
+            credentials JSON NOT NULL,
+            created_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP,
+            updated_at TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP
```
TBonnin: created_at and updated_at should be `not null`. no?

packages/server/lib/controllers/integrations/postIntegration.ts
```diff
@@ -76,7 +76,8 @@ export const postPublicIntegration = asyncWrapper<PostPublicIntegration>(async (
         unique_key: body.unique_key,
         custom: null,
         missing_fields: [],
-        forward_webhooks: body.forward_webhooks ?? true
+        forward_webhooks: body.forward_webhooks ?? true,
+        shared_credentials_id: null // to add logic later, right now we just have this as true since its user defined
```
TBonnin: comment isn't accurate I think

TBonnin (approved):

packages/database/lib/migrations/20250728111830_create_providers_shared_credentials_table.cjs
hassan254-prog: Good catch thanks, updated with the latest [commit](https://github.com/NangoHQ/nango/pull/4389/commits/1619981612445b95b5fe1ac6fd82f86b7aca6c75).

packages/server/lib/controllers/integrations/postIntegration.ts
hassan254-prog: comment left for the next pr 😄.

PR #4399, opened 2025-07-29
fix(connect-ui): use custom integration display name

kaposke:
- `provider.display_name`: the default display name set in providers.yaml.
- `integration.display_name`: the display name override that one can customize.

Fixes https://nangohq.slack.com/archives/C08C889EBND/p1753471505881179

---

This PR updates the Connect UI logic to display the custom integration display name, when available, instead of always using the provider's default display name. The changes ensure that all relevant UI strings reference the integration's overridden display name, enhancing user clarity for customized integrations.


bodinsamuel (approved):
Thanks I had push a PR but forgot to open it 🤦🏻‍♂️

PR #4403, opened 2025-07-30
feat(webapp): refresh usage on connection deleted

kaposke:
Small feat to make sidebar usage update immediately when a connection is deleted.
Also, making usage update more often. The usage query should be quick (fetches from redis), and it feels very nice to see it update "live".

---

This PR updates the webapp to immediately refresh the usage statistics shown in the sidebar when a connection is deleted. It also decreases the refetch interval for usage stats from 1 minute to 10 seconds, ensuring more real-time updates across the UI.

### DIFF
--- a/packages/webapp/src/hooks/usePlan.tsx
+++ b/packages/webapp/src/hooks/usePlan.tsx
@@ -63,7 +63,7 @@ export function useApiGetUsage(env: string) {
 
             return json;
         },
-        refetchInterval: 1000 * 60 // 1 minute
+        refetchInterval: 1000 * 10 // 10 seconds
     });
 }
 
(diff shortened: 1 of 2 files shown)


packages/webapp/src/hooks/usePlan.tsx
```diff
@@ -63,7 +63,7 @@ export function useApiGetUsage(env: string) {
 
             return json;
         },
-        refetchInterval: 1000 * 60 // 1 minute
+        refetchInterval: 1000 * 10 // 10 seconds
```
TBonnin, line 66: is 10s not a bit aggressive?

TBonnin (approved):

packages/webapp/src/hooks/usePlan.tsx
kaposke: Seems to align with other refetchIntervals, but I'm fine with reducing it as well.

PR #4404, opened 2025-07-30
fix(server): don't cap a connection from finishing

kaposke:
Following this discussion: https://linear.app/nango/issue/NAN-3382/explicit-caps-for-free-plans#comment-1ae77e1e

We don't want to block connect-sessions that where already started. We only want to cap a connect-session from being created. This means that it's possible to exceed the connection-cap by a few connections if multiple connect-sessions are opened before the limit is reached, but we're considering that acceptable.

Renamed: `resourceCapping` to `connectionCapping` - no further changes to file

> Note: we also already cap "importing" connections from `POST /connection`.

---

This PR updates the capping mechanism for user connections by replacing the 'resourceCapping' middleware with the new 'connectionCapping' middleware. The new logic ensures only the initiation of new connect sessions is blocked when the plan limit is reached, while sessions already in progress can finish unaffected. Associated middleware imports/usages and route registrations have been updated accordingly.

<details>

• Introduced 'connectionCapping' middleware to replace 'resourceCapping' for POST /connect/sessions.
• Removed 'resourceCapping' middleware file and references from public routes.
• Updated 'routes.public.ts' to use 'connectionCapping' and refactored auth handler arrays.
• Ensured that connect-session completion is not prevented after the connection cap is hit, per new requirements.

</details>

<details>

• Server middleware (packages/server/lib/middleware/connection-capping.middleware.ts)
• Public API routing (packages/server/lib/routes.public.ts)

</details>


packages/server/lib/routes.public.ts
```diff
@@ -66,11 +66,7 @@ const connectSessionAuth: RequestHandler[] = [authMiddleware.connectSessionAuth.
 const connectSessionAuthBody: RequestHandler[] = [authMiddleware.connectSessionAuthBody.bind(authMiddleware), rateLimiterMiddleware];
 const connectSessionOrApiAuth: RequestHandler[] = [authMiddleware.connectSessionOrSecretKeyAuth.bind(authMiddleware), rateLimiterMiddleware];
 
-const connectSessionOrPublicAuth: RequestHandler[] = [
-    authMiddleware.connectSessionOrPublicKeyAuth.bind(authMiddleware),
-    resourceCapping,
-    rateLimiterMiddleware
-];
+const connectSessionOrPublicAuth: RequestHandler[] = [authMiddleware.connectSessionOrPublicKeyAuth.bind(authMiddleware), rateLimiterMiddleware];
```
kaposke, line 69: Would appreciate a double-check, but from what I understand this is used in the different auth endpoints accessed by connect-ui / frontend sdk after a connect session is already created.

TBonnin (approved):

PR #4405, opened 2025-07-30
feat(providers): add example of workday username to use in docs

hassan254-prog:
- Add example of workday username to use in docs

---

This minor PR updates the Workday integration documentation to include a concrete example of the username format (`ISU Nango@acme`), clarifies format guidance, and syncs this example with the provider schema. The providers.yaml file is updated to use more representative example values for both tenant and username fields and now explicitly enforces a username pattern matching `^[^@]+@[^@]+$`.


packages/providers/providers.yaml
```diff
@@ -11328,14 +11328,16 @@ workday:
...
+            example: acme
             order: 2
             doc_section: '#step-2-finding-your-tenant'
     credentials:
         username:
             type: string
+            example: ISU Nango@acme
             title: Username
             description: Workday API Username
+            pattern: '^[^@]+@[^@]+$'
```
hassan254-prog, line 11340: This pattern should match everything that has an '@' in between.

khaliqgant (approved):

PR #4411, opened 2025-07-31
chore: logging performance experiments

TBonnin:
this is an experiment to evaluate the performance cost of using a custom formatter compared to winston default formatter


bodinsamuel:
I'm curious? did you try in prod? no results?

PR #4413, opened 2025-07-31
feat: usage notification emails

kaposke:
It's a simple task with a lot of quirks...

Ideally, this would be a simple reaction to an event, that by itself would be triggered as a chain reaction from a usage event (action, persist records, connection created).

While that's not possible, I'm adding another inlined side-effect to everywhere that increases one of the metrics. And for that to be accessible from persist, I had to move the email client module out of `server`. 

Rough look at the emails:
image

---

This PR introduces a major overhaul of the notification and email infrastructure in the codebase. It establishes a new event-driven system for usage cap notification emails, centralizes all email logic into a standalone, provider-agnostic @nangohq/email package, and refactors metric tracking for extensibility, reliability, and modularity. The update affects multiple packages, rewires how and where side effects (like usage notification emails) are triggered across the stack, introduces locking to mitigate race conditions, and retools Docker and build configs to support the new email architecture. Extensive dependency and type updates, improved usage endpoint consistency, and new testing 'dry run' logic for emails are included. The PR reflects a step towards a modern, event-based design while documenting known quirks and architectural trade-offs.

<details>

• Removal of all legacy/server-centric email logic and introduction of a new @nangohq/email package supporting SMTP, Mailgun, and a no-op logger mode.
• Centralization and standardization of email delivery logic, with HTML templates and pluggable providers.
• Introduction of an event-driven onUsageIncreased function that issues 80%/100% usage notification emails as a side-effect of metric increments across actions, records persistence, and connection creation.
• Major updates to Dockerfile, build pipeline, and package.json/lock to enable modular email package inclusion and correct dependency wiring.
• Implementation of a new withLock() pattern in @nangohq/kvstore for safe side-effect execution and race condition reduction.
• Refactoring and expansion of metric types (AccountUsageMetric vs. AccountUsageIncrementableMetric) and metric usage reporting APIs.
• Extensive rewiring of how account usage and notifications propagate through persist/sync/server/controllers/hooks for consistent event handling.
• Improvements to usage data fetching for metrics endpoints, and clean separation of usage tracking concerns.
• Add dry-run logic for notification emails to simplify safe testing (emails are logged, not sent during dry run).

</details>

<details>

• Email delivery infrastructure (server, persist, entrypoints, Dockerfile, build tooling)
• Account usage increment/decrement logic (actions, records, connections)
• Notification email templates and business logic
• Metric tracking and reporting types throughout account-usage
• KVStore locking & helper utilities
• Server controllers, persist package, hooks, and build/deploy scripts
• Dependency graph: package.json, package-lock, monorepo wiring

</details>


packages/shared/lib/utils/email.ts
```diff
@@ -111,3 +110,68 @@ Team Nango</p>
...
+}
+
+function formatUsage(usage: MetricUsage[]) {
+    const usageLines = usage.map((u) => {
+        const postFix = u.limit && u.usage >= u.limit ? '🚨' : u.limit && u.usage >= u.limit * 0.8 ? '⚠️' : '';
+        return `${u.label}: ${u.usage} / ${u.limit} ${postFix}`;
+    });
+
+    return ['<p>', usageLines.join('<br>'), '</p>'].join('\n');
+}
```
TBonnin: should this file be in the email package. it only depends on `utils` and `types`?

packages/shared/lib/utils/usage.ts
```diff
@@ -0,0 +1,96 @@
...
+        users.map((user) => {
+            // Full limit reached prioritized in case usage jumps directly to it
+            if (crossedLimit) {
+                return sendUsageLimitReachedEmail({ user, account, usage });
+            }
+
+            return sendUsageNearLimitEmail({ user, account, usage });
+        })
+    );
+}
```
TBonnin: I feel like those functions should be in the `account-usage` package, injecting the required dependency if necessary

packages/shared/lib/utils/usage.ts
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    const account = await accountService.getAccountById(db.knex, accountId);
+
+    if (!account) {
+        return;
+    }
+
+    const usage = await getAccountMetricsUsage(account, plan);
+    const users = await userService.getUsersByAccountId(accountId);
```
TBonnin: that's a lot of extra queries to happen every time records are being saved. Can we save the usage and have the check happens later? Does it have to be inlined?

rguldener:
@kaposke For the limits of actions & synced records, can you make sure there is "per month" at the end?

E.g.
Connections: XX / {{connections cap}}
Actions: YY / {{actions cap}} per month
Synced records: XX / {{records cap}} per month

packages/shared/lib/utils/usage.ts
kaposke: Notice it only happens at the point where the limit was crossed:
- Was below threshold and, in this exact increment operation, went above threshold with delta.
So I think it's safe given it's a rare occurrence. 
And it's the same reason why I can't do it later (unless I keep some intermediate state), since I'm trying to capture the exact moment the threshold was crossed.

So only the plan query happens every time. And notice this call was not awaited at any point, so at the very least, it's non-blocking.

Again, ideally those are all independent reactions to events.

packages/shared/lib/utils/usage.ts
kaposke: Yeah that's a fair point. 

Also, I could have added the trigger in `AccountTracker.incrementUsage` instead of everywhere where it's called, but I wanted to keep it separate to maintain SRP and to make the future transition to event-based more intuitive - the way it's now it's more easily comparable to what I wanted as an event reaction.

packages/account-usage/lib/events/onUsageIncreased.ts
```diff
@@ -0,0 +1,53 @@
...
+
+    const limit = accountUsageTracker.getLimit(plan, metric);
+    const currentUsage = await accountUsageTracker.getUsage({ accountId: accountId, metric });
+
+    if (limit === null || currentUsage === null || delta === 0) {
+        return;
+    }
+
+    const crossedLimit = currentUsage - delta < limit && currentUsage >= limit;
+    const crossed80Percent = currentUsage - delta < limit * 0.8 && currentUsage >= limit * 0.8;
```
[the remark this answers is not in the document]
kaposke: I don't think it can lead to duplicates, but it can in race conditions lead to no email being sent. This would be solved by having it actually be sequential in an event-based system, which is why I'm not obsessing about it. Crazy race conditions should be rare in free-plans.

packages/account-usage/lib/events/onUsageIncreased.ts
```diff
@@ -0,0 +1,88 @@
...
+            const crossed80Percent = currentUsage - delta < limit * 0.8 && currentUsage >= limit * 0.8;
+
+            if (!crossedLimit && !crossed80Percent) {
+                return;
+            }
+
+            const account = await accountService.getAccountById(db.knex, accountId);
+
+            if (!account) {
+                return;
```
TBonnin: maybe some logging here just in case

packages/account-usage/lib/events/onUsageIncreased.ts
```diff
@@ -0,0 +1,88 @@
...
+    metric,
+    delta,
+    plan
+}: {
+    accountId: number;
+    metric: AccountUsageMetric;
+    delta: number;
+    plan?: DBPlan | undefined;
+}) {
+    try {
```
TBonnin, line 34: can we had tracing (a single span for the entire function is ok for now) so we can keep an eye on the perf

TBonnin (approved):

PR #4421, opened 2025-08-04
feat: actually send usage emails

kaposke:
After confirming over the weekend that no unexpected spam behavior is happening, removing dryRun to actually send usage emails.

---

This PR deletes the 'dryRun' logic from the account usage notification email functions, resulting in actual sending of emails to users when their usage approaches or exceeds predefined limits. The change follows a monitoring period to confirm that no spam or unwanted email behavior results from enabling live emails.

### DIFF
--- a/packages/account-usage/lib/emails.ts
+++ b/packages/account-usage/lib/emails.ts
@@ -22,12 +22,6 @@ export async function sendUsageNearLimitEmail({
 
     logger.info(`Sending usage near limit email for ${metricName} to ${user.email}`);
 
-    // TODO: Remove after testing
-    const dryRun = true;
-    if (dryRun) {
-        return;
-    }
-
     const formattedUsage = formatUsage(usage, triggeringMetric);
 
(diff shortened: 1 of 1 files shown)


khaliqgant (approved):

packages/account-usage/lib/emails.ts
```diff
@@ -68,13 +62,7 @@ export async function sendUsageLimitReachedEmail({
...
     const metricName = usage.find((u) => u.metric === triggeringMetric)?.label ?? triggeringMetric;
 
-    logger.info(`Sending usage near limit email for ${metricName} to ${user.email}`);
-
-    // TODO: Remove after testing
-    const dryRun = true;
-    if (dryRun) {
-        return;
-    }
+    logger.info(`Sending usage limit reached email for ${metricName} to ${user.email}`);
```
TBonnin, line 65: I prefer not to log email since it is PII. UserId would be enough here imho

PR #4422, opened 2025-08-04
feat(action-usage): [nan-3644] log action size

khaliqgant:
---

This pull request enhances the SyncController to log an informational message when the response payload of a triggered action exceeds 0.5 MB. It adds logic to extract the 'content-length' header from each action's HTTP response and, if the size is above the threshold, logs the account ID, action name, and the response size in megabytes. The update also introduces a dedicated logger and ensures consistent processing and redaction of response headers in operation logs.

### DIFF
--- a/packages/server/lib/controllers/sync.controller.ts
+++ b/packages/server/lib/controllers/sync.controller.ts
@@ -303,7 +303,8 @@ class SyncController {
         } finally {
             const reqHeaders = getHeaders(req.headers);
             reqHeaders['authorization'] = 'REDACTED';
-            const contentLength = reqHeaders['content-length'];
+            const responseHeaders = getHeaders(res.getHeaders());
+            const contentLength = responseHeaders['content-length'];
             const lengthInMB = contentLength ? Number(contentLength) / (1024 * 1024) : 0;
             if (contentLength && lengthInMB > 0.5) {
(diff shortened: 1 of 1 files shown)


packages/server/lib/controllers/sync.controller.ts
```diff
@@ -302,6 +303,12 @@ class SyncController {
         } finally {
             const reqHeaders = getHeaders(req.headers);
             reqHeaders['authorization'] = 'REDACTED';
+            const responseHeaders = getHeaders(res.getHeaders());
+            const contentLength = responseHeaders['content-length'];
+            const lengthInMB = contentLength ? Number(contentLength) / (1024 * 1024) : 0;
+            if (contentLength && lengthInMB > 0.5) {
+                logger.info(`Action DEBUGGING: contentLength is ${lengthInMB.toFixed(2)} MB`);
```
TBonnin: do we want to also log the accountId?

TBonnin (approved):

packages/server/lib/controllers/sync.controller.ts
khaliqgant: Yeah, added with https://github.com/NangoHQ/nango/pull/4422/commits/e856023632711e454668e812929634111501b933

PR #4426, opened 2025-08-04
feat(db): connect_sessions overrides

kaposke:
Review https://github.com/NangoHQ/nango/pull/4427 to understand this.

Adds `overrides` column to `connect_sessions`. Isolated PR with migration.

> Why not use `integration_config_defaults`?

That's clearly meant to set defaults for fields that shouldn't be shown to the end-user, while `overrides` is for overriding settings.

---

This pull request introduces two database schema migrations and associated code changes to support session-specific configuration overrides and a plan-level feature flag. Specifically, it adds an `overrides` (nullable JSONB) column to the `connect_sessions` table for per-session override storage, and a `can_override_docs_connect_url` (boolean, default false) column to the `plans` table to control plans' ability to override the connect URL. The plan type definition and related seeder utility are updated to include this new property, ensuring type safety and test coverage.


packages/database/lib/migrations/202508042254_add_overrides_to_connect_session.cjs
```diff
@@ -0,0 +1,19 @@
+exports.config = { transaction: false };
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.schema.alterTable('connect_sessions', (table) => {
+        table.jsonb('overrides').nullable();
```
TBonnin: can we overrides everything? If not maybe we can find a name that is more descriptive

TBonnin (approved):

TBonnin (approved):

packages/database/lib/migrations/202508042254_add_overrides_to_connect_session.cjs
kaposke: Only docs_connect right now, but who knows.

PR #4428, opened 2025-08-04
feat(proxy): handle minutely rate limit

hassan254-prog:
- Handle minutely rate limit. Exact Online enforces a minutely rate limit. While retries using Nango’s default exponential backoff would eventually succeed, since these limits typically reset at the top of the minute, it's more efficient to retry `at` or `after` the minutely reset. This change first checks for the minutely rate limit and falls back to the default rate-limiting logic if needed.
```
{
  "headers": {
    "date": "Thu, 31 Jul 2025 23:02:07 GMT",
    "content-length": "0",
    "x-ratelimit-minutely-remaining": "0",
    "x-ratelimit-minutely-limit": "60",
    "x-ratelimit-minutely-reset": "1754002980000"
  },
  "code": 429
}

```

---

**Enhanced Proxy Rate Limiting: Multi-Header and Minutely/Daily Strategy Support**

This PR extends Nango's proxy rate-limiting system to support APIs that enforce multiple, concurrent rate-limits (e.g. minutely and daily, as with Exact Online). It generalizes the retry logic so that `retry.at` and `retry.after` fields may now be specified as arrays of headers (or single values for backward compatibility) in provider YAML configurations as well as in runtime proxy options. The proxy will scan all provided rate-limit headers and use the strictest (i.e., longest-required) wait time when deciding how long to wait before a retry. Type definitions, schema validation (JSON Schema), and unit tests have been updated accordingly.

<details>

• Generalized proxy retry logic to accept and process arrays of rate-limit headers for `retry.at` and `retry.after` in provider and runtime configurations
• Proxy now finds all relevant rate-limit headers in the response and uses the one requiring the longest wait time for its retry calculation
• Updated provider ``YAML`` (`packages/providers/providers.yaml`) to uniformly use arrays for retry headers in all relevant providers (backwards compatible)
• Extended provider validation schema (`scripts/validation/providers/schema.json`) to support arrays for `retry.at` and `retry.after` fields
• Revised internal type definitions (``RetryHeaderConfig`` in `api.ts`) to accept both string and array formats
• Refactored and extended unit tests (`retry.unit.test.ts`) to cover array header scenarios, multiple concurrent header cases, and correct wait-time calculations

</details>

<details>

• Proxy rate limit logic (`packages/shared/lib/services/proxy/retry.ts`)
• Provider ``YAML`` definitions (`packages/providers/providers.yaml`)
• Type definitions (`packages/types/lib/proxy/api.ts`)
• Provider schema validation (`scripts/validation/providers/schema.json`)
• Unit tests for proxy retry behavior (`retry.unit.test.ts`)

</details>

---


packages/providers/providers.yaml
```diff
@@ -3686,6 +3686,9 @@ exact-online:
             error_code:
                 - '429'
                 - '503'
+            minutely_rate_limit:
+                reset_time_key: 'x-ratelimit-minutely-reset'
+                strategy: at
```
TBonnin: why not set the `retry.at` value to `x-ratelimit-minutely-reset` above?

hassan254-prog: Exact Online can return either header, depending on which rate limit is exceeded. For the minutely limit, it returns `x-ratelimit-minutely-reset`, while for the daily limit, it returns `x-ratelimit-reset`. We need to handle both.

TBonnin: can we have a more generic approach where `retry.at` would support a list of headers instead of one?. I don't really like having some provider specific attributes and logic

hassan254-prog: I agree, I’ve updated this with the latest commit. This also maintains backward compatibility with syncs already using `retryHeader.at: string`, while introducing support for an array in both `custom` and `preconfigured` headers. I wasn’t sure whether we should also apply the changes to `retry.after`, let me know what you think.

TBonnin: retry.at and retry.after should probably have the same behavior. wdyt?

packages/shared/lib/services/proxy/retry.ts
```diff
@@ -68,22 +68,38 @@ export function getProxyRetryFromErr({ err, proxyConfig }: { err: unknown; proxy
...
+            retryHeaders = [proxyConfig.retryHeader.after!];
+        }
+
+        for (const retryHeader of retryHeaders) {
+            const res = getRetryFromHeader({ err, type, retryHeader });
+            if (res.found) {
+                return { retry: true, reason: `custom_${res.reason}`, wait: res.wait };
+            }
         }
     }
```
TBonnin, line 89: what if the response has the multiple specified headers with different values? Should we take into account the one with the most restrictive retry time instead of the first one found?

packages/shared/lib/services/proxy/retry.ts
```diff
@@ -174,7 +190,7 @@ function parseRetryValue({
 }: {
     type: 'at' | 'after';
     rawValue: string | undefined;
-    source: 'header' | 'body';
+    source: 'header' | 'body' | 'minutely_rate_limit';
```
TBonnin: is this necessary?

packages/providers/providers.yaml
hassan254-prog: I think that would be great, having a uniform behavior for both, as they work almost similarly. I’ve updated this with the latest commit.

packages/shared/lib/services/proxy/retry.ts
hassan254-prog: thanks, that’s actually a better approach and also reduces the chances of failure if not configured well. I’ve updated it with the latest [commit](https://github.com/NangoHQ/nango/pull/4428/commits/1a0a3865f752158543792b25eccd8f568bbb0a25) 🙏.

packages/shared/lib/services/proxy/retry.ts
```diff
@@ -68,22 +68,58 @@ export function getProxyRetryFromErr({ err, proxyConfig }: { err: unknown; proxy
...
+        let retryHeaders: string[];
 
-        const res = getRetryFromHeader({ err, type, retryHeader: retryHeader! });
-        if (res.found) {
-            return { retry: true, reason: `custom_${res.reason}`, wait: res.wait };
+        if (proxyConfig.retryHeader.at) {
+            retryHeaders = Array.isArray(proxyConfig.retryHeader.at) ? proxyConfig.retryHeader.at : [proxyConfig.retryHeader.at];
+        } else {
+            retryHeaders = Array.isArray(proxyConfig.retryHeader.after) ? proxyConfig.retryHeader.after : [proxyConfig.retryHeader.after!];
+        }
```
TBonnin: can this be simplified like this:
```
const retrySource = proxyConfig.provider.proxy.retry.at || proxyConfig.provider.proxy.retry.after || [];
const retryHeaders = Array.isArray(retrySource) ? retrySource : [retrySource];
```

packages/shared/lib/services/proxy/retry.ts
```diff
@@ -68,22 +68,58 @@ export function getProxyRetryFromErr({ err, proxyConfig }: { err: unknown; proxy
...
+
+        // Check all headers and find the most restrictive retry time
+        let bestRetry: { reason: string; wait: number } | null = null;
```
TBonnin, line 75: I think this is the other way around. We should pick the latest retry (highest wait). If one header says wait 20s, and the other one wait 60s, we should be conservative and pick the 60s. wdyt?

packages/shared/lib/services/proxy/retry.ts
```diff
@@ -68,22 +68,58 @@ export function getProxyRetryFromErr({ err, proxyConfig }: { err: unknown; proxy
...
+            const res = getRetryFromHeader({ err, type, retryHeader });
+            if (res.found) {
+                if (!bestRetry || res.wait < bestRetry.wait) {
+                    bestRetry = { reason: res.reason, wait: res.wait };
+                }
+            }
+        }
+
+        if (bestRetry) {
+            return { retry: true, reason: `preconfigured_${bestRetry.reason}`, wait: bestRetry.wait };
```
TBonnin, line 110: same comment as above

packages/shared/lib/services/proxy/retry.ts
hassan254-prog: That could cause issues. I was looking at Exact Online daily rate limit reset, it resets at midnight, while the minutely one resets top of the minute, if we have both it may cause us to process rate limiting slower than we should.

TBonnin: I am not sure of the point of having multiple rate limits if the one that matter is the shortest one.
if one says wait 20s and the other one wait 60s, why does the second one exists then if you can retry after 20s

hassan254-prog: Exact Online will return either the daily or minutely rate limit headers, each with different keys, depending on which limit is reached. This is why we maintain an array of headers to check. If, under any circumstance, both headers are returned simultaneously, we will use the shorter one, the minutely limit.

packages/shared/lib/services/proxy/retry.ts
```diff
@@ -68,22 +68,46 @@ export function getProxyRetryFromErr({ err, proxyConfig }: { err: unknown; proxy
...
+        const retryHeaders = Array.isArray(retrySource) ? retrySource : [retrySource];
+
+        // Check all headers and find the most restrictive retry time
+        let bestRetry: { reason: string; wait: number } | null = null;
+
+        for (const retryHeader of retryHeaders) {
+            const res = getRetryFromHeader({ err, type, retryHeader });
+            if (res.found) {
+                if (!bestRetry || res.wait < bestRetry.wait) {
+                    bestRetry = { reason: res.reason, wait: res.wait };
```
bodinsamuel: This doesn't work?
```suggestion
                    bestRetry = res
```

packages/shared/lib/services/proxy/retry.ts
```diff
@@ -68,22 +68,46 @@ export function getProxyRetryFromErr({ err, proxyConfig }: { err: unknown; proxy
...
```
bodinsamuel: same

packages/shared/lib/services/proxy/retry.ts
bodinsamuel: I agree with thomas, it's not possible to have two rate limits reset and not use the biggest one; because if the system says that our daily limit is reached, then the minutely limit doesn't matter (unless we are missing something)
If they always send those headers (even if daily limit hasn't been reached), and then we need to check which one is actually relevant with a second header

hassan254-prog: @bodinsamuel , @TBonnin ; I have updated the PR to check for the longest wait time. Here are 2 different examples showing the response headers for different status codes. The check between the longest is a precaution incase we get rate limited and get both headers. Do we leave it at that and hope they will only return a single header at a time, i.e for the different rate limits or improve the logic further and also check for remaining requests on both of them. 
I was thinking of this config.

```
retry:
    at:
        - 'x-ratelimit-minutely-reset': 'x-ratelimit-minutely-remaining'
        - 'x-ratelimit-reset': 'x-ratelimit-remaining'
```

Response headers:

```
"response": {
    "headers": {
      "date": "Fri, 01 Aug 2025 04:59:16 GMT",
      "content-length": "0",
      "x-ratelimit-minutely-remaining": "0",
      "x-ratelimit-minutely-limit": "60",
      "x-ratelimit-minutely-reset": "1754024400000"
    },
    "code": 429
  },
```
```
"response": {
    "headers": {
      "date": "Thu, 31 Jul 2025 23:05:57 GMT",
      "content-length": "162",
      "expires": "-1",
      "x-exact-correlation-id": "4f80cb65-6d36-4654-8917-ce5d54e1cf76",
      "x-ratelimit-limit": "5000",
      "x-ratelimit-minutely-limit": "60",
      "x-ratelimit-minutely-reset": "1754003160000",
      "x-ratelimit-reset": "1754006400000",
      "x-ratelimit-remaining": "4568",
      "pragma": "no-cache",
      "x-ratelimit-minutely-remaining": "46",
      "x-requesthandledby-eoldotnetapp": "0",
      "content-type": "application/json",
      "cache-control": "no-cache, no-store",
      "dataserviceversion": "1.0;"
    },
    "code": 500
  },
```

TBonnin: maybe I have no idea how our retry logic works but with the second example (the 500 one) why would the ratelimit reset be accounted for since it is not rate-limited?

TBonnin (approved):

packages/shared/lib/services/proxy/retry.ts
TBonnin: Discussed with @hassan254-prog 
Retry logic is already questionable since we wait until `reset` even if not a 429 but another retryable code.
PR as it is handles the exact online scenario but we could make the logic even more robust when more than 1 set of rate limit headers are returned by looking at the one with `remaining=0` instead of taking the max date.
We went for not complicating even more the retry logic for now since this is not a scenario we need to handle (exact online when 429 only returns one set of headers)

bodinsamuel (approved):

PR #4431, opened 2025-08-05
docs: add step for enabling 2GP in Salesforce

Tom-DK:
This step is important otherwise the following error will returned in step 9: 

`$ sf package list --target-dev-hub my-dev-hub  `

```
PackageErrorUsername,AppAnalyticsEnabled from Package2 ORDER BY NamespacePrefix
                                              ^
ERROR at Row:1:Column:261
sObject type 'Package2' is not supported
```

---

This PR updates the Salesforce integration documentation by adding an explicit instruction for enabling Second-Generation Packaging (2GP) during Developer Hub setup. The documentation change highlights that omitting this step can cause an error ('sObject type 'Package2' is not supported') during package listing, which is a common point of failure for users setting up external packaged integrations.


hassan254-prog:
Thanks for the contribution 🙏 .

hassan254-prog (approved):

PR #4432, opened 2025-08-05
feat: mar for usage should be aligned with billing

kaposke:
This PR refactors the logic controlling when Monthly Active Records (MAR) usage is tracked and billed. Now, the MAR tracking for usage capping is consistently conditioned on the same flag used for billing, ensuring both mechanisms run in parallel under the same circumstances.

### DIFF
--- a/packages/persist/lib/records.ts
+++ b/packages/persist/lib/records.ts
@@ -170,15 +170,16 @@ export async function persistRecords({
         const mar = new Set(summary.activatedKeys).size;
 
         // Billing metrics
-        if (plan && shouldBillMonthlyActiveRecords) {
-            billing.add('monthly_active_records', mar, { accountId, environmentId, providerConfigKey, connectionId, syncId, model });
-            metrics.increment(metrics.Types.BILLED_RECORDS_COUNT, mar, { accountId });
+        if (shouldBillMonthlyActiveRecords) {
+            if (plan) {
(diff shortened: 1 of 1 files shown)


packages/persist/lib/records.ts
```diff
@@ -170,16 +170,17 @@ export async function persistRecords({
...
-            metrics.increment(metrics.Types.BILLED_RECORDS_COUNT, mar, { accountId });
+        if (shouldBillMonthlyActiveRecords) {
+            if (plan) {
+                billing.add('monthly_active_records', mar, { accountId, environmentId, providerConfigKey, connectionId, syncId, model });
+                metrics.increment(metrics.Types.BILLED_RECORDS_COUNT, mar, { accountId });
+            }
+            const accountUsageTracker = await getAccountUsageTracker();
+            // Account usage tracking for capping
+            void accountUsageTracker.incrementUsage({ accountId, metric: 'active_records', delta: mar });
+            void onUsageIncreased({ accountId, metric: 'active_records', delta: mar, plan: plan ?? undefined });
```
TBonnin, line 181: Hasn't everybody a plan nowadays?

TBonnin (approved):
I guess the correct usage metrics will only show next month. right?

packages/persist/lib/records.ts
kaposke: I think so, but there's still a flag and the types being nullable. So not 100% sure

PR #4436, opened 2025-08-06
chore: add env var for server public routes body limit

TBonnin:
self-hosted customer might need to tweak the size limit of the public API


khaliqgant (approved):
Does this also need to go in the helm chart? Or at least be documented?

TBonnin:
> Does this also need to go in the helm chart? Or at least be documented?

Eventually yes but I would like to confirm there isn't another issue before we make it fully public

PR #4439, opened 2025-08-06
fix: activemq should be optional if not cloud/local

TBonnin:
Adding an env var to fully disable activemq if not cloud or local

@NangoHQ/engineers you will need to add this env var to your local setup if you want to run/test metering


kaposke (approved):
Let's add it to .env.example?

TBonnin:
> Let's add it to .env.example?

good point

PR #4441, opened 2025-08-06
fix(cron.exportUsage): prevent concurrent executions

TBonnin:
Jitter is not the solution to deal with multiple instances of server exporting usage at the same time.
Using a proper lock to prevent concurrent usage export

---

**Prevent Concurrent Executions of Usage Export Cron Job**

This PR updates the `exportUsage` cron job in the server code to prevent concurrent executions across multiple server instances. The former approach of adding random jitter to the cron schedule has been replaced with a distributed lock mechanism, using the locking interface from `@nangohq/kvstore`. The lock ensures only one instance can perform the usage export at any given scheduled run, skipping execution if the lock cannot be acquired. The schedule has been restored to a fixed interval, and lock acquisition/release is managed with proper safety to guarantee the lock is never leaked.

<details>

• Removed jitter-based scheduling and restored a fixed cron interval based on ``cronMinutes``.
• Integrated distributed locking via ``getLocking`` from `@nangohq/kvstore` for the cron job.
• The `exec` function now attempts to acquire a lock before running export logic and skips execution if the lock is already held.
• Added structured try/finally logic to guarantee lock release after execution.
• Refactored relevant imports to include lock types and the lock utility.

</details>

<details>

• packages/server/lib/crons/usage.ts

</details>

---

### DIFF
--- a/packages/server/lib/crons/usage.ts
+++ b/packages/server/lib/crons/usage.ts
@@ -2,12 +2,14 @@ import tracer from 'dd-trace';
 import * as cron from 'node-cron';
 
 import { billing as usageBilling } from '@nangohq/billing';
+import { getLocking } from '@nangohq/kvstore';
 import { records } from '@nangohq/records';
 import { connectionService } from '@nangohq/shared';
 import { flagHasUsage, getLogger, metrics, report } from '@nangohq/utils';
 
 import { envs } from '../env.js';
 
+import type { Lock } from '@nangohq/kvstore';
 import type { BillingMetric } from '@nangohq/types';
 
 const logger = getLogger('cron.exportUsage');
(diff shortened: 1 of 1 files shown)


kaposke (approved):

packages/server/lib/crons/usage.ts
```diff
@@ -31,11 +31,30 @@ export function exportUsageCron(): void {
...
-        await observability.exportRecordsMetrics();
-        logger.info(`✅ done`);
+
+        const locking = await getLocking();
+        const ttlMs = cronMinutes * 60 * 1000;
+        let lock: Lock | undefined;
+        const lockKey = `lock:cron:exportUsage`;
+
+        try {
+            lock = await locking.acquire(lockKey, ttlMs);
```
kaposke, line 41: Should it try for so long?

TBonnin: ttlMs is the ttl of the lock, not how long it try to acquire it.

PR #4469, opened 2025-08-11
feat(webhooks): add feature flags, revamp method signature

bodinsamuel:
## Changes

- Control execution with feature flags
Forward and scripts execution separately

- Revamp method signature
I needed to check the plan after getting the connectionIDs, but the current "nango" object was not helpful (stateless) and we were passing data around for no good reason. Simplify this with a class. We could go further and pass the body and headers too (like a regular nango script)

- Fixed webhookSubscriptions = '*' 
Maybe I'm wrong, but it wasn't working 

---

**Overhaul Webhook Infrastructure: Feature Flags, Plan Gating, and Standardized Handler Signatures**

This PR introduces a significant refactor of the webhook orchestration system. Major changes include the introduction of feature flags and plan-based gating for webhook forwarding and script execution, replacement of the stateless internalNango object with a stateful InternalNango class encompassing team, environment, plan, and integration context, and a full standardization of webhook handler function signatures across all integrations. All webhook routing logic, controllers, integration handlers, and tests are updated to work with the new contracts, and support for wildcard ('*') webhook subscriptions is made fully consistent and explicit. The goal is to provide explicit, plan-aware, and type-safe propagation of integration and environment context, and to allow for future extensibility in how webhooks are routed, executed, or skipped based on customer plan, feature flags, or integration support.

<details>

• Introduced plan-aware feature flags (`has_webhooks_script`, `has_webhooks_forward`) to control webhook script execution and forwarding independently, including support for self-hosted and multi-plan environments.
• Replaced stateless `internalNango` object/interface with a new `InternalNango` class that tracks team, environment, plan, integration, and logger context for every webhook request.
• Standardized all webhook handler method signatures to (`InternalNango`, headers, body, `rawBody`), deprecating prior positional arguments and eliminating legacy coupling to `integration` and logger objects.
• Refactored all webhook routing/controller logic (`routeWebhook`, `postWebhook`) to construct `InternalNango` with full context and pass it through the stack, including plan information for feature gating.
• Updated every webhook handler file (Hubspot, Github, Jira, Gmail, Sentry, Slack, Linear, Checkr, Xero, Attio, Salesforce, Airtable, Calendly, Microsoft Teams, unauthenticated, etc.) to the new function contract.
• Fixed and normalized support for wildcard `*` webhook subscriptions to ensure consistent matching and execution, as documented.
• Improved type definitions (types.ts, integration/db.ts) and added `DBIntegrationDecrypted` type for statically safe context propagation.
• Added and updated unit tests to cover proper handler invocation, execution ordering, catch-all matching, and complex webhook payload cases.

</details>

<details>

• Webhook manager/controller (webhook.manager.ts, `postWebhook`.ts)
• All integration-specific webhook routing/handler files
• Webhook internal orchestration library (internal-nango.ts)
• Type definition files for integration, environment, plan, context
• All webhook-related tests and mocking infrastructure

</details>

---


packages/server/lib/webhook/internal-nango.ts
```diff
@@ -87,23 +96,28 @@ export const internalNango: InternalNango = {
             }
 
             for (const webhook of webhook_subscriptions) {
-                if (type === webhook) {
+                if (type === webhook || webhook === '*') {
```
bodinsamuel, line 105: I'm not sure how it was supposed to work before; maybe I'm missing something? @khaliqgant

khaliqgant: The type for airtable would be `'*'` and the webhook would also be `'*'` so this additional condition isn't needed I don't think.

What were you seeing that prompted you to add this?

bodinsamuel: because I thought it was a feature, and it's been documented like this
https://docs.nango.dev/reference/scripts#param-webhook-subscriptions

I guess I found out handy to handle all webhooks at once, wdyt? otherwise I can remove, I'll need to change the doc 👌🏻

khaliqgant: Gotcha. Yeah, remembering discussing with Bastien a while back that the documentation was incorrect – guess it was never fixed.

This change would make the documentation correct and I agree would be a nice addition. Just would want to make sure it wouldn't break Airtable but I don't see how it couldin this case.

packages/server/lib/controllers/webhook/environmentUuid/postWebhook.ts
```diff
@@ -60,6 +61,19 @@ export const postWebhook = asyncWrapper<PostPublicWebhook>(async (req, res) => {
                 return;
             }
 
+            const resPlan = await getPlan(db.knex, { accountId: account.id });
```
TBonnin: does this need to be behind `flagHasPlan` for self-hosted instance?

TBonnin (commented):
This new logic must support self-hosted customers with `flagHasPlan` not being defined. Webhooks must work for them

packages/server/lib/controllers/webhook/environmentUuid/postWebhook.ts
bodinsamuel: oh yes good thinking!

packages/server/lib/webhook/internal-nango.ts
```diff
@@ -87,23 +102,28 @@ export const internalNango: InternalNango = {
...
                             input: body,
-                            logContextGetter
+                            logContextGetter: this.logContextGetter
                         });
                     }
+
+                    if (webhook === '*') {
+                        // Only trigger once since it will match all webhooks
+                        break;
+                    }
```
kaposke, line 122: It can technically do it twice if: `['webhookA', 'webhookB', '*']`
Not sure if we care though.

packages/server/lib/webhook/internal-nango.ts
```diff
@@ -87,23 +102,28 @@ export const internalNango: InternalNango = {
...
                     for (const connection of connections) {
                         await orchestrator.triggerWebhook({
-                            account,
-                            environment,
-                            integration,
+                            account: this.team,
+                            environment: this.environment,
+                            integration: this.integration as Config,
                             connection,
                             webhookName: webhook,
```
kaposke, line 112: Assuming the `triggerWebhook` knows how to deal with `*`?

packages/types/lib/integration/db.ts
```diff
@@ -20,6 +20,7 @@ export interface IntegrationConfig extends TimestampsAndDeleted {
     shared_credentials_id: number | null;
 }
 
+export type DBIntegrationDecrypted = IntegrationConfig; // TODO: tag this
```
kaposke, line 23: Does tagging it now break anything?

kaposke (approved):

packages/types/lib/integration/db.ts
bodinsamuel: yes the problem is that IntegrationConfig is used for both encrypted and decrypted (which is why we need to introduce both types). Making them mutually exclusive is breaking stuff

packages/server/lib/webhook/internal-nango.ts
bodinsamuel: Indeed, it's very unlikely yes

packages/server/lib/webhook/internal-nango.ts
bodinsamuel: yeah webhookName is not really used, it's the parentSyncName that matters

PR #4470, opened 2025-08-11
fix(integrations): loosen restrictions on jumpcloud api key pattern

gonzalde97:
---

This PR revises the validation pattern for the JumpCloud API key in the providers.yaml configuration file. The previous regex forced the key to match the format '^jca_[A-Za-z0-9]{32}$', but real-world keys vary more than previously expected. The new pattern now permits any combination of alphanumeric characters, underscores, and hyphens, providing broader compatibility with various JumpCloud API key formats.


packages/providers/providers.yaml
```diff
@@ -5988,7 +5988,7 @@ jumpcloud:
             title: API Key
             description: Your JumpCloud API Key.
             example: jca_9mK4pL2xR8vN6qW1sT7uY3eA5bC9dF0hG2jM
-            pattern: '^jca_[A-Za-z0-9]{32}$'
+            pattern: '^[A-Za-z0-9_-]+$'
```
bodinsamuel, line 5994: can you share what motivates this change?

gonzalde97: the 32 length was true for the test key but is actually has more variability than expected

bodinsamuel (approved):

PR #4471, opened 2025-08-11
fix: move usage tracking to metering

TBonnin:
Incrementing/checking account usage metric is not inlined anymore.
instead same events used for biliing are being dispatched and processed by metering

Note: server/persist will be deployed before metering, which will cause the lost of some events. Since usage metrics are only used to show in the dashboard and send an email when reaching the caps, @kaposke and I have agreed that loosing a few events is ok and putting in place a more complicated rollout strategy was not worth the extra effort

---

**Centralize Usage Tracking via Metering Events**

This PR refactors account usage tracking (covering monthly active records, actions, and connections) to remove inlined/in-service increment and cap-check logic from the persist and server packages. Instead, usage tracking is now achieved by dispatching standard `usage.*` events (previously mainly for billing) through pubsub, which the metering service processes centrally. The metering service now contains the unified usage metrics logic, updating usage statistics and performing cap checks in response to these events. Related code paths in persist and server have been simplified, with unused plan/usage-tracking fields and direct logic eliminated. Dependency references have been tidied to reflect these changes.

<details>

• Replaced inline account usage tracking logic in server and persist with event dispatch to metering via pubsub.
• Metering now handles all usage increment, cap check, and associated notifications based on standardized `usage.*` events.
• Added handling and publishing for a new `usage.connections` event type.
• Removed redundant plan/usage-tracking code and parameters from persist and corresponding ``API`` routes.
• Introduced a ``trackUsage`` helper in metering for usage/cap operations.
• Extended `UsageEvent` pubsub type to include `usage.connections`.
• Updated and cleaned up tsconfig and package.json references for proper dependency alignment.
• Improved error reporting (to Sentry/monitoring) in metering event processing as per code review feedback.

</details>

<details>

• Metering service (billing event processor, dependencies, tsconfig)
• Persist service (records and routes for data mutation)
• Server service (action trigger logic and connection creation hooks)
• Pubsub/event system (event types, dispatch)
• Supporting package manifests and types

</details>

---

### DIFF
--- a/packages/metering/lib/processors/billing.ts
+++ b/packages/metering/lib/processors/billing.ts
@@ -3,7 +3,7 @@ import { billing } from '@nangohq/billing';
 import db from '@nangohq/database';
 import { Subscriber } from '@nangohq/pubsub';
 import { connectionService, getPlan } from '@nangohq/shared';
-import { Err, Ok, metrics, stringifyError } from '@nangohq/utils';
+import { Err, Ok, metrics, report, stringifyError } from '@nangohq/utils';
 
 import { logger } from '../utils.js';
 
@@ -30,7 +30,7 @@ export class Billing {
                 logger.info(`Processing billing event`, { event });
                 const result = await process(event);
                 if (result.isErr()) {
-                    logger.error(`Failed to process billing event: ${result.error}`, { event });
+                    report(new Error(`Failed to process billing event: ${result.error}`), { event });
                     return;
                 }
             }
@@ -52,7 +52,7 @@ async function process(event: UsageEvent): Promise<Result<void>> {
                 }
                 const mar = event.payload.value;
                 metrics.increment(metrics.Types.BILLED_RECORDS_COUNT, mar, { accountId: event.payload.properties.accountId });
-                void trackUsage({
+                await trackUsage({
                     accountId: event.payload.properties.accountId,
                     metric: 'active_records',
                     delta: mar
@@ -65,7 +65,7 @@ async function process(event: UsageEvent): Promise<Result<void>> {
                 });
             }
             case 'usage.actions': {
-                void trackUsage({
+                await trackUsage({
                     accountId: event.payload.properties.accountId,
                     metric: 'actions',
                     delta: event.payload.value
@@ -77,7 +77,7 @@ async function process(event: UsageEvent): Promise<Result<void>> {
                 });
             }
             case 'usage.connections': {
-                void trackUsage({
+                await trackUsage({
                     accountId: event.payload.properties.accountId,
                     metric: 'connections',
                     delta: event.payload.value
@@ -97,7 +97,7 @@ async function trackUsage({ accountId, metric, delta }: { accountId: number; met
     try {
         if (metric !== 'connections') {
             const accountUsageTracker = await getAccountUsageTracker();
-            void accountUsageTracker.incrementUsage({
+            await accountUsageTracker.incrementUsage({
                 accountId,
                 metric,
                 delta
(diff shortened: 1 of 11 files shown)


packages/persist/lib/records.ts
```diff
@@ -61,14 +57,6 @@ export async function persistRecords({
 
     const logCtx = logContextGetter.getStateLess({ id: String(activityLogId), accountId });
 
-    const connection = await connectionService.getConnectionById(connectionId);
-    if (!connection) {
-        const err = new Error(`Connection ${connectionId} not found`);
-        void logCtx.error('Connection not found', { error: err, persistType });
-        span.setTag('error', err).finish();
-        return Err(err);
-    }
```
bodinsamuel: imo this check is useful

packages/metering/package.json
```diff
@@ -14,6 +14,8 @@
         "directory": "packages/metering"
     },
     "dependencies": {
+        "@nangohq/account-usage": "file:../account-usage",
+        "@nangohq/database": "file:../database",
```
bodinsamuel, line 18: missing tsconfig update

packages/metering/lib/processors/billing.ts
```diff
@@ -49,23 +52,69 @@ async function process(event: UsageEvent): Promise<Result<void>> {
...
+                void trackUsage({
+                    accountId: event.payload.properties.accountId,
+                    metric: 'actions',
+                    delta: event.payload.value
+                });
                 return billing.add('billable_actions', event.payload.value, {
                     idempotencyKey: event.idempotencyKey,
                     timestamp: event.createdAt,
                     ...event.payload.properties
                 });
```
bodinsamuel, line 77: We now have a queue that allows us to not lose events, so why don't we await all trackers?

packages/metering/lib/processors/billing.ts
```diff
@@ -49,23 +52,69 @@ async function process(event: UsageEvent): Promise<Result<void>> {
...
+            throw new Error(`Failed to get plan for account ${accountId}: ${plan.error.message}}`);
+        }
+        void onUsageIncreased({
+            accountId,
+            metric,
+            delta,
+            plan: plan.value
+        });
+    } catch (err) {
+        logger.error('Failed to track usage', { error: stringifyError(err) });
```
bodinsamuel: this error is lost

packages/metering/lib/processors/billing.ts
```diff
@@ -49,23 +52,69 @@ async function process(event: UsageEvent): Promise<Result<void>> {
...
+                    delta: event.payload.value
+                });
+                // No billing action for connections, just tracking usage
+                return Ok(undefined);
+            }
             default:
                 return Err(`Unknown billing event type: ${event.type}`);
         }
     } catch (err) {
         return Err(`Error processing billing event: ${stringifyError(err)}`);
```
bodinsamuel, line 92: nothing is reported to sentry

packages/persist/lib/records.ts
TBonnin: it is a extra query to fetch data that is not needed? why would it be useful to check the connection exist?

packages/metering/lib/processors/billing.ts
TBonnin: good point

packages/persist/lib/records.ts
bodinsamuel: well, because if it's not then we are storing records for an unknown/deleted connection and they might not get cleaned up

TBonnin: I was quite happy to remove one query to the db for each batchSave. It is true that in theory a connection could have been deleted between the time the script is executed and batch saved called and this code to run. Ok, I'll keep it

bodinsamuel (approved):

kaposke (approved):

PR #4478, opened 2025-08-12
fix(link-pagination): validate both possible pagination sources before returning

hassan254-prog:
- Validate both possible link pagination sources (header and body) before returning. This applies to Confluence v1, which is aliased to [Jira](https://github.com/NangoHQ/nango/blob/01d79e170c29811654a78152a398a8053b8de447/packages/providers/providers.yaml#L5764). In [v1](https://developer.atlassian.com/cloud/confluence/rest/v1/api-group-search/#api-wiki-rest-api-search-get), the link is included only in the body response and not in the header, which was causing an early return and breaking pagination.

---

**Improve Link Pagination: Sequential Validation of Header and Body Sources**

This PR refactors the pagination logic for APIs using link-based pagination by adjusting how the next page link is extracted. The updated logic ensures that both the response header (for standard REST link headers) and the response body (for APIs like Confluence v1, aliased to Jira, which return pagination links only in the body) are validated in sequence before determining if pagination should continue. The code now throws a descriptive error if neither of the pagination sources is configured, and the unit test suite is expanded to cover combinations of header- and body-based pagination links, preventing premature termination.

<details>

• Refactored `getNextPageLinkFromBodyOrHeaders` to first check the header, then fall back to body if the link is not in the header.
• Added an explicit early validation error if neither `link_rel_in_response_header` nor `link_path_in_response_body` is provided.
• Expanded and revised link pagination unit tests to cover header-only, body-only, missing/empty header cases, with improved scenario coverage.
• Updated type imports to include `LinkPagination` in the test suite.

</details>

<details>

• packages/runner-sdk/lib/paginate.service.ts
• packages/runner-sdk/lib/paginate.service.unit.test.ts

</details>

---


packages/runner-sdk/lib/paginate.service.ts
```diff
@@ -213,14 +213,21 @@ class PaginationService {
...
-            return get(response.data, linkPagination.link_path_in_response_body);
+            nextPageLink = linkHeader?.[linkPagination.link_rel_in_response_header]?.url;
         }
 
-        throw Error(`Either 'link_rel_in_response_header' or 'link_path_in_response_body' should be specified for '${paginationConfig.type}' pagination`);
+        if (!nextPageLink && linkPagination.link_path_in_response_body) {
+            nextPageLink = get(response.data, linkPagination.link_path_in_response_body);
+        }
+
+        if (!linkPagination.link_rel_in_response_header && !linkPagination.link_path_in_response_body) {
```
TBonnin: can the if be simplified to `if(!nextPageLink)`

TBonnin (approved):

packages/runner-sdk/lib/paginate.service.ts
hassan254-prog: That would cause issues, we need the function to either a `string` (for the next page link) or `undefined` (to indicate the end of pagination) so the link `paginator` can gracefully determine when to stop instead of throwing an error.

hassan254-prog: I think that might cause issues. The function should return either a `string` (for the next page link) or `undefined` (to indicate the end of pagination) so the link `paginator` can gracefully determine when to stop.

bodinsamuel (approved):

packages/runner-sdk/lib/paginate.service.ts
TBonnin: let's put the this early return first then

PR #4487, opened 2025-08-13
feat: getting started table migration

kaposke:
Creating 2 new tables for the new Getting Started flow:
- `getting_started_meta` (1 per account) used to store the environment and integration that the account will use for the getting started flow. Necessary because each account needs one integration, and every user going through the getting started will go through the same one.
- `getting_started_progress` (1 per user) used to store user specific progress on the getting started flow, and the connection he created within the getting started.

> I am aware of `onboarding_progress`, and it is pretty similar to what I needed, but I decided to avoid fixing up existing things and going from scratch. Will come back to delete `onboarding` left-overs later.

---

**Add Two New Tables and Types for 'Getting Started' Onboarding Flow**

This PR establishes the foundational database structures and corresponding TypeScript types to support a new 'Getting Started' onboarding flow distinct from the existing onboarding logic. It introduces two new tables: `getting_started_meta` (to record the integration and environment an account uses for onboarding) and `getting_started_progress` (to track each user's progress and created connections during the flow). These changes set the stage for a more structured and account-customized onboarding experience, ensuring clear separation of global (per-account) and user-specific data.

<details>

• Created migration for ``getting_started_meta`` table: stores one record per account with references to environment and integration; enforces uniqueness on ``account_id``.
• Created migration for ``getting_started_progress`` table: stores individual user progress with references to users, the account meta record, and optionally a connection; ensures one progress row per user; includes step tracking.
• Established foreign key constraints with ``ON`` ``DELETE`` ``CASCADE`` and ``SET`` ``NULL`` semantics for relevant relationships.
• Added corresponding `TypeScript` interfaces (`DBGettingStartedMeta`, `DBGettingStartedProgress`) to represent these tables in code.
• Removed the obsolete `complete` property from onboarding progress type.
• Migration and type definitions are isolated to avoid disturbing legacy onboarding structures (per author intent).

</details>

<details>

• Database schema: new migration files for onboarding tables
• `TypeScript` type definitions in `packages/types/lib/`gettingStarted`/db.ts`

</details>

---


packages/types/lib/gettingStarted/db.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * One per user. Holds progress for each user in the getting started flow.
+ */
+export interface DBGettingStartedProgress extends Timestamps {
+    id: number;
+    getting_started_meta_id: number;
+    user_id: number;
+    connection_id: number | null;
+    step: number;
+}
```
bodinsamuel, line 22: Since I'm lacking some context, I'm not sure I get why it's two tables? 
There will be an integration picker, and you can start the getting started multiple times with multiple integrations?

I feel like you can have as much flexibility by just adding all the fields the second table , wdyt?

kaposke: Second PR adds more context to this #4491. 

But no, there won't be any integration picker. We will always create an integration for them to go through the getting started flow. The integration will always be a google-calendar integration, with the unique key `google-calendar-getting-started` (and we support dealing with a conflict in case they create that themselves), and using pre-provisioned credentials.

The meta table is used to store a reference to that integration which we will use for the getting started of every user in that account. The progress table stores individual user progress in the getting started flow, including the  reference to the connection they've created.

It's all very similar to [try.nango.dev](https://try.nango.dev), except the integration and connections are all in their account. They can navigate to the integration and connection pages and see everything, as well as all the logs from the connection being created to the proxy request they'll be able to execute.
Here's a spoiler of the frontend:
image

TBonnin: I still don't follow why we need 2 tables.

kaposke: I guess I should ask how would you do it with one? Not saying it's not doable, but as far as I can see it's cleaner with 2.
You need individual user's progress, and you need a global notion of which integration the whole account uses for the flow (and in which environment that integration lives).

TBonnin: ok. got it. it is 1 meta entry per account (shared by all users). and 1 progress entry per users.

packages/database/lib/migrations/20250815160900_create_getting_started_meta_table.cjs
```diff
@@ -0,0 +1,23 @@
+exports.config = { transaction: false };
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.schema.createTable('getting_started_meta', (table) => {
+        table.increments('id').primary();
```
TBonnin, line 8: it is recommended to use Identity column in recent pg. https://wiki.postgresql.org/wiki/Don%27t_Do_This#Don.27t_use_serial

TBonnin (approved):
one comment about the type of `id` column.

PR #4491, opened 2025-08-14
feat(server): new getting started backend

kaposke:
> PR for Migrations: #4487 

> Long PR but a lot of it is long tests

This adds the necessary endpoints and behavior for the new getting started flow.

### The final desired flow of the getting started page
User can authenticate with Google Calendar via ConnectUI and make a proxy request with it. The integration used should belong to the account and be created automatically by default.

### What should happen behind the schemes
We create an integration with the unique key `google-calendar-getting-started`, using the `google-calendar` pre-provisioned provider (new shared credentials feature). This integration is tied to the user's account. Connections from the demo will be created in this integration. The user can navigate and see both the integration and created connections normally.

As such, we have to handle both the integration and the connection being deleted by the user. The flow needs to be ready for the scenario where the integration ceases to exist.

### My Approach
It basically all revolves around `getOrCreateX` happening to the whole dependency tree:
```
getOrCreate `getting_started_progress`
└── getOrCreate `getting_started_meta`
    └── getOrCreate `google-calendar-getting-started` integration
``` 

- User deleted the connection? That's low impact, it means we detach the connection from his `getting_started_progress`.
- User deleted the integration? In this case, both the `getting_started_meta` and `getting_started_progress`(es) attached to it are also deleted. It should all be created again if someone wants to go through the getting started flow.

### The fiddly details
Both connection and integration are deleted by soft-deletion. How I went around it:
- `getting_started_meta` shouldn't exist without an integration, so when an integration is soft-deleted, I hard delete `getting_started_meta` (which cascades to all related `getting_started_progress`(es)). I don't think it's worth adding soft-deletion to it, as it's a pain to maintain and it's not important.
- When fetching `getting-started-progress` I filter out soft-deleted `connections` when joining.

---

**Implement New Backend/API Endpoints for Getting Started Onboarding Flow**

This major pull request introduces the full backend and type support for a new 'Getting Started' onboarding flow, including API endpoints, service logic, supporting migrations (requires PR #4487), and comprehensive tests. It centers around creating and managing a special onboarding integration ('google-calendar-getting-started') using preprovisioned shared credentials, tracking user onboarding progress, and handling robust lifecycle management for onboarding resources (integration, meta, progress, and associated connections).

The implementation ensures the onboarding experience is always available: if onboarding meta, progress records, or the dedicated Google Calendar integration do not exist, they will be (re)created on demand. The PR includes endpoints for fetching and updating onboarding progress, sophisticated lifecycle rules for resource deletion (including cascading cleanup for soft-deleted integrations), and extensive integration tests for all critical flows and edge cases. All API contracts, DB types, and seeding/test utilities required for this flow are included.

<details>

• Introduces /api/v1/getting-started ``GET`` and ``PATCH`` endpoints for onboarding state
• Implements new onboarding service layer and supporting ``DB`` logic (with transactional safety and idempotency)
• Adds new tables: `getting_started_meta` and `getting_started_progress` (migrations are in a separate ``PR``)
• Automates provisioning and tying of a special `google-calendar-getting-started` integration using shared credentials
• Handles both soft and hard deletion of associated onboarding meta/progress when integration is deleted
• Ensures soft-deleted connections are filtered from onboarding progress, and detachment occurs on deletion
• Expands shared credentials service/utilities for preprovisioned provider/integration setup
• Adds and exposes all contract types, `DTOs`, and ``API`` types for new onboarding endpoints
• Includes 600+ lines of integration tests covering all onboarding scenarios and edge cases
• Updates seed utilities to streamline creation/testing of shared credentials and onboarding flows

</details>

<details>

• Server ``API`` controllers (new /getting-started endpoints, logic, and types)
• Service layer: new `gettingStartedService` and supporting `sharedCredentialsService` changes
• Database access layer: onboarding meta/progress, soft/hard deletion logic
• Types: types/lib/`gettingStarted`/*, api endpoints, integration with shared credentials
• Seeder/utilities: config and credentials creation helpers
• Test suite: new integration/controller tests for onboarding flows
• Integration deletion logic (now cascades to onboarding meta/progress)

</details>

---

### DIFF
--- a/packages/server/lib/controllers/integrations/uniqueKey/deleteIntegration.integration.test.ts
+++ b/packages/server/lib/controllers/integrations/uniqueKey/deleteIntegration.integration.test.ts
@@ -1,4 +1,4 @@
-import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+import { afterAll, assert, beforeAll, describe, expect, it } from 'vitest';
 
 import { gettingStartedService, seeders } from '@nangohq/shared';
 
@@ -52,6 +52,7 @@ describe(`DELETE ${endpoint}`, () => {
         isSuccess(res.json);
 
         const metaAfter = await gettingStartedService.getMetaByAccountId(account.id);
-        expect(metaAfter).toBeNull();
+        assert(!metaAfter.isErr(), 'Meta should be deleted');
+        expect(metaAfter.value).toBeNull();
     });
 });
--- a/packages/server/lib/controllers/integrations/uniqueKey/deleteIntegration.integration.test.ts
+++ b/packages/server/lib/controllers/integrations/uniqueKey/deleteIntegration.integration.test.ts
@@ -48,7 +48,7 @@ describe(`DELETE ${endpoint}`, () => {
             oauth_scopes: 'hello, world'
         });
 
-        const metaResult = await gettingStartedService.getOrCreateGettingStartedMeta(account.id, env.id);
+        const metaResult = await gettingStartedService.getOrCreateMeta(account.id, env.id);
         expect(metaResult.isOk()).toBe(true);
         const meta = metaResult.unwrap();
 
--- a/packages/server/lib/controllers/integrations/uniqueKey/deleteIntegration.integration.test.ts
+++ b/packages/server/lib/controllers/integrations/uniqueKey/deleteIntegration.integration.test.ts
@@ -1,8 +1,11 @@
 import { afterAll, beforeAll, describe, expect, it } from 'vitest';
 
-import { seeders } from '@nangohq/shared';
+import db from '@nangohq/database';
+import { gettingStartedService, seeders } from '@nangohq/shared';
 
 import { isSuccess, runServer, shouldBeProtected } from '../../../utils/tests.js';
+
+import type { DBGettingStartedMeta } from '@nangohq/types';
 
 let api: Awaited<ReturnType<typeof runServer>>;
 
@@ -33,5 +36,29 @@ describe(`DELETE ${endpoint}`, () => {
         expect(res.json).toStrictEqual<typeof res.json>({
             success: true
         });
+    });
+
+    it('should delete getting started meta', async () => {
+        const { env, account } = await seeders.seedAccountEnvAndUser();
+
+        // Getting started meta expects a preprovisioned provider config
+        await seeders.createPreprovisionedProviderConfigSeed(env, 'google-calendar-getting-started', 'google-calendar', {
+            oauth_client_id: 'foo',
+            oauth_client_secret: 'bar',
+            oauth_scopes: 'hello, world'
+        });
+
+        const metaResult = await gettingStartedService.getOrCreateGettingStartedMeta(account.id, env.id);
+        expect(metaResult.isOk()).toBe(true);
+        const meta = metaResult.unwrap();
+
+        const res = await api.fetch(endpoint, { method: 'DELETE', token: env.secret_key, params: { uniqueKey: 'google-calendar-getting-started' } });
+        isSuccess(res.json);
+        expect(res.json).toStrictEqual<typeof res.json>({
+            success: true
+        });
+
+        const metaAfter = await db.knex.from<DBGettingStartedMeta>('getting_started_meta').where({ id: meta.id }).first();
+        expect(metaAfter).toBeUndefined();
     });
 });
--- a/packages/server/lib/controllers/integrations/uniqueKey/deleteIntegration.integration.test.ts
+++ b/packages/server/lib/controllers/integrations/uniqueKey/deleteIntegration.integration.test.ts
@@ -1,11 +1,8 @@
 import { afterAll, beforeAll, describe, expect, it } from 'vitest';
 
-import db from '@nangohq/database';
 import { gettingStartedService, seeders } from '@nangohq/shared';
 
 import { isSuccess, runServer, shouldBeProtected } from '../../../utils/tests.js';
-
-import type { DBGettingStartedMeta } from '@nangohq/types';
 
 let api: Awaited<ReturnType<typeof runServer>>;
 
@@ -50,15 +47,11 @@ describe(`DELETE ${endpoint}`, () => {
 
         const metaResult = await gettingStartedService.getOrCreateMeta(account.id, env.id);
         expect(metaResult.isOk()).toBe(true);
-        const meta = metaResult.unwrap();
 
         const res = await api.fetch(endpoint, { method: 'DELETE', token: env.secret_key, params: { uniqueKey: 'google-calendar-getting-started' } });
         isSuccess(res.json);
-        expect(res.json).toStrictEqual<typeof res.json>({
-            success: true
-        });
 
-        const metaAfter = await db.knex.from<DBGettingStartedMeta>('getting_started_meta').where({ id: meta.id }).first();
-        expect(metaAfter).toBeUndefined();
+        const metaAfter = await gettingStartedService.getMetaByAccountId(account.id);
+        expect(metaAfter).toBeNull();
     });
 });
--- a/packages/server/lib/controllers/integrations/uniqueKey/deleteIntegration.integration.test.ts
+++ b/packages/server/lib/controllers/integrations/uniqueKey/deleteIntegration.integration.test.ts
@@ -1,5 +1,6 @@
 import { afterAll, assert, beforeAll, describe, expect, it } from 'vitest';
 
+import db from '@nangohq/database';
 import { gettingStartedService, seeders } from '@nangohq/shared';
 
 import { isSuccess, runServer, shouldBeProtected } from '../../../utils/tests.js';
@@ -45,13 +46,13 @@ describe(`DELETE ${endpoint}`, () => {
             oauth_scopes: 'hello, world'
         });
 
-        const metaResult = await gettingStartedService.getOrCreateMeta(account.id, env.id);
+        const metaResult = await gettingStartedService.getOrCreateMeta(db.knex, account.id, env.id);
         expect(metaResult.isOk()).toBe(true);
 
         const res = await api.fetch(endpoint, { method: 'DELETE', token: env.secret_key, params: { uniqueKey: 'google-calendar-getting-started' } });
         isSuccess(res.json);
 
-        const metaAfter = await gettingStartedService.getMetaByAccountId(account.id);
+        const metaAfter = await gettingStartedService.getMetaByAccountId(db.knex, account.id);
         assert(!metaAfter.isErr(), 'Meta should be deleted');
         expect(metaAfter.value).toBeNull();
     });
--- a/packages/server/lib/controllers/v1/gettingStarted/getGettingStarted.integration.test.ts
+++ b/packages/server/lib/controllers/v1/gettingStarted/getGettingStarted.integration.test.ts
@@ -1,9 +1,10 @@
 import { afterAll, beforeAll, describe, expect, it } from 'vitest';
 
 import db from '@nangohq/database';
-import { getProvider, seeders } from '@nangohq/shared';
+import { configService, getProvider, gettingStartedService, seeders } from '@nangohq/shared';
 
 import { authenticateUser, isSuccess, runServer, shouldBeProtected } from '../../../utils/tests.js';
+import { getOrchestrator } from '../../../utils/utils.js';
 
 import type { DBGettingStartedMeta, DBGettingStartedProgress } from '@nangohq/types';
 
@@ -151,18 +152,12 @@ describe(`GET ${endpoint}`, () => {
         expect(fresh?.connection_id).toBe(connection.id);
     });
 
-    it('returns connection as null when the linked connection is deleted', async () => {
-        const { env, user, account } = await seeders.seedAccountEnvAndUser();
+    it('returns connection as null when the linked connection is soft deleted', async () => {
+        const { env, user } = await seeders.seedAccountEnvAndUser();
         const session = await authenticateUser(api, user);
 
-        // Ensure integration exists
-        const integration = await seeders.createPreprovisionedProviderConfigSeed(env, 'google-calendar-getting-started', 'google-calendar');
-
-        // Create meta and a progress row with a connection
-        const [meta] = await db.knex
-            .from<DBGettingStartedMeta>('getting_started_meta')
-            .insert({ account_id: account.id, environment_id: env.id, integration_id: integration.id! })
-            .returning('*');
+        const gettingStartedProgress = await gettingStartedService.getOrCreateProgressByUser(user, env.id);
+        expect(gettingStartedProgress.isOk()).toBe(true);
 
         // Create a connection and link it to the progress
         const connection = await seeders.createConnectionSeed({
@@ -171,19 +166,11 @@ describe(`GET ${endpoint}`, () => {
             connectionId: 'test-conn-id'
         });
 
-        const [progress] = await db.knex
-            .from<DBGettingStartedProgress>('getting_started_progress')
-            .insert({ user_id: user.id, getting_started_meta_id: meta!.id, step: 2, complete: false, connection_id: connection.id })
-            .returning('*');
-        if (!progress) throw new Error('Failed to create progress');
+        // Attach the connection to the progress
+        await gettingStartedService.updateByUserId(user.id, { connection_id: connection.id });
 
-        // Verify the connection is initially linked
-        const initialRes = await api.fetch(endpoint, { method: 'GET', query: { env: 'dev' }, session });
-        isSuccess(initialRes.json);
-        expect(initialRes.json.data.connection).toMatchObject({ id: connection.id, connection_id: 'test-conn-id' });
-
-        // Delete the connection
-        await db.knex.from('_nango_connections').where({ id: connection.id }).delete();
+        // Soft delete the connection (not using service because it has lots of dependencies)
+        await db.knex.from('_nango_connections').where({ id: connection.id }).update({ deleted: true, deleted_at: new Date() });
 
         // Verify the connection is now null in the response
         const res = await api.fetch(endpoint, { method: 'GET', query: { env: 'dev' }, session });
@@ -193,14 +180,9 @@ describe(`GET ${endpoint}`, () => {
             data: {
                 meta: { environment: { id: env.id }, integration: { unique_key: 'google-calendar-getting-started' } },
                 connection: null,
-                step: 2,
+                step: 0,
                 complete: false
             }
         });
-
-        // Verify the progress row still exists but connection_id is null
-        const freshProgress = await db.knex.from<DBGettingStartedProgress>('getting_started_progress').where({ id: progress.id }).first();
-        expect(freshProgress).toBeTruthy();
-        expect(freshProgress?.connection_id).toBeNull();
     });
 });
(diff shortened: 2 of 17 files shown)


packages/server/lib/controllers/integrations/uniqueKey/deleteIntegration.integration.test.ts
```diff
@@ -34,4 +37,28 @@ describe(`DELETE ${endpoint}`, () => {
...
+
+        const metaResult = await gettingStartedService.getOrCreateGettingStartedMeta(account.id, env.id);
+        expect(metaResult.isOk()).toBe(true);
+        const meta = metaResult.unwrap();
+
+        const res = await api.fetch(endpoint, { method: 'DELETE', token: env.secret_key, params: { uniqueKey: 'google-calendar-getting-started' } });
+        isSuccess(res.json);
+        expect(res.json).toStrictEqual<typeof res.json>({
+            success: true
+        });
```
TBonnin: are we not testing the same thing twice?

packages/shared/lib/services/config.service.ts
```diff
@@ -145,25 +146,79 @@ class ConfigService {
...
+        provider: Provider;
+        display_name?: string;
+        unique_key?: string;
+    }): Promise<IntegrationConfig> {
         const sharedCredentialsId = await this.getSharedCredentialsId(providerName);
         if (!sharedCredentialsId) {
             throw new NangoError('shared_credentials_not_found');
         }
 
+        let finalUniqueKey = unique_key ?? providerName;
```
TBonnin: it is not this final since we mutate it again a couple of lines below ;) 
what about:
```
const providedUniqueKey = unique_key ?? providerName
const dbUniqueKey = await db.knex
            .select('unique_key')
            .from<ProviderConfig>(`_nango_configs`)
            .where({ unique_key: providedUniqueKey, environment_id, deleted: false })
            .first();

const config = await this.createProviderConfig(
            {
                environment_id,
                unique_key: dbUniqueKey ??  `${providedUniqueKey}-${nanoid(4).toLocaleLowerCase()}`,
```

packages/shared/lib/services/config.service.ts
```diff
@@ -145,25 +146,79 @@ class ConfigService {
...
+        const sharedCredentialsId = await this.getSharedCredentialsId(originalProviderName);
+        if (!sharedCredentialsId) {
+            throw new NangoError('shared_credentials_not_found');
+        }
+
+        const exists = await db.knex
+            .count<{ count: string }>('*')
+            .from<ProviderConfig>(`_nango_configs`)
+            .where({ provider: customProviderName, environment_id, deleted: false })
+            .first();
```
TBonnin: why using count instead of selecting the provider unique key?

packages/shared/lib/services/getting-started.service.ts
```diff
@@ -0,0 +1,264 @@
...
+    DBUser,
+    GettingStartedOutput as GettingStartedProgressOutput,
+    IntegrationConfig,
+    PatchGettingStartedInput
+} from '@nangohq/types';
+import type { Result } from '@nangohq/utils';
+
+export async function createGettingStartedMeta(accountId: number, environmentId: number, integrationId: number): Promise<Result<DBGettingStartedMeta>> {
+    try {
+        const result = await db.knex
```
TBonnin: you can do `const [inserted] = ...` so you don't have deal with indices below

packages/shared/lib/services/getting-started.service.ts
```diff
@@ -0,0 +1,264 @@
...
+                db.knex.raw('row_to_json(conn.*) as connection')
+            )
+            .leftJoin('getting_started_meta as meta', 'meta.id', 'progress.getting_started_meta_id')
+            .leftJoin('_nango_environments as env', 'env.id', 'meta.environment_id')
+            .leftJoin('_nango_configs as config', 'config.id', 'meta.integration_id')
+            .leftJoin('_nango_connections as conn', function () {
+                this.on('conn.id', '=', 'progress.connection_id').andOnNull('conn.deleted_at');
+            })
+            .where({ user_id: userId })
+            .first();
```
TBonnin, line 67: have you checked this query is fast enough?

packages/shared/lib/services/getting-started.service.ts
```diff
@@ -0,0 +1,264 @@
...
+        return Ok({
+            meta: {
+                integration: result.integration,
+                environment: result.environment
+            },
+            connection: result.connection ?? null,
+            step: result.progress.step
+        });
+    } catch (err) {
+        console.error('failed_to_get_getting_started_progress', err);
```
TBonnin: 🪓

packages/shared/lib/services/getting-started.service.ts
```diff
@@ -0,0 +1,264 @@
...
+        return Err(new Error('failed_to_get_getting_started_progress', { cause: err }));
+    }
+}
+
+export function updateByUserId(userId: number, data: Partial<DBGettingStartedProgress>) {
```
TBonnin: should return a `Result`

packages/shared/lib/services/getting-started.service.ts
```diff
@@ -0,0 +1,264 @@
...
+    const update: Partial<DBGettingStartedProgress> = {};
+
+    if (typeof input.step !== 'undefined') {
+        update.step = input.step;
+    }
+
+    if (typeof input.connection_id !== 'undefined') {
+        if (input.connection_id === null || input.connection_id === '') {
+            update.connection_id = null;
+        } else {
```
TBonnin: what about 
```
const update: Partial<DBGettingStartedProgress> = {
   ...(input.step ? { input.step } : {})
};

if(input.connectionId) {
      const connection = await connectionService.getConnectionById(connectionId)
      if(!connection) {
         return Err(...)
      }
} else {
   update.connectionId = null
}
```

using the connectionService to get the connection instead of having a connection related query in the getting-started-service

packages/shared/lib/services/getting-started.service.ts
```diff
@@ -0,0 +1,264 @@
...
+        const newMeta = await createGettingStartedMeta(accountId, currentEnvironmentId, googleCalendarIntegrationId.value);
+
+        if (newMeta.isErr()) {
+            return Err(new Error('failed_to_create_getting_started_meta'));
+        }
+
+        return Ok(newMeta.value);
+    } catch (err) {
+        return Err(new Error('failed_to_create_getting_started_meta', { cause: err }));
+    }
```
TBonnin: why the try catch?

packages/shared/lib/services/getting-started.service.ts
```diff
@@ -0,0 +1,264 @@
...
+}
+
+async function getOrCreateGoogleCalendarIntegration(environmentId: number): Promise<Result<number>> {
+    const existingIntegrationId = await configService.getIdByProviderConfigKey(environmentId, 'google-calendar-getting-started');
```
TBonnin: unfortunately `getIdByProviderConfigKey` is a legacy function that can throw. Might be a good time to refactor so it return a result

packages/shared/lib/services/getting-started.service.ts
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    const PROVIDER_NAME = 'google-calendar';
+
+    const provider = getProvider(PROVIDER_NAME);
+    if (!provider) {
+        return Err(new Error('google_calendar_provider_not_found'));
+    }
+
+    const newIntegration = await configService.createPreprovisionedProvider({
```
TBonnin: can we refactor createPreprovisionedProvider to return a result. I guess we missed it in the code review

packages/shared/lib/services/getting-started.service.ts
```diff
@@ -0,0 +1,264 @@
...
+
+    if (!id) {
+        return Err(new Error('failed_to_create_google_calendar_integration'));
+    }
+
+    return Ok(id);
+}
+
+export async function deleteMetaByIntegrationId(integrationId: number): Promise<Result<void>> {
+    const result = await db.knex.from<DBGettingStartedMeta>('getting_started_meta').where({ integration_id: integrationId }).delete();
```
TBonnin: db query must be try/catched

packages/shared/lib/services/getting-started.service.ts
```diff
@@ -0,0 +1,264 @@
...
+
+    if (result === 0) {
+        return Err(new Error('failed_to_delete_getting_started_meta'));
+    }
+
+    return Ok(undefined);
```
TBonnin: might be easier to reverse 
```
if(result) {
   return Ok(undefined)
}

return Err(...)
```

packages/server/lib/controllers/v1/gettingStarted/getGettingStarted.integration.test.ts
```diff
@@ -0,0 +1,182 @@
...
+        shouldBeProtected(res);
+    });
+
+    it("creates meta+integration when they don't exist", async () => {
+        const { env, user, account } = await seeders.seedAccountEnvAndUser();
+        await seeders.createSharedCredentialsSeed('google-calendar');
+        const session = await authenticateUser(api, user);
+
+        // Ensure no getting_started_meta exists for this account
+        await db.knex.from<DBGettingStartedMeta>('getting_started_meta').where({ account_id: account.id }).delete();
```
TBonnin: can we add functions to the service that will get/insert/delete meta and progress so they can be used internally but also reused in tests instead of having db queries all over the place

packages/types/lib/gettingStarted/dto.ts
```diff
@@ -0,0 +1,17 @@
...
+    meta: {
+        environment: DBEnvironment;
+        integration: IntegrationConfig;
+    };
+    connection: DBConnection | null;
+    step: number;
+}
+
+export interface PatchGettingStartedInput {
+    connection_id?: string | null | undefined;
```
TBonnin, line 16: I have the feeling modeling the missing connection_id as `none` (or similar) instead of `null` will help. So many way to get a condition wrong and mix things up between null and undefined

vite.integration.config.ts
```diff
@@ -17,7 +17,8 @@ export default defineConfig({
             NANGO_LOGS_ES_PREFIX: 'test',
             FLAG_PLAN_ENABLED: 'true',
             ORCHESTRATOR_SERVICE_URL: 'http://orchestrator',
-            RUNNER_NODE_ID: '1'
+            RUNNER_NODE_ID: '1',
+            FLAG_API_RATE_LIMIT_ENABLED: 'false'
```
TBonnin, line 21: is this flag actually being used?

packages/shared/lib/services/getting-started.service.ts
kaposke: He actually added it in the new PR with the endpoints. I'm rebasing and updating accordingly

packages/shared/lib/services/getting-started.service.ts
kaposke: I haven't, but I also don't think performance is critical for this one, as it will be low occurence. Probably fast enough though.

packages/shared/lib/services/getting-started.service.ts
kaposke: Ugh yeah, should definitely use the service. Though the method that allows the string `connection_id` also requires the provider-config-key, which is super annoying. Is it safe to do without nowadays?

packages/shared/lib/services/getting-started.service.ts
kaposke: Thanks, I think it was left-over.

packages/shared/lib/services/getting-started.service.ts
kaposke: Since it has 12 references to it, it likely belongs in a separate PR.

packages/types/lib/gettingStarted/dto.ts
kaposke: I think undefined and null serve exactly that purpose.
And yeah, if we do simple things like `if (connection_id)`, there's no differentiation, but it's weird to go out of our ways to avoid human error on a patch endpoint, where that sort of behavior is expected.

The same problem happens to `step?: number | undefined`. If I have `if (step) { // set step }`, I won't be able to set it to `0`. So it's down to checking explicitly for undefined, unfortunately.

packages/server/lib/controllers/integrations/uniqueKey/deleteIntegration.integration.test.ts
kaposke: This one tests explicitly that the integration being (soft) deleted also causes the `getting_started_meta` to be deleted. I had to add it explicitly to the `configService` unfortunately, to emulate cascading behavior but with the soft-delete.

Don't think I've tested the same thing elsewhere?

TBonnin: No I mean between the 2 lines: what does the isSuccess brings that the expect wouldn't pick up?

kaposke: Yes 🤦

packages/shared/lib/services/config.service.ts
kaposke: Improved after rebase.

packages/server/lib/controllers/v1/gettingStarted/getGettingStarted.integration.test.ts
kaposke: Did that, but also significantly simplified a lot of tests and removed some. Lots of them looked like bloat now that I'm looking back

vite.integration.config.ts
kaposke: Yeah, by the rateLimiterMiddleware. Causes problems now that I run a `signin` before every request.

packages/shared/lib/services/getting-started.service.ts
TBonnin: > Probably fast enough though

Famous last word. 😆  I would do a sanity check and run the query plan in prod just too be sure if I were you. I cannot count how many times I have said the same thing and happened to be wrong

packages/shared/lib/services/getting-started.service.ts
```diff
@@ -0,0 +1,290 @@
...
+        if (!result) {
+            return Ok(null);
+        }
+
+        return Ok({
+            meta: {
+                integration: pick(result.integration, ['id', 'unique_key', 'provider', 'display_name']),
+                environment: pick(result.environment, ['id', 'name'])
+            },
+            connection: result.connection ? pick(result.connection, ['id', 'connection_id']) : null,
```
TBonnin, line 78: do we really need `loadash.pick` here?

packages/shared/lib/services/getting-started.service.ts
```diff
@@ -0,0 +1,290 @@
...
+
+export async function createMeta(input: CreateGettingStartedMeta): Promise<Result<DBGettingStartedMeta>> {
+    try {
+        const [gettingStartedMeta] = await db.knex.from<DBGettingStartedMeta>('getting_started_meta').insert(input).returning('*');
+
+        if (!gettingStartedMeta) {
+            return Err(new Error('failed_to_create_getting_started_meta'));
+        }
+
+        return Ok(gettingStartedMeta);
```
TBonnin: can use a ternary here and in other functions below: 
```
return gettingStartedMeta ? Ok(gettingStartedMeta) : Err('failed_to_create_getting_started_meta')`
```

also `Err` accepts a string and internally creates a new Error. Saves a few keystrokes

packages/shared/lib/services/getting-started.service.ts
```diff
@@ -0,0 +1,290 @@
...
+    const newProgress = await getProgressByUserId(createdResult.value.user_id);
+    if (newProgress.isErr()) {
+        return Err(newProgress.error);
+    }
+
+    if (newProgress.value === null) {
+        return Err(new Error('failed_to_get_getting_started_progress'));
+    }
+
+    return Ok(newProgress.value);
```
TBonnin: all of those queries should be in a transaction to avoid race conditions and allow rollbacks

packages/shared/lib/services/getting-started.service.ts
```diff
@@ -0,0 +1,290 @@
...
+
+    try {
+        if (Object.keys(update).length > 0) {
+            await updateByUserId(user.id, update);
+        }
+    } catch (err) {
+        return Err(new Error('failed_to_update_getting_started_progress', { cause: err }));
+    }
+
+    return Ok(undefined);
```
TBonnin: same thing here. A transaction is needed

packages/types/lib/gettingStarted/dto.ts
TBonnin: the problem I have with `undefined` is that the  absence of value and the value being explicitly set to "none" is represented the same way, even though they don't necessarily means the same thing. I am not requesting any changes here, it is more a problem with the js type system. ;)

packages/shared/lib/services/getting-started.service.ts
kaposke: Well, we don't. But I saw no harm since it makes this significantly shorter. Looking back, i can also add these selections to the query instead, though it will get even more complex.

packages/shared/lib/services/getting-started.service.ts
kaposke: Sounds good. Locally it's all index scans, so looks ok tbh. I can check in prod once migrations are merged.

packages/shared/lib/services/getting-started.service.ts
```diff
@@ -0,0 +1,288 @@
...
+    DBGettingStartedProgress,
+    DBUser,
+    GettingStartedOutput as GettingStartedProgressOutput,
+    IntegrationConfig,
+    PatchGettingStartedInput
+} from '@nangohq/types';
+import type { Result } from '@nangohq/utils';
+import type { Knex } from 'knex';
+
+export async function createMeta(input: CreateGettingStartedMeta, trx?: Knex.Transaction): Promise<Result<DBGettingStartedMeta>> {
```
TBonnin: I'll set a default value for the optional parameter
```
function createMeta( db: knex.Knex, input: CreateGettingStartedProgress) ...
```

For consistency I will make the db param first and add it to all functions that talk to the db, even the one that are not used in a transaction, so if one day they must be called from within a transaction we don't forget.

I would also make it a required param. The caller will always have a db instance to pass

packages/shared/lib/services/getting-started.service.ts
```diff
@@ -0,0 +1,288 @@
...
+        if (typeof input.step !== 'undefined') {
+            update.step = input.step;
+        }
+
+        if (typeof input.connection_id !== 'undefined') {
+            if (input.connection_id === null || input.connection_id === '') {
+                update.connection_id = null;
+            } else {
+                try {
+                    const { error, response: connection } = await connectionService.getConnection(
```
TBonnin, line 177: that's why all functions that talk to the db should have a knex instance passed as argument. This query should be part of the transaction. I am not asking you to change it though.

packages/shared/lib/services/getting-started.service.ts
TBonnin: let' try catch it then ;)

TBonnin (approved):

packages/shared/lib/services/getting-started.service.ts
```diff
@@ -216,70 +203,80 @@ export async function patchProgressByUser(user: DBUser, input: PatchGettingStart
     });
 }
 
-export async function getOrCreateMeta(accountId: number, currentEnvironmentId: number, trx?: Knex.Transaction): Promise<Result<DBGettingStartedMeta>> {
-    const existingMeta = await getMetaByAccountId(accountId, trx);
+export async function getOrCreateMeta(dbInstance: Knex, accountId: number, currentEnvironmentId: number): Promise<Result<DBGettingStartedMeta>> {
+    return await dbInstance.transaction(async (trx) => {
```
TBonnin: you don't need the await

PR #4494, opened 2025-08-15
fix(ui): fix button text logic when creating integrations with provisioned apps

hassan254-prog:
- Fix the button text logic to display `Creating...` when creating integrations with provisioned apps across different envs.

---

This PR refines the logic controlling button text and disabled states when creating integrations using provisioned (preconfigured) apps in the integration creation dialog. The update ensures buttons now accurately display 'Creating...' and set their disabled state based on both the current environment (production or not) and the integration creation progress, improving visual feedback consistency for users across environments.


kaposke (approved):

packages/webapp/src/pages/Integrations/Create.tsx
```diff
@@ -222,10 +222,18 @@ export default function Create() {
...
                                 </Button>
-                                <Button type="button" onClick={() => handleIntegrationCreation(env === 'prod')} variant="secondary">
-                                    {env === 'prod' ? "Use Nango's developer app" : 'Use your own developer app'}
+                                <Button type="button" onClick={() => handleIntegrationCreation(env === 'prod')} disabled={isCreatingShared} variant="secondary">
+                                    {isCreatingShared && env === 'prod'
+                                        ? 'Creating...'
+                                        : env === 'prod'
+                                          ? "Use Nango's developer app"
+                                          : 'Use your own developer app'}
                                 </Button>
```
bodinsamuel, line 237: nb: Button have an `isLoading` prop that should be used instead of disabled

hassan254-prog: @bodinsamuel , thanks, I will open a pr to fix that.

PR #4499, opened 2025-08-18
feat: error message when 429s from nango API in scripts

TBonnin:
More self-explanatory message in Nango logs when the Nango API rate-limit is hit

We currently show a message that says "An error occurred during an HTTP call" when a customer hit the rate-limit from a script. 

---

**Add Specific Error Messaging for 429 Rate Limits from Nango API in Scripts**

This PR enhances error handling for scripts that interact with the Nango API. When a script receives an HTTP 429 (Too Many Requests) error from the Nango API, the system now issues a specific error type (`script_api_rate_limit_error`) with a clear, actionable message instructing users about mitigating rate limits and pointing them to support if necessary. The implementation ensures that users are not presented with ambiguous generic HTTP errors but are instead informed of the exact cause (Nango API rate limiting), improving troubleshooting and transparency in script logs. Updates are performed in both the script runner's error classification logic and the central error utility.

<details>

• Introduced logic in `packages/runner/lib/exec.ts` to specifically detect ``HTTP`` 429 errors from calls to the Nango ``API`` domains (`api.nango.dev`, `localhost`) during script execution.
• Assigned a new error type: ``script_api_rate_limit_error`` when such 429s are detected.
• Extended error message mapping in `packages/shared/lib/utils/error.ts` to give detailed user instructions for this new error, with status code 424.
• Old behavior (generic ``HTTP`` error message for 429s) is replaced with targeted, user-friendly instructions and mitigation suggestions.

</details>

<details>

• Script execution error handling (packages/runner/lib/exec.ts)
• Central error messaging and type mapping (packages/shared/lib/utils/error.ts)

</details>

---

### DIFF
--- a/packages/runner/lib/exec.ts
+++ b/packages/runner/lib/exec.ts
@@ -246,10 +246,19 @@ export async function exec({
                         errorResponse && typeof errorResponse === 'object' ? (errorResponse as Record<string, unknown>) : { message: errorResponse }
                     );
 
+                    let type = 'script_http_error';
+                    // If the error is a rate limit error from Nango API, we return a specific type/message
+                    if (err.response.status === 429 && err.response.config.url) {
+                        const url = new URL(err.response.config.url);
+                        if (url.hostname === 'api.nango.dev' || url.hostname === 'localhost') {
+                            type = 'script_api_rate_limit_error';
+                        }
+                    }
+
                     return {
                         success: false,
                         error: {
-                            type: 'script_http_error',
+                            type,
                             payload: responseBody,
                             status: err.response.status,
                             additional_properties: {
--- a/packages/shared/lib/utils/error.ts
+++ b/packages/shared/lib/utils/error.ts
@@ -481,7 +481,7 @@ export class NangoError extends NangoInternalError {
 
             case 'script_api_rate_limit_error':
                 this.status = 424;
-                this.message = `Too many operations performed against the Nango API. Reduce the number of actions and/or the frequency of your syncs. Contact Nango support to increase your limits.`; //TODO
+                this.message = `Script API rate limit exceeded. Your Nango scripts made too many calls to the Nango API within the allowed window. Mitigation: reduce how often scripts run (e.g. lower sync frequency, fewer actions). Need higher limits? Contact Nango support.`;
                 break;
 
             case 'script_internal_error':
--- a/packages/shared/lib/utils/error.ts
+++ b/packages/shared/lib/utils/error.ts
@@ -479,6 +479,11 @@ export class NangoError extends NangoInternalError {
                 this.message = `An error occurred during an HTTP call`;
                 break;
 
+            case 'script_api_rate_limit_error':
+                this.status = 424;
+                this.message = `Too many operations performed against the Nango API. Reduce the number of actions and/or the frequency of your syncs. Contact Nango support to increase your limits.`; //TODO
+                break;
+
             case 'script_internal_error':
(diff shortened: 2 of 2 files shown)


packages/shared/lib/utils/error.ts
```diff
@@ -479,6 +479,11 @@ export class NangoError extends NangoInternalError {
                 this.message = `An error occurred during an HTTP call`;
                 break;
 
+            case 'script_api_rate_limit_error':
+                this.status = 424;
+                this.message = `Too many operations performed against the Nango API. Reduce the number of actions and/or the frequency of your syncs. Contact Nango support to increase your limits.`; //TODO
```
TBonnin: @bastienbeurier can I have your input here? This is the log message that will be displayed to the user when a call to the Nango API fails from scripts. How should we phrase it? Thank you

Screenshot 2025-08-18 at 16 12 44

bastienbeurier: Proposed edit: 
> Script API rate limit exceeded.
> Your Nango scripts made too many calls to the Nango Script API within the allowed window.
> Mitigation: reduce how often scripts run (e.g. lower sync frequency, fewer actions) or reduce operations per run (e.g. batch saves, metadata reads or writes, logging).
> Need higher limits? Contact Nango support.

What do you think? I changed to scripts because this could also apply to webhook scripts and event-based scripts. 

Also, should we be explicit about whether this caused the script to fail or whether it will be retried, only slowing down the script execution?

cc @rguldener @khaliqgant @rossmcewan @bodinsamuel  @kaposke

bodinsamuel: This error is only logged if it's critical. But I don't think we have retries in place for internal API calls so I guess it's always critical

TBonnin: @bastienbeurier This PR is only changing the message. The log is still showing as an error that failed the script. 
I like your edit. I would only challenge this part  `or reduce operations per run (e.g. batch saves, metadata reads or writes, logging).` 
batchSave and logging are not hitting the API and are therefore not impacted by rate-limit, same for proxy. It is only calls like getConnection, get/setMetadata, getEnvVars, ... that could possibly getting a 429

bastienbeurier: Got it, makes sense to remove. Thanks!

packages/runner/lib/exec.ts
```diff
@@ -246,10 +246,19 @@ export async function exec({
                         errorResponse && typeof errorResponse === 'object' ? (errorResponse as Record<string, unknown>) : { message: errorResponse }
                     );
 
+                    let type = 'script_http_error';
+                    // If the error is a rate limit error from Nango API, we return a specific type/message
+                    if (err.response.status === 429 && err.response.config.url) {
+                        const url = new URL(err.response.config.url);
+                        if (url.hostname === 'api.nango.dev' || url.hostname === 'localhost') {
```
TBonnin, line 253: is there a better way to check the http call was to the nango api?

kaposke: My first thought was that the runner-sdk itself would detect 429's in the correct places, not for this to be handled in this global manner.

bodinsamuel (approved):

packages/runner/lib/exec.ts
TBonnin: was my initial idea but we would need to handle the error and format the response here anyway so I went for the solution where the error handling is in a single place

PR #4505, opened 2025-08-19
feat(runner-sdk): retry 429s from Nango API

TBonnin:
I have implemented the retry on 429s in the runner-sdk in this PR but after testing it for a bit locally I am starting to have second thoughts. 
When hitting the rate-limit, requests can starts to pile up very quickly and waiting time can be up to 60s, making every scripts just super slow. I am starting to think that failing quick and early and not retrying is probably the best course of action. @bastienbeurier @NangoHQ/engineers wdyt?

### DIFF
--- a/packages/runner/lib/sdk/sdk.ts
+++ b/packages/runner/lib/sdk/sdk.ts
@@ -1,3 +1,7 @@
+import { setTimeout } from 'node:timers/promises';
+
+import { isAxiosError } from 'axios';
+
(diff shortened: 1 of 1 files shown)


TBonnin:
closing. The UX is, in a lot of scenarios, not great when we retry 429s in the runner-sdk. Every script becomes slower and might timeout unexpectedly. We currently think it is better to fail quickly with a clear message.

PR #4507, opened 2025-08-20
fix(invitation): foundUs is not required

bodinsamuel:
## Changes

---

This pull request updates the signup form logic so that the 'How did you hear about Nango?' (foundUs) input is no longer required if the user is signing up from an invitation link. Previously, 'foundUs' was always rendered and required regardless of the signup context. The new logic conditionally shows and requires the field only for non-invited signups.


khaliqgant (approved):
thanks

PR #4517, opened 2025-08-21
fix(billing): sync data with orb and stripe

bodinsamuel:
## Changes

fixes https://linear.app/nango/issue/NAN-3655/stripe-to-orb-data-syncing

- Sync data with orb and stripe
When changing team name, and correctly replicate billing address 

---

**Event-Driven Team Data Synchronization with Stripe and Orb Billing Providers**

This PR introduces an event-driven architecture for keeping team data in sync across Stripe and Orb billing providers. When a team's name or billing address is updated, a 'team.updated' event is published and asynchronously processed by a new Team event processor, ensuring customer information is updated in both billing systems. The changes also clarify and separate what data is synchronized during webhooks vs. manual updates, centralize Stripe SDK import/export, and improve error handling and observability around billing sync failures. The interface for billing clients is expanded to expose a generic 'updateCustomer' method, and architectural decisions move side effects away from controllers in favor of retryable, resilient background processing. Related controller logic and type definitions are refactored to leverage these enhancements.

<details>

• Introduced `team.updated` pubsub event and Team event processor for asynchronous billing provider sync.
• Expanded `BillingClient` ``API`` with ``updateCustomer``, implemented for Orb.
• Refactored `putTeam` controller to publish events after local account changes, removing direct billing provider updates.
• Improved error handling and retry logic for third-party billing provider failures.
• Centralized Stripe ``SDK`` access and moved all imports to the billing package.
• Refined Stripe webhook handling: only billing address is updated on payment method events.
• Upgraded related type signatures, controller, and workflow code to support new event-driven sync.

</details>

<details>

• Team update ``API``/controller and endpoints
• Billing client abstraction, Orb and Stripe implementations
• Pubsub/event system-including event type declarations and processors
• Stripe payment, manual collection, and webhook flows
• Application startup/composition in metering app
• Type definitions for billing clients
• Package.json dependencies for billing/server

</details>

---


packages/server/lib/controllers/v1/team/putTeam.ts
```diff
@@ -29,11 +31,21 @@ export const putTeam = asyncWrapper<PutTeam>(async (req, res) => {
...
 
     await accountService.editAccount({ id: account.id, ...body });
 
+    if (plan?.stripe_customer_id && plan?.orb_customer_id) {
+        const stripe = getStripe();
+        await Promise.all([
+            stripe.customers.update(plan.stripe_customer_id, {
+                name: body.name
+            }),
+            billing.updateCustomer(plan.orb_customer_id, body.name)
```
TBonnin: can this lead to one being updated an not the other? Is that a problem?
also how it any error handled?

TBonnin (approved):

packages/server/lib/controllers/v1/team/putTeam.ts
```diff
@@ -29,11 +31,21 @@ export const putTeam = asyncWrapper<PutTeam>(async (req, res) => {
...
 
-    const { account } = res.locals;
+    const { account, plan } = res.locals;
     const body: PutTeam['Body'] = val.data;
 
     await accountService.editAccount({ id: account.id, ...body });
 
+    if (plan?.stripe_customer_id && plan?.orb_customer_id) {
+        const stripe = getStripe();
+        await Promise.all([
```
kaposke: These sound like side-effects that should happen as a reaction to an event. Should we start leveraging our new infra?

packages/server/lib/controllers/v1/team/putTeam.ts
kaposke: Another reason for it to be a retry-able side-effect.

packages/server/lib/controllers/v1/team/putTeam.ts
TBonnin: very good point actually 💯

packages/server/lib/controllers/v1/team/putTeam.ts
bodinsamuel: not really a problem it's just for display

packages/server/lib/controllers/v1/team/putTeam.ts
bodinsamuel: don't know if it's ready for this kind of use case? cc @TBonnin

bodinsamuel: okay let's try then

packages/server/lib/controllers/v1/team/putTeam.ts
```diff
@@ -37,17 +36,7 @@ export const putTeam = asyncWrapper<PutTeam>(async (req, res) => {
     await accountService.editAccount({ id: account.id, ...body });
 
     if (plan?.stripe_customer_id && plan?.orb_customer_id) {
```
TBonnin, line 38: I would always publish the event regardless of the plan and let the consumer decide what do to and when to do it

packages/server/lib/controllers/v1/team/putTeam.ts
```diff
@@ -37,17 +36,7 @@ export const putTeam = asyncWrapper<PutTeam>(async (req, res) => {
...
-            await Promise.all([
-                stripe.customers.update(plan.stripe_customer_id, {
-                    name: body.name
-                }),
-                billing.updateCustomer(plan.orb_customer_id, body.name)
-            ]);
-        } catch (err) {
-            report(new Error('Failed to update customer name', { cause: err }), { accountId: account.id });
-        }
+        void pubsub.publisher.publish({ subject: 'team', type: 'team.updated', payload: { id: account.id } });
```
TBonnin, line 39: do we only send the id or should we send the entire account object?

packages/server/package.json
```diff
@@ -62,7 +62,6 @@
         "semver": "7.6.3",
         "serialize-error": "11.0.3",
         "simple-oauth2": "5.1.0",
-        "stripe": "18.2.1",
```
TBonnin, line 65: 🫶

TBonnin (approved):

packages/server/lib/controllers/v1/team/putTeam.ts
bodinsamuel: This question is often hard to answer.  I feel like sending the whole object is most of the time "useless", but could also lead to desync if it takes a long time to process. Plus serialization is annoying (date etc.) breaks the expected types

TBonnin: valid point.
Wrt serialization: current pubsub implementation use v8 serialize and therefore supports most native js types ;) https://github.com/NangoHQ/nango/blob/d5b79df367b086790f11b64702703e607dc374f0/packages/pubsub/lib/utils/serde.ts#L1

PR #4546, opened 2025-08-26
fix: `patchUser` zod schema

kaposke:
This PR updates the zod schema used for validating PATCH requests to the user controller by renaming the field `closedGettingStarted` to `gettingStartedClosed`. This change ensures consistency between the request body schema and the corresponding database/user object fields.

### DIFF
--- a/packages/server/lib/controllers/v1/user/patchUser.ts
+++ b/packages/server/lib/controllers/v1/user/patchUser.ts
@@ -11,7 +11,7 @@ import type { DBUser, PatchUser } from '@nangohq/types';
 const validation = z
     .object({
         name: z.string().min(3).max(255).optional(),
-        closedGettingStarted: z.boolean().optional()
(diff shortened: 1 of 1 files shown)


TBonnin (approved):
I feel this should be caught by the compiler. I don't actually understand why `exactOptionalPropertyTypes` doesn't catch this tbh

packages/server/lib/controllers/v1/user/patchUser.ts
```diff
@@ -11,7 +11,7 @@ import type { DBUser, PatchUser } from '@nangohq/types';
 const validation = z
     .object({
         name: z.string().min(3).max(255).optional(),
-        closedGettingStarted: z.boolean().optional()
```
bodinsamuel, line 14: annoying that it's compatible with the types :/ 
we should find a better solution

PR #4547, opened 2025-08-26
fix: allow server to startup when ES isn't responding

TBonnin:
Server currently doesn't start if ES migration fails for whatever reason. Which means that during an ES outage server cannot be deployed or scaled up.
This commit ensures the server still starts up even if it cannot connect to ES (only if ConnectionError, other errors still cause the startup to fail)
The risk is to have the ES migration failing silently, the benefits is server can operate fully when ES is down. wdty @bodinsamuel 

 

---

**Graceful Server Startup When Elasticsearch Is Unresponsive**

This PR refactors the Elasticsearch (ES) initialization logic to allow the server to start even if ES is unreachable, preventing deployment and scaling blocks during ES outages. The key approach is to allow server startup to proceed when ES migration fails with a connection error (e.g., network issue), while still failing startup for any other migration or initialization errors. Migration attempts now return a Result<Ok/Err> type rather than throwing directly, with improved error logging and classification.

<details>

• Refactored ``migrateMapping`` to return a Result (Ok/Err) instead of throwing; distinguishes connection errors from other failures.
• The `start` function now inspects the Result: on ``ES`` connection errors, logs and skips migration allowing startup; for other errors, startup is still blocked.
• Updated imports to use `@elastic/elasticsearch` error types and Result helpers from `@nangohq/utils`.
• Improved logging to clarify ``ES`` initialization/migration success or specific error type.

</details>

<details>

• packages/logs/lib/es/helpers.ts

</details>

---

### DIFF
--- a/packages/logs/lib/es/helpers.ts
+++ b/packages/logs/lib/es/helpers.ts
@@ -1,4 +1,6 @@
-import { isTest } from '@nangohq/utils';
+import { errors } from '@elastic/elasticsearch';
+
+import { Err, Ok, isTest } from '@nangohq/utils';
 
 import { envs } from '../env.js';
 import { logger } from '../utils.js';
@@ -8,6 +10,8 @@ import { getFormattedMessage, getFormattedOperation } from '../models/helpers.js
 import { createMessage } from '../models/messages.js';
 import { createOperation } from '../models/operations.js';
 
+import type { Result } from '@nangohq/utils';
+
 export async function start() {
     if (!envs.NANGO_LOGS_ENABLED) {
         logger.warning('Elasticsearch is disabled, skipping');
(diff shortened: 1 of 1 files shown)


packages/logs/lib/es/helpers.ts
```diff
@@ -62,9 +75,11 @@ export async function migrateMapping() {
                 }
             }
         }
+        return Ok(undefined);
     } catch (err) {
-        logger.error(err);
-        throw new Error('failed_to_init_elasticsearch');
+        const errMsg = err instanceof errors.ConnectionError ? 'failed_to_connect_elasticsearch' : 'failed_to_init_elasticsearch';
```
TBonnin, line 80: should we also account for TimeoutError? https://github.com/elastic/elasticsearch-js/blob/80044746f0f8cf7ee1c06eb1ff882ae3c5bc76b5/docs/reference/connecting.md?plain=1#L468

bodinsamuel (approved):
With the risk of not migrating "silently".

TBonnin:
> With the risk of not migrating "silently".

yes. Whoever is deploying a ES schema change should check the log. Not ideal though.

PR #4549, opened 2025-08-27
feat(api): PATCH /connections/:id

bodinsamuel:
## Changes

- PATCH /connections/:id
Only accepts end_user for now

---

**Add PATCH /connections/:id Endpoint for Partial Connection Updates (end_user Only)**

This PR introduces a new PATCH /connections/:connectionId endpoint to the public API, enabling partial updates to a connection resource-currently limited to updating the end_user field. The change covers backend controller logic, OpenAPI specification updates, type and schema definitions, adjustments to the Node SDK, as well as user- and developer-facing documentation. Several related schemas, validation functions, and documentation navigation elements are also refactored to improve consistency and strictness.

<details>

• Introduces ``PATCH`` /connections/:`connectionId` endpoint for partial updates (currently only `end_user` can be updated).
• Implements new backend controller (`patchConnection`.ts) with strict schema validation for params, body, and query.
• Defines `PatchPublicConnection` type in @nangohq/types and updates type usage throughout the codebase.
• Updates `OpenAPI` spec (spec.yaml): documents new endpoint, response types, error handling, and consolidates schemas (notably, `EndUser` model).
• Extends Node ``SDK`` with `patchConnection` method and updates ``SDK`` documentation.
• Adds new ``API`` and ``SDK`` documentation for patching a connection and reorganizes doc navigation.
• Refactors shared type definitions for `EndUser`/`ConnectSessionInput` for clarity and reuse.
• Enhances strictness in connection creation validation, corrects `additionalProperties` and object schemas for stricter `OpenAPI` compliance.

</details>

<details>

• ``API`` surface (public ``HTTP`` ``API`` and `OpenAPI` specification)
• Backend controllers (connection: patch handler, validation logic)
• `TypeScript` ``API``/type definitions (@nangohq/types)
• Node ``SDK`` (`patchConnection` method and docs)
• Documentation (``API`` reference, ``SDK`` reference, navigation structure)
• Validation layers (for connection, syncs, end users, etc.)

</details>

---


docs-v2/spec.yaml
```diff
@@ -900,14 +867,48 @@ paths:
...
+                  required: true
+                  schema:
+                      type: string
+                  description: The connection ID used to create the connection.
+                - name: provider_config_key
+                  in: query
+                  required: true
+                  description: The integration ID used to create the connection (aka Unique Key).
+                  schema:
+                      type: string
```
kaposke, line 886: Are we close to not requiring this? Also, is this not usually a header?

docs-v2/spec.yaml
```diff
@@ -3294,6 +3272,22 @@ components:
...
+                id:
+                    type: string
+                    description: Uniquely identifies the end user.
+                email:
+                    type: string
+                display_name:
+                    type: string
+                tags:
+                    type: object
+                    additionalProperties: true
```
kaposke: I know we already use `true` everywhere, but `additionalProperties` can be a schema. It could literally be:
```yaml
additionalProperties:
  type: string
  maxLength: 64
```
> https://spec.openapis.org/oas/v3.1.0.html#model-with-map-dictionary-properties

packages/server/lib/controllers/connection/postConnection.ts
```diff
@@ -30,7 +30,7 @@ import { asyncWrapper } from '../../utils/asyncWrapper.js';
 
 import type { AuthOperationType, ConnectionConfig, ConnectionUpsertResponse, EndUser, PostPublicConnection, ProviderGithubApp } from '@nangohq/types';
 
-const schemaBody = z.object({
+const schemaBody = z.strictObject({
```
kaposke, line 33: Are we sure these unrelated changes are not breaking?

packages/types/lib/connection/api/get.ts
```diff
@@ -141,6 +141,22 @@ export type GetPublicConnection = Endpoint<{
...
+    Method: 'PATCH';
+    Path: '/connections/:connectionId';
+    Params: {
+        connectionId: string;
+    };
+    Querystring: {
+        provider_config_key: string;
+    };
+    Body: {
+        end_user?: ConnectSessionInput['end_user'] | undefined;
```
kaposke: Should this become it's own model then?

kaposke (approved):

packages/types/lib/connection/api/get.ts
```diff
@@ -141,6 +141,22 @@ export type GetPublicConnection = Endpoint<{
...
+    };
+    Success: { success: boolean };
```
TBonnin, line 156: should it return the updated connection?

TBonnin (approved):
one question about returning the updated connection but lgtm otherwise

docs-v2/spec.yaml
bodinsamuel: not close, we have kind of given up on this until next API

docs-v2/spec.yaml
bodinsamuel: thanks I wasn't aware of this feature 👌🏻

packages/server/lib/controllers/connection/postConnection.ts
bodinsamuel: yes it was missing everywhere it was added

packages/types/lib/connection/api/get.ts
bodinsamuel: indeed

packages/types/lib/connection/api/get.ts
bodinsamuel: I never find that useful tbh, but I changed my mind multiple times on this over the years.

General thought process: 
When you GET an object, you reply with an expected interface that contains specific fields. When you PATCH/PUT the same object, you often don't have the same fields, so before I answer, I need to query the same thing again, just to reply with the same interface as the GET. 
If most users are ignoring the reply for PATCH/PUT (which is my belief, but I may be wrong), then it's a waste of resources. And if you have more than one representation, you are kind of screwed.

NB: Not applicable for this specific endpoint, but in modern UI, you have a different call to fetch that will react when you mutate, so the reply will most of the time be ignored unless it's an error.

With a more concrete example of this: 
- `GET /integrations` I get a simplified list of integrations
- `GET /integrations/:id` has two shapes, depending on whether you need the credentials or not
- If I do `PATCH /integrations/:id`, which of the 3 shapes am I returning?

TBonnin: all good. I agree with your though process. I wanted to make sure it was an explicit decision.

PR #4564, opened 2025-08-29
feat(connect): stop normalizing endUser

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-3747/remove-normalization-of-userorg-id

> [!NOTE]
> This is a proposal. I have struggled more than anticipated, and I'm not satisfied with the current result. But the only other alternatives are even more unsatisfying

- Connect: stop normalizing endUser

**Before**, we were creating an end user directly when requesting a Connect Session. Most of the time the `endUser` would already be known, ending up just updating some values. So if a connect session end up not being used, there would be very little churn.
**After**, if we create an end user directly, we will end up with thousands and thousands of unused rows in the DB.

> I have decided not to do it because that would require an immediate need for a new cron to clean up the data.
> NB: we do have churn right now, but less intense

---

In an ideal world, `endUser` is now just a column in `connections`, there is no reason to have a dedicated table anymore, and it would simplify operations a lot.

> I have also decided not to do that, because that would require modifying a lot of code depending/updating/creating connections.

---

**So I took an intermediate step:**
- Add a column endUser to `connect_sessions`, this allows us to store the future endUser
- When we create a connection, we store the `endUser`
- When we update a connection, we update the `endUser`

---

**Refactor Connect Session: Defer End User Normalization**

This PR changes how end-user data is handled in the Connect (OAuth) flow by deferring normalization (database insertion/update) until a connection is actually established or updated, rather than at Connect Session creation. The end-user payload is now stored as a JSON column on the `connect_sessions` table (added via migration), reducing unnecessary database churn and large numbers of unused end user rows. All controller logic, service functions, and TypeScript types related to Connect Sessions and end users have been updated accordingly, and a broad set of test cases and endpoint handlers are refactored to support this new approach.

<details>

• Introduced a new ``end_user`` ``JSON`` column to the ``connect_sessions`` table via migration.
• Connect Session creation now stores the end user payload as ``JSON``; no end user is created in the database at this stage.
• Normalization (creation/update) of the end user occurs only when a connection is created or updated.
• Refactored service, controller, and middleware logic to accommodate in-memory payload and defer ``DB`` writes.
• Updated all relevant `TypeScript` types and interface contracts for Connect Sessions and end users.
• Rewrote/adjusted relevant integration and unit tests to match the new lifecycle.
• Refactored downstream flows to propagate new end user format, and updated ``API``: connect/reconnect, patch, `OAuth`, etc.

</details>

<details>

• Connect Session and `EndUser` service logic
• Database schema (migrations) for `connect_sessions`
• Controllers related to all Connect flows (`getSession`, `postSession`, reconnect, `OAuth` flows, etc.)
• Type definitions in @nangohq/types (`ConnectSession`, `EndUser`, `InternalEndUser`)
• Middleware and Express request/response typings
• Integration and unit tests for Connect Sessions and end user handling

</details>

---


packages/server/lib/controllers/auth/postApiKey.ts
```diff
@@ -90,7 +90,7 @@ export const postPublicApiKeyAuthorization = asyncWrapper<PostPublicApiKeyAuthor
                 : await logContextGetter.create(
                       {
                           operation: { type: 'auth', action: 'create_connection' },
-                          meta: { authType: 'apikey', connectSession: endUserToMeta(res.locals.endUser) },
+                          meta: { authType: 'apikey', connectSession: res.locals.endUser ? endUserToMeta(res.locals.endUser) : undefined },
```
TBonnin: should the condition be in `endUserToMeta`?

packages/server/lib/controllers/auth/postApiKey.ts
```diff
@@ -189,7 +189,7 @@ export const postPublicApiKeyAuthorization = asyncWrapper<PostPublicApiKeyAuthor
                 account,
                 auth_mode: 'API_KEY',
                 operation: updatedConnection.operation,
-                endUser: isConnectSession ? res.locals['endUser'] : undefined
+                endUser: res.locals.endUser ?? undefined
```
TBonnin: isn't the `?? undefined` redundant?

packages/shared/lib/services/endUser.service.ts
```diff
@@ -191,74 +200,107 @@ export async function upsertEndUser(
...
+): Promise<Result<EndUser, EndUserError>> {
+    // Check if the endUser exists in the database
+    const endUserRes = await getEndUser(db, { id, accountId: account.id, environmentId: environment.id }, { forUpdate: true });
+    if (endUserRes.isErr()) {
+        return endUserRes;
+    }
+
+    console.log('updating end user', {endUser,
+        previousEndUser: endUserRes.value
+    });
```
TBonnin: 🪓

packages/shared/lib/services/endUser.service.ts
```diff
@@ -191,74 +200,107 @@ export async function upsertEndUser(
...
-    const endUser = endUserRes.value;
+    return await getAndUpdateEndUser(db, {
+        id: connection.end_user_id,
+        account,
+        environment,
+        endUser
+    });
+}
+
+async function getAndUpdateEndUser(
```
TBonnin: would `updateIfModified` be a better name for this function?

TBonnin (approved):

packages/server/lib/controllers/auth/postApiKey.ts
bodinsamuel: It's actually null, but I'll allow null it will be cleaner

packages/shared/lib/services/endUser.service.ts
bodinsamuel: oops

packages/server/lib/controllers/auth/postApiKey.ts
bodinsamuel: yeah right I can change that too 👌🏻

packages/shared/lib/services/endUser.service.ts
bodinsamuel: renamed to `getAndUpdateIfModifiedEndUser`

bodinsamuel:
To merge next monday

PR #4587, opened 2025-09-02
fix(paginate): throw error if `response_path` isn't present on the first page

hassan254-prog:
- Throw an error if the `response_path` isn't present on the first page. This is in regards to slack where they reply with HTTP 200 but include the error message in the response body.

---

**Throw Error if `response_path` Missing on First Page in Pagination**

This PR updates the pagination logic so that if the expected `response_path` is missing from the first page response, an error is thrown immediately. This addresses cases such as Slack API responses where a 200 HTTP status might be returned but the actual data path is absent, indicating a silent API error. The change affects all three pagination strategies (cursor, link, offset) and introduces a dedicated validation function invoked on the first page of each pagination sequence. Comprehensive unit tests are added to verify the correctness of this validation for various edge cases.

<details>

• Adds `validateFirstPageResponse` method to `PaginationService`.
• Enforces a check for presence and array type at the configured ``response_path`` (or root) on the first page for all pagination methods.
• Throws descriptive errors when ``response_path`` is missing or is not an array.
• Extends unit tests for link pagination to cover new error conditions (missing path, non-array payload, etc).
• Updates cursor and offset pagination control flows to use the new first page validation.

</details>

<details>

• packages/runner-sdk/lib/paginate.service.ts
• packages/runner-sdk/lib/paginate.service.unit.test.ts

</details>

---


TBonnin:
@hassan254-prog can you share an example of a Slack response and an example of a valid paginated response. I am not sure I fully understand the problem ?

hassan254-prog:
> @hassan254-prog can you share an example of a Slack response and an example of a valid paginated response. I am not sure I fully understand the problem ?

Screenshot 2025-09-02 at 17 45 36
Screenshot 2025-09-02 at 17 47 28

TBonnin:
@hassan254-prog isn't checking for response path applying to every page?

hassan254-prog:
>Yes, but right now the generator exits if `response_path` is undefined. We should throw an error after validating the first page so that such issues can be detected and handled properly.

why not throw on every page and only the first one?

bodinsamuel (commented):
I'm not sure I get why it would be an issue for just the first page?
You can get an error at any moment no?

hassan254-prog:
I decided to only check the first page as the determinant of whether the request was successful and if pagination can proceed correctly. This avoids cases where a provider paginates through results but doesn’t return an empty array on the last page, which could otherwise be mistakenly treated as an error.

bodinsamuel (approved):

TBonnin (approved):

PR #4590, opened 2025-09-02
feat(server): restrict ConnectUI customization with flags

kaposke:
Adds flag restrictions to ConnectUI settings.
Anything that an account doesn't have access to results in default settings being returned/overwritten on save.

---

**Enforce Feature Flag Restrictions for ConnectUI Customization**

This PR introduces server-side enforcement of account-level feature flags for ConnectUI settings, specifically controlling access to UI theme and watermark customization based on account plan permissions. If an account lacks the relevant flags, attempts to customize restricted fields are silently overridden to system defaults on both read and write (GET and PUT) operations. The implementation updates ConnectUI controller logic to reference plan feature flags, restructures how settings are checked/applied, and significantly rewrites and expands the test suites to verify correct behavior across all combinations of feature flag states.

<details>

• Added plan feature flag checks in both ``GET`` and ``PUT`` `ConnectUI` settings controllers.
• Fields in settings (theme, watermark) default to system values if current plan prohibits customization, overriding any stored customizations.
• ``PUT``: On save, forbidden fields are overwritten with defaults-no error response-matching behavior expected by frontend and internal ``API`` usage.
• ``GET``: Always returns default values for fields the plan does not allow customizing, regardless of ``DB`` state.
• Comprehensive rewrite and expansion of integration tests to cover all permutations of feature flags and code paths.
• Cleaner use of types, improved fixtures, code refactors around settings manipulation for clarity and maintainability.

</details>

<details>

• packages/server/lib/controllers/v1/`connectUISettings`/`getConnectUISettings`.ts
• packages/server/lib/controllers/v1/`connectUISettings`/`putConnectUISettings`.ts
• packages/server/lib/controllers/v1/`connectUISettings`/`getConnectUISettings`.integration.test.ts
• packages/server/lib/controllers/v1/`connectUISettings`/`putConnectUISettings`.integration.test.ts

</details>

---


packages/server/lib/controllers/v1/connectUISettings/putConnectUISettings.ts
```diff
@@ -38,9 +38,20 @@ export const putConnectUISettings = asyncWrapper<PutConnectUISettings>(async (re
...
+    const newSettings: ConnectUISettings = body;
+
+    // Override settings to defaults if the plan does not have the feature
+    if (flagHasPlan && !plan?.can_customize_connect_ui_theme) {
+        newSettings.theme = connectUISettingsService.defaultConnectUISettings.theme;
+    }
+
+    if (flagHasPlan && !plan?.can_disable_connect_ui_watermark) {
+        newSettings.showWatermark = connectUISettingsService.defaultConnectUISettings.showWatermark;
+    }
```
TBonnin, line 53: would it make more sense to return an error (405 maybe) instead of accepting the request when plan doesn't allow customization

kaposke: Makes sense, but because watermark and themes are blocked by separate flags, I have to either:
1. Change this to a PATCH request to accept partial changes and validate they're all valid.
2. Keep it as a PUT request, still take full object as input, but validate they are not changing/deviating from default if not allowed.

Or keep what I did, overriding defaults. It makes sense from the perspective of our frontend (and it's a private route), but I can see how it's a bit non-conventional.

TBonnin: PATCH request seems to be the most conventional. Will return an error if a change is not allowed but still allow to set both at the same time. As you said it is a private route so up to you

TBonnin (approved):

packages/server/lib/controllers/v1/connectUISettings/getConnectUISettings.integration.test.ts
```diff
@@ -1,13 +1,35 @@
...
 import db from '@nangohq/database';
-import { connectUISettingsService, seeders } from '@nangohq/shared';
+import { connectUISettingsService, seeders, updatePlan } from '@nangohq/shared';
 
 import { isSuccess, runServer, shouldBeProtected, shouldRequireQueryEnv } from '../../../utils/tests.js';
 
 const route = '/api/v1/connect-ui-settings';
 let api: Awaited<ReturnType<typeof runServer>>;
 
+const customSettings = {
```
bodinsamuel: should it be typed?

packages/server/lib/controllers/v1/connectUISettings/getConnectUISettings.integration.test.ts
```diff
@@ -88,86 +92,111 @@ describe(`GET ${route}`, () => {
...
-                    primaryForeground: '#ffffff',
-                    textPrimary: '#ffffff',
-                    textMuted: '#999999'
-                }
-            }
+    it('should return default theme when plan does not have can_customize_connect_ui_theme flag', async () => {
+        const { env, plan } = await seeders.seedAccountEnvAndUser();
+
+        const testSettings = {
+            ...customSettings,
```
bodinsamuel: It's a shallow copy. usually it's safer to have a functions returns default settings

packages/server/lib/controllers/v1/connectUISettings/putConnectUISettings.integration.test.ts
```diff
@@ -10,6 +10,28 @@ import type { ConnectUISettings } from '@nangohq/types';
 const route = '/api/v1/connect-ui-settings';
 let api: Awaited<ReturnType<typeof runServer>>;
 
+const customSettings = {
```
bodinsamuel: same comment about types and ref

bodinsamuel (approved):

PR #4592, opened 2025-09-02
feat: add fallback support to connection config interpolation

hassan254-prog:

- Allow fallbacks during interpolations. This will be useful when interpolating `token_params` for `oauth2_cc` and we have an `automated: true` field. 

---

**Add Fallback Support to Connection Config Interpolation**

This PR enhances the utility for string interpolation in connection configuration by adding fallback support. You can now specify fallback values (using '||') within interpolation expressions so that if the primary value is missing or undefined, a default (literal) value is used instead. The implementation touches only the core interpolation utility and adds relevant tests to ensure both the primary and fallback behaviors are covered.

<details>

• Extended `interpolateStringFromObject` to handle `||` fallback syntax, supporting use of fallback literals if the left-side variable is missing or empty.
• Added tests to utils.unit.test.ts for both success (primary variable present) and fallback (variable missing) scenarios.
• No deletions and no impact outside interpolation logic and related tests.

</details>

<details>

• packages/shared/lib/utils/utils.ts (core interpolation logic)
• packages/shared/lib/utils/utils.unit.test.ts (test coverage)

</details>

---


packages/shared/lib/utils/utils.ts
```diff
@@ -233,6 +233,17 @@ export function interpolateStringFromObject(str: string, replacers: Record<strin
...
 
+    if (str.includes('||')) {
+        const parts = str.split('||').map((part) => part.trim());
+        const left = parts[0] ? interpolateStringFromObject(parts[0], replacers) : undefined;
+
+        if (left && left !== parts[0]) {
+            return left;
+        }
+
+        return parts[1] ? interpolateStringFromObject(parts[1], replacers) : '';
```
TBonnin: what about 
```
if (str.includes('||')) {
    const [left, right = ''] = str.split('||').map(part => part.trim());
    
    if (left) {
        const interpolated = interpolateStringFromObject(left, replacers);
        if (interpolated && interpolated !== left) return interpolated;
    }
    
    return right ? interpolateStringFromObject(right, replacers) : '';
}
```

TBonnin (approved):

packages/shared/lib/utils/utils.ts
```diff
@@ -233,6 +233,17 @@ export function interpolateStringFromObject(str: string, replacers: Record<strin
...
+    }
```
bodinsamuel, line 245: we didn't code this already for user-agent?

bodinsamuel (approved):

packages/shared/lib/utils/utils.ts
hassan254-prog: Yes, we did, but that was for a different utility function. We need to clean this up since some of the functions share the same logic.

packages/shared/lib/utils/utils.ts
hassan254-prog: Yes, simplifies it, updated with the latest [commit](https://github.com/NangoHQ/nango/pull/4592/commits/0ed6b3312a7042794cd90c0efdcc63da887f3529) .

PR #4611, opened 2025-09-04
feat(secondary-token-refresh): allow refreshGithubAppJwtToken to be specified in getConnection

khaliqgant:
---

**feat(secondary-token-refresh): Add explicit secondary JWT token refresh to getConnection**

This PR introduces an explicit mechanism for refreshing a secondary credential, specifically JWT tokens used by GitHub App and GitHub App (OAuth) integrations. It extends the getConnection API (REST, Node SDK, Runner SDK) with a new parameter, refreshGithubAppJwtToken, allowing clients to request on-demand JWT token regeneration. The update generalizes the credential refresh logic, solidifies backend orchestration, and adjusts provider-specific authentication flows-all while ensuring compatibility with existing access/refresh token behaviors. Type definitions, OpenAPI schemas, validation, tests, and developer documentation are updated to support and document this addition.

<details>

• Added ``refreshGithubAppJwtToken`` boolean parameter to `getConnection` ``API``, Node ``SDK``, and Runner ``SDK`` for explicit secondary (``JWT``) token refresh.
• Refactored backend credential refresh logic to support multi-token/credential scenarios and propagate the explicit refresh flag.
• Enhanced provider-specific (`GitHub` App, `GitHub` App `OAuth`) logic to regenerate/store `JWTs` on explicit request.
• Updated public `APIs`, types, `OpenAPI` schemas, input validation, and ``SDK`` documentation to reflect the new refresh flag.
• Extended test coverage and adapted test cases for new refresh flag and updated ``API`` contracts.
• Maintained backward compatibility for all existing single-token integrations and usage.

</details>

<details>

• Node ``SDK`` ``API`` (packages/node-client/lib/index.ts)
• Runner ``SDK`` (packages/runner-sdk/lib/action.ts)
• Backend credential/connection service logic (shared/lib/services/connections/credentials/refresh.ts, connection.service.ts)
• Provider-specific flows for `GitHub` App/``JWT`` (shared/lib/auth/`githubApp`.ts, shared/lib/auth/jwt.ts)
• ``REST`` ``API`` controllers/validation (server/lib/controllers/connection/`connectionId`/`getConnection`.ts)
• Type definitions and `OpenAPI` schemas (types/lib/connection/api/get.ts, docs-v2/spec.yaml)
• Developer and ``SDK`` documentation (docs-v2/reference/sdks/node.mdx)

</details>

---


packages/node-client/lib/index.ts
```diff
@@ -273,15 +273,17 @@ export class Nango {
...
+     * @param tokenName - Optional. Pass in a secondary token name to ensure the token is refreshed and is retrieved
      * @returns A promise that resolves with a connection object
      */
     public async getConnection(
         providerConfigKey: string,
         connectionId: string,
         forceRefresh?: boolean,
-        refreshToken?: boolean
+        refreshToken?: boolean,
+        tokenName?: string
```
TBonnin: should the type be 'jwtToken' instead of string?

khaliqgant: Well, right now that is the only option, but this could be extended to return the user credential from github-app-oauth, or even the slack user token

TBonnin: doesn't really matter if this is the only option or not. The function can be called by mistake with any string and the compiler won't complain. I would tighten up the type as much as possible

khaliqgant: That's fair – if anything we've learned the more strict and less footguns the better especially wrt the public api and sdk.

Updated with https://github.com/NangoHQ/nango/pull/4611/commits/29af3c5ea1e6226bdd69c09afe53f4c20a3b054a

bodinsamuel (approved):
Looks good 👌🏻 
I would go for something more specific that tokenName, because it feels generic while it only applies to githubApp, e.g, `githubAppSecondaryToken: boolean` ugly but removes all confusion. Unless we really see another use case for this prop

khaliqgant:
> Looks good 👌🏻 I would go for something more specific that tokenName, because it feels generic while it only applies to githubApp, e.g, `githubAppSecondaryToken: boolean` ugly but removes all confusion. Unless we really see another use case for this prop

Yeah, I was back and forth between this. Very fair feedback. Let me check if it makes sense that the Slacks secondary user token would need logic similar to this and can be extended to that too.

khaliqgant:
> > Looks good 👌🏻 I would go for something more specific that tokenName, because it feels generic while it only applies to githubApp, e.g, `githubAppSecondaryToken: boolean` ugly but removes all confusion. Unless we really see another use case for this prop
> 
> Yeah, I was back and forth between this. Very fair feedback. Let me check if it makes sense that the Slacks secondary user token would need logic similar to this and can be extended to that too.

After looking at this it can be isolated to just github-app / github-app-oauth. Updated accordingly.

PR #4614, opened 2025-09-04
fix(jobs): fail quicker when runner is not available

TBonnin:
Back in the old days, the requests dispatching tasks to runner were inlined :O and timeouts were disabled so they could accomodate long running scripts. It has not been working like this for a while now but we kept the no timeouts which mean the requests from jobs to runner can take a long time to fail when runner is overwhelmed/not reachable. (I have seen 2+ mins in traces).
This commit adds timeouts for requests to runner so it fails quicker. This is one of the reason webhooks couldn't be processed quickly enough when we experienced some spikes in the past few days. Not that they would have been executed since the customer runner was not healthy but they would have failed fast

 

---

**Add HTTP Timeouts to Runner Client for Faster Failure on Unavailability**

This PR introduces configurable HTTP timeouts to the runner client used by the jobs service when making requests to the runner. By replacing previously unlimited (0) timeouts with explicit (and now environment-driven) values for connect, headers, and response timeouts, the system ensures that job-to-runner requests fail faster if the runner is unavailable or unresponsive. Environmental variables are added (with sensible defaults) to control these timeout values, improving system responsiveness during runner overload or downtime scenarios. The changes also update all runner client instantiations and related test cases to support and validate the new timeout contract.

<details>

• Added environment variables ``RUNNER_CLIENT_HEADERS_TIMEOUT_MS``, ``RUNNER_CLIENT_CONNECT_TIMEOUT_MS``, and ``RUNNER_CLIENT_RESPONSE_TIMEOUT_MS`` (with defaults: 10s, 5s, 15s) to configure ``HTTP`` timeouts.
• Modified `getRunnerClient`() signature to accept a timeout config and applied it to undici.Agent in runner client requests.
• Refactored `FleetRunner` and `RemoteRunner` to inject timeouts via `runnerHttpOpts`.
• Defined and exported `runnerHttpOpts` as a central config in runner.ts.
• Updated environment parsing logic to support new timeout variables and defaults.
• Adjusted runner client unit test to use explicit timeout parameters.

</details>

<details>

• packages/runner/lib/client.ts (runner ``HTTP`` client creation)
• packages/jobs/lib/runner/runner.ts (runner config)
• packages/jobs/lib/runner/fleet.runner.ts (`FleetRunner` construction)
• packages/jobs/lib/runner/remote.runner.ts (`RemoteRunner` construction)
• packages/utils/lib/environment/parse.ts (environment variable parsing/validation)
• packages/runner/lib/client.unit.test.ts (unit test for client)

</details>

---

### DIFF
--- a/packages/runner/lib/client.ts
+++ b/packages/runner/lib/client.ts
@@ -19,9 +19,9 @@ export function getRunnerClient(url: string): ProxyAppRouter {
                     return fetch(url, {
                         ...options,
                         dispatcher: new Agent({
-                            headersTimeout: 0,
-                            connectTimeout: 0,
-                            bodyTimeout: 0
+                            headersTimeout: 10_000,
+                            connectTimeout: 5_000,
+                            bodyTimeout: 15_000
                         })
                     });
                 }
--- a/packages/runner/lib/client.ts
+++ b/packages/runner/lib/client.ts
@@ -8,7 +8,14 @@ import type { RequestInit } from 'undici';
 
 export type ProxyAppRouter = CreateTRPCProxyClient<AppRouter>;
 
-export function getRunnerClient(url: string): ProxyAppRouter {
+export function getRunnerClient(
+    url: string,
+    httpOpts: {
+        headersTimeoutMs: number;
+        connectTimeoutMs: number;
+        responseTimeoutMs: number;
+    }
+): ProxyAppRouter {
     return createTRPCProxyClient<AppRouter>({
         transformer: superjson,
         links: [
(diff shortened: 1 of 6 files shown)


packages/runner/lib/client.ts
```diff
@@ -19,9 +19,9 @@ export function getRunnerClient(url: string): ProxyAppRouter {
                     return fetch(url, {
                         ...options,
                         dispatcher: new Agent({
-                            headersTimeout: 0,
-                            connectTimeout: 0,
-                            bodyTimeout: 0
+                            headersTimeout: 10_000,
+                            connectTimeout: 5_000,
+                            bodyTimeout: 15_000
                         })
```
TBonnin, line 32: picked some aggressive but I think sensible values. let me know what you think?

rossmcewan: Can we not inject these with env vars with these defaults?

bodinsamuel (approved):

PR #4615, opened 2025-09-04
fix: don't emit task completion events for syncs/webhooks

TBonnin:
task completion events for webhooks and syncs are never listened to. Only actions and onEvents tasks are actually awaited. Which means that the majority of events are emitted and always discarded 
With this commit, task completion events will only be emitted if actions or onEvents, reducing even more the number of task events

 

---

**Restrict Task Completion Events to Actions and onEvents Only**

This PR modifies the orchestrator's event system to only emit 'task:completed' events for 'action' and 'onEvent' task types, stopping completion event emissions for 'sync' and 'webhook' tasks as these are never awaited or listened to. The changes reduce unnecessary event traffic and tighten logic around event emission and handling, with updates across event emission, route handling, and comprehensive tests to validate the new behavior.

<details>

• Updated `taskEvents`.`taskCompleted` to return an event name only for action and `onEvent` tasks, returning undefined for sync and webhook tasks.
• Changed event emission logic in `TaskEventsHandler` (events.ts) to emit task:completed only if an event name is provided.
• Modified task output route (`getOutput`.ts) to conditionally add event listeners only when relevant task completion events exist.
• Refactored and expanded integration and unit tests to ensure correct task completion event emission and coverage of all task types.

</details>

<details>

• packages/orchestrator/lib/events.ts
• packages/orchestrator/lib/routes/v1/tasks/`taskId`/`getOutput`.ts
• packages/orchestrator/lib/events.integration.test.ts

</details>

---


rossmcewan (approved):
I'm not disagreeing that this is the right thing to do for now, but do you think this points to listen/notify not being an appropriate choice for the scale of messages we want to send? I imagine this just kicks the can down the road from a volume perspective.

bodinsamuel (approved):

TBonnin:
> I'm not disagreeing that this is the right thing to do for now, but do you think this points to listen/notify not being an appropriate choice for the scale of messages we want to send? I imagine this just kicks the can down the road from a volume perspective.

I do think that single channel listen/notify like we do today is not appropriate for this use-case. But it was a hour of my time to give us some breathing room. This is definitely not the end of the story

PR #4618, opened 2025-09-05
chore: add more metrics

bodinsamuel:
## Changes

Fixes https://linear.app/nango/issue/NAN-3826/add-metrics-for-webhooks-actions-and-proxy
Fixes https://linear.app/nango/issue/NAN-3894/add-metrics-to-datadog

- Add metrics to graph payloads
- Add breakdown for logs

---

**Expanded Metrics Instrumentation for Telemetry and Observability**

This PR substantially expands the system's observability by introducing new and refined metrics across core communication and execution pathways. Metrics for incoming and outgoing payload sizes are now collected for actions, proxies, webhooks, and deploy confirmations, with unit suffixes (e.g., _BYTES) for disambiguation. The runner component reports periodic memory usage with account context. Logging metrics support richer dimensional analysis and source tagging, and some minor log source consistency adjustments are made. All new metrics are integrated for export to Datadog and similar monitoring backends, enabling improved visibility into system performance, data volume, and resource consumption.

<details>

• Added new metric types for payload byte sizes (incoming and outgoing) for actions, proxies, webhooks, and deployment confirmations in metrics.ts.
• Introduced runner memory usage metric, reported periodically and tagged by inferred account ``ID``.
• Instrumented controllers (action, proxy, webhook, deploy confirmation) to record payload size metrics per request, using Buffer.`byteLength`, tagged with `accountId`.
• Proxy controller now records both incoming and outgoing payload sizes, including error cases.
• Log ingestion tags metrics with log source property for granular breakdowns.
• ``SDK`` log source set to `internal` for specific model-merge logs for consistency.
• Metric units standardized and included in metric names (e.g., _BYTES), following best practices.
• Minor reorganization and clarification in code per review feedback and internal discussions.

</details>

<details>

• Telemetry and metrics definitions (packages/utils/lib/telemetry/metrics.ts)
• Runner memory monitoring and reporting (packages/runner/lib/monitor.ts)
• Action, Proxy, Webhook, Deploy Confirmation controllers (packages/server/lib/controllers/...)
• Logging metrics and source tagging (packages/logs/lib/transport.ts)
• ``SDK`` logging behaviors (packages/runner/lib/sdk/sdk.ts)

</details>

---

### DIFF
--- a/packages/runner/lib/monitor.ts
+++ b/packages/runner/lib/monitor.ts
@@ -1,6 +1,8 @@
 import fs from 'fs';
 import os from 'os';
 
+import { metrics } from '@nangohq/utils';
+
 import { persistClient } from './clients/persist.js';
 import { envs } from './env.js';
 import { idle } from './idle.js';
@@ -8,6 +10,7 @@ import { logger } from './logger.js';
 
 import type { NangoProps } from '@nangohq/types';
 
+const regexRunnerUrl = /^http:\/\/(production|staging)-runner-account-(\d+|default)-\d+/;
 export class RunnerMonitor {
     private runnerId: number;
     private tracked = new Map<number, { nangoProps: NangoProps; taskId: string }>();
@@ -16,12 +19,14 @@ export class RunnerMonitor {
     private lastMemoryReportDate: Date | null = null;
     private idleInterval: NodeJS.Timeout | null = null;
     private memoryInterval: NodeJS.Timeout | null = null;
+    private runnerAccountId: string | null = null;
 
     constructor({ runnerId }: { runnerId: number }) {
         this.runnerId = runnerId;
         this.memoryInterval = this.checkMemoryUsage();
         this.idleInterval = this.checkIdle();
         process.on('SIGTERM', this.onExit.bind(this));
+        this.runnerAccountId = envs.RUNNER_URL ? (regexRunnerUrl.exec(envs.RUNNER_URL)?.[2] ?? null) : null;
     }
 
     private onExit(): void {
@@ -65,6 +70,9 @@ export class RunnerMonitor {
             const rss = process.memoryUsage().rss;
             const total = getTotalMemoryInBytes();
             const memoryUsagePercentage = (rss / total) * 100;
+            if (this.runnerAccountId) {
+                metrics.gauge(metrics.Types.RUNNER_MEMORY_USAGE, memoryUsagePercentage, { accountId: this.runnerAccountId || 1 });
+            }
             if (memoryUsagePercentage > envs.RUNNER_MEMORY_WARNING_THRESHOLD) {
                 await this.reportHighMemoryUsage(memoryUsagePercentage);
             }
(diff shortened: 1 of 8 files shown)


packages/runner/lib/monitor.ts
```diff
@@ -16,12 +19,14 @@ export class RunnerMonitor {
...
     private idleInterval: NodeJS.Timeout | null = null;
     private memoryInterval: NodeJS.Timeout | null = null;
+    private runnerAccountId: string | null = null;
 
     constructor({ runnerId }: { runnerId: number }) {
         this.runnerId = runnerId;
         this.memoryInterval = this.checkMemoryUsage();
         this.idleInterval = this.checkIdle();
         process.on('SIGTERM', this.onExit.bind(this));
+        this.runnerAccountId = envs.RUNNER_URL ? (regexRunnerUrl.exec(envs.RUNNER_URL)?.[2] ?? null) : null;
```
bodinsamuel, line 29: Not sure why we don't pass the accountId directly, so I went with that lmk

TBonnin: we didn't need it before I guess. I would pass it so it is cleaner than doing regex matching

packages/runner/lib/monitor.ts
```diff
@@ -65,6 +70,9 @@ export class RunnerMonitor {
             const rss = process.memoryUsage().rss;
             const total = getTotalMemoryInBytes();
             const memoryUsagePercentage = (rss / total) * 100;
+            if (this.runnerAccountId) {
+                metrics.gauge(metrics.Types.RUNNER_MEMORY_USAGE, memoryUsagePercentage, { accountId: this.runnerAccountId || 1 });
```
TBonnin: you shouldn't need the `|| 1` inside the `if`

TBonnin: I would pass the total and not the percentage because it'd depend on the size of the runner

packages/utils/lib/telemetry/metrics.ts
```diff
@@ -94,7 +99,9 @@ export enum Types {
 
     RECORDS_TOTAL_COUNT = 'nango.records.total.count',
 
-    CRON_PERSIST_ACCOUNT_USAGE = 'nango.server.cron.persistAccountUsage'
+    CRON_PERSIST_ACCOUNT_USAGE = 'nango.server.cron.persistAccountUsage',
+
+    DEPLOY_INCOMING_PAYLOAD_SIZE = 'nango.server.deploy.incoming.payload.size'
```
TBonnin: I would have the unit in the name and value of the env var (ex: `DEPLOY_INCOMING_PAYLOAD_SIZE_BYTES`)  so there is no confusion possible

TBonnin (approved):

packages/runner/lib/monitor.ts
bodinsamuel: I have the opposite thought, if we don't log PCT this metric will be unusable to compare customer.

TBonnin: It depends what we want to do with the metric. Is it for monitoring, billing?

bodinsamuel: currently it's just to get an idea of the volume and monitoring.
Billing will be in GB

packages/runner/lib/monitor.ts
bodinsamuel: Since fleet/runner does not contain the `account_id` it's not possible anyway

PR #4620, opened 2025-09-05
feat(persist): add endpoint to delete outdated records

TBonnin:
The endpoint will be called by a new runner-sdk function that customers can call in their custom integrations to delete records from previous executions (equivalent to the track_deletes functionalities).
The endpoint is not used yet

---

**Add API Endpoint and Service Logic to Delete Outdated Persist Records**

This PR adds a new DELETE REST API endpoint to the Persist module, allowing deletion of outdated (i.e., prior-generation) records for a given connection, sync, and sync job. The records service is refactored to provide a type-safe, result-oriented `deleteOutdatedRecords` method (superseding `markPreviousGenerationRecordsAsDeleted`), with all usages updated accordingly. Input/output validation and API response types are strengthened, and the new endpoint is fully covered by unit, integration, and end-to-end tests. Server wiring is updated to expose the new endpoint, but it is not yet used in production.

<details>

• Introduced ``DELETE`` endpoint at /environment/:`environmentId`/connection/:`nangoConnectionId`/sync/:`syncId`/job/:`syncJobId`/outdated, including Zod-based validation, error handling, and logging.
• Implemented and exported records service method ``deleteOutdatedRecords``, refactoring and deprecating ``markPreviousGenerationRecordsAsDeleted``; all internal usages updated.
• Updated the sync job execution flow to consume ``deleteOutdatedRecords`` using typed results (Ok/Err).
• Expanded ``API`` response and types in @nangohq/types, e.g., added `DeleteOutdatedRecordsSuccess`.
• Unit, integration, and end-to-end ``API`` tests added to cover deletion of outdated records, batch operation behavior, and record count updates.
• Server routing and middleware expanded to mount the new endpoint; minimal `ESLint` rule update/override for template string array logging.

</details>

<details>

• persist ``API`` server and routing
• records service/data layer (delete/mark outdated records logic)
• jobs sync execution flow (`track_deletes` logic)
• ``API``/type definitions in @nangohq/types
• records and server test suites
• minor `ESLint` rule exception

</details>

---

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -149,6 +149,7 @@ export default tseslint.config(
                 {
                     allowNumber: true,
                     allowBoolean: true,
+                    allowArray: true,
                     allowNever: true
                 }
             ],
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -149,7 +149,6 @@ export default tseslint.config(
                 {
                     allowNumber: true,
                     allowBoolean: true,
-                    allowArray: true,
                     allowNever: true
                 }
             ],
--- a/packages/jobs/lib/execution/sync.ts
+++ b/packages/jobs/lib/execution/sync.ts
@@ -291,12 +291,16 @@ export async function handleSyncSuccess({ taskId, nangoProps }: { taskId: string
         for (const model of nangoProps.syncConfig.models || []) {
             let deletedKeys: string[] = [];
             if (nangoProps.syncConfig.track_deletes) {
-                deletedKeys = await records.markPreviousGenerationRecordsAsDeleted({
+                const res = await records.deleteOutdatedRecords({
                     connectionId: nangoProps.nangoConnectionId,
                     model,
                     syncId: nangoProps.syncId,
                     generation: nangoProps.syncJobId
                 });
+                if (res.isErr()) {
+                    throw res.error;
+                }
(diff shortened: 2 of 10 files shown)


eslint.config.mjs
```diff
@@ -149,6 +149,7 @@ export default tseslint.config(
                 {
                     allowNumber: true,
                     allowBoolean: true,
+                    allowArray: true,
```
TBonnin: not sure arrays were not already allowed but this is needed so the endpoint can return `{ deletedKeys: string[] }`

bodinsamuel: it's not your endpoint it's when using string template: 
`void logCtx.info(`${model}: "Deleted ${result.value} outdated records for model ${model}"`);` 
The problem with doing that is that it's often arrays of objects, so instead of doing what you want it's doing this: 
```ts
console.log(`hello ${[1, { foo: 'bar'}, 3]}`)

hello 1,[object Object],3
```

bodinsamuel (approved):

kaposke (approved):

eslint.config.mjs
TBonnin: ok. in this case let's keep the rule as it is and have a disable exception in the endpoint since we know there it is a array of string

PR #4627, opened 2025-09-09
feat(sdk): telemetry

bodinsamuel:
## Changes

- Add self-reported telemetry to runner SDK
It's quite simple, just a class that can easily be transmitted back to jobs. Not used right now. Feel free to suggest an alternative

- Bonus: Changed all responses type

---

**Introduce SDK Self-Reported Telemetry and Refactor Runner Response Types**

This PR introduces a self-reported telemetry mechanism within the runner SDK, enabling the system to track and return SDK metrics (e.g., custom logs, proxy calls) via a new `telemetryBag` property attached to all SDK outputs. Simultaneously, it overhauls the internal and API-facing response types, replacing the previous `{ success, error, response }` structure with a more idiomatic `Result` approach and central `ExecutionError` class to standardize and serialize error handling. Extensive changes update all internal method signatures, data propagation, test cases, API validation schemas, and type definitions to align with this new telemetry-aware, type-safe result model.

<details>

• Added `telemetryBag` to `NangoActionBase` and exposed on all runner SDK outputs to capture metrics (e.g., `customLogs`, `proxyCalls`).
• Replaced `RunnerOutput`'s `{ success, error, response }` with a `Result` type pattern: successful results contain `{ output, telemetryBag }`, errors use new serializable `ExecutionError` objects.
• Refactored all relevant APIs, OpenAPI/Zod schemas, job handlers, and type definitions (`RunnerOutput`, `RunnerOutputError`, `PutTask`, etc.) to incorporate the new telemetry structure.
• Updated test suites (`exec.unit.test.ts`) to match new `Result` and `ExecutionError` contract.
• Instrumented SDK core methods (`proxy`, `log`) to increment relevant `telemetryBag` counters.
• Standardized on a TypeScript `TelemetryBag` interface backed by simple number counters for extensibility.
• Reworked error serialization/deserialization via `ExecutionError` to ensure consistent API error propagation.

</details>

<details>

• `packages/runner-sdk/lib/action.ts`
• `packages/runner-sdk/lib/errors.ts`
• `packages/runner/lib/exec.ts`
• `packages/runner/lib/server.ts`
• `packages/types/lib/runner/index.ts`
• `packages/types/lib/runner/sdk.ts`
• `packages/types/lib/jobs/api.ts`
• `packages/jobs/lib/routes/tasks/putTask.ts`
• `packages/runner/lib/exec.unit.test.ts`

</details>

---


packages/runner/lib/sdk/sdk.ts
```diff
@@ -144,6 +145,7 @@ export class NangoActionRunner extends NangoActionBase<never, Record<string, str
         // arrays are not supported in the log meta, so we convert them to objects
         const meta = Array.isArray(payload) ? Object.fromEntries(payload.map((e, i) => [i, e])) : payload || undefined;
 
+        this.telemetryBag.customLogs += 1;
```
TBonnin: we increment proxy after making the request but increment logs before. I think we should be consistent because in case of failure they wouldn't be accounted the same way

packages/runner-sdk/lib/telemetry.ts
```diff
@@ -0,0 +1,13 @@
...
+    customLogs: number = 0;
+    proxyCalls: number = 0;
+
+    toJSON(): TelemetryBagJSON {
+        return {
+            customLogs: this.customLogs,
+            proxyCalls: this.proxyCalls
+        };
+    }
+}
```
TBonnin: Do we need this class? Can we just have a runner telemetry type in `types` and initialize it once in runner-sdk action base ?

packages/types/lib/runner/sdk.ts
```diff
@@ -45,3 +45,8 @@ export interface NangoProps {
 export interface UserLogParameters {
     level?: 'info' | 'debug' | 'error' | 'warn' | 'http' | 'verbose' | 'silly';
 }
+
+export interface TelemetryBagJSON {
```
TBonnin: I would just call it Telemetry. 

also what about extending `Record<string, JsonValue>` to ensure metrics will always be jsonable. We could even restrict the value to be `number`

bodinsamuel: I wanted to emphasize that it's not a regular Telemetry class and object that sends stuff directly (vs the dozen other telemetry things we already have). 
good idea for the extend

TBonnin: Good point. I think I had called in `Metrics` in a PR that didn't make it to master but TelemetryBag is fine with me.

packages/runner/lib/sdk/sdk.ts
bodinsamuel: true 👌🏻

packages/runner-sdk/lib/telemetry.ts
bodinsamuel: simplified.
I did that because I anticipated it would become more involved with time but that can be changed later

packages/types/lib/runner/sdk.ts
```diff
@@ -45,3 +45,8 @@ export interface NangoProps {
 export interface UserLogParameters {
     level?: 'info' | 'debug' | 'error' | 'warn' | 'http' | 'verbose' | 'silly';
 }
+
+export interface TelemetryBagJSON extends Record<string, number> {
```
TBonnin: I would call the interface TelemetryBag but up to you

TBonnin (approved):

PR #4633, opened 2025-09-09
feat: deleteRecordsFromPreviousExecution docs  and deprecate trackDeletes

TBonnin:
track_deletes is being deprecated:
- marked as @deprecated in zero yaml
- nango compile warning when set to true
- logs warning when set to true

Note: we don't deprecate it explicitly in legacy nango.yaml since it is EOL in december

This commit replace trackDeletes documentation by documentation about the `deleteRecordsFromPreviousExecution` runner-sdk function.

 

---

**Deprecate `trackDeletes` and Transition to `deleteRecordsFromPreviousExecutions` for Deletion Detection**

This PR deprecates the `trackDeletes` flag in favor of the new `deleteRecordsFromPreviousExecutions` function for handling automatic deletion detection in sync integrations. The documentation, SDK API, CLI behavior, and associated type definitions have been updated throughout the codebase to both discourage the use of `trackDeletes` and encourage correct, explicit usage of the new function via scripts. Compiler- and runtime-level warnings now signal deprecated usage, and additional static and runtime checks enforce correct sequence and count of `deleteRecordsFromPreviousExecutions` calls. All public and internal documentation, including guides, SDK reference, and code comments, have been updated to reflect this new approach, with details on expected usage patterns and possible error scenarios.

<details>

• Marked `trackDeletes` as `@deprecated` in documentation, SDK type definitions, and zero YAML schemas.
• Added compile-time and runtime warnings in `packages/cli/lib/zeroYaml/compile.ts` and job execution logic if `trackDeletes` is set to `true`.
• Updated all documentation (`docs-v2/guides/syncs/detecting-deletes.mdx`, `docs-v2/reference/scripts.mdx`, `docs-v2/guides/custom-integrations/build-a-custom-integration.mdx`) to remove or rewrite `trackDeletes` sections in favor of `deleteRecordsFromPreviousExecutions`.
• Refactored CLI, SDK, runtime jobs, and parser logic to use pluralized `deleteRecordsFromPreviousExecutions` instead of singular/interim forms, ensuring API consistency.
• Enforced usage restrictions: `deleteRecordsFromPreviousExecutions` can only be called once per sync and only after all `batchSave` calls; violations emit compile-time errors.
• Amended and added tests, code snapshots, and type declarations to ensure correct name and enforcement patterns.
• Added explicit warnings in logs and docs about potential error scenarios and safe usage of deletion detection.

</details>

<details>

• `docs-v2/guides/syncs/detecting-deletes.mdx`
• `docs-v2/reference/scripts.mdx`
• `docs-v2/guides/custom-integrations/build-a-custom-integration.mdx`
• `packages/cli/lib/zeroYaml/compile.ts`
• `packages/cli/lib/services/parser.service.ts`
• `packages/cli/lib/services/sdk.ts`
• `packages/jobs/lib/execution/sync.ts`
• `packages/runner-sdk/models.d.ts`
• `packages/runner-sdk/lib/scripts.ts`
• `packages/runner-sdk/lib/sync.ts`
• `packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap`

</details>

---

### DIFF
--- a/docs-v2/guides/syncs/detecting-deletes.mdx
+++ b/docs-v2/guides/syncs/detecting-deletes.mdx
@@ -120,10 +120,8 @@ export default createSync({
 <Warning>
 **Be careful with exception handling when using** `deleteRecordsFromPreviousExecutions`
 
-Nango only performs deletion detection (the “diff”) if a sync run completes successfully without any uncaught exceptions.
-
 If you’re using `deleteRecordsFromPreviousExecutions`, exception handling becomes critical:
-- If your sync doesn’t fetch the full dataset, but still completes (e.g. you catch and swallow an exception), Nango will attempt the diff on an incomplete dataset.
+- If your sync doesn’t fetch the full dataset, but still call `deleteRecordsFromPreviousExecutions` (e.g. you catch and swallow an exception), Nango will attempt the diff on an incomplete dataset.
 - This leads to false positives, where valid records are mistakenly considered deleted.
 
 ✅ What You Should Do
--- a/docs-v2/guides/syncs/detecting-deletes.mdx
+++ b/docs-v2/guides/syncs/detecting-deletes.mdx
@@ -69,15 +69,10 @@ export default createSync({
 });
 ```
 
-## Detecting deletes in full refresh syncs
+## Detecting deletes in syncs
 
-Full refresh syncs download all records on every run.
-
-Nango can therefore detect removals by computing the diff between two consecutive result sets. Enable this behaviour with the `trackDeletes` flag ([full reference](/reference/integration-configuration#param-track-deletes)).
-
-<Note>
-  `trackDeletes` does not work with incremental syncs because fetching the data incrementally prevents performing a diff and automatically detecting deletions.
-</Note>
+Nango can detect removals by computing the diff between two consecutive result sets.
+Enable this behaviour by calling the `deleteRecordsFromPreviousExecution` function. ([full reference](/reference/scripts#detect-deletions-automatically)).
 
 ### Example full refresh sync with deletion detection
 
@@ -95,7 +90,6 @@ export default createSync({
   description: 'Fetch all help-desk tickets',
   frequency: 'every day',
   syncType: 'full',
-  trackDeletes: true,       // Enables deletion detection for the sync
   endpoints: [{ method: 'GET', path: '/tickets', group: 'Tickets' }],
   models: { Ticket: TicketSchema },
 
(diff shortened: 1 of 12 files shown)


docs-v2/guides/syncs/detecting-deletes.mdx
```diff
@@ -110,45 +104,48 @@ export default createSync({
     for await (const page of tickets) {
       await nango.batchSave(page, 'Ticket');
     }
+
+    // Delete records that don't exist anymore
+    await nango.deleteRecordsFromPreviousExecution('Ticket');
```
bastienbeurier: Curious about whether `Execution` should be singular or plural?

docs-v2/guides/syncs/detecting-deletes.mdx
```diff
@@ -110,45 +104,48 @@ export default createSync({
...
-2. On the next run, Nango compares the new list with the old one.
+1. During each execution, Nango stores the list of record IDs.
+2. When calling `deleteRecordsFromPreviousExecution`, Nango compares the new list with the old one.
 3. Any records missing in the new list are marked as deleted (soft delete). They remain accessible from the Nango cache, but with `record._metadata.deleted === true`.
 
 <Warning>
-**Be careful with exception handling when using** `trackDelete`
+**Be careful with exception handling when using** `deleteRecordsFromPreviousExecution`
 
 Nango only performs deletion detection (the “diff”) if a sync run completes successfully without any uncaught exceptions.
```
bastienbeurier: Is this still true? 

If they call `deleteRecordsFromPreviousExecution`, then the script crashes, the diff won't happen from what I read below?

So even the script succeeds, I imagine the diffing operation will not be executed when the method is called, but after script completion? Probably worth making this explicit if it's not already.

docs-v2/reference/scripts.mdx
```diff
@@ -774,6 +778,33 @@ await nango.batchDelete(githubIssuesToDelete, 'GitHubIssue');
...
 
+### Detect deletions automatically
+
+Automatically detects and marks records as deleted by comparing the current sync execution with the previous one.
+
+```js
+await nango.deleteRecordsFromPreviousExecution('ModelName');
+```
+
+This function should be called at the end of your sync execution, after all records have been saved with `batchSave`. Nango will compare the records saved in the current execution with those from the previous execution and mark any missing records as deleted.
```
bastienbeurier, line 789: What happens if `deleteRecordsFromPreviousExecution` is called before some calls to `batchSave` for the same model?

bastienbeurier (commented):
@khaliqgant we should create a task to migrate all templates to this (vs. track_delete) please.

bodinsamuel (approved):

docs-v2/guides/syncs/detecting-deletes.mdx
TBonnin: good point. Currently track_deletes is on or off so it always only delete from the previous execution but in a world where this function is called dynamically it would be more accurate to use the plural form. I will change

docs-v2/guides/syncs/detecting-deletes.mdx
TBonnin: you are right. This is not true anymore. I will fix

docs-v2/reference/scripts.mdx
TBonnin: It is prevented at compilation time: https://github.com/NangoHQ/nango/blob/master/packages/cli/lib/zeroYaml/compile.ts#L254-L260

If it was happening though, it would mark as deleted some records than potentially still exist.. They would be re-instated by following batchSave operations but during a short time period of time they would show as deleted.

docs-v2/guides/syncs/detecting-deletes.mdx
TBonnin: https://github.com/NangoHQ/nango/pull/4633/commits/35b25e7ea5eacbf9c7461dd06840fd0c7d853451

docs-v2/guides/syncs/detecting-deletes.mdx
TBonnin: https://github.com/NangoHQ/nango/pull/4633/commits/07a9b4ff3d25965266c59f74978fcd4d29cdaa23

docs-v2/guides/syncs/detecting-deletes.mdx
```diff
@@ -69,15 +69,10 @@ export default createSync({
 });
 ```
 
-## Detecting deletes in full refresh syncs
```
bastienbeurier, line 72: @TBonnin just realized the "Full refresh" has been deleted here.

PR #4637, opened 2025-09-10
fix: Revert "fix(paginate): throw error if `response_path` isn't present on the first page"

hassan254-prog:
Reverts NangoHQ/nango#4587


hassan254-prog:
It’s breaking for integrations that don’t return an empty array, i.e quickbooks cdc.

bodinsamuel (approved):

PR #4639, opened 2025-09-11
feat: store records size in bytes

TBonnin:
Storing the size in bytes of the records for each connection/model in the `record_counts` table

Note: the reported size is the compressed size reported by pg of the encrypted payload (with iv and authTag). So not the size as seen by the customer but the size taken on disk from the db point of view

---

**Store Record Size in Bytes per Connection/Model**

This PR introduces tracking and storage of the compressed size in bytes of records for each connection and model in the `record_counts` table. The change adapts all critical layers-model logic, API/service, types, and integration tests-to support and expose the new `size_bytes` field. Compressed size, as reported by PostgreSQL (`pg_column_size`), is now calculated for each upsert, update, and delete operation, and stored/update in the database alongside the existing record counts. All downstream consumers and statistics APIs are updated to work with this extended metric.

<details>

• Added `size_bytes` column (bigint, NOT NULL, default 0) to `record_counts` via migration `packages/records/lib/db/migrations/20250910200700_record_counts_size_bytes.ts`
• Updated `RecordCount` type in `packages/records/lib/types.ts` to include `size_bytes`
• Implemented calculation of payload size using `pg_column_size(json)` during record upsert and update in `packages/records/lib/models/records.ts`
• Tracked `deltaSizeInBytes` changes in transactional operations, updating/incrementing `size_bytes` in `record_counts` accordingly
• Replaced `getRecordCountsByModel` APIs/usages with `getRecordStatsByModel`, now returning both `count` and `size_bytes`
• Updated orchestrator client, sync flow management, and controller to pass and utilize the new statistics signature (`packages/shared/lib/clients/orchestrator.ts`, `packages/shared/lib/services/sync/manager.service.ts`, `packages/server/lib/controllers/sync.controller.ts`)
• Extended integration tests in `packages/records/lib/models/records.integration.test.ts` to validate correctness of `size_bytes` in key scenarios (insert, update, delete, soft delete, batch/upsert concurrency)
• All API/service layers now expect and populate model statistics including size in bytes

</details>

<details>

• `record_counts` table and migration logic
• `RecordCount` type definitions in `packages/records/lib/types.ts`
• Upsert, update, and delete logic in `packages/records/lib/models/records.ts`
• Integration tests in `packages/records/lib/models/records.integration.test.ts`
• Service and controller code relying on model statistics
• Orchestrator client/service interface and consumers

</details>

---

### DIFF
--- a/packages/records/lib/db/migrations/20250910200700_record_counts_size_bytes.ts
+++ b/packages/records/lib/db/migrations/20250910200700_record_counts_size_bytes.ts
@@ -1,0 +1,14 @@
+import type { Knex } from 'knex';
+
+const TABLE = 'record_counts';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.transaction(async (trx) => {
+        await trx.raw(`
+            ALTER TABLE "${TABLE}"
+            ADD COLUMN IF NOT EXISTS size_bytes bigint NOT NULL DEFAULT 0;
+        `);
+    });
+}
+
+export async function down(_knex: Knex): Promise<void> {}
--- a/packages/records/lib/models/records.integration.test.ts
+++ b/packages/records/lib/models/records.integration.test.ts
@@ -9,7 +9,7 @@ import { formatRecords } from '../helpers/format.js';
 import * as Records from '../models/records.js';
 
 import type { FormattedRecord, UnencryptedRecordData, UpsertSummary } from '../types.js';
-import type { MergingStrategy } from '@nangohq/types';
+import type { MergingStrategy, Result } from '@nangohq/types';
 
 describe('Records service', () => {
     beforeAll(async () => {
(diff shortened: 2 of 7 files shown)


packages/records/lib/db/migrations/20250910200700_record_counts_size_bytes.ts
```diff
@@ -0,0 +1,14 @@
...
+
+const TABLE = 'record_counts';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.transaction(async (trx) => {
+        await trx.raw(`
+            ALTER TABLE "${TABLE}"
+            ADD COLUMN IF NOT EXISTS size_bytes bigint NOT NULL DEFAULT 0;
+        `);
+    });
```
TBonnin, line 11: I will make sure to deploy server first so the migration run before persist is deployed

I will need to run a one-off script to populate the values. It would be way to slow to do that as part of the migration

rossmcewan (approved):

PR #4650, opened 2025-09-12
feat: max concurrency per environment for function executions

TBonnin:
We currently only have a global concurrency limit at the processor level that limit how many functions are executed at the same time. It is nice to protect our infra but if a large number of tasks is queued for the same environment they are all dispatched to a single runner that cannot handle the load.
The orchestrator has a mechanism to limit how many tasks for a specific group can be executed at the same time. We currently only use to to limit the concurrency of async actions per environment. I propose to set a upper-limit of how many functions can be executed per environment and type of functions to avoid dispatching tons of functions to a single runner at the same time.
This commit introduce global limit per environment and per function types but a possible improvement would be to have dedicated values in plan (the same way we are doing for the API rate-limit) so different customers can have different limits

 

---

**Add Per-Environment Max Concurrency Limits for Function Executions**

This PR introduces configurable maximum concurrency limits per environment and function type (action, webhook, on-event script, sync) for function executions. Previously, only a global processor-level concurrency limit was enforced, which could cause overloading of individual runners if many tasks from one environment were queued. The new mechanism adds environment-scoped caps and exposes configuration via new environment variables, ensuring that no environment can overwhelm processing resources. Relevant orchestrator, service, and handler logic is updated to utilize and forward the per-environment concurrency limit values. Some minor improvements, such as increasing timeouts for webhook execution scheduling, are included.

<details>

• Added environment variables: `ACTION_ENVIRONMENT_MAX_CONCURRENCY`, `WEBHOOK_ENVIRONMENT_MAX_CONCURRENCY`, `ON_EVENT_ENVIRONMENT_MAX_CONCURRENCY`, `SYNC_ENVIRONMENT_MAX_CONCURRENCY`, each with sensible default values
• Forwarded `maxConcurrency` values from orchestrator API entrypoints through to execution logic for all function types
• Updated group key generation logic for orchestrator tasks to ensure per-environment separation/limits
• Injected `maxConcurrency` into all relevant calls for actions, webhooks, and on-event scripts (including controller and hook logic)
• Increased the webhook execution `createdToStarted` timeout from 30 seconds to 5 minutes to avoid premature task expiration for queued webhooks
• Expanded the environment schema in `packages/utils/lib/environment/parse.ts` to validate and provide default values for the new variables

</details>

<details>

• `packages/shared/lib/clients/orchestrator.ts`
• `packages/server/lib/controllers/action/postTriggerAction.ts`
• `packages/server/lib/controllers/mcp/server.ts`
• `packages/server/lib/hooks/connection/on/post-connection-creation.ts`
• `packages/server/lib/hooks/connection/on/pre-connection-deletion.ts`
• `packages/server/lib/hooks/connection/on/validate-connection.ts`
• `packages/server/lib/webhook/internal-nango.ts`
• `packages/orchestrator/lib/clients/client.ts`

</details>

---

### DIFF
--- a/packages/orchestrator/lib/clients/client.ts
+++ b/packages/orchestrator/lib/clients/client.ts
@@ -259,7 +259,7 @@ export class OrchestratorClient {
             ...rest,
             retry: { count: 0, max: 0 },
             timeoutSettingsInSecs: {
-                createdToStarted: 30,
+                createdToStarted: 5 * 60,
                 startedToCompleted: 60 * 60,
                 heartbeat: 5 * 60
             },
--- a/packages/utils/lib/environment/parse.ts
+++ b/packages/utils/lib/environment/parse.ts
@@ -76,6 +76,10 @@ export const ENVS = z.object({
     ACTION_PROCESSOR_MAX_CONCURRENCY: z.coerce.number().optional().default(200),
     WEBHOOK_PROCESSOR_MAX_CONCURRENCY: z.coerce.number().optional().default(200),
     ONEVENT_PROCESSOR_MAX_CONCURRENCY: z.coerce.number().optional().default(50),
+    SYNC_ENVIRONMENT_MAX_CONCURRENCY: z.coerce.number().optional().default(100),
+    ACTION_ENVIRONMENT_MAX_CONCURRENCY: z.coerce.number().optional().default(100),
+    WEBHOOK_ENVIRONMENT_MAX_CONCURRENCY: z.coerce.number().optional().default(50),
+    ON_EVENT_ENVIRONMENT_MAX_CONCURRENCY: z.coerce.number().optional().default(50),
 
     // Runner
     RUNNER_TYPE: z.enum(['LOCAL', 'REMOTE', 'RENDER', 'KUBERNETES']).default('LOCAL'),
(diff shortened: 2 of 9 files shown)


packages/orchestrator/lib/clients/client.ts
```diff
@@ -259,7 +259,7 @@ export class OrchestratorClient {
             ...rest,
             retry: { count: 0, max: 0 },
             timeoutSettingsInSecs: {
-                createdToStarted: 30,
+                createdToStarted: 5 * 60,
```
TBonnin, line 262: being a bit more generous so queued webhooks are not expired too fast

packages/utils/lib/environment/parse.ts
```diff
@@ -76,6 +76,10 @@ export const ENVS = z.object({
     ACTION_PROCESSOR_MAX_CONCURRENCY: z.coerce.number().optional().default(200),
     WEBHOOK_PROCESSOR_MAX_CONCURRENCY: z.coerce.number().optional().default(200),
     ONEVENT_PROCESSOR_MAX_CONCURRENCY: z.coerce.number().optional().default(50),
+    SYNC_ENVIRONMENT_MAX_CONCURRENCY: z.coerce.number().optional().default(100),
+    ACTION_ENVIRONMENT_MAX_CONCURRENCY: z.coerce.number().optional().default(100),
+    WEBHOOK_ENVIRONMENT_MAX_CONCURRENCY: z.coerce.number().optional().default(50),
+    ON_EVENT_ENVIRONMENT_MAX_CONCURRENCY: z.coerce.number().optional().default(50),
```
TBonnin, line 82: Values are still pretty generous based on current usage. We might need to tweak them down a bit but at least we now have an upper limit per environment

rossmcewan (approved):

kaposke (approved):

PR #4692, opened 2025-09-18
feat(credentials): keep credentials until they are hard deleted at 30 days

khaliqgant:
---

**Retain Credentials Until Hard Deletion for Soft-Deleted Connections**

This PR modifies the `deleteConnection` logic in `packages/shared/lib/services/connection.service.ts` so that sensitive fields (`credentials`, `credentials_iv`, `credentials_tag`) are no longer wiped during a soft delete. Instead, these fields are now only deleted during hard deletion, usually performed 30 days after the soft delete. The change ensures that credentials remain recoverable during the soft-deleted period, aligning data retention with the system's deletion policy.

<details>

• Updated the `.update` call in the `deleteConnection` method to remove wiping of `credentials`, `credentials_iv`, and `credentials_tag` fields.
• Now, only the `deleted` flag is set to `true` and `deleted_at` timestamp is set during a soft delete; credentials remain intact.
• The data-wiping of credentials for soft-deleted connections is deferred to the hard deletion phase.

</details>

<details>

• `deleteConnection` method in `packages/shared/lib/services/connection.service.ts`
• Soft-delete mechanics and state of credentials for records in the `_nango_connections` database table

</details>

---

### DIFF
--- a/packages/shared/lib/services/connection.service.ts
+++ b/packages/shared/lib/services/connection.service.ts
@@ -841,7 +841,7 @@ class ConnectionService {
                 environment_id: environmentId,
                 deleted: false
             })
-            .update({ deleted: true, credentials: {}, credentials_iv: null, credentials_tag: null, deleted_at: new Date() });
+            .update({ deleted: true, deleted_at: new Date() });
 
         // TODO: move the following side effects to a post deletion hook
         // so we can remove the orchestrator dependencies


TBonnin (approved):
could it cause compliance issue not to delete credentials when customer hits delete?

bodinsamuel:
> could it cause compliance issue not to delete credentials when customer hits delete?

I don't think so, they usually mandate encryption and deletion when requested, with a tolerated grace period depending on law requirements and legitimate interest.

bodinsamuel (approved):

PR #4704, opened 2025-09-19
fix(fleet): race condition when node being created

TBonnin:
multiple execution of getRunningNode could enter the if condition and create node (even if the creation itself is protected by a lock. The solution is the classic lock double check. It is an extra query when a node needs to be created but that doesn't happen very often, while keeping the more common scenario (getting a running node) fast and lock-free

---

**Fix Race Condition in Fleet Node Creation**

This PR resolves a race condition in the `Fleet.getRunningNode` method where concurrent invocations could lead to multiple nodes being created for the same `routingId`. The root cause was that, despite locking around node creation, multiple threads could still pass an outer existence check before the lock was acquired. The refactor introduces a double-checked locking pattern-revalidating node existence within the lock before creating a new node. Node search logic is extracted into a helper (`searchNode`) that is reused both before and after acquiring the lock. An integration test has been added to ensure that, even under heavy concurrency (100 simultaneous calls), only one node is created.

<details>

• Refactored `Fleet.getRunningNode()` to use double-checked locking before creating a node.
• Extracted node search logic to new helper function `searchNode`.
• Added an integration test in `fleet.integration.test.ts` to verify single node creation under concurrent `getRunningNode` calls.
• Added explicit import of `setTimeout` for type safety and clarity.

</details>

<details>

• `packages/fleet/lib/fleet.ts` (main node management and locking logic)
• `packages/fleet/lib/fleet.integration.test.ts` (integration test coverage for concurrency)

</details>

---

### DIFF
--- a/packages/fleet/lib/fleet.ts
+++ b/packages/fleet/lib/fleet.ts
@@ -17,6 +17,7 @@ import type { NodeProvider } from './node-providers/node_provider.js';
 import type { Node, NodeConfigOverride } from './types.js';
 import type { Deployment, RoutingId } from '@nangohq/types';
 import type { Result } from '@nangohq/utils';
+import type knex from 'knex';
 
 const defaultDbUrl =
     envs.RUNNERS_DATABASE_URL ||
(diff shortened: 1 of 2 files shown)


packages/fleet/lib/fleet.ts
```diff
@@ -81,36 +82,49 @@ export class Fleet {
...
                     fn: async (trx): Promise<Result<Node>> => {
+                        // double check node was not created while acquiring lock
+                        const node = await searchNode(trx);
+                        if (node.isErr()) {
+                            return Err(node.error);
+                        }
+                        if (node.value) {
+                            return Ok(node.value);
+                        }
+
```
TBonnin, line 127: This is the fix. Double checking after acquiring the lock that the node was not created while waiting to acquire the lock
The rest is reorganizing and optimizing the code

rossmcewan (approved):

PR #4706, opened 2025-09-19
fix(slack): Handle Slack returning 200 when posting message

TBonnin:
Slack API returns 200 with `{ ok: false, error: ... }` when request fails We handle it when joining the channel but not when posting the message. Before I refactored it earlier this week we will always try to join the channel so error will be caught there and fail silently for the customer. After my changes we post first and only handled the not_in_channel error.

 

---

**Improve Slack API Error Handling in Notification Service**

This pull request refines how the Slack notification service processes responses from the Slack API, specifically addressing the scenario where a message post returns HTTP 200 with `{ ok: false, error: ... }` in the payload, indicating an application-level error rather than a true success. Earlier error handling covered this for channel-join operations but did not correctly surface such failures when posting messages, leading to silent delivery issues. The implementation now introduces explicit checks for missing or unsuccessful responses when posting messages and during channel joins, ensuring all errors are detected and handled uniformly. Additionally, resend-on-join logic is simplified and some redundant code is removed.

<details>

• Added explicit verification for `slackMessage.value.data` presence and correctness when posting messages and joining channels in `packages/shared/lib/services/notification/slack.service.ts`.
• Returns structured errors for all Slack API responses with `ok: false`, not just `not_in_channel` errors.
• Removed redundant error logging for missing Slack data, using early error returns instead.
• Simplified error handling and message resend logic for the `not_in_channel` case, reducing unnecessary nesting.

</details>

<details>

• `packages/shared/lib/services/notification/slack.service.ts`
• `proxySlackMessage` function and related error handling code

</details>

---

### DIFF
--- a/packages/shared/lib/services/notification/slack.service.ts
+++ b/packages/shared/lib/services/notification/slack.service.ts
@@ -720,6 +720,9 @@ export class SlackService {
             if (join.isErr()) {
                 return Err(new Error('slack_join_channel_failed', { cause: join.error }));
             }
+            if (!join.value.data) {
+                return Err(new Error('slack_join_no_response'));
+            }
             if (!join.value.data.ok) {
                 return Err(join.value.data.error);
             }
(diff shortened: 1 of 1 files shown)


bodinsamuel (approved):

khaliqgant (approved):
Thanks!

packages/shared/lib/services/notification/slack.service.ts
```diff
@@ -805,18 +807,14 @@ export class SlackService {
...
                     return Err(join.error);
                 }
-                const resend = await sendMessage();
-                if (resend.isErr()) {
-                    return Err(resend.error);
-                }
-                return resend;
+                return sendMessage();
             }
-            return Err(send.error);
```
TBonnin, line 819: unrelated but checking `isErr` to resend the error is unecessary 😀

PR #4711, opened 2025-09-22
fix(webhook): correctly send auth override webhook

bodinsamuel:
## Changes

- fix(webhook): correctly send auth override webhook
The logic is a bit convoluted but it's now working

---

**Correct Webhook Dispatch for `auth_override` Operations**

This PR updates the webhook handling logic to ensure that `auth_override` authentication operations trigger the correct webhook events. The changes add support for `auth_override` as a specific webhook event type, refactoring operation-to-type mapping and ensuring it propagates through the relevant webhook functions and their call sites.

<details>

• Extended the `shouldSend` function in `packages/webhooks/lib/utils.ts` to recognize `auth_override` as a valid event type.
• Introduced the `AUTH_OPERATION_TO_TYPE` constant in `packages/webhooks/lib/auth.ts` for consistent mapping of authentication operations (`creation`, `override`, `refresh`) to webhook type strings.
• Refactored the `sendAuth` function in `packages/webhooks/lib/auth.ts` to use the new mapping and ensure proper handling of all supported operation types including `auth_override`.
• Updated destructuring and parameter passing in `connectionCreated` in `packages/server/lib/hooks/hooks.ts` to support dynamic `operation` values instead of hardcoded types.

</details>

<details>

• `packages/webhooks/lib/auth.ts`
• `packages/webhooks/lib/utils.ts`
• `packages/server/lib/hooks/hooks.ts`

</details>

---


packages/webhooks/lib/auth.ts
```diff
@@ -45,11 +45,17 @@ export async function sendAuth({
...
     }
 
-    if (!shouldSend({ success, type: operation === 'creation' ? 'auth_creation' : 'auth_refresh', webhookSettings })) {
+    if (
+        !shouldSend({
+            success,
+            type: operation === 'creation' ? 'auth_creation' : operation === 'override' ? 'auth_override' : 'auth_refresh',
+            webhookSettings
+        })
+    ) {
```
TBonnin: I personally have a hard time parsing double ternary operator. Maybe that's just me but I would prefer a switch or a Map 
```
const getAuthType = (op: string) => {
  switch (op) {
    case 'creation': return 'auth_creation';
    case 'override': return 'auth_override';
    default: return 'auth_refresh';
  }
};

...
type: getAuthType(operation)
...
```

TBonnin (approved):

packages/webhooks/lib/auth.ts
bodinsamuel: or a map yes

bodinsamuel: (tbf this whole rewrite is useless, it should be drop)

PR #4712, opened 2025-09-22
fix(endUsers): duplicate all relationships

bodinsamuel:
## Changes

- duplicate all relationships
Now that we have something working and people are actively using it, we need to duplicate all `endUsers` that were tied to multiple connections.

I ran it locally and it worked. I would appreciate a test on your side, just in case ☺️ 

---

**Database Migration: One-to-One Mapping for `end_users` and `_nango_connections`**

This PR introduces a one-way, irreversible database migration to normalize the relationship between the `end_users` and `_nango_connections` tables. Currently, a single `end_user` can be tied to multiple `_nango_connections`, creating a many-to-many mapping. The migration script iterates through all `end_users` associated with more than one `_nango_connections` record. For each such user, it creates a copy of the `end_user` for every additional connection and updates the corresponding `_nango_connections` to point to these new records. The migration runs non-transactionally, prioritizing simplicity over strict atomicity and is intended to be re-executed as needed.

<details>

• Added migration script `packages/database/lib/migrations/20250922111100_end_users_dup.cjs`
• Implemented `exports.up` function to duplicate `end_users` for each associated `_nango_connections` record beyond the first
• Updated `_nango_connections` to point to newly created `end_users`
• Set `exports.config = { transaction: false }`, disabling transaction wrapping for the migration
• Provided an empty, non-functional `exports.down` (irreversible migration)
• Removed batching timing/wait logic after review discussion

</details>

<details>

• `packages/database/lib/migrations/20250922111100_end_users_dup.cjs`
• `end_users` table
• `_nango_connections` table

</details>

---


packages/database/lib/migrations/20250922111100_end_users_dup.cjs
```diff
@@ -0,0 +1,55 @@
...
+            break;
+        }
+
+        // Grab all connections for the end user that needs to be duplicated
+        const connections = await knex.raw(`SELECT
+	row_to_json(end_users.*) as end_user, _nango_connections.id as connection_id
+FROM
+	end_users
+	JOIN _nango_connections ON end_users.id = end_users.id
+    WHERE end_users.id IN (${endUsers.rows.map((endUser) => endUser.id).join(',')})`);
```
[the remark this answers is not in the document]
bodinsamuel: thanks but no

packages/database/lib/migrations/20250922111100_end_users_dup.cjs
```diff
@@ -0,0 +1,55 @@
...
+        const endUsers = await knex.raw(`SELECT
+			end_users.id,
+			COUNT(1) as total
+		FROM
+			end_users
+			JOIN _nango_connections AS nc ON nc.end_user_id = end_users.id
+		GROUP BY
+			end_users.id
+		HAVING
+			COUNT(1) > 1
```
kaposke:
```suggestion
			total > 1
```

packages/database/lib/migrations/20250922111100_end_users_dup.cjs
```diff
@@ -0,0 +1,55 @@
...
+		LIMIT 1`);
```
kaposke: Is it intentionally 1 by 1? This could be batches of like 100 or 1000, and forEach below.

packages/database/lib/migrations/20250922111100_end_users_dup.cjs
```diff
@@ -0,0 +1,55 @@
...
+            break;
+        }
+
+        // Grab all connections for the end user that needs to be duplicated
+        const connections = await knex.raw(`SELECT
+	row_to_json(end_users.*) as end_user, _nango_connections.id as connection_id
+FROM
+	end_users
+	JOIN _nango_connections ON end_users.id = _nango_connections.end_user_id
+    WHERE end_users.id IN (${endUsers.rows.map((endUser) => endUser.id).join(',')})`);
```
kaposke: When batching 1 by 1, this `IN` doesn't make sense because it has a single ID. 
If you increase the batch in the query above, it won't make sense because you want to look for each end_user at a time.

Should this be (after increasing limit above):
```ts
for (const endUser of endUsers) {
  // Grab all connections for the end user that needs to be duplicated
  const connections = await knex.raw(`
    SELECT
	  row_to_json(end_users.*) as end_user, _nango_connections.id as connection_id
    FROM
	  end_users
	JOIN 
      _nango_connections ON end_users.id = _nango_connections.end_user_id
    WHERE 
      end_users.id = ${endUser.id}`);
  // ...
}
```

packages/database/lib/migrations/20250922111100_end_users_dup.cjs
```diff
@@ -0,0 +1,55 @@
...
+
+        console.log('[endUser dup] changing', endUsers.rows[0].total, 'connections, for end user', endUsers.rows[0].id);
+
+        // Insert a new end user for each connection
+        // And update the related connection with the new end_users.id
+        for (const connection of connections.rows) {
```
kaposke: should we make this a regular i++ for, and start at i = 1? Currently will end up with hanging endUsers with no relationships.

packages/database/lib/migrations/20250922111100_end_users_dup.cjs
bodinsamuel: computed values are not available in where and having

packages/database/lib/migrations/20250922111100_end_users_dup.cjs
bodinsamuel: yes, it's intentional to simplify the batching we get all connections tied to a single end user

packages/database/lib/migrations/20250922111100_end_users_dup.cjs
bodinsamuel: yeah okay I can rewrite it like this 👌🏻

packages/database/lib/migrations/20250922111100_end_users_dup.cjs
```diff
@@ -0,0 +1,63 @@
...
+                const connection = connections.rows[i];
+                const inserted = await knex
+                    .from('end_users')
+                    .insert({ ...connection.end_user, id: undefined })
+                    .returning('id');
+                await knex.raw(`UPDATE _nango_connections SET end_user_id = ${inserted[0].id} WHERE id = ${connection.connection_id}`);
+            }
+
+            // Be nice to the database and cpu
+            await wait(100);
```
TBonnin: how many end users do we have in prod and how long is this migration gonna take?

bodinsamuel: there are 60k end users, about 26k to be created. Not really big, not small

TBonnin: are we gonna wait 100ms between every single end users?

bodinsamuel: waiting after each batch of connections, but I could lower the wait indeed

TBonnin: it depends. if the migrations is a couple of seconds long, it is fine, if we are talking minutes it might cause the render deploy to timeout
To be honest if I had been doing it I will have let pg do everything in a single transaction.

kaposke: The first query returns about 7340 connections. If we wait 5ms for each, that's 37 seconds in wait-time, without counting query time. Should we only apply a wait time after bigger batches (more than one connection), or do it differently entirely? Or is 37+query time fast enough?

bodinsamuel: If I use a transaction, it will lock all those connections (for update). I'm not really worried about timing tbh, and it can be re-run until completion. I can add some conditions around the number of impacted connections for sure.

> To be honest if I had been doing it I will have let pg do everything in a single transaction.

I tried, but it's not ideal to write and review. This is simple and easy to debug

bodinsamuel: I'll just remove the wait since it's creating some debate, but would love to know if there is a strong push back so I can finally merge

TBonnin: no push back. I just wanted to make sure the migration could run quickly enough not to timeout the deployment

TBonnin (approved):

packages/database/lib/migrations/20250922111100_end_users_dup.cjs
kaposke: Same, sounds good if it doesn't time-out. One-time heavy operations are ok by me.

PR #4796, opened 2025-10-02
fix(kvstore): use kvstore across the board and allow use of ioredis

rossmcewan:
---

**Refactor `kvstore` to Add Pluggable Redis Backend, Unified API, and App-Wide Usage**

This PR introduces a major refactor to the `kvstore` subsystem. It replaces all direct Redis usages across the application (including middleware, background jobs, pub/sub, metering, and crons) with a unified `KVStore` abstraction layer. The new system supports both `node-redis` and `ioredis` as runtime-selectable Redis clients, determined via the `NANGO_REDIS_CLIENT_LIBRARY` environment variable. Implementation uses an adapter pattern for backend/client selection, supports named/shared Redis client instances, centralizes resource cleanup, and standardizes the public API for KV operations-now including pub/sub support. Unit tests are extensively expanded for the new API, and dependencies for both Redis libraries are pinned and updated. Architectural and style discussions (such as config source and instantiation ergonomics) are partially addressed and some are planned for later iteration.

<details>

• Implement unified `KVStore` abstraction across all Redis-dependent features.
• Add adapter-based backend selection: support for `node-redis` and `ioredis` toggled by `NANGO_REDIS_CLIENT_LIBRARY`.
• Introduce new files: `IORedisStore.ts` and `utils.ts` for client management and utility instantiation.
• Expand `KVStore` and `KVStoreRedis` API to support pub/sub (`publish`, `subscribe`, `unsubscribe`), and scanning.
• Support named Redis clients (multi-instance) and provide shared connection management.
• Replace all direct Redis usages (middleware, jobs, server pub/sub, metering, rate limiting, etc.) with the new `KVStore` interface.
• Enhance in-memory store to provide full feature parity (including event-driven pub/sub) for testing.
• Revise teardown and resource management (e.g., `destroyAll`, client cleanup).
• Refactor and expand all unit tests for coverage of the pluggable backend and multi-client use cases.
• Update dependencies and pin versions for both `redis` and `ioredis` in `package.json` and `package-lock.json`.

</details>

<details>

• `packages/kvstore/lib/` (all core, adapters, tests, utils)
• `packages/server/lib/clients/publisher.client.ts`
• `packages/server/lib/middleware/ratelimit.middleware.ts`
• `packages/account-usage/lib/index.ts`
• `packages/server/lib/server.ts`
• `packages/jobs/lib/app.ts`
• `packages/metering/lib/crons/persistAccountUsage.ts`
• `packages/kvstore/package.json`
• `package-lock.json`

</details>

---


packages/kvstore/package.json
```diff
@@ -15,7 +15,8 @@
     },
     "dependencies": {
         "@nangohq/utils": "file:../utils",
-        "redis": "4.6.13"
+        "redis": "4.6.13",
+        "ioredis": "^5.8.0"
```
TBonnin: let's pin the version like we do for the other one

packages/server/lib/middleware/ratelimit.middleware.ts
```diff
@@ -43,25 +44,7 @@ async function getRateLimiter(size: DBPlan['api_rate_limit_size']) {
...
-        const redisClient = await createClient({
-            url: url,
-            disableOfflineQueue: true,
-            pingInterval: 30_000,
-            socket
-        }).connect();
-        redisClient.on('error', (err) => {
-            logger.error(`Redis (rate-limiter) error: ${err}`);
-        });
+        const redisClient = ((await getKVStore(url)) as KVStoreRedis).getClient();
```
TBonnin: why not calling `getRedis` and only get the client instead of instanciating a KVStore and only get the client?
Also, it is using the url as the name. is it intended?

packages/kvstore/lib/index.ts
```diff
@@ -1,91 +1,100 @@
...
-let redis: RedisClientType | undefined;
-async function getRedis(url: string): Promise<RedisClientType> {
-    if (redis) {
-        return redis;
+const redisClients = new Map<string, RedisClient>();
+export function getRedis(name: string, url: string): RedisClient {
+    if (redisClients.has(name)) {
+        return redisClients.get(name)!;
+    }
+    const clientLibrary = process.env['NANGO_REDIS_CLIENT_LIBRARY'] || 'node-redis';
```
TBonnin: let's not use an env var in a library type package, I know there might still be some places in the code where we used to do that but it is source of errors and env vars cannot be validated unless we import utils. Any config value should be passed as constructor argument

packages/kvstore/lib/index.ts
```diff
@@ -1,91 +1,100 @@
...
+    switch (clientLibrary) {
+        case 'node-redis':
+            redisClients.set(name, getNodeRedis(url));
+            break;
+        case 'ioredis':
+            redisClients.set(name, getIORedis(url));
+            break;
```
TBonnin, line 35: the switch will fall through if `NANGO_REDIS_CLIENT_LIBRARY` env var is set to an unexpected value. Another good reason not use an untyped env var

packages/kvstore/lib/index.ts
```diff
@@ -1,91 +1,100 @@
...
 export { type Lock, Locking } from './Locking.js';
 
 // Those getters can be accessed at any point so we store the promise to avoid race condition
 // Not my best code
-
-let redis: RedisClientType | undefined;
-async function getRedis(url: string): Promise<RedisClientType> {
-    if (redis) {
-        return redis;
+const redisClients = new Map<string, RedisClient>();
```
TBonnin, line 22: the stored promise was there so we could create/reuse a single redis client. if we are gonna have multiple Redis clients we can create a new one every time without worrying about race condition.

What's the overhead of having multiple clients? The only place we really need 2 is the server publisher/subscriber.

packages/kvstore/lib/index.ts
```diff
@@ -1,91 +1,100 @@
...
+        await destroy(name, hard);
     }
-    if (redis) {
-        await redis.disconnect();
+}
+
+export async function destroy(name: string, hard: boolean = false) {
+    if (kvstorePromises.has(name)) {
+        await (await kvstorePromises.get(name)!).destroy();
+        if (hard) {
```
TBonnin, line 74: if hard is false the promise is destroyed but kept in the map

packages/server/lib/middleware/ratelimit.middleware.ts
rossmcewan: - good catch on the URL, that was not intended. 
- as far as constructing the KVStore, I didn't intend to keep the getRedis exported so there was only a single way of constructing the KV store.

packages/kvstore/lib/index.ts
rossmcewan: Yeah, agreed. This code already had a few env vars dotted through it so I went with the lazy option. I'm cleaning it up.

packages/kvstore/lib/index.ts
rossmcewan: The destroy is not the promise - it's the KVStore instance - but it doesn't do anything. This was what it already did.

packages/kvstore/lib/index.ts
rossmcewan: We can still share the redis clients if we give them a shared name - something like default - and then just give the pulish subscribe their own one.

TBonnin: in this case I would make the name optional so you don't have to specify it unless you want a specific one and maybe calling the param `key` or `identifier` to hint at its uniqueness. I am also gonna reiterate but url should be passed as a mandatory param and not rely on an env var that is not validated in this library package. There is already a shared function to get the redis url: https://github.com/NangoHQ/nango/blob/ab5b676c3731d39b691113b3cd3fbcbd20a1115f/packages/shared/lib/utils/utils.ts#L56
Something like
```
getRedis(opts: {
   url: string;
   identifier?: string;
   ...
}) 
// or 
getRedis(url: string, opts: {
   identifier?: string;
   ...
}) 
```
Same signature for `getKVStore` so it can be used as
```
const kvstore = getKVStore({ url: getRedisUrl()}) // default client
// or
const url = getRedisUrl();
const pub = getKVStore({ identifier: "publisher", url })
const sub = getKVStore({ identifer: "subscriber", url})
```

packages/server/lib/middleware/ratelimit.middleware.ts
```diff
@@ -43,25 +44,7 @@ async function getRateLimiter(size: DBPlan['api_rate_limit_size']) {
...
-        const redisClient = await createClient({
-            url: url,
-            disableOfflineQueue: true,
-            pingInterval: 30_000,
-            socket
-        }).connect();
-        redisClient.on('error', (err) => {
-            logger.error(`Redis (rate-limiter) error: ${err}`);
-        });
+        const redisClient = ((await getKVStore('default', { url })) as KVStoreRedis).getClient();
```
TBonnin: why not calling getRedis directly instead of instanciating a KVStore just to use its client?

packages/kvstore/lib/index.ts
```diff
@@ -1,115 +1,127 @@
...
 // Not my best code
-
-let redis: RedisClientType | undefined;
-async function getRedis(url: string): Promise<RedisClientType> {
-    if (redis) {
-        return redis;
+const redisClients = new Map<string, RedisClient>();
+export function getRedis(name: string, options: KVStoreOptions): RedisClient {
+    if (redisClients.has(name)) {
+        return redisClients.get(name)!;
```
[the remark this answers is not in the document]
TBonnin: are you sure propel? what race condition are you talking about? 😉

packages/kvstore/lib/index.ts
TBonnin: before the promise was unset `kvstorePromise = undefined` so it couldn't be accessed. it is not the case here

rossmcewan: It didn't used to. I made a change last week that did that because tests were unable to pass without promises being reset. Now that same behavior is specified with the 'hard=true' - that deletes the promise from the map.

packages/server/lib/middleware/ratelimit.middleware.ts
rossmcewan: Thought I answered this. I didn't want to expose the getRedis directly so there was only a single entry point to KV functionality.

packages/kvstore/lib/index.ts
rossmcewan: The name one is fair - happy to change that. 
The URL one - getRedisUrl also relies on unvalidated env vars too. Not sure I agree with this one.

TBonnin (approved):

PR #4800, opened 2025-10-02
fix: modify task processor tests to use wildcard groupKeyPattern

TBonnin:
we recently broke the dequeue long polling by adding wildcard to the groupKeyPattern.
This commit modifies the processor tests to use a wildcard groupKeyPattern, mimicing real processor. Test now will now fail if the task event logic around discarding wildcard is modified because it will timeout before the dequeue times out

It doesn't address all the possible ways dequeue long polling can be broken but it will detect the same kind of bug we encounter a few weeks ago.

---

**Update `OrchestratorProcessor` Tests to Use Wildcard `groupKeyPattern`**

This pull request updates the `processN` test utility in `packages/orchestrator/lib/clients/processor.integration.test.ts` to use a wildcard pattern for the `groupKeyPattern` property, matching the real processor behavior. The test logic is adjusted to prevent silent test passes if the task event handling for wildcard patterns is broken, by introducing a stricter check on the processed tasks and a shorter timeout.

<details>

• Changed `groupKeyPattern` from `groupKey` to ``${groupKey}*`` in the processor test utility.
• Inserted a brief delay with `await setTimeout(100)` after starting the processor to improve timing reliability.
• Adjusted the `timeout` value for task processing from 3,000 ms to 1,000 ms to fail the test faster if the wildcard logic is incorrect.
• Refined the processing logic to require exactly `n` tasks and to check all tasks with `tasks.length == n && tasks.every(waitUntil)`.

</details>

<details>

• `packages/orchestrator/lib/clients/processor.integration.test.ts`

</details>

---

### DIFF
--- a/packages/orchestrator/lib/clients/processor.integration.test.ts
+++ b/packages/orchestrator/lib/clients/processor.integration.test.ts
@@ -71,10 +71,11 @@ async function processN({
     const processor = new OrchestratorProcessor({
         handler,
         orchestratorClient,
-        groupKeyPattern: groupKey,
+        groupKeyPattern: `${groupKey}*`, // using wildcard like the real processor
         maxConcurrency: n
     });
     processor.start();
+    await setTimeout(100);
 
     for (let i = 0; i < n; i++) {
         await immediateTask({ groupKey });
@@ -82,11 +83,11 @@ async function processN({
 
     let processed = false;
     const start = Date.now();
-    const timeout = 3_000;
+    const timeout = 1_000;
     while (!processed) {
         await setTimeout(100);
         const tasks = (await scheduler.searchTasks({ groupKey })).unwrap();
-        processed = tasks.every(waitUntil);
+        processed = tasks.length == n && tasks.every(waitUntil);
         if (!processed && Date.now() - start > timeout) {
             throw new Error(`Timeout: expected ${n} tasks to be processed, but tasks are still in states: ${tasks.map((task) => task.state).join(', ')}`);
         }


packages/orchestrator/lib/clients/processor.integration.test.ts
```diff
@@ -71,22 +71,23 @@ async function processN({
     const processor = new OrchestratorProcessor({
         handler,
         orchestratorClient,
-        groupKeyPattern: groupKey,
+        groupKeyPattern: `${groupKey}*`, // using wildcard like the real processor
         maxConcurrency: n
     });
     processor.start();
+    await setTimeout(100);
```
TBonnin, line 78: waiting for the processor to be already dequeueing

packages/orchestrator/lib/clients/processor.integration.test.ts
```diff
@@ -71,22 +71,23 @@ async function processN({
...
 
     for (let i = 0; i < n; i++) {
         await immediateTask({ groupKey });
     }
 
     let processed = false;
     const start = Date.now();
-    const timeout = 3_000;
+    const timeout = 1_000;
```
TBonnin, line 86: we actually don't need to wait this long, 1s is more than enough

packages/orchestrator/lib/clients/processor.integration.test.ts
```diff
@@ -71,22 +71,23 @@ async function processN({
...
     while (!processed) {
         await setTimeout(100);
         const tasks = (await scheduler.searchTasks({ groupKey })).unwrap();
-        processed = tasks.every(waitUntil);
+        processed = tasks.length == n && tasks.every(waitUntil);
```
TBonnin, line 90: not absolutely required but if somehow `searchTasks` return an empty array, `every` would return true, which doesn't mean the tasks were processed

rossmcewan (approved):

PR #4871, opened 2025-10-16
chore: add plan.webhook_forwards_max column

TBonnin:
I misunderstood one of the capping metric. `external_webhooks` should actually be `webhook_forwards`.
Renaming the column in plan. Next PR will modify the code


kaposke (approved):
Note that the other migration (removing the old key) is in the next PR

PR #4874, opened 2025-10-16
fix(runner): annotate pods to prevent disruption

rossmcewan:

### DIFF
--- a/packages/jobs/lib/runner/kubernetes.ts
+++ b/packages/jobs/lib/runner/kubernetes.ts
@@ -171,7 +171,8 @@ class Kubernetes {
                 template: {
                     metadata: {
                         annotations: {
-                            [`ad.datadoghq.com/runner.logs`]: `[{"source":"nango","service":"${name}"}]`
+                            [`ad.datadoghq.com/runner.logs`]: `[{"source":"nango","service":"${name}"}]`,
+                            ['karpenter.sh/do-not-disrupt']: 'true'
                         },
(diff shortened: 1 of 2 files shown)


packages/jobs/lib/runner/kubernetes.ts
```diff
@@ -171,7 +171,8 @@ class Kubernetes {
                 template: {
                     metadata: {
                         annotations: {
-                            [`ad.datadoghq.com/runner.logs`]: `[{"source":"nango","service":"${name}"}]`
+                            [`ad.datadoghq.com/runner.logs`]: `[{"source":"nango","service":"${name}"}]`,
+                            ['karpenter.sh/do-not-disrupt']: `${envs.RUNNER_DO_NOT_DISRUPT}`
```
TBonnin, line 175: does it need to be an env var and be always true?
if yes, what about `RUNNER_K8S_DO_NOT_DISRUPT` to make it even clearer?

TBonnin (approved):

packages/jobs/lib/runner/kubernetes.ts
rossmcewan: I want it to be an env var because staging and development shouldn't use this - they only have a couple of nodes so it would pretty much prevent any maintenance operations happening.

I also didn't make it kubernetes-specific - that's our only implementation for now - but would hold for other container orchestrators like ECS too.

PR #4890, opened 2025-10-20
fix: orb_subcribed_at backfill

kaposke:
Previous `start_date` included `free` subscription start date, didn't account for plan changes.
The correct way to get the plan change dates, is to get the `/subscriptions/:id/schedule` endpoint. It returns the history of plans in a subscription. I'm getting the earliest paid plan and using it as the new date.

Dryran and checked 10 accounts, they were all correct. Total rows to update: 76.

---

**Correct Backfill of `orb_subscribed_at` Dates Using Plan Change History**

This PR updates the logic for backfilling the `orb_subscribed_at` field in the `nango.plans` table. Previously, the script used the `start_date` of Orb subscriptions, which included free plans and did not correctly account for upgrades or plan transitions. The new approach retrieves the subscription schedule from the `/subscriptions/:id/schedule` Orb API endpoint, analyzes the plan history, and determines the correct date when an account first became a paying (non-free) subscriber, specifically targeting upgrade transitions from free to paid or, if that isn't found, the earliest paid plan. The script is intended as a one-off for ~76 subscriptions, and has been tested in dry-run mode on a sample of accounts.

<details>

• Removed use of Orb subscription summary and switched to querying `/subscriptions/:id/schedule` for full plan history.
• Introduced interface `OrbSubscriptionScheduleItem` to model plan schedule entries.
• Implemented the `fetchOrbSubscriptionSchedule` function to retrieve historical plan data.
• Added the `getLastUpgradeDate` function to determine the correct paid plan activation date, handling free-to-paid transitions.
• Refactored the main backfill logic to use the schedule-based upgrade date instead of the simple `start_date`.
• Adjusted database queries to use the correct table and updated output/logging to reflect new date logic.
• Dry-run mode remains for safe trial runs.

</details>

<details>

• scripts/one-off/backfill-orb-subscription-dates/backfill.ts

</details>

---

### DIFF
--- a/scripts/one-off/backfill-orb-subscription-dates/backfill.ts
+++ b/scripts/one-off/backfill-orb-subscription-dates/backfill.ts
@@ -19,7 +19,6 @@ const db = knex({
     client: 'pg',
     connection: {
         connectionString: DB_CONNECTION_STRING,
-        ssl: 'no-verify',
         statement_timeout: 60000
     },
     pool: {
@@ -27,8 +26,18 @@ const db = knex({
         max: 10
     }
 });
-async function fetchOrbSubscription(subscriptionId: string) {
-    const response = await fetch(`https://api.withorb.com/v1/subscriptions/${subscriptionId}`, {
+
+interface OrbSubscriptionScheduleItem {
+    start_date: string;
+    end_date: string;
+    created_at: string;
+    plan: {
+        external_plan_id: string;
+    };
+}
+
+async function fetchOrbSubscriptionSchedule(subscriptionId: string): Promise<OrbSubscriptionScheduleItem[]> {
+    const response = await fetch(`https://api.withorb.com/v1/subscriptions/${subscriptionId}/schedule`, {
         headers: {
             Authorization: `Bearer ${ORB_API_KEY}`
         }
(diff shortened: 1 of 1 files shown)


scripts/one-off/backfill-orb-subscription-dates/backfill.ts
```diff
@@ -38,11 +47,29 @@ async function fetchOrbSubscription(subscriptionId: string) {
         throw new Error(`Failed to fetch Orb subscription: ${response.statusText}`);
     }
 
-    return response.json();
+    return (await response.json()).data as OrbSubscriptionScheduleItem[];
+}
+
+function getFirstPaidPlanDate(subscriptionSchedules: OrbSubscriptionScheduleItem[]): Date {
```
TBonnin: should it be the last paid plan. what if a customer upgraded multiple time or upgraded>downgraded>upgraded?

TBonnin (approved):

scripts/one-off/backfill-orb-subscription-dates/backfill.ts
kaposke: So, we want the start of the paying subscription. Ideally we do: date of latest non-free plan right after a free plan (so last upgrade). But I though complicating this rule would make for higher risk and lower confidence, so I thought this simplification was good enough. Don't think we have many customers who became free then paying again? Do you think it's worth the complication?

kaposke: @TBonnin Ended up doing the complication. Re-tested 10 users, including some that upgraded, then downgraded, then upgraded again, and they worked correctly, but can you please review the `getLatestUpgrade` function?

PR #4895, opened 2025-10-21
fix: pagination memory leak investigation

TBonnin:
investigating a potential memory leak in long running pagination


rossmcewan (approved):

TBonnin:
closing. not the cause of the memory leak

PR #4897, opened 2025-10-21
feat: new plan definitions (proposal)

kaposke:
> [!NOTE]
> This is my [McDonald's idea](https://jonbell.medium.com/mcdonalds-theory-9216e1c9da7d) to get things started. It's not well tested nor intended to be final.

We previously only differentiated (orb) plans by distinct "orbId"s (defined in orb as "External ID"). Now, we also need to differentiate by it's version. `growth` version 4 needs to be treated differently than `growth` version 3. Only the latest defined versions for each plan will have self-upgrade capabilities.
> Considering intended behavior defined [here](https://docs.google.com/spreadsheets/d/1ePbNSizlatHJ8Jhjk-in_0HKCgvXXhuPEzM_gJqktuc/edit?gid=0#gid=0)

So in short, we need to have some sort of versioning for our local definitions of plans. This simple requirement lead me to the following changes:
- No more `free`, `starter`, `growth`. Instead, current plans  should migrate to `free-v1`, `starter-v1`, `growth-v1`.
- We add a `orbVersion` property, and use it to map from Orb plans to our local plans. `starter-v1` encapsulates any `starter` orb version from 1 - 5, and `growth-v1` encapsulates any `growth` version from 1 - 3.
- We create the all-new plans: `free-v2`, `starter-v2`, `growth-v2`. They map to orb: free v2, starter v6 and growth v4, respectively.
- We remove the ability to upgrade from the `v1` plans, but let them downgrade to `free-v2`.
- Added `isFree` property to plan definitions, to facilitate checks related to free plans now that it's not a single code.

Caveats:
- We need to migrate the `plans` table to use the new v1 names.
- The orb testing and prod environments get out of sync with versions
  > I've manually matched `starter` and `growth` plans in both environments to arrive to the same version. But the `free` plan is a version ahead in the test env. So what to do? And even then, can we rely that they'll stay in sync?

I'm considering:
- Instead of using the plan name to track plans in the plans table(`plan.name = 'free-v1'`), we can track the version separately (`plan.name = 'free'; plan.version = 2`). But then that version better match the orb version.
- We store `orb_future_plan`, and that just stores the plan name. We also need to store the version now, I guess?

## Second pass
Trying out the idea of having `name` and `version` separately, instead of all in the name. Practical changes:
- In plan definitions, I've simplified `code` and `orbId` into a single `name` property. Also, omitted the name from the flags there.
- Removed versioning from the names. All versions of the free plan will be called `free` and so on. (still differentiating legacy pre-orb though)
- Because different versions of definitions are not distinguished by their names anymore, made the `prevPlans` and `nextPlans` properties reference actual full definitions.
- Added `version` to the plans table. (no migration yet though)
- Changed code throughout the code-base to consider `name` and `version`.

Caveats:
- Got to create the migration for adding `version` to the plans table, then backfill with current versions by fetching from the orb api.
- Not sure my checks throughout the codebase comparing `planDefinition1 === planDefinition2` will all work. Some might not because it's comparing references. For now, just showing the concept though. Might have to change.


packages/server/lib/controllers/v1/orb/postWebhooks.ts
```diff
@@ -72,14 +72,39 @@ interface SubscriptionPlanChangedScheduledEvent extends SubscriptionEvent {
...
+}
+
+interface SubscriptionPlanVersionChangedScheduledEvent extends SubscriptionEvent {
+    type: 'subscription.plan_version_change_scheduled';
+    properties: {
+        effective_date: string;
+        previous_plan_version_number: number;
+        new_plan_version_number: number;
+    };
+}
```
kaposke, line 91: https://docs.withorb.com/integrations-and-exports/webhooks#subscription-plan-version-change-scheduled

packages/shared/lib/services/plans/definitions.ts
```diff
@@ -31,37 +37,26 @@ export const freePlan: PlanDefinition = {
...
-            { title: '10 connections' },
-            { title: '1k actions' },
-            { title: '5k synced records' },
-            { title: '2 environments' },
-            { title: 'API authorization' },
-            { title: 'Syncs & actions' },
-            { title: 'MCP & AI Tools' },
-            { title: 'Proxy requests' }
-        ],
-        sub: 'Certain features of free projects (syncs & actions) are paused after two weeks of inactivity.'
```
kaposke, line 46: We don't display cards anymore

packages/shared/lib/services/plans/definitions.ts
```diff
@@ -184,17 +163,142 @@ export const enterprisePlan: PlanDefinition = {
...
+        name: 'free-v1',
+        sync_frequency_secs_min: 3600,
+        auto_idle: true,
+        monthly_actions_max: 1000,
+        monthly_active_records_max: 5000,
+        has_webhooks_script: false,
+        has_webhooks_forward: false,
+        can_override_docs_connect_url: false,
+        can_customize_connect_ui_theme: false,
+        can_disable_connect_ui_watermark: false
```
kaposke: Could share flags

packages/shared/lib/services/plans/definitions.ts
```diff
@@ -21,7 +27,7 @@ export const freePlan: PlanDefinition = {
         function_compute_gbms_max: 50_000_000,
         webhook_forwards_max: 100_000,
         function_logs_max: 100_000,
-        name: 'free',
+        name: 'free-v2',
```
kaposke: It's a big assumption that `code` and `name` have to be the same.

packages/utils/lib/telemetry/tracer.ts
```diff
@@ -6,7 +6,7 @@ export function tagTraceUser({ account, environment, plan }: { account: DBTeam;
     tracer.setUser({
         id: String(account.id),
         environmentId: String(environment.id),
-        paying: plan?.name !== 'free' ? 'paying' : 'free',
+        paying: plan?.name !== 'free-v1' && plan?.name !== 'free-v2' ? 'paying' : 'free',
```
kaposke: Don't have access to `shared` package...

packages/webapp/src/components-v2/AppSidebar/index.tsx
```diff
@@ -89,11 +87,9 @@ export const AppSidebar: React.FC = () => {
...
             </SidebarContent>
             <SidebarFooter className="p-0">
-                {plan?.name === 'free' && (
-                    <div className="px-3 mb-8">
-                        <UsageCard />
-                    </div>
-                )}
+                <div className="px-3 mb-8">
+                    <UsageCard />
+                </div>
```
kaposke, line 92: I removed this condition in my UsageCard redesign PR, but it's not merged yet

packages/billing/lib/clients/orb.ts
```diff
@@ -251,7 +252,8 @@ export class OrbClient implements BillingClient {
 
             return Ok({
                 id: res.subscription.id,
-                planExternalId: res.subscription.plan!.external_plan_id!
+                planExternalId: res.subscription.plan!.external_plan_id!,
+                planVersion: res.subscription.plan!.version
```
kaposke, line 256: https://docs.withorb.com/api-reference/subscription-change/apply-subscription-change#response-subscription-plan-version

packages/billing/lib/clients/orb.ts
```diff
@@ -140,7 +140,8 @@ export class OrbClient implements BillingClient {
             return Ok({
                 id: sub.id,
                 pendingChangeId: sub.pending_subscription_change?.id,
-                planExternalId: sub.plan?.external_plan_id || ''
+                planExternalId: sub.plan?.external_plan_id || '',
+                planVersion: sub.plan?.version || 0
```
kaposke, line 144: https://docs.withorb.com/api-reference/subscription/fetch-subscription#response-plan-version

packages/billing/lib/clients/orb.ts
```diff
@@ -123,7 +123,7 @@ export class OrbClient implements BillingClient {
                 external_plan_id: planExternalId,
                 start_date: startDate
             });
-            return Ok({ id: subscription.id, planExternalId: planExternalId });
+            return Ok({ id: subscription.id, planExternalId: planExternalId, planVersion: subscription.plan?.version || 0 });
```
kaposke, line 126: https://docs.withorb.com/api-reference/subscription/create-subscription#response-plan-version

packages/utils/lib/telemetry/tracer.ts
TBonnin: can we use `includes('free-')`

packages/shared/lib/services/plans/definitions.ts
TBonnin: is `flags.name` required? can we get rid of it? name isn't really a flag anyway

packages/shared/lib/services/plans/definitions.ts
TBonnin: i would reuse flags, at least between versions

packages/utils/lib/telemetry/tracer.ts
TBonnin: should `isPaid` added to the plan table?

packages/shared/lib/services/plans/definitions.ts
```diff
@@ -1,237 +1,137 @@
...
-            { title: '5k synced records' },
-            { title: '2 environments' },
-            { title: 'API authorization' },
-            { title: 'Syncs & actions' },
-            { title: 'MCP & AI Tools' },
-            { title: 'Proxy requests' }
-        ],
-        sub: 'Certain features of free projects (syncs & actions) are paused after two weeks of inactivity.'
-    }
+    flags: freePlanFlags
```
TBonnin, line 19: instead of creating a flags file you could set them in one planDefinition and reference it in the other one `flags: freeV2Plan.flags`. It keeps everything in one file but remove the duplication

packages/shared/lib/services/plans/definitions.ts
```diff
@@ -1,237 +1,137 @@
...
-        ]
-    }
+    flags: growthPlanFlags
 };
 
+// Set up cross-references after all plans are defined
+freeV2Plan.nextPlan = [starterV2Plan, growthV2Plan];
+starterV2Plan.prevPlan = [freeV2Plan];
+starterV2Plan.nextPlan = [growthV2Plan];
+growthV2Plan.prevPlan = [freeV2Plan, starterV2Plan];
```
TBonnin, line 52: the compiler doesn't like the cross references?

packages/shared/lib/services/plans/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+
+import type { PlanDefinition } from '@nangohq/types';
+
+export function getLatestVersionInDefinition(definition: PlanDefinition): number {
+    if (Array.isArray(definition.orbVersion)) {
+        return Math.max(...definition.orbVersion);
+    } else if (typeof definition.orbVersion === 'number') {
+        return definition.orbVersion;
+    }
+    return -1;
```
TBonnin, line 11: can we return undefined or null? something that is basically not a number?

packages/shared/lib/services/plans/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * Returns plan matching the given name and version, or latest version if no version is provided
+ */
+export function getMatchingPlanDefinition(name: string, version?: number | null): PlanDefinition | undefined {
+    if (!version) {
+        return getLatestVersion(name);
+    }
+
+    return plansList.find(
+        (p) => (p.name === name && p.orbVersion === version) || (p.orbVersion && Array.isArray(p.orbVersion) && p.orbVersion.includes(version))
```
TBonnin, line 32: isn't p.orbVersion always an array? also should the right part of the OR also check for the name?

packages/webapp/src/pages/Team/Billing/types.ts
```diff
@@ -7,3 +7,12 @@ export interface PlanDefinitionList {
...
     isUpgrade?: boolean;
 }
+
+export function getLatestVersionInDefinition(definition: PlanDefinition): number {
+    if (Array.isArray(definition.orbVersion)) {
+        return Math.max(...definition.orbVersion);
+    } else if (typeof definition.orbVersion === 'number') {
+        return definition.orbVersion;
+    }
+    return -1;
```
TBonnin, line 17: here again. I am not a huge fan of returning the same type when no version is found

packages/shared/lib/services/plans/definitions.ts
kaposke: Because of declaration orders :(

packages/shared/lib/services/plans/utils.ts
kaposke: I've made it a single number or array of numbers. Could change it to always be an array though.
And yeah, the condition is wrong, nice catch.

PR #4902, opened 2025-10-22
fix: record schema compilation caching

TBonnin:
when validating saved records we compile and cache the compiled model schema.
If the compilation fails for some reason the compilation is attempted
again on all batchSave (and keeps failing).
Ajv compilation seems to hold some references/data in memory and since
we use a static ajv instance the memory is never released and is piling up.

This commit ensures we also cache compilation failure and not re-attempt
to compile a failing schema.
It doesn't solve the static ajv instance holding on memory but it would
take 1000s of different models/versions to run into some kind of memory
leak. To keep in mind though.

---

**Cache Failed JSON Schema Compilations in `dataValidation.ts`**

This pull request updates the schema validation logic in `packages/runner-sdk/lib/dataValidation.ts` so that schema compilation failures (via Ajv) are cached, not just successful validator functions. By storing the `Error` instance directly in the cache keyed by `modelName-version`, the system avoids retrying compilation on invalid schemas during subsequent batch operations, which prevents both redundant computations and unbounded memory use caused by the static Ajv instance accumulating data. The change includes explanatory comments and minimal control flow adjustments to implement this error-caching behavior.

<details>

• Modified `cache` definition in `dataValidation.ts` to allow both `ValidateFunction<any>` and `Error` as values, keyed by `modelName-version`.
• On schema compilation failure, store the `Error` object directly in the `cache` and return it as an array to the caller.
• During validation, added logic to immediately return the cached error if present, thus bypassing re-compilation attempts.
• Added inline comments clarifying the rationale and flow of the error-caching strategy.
• Made minor restructuring in `validateData` to separate handling for cache hits and compilation attempts.

</details>

<details>

• `packages/runner-sdk/lib/dataValidation.ts`
• `validateData` function
• Validator cache logic and error handling code

</details>

---

### DIFF
--- a/packages/runner-sdk/lib/dataValidation.ts
+++ b/packages/runner-sdk/lib/dataValidation.ts
@@ -12,7 +12,7 @@ export interface ValidateProps {
 }
 
 let ajv: Ajv;
-const cache = new Map<string, ValidateFunction<any>>();
+const cache = new Map<string, ValidateFunction<any> | Error>(); // caching compiled validators (or error if compilation fails) by modelName-version
 
 export function clearValidationCache() {
     cache.clear();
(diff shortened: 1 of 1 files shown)


rossmcewan (approved):

TBonnin:
@rossmcewan I'll let you merge this PR and deploy the runner whenever you prefer

TBonnin:
@rossmcewan can we deploy this change?

khaliqgant (approved):

PR #4905, opened 2025-10-23
fix(runner): be more explicit with the runner sizing

rossmcewan:
---

**Formula-based Kubernetes runner sizing with env-driven guardrails**

The PR removes the hard-coded, tiered sizing table for Kubernetes runners and replaces it with a proportional formula. CPU and memory requests are now derived as a percentage (default 60 %) of the underlying node capacity, then clamped between new minimum/maximum values defined via environment variables. The same env vars are parsed and validated in `parse.ts`, making resource envelopes configurable without code changes.

<details>

• Deleted legacy decision tree in `getResourceLimits()` and introduced formula using `REQUEST_CPU_MULTIPLIER` / `REQUEST_MEMORY_MULTIPLIER`
• Added eight private constants (`MAX_REQUEST_CPU`, `MIN_REQUEST_CPU`, etc.) sourced from `envs` for clamping
• Extended `packages/utils/lib/environment/parse.ts` with six new Zod-validated env vars: `RUNNER_MAX_REQUEST_CPU`, `RUNNER_MAX_REQUEST_MEMORY`, `RUNNER_MIN_REQUEST_CPU`, `RUNNER_MIN_REQUEST_MEMORY`, `RUNNER_REQUEST_CPU_MULTIPLIER`, `RUNNER_REQUEST_MEMORY_MULTIPLIER`
• Default upper bound lowered from 8 CPU / 32 GiB to 4 CPU / 16 GiB unless overridden by env
• Net diff: −64 lines (deleted logic) +21 lines (new formula and env parsing)

</details>

<details>

• `packages/jobs/lib/runner/kubernetes.ts` (resource calculation, new fields)
• `packages/utils/lib/environment/parse.ts` (env parsing schema)
• Runner provisioning behaviour in Kubernetes clusters

</details>

---

### DIFF
--- a/packages/jobs/lib/runner/kubernetes.ts
+++ b/packages/jobs/lib/runner/kubernetes.ts
@@ -427,74 +427,18 @@ class Kubernetes {
     }
 
     private getResourceLimits(node: Node): { requests: { cpu: string; memory: string }; limits: { cpu: string; memory: string } } {
-        if (node.cpuMilli >= 8000 && node.memoryMb >= 32000) {
-            return {
-                requests: {
-                    cpu: '4000m',
-                    memory: '16384Mi'
-                },
-                limits: {
-                    cpu: '8000m',
-                    memory: '32768Mi'
-                }
-            };
-        }
-        if (node.cpuMilli >= 4000 && node.memoryMb >= 16000) {
-            return {
-                requests: {
-                    cpu: '4000m',
-                    memory: '8192Mi'
-                },
-                limits: {
-                    cpu: '4000m',
-                    memory: '16384Mi'
-                }
-            };
-        }
-        if (node.cpuMilli >= 4000 && node.memoryMb >= 8000) {
-            return {
-                requests: {
-                    cpu: '2000m',
-                    memory: '4096Mi'
-                },
-                limits: {
-                    cpu: '4000m',
-                    memory: '8192Mi'
-                }
-            };
-        }
-        if (node.cpuMilli > 2000 || node.memoryMb >= 4000) {
-            return {
-                requests: {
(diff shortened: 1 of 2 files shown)


packages/jobs/lib/runner/kubernetes.ts
```diff
@@ -426,75 +426,23 @@ class Kubernetes {
         return accountId ? envs.RUNNER_PROFILED_ACCOUNTS.includes(accountId) : false;
     }
 
+    private MAX_REQUEST_CPU = 4000;
+    private MAX_REQUEST_MEMORY = 16384;
```
TBonnin: we could theoretically support 8cpu/32gb runner before. I think we used it temporarily once. Are we sure we will never need to have this option available?

packages/jobs/lib/runner/kubernetes.ts
```diff
@@ -426,75 +426,23 @@ class Kubernetes {
...
+    private MIN_REQUEST_CPU = 100;
+    private MIN_REQUEST_MEMORY = 128;
```
TBonnin: it is not guaranteed the runner can even be loaded with 128mb ;)

TBonnin (approved):

packages/jobs/lib/runner/kubernetes.ts
rossmcewan: Nope. Not sure. Happy to increase or make it an environment variable.

packages/jobs/lib/runner/kubernetes.ts
rossmcewan: What do you suggest as a healthy minimum?

PR #4919, opened 2025-10-27
fix(record count): do not count records from deleted connections

TBonnin:
Record counts table entries are not deleted when a connection is soft deleted (only when hard deleted) which means that records are still counted until the connection is hard deleted.
Since we delete connections from different places in the codebase it would require a big refactoring to delete the record counts table when connections is soft deleted. Main pain point is connections are soft-deleted when an integration is soft-deleted, deep inside the shared package, which hasn't any knowledge of the records packages :(
Instead I decided to account for the deleted connections when we report the record counts. The logic was already quite cumbersome and wasn't supporting pagination.

 

---

**Exclude soft-deleted connections from record metrics + add pagination helpers**

Soft-deleted connections were still represented in the `record_counts` table, so the "Records" usage metric billed/observed more data than actually active. This PR avoids modifying the underlying table and instead filters record-count rows at read-time. It introduces paginated accessors in both the Records and Connections layers and rewrites the metering cron to join these two paginated streams, aggregating only active (non-deleted) connections. Additional integration tests ensure the new iterators and counting logic behave as expected.

<details>

• Added `records.paginateRecordCounts()` – offset-based async generator that yields batches from `RECORD_COUNTS_TABLE`.
• Added `connectionService.paginateConnections()` – cursor-based async generator returning active connections with related account/env rows; supports optional `connectionIds` filter.
• Re-implemented `observability.exportRecordsMetrics()` in `packages/metering/lib/crons/usage.ts` to: a) page through record counts, b) fetch matching connections, c) skip deleted connections, and d) aggregate metrics per account.
• Updated unit/integration tests for Records and Connection services to cover new pagination helpers and counting logic.
• Minor cleanup: removed unused `environmentService` import, dropped redundant warning log in usage tracker.

</details>

<details>

• `packages/metering/lib/crons/usage.ts`
• `packages/records/lib/models/records.ts`
• Associated integration test suites
• `packages/account-usage/lib/usage.ts` (small log change)

</details>

---

### DIFF
--- a/packages/account-usage/lib/usage.ts
+++ b/packages/account-usage/lib/usage.ts
@@ -196,7 +196,6 @@ export class UsageTracker implements IUsageTracker {
                 case 'function_logs': {
                     const billingUsage = await this.getBillingUsage(accountId);
                     if (billingUsage.isErr()) {
-                        logger.warning(`Failed to fetch billing usage for accountId: ${accountId}`, billingUsage.error);
                         if (billingUsage.error.message === 'rate_limit_exceeded') {
                             span?.setTag('rate_limited', true);
                         }
--- a/packages/metering/lib/crons/usage.ts
+++ b/packages/metering/lib/crons/usage.ts
@@ -4,7 +4,7 @@ import * as cron from 'node-cron';
 import { billing as usageBilling } from '@nangohq/billing';
 import { getLocking } from '@nangohq/kvstore';
 import { records } from '@nangohq/records';
-import { connectionService, environmentService } from '@nangohq/shared';
+import { connectionService } from '@nangohq/shared';
 import { flagHasUsage, getLogger, metrics } from '@nangohq/utils';
 
 import { envs } from '../env.js';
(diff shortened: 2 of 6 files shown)


packages/account-usage/lib/usage.ts
```diff
@@ -196,7 +196,6 @@ export class UsageTracker implements IUsageTracker {
                 case 'function_logs': {
                     const billingUsage = await this.getBillingUsage(accountId);
                     if (billingUsage.isErr()) {
-                        logger.warning(`Failed to fetch billing usage for accountId: ${accountId}`, billingUsage.error);
```
TBonnin, line 199: unrelated. I just noticed we were logging the error twice. Here and in the catch

packages/metering/lib/crons/usage.ts
```diff
@@ -95,31 +95,43 @@ const observability = {
...
-                        existing.sizeBytes += metric.sizeBytes;
-                    } else {
-                        metricsByAccount.set(metric.accountId, { count: metric.count, sizeBytes: metric.sizeBytes });
+                    const connectionIds = recordCounts.value.map((r) => r.connection_id);
+                    for await (const res of connectionService.paginateConnections({ connectionIds })) {
+                        if (res.isErr()) {
+                            throw res.error;
+                        }
+                        for (const value of res.value) {
+                            const recordCount = recordCounts.value.find((r) => r.connection_id === value.connection.id);
```
kaposke, line 121: Potentially avoidable o(n^2), but probably negligible compared to queries?

kaposke (approved):

packages/metering/lib/crons/usage.ts
TBonnin: This entire analytics logic only works because the scale is small. We are gonna need to come up with a better solution as our user base of customers saving records expands

PR #4927, opened 2025-10-28
fix(runner): error handling and retry in jobs client

TBonnin:
retry mechanism relies on a rejected promise so we cannot wrap the output with Result.
But we can catch everything in the jobs client functions so downstream code doesn't have to deal with exception but only Result

 

---

**Refactor runner HTTP layer to return `Result` objects and centralize retry logic**

This moderate-sized PR (~170 LOC touched) removes in-function `try/catch` blocks from runner call-sites and moves retry/error handling down into the shared HTTP utilities. `packages/runner/lib/clients/jobs.ts` now always returns a `Result` (`Ok`/`Err`) so callers (`idle.ts`, `register.ts`) can use `isErr()` instead of their own exception logic. The new implementation delegates network retries to an enhanced `httpFetch`, which retries only on network failures, 5xx, and 429 status codes and logs errors consistently. Special handling for 413 (payload too large) is preserved.

Default back-off parameters are defined once (`defaultRetryOptions`) and passed through, reducing duplicate code.

<details>

• Added centralized retry/error logic in `httpFetch` using `retryWithBackoff` and status-code filtering
• Updated `packages/runner/lib/clients/jobs.ts` methods (`postHeartbeat`, `putTask`, `postRegister`, `postIdle`) to wrap responses in `Result` and replace local retry blocks
• Changed error constant names to snake_case (e.g., `heartbeat_failed`, `put_task_failed`)
• Simplified `packages/runner/lib/idle.ts` and `packages/runner/lib/register.ts` – they now check `res.isErr()` instead of catching exceptions
• Introduced `defaultRetryOptions` constant and optional back-off options parameter in `httpFetch`

</details>

<details>

• `packages/runner/lib/clients/http.ts`
• `packages/runner/lib/clients/jobs.ts`
• `packages/runner/lib/idle.ts`
• `packages/runner/lib/register.ts`

</details>

---

### DIFF
--- a/packages/runner/lib/clients/http.ts
+++ b/packages/runner/lib/clients/http.ts
@@ -1,17 +1,46 @@
+import { retryWithBackoff } from '@nangohq/utils';
+
 import { logger } from '../logger.js';
 
-import type { RequestInfo } from 'undici/types/index.js';
-
-export async function httpFetch(resource: RequestInfo | URL, init?: RequestInit): Promise<Response> {
-    const res = await fetch(resource, init);
-    if (res.status >= 400) {
-        let content;
-        if (res.headers.get('content-type')?.includes('application/json')) {
-            content = await res.json().catch(() => 'Cannot parse response as JSON.');
-        } else {
-            content = await res.text();
-        }
-        logger.error(`Error: ${init?.method} ${resource as string} -> ${res.status} ${JSON.stringify(content)}`);
+export async function httpFetch(
+    url: string | URL,
+    init?: RequestInit,
+    backoffOptions?: {
+        startingDelay?: number;
+        timeMultiple?: number;
+        numOfAttempts?: number;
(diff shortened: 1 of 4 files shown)


rossmcewan (approved):
Would there be any benefit to adding some error logging here? It seems like we lose any relevant information once we return the error.

khaliqgant (approved):

TBonnin:
> Would there be any benefit to adding some error logging here? It seems like we lose any relevant information once we return the error.

error logging happens in `httpFetch` https://github.com/NangoHQ/nango/blob/master/packages/runner/lib/clients/http.ts#L14 but I realize only when getting a response, not when it throws. I'll add that. 

I also realize we shouldn't retry all failed requests. 4XX shouldn't be retried. I will fix that also tomorrow

PR #4928, opened 2025-10-28
fix(usage): account for deleted connections when revalidating records

TBonnin:
I fixed the issue where we were not accounting for deleted connections when reporting records count but I forgot the records count data was also used directly in usage revalidation.
This commit remove the records.metrics function that cannot be used directly and ensure records for deleted connections are not counted. Because records and connections don't live in the same db, joins aren't possible and we have to do some cumbersome gymnastic in the code. It is not super scalable but since we are dealing with a very small number of environments it is still very manageable in term of performance

 

---

**Exclude deleted connections from record-usage revalidation & remove obsolete metrics helper**

Reworks the record-count flow used by the `UsageTracker` to ensure deleted connections are no longer included when re-validating account usage. The obsolete aggregation helper `records.metrics()` (and its tests) is dropped, `records.paginateRecordCounts()` gains optional `environmentIds` filtering, and `UsageTracker` now iterates through record/count pages and live connections via the new helper `getRecordsUsage()`.

Changes touch only three files (~150 net LOC) but modify a cross-package contract (records ⇄ account-usage) and introduce a nested pagination loop that replaces a previous single query.

<details>

• Deleted `records.metrics()` and related test block in `packages/records/lib/models/records.integration.test.ts`
• Extended `records.paginateRecordCounts()` to accept `environmentIds` filter and adjusted query building
• Added helper method `UsageTracker.getRecordsUsage()`; `UsageTracker.revalidate()` ‘records’ branch now calls this and writes to cache
• Updated integration tests: removed `metrics` suite, added environment-filtered pagination tests
• Minor typing / import updates; commit message clarifies performance/scale expectations

</details>

<details>

• `packages/records/lib/models/records.ts` – pagination API & removal of metrics
• `packages/account-usage/lib/usage.ts` – revalidation logic & new helper
• Record integration tests

</details>

---

### DIFF
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -88,52 +88,23 @@ export async function getRecordStatsByModel({
     }
 }
 
-export async function metrics({ environmentIds }: { environmentIds?: number[] } = {}): Promise<
-    Result<{ environmentId: number; count: number; sizeBytes: number }[]>
-> {
-    try {
-        if (environmentIds && environmentIds.length === 0) {
-            return Err(new Error(`No environment provided`));
-        }
-
-        const query = db
-            .from(RECORD_COUNTS_TABLE)
-            .select<
-                { environment_id: number; count: number; size_bytes: number }[]
-            >('environment_id', db.raw('SUM(count) as count'), db.raw('SUM(size_bytes) as size_bytes'))
-            .groupBy('environment_id')
-            .having(db.raw('sum(size_bytes) > 0 OR sum(count) > 0')); // only return entries with records
-
-        if (environmentIds) {
-            query.whereIn('environment_id', environmentIds);
-        }
-
-        const res = await query;
-        if (!res) {
-            return Err(new Error(`Failed to count records`));
-        }
-        const metrics = res.map((r) => ({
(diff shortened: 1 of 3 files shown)


rossmcewan (approved):

packages/records/lib/models/records.ts
```diff
@@ -88,52 +88,23 @@ export async function getRecordStatsByModel({
...
+    environmentIds?: number[];
     batchSize?: number;
 } = {}): AsyncGenerator<Result<RecordCount[]>> {
     let offset = 0;
 
     try {
         while (true) {
-            // TODO: optimize with cursor pagination when needed
-            const results = await db.select('*').from(RECORD_COUNTS_TABLE).orderBy('connection_id', 'model').limit(batchSize).offset(offset);
+            // TODO: optimize with cursor pagination
```
khaliqgant, line 102: This captured in linear?

khaliqgant (approved):

packages/records/lib/models/records.ts
TBonnin: not really because it is very very long term (assuming we will ever need it)

PR #4931, opened 2025-10-29
fix(server): add ready endpoint

rossmcewan:
---

**Add `/ready` readiness endpoint with shutdown flag & delay**

Introduces a lightweight readiness probe so orchestrators can stop routing traffic before the process terminates. A new global flag tracks shutdown state; `process.on('SIGTERM')` now sets this flag and waits an optional `SERVER_SHUTDOWN_DELAY` before invoking the existing `close()` routine. The new `/ready` route returns 500 while the flag is true, 200 otherwise.

<details>

• Created `packages/server/lib/utils/state.ts` with `setShuttingDown()` and `isShuttingDown()` helpers
• Added `/ready` route in `packages/server/lib/routes.ts`; returns 500 when `isShuttingDown()` is true
• Updated `packages/server/lib/server.ts`: on `SIGTERM` call `setShuttingDown(true)` and `setTimeout(close, envs.SERVER_SHUTDOWN_DELAY)`
• Extended env schema in `packages/utils/lib/environment/parse.ts` to include `SERVER_SHUTDOWN_DELAY` (number, default 0)

</details>

<details>

• `packages/server/lib/utils/state.ts`
• `packages/server/lib/routes.ts`
• `packages/server/lib/server.ts`

</details>

---

### DIFF
--- a/packages/server/lib/server.ts
+++ b/packages/server/lib/server.ts
@@ -26,6 +26,7 @@ import { runnersFleet } from './fleet.js';
 import { pubsub } from './pubsub.js';
 import { router } from './routes.js';
 import migrate from './utils/migrate.js';
+import { setShuttingDown } from './utils/state.js';
 
 import type { WebSocket } from 'ws';
 
@@ -106,6 +107,7 @@ server.listen(port, () => {
 
 // --- Close function
 const close = once(() => {
+    setShuttingDown(true);
     logger.info('Closing...');
 
     cron.getTasks().forEach((task) => task.stop());
--- a/packages/server/lib/server.ts
+++ b/packages/server/lib/server.ts
@@ -107,7 +107,6 @@ server.listen(port, () => {
 
 // --- Close function
 const close = once(() => {
-    setShuttingDown(true);
     logger.info('Closing...');
 
     cron.getTasks().forEach((task) => task.stop());
@@ -139,5 +138,6 @@ process.on('SIGINT', () => {
 
 process.on('SIGTERM', () => {
     logger.info('Received SIGTERM...');
-    close();
+    setShuttingDown(true);
+    setTimeout(close, envs.SERVER_SHUTDOWN_DELAY);
 });
--- a/packages/server/lib/routes.ts
+++ b/packages/server/lib/routes.ts
@@ -11,6 +11,7 @@ import { securityMiddlewares } from './middleware/security.js';
 import { internalApi } from './routes.internal.js';
 import { privateApi } from './routes.private.js';
 import { publicAPI } from './routes.public.js';
+import { isShuttingDown } from './utils/state.js';
 import { dirname } from './utils/utils.js';
 
 import type { ApiError } from '@nangohq/types';
@@ -23,6 +24,13 @@ router.use(...securityMiddlewares());
 // -------
 // No auth routes
 router.get('/health', (_, res) => {
+    res.status(200).send({ result: 'ok' });
+});
+router.get('/ready', (_, res) => {
+    if (isShuttingDown()) {
+        res.status(500).send({ result: 'shutting down' });
+        return;
+    }
     res.status(200).send({ result: 'ok' });
 });
 router.get('/env.js', getEnvJs);
--- a/packages/server/lib/utils/state.ts
+++ b/packages/server/lib/utils/state.ts
@@ -1,0 +1,9 @@
+let shuttingDown = false;
+
+export function setShuttingDown(value: boolean) {
+    shuttingDown = value;
+}
+
+export function isShuttingDown() {
+    return shuttingDown;
+}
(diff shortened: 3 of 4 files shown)


TBonnin:
I am not sure I understand. `close` is called when receiving SIGTERM. At this point k8s should have removed it from the service endpoints and stopped sending traffic to the pod anyway. What am I missing?

rossmcewan:
> I am not sure I understand. `close` is called when receiving SIGTERM. At this point k8s should have removed it from the service endpoints and stopped sending traffic to the pod anyway. What am I missing?

My best understanding:
This is about handling readiness correctly - readiness and health probes shouldn't be the same. Deletions are not instantaneous and the readiness probe is still called during termination. There are brief periods where the pod can still receive traffic. This is especially important when load balancers are in play. 

Some more reading [here](https://blog.risingstack.com/graceful-shutdown-node-js-kubernetes/) and [here](https://github.com/kubernetes/kubernetes/pull/110191)

TBonnin:
> > I am not sure I understand. `close` is called when receiving SIGTERM. At this point k8s should have removed it from the service endpoints and stopped sending traffic to the pod anyway. What am I missing?
> 
> My best understanding: This is about handling readiness correctly - readiness and health probes shouldn't be the same. Deletions are not instantaneous and the readiness probe is still called during termination. There are brief periods where the pod can still receive traffic. This is especially important when load balancers are in play.
> 
> Some more reading [here](https://blog.risingstack.com/graceful-shutdown-node-js-kubernetes/) and [here](https://github.com/kubernetes/kubernetes/pull/110191)

An important point in the link you mentioned is:
> App waits (failureThreshold * periodSecond) before it starts to shut down – to make sure that the Service is getting notified via readinessProbe fail

Without it there is still a potential race condition where server stop listening when `close` is called but readiness hasn't failed yet. I don't think the /ready endpoint alone is enough to ensure no traffic is send to a closing pod

rossmcewan:
> > > I am not sure I understand. `close` is called when receiving SIGTERM. At this point k8s should have removed it from the service endpoints and stopped sending traffic to the pod anyway. What am I missing?
> > 
> > 
> > My best understanding: This is about handling readiness correctly - readiness and health probes shouldn't be the same. Deletions are not instantaneous and the readiness probe is still called during termination. There are brief periods where the pod can still receive traffic. This is especially important when load balancers are in play.
> > Some more reading [here](https://blog.risingstack.com/graceful-shutdown-node-js-kubernetes/) and [here](https://github.com/kubernetes/kubernetes/pull/110191)
> 
> An important point in the link you mentioned is:
> 
> > App waits (failureThreshold * periodSecond) before it starts to shut down – to make sure that the Service is getting notified via readinessProbe fail
> 
> Without it there is still a potential race condition where server stop listening when `close` is called but readiness hasn't failed yet. I don't think the /ready endpoint alone is enough to ensure no traffic is send to a closing pod

The missing piece being this:
```
process.on('SIGTERM', function onSigterm () {
  console.info('Got SIGTERM. Graceful shutdown start', new Date().toISOString())

  // Wait a little bit to give enough time for Kubernetes readiness probe to fail 
  // (we are not ready to serve more traffic)
  // Don't worry livenessProbe won't kill it until (failureThreshold: 3) => 30s
  setTimeout(greacefulStop, READINESS_PROBE_DELAY)
})
```

That's a good point. Our default values for failure threshold is 3 and period seconds 5 right now which would require a 15 second delay before we begin the close operation.

We could reduce these - maybe 2/2 like the example.

TBonnin:
I find it so backward: _k8s to tell pod to stop, pod to start saying it is not ready anymore, k8s to poll for a bit to ensure pod is *really* not ready anymore, pod to wait for a magic number of seconds to let k8s realize it is *actually* not ready anymore, ..._ 🤮

rossmcewan:
> I find it so backward: _k8s to tell pod to stop, pod to start saying it is not ready anymore, k8s to poll for a bit to ensure pod is _really_ not ready anymore, pod to wait for a magic number of seconds to let k8s realize it is _actually_ not ready anymore, ..._ 🤮

I don't think that's a good representation of what happens though. Everything is eventually consistent and we are trying to close the gaps so that requests don't slip through in those periods. We haven't implemented the probes correctly at the moment with liveness, readiness, and health all using the same endpoint at the moment. 

But you ou can have that conversation [here](https://github.com/kubernetes/kubernetes) 😁 Can I get this one approved though. It's a step in the right direction.

packages/server/lib/server.ts
```diff
@@ -137,5 +138,6 @@ process.on('SIGINT', () => {
 
 process.on('SIGTERM', () => {
     logger.info('Received SIGTERM...');
-    close();
+    setShuttingDown(true);
+    setTimeout(close, envs.SERVER_SHUTDOWN_DELAY);
```
TBonnin, line 142: should we also add the same mechanism to all of our internal apps?

TBonnin (approved):

packages/server/lib/server.ts
rossmcewan: I was thinking that. It would be the "correct" thing to do. I'll put a ticket in.

PR #4934, opened 2025-10-29
fix: keep feature flag overrides when upgrading

TBonnin:
Currently any plan change results in the override of the plan flags to the default values of the new plan. Any override that was set is discarded. (ie: an enabled flag or a manually increased limit will be lost)
This PR is ensuring flags aren't reset if more generous that the new plan default when upgrading or migrating. Downgrading always results in applying default flags for new plan.

---

**Preserve feature-flag overrides on plan upgrades/migrations**

Adds logic so that when a team upgrades or migrates to a different billing plan, any flag that was previously overridden to a *more generous* value is retained. A full reset to new-plan defaults still happens on downgrades.

Key pieces are the new `mergeFlags` helper in `packages/shared/lib/services/plans/plans.ts` and the downgrade-detection utility `isPotentialDowngrade` in `packages/shared/lib/services/plans/definitions.ts`. `handlePlanChanged()` now calls `mergeFlags` and applies the merged set instead of blindly spreading `newPlan.flags`. A comprehensive Vitest suite validates merge behaviour across upgrade, migration and downgrade scenarios.

<details>

• Introduced `mergeFlags` to combine `currentPlan` and `newPlanDefinition.flags` while preserving overrides when appropriate
• Injected `mergedFlags` into `updatePlanByTeam()` inside `handlePlanChanged()` instead of raw `newPlan.flags`
• Added downgrade detection via `isPotentialDowngrade` with an explicit downgrade matrix
• Exposed helper `getPlanDefinition` for tests and internal lookup
• Added 100+ line unit test `plans.unit.test.ts` covering downgrade, upgrade and migration paths
• Minor type additions (`PlanDefinition`) and import adjustments

</details>

<details>

• `packages/shared/lib/services/plans/plans.ts` (plan update flow)
• `packages/shared/lib/services/plans/definitions.ts` (plan metadata helpers)
• `packages/shared/lib/services/plans/plans.unit.test.ts` (new tests)

</details>

---

### DIFF
--- a/packages/shared/lib/services/plans/plans.ts
+++ b/packages/shared/lib/services/plans/plans.ts
@@ -2,10 +2,10 @@ import ms from 'ms';
 
 import { Err, Ok, flagHasPlan } from '@nangohq/utils';
 
-import { freePlan, plansList } from './definitions.js';
+import { freePlan, isPotentialDowngrade, plansList } from './definitions.js';
 import { productTracking } from '../../utils/productTracking.js';
 
-import type { DBPlan, DBTeam } from '@nangohq/types';
+import type { DBPlan, DBTeam, PlanDefinition } from '@nangohq/types';
 import type { Result } from '@nangohq/utils';
 import type { Knex } from 'knex';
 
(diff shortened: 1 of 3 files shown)


packages/shared/lib/services/plans/plans.ts
```diff
@@ -175,3 +178,118 @@ export async function handlePlanChanged(
...
+                const newIndex = sizeIndex[newPlanDefinition.flags[key]];
+                if (currentIndex > newIndex) {
+                    overrides[key] = currentPlan[key];
+                }
+                break;
+            }
+            default:
+                ((_exhaustiveCheck: never) => {
+                    throw new Error(`Unhandled plan flag key in mergeFlags: ${key}`);
+                })(key);
```
TBonnin, line 290: exhaustive check to make it a bit future proof. Any new plan attributes will require adding it to the `switch`

rossmcewan (approved):

PR #4940, opened 2025-10-30
feat(webhooks): add support for pagerduty webhooks

hassan254-prog:

- add support for pagerduty webhooks

---

**feat(webhooks): introduce first-class PagerDuty webhook ingestion with HMAC verification**

This PR adds native support for receiving PagerDuty incident webhooks. A new Express router mounted at /webhooks/pagerduty validates the X-PagerDuty-Signature header using an HMAC-SHA256 shared secret, converts PagerDuty lifecycle events (triggered, acknowledged, resolved, etc.) to the platform’s internal event schema, and forwards them to the existing processing pipeline.

The change is strictly additive—no existing providers or routes are altered—so risk to current functionality is low. Once the signing secret is configured, customers can automate incident-response workflows triggered directly from PagerDuty, integrating them seamlessly with other provider events already supported by the platform.

<details>

• Added packages/server/lib/webhook/pagerduty-webhook-routing.ts with signature validation and payload normalisation.
• Registered the new router in packages/server/lib/webhook/index.ts so the endpoint is live by default.
• Extended provider type system (packages/server/lib/webhook/types.ts) to include PagerDuty-specific payload typings and WebhookProvider enum update.
• Updated provider catalogue (packages/providers/providers.yaml) to expose PagerDuty in UI/CLI.
• Generated user documentation snippets explaining setup and secret management.
• Enforced single HMAC signing secret for PagerDuty verification.

</details>

<details>

• Webhook routing layer
• Type definitions for webhook providers
• Provider catalogue / configuration surfaces
• Developer & user documentation

</details>

---


khaliqgant:
@hassan254-prog think we need some documentation on this as it stands now I'm not sure how this works exactly

packages/server/lib/webhook/pagerduty-webhook-routing.ts
```diff
@@ -0,0 +1,84 @@
+import { NangoError } from '@nangohq/shared';
+import { Err, Ok } from '@nangohq/utils';
+
+import type { PagerDutyWebhookPayload, WebhookHandler } from './types.js';
+
+function isIpInWhitelist(ip: string, whitelist: string[]): boolean {
```
TBonnin: IP whitelist can easily be bypassed. Isn't PagerDuty webhook supporting signature?

hassan254-prog:
> @hassan254-prog think we need some documentation on this as it stands now I'm not sure how this works exactly

@khaliqgant yes, as we continue adding more webhooks support, it would be great to have a dedicated section for webhooks per provider. Where do you suggest we should place this?

packages/server/lib/webhook/pagerduty-webhook-routing.ts
hassan254-prog: yes, they do, but they generate a new key for each subscription created, which makes it difficult to use this as a method for webhook verification. However, they support other methods, including:

* IP Safelists
* Basic Auth
* Webhook Signature Validation
* OAuth Auth
* Mutual TLS Auth

I see IP Safelists as a method that could work with the current Nango setup, wdyt?
https://developer.pagerduty.com/docs/webhook-behavior

TBonnin: IP isn't really reliable (headers can be spoofed, lost/modify by LB/proxy/etc... I wouldn't consider IP whitelist to be a good verification mechanism. 
You mentioned _they generate a new key for each subscription created_, does it mean a new key per nango connection?

hassan254-prog: they generate a new key for each subscription, which can still work, we could maintain an array of these secret keys. My main concern is performance, since each webhook request can include one or more [signatures](https://developer.pagerduty.com/docs/verifying-webhook-signatures) in the header (to support zero-downtime secret rotation). If we accumulate many webhook secret keys and multiple signatures per request, processing a single webhook could become slower, wdyt?

TBonnin: what about their recommended option: TLS authentication?

hassan254-prog: Following our Slack con, since Mutual TLS requires specific configurations on Nango’s server that are tightly coupled to pagerduty, we can enforce a single secret per connection to avoid slowing down webhook processing. I’ve updated this in the latest commit.

packages/server/lib/webhook/types.ts
```diff
@@ -234,3 +234,26 @@ export interface PagerDutyWebhookPayload {
...
+        key_url?: string;
+    };
+    CallbackObjectRecId?: number;
+    [key: string]: any;
+}
+
+export interface ShipStationWebhook {
+    resource_url: string;
+    resource_type: string;
+}
```
TBonnin, line 259: is that from another PR?

TBonnin (approved):

packages/server/lib/webhook/types.ts
hassan254-prog: Yes, it's strange that this is shown as a new change to the file, considering it was already merged and isn’t part of this PR.

PR #4950, opened 2025-11-02
feat(integrations): Add support for Connectwise PSA webhooks

jeremydhoon:
This PR adds webhook support for Connectwise PSA. It authenticates webhooks by treating the user-provided webhook secret as the trusted subdomain from which to fetch each webhook's signing key.

This PR is NOT ready to be merged. There are two major issues:
- It includes a hacky workaround to allow for unexpected query parameters, which are always sent by Connectwise.
- ~It relies on using the webhook secret as the trusted subdomain for verifying Connectwise webhook signatures. The background here is that Connectwise provides a secure URL in every webhook that the recipient is meant to fetch in order to obtain the signing key for the URL. In order for this to be secure, keys can only be fetched from known, trusted sources. Unfortunately, the domain serving the signing key is not fixed, and hence we need to make this a configurable parameters for each Connectwise PSA integration.~ I actually think there are a limited number of subdomains that could host signing keys, and none of these would be specific to any one Connectwise user, so we can maybe just check that the root domain is `myconnectwise.net`.

I'd like to discuss possible solutions here with the Nango team.

**Query parameters**: I agree that sending and receiving query parameters isn't a great practice because they can't be included in the signed body of the webhook and therefore can be trivially forged in a replay attack. Would we consider allowing query parameters, but ignoring them (or possibly logging their existence but not their values), rather than passing them on to the webhook handler? We could also consider making this behavior configurable on a per-integration basis in providers.yaml.

**Trusted subdomains**: ~should we allow providers.yaml to declare integration-wide fields, perhaps specifically for webhooks?~ We can likely just look at the root domain. 

**Testing**

Added a webhook endpoint pointing to my dev machine's 
Screenshot 2025-11-02 at 23 03 18

---

**Add first-class ConnectWise PSA webhook ingestion with signature validation & dynamic key retrieval**

This PR introduces real-time webhook ingestion for the ConnectWise PSA integration. It adds a dedicated webhook router that downloads the tenant-specific public key advertised in each payload, validates the HMAC signature, and forwards verified events to the generic dispatcher. Supporting changes update routing tables, typings, provider metadata, and documentation so that ConnectWise PSA is now treated as a full webhook-capable provider. The work is self-contained and does not change behaviour for any existing integrations, but it introduces a new code path that depends on network retrieval of public keys and looser query-parameter handling.

<details>

• New connectwise-psa-webhook-routing.ts handler that fetches public keys, validates HMAC, and dispatches events
• Webhook registry updated to route provider=connectwise-psa traffic to the new handler
• postWebhook controller relaxed to ignore provider-scoped query parameters appended by ConnectWise
• providers.yaml updated to mark ConnectWise PSA as "webhooks: true" with domain whitelist *.myconnectwise.net
• Added TypeScript typings and autogenerated docs for ConnectWise PSA event shapes and setup instructions

</details>

<details>

• Webhook routing layer
• Provider registry (providers.yaml)
• postWebhook controller / request validation
• TypeScript type definitions for webhook payloads
• Generated developer documentation

</details>

---


jeremydhoon:
Update: I've implemented a trusted URL allowlist.

docs/integrations/all/connectwise-psa/webhooks.mdx
```diff
@@ -0,0 +1,96 @@
...
+  </Step>
+  <Step title="Configure your trusted ConnectWise subdomain in Nango">
+    1. In the Nango UI _Settings_ page for your ConnectWise PSA integration
+    2. In the **Webhook Configuration** section, enter your ConnectWise subdomain in the **Webhook Secret** field
+    3. The subdomain should be just the subdomain portion (
+       e.g., `mycompany` if you log into Connectwise at `mycompany.myconnectwise.net`)
+    4. Click **Save**
+
+    <Warning>
+      This subdomain is used for security validation. Only webhooks from this specific ConnectWise environment will be accepted.
```
TBonnin:
```suggestion
      This subdomain is used for security validation. Only webhooks from this specific ConnectWise subdomain will be accepted.
```

packages/server/lib/controllers/webhook/environmentUuid/postWebhook.ts
```diff
@@ -21,11 +21,12 @@ const paramValidation = z
...
-    if (emptyQuery) {
-        res.status(400).send({ error: { code: 'invalid_query_params', errors: zodErrorToHTTP(emptyQuery.error) } });
-        return;
-    }
+    // TEMPORARY: Query parameter validation disabled for testing ConnectWise webhooks
+    // const emptyQuery = requireEmptyQuery(req);
+    // if (emptyQuery) {
+    //     res.status(400).send({ error: { code: 'invalid_query_params', errors: zodErrorToHTTP(emptyQuery.error) } });
+    //     return;
+    // }
```
TBonnin: let's allow query string parameters but discard them. This can be done by removing `strict` above. let's add a comment though to mention some 3rd parties are adding query string parameters that must be discarded

packages/server/lib/webhook/connectwise-psa-webhook-routing.ts
```diff
@@ -0,0 +1,195 @@
...
+        }
+
+        logger.info('Webhook signature validated successfully', { configId: nango.integration.id });
+    } else {
+        // No trustedSubdomain configured - skip validation but log a warning
+        logger.warn('Webhook signature validation skipped - no trustedSubdomain configured', {
+            configId: nango.integration.id,
+            message:
+                'Configure the Webhook Secret field in integration settings with your trusted ConnectWise subdomain (e.g., "sandbox-na" or "api-na") to enable webhook signature validation'
+        });
```
TBonnin: I don't think we should proceed without validation. `trustedSubdomain` should be mandatory and logic fail if we cannot validate the webhook

packages/server/lib/controllers/webhook/environmentUuid/postWebhook.ts
jeremydhoon: I don't believe we need to remove `.strict()` above. The commented code was responsible for disallowing query parameters. I've uncommented this code, moved it below (to where we've parsed `providerConfigKey` and `environmentUuid`), and replaced returning a 400 with logging that parameters were discarded.

packages/server/lib/webhook/connectwise-psa-webhook-routing.ts
jeremydhoon: Great, I'm happy to be stricter here. I've implemented this.

jeremydhoon:
@TBonnin I believe my latest changes should address your initial feedback. Happy to make further changes (e.g. in case you'd prefer not to log if we discarded query parameters).

TBonnin:
> @TBonnin I believe my latest changes should address your initial feedback. Happy to make further changes (e.g. in case you'd prefer not to log if we discarded query parameters).

@jeremydhoon Thank you very much. I have just pushed a commit modifying the error handling and logging. 

@hassan254-prog can you review the PR tomorrow? thank you very much

packages/server/lib/controllers/webhook/environmentUuid/postWebhook.ts
hassan254-prog, line 1: @jeremydhoon to simplify this, we can retrieve the `keyUrl` from the webhook response , then use it to obtain the `signing_key` and verify the webhook. After that, we can extract the `subdomain` from the `keyUrl` and use it as the `connectionIdentifierValue` to fetch the corresponding connection IDs, instead of always selecting the first one. This approach ensures the `subdomain` is defined only once at the connection level. 
Since `ConnectWise PSA (Staging)` doesn't require one, we can have the `Subdomain` as a hidden field within the connect-ui form with `default_value: api-staging`. 
As a way of validating the `keyUrl`, you can check for connections with the subdomain as the `connectionIdentifierValue` and drop webhooks if you don't find a valid connection. 
Happy to take this over and push to the finish line.

packages/server/lib/webhook/connectwise-psa-webhook-routing.ts
```diff
@@ -0,0 +1,147 @@
...
+ */
+async function fetchSigningKey(keyUrl: string, trustedSubdomain: string): Promise<Result<string>> {
+    try {
+        // Validate that the key URL is from a trusted subdomain BEFORE fetching
+        const trusted = trustedKeyUrl(keyUrl, trustedSubdomain);
+        if (trusted.isErr()) {
+            return trusted;
+        }
+
+        const response = await axiosInstance.get<SigningKeyResponse>(keyUrl);
```
hassan254-prog, line 69: It’s strange that this endpoint doesn’t require authentication. 
Also curious, does the `signing_key` ever change? If it’s constant, we could have the user use it as Nango’s webhook secret and avoid making an extra request.

docs/integrations/all/connectwise-psa/webhooks.mdx
hassan254-prog, line 1: You can omit this for now, since we don’t yet have an official unified location for the webhook documentation.

packages/server/lib/webhook/connectwise-psa-webhook-routing.ts
jeremydhoon: I can confirm that these URLs change. Whether they are per-webhook or simply short-lived and subject to frequent rotation, I haven't taken the time to establish. Here are examples of two (now expired) links:

https://sandbox-na.myconnectwise.net/v2025_1//system/callbackkey/index.rails?companyName=ong_cs1&id=ad48c0e7-a9f9-4535-9c5e-6c46d8491f5a

https://sandbox-na.myconnectwise.net/v2025_1//system/callbackkey/index.rails?companyName=ong_cs1&id=38a28308-c33b-4893-95de-4f543106cc51

I would hope, but have not tested, that their signing keys change as well. Given these are symmetric secrets, I imagine it would not have been a huge burden for Connectwise to implement key rotation along with URL rotation.

This leads me to authentication: the IDs in these URLs are pretty long and entropic, so I think that provides enough security, especially given that they seem to be rotated with some frequency.

docs/integrations/all/connectwise-psa/webhooks.mdx
jeremydhoon: ~Feel free to remove webhooks.mdx. I was basing this off of Xero's webhook documentation: https://nango.dev/docs/integrations/all/xero/webhooks~

I've gone ahead and removed this file since it was causing a lint failure.

packages/server/lib/controllers/webhook/environmentUuid/postWebhook.ts
```diff
@@ -21,11 +21,10 @@ const paramValidation = z
...
 export const postWebhook = asyncWrapper<PostPublicWebhook>(async (req, res) => {
-    const emptyQuery = requireEmptyQuery(req);
-    if (emptyQuery) {
-        res.status(400).send({ error: { code: 'invalid_query_params', errors: zodErrorToHTTP(emptyQuery.error) } });
-        return;
-    }
+    // discard all query parameters
+    // some providers send unverifiable query parameters
+    // we ignore them for security reasons
+    req.query = {} as PostPublicWebhook['Querystring'];
```
jeremydhoon: FYI this line is giving me the following error when handling webhooks:

```
TypeError: Cannot set property query of #<IncomingMessage> which has only a getter
    at <anonymous> (/Users/jhoon/code/nango/packages/server/lib/controllers/webhook/environmentUuid/postWebhook.ts:27:9)
    at <anonymous> (/Users/jhoon/code/nango/packages/server/lib/utils/asyncWrapper.ts:31:21)
    at /Users/jhoon/code/nango/node_modules/dd-trace/packages/datadog-instrumentations/src/router.js:51:25
    at Layer.handleRequest (/Users/jhoon/code/nango/node_modules/router/lib/layer.js:152:17)
    at Layer.handleRequest (/Users/jhoon/code/nango/node_modules/dd-trace/packages/datadog-instrumentations/src/router.js:225:21)
    at next (/Users/jhoon/code/nango/node_modules/router/lib/route.js:157:13)
    at Route.dispatch (/Users/jhoon/code/nango/node_modules/router/lib/route.js:117:3)
    at handle (/Users/jhoon/code/nango/node_modules/router/index.js:435:11)
    at handle (/Users/jhoon/code/nango/node_modules/dd-trace/packages/datadog-instrumentations/src/router.js:51:25)
    at Layer.handleRequest (/Users/jhoon/code/nango/node_modules/router/lib/layer.js:152:17)
    at Layer.handleRequest (/Users/jhoon/code/nango/node_modules/dd-trace/packages/datadog-instrumentations/src/router.js:225:21)
    at /Users/jhoon/code/nango/node_modules/router/index.js:295:15
    at param (/Users/jhoon/code/nango/node_modules/router/index.js:600:14)
    at param (/Users/jhoon/code/nango/node_modules/router/index.js:610:14)
    at param (/Users/jhoon/code/nango/node_modules/router/index.js:610:14)
    at processParams (/Users/jhoon/code/nango/node_modules/router/index.js:664:3)
```

jeremydhoon:
@hassan254-prog I've simplified things by moving to a short allowlist of trusted domains. We no longer have to specify a webhook secret.

@TBonnin I edited your query parameter discarding logic since this failed in a way that should have been predictable but was not caught by the Typescript compiler.

packages/server/lib/webhook/connectwise-psa-webhook-routing.ts
```diff
@@ -0,0 +1,151 @@
...
+ * @param keyUrl - The URL from the webhook's Metadata.key_url field (UNTRUSTED)
+ * @returns The validated key URL or an error
+ */
+function trustedKeyUrl(keyUrl: string): Result<string> {
+    try {
+        const keyUrlParsed = new URL(keyUrl);
+
+        // Only accept HTTPS protocol
+        if (keyUrlParsed.protocol !== 'https:') {
+            return Err(new Error('webhook_invalid_key_url', { cause: 'Key URL must use HTTPS protocol' }));
```
jeremydhoon, line 32: Just want to check @TBonnin you're okay with the behavior here of returning a 500 instead of a 401? If we want a 401 we'd have to define a new `NangoError` type.

packages/providers/providers.yaml
```diff
@@ -3143,6 +3144,8 @@ connectwise-psa-staging:
         - support
         - ticketing
     auth_mode: BASIC
+    webhook_routing_script: connectwisePsaWebhookRouting
+    webhook_user_defined_secret: true
```
hassan254-prog:
```suggestion
    
```
You can omit this since no **webhook secret** is required.

packages/server/lib/webhook/connectwise-psa-webhook-routing.ts
```diff
@@ -0,0 +1,151 @@
...
+    if (!validateSignature(signingKey.value, signature, rawBody)) {
+        return Err(new Error('webhook_invalid_signature', { cause: 'Signature validation failed' }));
+    }
+
+    const response = await nango.executeScriptForWebhooks({
+        body,
+        webhookType: 'Type' // ConnectWise webhook type field
+    });
+
+    const connectionId = response?.connectionIds?.[0];
```
hassan254-prog: If we can’t retrieve a value from the webhook that can be used to identify a connection, I suggest returning all of them instead of selecting only the first one.

jeremydhoon: Oh I see -- apologies. Yes let me do that.

hassan254-prog (approved):

PR #4952, opened 2025-11-03
feat(fleet): add idle timeout to node config and overrides

rossmcewan:
---

**Introduce per-node idle timeout (`idleMaxDurationMs`) across Fleet**

This PR adds first-class support for a configurable idle timeout at node level. It extends the DB schema, domain models, supervisor logic, node providers and public type packages so each node (or override) can specify `idleMaxDurationMs`, replacing the former global constant. A new migration seeds existing rows, default values are injected by every provider, and CRUD paths/tests are updated accordingly.

<details>

• DB migration `packages/fleet/lib/db/migrations/20251103174100_add_idle_timeout.ts` adds `idle_max_duration_ms` to `nodes` (NOT NULL, default `1800000`) and `node_config_overrides` (nullable).
• Updated domain entities (`Node`, `NodeConfigOverride`, `NodeConfig`) and generated typings in `packages/types` to include `idleMaxDurationMs`.
• Extended `packages/fleet/lib/models` CRUD to read/write the new column; updated helper utilities and integration tests.
• Supervisor (`packages/fleet/lib/supervisor/supervisor.ts`) now considers `idleMaxDurationMs` when deciding to OUTDATE nodes and when creating nodes from overrides.
• All node providers (`noop`, `local`, `kubernetes`, `render`) expose the new default and propagate it to runner env via `IDLE_MAX_DURATION_MS`.
• Fleet service (`packages/fleet/lib/fleet.ts`) updates equality check in `overrideNodeConfig()` to account for the new field.
• Large test refresh to assert the additional attribute everywhere.

</details>

<details>

• `nodes` and `node_config_overrides` tables
• `packages/fleet/lib/models/*`
• `packages/fleet/lib/supervisor`
• `packages/fleet/lib/fleet.ts`
• All node provider implementations
• Public `@nangohq/types` package

</details>

---

### DIFF
--- a/packages/fleet/lib/db/migrations/20251103174100_add_idle_timeout.ts
+++ b/packages/fleet/lib/db/migrations/20251103174100_add_idle_timeout.ts
@@ -4,9 +4,9 @@ import { NODE_CONFIG_OVERRIDES_TABLE } from '../../models/node_config_overrides.
 
 export async function up(knex: Knex): Promise<void> {
     await knex.raw(`ALTER TABLE ${NODES_TABLE}
-      ADD COLUMN IF NOT EXISTS idle_timeout int NOT NULL DEFAULT 1800`);
+      ADD COLUMN IF NOT EXISTS idle_timeout_secs int NOT NULL DEFAULT 1800`);
     await knex.raw(`ALTER TABLE ${NODE_CONFIG_OVERRIDES_TABLE}
-      ADD COLUMN IF NOT EXISTS idle_timeout int DEFAULT NULL`);
+      ADD COLUMN IF NOT EXISTS idle_timeout_secs int DEFAULT NULL`);
 }
 
 export async function down(): Promise<void> {}
--- a/packages/fleet/lib/db/migrations/20251103174100_add_idle_timeout.ts
+++ b/packages/fleet/lib/db/migrations/20251103174100_add_idle_timeout.ts
@@ -6,7 +6,7 @@ export async function up(knex: Knex): Promise<void> {
     await knex.raw(`ALTER TABLE ${NODES_TABLE}
       ADD COLUMN IF NOT EXISTS idle_timeout int NOT NULL DEFAULT 1800`);
     await knex.raw(`ALTER TABLE ${NODE_CONFIG_OVERRIDES_TABLE}
-      ADD COLUMN IF NOT EXISTS idle_timeout int NULLABLE DEFAULT NULL`);
+      ADD COLUMN IF NOT EXISTS idle_timeout int DEFAULT NULL`);
 }
 
 export async function down(): Promise<void> {}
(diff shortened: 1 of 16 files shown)


packages/fleet/lib/db/migrations/20251103174100_add_idle_timeout.ts
```diff
@@ -0,0 +1,12 @@
+import type { Knex } from 'knex';
+import { NODES_TABLE } from '../../models/nodes.js';
+import { NODE_CONFIG_OVERRIDES_TABLE } from '../../models/node_config_overrides.js';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.raw(`ALTER TABLE ${NODES_TABLE}
+      ADD COLUMN IF NOT EXISTS idle_timeout int NOT NULL DEFAULT 1800`);
```
TBonnin: let's stick to the convention to have a suffix that represent the unit `idle_timeout_secs`

packages/jobs/lib/runner/kubernetes.ts
TBonnin, line 411: you need to pass the `node.idleTimeoutSecs` * 1000` here now

packages/fleet/lib/supervisor/supervisor.ts
```diff
@@ -247,7 +250,8 @@ export class Supervisor {
                 // Timeout IDLE nodes if they are taking too long (nodeProvider probably failed to terminate the node)
                 plan.push(
                     ...(nodes.IDLE || []).flatMap<Operation>((node) => {
-                        if (Date.now() - node.lastStateTransitionAt.getTime() > STATE_TIMEOUT_MS.IDLE) {
+                        const idleTimeout = node.idleTimeoutSecs ? node.idleTimeoutSecs * 1000 : STATE_TIMEOUT_MS.IDLE;
+                        if (Date.now() - node.lastStateTransitionAt.getTime() > idleTimeout) {
```
TBonnin: I think there is a confusion here. This condition is to terminate the node once it is already in idle state, meaning it has already notified fleet supervisor that it as stopped processing any task. The timeout is to enforce cleaning up idle nodes even if the termination operation failed (network issue, etc...). We shouldn't modified this behavior

TBonnin: it might be confusing to call this new node attribute idle timeout though. The env var controlling the idling behavior is called `IDLE_MAX_DURATION_MS`. Maybe we call this new field `idleMaxDurationMs` or `idleMaxDurationSecs`

rossmcewan: You're right. There was confusion. I have renamed the field and applied it in the correct place.

packages/jobs/lib/runner/render.ts
```diff
@@ -56,7 +57,7 @@ export const renderNodeProvider: NodeProvider = {
                     { key: 'NODE_OPTIONS', value: `--max-old-space-size=${Math.floor((node.memoryMb / 4) * 3)}` },
                     { key: 'RUNNER_NODE_ID', value: `${node.id}` },
                     { key: 'RUNNER_URL', value: `http://${name}` },
-                    { key: 'IDLE_MAX_DURATION_MS', value: `${25 * 60 * 60 * 1000}` }, // 25 hours
+                    { key: 'IDLE_MAX_DURATION_MS', value: `${node.idleMaxDurationMs}` }, // 25 hours
```
TBonnin:
```suggestion
                    { key: 'IDLE_MAX_DURATION_MS', value: `${node.idleMaxDurationMs}` },
```

TBonnin (approved):

PR #4955, opened 2025-11-03
fix: create an billing subscription when signup via Google

TBonnin:
Signing up by email not only trigger the creation of a customer in Orb but also create a subscription for it. This logic is missing when signing up via Google auth.

 

---

**Add missing billing subscription creation on Google-based signup**

Extends the Google OAuth signup flow so that, in addition to creating an Orb customer, it also creates an initial free subscription. The change mirrors the behaviour already present for email-based signups by invoking the new helper `linkBillingFreeSubscription()` after successfully linking the Orb customer.

<details>

• Added import of `linkBillingFreeSubscription` in `packages/server/lib/controllers/v1/account/managed/getCallback.ts`
• Replaced single call to `linkBillingCustomer()` with two-step flow: `linkBillingCustomer()` followed by `linkBillingFreeSubscription()` when the first succeeds
• Introduced separate error handling/reporting blocks for both customer and subscription linking results

</details>

<details>

• `getCallback.ts` (Google managed signup controller)
• Billing utility integration (`utils/billing.js` import)

</details>

---

### DIFF
--- a/packages/server/lib/controllers/v1/account/managed/getCallback.ts
+++ b/packages/server/lib/controllers/v1/account/managed/getCallback.ts
@@ -6,7 +6,7 @@ import { basePublicUrl, flagHasUsage, getLogger, nanoid, report } from '@nangohq
 
 import { getWorkOSClient } from '../../../../clients/workos.client.js';
 import { asyncWrapper } from '../../../../utils/asyncWrapper.js';
-import { linkBillingCustomer } from '../../../../utils/billing.js';
+import { linkBillingCustomer, linkBillingFreeSubscription } from '../../../../utils/billing.js';
 
 import type { InviteAccountState } from './postSignup.js';
 import type { DBInvitation, DBTeam, GetManagedCallback } from '@nangohq/types';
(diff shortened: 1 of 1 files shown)


packages/server/lib/controllers/v1/account/managed/getCallback.ts
```diff
@@ -118,9 +118,14 @@ export const getManagedCallback = asyncWrapper<GetManagedCallback>(async (req, r
         }
 
         if (isNewTeam && flagHasUsage) {
-            const res = await linkBillingCustomer(account, user);
-            if (res.isErr()) {
-                report(res.error);
+            const linkOrbCustomerRes = await linkBillingCustomer(account, user);
```
khaliqgant, line 121: This logic seems a bit gnarly and nested / hard to reason about. Any way to simplify?

TBonnin: yes. I should have mentioned this is a quickfix to ensure subscription are properly being associated with new users. At the moment they are not. I have just duplicated the logic from email signup https://github.com/NangoHQ/nango/blob/master/packages/server/lib/controllers/v1/account/signup.ts#L114-L122 
I don't know if this was an oversight or done on purpose. I need to sync up with @kaposke when he is back.

khaliqgant (approved):

PR #4962, opened 2025-11-05
fix(cli): add normalized hash fallback for mock file lookup

hassan254-prog:

- add normalized hash fallback for mock file lookup

---

**CLI test-mock lookup now retries with normalized hash**

Adds a fallback flow in `getMockFile` so that when a mock file keyed by the strict SHA-1 of the request identity is not found, the code retries using a "normalized" hash where every param value is coerced to string. This prevents lookup failures caused by numeric vs string representation mismatches in query/body params.

<details>

• Extended `getMockFile` to catch lookup errors and attempt a second read using a normalized hash
• Introduced helper `computeNormalizedHash(requestIdentity)` that stringifies param values before hashing
• Adjusted fallback path construction to avoid missing slash and mirror original directory layout

</details>

<details>

• `packages/cli/lib/testMocks/utils.ts`

</details>

---

### DIFF
--- a/packages/cli/lib/testMocks/utils.ts
+++ b/packages/cli/lib/testMocks/utils.ts
@@ -86,8 +86,17 @@ class NangoActionMock {
         const hashBasedPath = `${dir}/${identity.requestIdentityHash}`;
 
         if (await this.hashDirExists(dir)) {
-            const data = await this.getMockFile(hashBasedPath, true, identity);
-            return data;
+            try {
+                const data = await this.getMockFile(hashBasedPath, true, identity);
+                return data;
+            } catch (err) {
+                const normalizedHash = computeNormalizedHash(identity.requestIdentity);
+                if (normalizedHash !== identity.requestIdentityHash) {
+                    const data = await this.getMockFile(`${dir}${normalizedHash}`, true, identity);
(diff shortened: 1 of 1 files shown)


packages/cli/lib/testMocks/utils.ts
```diff
@@ -86,8 +86,17 @@ class NangoActionMock {
...
+                const data = await this.getMockFile(hashBasedPath, true, identity);
+                return data;
+            } catch (err) {
+                const normalizedHash = computeNormalizedHash(identity.requestIdentity);
+                if (normalizedHash !== identity.requestIdentityHash) {
+                    const data = await this.getMockFile(`${dir}${normalizedHash}`, true, identity);
+                    return data;
+                }
+                throw err;
+            }
```
TBonnin: would it make sense to move all of this fallback logic inside `getMockFile`?

hassan254-prog: yes, I've updated this with the latest [commit](https://github.com/NangoHQ/nango/pull/4962/commits/b5c2e19ddb758c4cecf5261b57d4afe7ad86e28f).

TBonnin (approved):

PR #4987, opened 2025-11-11
feat(dryrun-diagnostics): add diagnostics service for dryrun

khaliqgant:
---

**Add `--diagnostics` flag with CPU/Memory monitoring to `dryrun` command**

The PR introduces an opt-in lightweight diagnostics facility for the CLI `dryrun` path. When the new `--diagnostics` flag is present, a `DiagnosticsMonitor` is instantiated before user code executes and stopped afterwards, producing a formatted summary of average and peak RSS/heap/external memory plus CPU utilisation. The feature is fully plumbed through the public `DryRunService`, its private `runScript()` helper, and the CLI entry-point without altering default behaviour when the flag is omitted.

<details>

• Created `packages/cli/lib/services/diagnostics-monitor.service.ts` implementing `DiagnosticsMonitor`, adaptive 50-100 ms sampling, bounded sample buffer (`MAX_SAMPLES = 5000`), statistics calculators and `formatDiagnostics()` pretty printer
• Extended `RunArgs`, `runScript()` and `DryRunService.run()` to accept optional `diagnostics?: boolean` and to start/stop the monitor around script execution
• Added CLI flag registration in `packages/cli/lib/index.ts` for top-level and `dryrun` sub-command
• Inserted report printing in `finally` blocks so summary always appears even on error paths
• Addressed review feedback by bounding sample array growth

</details>

<details>

• `packages/cli/lib/services/diagnostics-monitor.service.ts`
• `packages/cli/lib/services/dryrun.service.ts`
• `packages/cli/lib/index.ts`

</details>

---

### DIFF
--- a/packages/cli/lib/services/diagnostics-monitor.service.ts
+++ b/packages/cli/lib/services/diagnostics-monitor.service.ts
@@ -1,0 +1,246 @@
+import chalk from 'chalk';
+
+interface MemorySample {
+    rss: number;
+    heapUsed: number;
+    heapTotal: number;
+    external: number;
+}
+
+interface CPUSample {
+    user: number;
+    system: number;
+}
+
+interface Sample {
+    timestamp: number;
+    memory: MemorySample;
+    cpu: CPUSample;
+}
+
+interface MemoryStats {
+    avg: number;
+    peak: number;
+}
+
+interface CPUStats {
+    avg: number;
+    peak: number;
+}
+
+export interface DiagnosticsStats {
+    duration: number;
+    memory: {
+        rss: MemoryStats;
+        heapUsed: MemoryStats;
+        heapTotal: MemoryStats;
+        external: MemoryStats;
+    };
+    cpu: CPUStats;
+}
+
+export class DiagnosticsMonitor {
+    private samples: Sample[] = [];
+    private startTime: number = 0;
+    private startCpu: CPUSample = { user: 0, system: 0 };
+    private intervalId: NodeJS.Timeout | null = null;
+    private currentInterval: number = 100; // Start with 100ms baseline
+    private lastSample: Sample | null = null;
+    private stableCount: number = 0;
+
+    start(): void {
+        this.startTime = Date.now();
+        this.startCpu = process.cpuUsage();
+        this.takeSample(); // Initial sample
+        this.scheduleNextSample();
+    }
+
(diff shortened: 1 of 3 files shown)


rossmcewan (commented):
Thoughts on measuring the memory of the VM as opposed to the current process (or maybe in addition to)?

https://nodejs.org/api/vm.html#vmmeasurememoryoptions

khaliqgant:
> Thoughts on measuring the memory of the VM as opposed to the current process (or maybe in addition to)?
> 
> https://nodejs.org/api/vm.html#vmmeasurememoryoptions

Yeah, good thought. Ran into some errors when trying to add that in. I think for a v1 this is sufficient to just offer _some_ insight into the sync metrics. wdyt?

rossmcewan (approved):

packages/cli/lib/services/diagnostics-monitor.service.ts
```diff
@@ -0,0 +1,246 @@
...
+                heapTotal: memory.heapTotal,
+                external: memory.external
+            },
+            cpu: {
+                user: cpu.user,
+                system: cpu.system
+            }
+        };
+
+        this.samples.push(sample);
```
TBonnin, line 101: should we have an upper bound limit to prevent `samples` to grow indefinitely?

packages/cli/lib/services/diagnostics-monitor.service.ts
```diff
@@ -0,0 +1,246 @@
...
+        }
+
+        // Check CPU change (approximate percentage)
+        const prevCpuTotal = prev.cpu.user + prev.cpu.system;
+        const currentCpuTotal = current.cpu.user + current.cpu.system;
+        const timeDiff = current.timestamp - prev.timestamp;
+
+        if (timeDiff > 0) {
+            const prevCpuPercent = (prevCpuTotal / (timeDiff * 1000)) * 100;
+            const currentCpuPercent = (currentCpuTotal / (timeDiff * 1000)) * 100;
```
TBonnin: can we have all variables in this file have a suffix with their unit and maybe transform all time variables to `ms` and memory ones to `bytes`. 
Otherwise this kind of formula is quite hard to understand. I had to lookup the unit of `process.cpuUsage` for instance to realize it was in microseconds while timeDiff is in millis

TBonnin (approved):

PR #4994, opened 2025-11-12
feat(webhooks): add support for google calendar webhooks

hassan254-prog:

- add support for google calendar webhooks

---

**Add first-class Google Calendar webhook support & improve generic webhook manager**

This PR introduces native routing for Google Calendar push notifications, enabling real-time event sync for the Google Calendar connector and reducing operational noise from verification pings. A new routing script extracts the account identifier from incoming Google headers, integrates it with the universal webhook execution flow, and immediately acknowledges delivery. The generic webhook manager has also been hardened to gracefully handle header-only health-check calls from Google by returning HTTP 204 instead of logging errors or returning 500s. Provider registration and exports have been updated so the new router is automatically discovered at runtime.

<details>

• New google-calendar-webhook-routing.ts that parses `x-goog-resource-uri`, derives end-user email, invokes `nango.executeScriptForWebhooks()`, and returns 200 OK
• providers.yaml updated with `webhook_routing_script: googleCalendarWebhookRouting` for the google-calendar provider
• webhook/index.ts export added for the new routing script so it is auto-loaded by the server
• webhook.manager.ts now short-circuits with HTTP 204 when both body and headers are empty (Google subscription ping safeguard)

</details>

<details>

• webhook routing layer
• provider registry (providers.yaml)
• generic webhook manager
• server bootstrap (webhook/index.ts)

</details>

---


packages/server/lib/webhook/webhook.manager.ts
```diff
@@ -38,7 +38,11 @@ export async function routeWebhook({
     rawBody: string;
     logContextGetter: LogContextGetter;
 }): Promise<WebhookResponse> {
-    if (!body) {
+    // Check if both body and headers are empty
+    const hasBody = body && (typeof body === 'object' ? Object.keys(body).length > 0 : true);
+    const hasHeaders = headers && Object.keys(headers).length > 0;
+
+    if (!hasBody && !hasHeaders) {
```
TBonnin, line 45: that's a change of behavior that is gonna impact all providers. Are we sure this is not gonna break anything else?

TBonnin (approved):

packages/server/lib/webhook/webhook.manager.ts
hassan254-prog: The issue here is that we’ll now return a 200 status code even for webhooks with an empty body, which wasn’t the case before, we previously returned 204. To avoid changing this behavior, we could consider adding another property in the providers.yaml to control this behavior, wdyt?

TBonnin: I will keep things simple. I think the provider except a successful response. I don't think they care about 200 vs 204

PR #4996, opened 2025-11-12
feat(cli-nango-test): nango test implementation

khaliqgant:
---

**CLI: Introduce `nango test` command for local & remote integration testing with CI-ready reporting**

This PR adds a first-class end-to-end testing workflow to the Nango CLI by introducing the `nango test` command. The new command discovers and runs exported test functions in an integration workspace, either locally in the sandbox or remotely against a deployed Nango environment (via a new `/v1/tests` backend endpoint). Results are captured in a structured format, summarized in a Jest-style report, and the CLI now exits with the number of failed tests—making it immediately usable as a quality gate in CI pipelines.

Supporting work includes refactoring shared execution logic out of the dry-run and response-saver services, extending the SDK for remote test invocation, adding unit tests and dev dependencies, updating provider-validation scripts to invoke the new command, and documenting how to write and run integration tests. The change set is purely additive (no breaking API changes) but requires a backend version that exposes the new endpoint for remote execution.

---

### DIFF
--- a/docs/implementation-guides/actions/implement-an-action.mdx
+++ b/docs/implementation-guides/actions/implement-an-action.mdx
@@ -156,6 +156,10 @@ You can also pass input data to the action. Run `nango dryrun --help` to see all
     By default, `dryrun` retrieves connections from your `dev` environment. You can change this with a CLI flag.
 </Tip>
 
+<Tip>
+    You can also use `nango test` to run your action with output validation, or `nango test --local` for fast testing with mocked API responses. See the [Test functions guide](/implementation-guides/building-integrations/test-functions) for more details.
+</Tip>
+
 ### Step 6 - Deploy your action
 
 To run your action in Nango, you need to deploy it to an environment in your Nango account.
--- a/docs/implementation-guides/building-integrations/test-functions.mdx
+++ b/docs/implementation-guides/building-integrations/test-functions.mdx
@@ -1,0 +1,224 @@
+---
+title: 'Test functions'
+sidebarTitle: 'Test functions'
+description: 'How to test your syncs and actions locally'
+---
+
+Testing your integration functions ensures they work correctly before deploying to production. Nango provides two testing modes:
+
+- **Remote testing** (default): Tests with real API calls and validates output against your schema
+- **Local testing**: Tests with mocked API responses for fast, repeatable regression testing
+
+## Remote testing
+
+Remote testing executes your function making real API calls to verify it works end-to-end.
+
+### Run a remote test
+
+```bash
+nango test <script-name> <connection-id>
+```
+
+This command:
+1. Compiles your function
+2. Executes it against a real connection
+3. Validates output against your model schema
+4. Reports any validation errors
+
+Example:
+```bash
+nango test salesforce-contacts conn_123abc
+```
+
+<Tip>
+By default, tests use connections from your `dev` environment. Change this with the `-e` flag.
+</Tip>
+
+### Pass input to actions
+
+For actions that require input, use the `--input` flag:
+
+```bash
+nango test create-contact conn_123abc --input '{"email": "test@example.com"}'
+```
+
+You can also load input from a file:
+```bash
+nango test create-contact conn_123abc --input @fixtures/contact.json
+```
+
+## Local testing
+
+Local testing uses mocked API responses to test your functions without making real API calls. This is useful for:
+
+- Fast iteration during development
+- Regression testing in CI/CD
+- Testing without rate limits
+- Reproducible test results
+
+### Set up local testing
+
+**Step 1: Capture API responses**
+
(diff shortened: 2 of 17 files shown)


docs/implementation-guides/building-integrations/test-functions.mdx
```diff
@@ -0,0 +1,224 @@
...
+- Reproducible test results
+
+### Set up local testing
+
+**Step 1: Capture API responses**
+
+First, save real API responses from a working connection:
+
+```bash
+nango dryrun <script-name> <connection-id> --save-responses
```
TBonnin, line 66: should we have an alias `nango test <script-name> <connection-id> --record` to avoid having to juggle between different subcommand
or even more simple to have the test command to record API responses if they don't exist yet.
```
// 1. records API responses if they don't exist
nango test <script-name> <connection-id> 

// 2. use the saved API responses by default
nango test <script-name> <connection-id> 

// 3. option to force saving/overwriting API responses
nango test <script-name> <connection-id> --record
```

docs/implementation-guides/building-integrations/test-functions.mdx
```diff
@@ -0,0 +1,224 @@
...
+echo '{"id": "123", "name": "John Doe"}' > nango-integrations/<integration>/mocks/<action-name>/output.json
+```
+
+For syncs, create expected batch save/delete files:
+```bash
+# Create the model directory
+mkdir -p nango-integrations/<integration>/mocks/<sync-name>/<ModelName>
+
+# Add expected records
+echo '[{"id": "1", "name": "Contact 1"}]' > nango-integrations/<integration>/mocks/<sync-name>/<ModelName>/batchSave.json
```
TBonnin, line 106: I think this file should be created automatically when running recording the real API response. It could contains some default value that will always make the test fail like `{ status: 'not implemented yet' }`

docs/implementation-guides/building-integrations/test-functions.mdx
```diff
@@ -0,0 +1,224 @@
...
+- **Local testing**: Tests with mocked API responses for fast, repeatable regression testing
+
+## Remote testing
+
+Remote testing executes your function making real API calls to verify it works end-to-end.
+
+### Run a remote test
+
+```bash
+nango test <script-name> <connection-id>
```
TBonnin, line 19: would it make more sense to have the default mode to be using saved api responses?

docs/implementation-guides/actions/implement-an-action.mdx
```diff
@@ -156,6 +156,10 @@ You can also pass input data to the action. Run `nango dryrun --help` to see all
     By default, `dryrun` retrieves connections from your `dev` environment. You can change this with a CLI flag.
 </Tip>
 
+<Tip>
+    You can also use `nango test` to run your action with output validation, or `nango test --local` for fast testing with mocked API responses. See the [Test functions guide](/implementation-guides/building-integrations/test-functions) for more details.
```
TBonnin, line 160: I don't think we should used the work mocks. They are real API responses that have been saved/recorded, not faked

docs/implementation-guides/building-integrations/test-functions.mdx
```diff
@@ -0,0 +1,224 @@
+---
+title: 'Test functions'
+sidebarTitle: 'Test functions'
+description: 'How to test your syncs and actions locally'
+---
+
+Testing your integration functions ensures they work correctly before deploying to production. Nango provides two testing modes:
+
+- **Remote testing** (default): Tests with real API calls and validates output against your schema
+- **Local testing**: Tests with mocked API responses for fast, repeatable regression testing
```
TBonnin, line 10: is `local vs remote` the best names? what about `offline vs live`? or `replay vs live`

docs/implementation-guides/building-integrations/test-functions.mdx
```diff
@@ -0,0 +1,224 @@
...
+nango dryrun salesforce-contacts conn_123abc --save-responses
+```
+
+**Test both success and error cases**
+
+Create separate mocks for error scenarios to ensure your error handling works correctly.
+
+**Use meaningful connection IDs**
+
+In local mode, the connection ID doesn't need to be real, but use descriptive names:
```
TBonnin, line 196: I find that super confusing.
It would make more sense to not ask for a connection id unless in live/recording mode. no?

docs/implementation-guides/building-integrations/test-functions.mdx
TBonnin: thinking more about it and since connection is only required for recording or live mode, what about:
```
nango test <script-name> // run with recorded responses. Error if no recordings
nango test <script-name> --record <connection-id> // save/overwrite recordings and run test with them
nango test <script-name> --live <connection-id> // run test calling Nango live API
```

PR #4998, opened 2025-11-12
feat(lambda): lambda runtime for function execution

rossmcewan:
---

**Add AWS Lambda Runtime and Multi-Fleet Orchestration for Function Execution**

Introduces a Lambda-based execution path alongside the existing container runner, with a pluggable runtime selection layer driven by plan metadata and rules. Adds a dedicated `@nangohq/lambda-runner` package, extends fleet supervision to support multiple fleets and Lambda-specific node providers, and updates infrastructure (Redis, plans schema, deployment workflow) to support the new runtime while maintaining backward compatibility.

<details>

• Adds Lambda runtime adapter, node provider, and handler (`packages/jobs/lib/runtime/lambda.adapter.ts`, `packages/jobs/lib/runner/lambda.ts`, `packages/lambda-runner/lib/index.ts`) for invoking AWS Lambda functions asynchronously with locking, heartbeat, and abort handling.
• Introduces runtime orchestration layer (`packages/jobs/lib/runtime/runtimes.ts`, `packages/jobs/lib/runtime/runtimes.rules.ts`) that selects fleets based on JSON rules and per-plan runtime flags added via migration `20260108145100_add_function_runtime_flags.cjs` and plan schema updates.
• Extends fleet management to support multiple fleets and Lambda nodes (new image verifiers, supervisor updates, `fleet_id` persistence, node provider API changes) and separates Redis usage into system/customer boundaries with locking enhancements (`packages/kvstore/lib/index.ts`, `packages/kvstore/lib/Locking.ts`).
• Adds new `@nangohq/lambda-runner` package with build pipeline (`Dockerfile.lambda`, tsconfig, schema validations) and updates CI/CD workflow to deploy Lambda images and invoke rollout API with image type.
• Updates job execution paths (sync/action/webhook/on-event) to pass runtime context and use new runtime adapter, while ensuring abort propagation through Redis flags and Lambda heartbeat polling.

</details>

<details>

• @nangohq/jobs runtime orchestration
• @nangohq/fleet supervision and node models
• @nangohq/kvstore locking and Redis clients
• @nangohq/server fleet rollout API
• @nangohq/utils environment parsing and detection
• @nangohq/lambda-runner package and deployment workflow
• Database schemas for plans and fleet nodes

</details>

---


packages/fleet/lib/models/node_config_overrides.ts
```diff
@@ -19,6 +19,8 @@ interface DBNodeConfigOverride {
     readonly is_tracing_enabled: boolean | null;
     readonly is_profiling_enabled: boolean | null;
     readonly idle_max_duration_ms: number | null;
+    readonly execution_timeout_secs: number | null;
+    readonly provisioned_concurrency: number | null;
     readonly created_at: Date;
```
rossmcewan, line 24: These are required for function configuration. 
- `execution_timeout_secs` is how long a function is allowed to run before it is terminated by the runtime
- `provisioned_concurrency` is a Lambda specific feature that helps mitigate cold-starts. It's configurable as we would likely want different values for different function profiles.

packages/fleet/lib/models/nodes.ts
```diff
@@ -54,6 +55,8 @@ interface DBNode {
     readonly is_tracing_enabled: boolean;
     readonly is_profiling_enabled: boolean;
     readonly idle_max_duration_ms: number | null;
+    readonly execution_timeout_secs: number | null;
+    readonly provisioned_concurrency: number | null;
```
rossmcewan, line 59: These are required for function configuration.

- `execution_timeout_secs` is how long a function is allowed to run before it is terminated by the runtime
- `provisioned_concurrency` is a Lambda specific feature that helps mitigate cold-starts. It's configurable as we would likely want different values for different function profiles.

packages/fleet/lib/node-providers/node_provider.ts
```diff
@@ -7,4 +7,6 @@ export interface NodeProvider {
     start(node: Node): Promise<Result<void>>;
     terminate(node: Node): Promise<Result<void>>;
     verifyUrl(url: string): Promise<Result<void>>;
+    finish(node: Node): Promise<Result<void>>;
```
rossmcewan, line 10: The way nodes are `finished` was specific to instance-based nodes (i.e. Render compute instances or Kubernetes pods). Offloading this capability to the node provider allows us to have custom functionality for different kinds of nodes.

packages/fleet/lib/node-providers/node_provider.ts
```diff
@@ -7,4 +7,6 @@ export interface NodeProvider {
...
+    waitUntilHealthy(options: { nodeId: number; url: string; timeoutMs: number }): Promise<Result<void>>;
```
rossmcewan, line 11: Similar to `finishing` a node, the concept of `waiting until healthy` was also specific to instance-based nodes. Having this functionality encapsulated in the node provider allows for different implementations.

packages/fleet/lib/supervisor/operation.ts
```diff
@@ -2,7 +2,7 @@ import type { Node } from '../types.js';
 import type { Deployment, NodeConfig } from '@nangohq/types';
 
 export type Operation =
-    | { type: 'CREATE'; routingId: Node['routingId']; deployment: Deployment; nodeConfig?: NodeConfig | undefined }
+    | { type: 'CREATE'; routingId: Node['routingId']; deployment: Deployment; nodeConfig?: NodeConfig | undefined; fleetId: string }
```
rossmcewan, line 5: The fleet id has been added to a node so it is required to be passed in during the create. 
Having the node aware of the fleet id allows for a multi-fleet runtime.

packages/fleet/lib/supervisor/supervisor.ts
```diff
@@ -41,11 +41,13 @@ export class Supervisor {
     private state: SupervisorState = 'stopped';
     private dbClient: DatabaseClient;
     private tickCancelled: boolean = false;
+    private fleetId: string;
     public nodeProvider: NodeProvider;
 
-    constructor({ dbClient, nodeProvider }: { dbClient: DatabaseClient; nodeProvider: NodeProvider }) {
+    constructor({ dbClient, nodeProvider, fleetId }: { dbClient: DatabaseClient; nodeProvider: NodeProvider; fleetId: string }) {
```
rossmcewan, line 47: The supervisor needs to be aware of which fleet it is supervising. This is propagated to nodes when they are created and is used to create a dynamic lock key to perform fleet operations.

packages/fleet/lib/utils/errors.ts
```diff
@@ -27,6 +27,8 @@ type FleetErrorCode =
     | 'fleet_misconfigured'
     | 'fleet_rollout_invalid_image'
     | 'fleet_rollout_image_not_found'
+    | 'fleet_rollout_image_verification_failed'
+    | 'fleet_rollout_invalid_image_type'
```
rossmcewan, line 31: An `image type` was introduced to the rollout operation so that we can support additional repositories. Currently values are either `ecr` or `docker`.

packages/fleet/lib/fleet.ts
```diff
@@ -29,14 +31,17 @@ export class Fleet {
...
     constructor({ fleetId, dbUrl = defaultDbUrl, nodeProvider }: { fleetId: string; dbUrl?: string | undefined; nodeProvider?: NodeProvider }) {
         this.fleetId = fleetId;
         this.dbClient = new DatabaseClient({ url: dbUrl, schema: fleetId });
         if (nodeProvider) {
-            this.supervisor = new Supervisor({ dbClient: this.dbClient, nodeProvider: nodeProvider });
+            this.supervisor = new Supervisor({ dbClient: this.dbClient, nodeProvider: nodeProvider, fleetId: this.fleetId });
         }
         this.nodeProvider = nodeProvider || noopNodeProvider;
+        this.imageVerifiers.set('docker', new DockerImageVerifier());
+        this.imageVerifiers.set('ecr', new ECRImageVerifier());
```
rossmcewan, line 44: Image verifiers are configured per image type.

packages/jobs/lib/runner/lambda.ts
```diff
@@ -0,0 +1,195 @@
...
+
+    static getInstance(): Lambda {
+        if (!Lambda.instance) {
+            Lambda.instance = new Lambda();
+        }
+        return Lambda.instance;
+    }
+
+    async createFunction(node: Node): Promise<Result<void>> {
+        setImmediate(async () => {
```
rossmcewan, line 63: This is performed in the background. A potential optimization here is to offload this to some kind of durable process (e.g. durable Lambda, StepFunction, or even a Kubernetes Job with durable capabilities).

packages/jobs/lib/runner/lambda.ts
```diff
@@ -0,0 +1,195 @@
...
+                    Role: envs.LAMBDA_EXECUTION_ROLE_ARN,
+                    Code: {
+                        ImageUri: `${envs.LAMBDA_ECR_REGISTRY}/${node.image}`
+                    },
+                    PackageType: 'Image',
+                    Architectures: [envs.LAMBDA_ARCHITECTURE],
+                    MemorySize: getSize(node),
+                    Timeout: node.executionTimeoutSecs || envs.LAMBDA_EXECUTION_TIMEOUT_SECS,
+                    Environment: this.getEnvironmentVariables(node),
+                    VpcConfig: {
```
rossmcewan, line 77: The function is connected to the same VPC that the cluster is in - this allows access to the service endpoints without having to traverse the internet.

packages/jobs/lib/runner/lambda.ts
```diff
@@ -0,0 +1,195 @@
...
+                        FunctionVersion: pvResult.Version
+                    })
+                );
+                await lambdaClient.send(
+                    new PutProvisionedConcurrencyConfigCommand({
+                        FunctionName: pvResult.FunctionName,
+                        Qualifier: envs.LAMBDA_FUNCTION_ALIAS,
+                        ProvisionedConcurrentExecutions: node.provisionedConcurrency || envs.LAMBDA_PROVISIONED_CONCURRENCY
+                    })
+                );
```
rossmcewan, line 131: We don't wait for the provisioned concurrency operation to complete as it can take a really long time and is not necessary to begin invocations. This process will complete in the background. We can monitor performance during function updates to see if this is necessary.

packages/jobs/lib/runner/lambda.ts
```diff
@@ -0,0 +1,195 @@
...
+                const fleetId = node.fleetId || envs.RUNNER_LAMBDA_FLEET_ID;
+                const result = await registerWithFleet(fleetId, {
+                    nodeId: node.id,
+                    url: `${aResult.AliasArn}`
+                });
```
rossmcewan, line 136: Instance-based runners report their readiness to the `jobs-service` when they have started - this is not inline with how function-based runners will work so the provider itself will register the created function with the fleet when all creation operations have completed.

packages/jobs/lib/runner/lambda.ts
```diff
@@ -0,0 +1,195 @@
...
+        };
+    }
+
+    async terminateFunction(node: Node): Promise<Result<void>> {
+        const name = getFunctionName(node);
+        await lambdaClient.send(
+            new DeleteFunctionCommand({
+                FunctionName: name
+            })
+        );
```
rossmcewan, line 174: This has a cascading effect and removes all associated resources like aliases and provisioned concurrency settings. It has no effect on running functions so is safe to perform while invocations are in flight.

packages/jobs/lib/runner/render.ts
rossmcewan, line 1: Should we remove this file?

packages/jobs/lib/runtime/lambda.adapter.ts
```diff
@@ -0,0 +1,63 @@
...
+        const command = new InvokeCommand({
+            FunctionName: func.arn,
+            Payload: JSON.stringify({
+                taskId: params.taskId,
+                nangoProps: params.nangoProps,
+                code: Buffer.from(params.code).toString('base64'),
+                codeParams: params.codeParams
+            }),
+            InvocationType: 'Event'
+        });
```
rossmcewan, line 46: This uses an `Event` invocation so will run asynchronously.

packages/jobs/lib/runtime/runtimes.rules.ts
```diff
@@ -0,0 +1,33 @@
...
+
+const rules = envs.RUNNER_FLEET_RULES.length
+    ? envs.RUNNER_FLEET_RULES
+    : [
+          {
+              priority: 0,
+              conditions: { all: [{ fact: 'always', operator: 'equal', value: true }] },
+              event: { type: envs.RUNNER_FLEET_ID }
+          }
+      ];
```
rossmcewan: This is safe operation so that the default fleet id is returned.

packages/jobs/lib/runtime/runtimes.ts
```diff
@@ -0,0 +1,87 @@
...
+import type { RuntimeAdapter } from './adapter.js';
+import type { NangoProps } from '@nangohq/types';
+import type { Result } from '@nangohq/utils';
+
+interface Runtime {
+    readonly adapter: RuntimeAdapter;
+    readonly fleet: Fleet;
+}
+
+const runtimes: Map<string, Runtime> = new Map<string, Runtime>();
```
rossmcewan, line 21: Runtimes are stored in a map keys on fleet id.

packages/jobs/lib/runtime/runtimes.ts
```diff
@@ -0,0 +1,87 @@
...
+    if (result.isErr()) {
+        logger.error('Error while getting fleet id for runtime', result.error);
+    } else {
+        if (result.value) {
+            const runtime = runtimes.get(result.value);
+            if (runtime) return Ok(runtime.adapter);
+        }
+    }
+    return Ok(defaultRuntime.adapter);
+}
```
rossmcewan, line 53: This uses the configured rules to determine the fleet id to use based on values in the `NangoProps`. If an error occurs or no value is returned from the rule evaluation, the default runtime adapter is returned (i.e. current behavior)

packages/server/lib/controllers/fleet/postRollout.ts
```diff
@@ -5,13 +5,23 @@ import { requireEmptyQuery, zodErrorToHTTP } from '@nangohq/utils';
 
 import { asyncWrapper } from '../../utils/asyncWrapper.js';
 
-import type { PostRollout } from '@nangohq/types';
+import type { ImageType, PostRollout } from '@nangohq/types';
 
 const bodyValidation = z
     .object({
-        image: z.string().regex(/^[a-z0-9-]+\/[a-z0-9-]+:[a-f0-9]{40}$/, { message: "Invalid image. Must be 'repository/image:commit'" })
+        imageType: z.enum(['docker', 'ecr']).optional(),
```
rossmcewan, line 12: We allow an `imageType` to be passed to `rollout`

packages/shared/lib/services/file/remote.service.ts
```diff
@@ -11,29 +11,52 @@ import { NangoError } from '../../utils/error.js';
...
+    const accessKeyId = process.env['AWS_INTEGRATIONS_ACCESS_KEY_ID'];
+    const secretAccessKey = process.env['AWS_INTEGRATIONS_SECRET_ACCESS_KEY'];
+    if (!accessKeyId || !secretAccessKey) {
+        return undefined;
+    }
+    return {
+        accessKeyId,
+        secretAccessKey
+    };
+}
```
rossmcewan, line 28: The names of the AWS credentials environment variables have to be changed because the permissions are provided to the runtime environment at an infrastructure level - keeping the environment variables named as the default (i.e. `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY_ID`) overrides any profile assumed by the runtime. The credential provider for AWS evaluates environment variables to have higher priority.

packages/utils/lib/environment/parse.ts
```diff
@@ -123,6 +125,19 @@ export const ENVS = z.object({
         .regex(/^[a-zA-Z0-9_-]+$/)
         .optional()
         .default('nango_runners'),
+    RUNNER_LAMBDA_FLEET_ID: z.string().optional().default('nango_runners_lambda'),
+    RUNNER_FLEET_RULES: z
```
rossmcewan: Rules are provided as an environment variable. This should be a JSON array of [json-rules-engine](https://github.com/cachecontrol/json-rules-engine) rules.

packages/utils/lib/environment/parse.ts
```diff
@@ -351,6 +372,45 @@ export const ENVS = z.object({
     NANGO_ACTIVEMQ_PASSWORD: z.string().optional().default('admin'),
     NANGO_ACTIVEMQ_CONNECT_TIMEOUT_MS: z.coerce.number().optional().default(10_000),
 
+    // Lambda
+    LAMBDA_ENABLED: z.stringbool().optional().default(false),
```
rossmcewan, line 383: Controls whether Lambda is enabled at all. If this is set to false, no Lambda capabilities can be used as neither the adapter nor the fleet is created.

packages/utils/lib/environment/parse.ts
```diff
@@ -351,6 +372,45 @@ export const ENVS = z.object({
...
+    LAMBDA_DEFAULT_SIZE: z.coerce.number().default(512),
```
rossmcewan, line 384: The default size for Lambda functions. This is used as part of the routing id.

packages/jobs/lib/runner/lambda.ts
```diff
@@ -0,0 +1,188 @@
...
+export const logger = getLogger('Lambda');
+
+const lambdaClient = new LambdaClient();
+
+export function getFunctionName(node: Node): string {
+    return `${node.routingId}-${node.id}`;
+}
+
+function getSize(node: Node): number {
+    return getSizeFromRoutingId(node.routingId);
```
rossmcewan, line 38: We get the size of the Lambda from the routing id - this decision is made based on the NangoProps when creating the Node.

packages/lambda/lib/index.ts
```diff
@@ -0,0 +1,76 @@
...
+        if (res.isOk()) {
+            lastSuccessHeartbeatAt = Date.now();
+        }
+    }, heartbeatIntervalMs);
+    try {
+        const payload = {
+            nangoProps: { ...(request.nangoProps as unknown as NangoProps), host: getNangoHost() },
+            code: request.code,
+            codeParams: request.codeParams,
+            locks: new MapLocks(), //new KVLocks(await getLocking()),
```
rossmcewan: This should change to a `KVLocks` instance but need to test the performance

packages/lambda/lib/index.ts
```diff
@@ -0,0 +1,77 @@
...
+        if (res.isOk()) {
+            lastSuccessHeartbeatAt = Date.now();
+        }
+    }, heartbeatIntervalMs);
+    try {
+        const payload = {
+            nangoProps: { ...(request.nangoProps as unknown as NangoProps), host: getNangoHost() },
+            code: request.code,
+            codeParams: request.codeParams,
+            locks: new KVLocks(await getLocking()),
```
rossmcewan: Uses a redis-backed locking implementation

packages/fleet/lib/db/migrations/20251203172800_add_lambda_fields.ts
```diff
@@ -0,0 +1,18 @@
...
+    await knex.raw(`ALTER TABLE ${NODES_TABLE}
+      ADD COLUMN IF NOT EXISTS fleet_id varchar(255) DEFAULT NULL`);
+    await knex.raw(`ALTER TABLE ${NODES_TABLE}
+      ADD COLUMN IF NOT EXISTS execution_timeout_secs int DEFAULT NULL`);
+    await knex.raw(`ALTER TABLE ${NODES_TABLE}
+      ADD COLUMN IF NOT EXISTS provisioned_concurrency int DEFAULT NULL`);
+    await knex.raw(`ALTER TABLE ${NODE_CONFIG_OVERRIDES_TABLE}
+      ADD COLUMN IF NOT EXISTS execution_timeout_secs int DEFAULT NULL`);
+    await knex.raw(`ALTER TABLE ${NODE_CONFIG_OVERRIDES_TABLE}
+      ADD COLUMN IF NOT EXISTS provisioned_concurrency int DEFAULT NULL`);
```
TBonnin, line 15: twice`provisioned_concurrency` and `execution_timeout_secs`

packages/fleet/lib/db/migrations/20251203172800_add_lambda_fields.ts
```diff
@@ -0,0 +1,18 @@
+import type { Knex } from 'knex';
+import { NODES_TABLE } from '../../models/nodes.js';
+import { NODE_CONFIG_OVERRIDES_TABLE } from '../../models/node_config_overrides.js';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.raw(`ALTER TABLE ${NODES_TABLE}
+      ADD COLUMN IF NOT EXISTS fleet_id varchar(255) DEFAULT NULL`);
```
TBonnin, line 7: I'd make it non null and set a value for the existing entries. It will then be one less thing to think about instead of wondering what a null value means.

packages/fleet/lib/db/migrations/20251203172800_add_lambda_fields.ts
```diff
@@ -0,0 +1,18 @@
...
+    await knex.raw(`ALTER TABLE ${NODES_TABLE}
+      ADD COLUMN IF NOT EXISTS execution_timeout_secs int DEFAULT NULL`);
```
TBonnin, line 9: would it make sense to have those new fields non null with default 0

packages/fleet/lib/models/node_config_overrides.ts
TBonnin: isn't `execution_timeout_secs` already handled by the orchestrator?

packages/fleet/lib/image-verifier/verifier.ts
TBonnin, line 1: we usually use underscore in file name. We could also not have this devate and move those files to `lib/verifier/image.ts`, `lib/verifier/docker.ts`, etc...

packages/jobs/lib/runner/lambda.ts
```diff
@@ -0,0 +1,189 @@
...
+}
+
+export function getFunctionQualifier(node: Node): string {
+    //need to get the qualifier from the function url
+    //arn is made up 8 segments: arn:aws:lambda:region:account:function:functionName:qualifier
+    const arnSegments = node.url?.split(':');
+    return arnSegments && arnSegments.length >= 8 ? arnSegments[7]! : '';
```
TBonnin, line 45: can this logic ever fail?

packages/jobs/lib/runner/lambda.ts
TBonnin: does that mean that there is no guarantee the lambda function will be running/healthy when invoked? 
I get why registration isn't necessary anymore in lambda world but ensuring `/health` answer at least once would be meaningful I think

packages/jobs/lib/runner/render.ts
TBonnin: 🪓

packages/jobs/lib/utils/runner.ts
```diff
@@ -0,0 +1,19 @@
+import { Ok, retryWithBackoff } from '@nangohq/utils';
+
+import type { Node } from '@nangohq/fleet';
+import type { Result } from '@nangohq/utils';
+
+export async function onFinishing(node: Node): Promise<Result<void>> {
```
TBonnin: can we name the function `notifyWhenIdle` after what it does instead of the most abstract `onFinishing`. Should it also live in the `lib/runner` folder and why not even in the `lib/runner/runner.ts` file?

packages/jobs/lib/runtime/lambda.adapter.ts
TBonnin: Let's make it a comment so I can remember in 6 months when I read the code ;)

packages/jobs/lib/runtime/runtimes.rules.ts
```diff
@@ -0,0 +1,33 @@
+import { Engine } from 'json-rules-engine';
+
+import { Err, Ok } from '@nangohq/utils';
+
+import { envs } from '../env.js';
+
+import type { NangoProps, Result } from '@nangohq/types';
+import type { RuleProperties } from 'json-rules-engine';
```
TBonnin: can we express the routing logic in pure typescript instead of adding a dependency and a custom DSL?

packages/jobs/lib/runtime/runtimes.ts
```diff
@@ -0,0 +1,87 @@
...
+export function getDefaultFleet(): Fleet {
+    return runnersFleet;
+}
+
+export async function getRuntimeAdapter(nangoProps: NangoProps): Promise<Result<RuntimeAdapter>> {
+    const result = await getFleetId(nangoProps);
+    if (result.isErr()) {
+        logger.error('Error while getting fleet id for runtime', result.error);
+    } else {
+        if (result.value) {
```
TBonnin, line 47: can result.value ever be null?

packages/kvstore/lib/Locking.ts
```diff
@@ -45,10 +45,20 @@ export class Locking {
         return { key };
     }
 
+    public async releaseAll(prefix: string): Promise<void> {
+        for await (const key of this.store.scan(`${prefix}:*`)) {
```
TBonnin, line 49: `scan` can be quite an expensive operation? Are we sure we want to allow it? Where is it being used?

packages/lambda/lib/index.ts
```diff
@@ -0,0 +1,123 @@
...
+
+function getNangoHost() {
+    return process.env['NANGO_HOST'] || 'http://server.internal.nango';
+}
+
+export const handler = async (event: zod.infer<typeof requestSchema>, context: Context) => {
+    context.callbackWaitsForEmptyEventLoop = false;
+    const request = requestSchema.parse(event);
+    const nangoProps = { ...(request.nangoProps as unknown as NangoProps) };
+    const locking = await getLocking();
```
TBonnin: this is a big architecture change, where runner now talk to a shared kvstore. Let's talk about that

packages/lambda-runner/lib/index.ts
```diff
@@ -0,0 +1,123 @@
...
+        logger.error('Conflicting sync detected', { syncId: nangoProps.syncId });
+        throw new Error('Conflicting sync detected');
+    }
+    const locks = new KVLocks(locking);
+
+    let lastSuccessHeartbeatAt: number | null = null;
+    const startTime = Date.now();
+
+    const abortController = new AbortController();
+    const heartbeatTimeoutMs = request.nangoProps.heartbeatTimeoutSecs ? request.nangoProps.heartbeatTimeoutSecs * 1000 : heartbeatIntervalMs * 3;
```
TBonnin, line 66: would be a good time to clean NangoProps code and ensure heartbeatTimeoutSecs is always set. It is probably optional because of legacy/migration reason

packages/lambda-runner/lib/index.ts
```diff
@@ -0,0 +1,123 @@
...
+                },
+                status: 500
+            },
+            telemetryBag: {
+                customLogs: 0,
+                proxyCalls: 0,
+                durationMs: Date.now() - startTime,
+                memoryGb: Number(context.memoryLimitInMB) / 1024
+            }
+        });
```
TBonnin, line 126: is it worth consolidating the hearthbeat and exec logic into `@nangohq/runner`?

packages/lambda-runner/lib/schemas.ts
```diff
@@ -0,0 +1,93 @@
...
+            request: zod.any().optional(), // AxiosInterceptorManager cannot be validated
+            response: zod
+                .object({
+                    onFulfilled: zod.function(),
+                    onRejected: zod.function()
+                })
+                .optional()
+        })
+        .optional()
+});
```
TBonnin, line 81: can we be less strict and ignore field like `axios`, etc...

packages/lambda-runner/Dockerfile.lambda
```diff
@@ -0,0 +1,99 @@
...
+  && rm -rf packages/*/src \
+  # && rm -rf packages/*/lib will break database migrations because they are not compiled; barely saves a few MBs
+  && rm -rf packages/*/public
+
+# Clean dev dependencies
+RUN true \
+  && npm prune --omit=dev --omit=peer --omit=optional
+
+RUN npm install -g npm@11.6.4
+
```
TBonnin, line 77: can we consolidate everything above? I am worried about both Dockerfile to accidentally fall out of sync

packages/lambda/package.json
```diff
@@ -0,0 +1,31 @@
...
+        "@nangohq/kvstore": "file:../kvstore",
+        "@nangohq/node": "file:../node-client",
+        "@nangohq/runner": "file:../runner",
+        "@nangohq/types": "file:../types",
+        "@nangohq/utils": "file:../utils",
+        "botbuilder": "4.23.3",
+        "datadog-lambda-js": "12.130.0",
+        "dd-trace": "5.80.0",
+        "soap": "1.6.0",
+        "unzipper": "0.12.3",
```
TBonnin: do we need soap and unzipper?

packages/utils/lib/environment/parse.ts
TBonnin: linked to my question above. Are rules passed via env var living in another repo necessary. Can they not be expressed in pure code?

.github/workflows/deploy.yaml
```diff
@@ -195,4 +196,48 @@ jobs:
                   curl -sS --fail-with-body --request POST "${{ vars.API_DOMAIN }}/internal/fleet/nango_runners_k8s/rollout" \
                     --header "authorization: Bearer $INTERNAL_API_KEY"\
                     --header "content-type: application/json"\
-                    --data "{ \"image\": \"nangohq/nango:${{ github.sha }}\" }"
+                    --data "{ \"image\": \"nangohq/nango:${{ github.sha }}\", \"imageType\":\"docker\" }"
+    deploy_lambda:
```
TBonnin, line 200: shouldn't runner and lambda be deployed at the same time?

packages/lambda-runner/lib/index.ts
TBonnin, line 1: lambda is quite a generic name. What if we run another component in lambdas in the future? Should we call this package `runner-lambda`?

packages/jobs/lib/runner/lambda.ts
```diff
@@ -0,0 +1,189 @@
...
+                const fleetId = node.fleetId || envs.RUNNER_LAMBDA_FLEET_ID;
+                const result = await registerWithFleet(fleetId, {
+                    nodeId: node.id,
+                    url: `${aResult.AliasArn}`
+                });
+                if (result.isErr()) {
+                    logger.error(`Error registering node ${node.id} to fleet ${fleetId}`, result.error);
+                }
+            } catch (err: any) {
+                logger.error(`Error creating function ${name}`, err);
```
TBonnin, line 141: should it return an error?

packages/jobs/lib/runner/lambda.ts
```diff
@@ -0,0 +1,189 @@
...
+        await lambdaClient.send(
+            new DeleteFunctionCommand({
+                FunctionName: name
+            })
+        );
+        return Ok(undefined);
+    }
+
+    async verifyUrl(_url: string): Promise<Result<void>> {
+        //this needs to be validates against a lambda function arn with qualifier using regex
```
TBonnin: not sure I understand the comment

packages/jobs/lib/utils/lambda.ts
```diff
@@ -0,0 +1,20 @@
...
+export function getSizeFromRoutingId(name: string): number {
+    const parts = name.split('-');
+    const size = Number.parseInt((parts.length && parts[parts.length - 1]) || '');
+    if (Number.isNaN(size)) return envs.LAMBDA_DEFAULT_SIZE;
+    return size;
+}
+
+export function getRoutingId(nangoProps: NangoProps): string {
+    const size = getSizeFromProps(nangoProps);
+    return `nango-function-${size}`;
```
TBonnin: `nango-runner-function`? in case one day we use lambdas for something else than the runner

packages/jobs/lib/runner/lambda.ts
```diff
@@ -0,0 +1,189 @@
...
+        const regex = /^arn:aws:lambda:.*:function:nango-function-\d+-[a-f0-9-]+:.*$/;
```
TBonnin: the name of the function should be tied up with the function named set in the utils file

packages/lambda-runner/lib/index.ts
```diff
@@ -0,0 +1,123 @@
...
+        try {
+            if (nangoProps.scriptType !== 'sync') return { allowed: true };
+            const key = this.getKey(nangoProps);
+            const lock = await this.locking.tryAcquire(key, opts.ttlMs, 1000);
+            return {
+                allowed: true,
+                lock
+            };
+        } catch {
+            return { allowed: false };
```
TBonnin, line 36: would it make sense to return the error?

packages/jobs/lib/runtime/lambda.adapter.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    async invoke(params: { taskId: string; nangoProps: NangoProps; code: string; codeParams: object }): Promise<Result<boolean>> {
+        const func = await this.getFunction(params.nangoProps);
+        const command = new InvokeCommand({
+            FunctionName: func.arn,
+            Payload: JSON.stringify({
+                taskId: params.taskId,
+                nangoProps: params.nangoProps,
+                code: Buffer.from(params.code).toString('base64'),
```
TBonnin: to we need to b64 encode?

packages/jobs/lib/runtime/runtimes.ts
```diff
@@ -0,0 +1,87 @@
...
+    adapter: new RunnerRuntimeAdapter(),
+    fleet: runnersFleet
+};
+runtimes.set(runnersFleet.fleetId, defaultRuntime);
+
+export function getDefaultFleet(): Fleet {
+    return runnersFleet;
+}
+
+export async function getRuntimeAdapter(nangoProps: NangoProps): Promise<Result<RuntimeAdapter>> {
```
TBonnin: is it possible to add a  test for this function?

TBonnin:
as mentioned in our discussion it would be helpful to add a `runtime` dimension to the function execution metric sent to datadog so we can break it down

packages/jobs/lib/runner/lambda.ts
rossmcewan: There is no "/health" for a Lambda function - the equivalent of the runner health would be getting the status of the function arn defined in the nodes.url and that would return 'Active'. The function doesn't serve requests in the same way the runner server does, it is invoked with the function payload. 

In this case, when the Lambda is registered as RUNNING with the fleet, it is semantically equivalent to a /health endpoint return healthy.

packages/utils/lib/environment/parse.ts
rossmcewan: Will refer to our discussion last night:
- I would prefer to keep it like this for now as we have a very loose understanding of the kinds of rules we would like to evaluate. 
- Doing it this way keeps the rule and value co-located - using code, we would still need a set of environment variables for the rules to evaluate against (e.g. environment ids). I would rather have a PR against a single repo (i.e. nango-environments) as opposed to one to change the code and another to change the environment variables.
- We can revisit this in the future - perhaps we get a better idea of the kinds of rules we are using OR (best case) we have a better feature management capability.

packages/fleet/lib/db/migrations/20251203172800_add_lambda_fields.ts
rossmcewan: What do you mean twice? One is in nodes and one is in the overrides.

packages/fleet/lib/models/node_config_overrides.ts
rossmcewan: It can't be. There is no way to interact with a running lambda function. This timeout is the maximum time that the Lambda runtime will allow the function to run (we use 900s by default). Currently, functions that timeout will stop sending a heartbeat and then the orchestrator can update the task appropriately.

packages/fleet/lib/db/migrations/20251203172800_add_lambda_fields.ts
rossmcewan: What value though? The default fleet id is passed in as an environment variable. We could use a magic string but I prefer NULL indicating the default (i.e. non-breaking) value.

packages/fleet/lib/image-verifier/verifier.ts
rossmcewan: Will change. Used hyphens more out of habit without checking what the default is.

packages/jobs/lib/runner/lambda.ts
rossmcewan: Not that I can think of. Did you have something specific in mind?

packages/jobs/lib/runtime/runtimes.rules.ts
rossmcewan: Will refer to our discussion last night:

- I would prefer to keep it like this for now as we have a very loose understanding of the kinds of rules we would like to evaluate.
- Doing it this way keeps the rule and value co-located - using code, we would still need a set of environment variables for the rules to evaluate against (e.g. environment ids). I would rather have a PR against a single repo (i.e. nango-environments) as opposed to one to change the code and another to change the environment variables.
- We can revisit this in the future - perhaps we get a better idea of the kinds of rules we are using OR (best case) we have a better feature management capability.

packages/jobs/lib/runtime/runtimes.ts
rossmcewan: It shouldn't happen, but yes - if the rules evaluate to no valid value. This is catch-all to return the default runtime if that happens.

packages/kvstore/lib/Locking.ts
rossmcewan: We have a function `releaseAllLocks` on the nango sdk.

packages/lambda/lib/index.ts
rossmcewan: Will implement changes as per our discussion.

packages/lambda-runner/lib/index.ts
rossmcewan: Will do as follow up - trying to reduce the amount of changes that are not directly related.

packages/lambda-runner/lib/index.ts
rossmcewan: Potentially. Will have a look but will probably schedule as a follow up task.

packages/lambda-runner/lib/schemas.ts
rossmcewan: Absolutely. This was a generated schema - will review the structure.

packages/jobs/lib/runner/render.ts
rossmcewan: Will do in separate PR.

packages/jobs/lib/runtime/runtimes.rules.ts
TBonnin: because we have a very loose understanding of the kinds of rules I would take the opposite approach and define everything as code, until we figure out what we need. I don't really like that some product logic are gonna be living in a separate repo.
I am not proposing to use env vars AND code but code only, with all the logic hardcoded in a single place that requires a single deployment.
```
// pseudo-code
function getRuntime(props: nangoProps): 'lambda' | 'k8s' {
    const  environmentShouldUseLambda = () => {
        if(isProd) return [1, 2, 3].includes(nangoProps.environmentId)
        return  true     
   }
   const actionShouldUseLambda = () => {
       return nangoProps.type == 'action'
   }
   
   if(environmentShoudUseLambda && actionShouldUseLambda) {
       return 'lambda'
   }
}




```

packages/lambda-runner/lib/index.ts
rossmcewan: For what purpose - logging? Failure to acquire the lock always throws an error.

packages/jobs/lib/runtime/lambda.adapter.ts
rossmcewan: No. This came from testing - I was having issues getting escaped payloads to work. Will change it back to raw - more efficient anyway.

packages/jobs/lib/runner/lambda.ts
rossmcewan: Just describing that the arn is a lambda function WITH qualifier - you can still have a valid lambda ARN without qualifier. Can remove though.

packages/jobs/lib/runner/lambda.ts
rossmcewan: It can't return it because it's running in the background - i.e. caller has already returned. The supervisor would fail it as it takes too long to start.

.github/workflows/deploy.yaml
rossmcewan: I don't think so. There are certainly changes we would want to deploy to Lambda that shouldn't necessitate a full rollout of runners (and vice versa).

packages/lambda/package.json
rossmcewan: Can remove it from the lambda package.json now - layover from when I wasn't using the exec function directly. Will remove.

packages/lambda-runner/Dockerfile.lambda
rossmcewan: What do you mean consolidate?

rossmcewan:
> as mentioned in our discussion it would be helpful to add a `runtime` dimension to the function execution metric sent to datadog so we can break it down

I will capture a separate task for this - it's quite a large change as it needs to be tied in to the task success/task failure events.

packages/fleet/lib/db/migrations/20251203172800_add_lambda_fields.ts
rossmcewan: I don't think so. I prefer NULL as a default here as 0 could carry meaning.

packages/jobs/lib/runtime/runtimes.rules.ts
rossmcewan: I don't think exposing actual data of the nango props is a good idea in our codebase.

packages/lambda-runner/lib/index.ts
TBonnin: mostly debugging but no big deal

.github/workflows/deploy.yaml
TBonnin: ok. We will have to remember that any changes to the runner-sdk would require 2 deploys

packages/lambda-runner/Dockerfile.lambda
TBonnin: I mean not having 2 Dockerfile with roughly the same content that someone (probably me) is going to forget to update when a change is needed

packages/fleet/lib/db/migrations/20251203172800_add_lambda_fields.ts
TBonnin: nvm. I don't know how to read 🙈

packages/jobs/lib/runtime/runtimes.rules.ts
TBonnin: fair. I still think introducing an extra dependency, a DSL and splitting product logic as json into a separate codebase is too complicated.

.github/workflows/deploy.yaml
rossmcewan: We could probably create a step flow that does both to help with that but I don't think it should be the default, especially early on. Maybe we change that in the future.

packages/kvstore/lib/index.ts
```diff
@@ -14,13 +15,14 @@ export { RedisKVStore } from './RedisStore.js';
 export type { KVStore } from './KVStore.js';
 export { type Lock, Locking } from './Locking.js';
 
+type KvBoundary = 'system' | 'customer';
```
rossmcewan, line 18: Introduced a boundary to define the different kinds of kvstores we might want - currently have system and customer. Open to discussion on the best way to implement this but was more complicated than I expected.

packages/lambda-runner/Dockerfile.lambda
rossmcewan: They aren't identical though. Different base images. Different package. I originally started by trying to do a new image completely but the structure of the repo and the dependencies made that quite challenging so decided to use the main one as a base. I would like to trim the Lambda one as much as possible though - it just requires some iteration. I don't think they should be shared.

packages/jobs/lib/runtime/runtimes.rules.ts
rossmcewan: Will try and lay out what rules we want to start with and can look at a code implementation.

packages/jobs/lib/runner/lambda.ts
rossmcewan: Captured issue in Linear

TBonnin (approved):

PR #5009, opened 2025-11-13
fix(proxy): resolve issue where authorization headers are getting dropped during redirects

hassan254-prog:

- axios uses the `follow-redirects` library to handle redirects, which, by design, [removes](https://github.com/follow-redirects/follow-redirects/blob/86dc1f86e4b56bcd642c78384d51f10f123aea75/index.js#L471) sensitive headers such as `authorization` when redirecting to a different origin. In the case of Dayforce, we need to ensure that the redirect is followed while preserving the authorization header.

---

**Proxy Redirect Handling: Preserve Authorization Header Across Safe 3xx Redirects**

This PR fixes a long-standing proxy bug where Authorization headers were silently dropped when the proxy auto-followed HTTP redirects (307, 308, etc.). The updated logic now explicitly carries the Authorization header to the redirected request whenever the destination host is the same (or otherwise whitelisted), preventing unexpected 401/403 responses from downstream services. Additional regression tests and minor provider tweaks were added to ensure the new behavior is continuously verified.

<details>

• Refactored redirect logic in packages/shared/lib/services/proxy/utils.ts to copy Authorization header on safe redirects and centralize a keep/strip header list.
• Added end-to-end unit tests in packages/runner/lib/sdk/sdk.unit.test.ts that spin up a redirecting stub and assert header preservation.
• Updated providers.yaml entries for services that rely on redirect-based authentication to exercise the new path in CI.

</details>

<details>

• proxy redirect handler (utils.ts)
• SDK tests
• provider configuration

</details>

---

### DIFF
--- a/packages/providers/providers.yaml
+++ b/packages/providers/providers.yaml
@@ -3542,6 +3542,10 @@ dayforce:
         client_id: Dayforce.HCMAnywhere.Client
     token_headers:
         content-type: application/x-www-form-urlencoded
+    token_response:
+        token: access_token
+        token_expiration: expires_in
+        token_expiration_strategy: expireIn
     docs: https://nango.dev/docs/integrations/all/dayforce
     docs_connect: https://nango.dev/docs/integrations/all/dayforce/connect
     connection_config:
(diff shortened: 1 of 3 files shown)


TBonnin (approved):
are propel comments legit?

hassan254-prog:
> are propel comments legit?

I don't think so, in dayforce, we know that its going to be a cross origin redirect and for the authentication to work we need to also pass this header along. we also need to pass this header as is from nango `buildProxyHeaders`.

PR #5026, opened 2025-11-18
fix: avoid excessive locking when updating connection last_fetched_at

TBonnin:
The TODO comment state that calling `updateLastFetched` could be removed. However it is still used in `getStaleConnections` https://github.com/NangoHQ/nango/blob/cc93b91b3bc4445c70c8cf84783cb0e0c26d34fb/packages/shared/lib/services/connection.service.ts#L572
so not sure removing is safe. 
pinging @bodinsamuel in case you remember (feel free to ignore if you are too busy enjoying köttbullar) 

In the meantime, this commit is making the query more concurrency-safe by skipping the update if another transaction is locking the row.

---

**Make `updateLastFetched()` non-blocking under concurrent access**

Rewrites the `updateLastFetched()` query so that it *attempts* to acquire a row-level lock and immediately skips the update if the row is already locked.  This prevents connection-refresh workers that run in parallel from piling up on the same row and keeps overall DB latency low.  Call-sites remain unchanged except that an outdated TODO comment was removed.

<details>

• Refactored `updateLastFetched(id)` in `packages/shared/lib/services/connection.service.ts` to: (1) sub-select the row with `forUpdate().skipLocked()`, (2) update with `db.knex.fn.now()` instead of `new Date()`
• Deleted stale TODO line in `packages/shared/lib/services/connections/credentials/refresh.ts` (functional behaviour unchanged)

</details>

<details>

• `packages/shared/lib/services/connections/credentials/refresh.ts`

</details>

---

### DIFF
--- a/packages/shared/lib/services/connection.service.ts
+++ b/packages/shared/lib/services/connection.service.ts
@@ -842,8 +842,17 @@ class ConnectionService {
         return del;
     }
 
-    public async updateLastFetched(id: number) {
-        await db.knex.from<DBConnection>(`_nango_connections`).where({ id, deleted: false }).update({ last_fetched_at: new Date() });
+    public async updateLastFetched(id: number): Promise<void> {
+        // the query can be called concurrently
+        // we therefore first attempt to lock the row
+        // if lock can't be acquired, the update is skipped without waiting
+        // in order to avoid excessive db locking/waiting
+        await db
+            .knex('_nango_connections')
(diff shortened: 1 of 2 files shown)


rossmcewan (approved):

bodinsamuel:
Sorry I don't remember, with the new colums (last_refreshed_at, etc.) I think it might be fine.

PR #5041, opened 2025-11-19
feat: add AWS SigV4 proxy integration

nickvelloff:
## Summary
  - introduce the AWS SigV4 provider (docs, spec entry, logo, generated templates, and connect UI copy) so customers can create proxy integrations against AWS services
  - add the SigV4 auth flow on the backend (new /auth/aws-sigv4 controller, session/reconnect handling, validation, and reconnect defaults) plus expose the new auth type through the shared types & SDKs
  - add the signing/proxy helpers (shared proxy service + unit tests) and hook them into the connection service so signed calls pick up the correct base URL and ARN context

  ## Testing
  - npm run test packages/shared/lib/services/proxy/aws-sigv4.unit.test.ts
  - npm run test packages/server/test/integration/connect/postReconnect.integration.test.ts
  - npm run test packages/server/test/integration/providerConfigKey/patchIntegration.integration.test.ts

---

In addition to introducing end-to-end `AWS_SIGV4` support, this PR also strengthens configuration security and lifecycle handling by separating sensitive SigV4 credentials from public integration config, applying encrypted secret storage, and ensuring redaction/normalization in integration responses. It further expands the experience with dedicated AWS SigV4 settings behavior in the web app (including built-in vs custom STS configuration) while preserving reconnect/default flows and legacy compatibility expectations.

---


TBonnin:
A few questions:
- who is running/providing the STS helper service?
- can you share the cloudformation template?

khaliqgant:
> * who is running/providing the STS helper service?

Our users would have to run the STS helper service

TBonnin:
> > * who is running/providing the STS helper service?
> 
> Our users would have to run the STS helper service

With some cross-account sts assume role policy I imagine. I guess that what the cloudformation template is setting up. I would like to see the intended template.

nickvelloff:
The intent is that the Nango customer runs their own STS helper container in their AWS account; it assumes roles on behalf of their end customers. That keeps Nango out of operating that lifecycle and cleanly separates responsibilities.

Nango could optionally “provide” a boilerplate service, but only as a contract/example (see https://github.com/nickvelloff/nango-sts-boilerplate). It’s deliberately simple.

CloudFormation templates are owned/hosted by the customer because they reflect their specific role needs. The only requirement is that the template URL is publicly readable (CORS-friendly) so the UI can hydrate parameters. Nango could host a few generic templates as defaults, but that’s just a nice-to-have. We assume anyone needing this enterprise-style integration already has the CloudFormation/DevOps maturity to host their own templates.

Cloudformation template example (https://raw.githubusercontent.com/nickvelloff/nango-sts-boilerplate/refs/heads/main/infra/cloudformation/s3-readonly.json) - simply hosted publicly.

Importantly: This file explains the usage and validation flows a customer would want to leverage when setting this up - https://github.com/nickvelloff/nango-sts-boilerplate/blob/main/docs/aws-sigv4-e2e-validation.md

packages/server/lib/controllers/v1/integrations/providerConfigKey/patchIntegration.ts
```diff
@@ -139,6 +140,33 @@ export const patchIntegration = asyncWrapper<PatchIntegration>(async (req, res)
...
 
+    if ('custom' in body && body.custom) {
+        let nextCustom: Record<string, string> = { ...(integration.custom || {}) };
+        for (const [key, value] of Object.entries(body.custom)) {
+            if (value === null || value === '') {
+                const { [key]: _removed, ...rest } = nextCustom;
+                nextCustom = rest;
+                continue;
+            }
+            if (key === 'aws_sigv4_config') {
```
agusayerza: nit: We should use a constant

packages/shared/lib/auth/aws-sigv4.ts
```diff
@@ -0,0 +1,147 @@
...
+
+export interface AwsSigV4TemporaryCredentials {
+    accessKeyId: string;
+    secretAccessKey: string;
+    sessionToken: string;
+    expiresAt: Date;
+}
+
+export function getAwsSigV4Settings(config: ProviderConfig): Result<AwsSigV4IntegrationSettings, NangoError> {
+    const rawSettings = config.custom?.[AWS_SIGV4_CUSTOM_KEY];
```
agusayerza: STS endpoint auth secrets (API key, basic auth password) are stored inside an unstructured JSON blob in the custom field rather than in dedicated secret fields, right? Those secrets shouldn't be buried in a nested JSON structure inside `custom` as they can't be individually rotated, audited, or access-controlled separately from the rest of the integration config. Plus, it would be super simple to accidentally expose them.

packages/server/lib/controllers/auth/postAwsSigV4.ts
```diff
@@ -0,0 +1,365 @@
...
+
+        const resolvedRegion = body.region || (connectionConfig['region'] as string) || settings.defaultRegion;
+        if (!resolvedRegion) {
+            const err = new NangoError('missing_aws_sigv4_region');
+            errorManager.errResFromNangoErr(res, err);
+            await logCtx.failed();
+            return;
+        }
+
+        const externalId = (connectionConfig['external_id'] as string) || uuidv4();
```
agusayerza: We shouldn't accept the `external_id` provided by the client without validation. ([The External ID is a critical AWS confused deputy protection](https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html)). If an attacker passes a known `external_id` that matches one already trusted in a victim's role trust policy, they could assume that role. I think for new connections Nango should always generate the `external_id` server-side. For reconnections, it should only reuse the `external_id` already stored in the existing connection's `connection_config`, never accept a new one from the client.

packages/shared/lib/services/proxy/aws-sigv4.ts
```diff
@@ -0,0 +1,120 @@
...
+    entries.sort(([aKey, aValue], [bKey, bValue]) => {
+        if (aKey === bKey) {
+            return aValue < bValue ? -1 : aValue > bValue ? 1 : 0;
+        }
+        return aKey < bKey ? -1 : 1;
+    });
+    return entries.map(([key, value]) => `${key}=${value}`).join('&');
+}
+
+function hashPayload(body?: string | Buffer | null): string {
```
agusayerza: Function returns `UNSIGNED-PAYLOAD` when body is null but hashes an empty string when body is undefined. This inconsistency will cause signature mismatches for non-S3 services. The hashing decision should be driven by `credentials.service === 's3'`, not by the body type.

packages/shared/lib/services/proxy/utils.ts
```diff
@@ -208,7 +212,17 @@ export function buildProxyURL({ config, connection }: { config: ApplicationConst
...
     let apiBase = config.baseUrlOverride || templateApiBase;
 
-    if (apiBase?.includes('${') && apiBase?.includes('||')) {
+    if (!apiBase && connection.credentials.type === 'AWS_SIGV4') {
+        const awsCreds = connection.credentials as AwsSigV4Credentials;
+        const awsRegion = awsCreds.region || (connection.connection_config?.['region'] as string | undefined);
+        const awsService = awsCreds.service || (connection.connection_config?.['service'] as string | undefined);
+
+        if (awsRegion && awsService) {
+            apiBase = `https://${awsService}.${awsRegion}.amazonaws.com`;
```
agusayerza: There’s no integration-level base URL for AWS SigV4. The provider has no `proxy.base_url`, so `buildProxyURL` hardcodes `https://${service}.${region}.amazonaws.com`. That breaks for services with non-standard endpoint patterns (e.g., S3 uses bucket/region or global endpoints and has path-style vs virtual-hosted style) aren’t supported unless callers pass `baseUrlOverride` on *every proxy call*. We should add an integration-level override (or partition-aware resolution) so this is reusable/configurable across AWS environments.

packages/providers/providers.yaml
```diff
@@ -1560,6 +1560,28 @@ aws-scim:
             order: 1
             doc_section: '#step-2-setting-up-automatic-provisioning'
 
+aws-sigv4:
```
agusayerza, line 2358: The provider has no proxy.base_url defined. Consider defining `proxy.base_url: https://${connectionConfig.service}.${connectionConfig.region}.amazonaws.com`

packages/connect-ui/src/views/Go.tsx
```diff
@@ -66,6 +96,10 @@ const formSchema: Record<AuthModeType, z.ZodObject> = {
         username: z.string().min(1),
         password: z.string().min(1)
     }),
+    AWS_SIGV4: z.object({
+        role_arn: z.string().min(1),
+        region: z.string().min(1)
```
agusayerza: Here `AWS_SIGV4` uses .min(1) making it required, but `providers.yaml` and the backend both say it's optional.

packages/webapp/src/pages/Integrations/providerConfigKey/Settings/components/General.tsx
```diff
@@ -310,6 +678,271 @@ export const SettingsGeneral: React.FC<{
                     )}
                 </div>
             )}
+
+            {template.auth_mode === 'AWS_SIGV4' && (
```
agusayerza: Let's extract this to a component so we don't add ~300 lines of AWS-specific UI directly in `General.tsx`, which is the shared settings component for all auth types

packages/webapp/src/pages/Integrations/providerConfigKey/Settings/components/General.tsx
```diff
@@ -310,6 +678,271 @@ export const SettingsGeneral: React.FC<{
...
+                <div className="grid grid-cols-1 gap-10">
+                    <InfoBloc title="AWS SigV4 Settings" help={<p>Configure how this integration issues temporary AWS credentials via your STS endpoint.</p>}>
+                        <div className="flex flex-col gap-4">
+                            <div className="grid grid-cols-2 gap-4">
+                                <div className="flex flex-col gap-1">
+                                    <label className="text-xs text-white font-semibold">AWS Service</label>
```
agusayerza: Hardcoded color values violate project conventions. Instead of `text-white` use `text-text-primary`

packages/webapp/src/pages/Integrations/providerConfigKey/Settings/components/General.tsx
```diff
@@ -310,6 +678,271 @@ export const SettingsGeneral: React.FC<{
...
+                                    <label className="text-xs text-white font-semibold">CloudFormation Templates (optional)</label>
+                                    <p className="text-xs text-text-soft">
+                                        Store one or more named templates for this integration. They will appear in Connect as one-click deployment buttons for
+                                        your customers.
+                                    </p>
+                                </div>
+                                <div className="flex flex-col gap-4">
+                                    {awsSigV4Config?.templates && awsSigV4Config.templates.length > 0 ? (
+                                        awsSigV4Config.templates.map((template, index) => (
+                                            <div key={`template-${index}`} className="border border-subtle rounded-md p-4 flex flex-col gap-3 bg-[#0f1117]">
```
agusayerza: Hardcoded color `bg-[#0f1117]`

packages/webapp/src/pages/Integrations/providerConfigKey/Settings/components/General.tsx
```diff
@@ -310,6 +678,271 @@ export const SettingsGeneral: React.FC<{
...
+                                    }
+                                    placeholder="https://sts.example.com/assume"
+                                    variant={'flat'}
+                                />
+                            </div>
+
+                            <div className="grid grid-cols-3 gap-3">
+                                <div className="flex flex-col gap-1">
+                                    <label className="text-xs text-white font-semibold">Auth Type</label>
+                                    <select
```
agusayerza: Use Select component

packages/server/lib/controllers/auth/postAwsSigV4.ts
```diff
@@ -0,0 +1,365 @@
...
+        if (logCtx) {
+            void logCtx.error('uncaught error', { error: err });
+            await logCtx.failed();
+        }
+        metrics.increment(metrics.Types.AUTH_FAILURE, 1, { auth_mode: 'AWS_SIGV4', ...(config ? { provider: config.provider } : {}) });
+        next(err);
+    }
+});
+
+async function verifyAwsCredentials({
```
agusayerza, line 354: Q: Should we check the expected role after `GetCallerIdentity` ?

package.json
```diff
@@ -92,5 +92,6 @@
     },
     "engines": {
         "node": ">=20.0.0"
-    }
+    },
+    "packageManager": "pnpm@10.15.1+sha512.34e538c329b5553014ca8e8f4535997f96180a1d0f614339357449935350d924e22f8614682191264ec33d1462ac21561aff97f6bb18065351c162c7e8f6de67"
```
agusayerza: This shouldn't be included as part of the PR

packages/types/lib/api.ts
```diff
@@ -36,7 +36,11 @@ export type ResDefaultErrors =
     | ApiError<'invalid_cli_version'>
     | ApiError<'invalid_permissions'>
     | ApiError<'invalid_connect_session_token_format'>
-    | ApiError<'request_too_large'>;
+    | ApiError<'request_too_large'>
+    | ApiError<'missing_aws_sigv4_config'>
+    | ApiError<'invalid_aws_sigv4_config'>
+    | ApiError<'missing_aws_sigv4_service'>
+    | ApiError<'missing_aws_sigv4_sts_endpoint'>;
```
agusayerza: These should be scoped to `PostPublicAwsSigV4Authorization['Error']` in `packages/types/lib/auth/http.api.ts` only.

docs/spec.yaml
```diff
@@ -91,6 +91,7 @@ paths:
                     $ref: '#/components/responses/NotFound'
 
         post:
+            operationId: postAuthAwsSigV4
```
agusayerza: I agree with the bot here, we should remove this line. This has only been added to `/auth/aws-sigv4/{providerConfigKey}` right?

agusayerza (changes requested):
The STS endpoint design makes this non-reusable. We should support direct AWS STS `AssumeRole` calls as a built-in option. Nango should be able to call AWS STS `AssumeRole` directly using the integration owner's own AWS credentials stored in the encrypted credential fields. This would remove the need to deploy a bespoke microservice for most of use cases and make the provider reusable. The custom STS endpoint can remain as an advanced/power-user option for users who need custom validation or multi-tenant routing logic between Nango and AWS.)

packages/shared/lib/auth/aws-sigv4.ts
agusayerza: Expanding on this as requested on slack:

- Keep `custom.aws_sigv4_config` for non-secret settings only (`service`, `defaultRegion`, `stsEndpoint.url`, templates/instructions).
- Move STS auth secret material into dedicated encrypted integration fields (same pattern as `oauth_client_secret`):
  - `aws_sigv4_sts_auth_type`
  - `aws_sigv4_sts_auth_header`
  - `aws_sigv4_sts_auth_username`
  - `aws_sigv4_sts_auth_secret`

This gives us rotation/audit/access-control per secret and avoids exposing credentials through `custom` JSON.

What I'm proposing is something like this: 
```javascript
// packages/types/lib/integration/db.ts
type EncryptedSecret = {
    encryptedValue: string;
    iv: string;
    authTag: string;
};

type AwsSigV4StsAuthSecret =
    | { type: 'none' }
    | { type: 'api_key'; header: string; secret?: EncryptedSecret }
    | { type: 'basic'; username: string; secret?: EncryptedSecret };

export interface IntegrationConfig {
    // existing fields...
    integration_secrets?: {
        aws_sigv4?: {
            sts_auth?: AwsSigV4StsAuthSecret;
        };
    } | null;
}
```
Where the `integration_secrets` are part of the config, but on its own field so we can filter them out when returning on public endpoints, we can audit access, they are properly encrypted etc.

arsalazarjr: Thanks for that. I just pushed some changes for this. Here's what was done:

Database: Added integration_secrets JSONB column to _nango_configs with a migration.

Encryption: The column is encrypted/decrypted by the existing encryption manager using the same {encryptedValue, iv, authTag} pattern as custom.

Write path: When the PATCH endpoint receives aws_sigv4_config, it extracts the STS auth (API key value or basic auth password) into integration_secrets.aws_sigv4.sts_auth and stores the cleaned config (without auth) in custom. Re-saving without providing a new secret preserves the existing one. Removing the config clears both columns.

Read path: getAwsSigV4Settings reads auth from integration_secrets first, falls back to custom for pre-migration data. No bulk migration needed. Secrets move on next save.

API response: A redactAwsSigV4Secrets formatter replaces secret values with "***" in responses, handling both migrated and legacy data. integration_secrets is excluded from the API type entirely.

Frontend: Treats "***" as "secret configured" (shows empty field), omits empty secrets on save so the backend preserves existing values.

packages/server/lib/controllers/auth/postAwsSigV4.ts
arsalazarjr: Yes, we already do. Lines 387-405 parse the GetCallerIdentity response, extract the Arn, and compare the role name portion against the expected role_arn. It's a warning-only log today (not a hard failure) since the assumed-role ARN format differs from the IAM role ARN (sts::123:assumed-role/MyRole/session vs iam::123:role/MyRole).

packages/shared/lib/auth/aws-sigv4.ts
arsalazarjr: @agusayerza 
Built-in AWS STS AssumeRole support is complete. Integrations can now choose between two STS modes:
- Built-in AWS STS: Nango calls sts:AssumeRole directly with the integration owner's AWS credentials. No external service needed.
- Custom STS Endpoint: Existing behavior, unchanged.

What changed
- Backend: Added stsMode (builtin | custom) to settings. Builtin mode signs and sends AssumeRole requests directly, with credentials stored in the encrypted integration_secrets column. Response parser handles both JSON and XML from AWS STS.
- PATCH endpoint: Extracts and encrypts builtin credentials on save, cleans up stale secrets on mode switch, validates credentials exist.
- Frontend: STS Mode selector that conditionally shows either AWS credential fields (builtin) or STS endpoint/auth fields (custom).
- Tests: 11 unit tests + 2 integration tests, all passing.
- Docs: Updated with separate setup guides for each mode.

Tested end-to-end: created a connection via UI in builtin mode, then successfully listed S3 bucket contents through the Nango proxy.

packages/connect-ui/src/views/Go.tsx
```diff
@@ -103,7 +139,16 @@ export const Go: React.FC = () => {
...
-    const preconfigured = session && integration ? session.integrations_config_defaults?.[integration.unique_key]?.connection_config || {} : {};
+    const preconfiguredCredentials = session && integration ? session.integrations_config_defaults?.[integration.unique_key]?.credentials || {} : {};
+    const preconfiguredParams = session && integration ? session.integrations_config_defaults?.[integration.unique_key]?.connection_config || {} : {};
+    const preconfigured = { ...preconfiguredCredentials, ...preconfiguredParams };
+    const initialExternalId = useMemo(() => {
+        const value = (preconfiguredParams['external_id'] as string | undefined) || (preconfiguredCredentials['external_id'] as string | undefined);
+        return value && value.length > 0 ? value : generateExternalId();
+    }, [preconfiguredParams, preconfiguredCredentials]);
+    const [awsExternalId] = useState(initialExternalId);
+    const [externalIdCopied, setExternalIdCopied] = useState(false);
```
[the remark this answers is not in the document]
arsalazarjr: The re-render concern is mitigated by the existing code: useState(initialExternalId) on line 149 locks the value on first render, even if the useMemo recomputes due to dependency changes, awsExternalId never changes for the lifetime of the component.

For page refreshes, the connect session (including integrations_config_defaults) is fetched from the backend, so if the caller pre-populated external_id in the session, it survives. And the backend now honors client-provided external_id from params, so the UI-generated value is what gets stored and used for AssumeRole.

The suggestion to generate it in postConnectSessions is worth considering in a future iteration, but it would make the generic connect session endpoint AWS SigV4-aware, which breaks separation of concerns. I'd rather keep that endpoint auth-mode-agnostic.

agusayerza:
I did a test run of the integration and its working as expected. 

Couple of pending items and questions:
- [ ] Connect UI: the “copy external ID” button doesn’t copy on my end (seems broken)
- [ ] Can you rebase onto master?
- [ ] Custom STS mode: can you share a concrete example custom STS endpoint config (URL + auth/headers) so we can test that flow too? We need to be able to manually test this on our side
- [ ] Config scoping question: do you intend `service` + `aws_access_key_id`/`aws_secret_access_key` to be integration-level config? Shouldn't this be a connection-level config? Might be confused here, but I would expect those keys to be the customer's credentials to perform the `assumeRole`, no?
- [ ] CloudFormation template: Feels like product/business onboarding logic creeping into Nango. I think this should be handled on your side once the connection is established. I’m open to the rationale, but my default would be keeping this out unless we have a strong platform reason. Also, is the intent having one integration-level template reused for all connections?
- [ ] UI styling doesn’t fully match our current design system (probably minor and we can fix as follow up)

arsalazarjr:
@agusayerza I've made the changes requested in your last reply (except UI since you said it can be deferred in follow-up PR). Here are the answers to the questions you had there.
> * [ ]  Config scoping question: do you intend `service` + `aws_access_key_id`/`aws_secret_access_key` to be integration-level config? Shouldn't this be a connection-level config? Might be confused here, but I would expect those keys to be the customer's credentials to perform the `assumeRole`, no?

The scoping is intentional. There are two layers:

- Integration-level: service, aws_access_key_id, aws_secret_access_key -- these belong to the Nango customer (the integration owner), not the end customer. The access key/secret key are the integration owner's long-lived credentials used to call sts:AssumeRole on behalf of each end customer. They're stored encrypted in integration_secrets, never exposed to end customers.
- Connection-level: role_arn, external_id, region, and the resulting temporary credentials (access_key_id, secret_access_key, session_token, expires_at). These are per end customer and stored in the connection's encrypted credentials column.

The flow is: integration owner provides their AWS credentials once at setup, then each end customer provides their IAM Role ARN during Connect, then Nango uses the owner's credentials to AssumeRole into the customer's account, and stores the temp creds on the connection and refreshes them automatically.

So the AWS keys aren't the customer's credentials. They're the integration owner's credentials that enable assuming roles across all their customers' accounts.

> * [ ]  CloudFormation template: Feels like product/business onboarding logic creeping into Nango. I think this should be handled on your side once the connection is established. I’m open to the rationale, but my default would be keeping this out unless we have a strong platform reason. Also, is the intent having one integration-level template reused for all connections?

Yes, the templates are integration-level and reused across all connections. The integration owner defines one (or a few) templates, and every end customer sees the same one-click deploy buttons in Connect UI.

Fair pushback. The rationale is that the IAM trust-policy setup is a prerequisite to establishing the connection. Without it, the AssumeRole call fails and the connection can't be created. It's analogous to how OAuth providers need redirect URIs configured before auth works, except here the "configuration" happens in the customer's AWS account. Putting it in Connect means the end customer can deploy the stack and paste the Role ARN in a single flow rather than going back and forth.

That said, if the preference is to keep this out of Nango and have it handled externally (e.g., via the instructions_url or custom pre-connect documentation), that's doable -- the CloudFormation template feature could be stripped without breaking the core SigV4 auth flow. The minimum viable integration only needs: role ARN + external ID collection in Connect, and the STS assume-role call.

packages/server/lib/controllers/v1/integrations/providerConfigKey/patchIntegration.ts
```diff
@@ -74,6 +77,121 @@ export const patchIntegration = asyncWrapper<PatchIntegration>(async (req, res)
...
+                const validation = awsSigV4Client.getAwsSigV4Settings(simulated);
+                if (validation.isErr()) {
+                    res.status(400).send({ error: { code: validation.error.type, message: validation.error.message } } as PatchIntegration['Errors']);
+                    return;
+                }
+
+                // Update integration_secrets based on mode
+                let parsedConfig: Record<string, any> = {};
+                try {
+                    parsedConfig = JSON.parse(value);
```
agusayerza: Why are we parsing the value again?
We have parsed
1. To validate its JSON
2. To extract the secrets
3. To get the `stsMode`

agusayerza:
Hey @arsalazarjr , thanks for addressing the feedback, the config scoping explanation makes sense (integration-level owner credentials vs. connection-level customer credentials), the rebase looks clean, and the custom STS reference in #integration-templates/443 answers the testing question well.

I will be faster to review the follow up, please let me know when done.

## CloudFormation templates:
This is out of scope for Nango's current capabilities and should be handled on the integrator's side. Let strip it. The IAM setup is a prerequisite the integrator controls, not something Nango should guide through its UI. Keeping it in means we're taking on product/onboarding logic that belongs to the customer's app. Please remove the CloudFormation template feature from the integration (the template catalog, the "Deploy in AWS Console" buttons, the `instructions` block in Connect UI). The core flow: collect Role ARN + ExternalId, call STS, store credentials, is self-contained without it.

---

Other than that, there are small nits raised by Codex and 1 by me and we are done.

arsalazarjr:
> Hey @arsalazarjr , thanks for addressing the feedback, the config scoping explanation makes sense (integration-level owner credentials vs. connection-level customer credentials), the rebase looks clean, and the custom STS reference in #integration-templates/443 answers the testing question well.
> 
> I will be faster to review the follow up, please let me know when done.
> 
> ## CloudFormation templates:
> This is out of scope for Nango's current capabilities and should be handled on the integrator's side. Let strip it. The IAM setup is a prerequisite the integrator controls, not something Nango should guide through its UI. Keeping it in means we're taking on product/onboarding logic that belongs to the customer's app. Please remove the CloudFormation template feature from the integration (the template catalog, the "Deploy in AWS Console" buttons, the `instructions` block in Connect UI). The core flow: collect Role ARN + ExternalId, call STS, store credentials, is self-contained without it.
> 
> Other than that, there are small nits raised by Codex and 1 by me and we are done.

@agusayerza Thanks for flagging this. I want to push back because removing the CloudFormation piece now would undo work we explicitly aligned on at the start of this effort and reshape the process we've already built around.

On scope: The CloudFormation templates aren't "onboarding logic for the customer's app." They're a self-service path for the integrator's end users to provision the exact IAM role (trust policy + ExternalId) that STS AssumeRole requires. Without it, every integrator using AWS Sig4 has to write the same IAM setup documentation and guide their users through console clicks. Bundling a known-good template is the lowest-friction way to avoid misconfigured trust policies.

On "product logic": The templates are static assets + a pre-filled console URL. There's no business logic, no runtime behavior, no state to maintain. It's documentation-as-a-link, not a product feature that accrues complexity.

What I'd propose instead: If the concern is that this feels like it belongs outside the core SDK, we could keep the templates but gate them behind a provider-level flag so integrators can opt out while still allowing us to use it.

arsalazarjr:
> Other than that, there are small nits raised by Codex and 1 by me and we are done.

@agusayerza 
I addressed the Codex feedback and your comment [here](https://github.com/NangoHQ/nango/pull/5041/#discussion_r3088744050) in my latest commit. Here's a summary of those changes. Please lmk if it looks good.

#### Security / correctness fixes

SigV4 canonical header whitespace — packages/shared/lib/services/proxy/aws-sigv4.ts
Canonical header values now collapse internal whitespace to a single space per the SigV4 spec, not just trim ends. Prevents SignatureDoesNotMatch for headers with repeated whitespace.

S3 canonical path double-encoding — packages/shared/lib/services/proxy/aws-sigv4.ts + test
buildCanonicalUri now branches on service: S3 gets single-encoding (pathname as-is, already encoded by the URL parser), non-S3 services keep the current double-encoding. Fixes the case where S3 object keys containing %2F were being re-encoded to %252F and returning NoSuchKey. Added a regression test that locks in the S3 signature for a %2F-containing path and asserts it differs from the non-S3 signature for the same URL.

Numeric expiresAt parsing — packages/shared/lib/auth/aws-sigv4.ts
Custom STS endpoints returning Unix seconds were interpreted as milliseconds (e.g. 1772146564 → Jan 1970), causing instant-expired credentials and refres loops. Fixed with a magnitude check: numeric values below 1e12 are treated as seconds and multiplied by 1000. Diverged from the suggested blind * 1000 because that would break endpoints already returning milliseconds.

Silent fallback in getStsAuthFromConfig — packages/shared/lib/auth/aws-sigv4.ts
Post-migration, a malformed structured secret (api_key type with empty value) now returns undefined instead of falling back to the raw custom blob.
Unknown auth types also explicitly return undefined.

Unknown-auth-type leak in redactAwsSigV4Secrets — packages/server/lib/formatters/integration.ts
Legacy custom-blob path: when parsed.stsEndpoint.auth has an unknown type or missing required secret, the auth field is now stripped entirely rather than returning the raw value.

Specific error codes for missing ARN / external ID / region — packages/shared/lib/services/connection.service.ts, packages/shared/lib/utils/error.ts
The lumped !roleArn || !externalId || !region check always returned missing_aws_sigv4_region, which was misleading for connections created before the ExternalId requirement. Split into three sequential guards; registered two new error codes: missing_aws_sigv4_role_arn, missing_aws_sigv4_external_id.

Random RoleSessionName instead of Date.now() — packages/shared/lib/auth/aws-sigv4.ts
Session name is now nango-<16 hex chars> generated via crypto.randomBytes(8) — collision-free in practice, matches the convention in packages/shared/lib/auth/signature.ts. Dropped the timestamp. Diverged from the suggested Math.random() fix because it's not cryptographically secure.

Client-side SSRF guard on template fetch — packages/webapp/src/pages/Integrations/providerConfigKey/Settings/components/AwsSigV4Settings.tsx
loadAwsSigV4TemplateFromUrl now rejects non-HTTPS URLs and literal private/loopback/link-local addresses (including 169.254.169.254 — the AWS instance metadata endpoint). Added an isPrivateOrLocalHost helper covering IPv4 RFC1918, CGNAT, loopback, IPv6 unique-local and link-local. DNS-based bypass still possible, so the comment notes the server must not trust fetched template bodies.

window.open reverse-tabnapping — packages/connect-ui/src/views/Go.tsx
Both external-link buttons (awsSigV4Instructions.url and the "Deploy in AWS Console" launchUrl) now pass 'noopener,noreferrer' features to window.open.

Creation-failure hook mislabels reconnects — packages/server/lib/controllers/auth/postAwsSigV4.ts
The catch block previously reported operation: 'creation' unconditionally. Now tracks attemptedOperation across the try/catch: starts 'unknown', becomes 'creation'/'override' after the existing-connection lookup, and is refined to storedConnection.operation after upsert.

#### Refactors (same-behavior cleanups from earlier review rounds):

Parse aws_sigv4_config value once, reuse everywhere —packages/server/lib/controllers/v1/integrations/providerConfigKey/patchIntegration.ts, packages/shared/lib/auth/aws-sigv4.ts, packages/shared/lib/auth/aws-sigv4.unit.test.ts

Previously value was parsed three times in the caller (for validation, for secret extraction, and for stsMode lookup). Changed extractSecretsFromConfig to accept an already-parsed Record<string, any> instead of a raw JSON string. Caller now parses once, handles parse-failure locally, and reuses the parsed object throughout. Deleted the unused deprecated extractStsAuthFromConfig wrapper. Updated unit tests to pass parsed objects (removed the now-moot invalid-JSON test).

Stage integration_secrets mutations inside the body.custom loop — packages/server/lib/controllers/v1/integrations/providerConfigKey/patchIntegration.ts

Previously integration.integration_secrets was mutated directly across several branches inside the loop, so any early-return after the first mutation could leave integration in a partially-applied state. Introduced a nextSecrets staging variable and a secretsTouched flag; the assignment integration.integration_secrets = nextSecrets only happens once, after the loop completes, and only if any branch actually touched secrets (preserving the "don't overwrite when no aws_sigv4 key was in the patch" behavior).

#### Documented (intentionally minimal change)

FormData + S3 UNSIGNED-PAYLOAD — packages/shared/lib/services/proxy/aws-sigv4.ts
resolveSigV4Payload returns null for FormData, which falls into the S3 UNSIGNED-PAYLOAD branch. Extended the comment in hashPayload to document that signed FormData uploads to S3 are not supported; callers needing signed multipart uploads must serialize to a Buffer before signing. Treating this as a known limitation rather than a silent edge case; full FormData signing deferred as a separate design decision.

#### Pushed back on

"Builtin-mode null-creds silent success" in patchIntegration.ts — false positive
The review claimed that when builtinCredentials is null and existingStsCredentials is also null, validation passes because simulatedSecrets "does include sts_credentials," leading to a silent bad write. simulatedSecrets only injects sts_credentials when builtinCredentials is truthy:

```
  const simulatedSecrets = builtinCredentials
      ? { ...existingSecrets, aws_sigv4: { ...existingAwsSigV4, sts_credentials: builtinCredentials } }
      : existingSecrets;
```

When builtinCredentials is null, simulatedSecrets = existingSecrets (no sts_credentials), so getBuiltinCredentialsFromConfig returns null and validation fails with missing_aws_sigv4_builtin_credentials before the secret-write logic runs. The fallthrough path the review describes is unreachable.

"Hard delete destroys live connection on update" in postAwsSigV4.ts — partially incorrect
The review said hardDelete runs on update-validation failure. It doesn't:

```
  if (storedConnection.operation === 'creation') {
      await connectionService.hardDelete(storedConnection.connection.id);
  }
```

hardDelete is gated on 'creation' only. What is true: on update-validation failure, the new (failing) credentials have already overwritten the old ones via upsertAuthConnection, with no rollback — the user is stuck with broken credentials. This is a real UX issue but present in six other auth controllers (postTba, postApiKey, postTwoStep, postBasic, postSignature, postOAuth2CC). Also, validateConnection requires the DB row (uses connection.id), so simply moving validation before the upsert isn't possible. Better addressed as a coordinated PR across all auth controllers.

packages/connect-ui/src/views/Go.tsx
```diff
@@ -103,7 +116,16 @@ export const Go: React.FC = () => {
     const [connectionFailed, setConnectionFailed] = useState(false);
     const [showErrorDetails, setShowErrorDetails] = useState(false);
 
-    const preconfigured = session && integration ? session.integrations_config_defaults?.[integration.unique_key]?.connection_config || {} : {};
+    const preconfiguredCredentials = session && integration ? session.integrations_config_defaults?.[integration.unique_key]?.credentials || {} : {};
+    const preconfiguredParams = session && integration ? session.integrations_config_defaults?.[integration.unique_key]?.connection_config || {} : {};
```
agusayerza, line 113: On reconnect we are not populating this value and thus we show a new External id. Then, when actually creating the connection we reuse the old external id (correctly). We should probably populate `session.integrations_config_defaults.[integration.unique_key]?.connection_config` with the known `external_id`

packages/server/lib/controllers/auth/postAwsSigV4.ts
```diff
@@ -0,0 +1,422 @@
...
+
+        const [storedConnection] = await connectionService.upsertAuthConnection({
+            connectionId,
+            providerConfigKey,
+            credentials,
+            connectionConfig,
+            metadata: {},
+            config,
+            environment
+        });
```
agusayerza, line 274: We are dropping `connectSession?.tags`, please pipe them. 

```suggestion
        const [storedConnection] = await connectionService.upsertAuthConnection({
            connectionId,
            providerConfigKey,
            credentials,
            connectionConfig,
            metadata: {},
            config,
            environment,
            tags: connectSession?.tags
        });
```

packages/shared/lib/services/proxy/utils.ts
agusayerza, line 664: `config.headers` are the caller-provided request headers, and they’re merged here after the `AWS_SIGV4` branch has already called `signAwsSigV4Request()`. So headers like `x-amz-target` can be sent on the final request without being included in the canonical request `SignedHeaders`, which makes the SigV4 signature invalid for APIs that require them.

packages/database/lib/migrations/20260224120000_add_integration_secrets_to_configs.cjs
```diff
@@ -0,0 +1,11 @@
+exports.up = async function (knex) {
+    return knex.schema.alterTable('_nango_configs', function (table) {
+        table.jsonb('integration_secrets').nullable();
```
agusayerza: This column is encrypted at rest

agusayerza:
Hello,

Overall code and strategy looks good. Couple of bugs/comments to fix before merging:
1. Reconnect UI shows an incorrect external id. (https://github.com/NangoHQ/nango/pull/5041#discussion_r3337459688)
2. We are dropping connection tags (https://github.com/NangoHQ/nango/pull/5041#discussion_r3337476327)
3. We are signing the headers before all potential headers have been added (https://github.com/NangoHQ/nango/pull/5041#discussion_r3337545768)

From the bots:
1. `getSession.ts` credential exposure (https://github.com/NangoHQ/nango/pull/5041#discussion_r3263577703) / (https://github.com/NangoHQ/nango/pull/5041#discussion_r3263589149)
2. AWS SDK version pin (https://github.com/NangoHQ/nango/pull/5041#discussion_r3263577709)
3. Incorrect initial UI (https://github.com/NangoHQ/nango/pull/5041#discussion_r3263589157)
4. Reconnect does not reuse region (https://github.com/NangoHQ/nango/pull/5041#discussion_r3263589164)

packages/server/lib/controllers/connect/getSession.ts
```diff
@@ -82,19 +82,52 @@ export const getConnectSession = asyncWrapper<GetConnectSession>(async (req, res
...
+            // is string | string[]); cast through Record so the spread satisfies the API contract.
+            const storedConfig = existing.connection_config as Record<string, unknown>;
+            data.integrations_config_defaults[key] = {
+                ...existingDefaults,
+                connection_config: {
+                    ...(existingDefaults.connection_config || {}),
+                    ...storedConfig
+                }
+            };
+        }
```
agusayerza: This will leak secrets. We should not spread the full stored `connection_config` into the browser response. Only hydrate `AWS_SIGV4`, and only via an explicit allowlist of safe fields:
- `connection_config.external_id`
- `credentials.role_arn`
- `credentials.region`
Same issue exists in the `integrations_config_defaults` mapping above: even for `AWS_SIGV4` we should not echo the raw `credentials` blob, only the known non-sensitive fields.

Let's add test coverage for this as well:
1. AWS reconnect returns only:
- `connection_config.external_id`
- `credentials.role_arn`
- `credentials.region`
2. AWS reconnect does not echo extra credential keys like `secret_access_key`, `session_token`, etc.
3. Reconnect for a provider with secret `connection_config` does not echo those stored secret fields.
4. Non-`AWS_SIGV4` `integrations_config_defaults.credentials` are not echoed at all.

agusayerza (approved):

PR #5043, opened 2025-11-20
fix(webhook): use event-type header and minor clean up

khaliqgant:
---

**Route GitHub webhooks by `x-github-event` header & prevent duplicate script execution**

This patch switches GitHub webhook routing from relying on the body’s `action` field to the `x-github-event` request header. The change is threaded through `executeScriptForWebhooks()` and both GitHub routing files. While only ~22 LOC are touched, the execution path for all GitHub-App webhooks is affected. Additional tidy-ups include stricter connection-lookup logic, a safeguard against double triggers, and use of the modern `node:crypto` import.

<details>

• Extended `executeScriptForWebhooks()` signature with `webhookHeaderValue` and matched it against `webhook_subscriptions`
• Added local `triggered` flag to avoid multiple `orchestrator.triggerWebhook()` calls when `webhook_subscriptions` contains overlapping entries (e.g., `'*'` and specific events)
• Relaxed connection lookup condition from `(!connectionIdentifier || connectionIdentifier === '')` to `!connectionIdentifier`
• Returned early with `connections.map(...)` in plan-restricted branch to avoid optional chaining on possibly null list
• Replaced legacy `crypto` import with `node:crypto`
• Updated `github-app-webhook-routing.ts` & `github-app-oauth-webhook-routing.ts` to pass `headers['x-github-event']` instead of `webhookType:'action'`
• Guard change for installation-ID check now allows `0` but still blocks `undefined|null`

</details>

<details>

• `packages/server/lib/webhook/internal-nango.ts` (core webhook execution logic)
• `packages/server/lib/webhook/github-app-webhook-routing.ts`
• `packages/server/lib/webhook/github-app-oauth-webhook-routing.ts`

</details>

---

### DIFF
--- a/packages/server/lib/webhook/github-app-oauth-webhook-routing.ts
+++ b/packages/server/lib/webhook/github-app-oauth-webhook-routing.ts
@@ -51,8 +51,7 @@ const route: WebhookHandler = async (nango, headers, body, rawBody) => {
 
     const response = await nango.executeScriptForWebhooks({
         body,
-        headers,
-        webhookHeader: headers['x-github-event'] as string,
+        webhookHeaderValue: headers['x-github-event'] as string,
         connectionIdentifier: 'installation.id',
         propName: 'installation_id'
     });
--- a/packages/server/lib/webhook/github-app-oauth-webhook-routing.ts
+++ b/packages/server/lib/webhook/github-app-oauth-webhook-routing.ts
@@ -1,4 +1,4 @@
-import crypto from 'crypto';
+import crypto from 'node:crypto';
 
 import get from 'lodash-es/get.js';
 
@@ -51,7 +51,8 @@ const route: WebhookHandler = async (nango, headers, body, rawBody) => {
 
     const response = await nango.executeScriptForWebhooks({
         body,
-        webhookType: 'action',
+        headers,
+        webhookHeader: headers['x-github-event'] as string,
         connectionIdentifier: 'installation.id',
         propName: 'installation_id'
     });
@@ -90,7 +91,7 @@ async function handleCreateWebhook(nango: InternalNango, body: any): Promise<Res
         const [connection] = connections;
 
         // if there is no matching connection or if the connection config already has an installation_id, exit
-        if (!connection || connection.connection_config['installation_id']) {
+        if (!connection || (connection.connection_config['installation_id'] !== undefined && connection.connection_config['installation_id'] !== null)) {
             logger.error('no connection or existing installation_id');
             return Err(new NangoError('webhook_no_connection_or_existing_installation_id'));
         }
--- a/packages/server/lib/webhook/github-app-oauth-webhook-routing.ts
+++ b/packages/server/lib/webhook/github-app-oauth-webhook-routing.ts
@@ -90,7 +90,7 @@ async function handleCreateWebhook(nango: InternalNango, body: any): Promise<Res
         const [connection] = connections;
 
         // if there is no matching connection or if the connection config already has an installation_id, exit
-        if (!connection || (connection.connection_config['installation_id'] !== undefined && connection.connection_config['installation_id'] !== null)) {
+        if (!connection || connection.connection_config['installation_id']) {
             logger.error('no connection or existing installation_id');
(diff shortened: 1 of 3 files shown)


khaliqgant:
Note that I verified that this will not be a breaking change for any users

packages/server/lib/webhook/github-app-oauth-webhook-routing.ts
```diff
@@ -51,7 +51,8 @@ const route: WebhookHandler = async (nango, headers, body, rawBody) => {
 
     const response = await nango.executeScriptForWebhooks({
         body,
-        webhookType: 'action',
+        headers,
+        webhookHeader: headers['x-github-event'] as string,
```
michaelSaid: webhookHeader should be `x-github-event` not the value of it?!

packages/server/lib/webhook/github-app-webhook-routing.ts
```diff
@@ -34,7 +34,8 @@ const route: WebhookHandler = async (nango, headers, body) => {
 
     const response = await nango.executeScriptForWebhooks({
         body,
-        webhookType: 'action',
+        headers,
+        webhookHeader: headers['x-github-event'] as string,
         connectionIdentifier: 'installation.id',
```
michaelSaid: dito

packages/server/lib/webhook/github-app-oauth-webhook-routing.ts
khaliqgant: Value is grabbed here: https://github.com/NangoHQ/nango/pull/5043/files#diff-18b5c62ee292ba65d293e837e8c4b1ded0b4231134c48b0f21c8b2aaef48e8b7R111

packages/server/lib/webhook/github-app-webhook-routing.ts
khaliqgant: https://github.com/NangoHQ/nango/pull/5043#discussion_r2545638494

packages/server/lib/webhook/github-app-oauth-webhook-routing.ts
michaelSaid: I think `webhookHeader` is supposed to be the header name/key (like 'x-github-event') because the code does `headers[webhookHeader]` to get the value. Passing the value directly might not work as intended. 
Can we confirm which one it should be?

packages/server/lib/webhook/internal-nango.ts
```diff
@@ -130,6 +138,8 @@ export class InternalNango {
                         });
                     }
 
+                    triggered = true;
+
                     if (webhook === '*') {
                         // Only trigger once since it will match all webhooks
                         break;
```
TBonnin, line 145: can we not just break in any case? and get rid of triggered

packages/server/lib/webhook/github-app-oauth-webhook-routing.ts
```diff
@@ -90,7 +90,7 @@ async function handleCreateWebhook(nango: InternalNango, body: any): Promise<Res
         const [connection] = connections;
 
         // if there is no matching connection or if the connection config already has an installation_id, exit
-        if (!connection || connection.connection_config['installation_id']) {
+        if (!connection || (connection.connection_config['installation_id'] !== undefined && connection.connection_config['installation_id'] !== null)) {
```
TBonnin: can installation_id be false or 0? if not I don't think it is worth the extra condition

packages/server/lib/webhook/internal-nango.ts
khaliqgant: The extra break here is handling an edge case if a user for some reason has

```
webhookSubscriptions: ['anEvent', '*']
```

As the code is now it would actually fire twice

packages/server/lib/webhook/github-app-oauth-webhook-routing.ts
khaliqgant: theoretically it could be 0, but highly unlikely, but it is so unlikely I'm inclined to agree with you

rossmcewan (approved):

PR #5068, opened 2025-11-28
feat(post-connection): add zoho post connection script

hassan254-prog:

- add zoho post connection script to extract extension from `api_domain`. Users who connect to nango using headless client, can create a zoho connection without passing an extension as it will use the default `.com` for the `authorization_url` and use the `accounts-server` url as the `token_url` base-url, which means they might end up creating a connection without supplying the extension. To fill this we can use the post_connection script to extract the domain extension and only add it only if its not present, preserving user's input. This will only be used as a fallback.

- Update the Sellsy webhook handling logic by replacing the use of `webhookType` with `webhookTypeValue`

---

**Add Zoho post-connection hook to infer domain extension**

Adds a Zoho-specific post-connection hook that inspects the Zoho OAuth response metadata to backfill the `extension` connection config when it is not supplied by the user. The script reads `api_domain` from either the existing connection config or the raw OAuth2 credential payload, extracts the regional suffix from the Zoho API hostname, and updates the connection configuration accordingly.
Provider metadata is updated so that the new hook runs automatically after Zoho OAuth completions, and the hook is exported through the connection hook index for runtime registration.

<details>

• Add `packages/server/lib/hooks/connection/providers/zoho/post-connection.ts` implementing Zoho extension extraction and connection config update
• Register the Zoho post-connection hook in `packages/server/lib/hooks/connection/index.ts`
• Declare the hook for the Zoho provider in `packages/providers/providers.yaml` via `post_connection_script: zohoPostConnection`

</details>

<details>

• Zoho post-connection automation
• Connection hook registry
• Provider metadata definitions

</details>

---


packages/server/lib/hooks/connection/providers/zoho/post-connection.ts
```diff
@@ -0,0 +1,62 @@
...
+        }
+
+        // Get everything after "zohoapis" as this is returned as https://www.zohoapis.{extension}
+        const afterMarker = apiDomain.substring(index + marker.length);
+
+        const extension = afterMarker.replace(/^\./, '').split('/')[0];
+
+        return extension || null;
+    } catch (err) {
+        logger.info('Failed to parse api_domain URL', { apiDomain, error: err });
```
TBonnin: let's try to be make sure every log we are adding are meaningful. Would this log be helpful for debugging? is it absolutely required?

packages/server/lib/hooks/connection/providers/zoho/post-connection.ts
```diff
@@ -0,0 +1,62 @@
...
+    }
+}
+
+export default async function execute(nango: Nango) {
+    const connection = await nango.getConnection();
+    const connectionConfig = connection.connection_config || {};
+
+    //if extension is already present, don't override user input
+    const existingExtension = connectionConfig['extension'];
+    if (existingExtension && typeof existingExtension === 'string' && existingExtension.trim() !== '') {
```
TBonnin: i don't think you need `existingExtension && `. It is part of the second subcondition check

packages/server/lib/hooks/connection/providers/zoho/post-connection.ts
```diff
@@ -0,0 +1,62 @@
...
+        return;
+    }
+
+    let apiDomain = connectionConfig['api_domain'];
+
+    //if not in connection_config, try to get it from credentials metadata
+    if (!apiDomain) {
+        const credentials = connection.credentials as OAuth2Credentials;
+        if (credentials?.raw && typeof credentials.raw === 'object' && 'api_domain' in credentials.raw) {
```
TBonnin: same. I don't think that checking for undefined is necessary

packages/server/lib/hooks/connection/providers/zoho/post-connection.ts
```diff
@@ -0,0 +1,62 @@
...
+            const rawApiDomain = credentials.raw['api_domain'];
+            if (typeof rawApiDomain === 'string') {
+                apiDomain = rawApiDomain;
+            }
+        }
+    }
+
+    if (!apiDomain) {
+        logger.info('no api_domain found in connection config or credentials.');
```
TBonnin: same comment about the log. Is it a useful one?

TBonnin (approved):

khaliqgant (approved):

PR #5090, opened 2025-12-02
fix(server): coalesce concurrent token refresh requests

TBonnin:
_To review with Github diff option: Hide whitespaces_

When multiple requests attempt to refresh the same connection simultaneously,
they all queue up waiting for the Redis lock causing unnecessary contention
that can lead to server resource exhaustion and timeouts. It can be
easily reproduced by high-traffic connection requests with`force_refresh=true` 
or for failing connections (refresh will then be attempted every time)

## Solution

Add an in-memory promise cache that deduplicates concurrent refresh requests
before they reach the Redis lock. If a refresh is already in-flight for a
connection, subsequent requests return the existing promise instead of
attempting another refresh.
  
In case of successful refresh, it saves the extra db queries to fetch the refreshed connections and no change in external behavior. 
In case of repetitive refresh failures, no change in behavior neither but preventing all the unnecessary refresh attempts.
The only change in behavior is in the scenario where a refresh fails and then succeed quickly after. With this commit all requests awaiting will return the failed refresh where before the refresh attempts could have been go through and eventually succeed. I think this change of behavior is worth the extra level of reliability/performance. Let me know if you disagree

  This creates two layers of protection:
  - In-memory cache: Handles concurrency within a single instance
  - Redis lock: Handles distributed coordination across multiple processes

 

---

**Add in-memory deduplication for token refreshes**

Introduces an in-memory promise cache inside `refreshCredentialsIfNeeded` so concurrent refresh attempts for the same `(environment, providerConfig, connection)` return the same in-flight promise instead of each acquiring the Redis lock. The previous refresh workflow is refactored into an inner `executeRefresh` helper that retains the existing Redis locking logic and credential update flow, with cache cleanup happening in a `finally` block to avoid leaking entries.

<details>

• Declare module-level `inFlightRefreshes` map keyed by `${environment_id}:${providerConfig.unique_key}:${connectionId}` to track in-flight refresh promises
• Short-circuit `refreshCredentialsIfNeeded` to return an existing cached promise when available
• Wrap the original refresh logic in `executeRefresh` and ensure map entry removal in a `finally` block after the promise settles
• Move acquisition of the Redis lock via `getLocking()` into `executeRefresh` to keep prior distributed coordination behavior

</details>

<details>

• `packages/shared/lib/services/connections/credentials/refresh.ts`

</details>

---

### DIFF
--- a/packages/shared/lib/services/connections/credentials/refresh.ts
+++ b/packages/shared/lib/services/connections/credentials/refresh.ts
@@ -62,6 +62,12 @@ interface RefreshProps {
 
 const logger = getLogger('connectionRefresh');
 
+// In-memory cache for in-flight refresh operations to collapse concurrent requests
+const inFlightRefreshes = new Map<
+    string,
+    Promise<Result<{ connection: DBConnectionDecrypted; refreshed: boolean; credentials: RefreshableCredentials }, NangoInternalError>>
+>();
+
 /**
  * Take a connection and try to refresh or test based on it's type
  * If instantRefresh === false, we will not refresh if not necessary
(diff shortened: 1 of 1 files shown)


packages/shared/lib/services/connections/credentials/refresh.ts
```diff
@@ -62,6 +62,12 @@ interface RefreshProps {
 
 const logger = getLogger('connectionRefresh');
 
+// In-memory cache for in-flight refresh operations to collapse concurrent requests
+const inFlightRefreshes = new Map<
+    string,
+    Promise<Result<{ connection: DBConnectionDecrypted; refreshed: boolean; credentials: RefreshableCredentials }, NangoInternalError>>
+>();
```
TBonnin, line 69: I am planning to implement a simple fixed size map util to use here and in account.service.ts to cache the environment. I'll raise a PR then

kaposke (approved):
I've never seen this "promise caching" technique before, but it does sound pretty smart. Is this something you've done before?

TBonnin:
> I've never seen this "promise caching" technique before, but it does sound pretty smart. Is this something you've done before?

already used in the codebase to prevent concurrent async initialization. Ex: https://github.com/NangoHQ/nango/blob/7fda398e7d7de875030ad22f3d19d8d1ed502509/packages/kvstore/lib/index.ts#L82

marcindobry (approved):

PR #5098, opened 2025-12-03
perf(records): use tuple comparisons for cursor pagination

TBonnin:
Replace OR-based conditions with tuple comparisons in records package. This enables more efficient index usage on (updated_at, id) composite indexes, improving query performance for cursor-based pagination.

Query plan before:
```
Limit  (cost=0.56..99.84 rows=100 width=959)
  ->  Index Scan using records_p67_connection_id_model_updated_at_id_idx on records_p67 records  (cost=0.56..197387.66 rows=198802 width=959)
        Index Cond: ((connection_id = 123456) AND ((model)::text = 'MyModel'::text))
        Filter: ((updated_at > '2025-04-29 12:31:34.836058+00'::timestamp with time zone) OR ((updated_at = '2025-04-29 12:31:34.836058+00'::timestamp with time zone) AND (id > 'UUID'::uuid)))
```
Applies the OR logic as a Filter after the index scan, which requires scanning more rows and then filtering them.

Query plan after:
```
Limit  (cost=0.56..77.76 rows=100 width=959)
  ->  Index Scan using records_p67_connection_id_model_updated_at_id_idx on records_p67 records  (cost=0.56..153483.84 rows=198798 width=959)
        Index Cond: ((connection_id = 123456) AND ((model)::text = 'MyModel'::text) AND (ROW(updated_at, id) > ROW('2025-04-29 12:31:34.836058+00'::timestamp with time zone, 'UUID'::uuid)))
```
The entire condition is pushed into the index scan. Pg can seek directly to the first matching row in the index, then scan forward.

### DIFF
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -165,12 +165,8 @@ export async function getRecords({
                 return Err(error);
             }
 
-            query = query.where(
-                (builder) =>
-                    void builder
-                        .where('updated_at', '>', decodedCursor.sort)
-                        .orWhere((builder) => void builder.where('updated_at', '=', decodedCursor.sort).andWhere('id', '>', decodedCursor.id))
-            );
+            // Tuple comparison for efficient index usage
+            query = query.whereRaw('(updated_at, id) > (?, ?)', [decodedCursor.sort, decodedCursor.id]);
         }
 
         if (externalIds) {
@@ -395,10 +391,7 @@ export async function upsert({
                                 if (merging.strategy === 'ignore_if_modified_after_cursor' && merging.cursor) {
                                     const cursor = Cursor.from(merging.cursor);
                                     if (cursor) {
-                                        qb.whereRaw(`${RECORDS_TABLE}.updated_at < ?`, [cursor.sort]).orWhereRaw(
-                                            `${RECORDS_TABLE}.updated_at = ? AND ${RECORDS_TABLE}.id <= ?`,
-                                            [cursor.sort, cursor.id]
-                                        );
+                                        qb.whereRaw(`(${RECORDS_TABLE}.updated_at, ${RECORDS_TABLE}.id) <= (?, ?)`, [cursor.sort, cursor.id]);
                                     }
                                 }
                             })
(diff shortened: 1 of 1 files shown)


kaposke (approved):

packages/records/lib/models/records.ts
```diff
@@ -165,12 +165,8 @@ export async function getRecords({
...
             }
 
-            query = query.where(
-                (builder) =>
-                    void builder
-                        .where('updated_at', '>', decodedCursor.sort)
-                        .orWhere((builder) => void builder.where('updated_at', '=', decodedCursor.sort).andWhere('id', '>', decodedCursor.id))
-            );
+            // Tuple comparison for efficient index usage
+            query = query.whereRaw('(updated_at, id) > (?, ?)', [decodedCursor.sort, decodedCursor.id]);
```
rossmcewan, line 169: Does this not lose the second part of the condition by only checking > (or I'm misunderstanding something)?
before: `updated_at > decodedCursor.sort OR (updated_at = decodedCursor.sort AND id > decodedCursor.id)`
after: `updatedAt > decodedCursor.sort AND id > decodedCursor.id`

kaposke: It took me a while to wrap my head around it, but looks like the tuple comparison doesn't result in that AND comparison. It actually results in `updated_at > decodedCursor.sort OR (updated_at = decodedCursor.sort AND id > decodedCursor.id)`

Here's the TLDR from a long chatGPT convo:

> TL;DR
> 
> No, tuple comparison is not translated to
> updated_at > lastUpdatedAt AND id > lastId.
> 
> Instead, SQL uses lexicographic comparison, meaning:
> updated_at > lastUpdatedAt
> OR (they’re equal and id > lastId).
> 
> This is the correct behavior for ordered pagination.

TBonnin: from pg docs: https://www.postgresql.org/docs/current/functions-comparisons.html

> For the <, <=, > and >= cases, the row elements are compared left-to-right, stopping as soon as an unequal or null pair of elements is found... For example, ROW(1,2,NULL) < ROW(1,3,0) yields true...

in our case pg compare first the `updatedAt < decodedCursor.sort`, if not equals it returns, if equal move to the second comparison (the `id`)

marcindobry: From how I interpret [the docs](https://www.postgresql.org/docs/current/functions-comparisons.html#ROW-WISE-COMPARISON) this is correct

> For the <, <=, > and >= cases, the row elements are compared left-to-right, stopping as soon as an unequal or null pair of elements is found. If either of this pair of elements is null, the result of the row comparison is unknown (null); otherwise comparison of this pair of elements determines the result. For example, ROW(1,2,NULL) < ROW(1,3,0) yields true, not null, because the third pair of elements are not considered.

The way I understand it is that it compares `updated_at` with `<`. If they aren't equal, the result of that comparison is used. If they are equal, then it compares `id` with `<` and returns that result of the comparison.

🤯

marcindobry (approved):

PR #5105, opened 2025-12-04
fix(two_step): enrich logs with provider specific error logs

hassan254-prog:

- enrich logs with provider specific error logs

---

**Enhance `stringifyError` for provider-aware logging**

Updates `stringifyError` to whitelist provider message fields and capture them under `provider_error_payload`, then reuses the normalized string output wherever two-step credential fetches and Stripe app flows surface errors. Adds a comprehensive Vitest suite that exercises filtering, precedence, malformed inputs, pretty/stack options, and a coarse performance guard. Service and client layers now pass the enriched serialized string to `NangoError` and loggers, replacing bespoke payload assembly.

<details>

• Introduce `PROVIDER_ERROR_MESSAGE_FIELDS` and enrich `stringifyError` in `packages/utils/lib/errors.ts` to extract provider payloads while restricting the serialized keys
• Adopt the new `stringifyError` output when logging and constructing `NangoError` instances in `packages/shared/lib/services/connection.service.ts` and `packages/shared/lib/clients/provider.client.ts`
• Add detailed unit and performance tests for `stringifyError` behavior in `packages/utils/lib/errors.unit.test.ts`

</details>

<details>

• `packages/utils/lib/errors.ts`
• `packages/utils/lib/errors.unit.test.ts`
• `packages/shared/lib/clients/provider.client.ts`

</details>

---


packages/shared/lib/clients/provider.client.ts
hassan254-prog, line 1: this entire file needs to be refactored.

packages/utils/lib/errors.ts
```diff
@@ -25,7 +25,29 @@ export function errorToObject(err: unknown): ErrorObject {
...
+        ...(serialized && typeof serialized === 'object' ? serialized : {})
+    };
+
+    if (typeof err === 'object' && err != null) {
+        const anyErr = err as any;
+
+        // handle axios response data
+        if (anyErr.response?.data) {
+            enriched['provider_error_payload'] = anyErr.response.data;
+        }
```
TBonnin, line 57: I am personally not a fan of logging arbitrary response coming from the internet. What if the response contains any sensitive data?

hassan254-prog: fair point. However, it’s difficult trying to debug issues when we only get generic error codes with no context, especially since providers can return same error codes with different error messages, how do you suggest we handle that?

TBonnin: good question. Can you show what would a failed response would log with this PR compared to before?

hassan254-prog: This is the error response the user is currently getting:
```
{
  "provider": "valley",
  "error": {
    "message": "Error fetching Two Step credentials",
    "payload": {
      "name": "AxiosError",
      "message": "Request failed with status code 401"
    }
  }
}
```

after these changes, it will be:

```
{
  "provider": "valley",
  "error": {
    "message": "Error fetching Two Step credentials",
    "payload": "{\n  \"message\": \"Request failed with status code 401\",\n  \"name\": \"AxiosError\",\n  \"provider_error_payload\": {\n    \"error\": \"invalid_credentials\",\n    \"message\": \"Invalid email or password\"\n  }\n}"
  }
}
```

TBonnin: can we be a bit defensive here and only log the `error` field of the response if it exists. I know not all provider failure response is gonna be structured this way but I would prefer filtering too much than not enough

hassan254-prog: updated with the latest [commit](https://github.com/NangoHQ/nango/pull/5105/commits/de72c922cd856e78ba0789b5c3dd5e4f1b982bfd). I think this is better, and we can add the field names as the list grows

marcindobry (commented):
Now that this function does a bunch more stuff, could we add a few unit tests for the various cases like filtering out fields that aren't whitelisted?

hassan254-prog:
> Now that this function does a bunch more stuff, could we add a few unit tests for the various cases like filtering out fields that aren't whitelisted?

I agree, updated with the latest [commit](https://github.com/NangoHQ/nango/pull/5105/commits/0cedc302d01a463ffc5eff67f9b1fd08d5c177d8).

packages/utils/lib/errors.ts
TBonnin: thank you. do we know how much of a perf overhead this new logic is introducing? Maybe we can add a test to make sure it doesn't get out of hand

marcindobry (approved):

packages/utils/lib/errors.ts
hassan254-prog: added a test for this in the latest [commit](https://github.com/NangoHQ/nango/pull/5105/commits/0ed1d08108a0909bc731d242d482cff9d2ced314).

packages/utils/lib/errors.unit.test.ts
```diff
@@ -115,5 +115,73 @@ describe('stringifyError', () => {
...
         });
+
+        describe('stringifyError performance - NEW VERSION', () => {
+            it('should not significantly regress from baseline performance', () => {
+                // baseline from old implementation (measured in gh actions ci environment)
+                const baseline: Record<string, number> = {
+                    'Simple Error': 0.002882297399999993,
+                    'Axios Error': 0.004207618500000001,
+                    'Boom Error': 0.0029855881999999953
+                };
```
TBonnin: i wouldn't worry about specific baseline. Pick one upper bound number that we are we are happy with for 100_000 iterations and let's make sure it doesn't go above it in the future

packages/utils/lib/errors.ts
TBonnin: so how much slower it is?

TBonnin (approved):

packages/utils/lib/errors.ts
hassan254-prog: For simple error objects, the observed performance overhead is negligible `(~+0.2%)`, while for axios and boom error types incur a latency increase of approximately `1.51×`.

PR #5116, opened 2025-12-08
fix: revalidate usage on get

kaposke:
**Ensure `UsageTracker` revalidates cache entries on read paths**

Updates `packages/account-usage/lib/usage.ts` so that `UsageTracker.get` and `UsageTracker.getAll` trigger `revalidate` when cached values are missing or their `revalidateAfter` timestamp is in the past, aligning read-path behavior with the existing logic in `incr`. Extends `packages/account-usage/lib/usage.integration.test.ts` with `vi` spies and fake timers to cover null, stale, invalid, and fresh cache scenarios for both `get` and `getAll`, ensuring revalidation is invoked (or not) as expected.

<details>

• Added conditional revalidation in `UsageTracker.get` when `entry.value` is `null` or stale (`entry.value.revalidateAfter < now`)
• Mirrored the same revalidation trigger inside `UsageTracker.getAll` for each metric retrieved
• Expanded integration tests to spy on `revalidate`, simulate stale windows via fake timers, and assert invocation counts/arguments for `get`, `getAll`, and `incr` flows

</details>

<details>

• `packages/account-usage/lib/usage.ts`
• `packages/account-usage/lib/usage.integration.test.ts`

</details>

---

### DIFF
--- a/packages/account-usage/lib/usage.ts
+++ b/packages/account-usage/lib/usage.ts
@@ -88,6 +88,9 @@ export class UsageTracker implements IUsageTracker {
         if (entry.isErr()) {
             return Err(entry.error);
         }
+        if (entry.value === null || entry.value.revalidateAfter < now.getTime()) {
+            void this.revalidate({ accountId, metric });
+        }
         return Ok({
             accountId,
             metric,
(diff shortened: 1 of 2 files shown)


TBonnin (approved):
can we add/modify tests to check the new behavior

packages/account-usage/lib/usage.ts
```diff
@@ -88,6 +88,9 @@ export class UsageTracker implements IUsageTracker {
         if (entry.isErr()) {
             return Err(entry.error);
         }
+        if (entry.value === null || entry.value.revalidateAfter < now.getTime()) {
```
TBonnin, line 91: how can value be null?

kaposke: On the first time trying to get it from the cache, when it doesn't exist yet. Down there it defaults to 0, but ideally we take the hint to revalidate.

PR #5117, opened 2025-12-08
feat(records): add cursor-based deletion to deleteRecords

TBonnin:
Add support for bounded record deletion:
- Add `toCursorIncluded` parameter to delete records up to a cursor (inclusive)
- Add `limit` parameter to limit number of records deleted
- Return `lastDeletedCursor`

The cursor logic will be used by the DELETE /records endpoint so
customers can explicitely prune records until a specific record/cursor

---

**Cursor- and limit-aware record deletion with count updates**

Extends `deleteRecords` in `packages/records/lib/models/records.ts` to support bounded deletions via optional `limit` and `toCursorIncluded`, returning both the total deleted row count and the `lastDeletedCursor`. The update now tracks deleted byte totals, reuses `incrCount` to adjust `record_counts`, and prunes the count row when the remaining total reaches zero. Integration coverage in `packages/records/lib/models/records.integration.test.ts` has been expanded to exercise the new deletion pathways (limits, cursors, batch execution, validation) while consolidating setup/teardown to a single migration run per suite. The `incrCount` helper now returns the updated `RecordCount`, enabling callers to check whether counts hit zero.

<details>

• Augmented `deleteRecords` to accept `limit`, `toCursorIncluded`, and return `{ totalDeletedRecords, lastDeletedCursor }`, with advisory locking and ordered batched deletions that respect cursor boundaries.
• Accumulates deleted row counts and byte sizes, invokes `incrCount` for deltas, and removes the record-count row when totals reach zero.
• Reworked `incrCount` to always upsert counts, clamp at zero via `GREATEST`, and return the new `RecordCount` (with `size_bytes` normalized to number).
• Expanded `records.integration.test.ts` with cursor/limit deletion scenarios, invalid input checks, and updated expectations for count bookkeeping while switching to `beforeAll`/`afterAll` migration control.

</details>

<details>

• `packages/records/lib/models/records.ts`
• `packages/records/lib/models/records.integration.test.ts`

</details>

---

### DIFF
--- a/packages/records/lib/models/records.integration.test.ts
+++ b/packages/records/lib/models/records.integration.test.ts
@@ -1,6 +1,6 @@
 import dayjs from 'dayjs';
 import * as uuid from 'uuid';
-import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
 
 import { RECORDS_TABLE, RECORD_COUNTS_TABLE } from '../constants.js';
 import { db } from '../db/client.js';
@@ -12,11 +12,11 @@ import type { FormattedRecord, UnencryptedRecordData, UpsertSummary } from '../t
 import type { MergingStrategy, Result } from '@nangohq/types';
 
 describe('Records service', () => {
-    beforeEach(async () => {
+    beforeAll(async () => {
         await migrate();
     });
 
-    afterEach(async () => {
+    afterAll(async () => {
         await db(RECORDS_TABLE).truncate();
         await db(RECORD_COUNTS_TABLE).truncate();
     });
@@ -905,6 +905,10 @@ describe('Records service', () => {
 
     describe('paginateCounts', () => {
         it('should paginate through record counts', async () => {
+            // Clear existing records and counts
+            await db(RECORDS_TABLE).truncate();
+            await db(RECORD_COUNTS_TABLE).truncate();
+
             const res1 = await upsertNRecords(15);
             const res2 = await upsertNRecords(25);
             const res3 = await upsertNRecords(35);
@@ -987,7 +991,7 @@ describe('Records service', () => {
     });
 
     describe('deleteRecords', () => {
-        it('should delete records for given connection/model', async () => {
+        it('should delete all records for given connection/model', async () => {
             const connectionId = rnd.number();
             const environmentId = rnd.number();
             const model = rnd.string();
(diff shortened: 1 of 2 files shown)


packages/records/lib/models/records.integration.test.ts
```diff
@@ -12,11 +12,11 @@ import type { FormattedRecord, UnencryptedRecordData, UpsertSummary } from '../t
...
-    beforeEach(async () => {
+    beforeAll(async () => {
         await migrate();
     });
 
-    afterEach(async () => {
+    afterAll(async () => {
         await db(RECORDS_TABLE).truncate();
         await db(RECORD_COUNTS_TABLE).truncate();
     });
```
TBonnin, line 22: unrelated but much faster not to reset the db after each test. All tests but one below are independent since they use random environment

packages/records/lib/models/records.ts
```diff
@@ -694,60 +694,127 @@ export async function update({
...
+                    lastDeletedCursor = Cursor.new({ id: lastDeletedRecord.id, last_modified_at: lastDeletedRecord.updated_at });
+                }
             } while (deletedRecords > 0);
 
-            await deleteCount(trx, {
-                environmentId,
-                connectionId,
-                model
-            });
+            if (limit) {
```
marcindobry: If the limit is larger than the number of records we've deleted (meaning there is nothing left in the DB) would it make sense to call `deleteCount` instead?

marcindobry: Also, what if the user specifies `toCursorIncluded` instead of/in addition to `limit`? Shouldn't we `incrCount` in that case too?

packages/records/lib/models/records.integration.test.ts
```diff
@@ -1006,6 +1010,166 @@ describe('Records service', () => {
...
+            const twelfthRecordCursor = allRecords.records[11]?._nango_metadata.cursor;
+            expect(twelfthRecordCursor).toBeDefined();
+
+            const deleteResult = (
+                await Records.deleteRecords({
+                    connectionId,
+                    environmentId,
+                    model,
+                    toCursorIncluded: twelfthRecordCursor!,
+                    batchSize: 3
```
kaposke: 12 is a multiple of 3. Can this test use 5, or any batchSize that is not a multiple of the cursor position?

TBonnin: sure

packages/records/lib/models/records.ts
TBonnin: the idea was to make the behavior (remove the count row or not) depends on (and be consistent with) the input (aka: limit isn't defined => I want everything to be deleted including the count/size) and not depends on the content of the db.
Your second point though shows that depending on the combination of the input params the behavior might be different. Maybe it is simpler to reconsider the decision and delete the row when everything is deleted regardless of the input. I'll revisit

TBonnin: @marcindobry addressed https://github.com/NangoHQ/nango/pull/5117/commits/9e0e369c99675035e2613b1330e13c53ed59b456#diff-6e68d35e254c27b00db6fb2a4642bdeb350eaa08c7429e34e0df720a75c2b03aL796-L1004

making an extra query when count reach 0 to avoid error-prone logic when limit/cursor === exact number of records

marcindobry (approved):

PR #5128, opened 2025-12-10
feat: add /DELETE records endpoint

TBonnin:
API endpoint to delete records for a given connection/model
Parameters:
- hard (deletes entries) or soft (empties payload)
- until_cursor (deletes all records until cursor)
- limit (optional - limit the number of deleted records)

# How to test

```
curl --request DELETE \  
  --url http://localhost:3003/records\?model\=YOUR_MODEL\&until_cursor\=CURSOR\&mode\=soft \
  --header 'Authorization: Bearer ...' \
  --header 'Connection-Id: ...' \
  --header 'Provider-Config-Key: ...'

```

---

**Add cursor-aware `DELETE /records` API with backend pruning refinements**

Introduces a new public `DELETE /records` endpoint with `soft` and `hard` deletion modes, cursor-bounded pagination, header/query validation via Zod, and integration into the public route/auth pipeline. The controller relies on `connectionService.getConnection`, applies a default page size of 1000 records, and reports `{ count, has_more }` based on the last deleted cursor.

Underlying record store changes ensure `records.deleteRecords` supports both deletion modes without mutating `updated_at` during soft deletes, always echoes the external identifier in the response `id`, and prevents updates to already-deleted rows. Types, OpenAPI docs, and extensive integration tests were updated to cover the new behaviors and edge cases (limits, cursor drift, protection, and updated records).

<details>

• Added `deletePublicRecords` controller with Zod-based validation, connection lookup, `records.deleteRecords` invocation, and `has_more` computation, exposed through the public router.
• Enhanced `records.deleteRecords` implementation to handle soft vs. hard modes, iterate in batches up to the requested cursor/limit, return the last deleted cursor, and keep soft-deleted records' `updated_at` stable while recording `deleted_at`.
• Adjusted `getRecords` to set response `id` from `external_id`, blocked updates on already-deleted rows via `getRecordsToUpdate`, and refreshed OpenAPI spec/types and integration tests for deletion scenarios.

</details>

<details>

• packages/server/lib/controllers/records
• packages/server/lib/routes.public.ts
• packages/records/lib/models/records.ts
• packages/types/lib/record/api.ts
• docs/spec.yaml
• packages/records/lib/models/records.integration.test.ts

</details>

---

### DIFF
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -176,6 +176,7 @@ export async function getRecords({
             db.raw(`
                 tableoid::regclass as partition,
                 id,
+                external_id,
                 json,
                 to_json(created_at) as first_seen_at,
                 to_json(updated_at) as last_modified_at,
@@ -199,6 +200,7 @@ export async function getRecords({
             const decryptedData = await decryptRecordData(item);
             results.push({
                 ...decryptedData,
+                id: item.external_id, // record payload can be empty (when pruned), always use external_id as id
                 _nango_metadata: {
                     first_seen_at: item.first_seen_at,
                     last_modified_at: item.last_modified_at,
@@ -791,8 +793,8 @@ export async function deleteRecords({
                     case 'soft':
                         query.update({
                             deleted_at: now,
-                            updated_at: now,
                             json: {}
+                            // IMPORTANT: updated_at isn't updated because it would cause the record cursor to also change
                         });
                         break;
                     case 'hard':
@@ -1088,6 +1090,7 @@ async function getRecordsToUpdate({
             connection_id: connectionId,
             model
         })
+        .whereNull('deleted_at') // only non-deleted records can be updated
         .whereIn('external_id', keys)
         .whereNotIn(['external_id', 'data_hash'], keysWithHash);
 }
--- a/packages/server/lib/controllers/records/deleteRecords.ts
+++ b/packages/server/lib/controllers/records/deleteRecords.ts
@@ -1,0 +1,83 @@
+import * as z from 'zod';
+
+import { records } from '@nangohq/records';
+import { connectionService } from '@nangohq/shared';
+import { zodErrorToHTTP } from '@nangohq/utils';
+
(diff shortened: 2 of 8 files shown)


packages/records/lib/models/records.ts
```diff
@@ -791,8 +793,8 @@ export async function deleteRecords({
                     case 'soft':
                         query.update({
                             deleted_at: now,
-                            updated_at: now,
                             json: {}
+                            // IMPORTANT: updated_at isn't updated because it would cause the record cursor to also change
```
TBonnin, line 797: quite an important behavior. Updating the `updated_at` would make the records cursor be updated and "move" the records to the end of the list when deleted explicitely. I thought it would make more sense if the position of a record in the ordered list doesn't change when deleted.

packages/records/lib/models/records.ts
```diff
@@ -199,6 +200,7 @@ export async function getRecords({
             const decryptedData = await decryptRecordData(item);
             results.push({
                 ...decryptedData,
+                id: item.external_id, // record payload can be empty (when pruned), always use external_id as id
```
TBonnin, line 203: doesn't change anything for non-pruned records (`payload.id == external_id`) but necessary for pruned records where the payload is empty

packages/server/lib/controllers/records/deleteRecords.ts
```diff
@@ -0,0 +1,83 @@
...
+        return;
+    }
+
+    // IMPORTANT:
+    // if until_cursor doesn't exist anymore (e.g. records was updated or deleted)
+    // we can't be sure if there are more records to delete
+    // so we optimistically return has_more = true in that case
+    // This may lead to one extra call to this endpoint
+    // but ensures all records are deleted as expected
+    // without extra code/query to maintain and execute on every request
```
TBonnin, line 76: tradeoff: possible extra call when until_cursor has been modified vs extra logic to execute on every request.

kaposke: I think this makes sense

kaposke (approved):

marcindobry (approved):

PR #5149, opened 2025-12-15
fix(auth): allow/disallow regular signup based on environment variable

rossmcewan:
---

**Signup flow gated by new env toggle**

Introduces a configurable guard on non-invite signup flows by adding the `AUTH_ALLOW_SIGNUP` environment flag (default `true`). The flag is parsed via `packages/utils/lib/environment/parse.ts`, injected into both regular and WorkOS-managed signup controllers, and surfaced to the client configuration payload through `packages/server/lib/controllers/v1/getEnvJs.ts` and `packages/types/lib/web/env.ts`.

<details>

• Parsed new `AUTH_ALLOW_SIGNUP` flag in `packages/utils/lib/environment/parse.ts` with a default of `true`.
• Blocked regular account creation in `packages/server/lib/controllers/v1/account/signup.ts` when `envs.AUTH_ALLOW_SIGNUP` is falsy, returning `403`.
• Guarded WorkOS-managed regular signups in `packages/server/lib/controllers/v1/account/managed/getCallback.ts` using the same `envs.AUTH_ALLOW_SIGNUP` check.
• Exposed the flag to the dashboard environment payload via `packages/server/lib/controllers/v1/getEnvJs.ts` and updated the `WindowEnv` type in `packages/types/lib/web/env.ts`.

</details>

<details>

• `packages/server/lib/controllers/v1/account/signup.ts`
• `packages/server/lib/controllers/v1/account/managed/getCallback.ts`
• `packages/server/lib/controllers/v1/getEnvJs.ts`
• `packages/types/lib/web/env.ts`

</details>

---

### DIFF
--- a/packages/server/lib/controllers/v1/account/managed/getCallback.ts
+++ b/packages/server/lib/controllers/v1/account/managed/getCallback.ts
@@ -99,7 +99,7 @@ export const getManagedCallback = asyncWrapper<GetManagedCallback>(async (req, r
         } else {
             // Regular signup
             if (!envs.AUTH_ALLOW_SIGNUP) {
-                res.status(500).send({ error: { code: 'server_error', message: 'Signup is disabled.' } });
+                res.status(403).send({ error: { code: 'forbidden', message: 'Signup is disabled.' } });
                 return;
             }
             const resAccount = await accountService.createAccount({ name, email: authorizedUser.email });
--- a/packages/server/lib/controllers/v1/account/managed/getCallback.ts
+++ b/packages/server/lib/controllers/v1/account/managed/getCallback.ts
@@ -5,6 +5,7 @@ import { acceptInvitation, accountService, expirePreviousInvitations, getInvitat
 import { basePublicUrl, flagHasUsage, getLogger, nanoid, report } from '@nangohq/utils';
 
 import { getWorkOSClient } from '../../../../clients/workos.client.js';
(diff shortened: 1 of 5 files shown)


packages/server/lib/controllers/v1/account/managed/getCallback.ts
```diff
@@ -97,6 +98,10 @@ export const getManagedCallback = asyncWrapper<GetManagedCallback>(async (req, r
             account = (await accountService.getAccountById(db.knex, invitation.account_id))!;
         } else {
             // Regular signup
+            if (!envs.AUTH_ALLOW_SIGNUP) {
+                res.status(500).send({ error: { code: 'server_error', message: 'Signup is disabled.' } });
```
marcindobry: Should we return a 4xx instead? These would be "false positive" 500s in a sense that there is nothing we can do about it and can be annoying to be alerted about

marcindobry (approved):
Approving to unblock, left 1 question

packages/server/lib/controllers/v1/account/managed/getCallback.ts
rossmcewan: was torn as there's nothing the user can change - but you're right, 500s will give the wrong impression from an observability perspective. I'll change to 403 - is a better fit too.

PR #5152, opened 2025-12-16
feat(records): add pruned_at column to records table

TBonnin:
companion PR (Read it first): https://github.com/NangoHQ/nango/pull/5152

---

**Introduce record pruning API and metadata support**

Adds a nullable `pruned_at` column to records, exposes it via the records API and types, and replaces the former delete endpoint with a dedicated prune workflow. The records model now supports three deletion modes (hard/soft/prune), client SDKs and orchestrator callers are updated to consume the new PATCH `/records/prune` route, and documentation/tests reflect the revised metadata and request format.

<details>
<summary><strong>Key Changes</strong></summary>

• Added migration `packages/records/lib/db/migrations/20251215191200_records_pruned_at.ts` introducing nullable timestamp column `pruned_at` on `records`.
• Extended `packages/records/lib/models/records.ts` deletion logic to accept `mode: 'hard' | 'soft' | 'prune'`, emit `{ count, lastCursor }`, and track `pruned_at`/size deltas without adjusting counts for prune operations.
• Replaced the public DELETE `/records` route with PATCH `/records/prune` (`packages/server/lib/controllers/records/patchPruneRecords.ts`) including validation, response shape, and routing updates.
• Exposed pruning in clients and docs: added `pruneRecords` helper in `packages/node-client/lib/index.ts`, updated OpenAPI (`docs/spec.yaml`) and Node SDK docs.
• Updated orchestrator/cron consumers and tests to pass explicit deletion modes and expect the new `{ count, lastCursor }` structure.

</details>

<details>
<summary><strong>Affected Areas</strong></summary>

• `records` service deletion logic
• `/records` public API contract
• `node-client` SDK
• `docs/spec.yaml` and Node SDK docs
• Orchestrator cleanup flows and related tests

</details>

---
*This summary was automatically generated by @propel-code-bot*

PR #5153, opened 2025-12-16
feat: records pruning

TBonnin:
Prerequesite: https://github.com/NangoHQ/nango/pull/5152

I misunderstood the requirements and thought pruning was linked to records deletion but records pruning is actually its own operation, independent to the concept of deleting a record.
This commit is removing the DELETE /records endpoint and replace it by PATCH /records/prune.
Records now have a pruned_at metadata to be able to track already pruned records since payload cannot be used at query time since it is encrypted. It is also a good idea to have it as an audit trail

I have kept the internal deleteRecords function in order to take advantage of the shared logic (pagination, etc...) even though the soft delete mode isn't currently used

Documentation update to come in next PR

---

**Add record pruning endpoint and metadata tracking**

This PR replaces the previous `DELETE /records` API with a dedicated `PATCH /records/prune` endpoint that empties record payloads, sets a new `pruned_at` timestamp, and leaves metadata/counts in place. The change propagates through the records service, OpenAPI spec, node SDK, orchestrator cleanup flow, and both unit and integration tests so pruning can be initiated programmatically and reported back to clients.

<details>

• Expose `pruned_at` in records metadata, OpenAPI schema, integration tests, and the node client response types.
• Extend `records.deleteRecords` to support a new `'prune'` mode (no count decrement, size reduction only) and return `{ count, lastCursor }` for all modes.
• Replace the public `DELETE /records` route with `PATCH /records/prune`, validating request bodies and wiring the new handler through routing and SDK support.
• Update orchestrator and cleanup cron callers to pass an explicit deletion mode and to adapt to the updated result shape.
• Document the pruning workflow in the public API spec and Node SDK guide.

</details>

<details>

• Public records API routing
• Records service deletion logic
• Node SDK (`packages/node-client/lib/index.ts`)
• OpenAPI spec and documentation
• Orchestrator cleanup flows and cron

</details>

---

### DIFF
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -181,6 +181,7 @@ export async function getRecords({
                 to_json(created_at) as first_seen_at,
                 to_json(updated_at) as last_modified_at,
                 to_json(deleted_at) as deleted_at,
+                to_json(pruned_at) as pruned_at,
                 CASE
                     WHEN deleted_at IS NOT NULL THEN 'DELETED'
                     WHEN created_at = updated_at THEN 'ADDED'
@@ -206,6 +207,7 @@ export async function getRecords({
                     last_modified_at: item.last_modified_at,
                     last_action: item.last_action,
                     deleted_at: item.deleted_at,
+                    pruned_at: item.pruned_at,
                     cursor: Cursor.new(item)
                 }
             });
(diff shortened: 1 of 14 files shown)


packages/records/lib/models/records.ts
```diff
@@ -703,39 +705,39 @@ export async function update({
...
     limit,
     toCursorIncluded,
-    batchSize = 1000,
-    mode = 'hard'
+    batchSize = 1000
 }: {
     connectionId: number;
     environmentId: number;
     model: string;
+    mode: 'hard' | 'soft' | 'prune';
```
TBonnin, line 725: For now soft delete, hard delete, prune are disjoint options of the deleteRecords function but could make it a bitwise parameter (or similar) if we need to combine them in the future.

rossmcewan (commented):
I'm not sure a PATCH is the correct approach for this. Semantically, a prune is like a DELETE (usually) and that is reflected in the internal implementation too. Would it not be more correct to offer the DELETE option with a mode? I missed the conversation on this so may have already been discussed.

Also not a big fan of verbs in the path.

packages/records/lib/models/records.ts
```diff
@@ -703,39 +705,39 @@ export async function update({
...
-        tags: { 'nango.connectionId': connectionId, 'nango.model': model }
+        tags: { 'nango.environmentId': environmentId, 'nango.connectionId': connectionId, 'nango.model': model, 'nango.deletionMode': mode }
     });
 
     let partition: string | undefined = undefined;
-    let totalDeletedRecords = 0;
-    let totalDeletedSizeInBytes = 0;
-    let deletedRecords = 0;
-    let lastDeletedCursor: string | null = null;
+    let totatRecords = 0;
```
kaposke:
```suggestion
    let totalRecords = 0;
```

kaposke (approved):
Valid points by Ross.

So the difference between soft deletion and pruning is that pruning is just deleting the data, but the record is still accounted for?

TBonnin:
> I'm not sure a PATCH is the correct approach for this. Semantically, a prune is like a DELETE (usually) and that is reflected in the internal implementation too. Would it not be more correct to offer the DELETE option with a mode? I missed the conversation on this so may have already been discussed.

That was my initial understanding of the requirements (and exactly how I initially implemented it) but emptying the payload shouldn't delete the entry or even marking it as deleted. The records are still accounted for, we just stop holding any data (other than the id and hash). Bastien is worried that using the DELETE http verb would be confusing.

> 
> Also not a big fan of verbs in the path.

Me neither but the HTTP verbs, as demonstrated in this example, are very limited. If we don't want to use DELETE we are left with using a verb in the path, unless you have a better idea?

TBonnin:
> So the difference between soft deletion and pruning is that pruning is just deleting the data, but the record is still accounted for?

exactly, pruning is just emptying the payload but the records still exists as before

packages/records/lib/models/records.ts
marcindobry: Something to think about maybe for the future - if the same method is basically capable of updating and deleting records in a batched way, maybe this can eventually live as an "updateRecords" method instead

marcindobry (approved):

PR #5164, opened 2025-12-17
feat(connect-ui): allow override for client id and client secret for OAUTH2

khaliqgant:
If a user wants to allow the client id and secret to be overriden for OAUTH2 they can use `integrations_config_defaults`
```
integrations_config_defaults: {
                    google: {
                        connection_config: {
                            oauth_client_id_override: '',
                            oauth_client_secret_override: ''
                        }
                    }
                }
````

---

**Allow OAuth2 client credential overrides in Connect UI**

Adds conditional handling in `packages/connect-ui/src/views/Go.tsx` so that OAuth2 integrations with `integrations_config_defaults` exposing empty `oauth_client_id_override` and `oauth_client_secret_override` values render corresponding optional credential fields. Adjusts the memoization dependencies and visibility logic so the override fields appear only when allowed and remain hidden for other auth modes or pre-populated secrets.

<details>

• Injects `oauth_client_id_override` and `oauth_client_secret_override` into the `baseForm.shape` when `provider.auth_mode` is `OAUTH2` and `preconfigured` contains the keys with empty values
• Ensures `orderedFields` includes the newly added override credentials and keeps the `useMemo` hook reactive to `preconfigured` changes
• Updates `FormItem` visibility conditions to surface the override inputs only when `preconfigured` provides empty placeholders for OAuth2

</details>

<details>

• `packages/connect-ui/src/views/Go.tsx`

</details>

---

### DIFF
--- a/packages/connect-ui/src/views/Go.tsx
+++ b/packages/connect-ui/src/views/Go.tsx
@@ -187,6 +187,18 @@ export const Go: React.FC = () => {
             }
         }
 
+        if (provider.auth_mode === 'OAUTH2' && Object.keys(preconfigured).length > 0) {
+            // For OAUTH2, allow users to override client credentials if preconfigured with empty values
+            const allowedOverrides = ['oauth_client_id_override', 'oauth_client_secret_override'];
+            for (const key of allowedOverrides) {
+                if (key in preconfigured && !additionalFields[key] && !(key in baseForm.shape)) {
+                    baseForm.shape[key] = z.string().optional();
+                    order += 1;
+                    orderedFields[`credentials.${key}`] = order;
+                }
+            }
+        }
+
         // Only add objects if they have something otherwise it breaks react-form
         const fields = z.object({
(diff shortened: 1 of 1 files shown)


kaposke (approved):
eLooks ok I think. Guessing there are no precedents of "ask final customer if empty"? 

Also, would we want that to be the behavior when the client_id and secret are empty in the first place (via integration settings and not overrides)?

My only concern is the condition on the className does look kind of hacky. This addition makes me think that wether it's hidden or not should be in the orderedField's metadata now (because the condition got complex and in 2 places).

Approving but happy to discuss further.

khaliqgant:
Thanks @kaposke! Makes sense, had a chat about this `className` check https://myseniordev.com/repos/NangoHQ/nango/pulls/5164?focus=chat%3A71f8b252-c766-44b2-9255-d7d232d067ce%3A6 -- wasn't satisfied with the proposal. Will merge this as is with a documentation PR to follow and can improve on this later.

hassan254-prog (approved):

PR #5169, opened 2025-12-17
fix(proxy): preserve BigInt precision in JSON responses

TBonnin:
Avoid precision loss with big integers by forwarding JSON responses as raw buffers instead of re-parsing/re-serializing.

Keeping the JSON validation behavior for now but we could revisit it at some point. The objective here is to unblock customers and fix the bigInt precision loss bug

---

**Preserve BigInt precision in proxy JSON responses**

Exports `handleResponse` from `packages/server/lib/controllers/proxy/allProxy.ts` and changes JSON proxy handling so the upstream buffer is returned directly. JSON payloads are still validated by parsing a copy, while the response headers and logging/metrics flows remain intact. Unit coverage in `packages/server/lib/controllers/proxy/allProxy.unit.test.ts` now exercises 204 responses, invalid JSON handling, and a BigInt regression case.

<details>

• Updated `handleResponse` to stream non-buffered payloads unchanged and forward buffered responses using `res.send(Buffer.concat(responseData))` while setting `Content-Type` explicitly for JSON responses.
• Retained JSON validation by parsing a cloned buffer to detect invalid payloads without mutating the original stream, preserving BigInt precision.
• Exported `handleResponse` and expanded Vitest coverage to include BigInt, invalid JSON, and 204 No Content scenarios with mock `Response` objects.

</details>

<details>

• packages/server/lib/controllers/proxy/allProxy.ts
• packages/server/lib/controllers/proxy/allProxy.unit.test.ts

</details>

---

### DIFF
--- a/packages/server/lib/controllers/proxy/allProxy.ts
+++ b/packages/server/lib/controllers/proxy/allProxy.ts
@@ -303,7 +303,7 @@ export function parseHeaders(req: Pick<Request, 'rawHeaders'>) {
     return forwardedHeaders;
 }
 
-async function handleResponse({ res, responseStream, logCtx }: { res: Response; responseStream: AxiosResponse; logCtx: LogContext }) {
+export async function handleResponse({ res, responseStream, logCtx }: { res: Response; responseStream: AxiosResponse; logCtx: LogContext }) {
     const contentType = responseStream.headers['content-type'] || '';
     const contentDisposition = responseStream.headers['content-disposition'] || '';
     const transferEncoding = responseStream.headers['transfer-encoding'] || '';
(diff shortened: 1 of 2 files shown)


packages/server/lib/controllers/proxy/allProxy.ts
```diff
@@ -303,7 +303,7 @@ export function parseHeaders(req: Pick<Request, 'rawHeaders'>) {
     return forwardedHeaders;
 }
 
-async function handleResponse({ res, responseStream, logCtx }: { res: Response; responseStream: AxiosResponse; logCtx: LogContext }) {
+export async function handleResponse({ res, responseStream, logCtx }: { res: Response; responseStream: AxiosResponse; logCtx: LogContext }) {
```
TBonnin, line 306: not ideal to export this private logic just to be able to unit test it but the alternative  (setting up all the machinery in the tests or refactoring/extracting this logic into its own file) is also adding complexity

agusayerza (approved):

marcindobry (approved):

PR #5170, opened 2025-12-17
feat(cli): Add interactive prompts for commands

agusayerza:
This PR improves the CLI's user experience by introducing an interactive mode for the `create`, `dryrun`, `deploy`, and `init` commands.
​
 When a user runs one of these commands without providing all the required arguments, the CLI will now launch an interactive prompt to guide them through the missing options. This makes the CLI easier to use, as it no longer requires users to memorize all command-line arguments.
​
 The interactive prompts are built using `inquirer` to provide a user-friendly selection list for known values (like function types or environments).
​
 The interactive mode is enabled by default and can be disabled with the `--no-interactive` flag or by setting the `CI` environment variable, ensuring backward compatibility and scriptability.
​
 Fixes NAN-4237
​
Testing:
 1. Run `npm install` to add the new `inquirer` dependency.
 2. Run `nango create` without any arguments. Verify that you are prompted to select a function type, integration, and provide a name.
 3. Run `nango dryrun` without any arguments. Verify that you are prompted to select an environment, a function, and a connection.
 4. Run `nango deploy` without an environment. Verify that you are prompted to select one.
 5. Run `nango init` without a path. Verify that you are prompted to enter one.
 6. Run any of the above commands with the `--no-interactive` flag to confirm that the prompts are skipped and the command fails due to missing arguments.

---

A shared Ensure utility now orchestrates the interactive prompts, translating missing-argument errors into actionable guidance while preserving non-interactive execution when required.

<details>

• packages/cli/lib/index.ts
• packages/cli/lib/services/interactive.service.ts
• packages/cli/lib/services/ensure.service.ts
• packages/cli/lib/services/function-create.service.ts
• packages/cli/lib/types.ts
• packages/cli/lib/services/sdk.ts
• packages/cli/lib/utils/errors.ts
• packages/cli/tsconfig.json
• packages/cli/package.json
• docs/reference/cli.mdx
• package-lock.json

</details>

---


packages/cli/lib/services/interactive.service.ts
```diff
@@ -0,0 +1,183 @@
...
+            }
+        } else {
+            const parsing = parse(fullPath, debug);
+            if (parsing.isOk()) {
+                integrations = parsing.value.parsed?.integrations.map((i) => i.providerConfigKey) || [];
+            } else {
+                console.error(chalk.red(parsing.error));
+            }
+        }
+    }
```
TBonnin: would it make sense to pass the integrations as argument and move the the parsing/loading outside of the interactive logic? Same for fetching the connections below?

packages/cli/lib/services/interactive.service.ts
```diff
@@ -0,0 +1,183 @@
...
+    return name;
+}
+
+export async function promptForEnvironment(): Promise<string> {
+    const { env } = await inquirer.prompt([
+        {
+            type: 'rawlist',
+            name: 'env',
+            message: 'Which environment do you want to use?',
+            choices: ['dev', 'prod', new inquirer.Separator(), OTHER_CHOICE]
```
TBonnin, line 77: why not fetching the environments from the API?

packages/cli/lib/index.ts
```diff
@@ -35,23 +44,32 @@ import { dev } from './zeroYaml/dev.js';
...
+import type { DeployOptions, FunctionType, GlobalOptions } from './types.js';
 import type { NangoYamlParsed } from '@nangohq/types';
 
 class NangoCommand extends Command {
     override createCommand(name: string) {
         const cmd = new Command(name);
         cmd.option('--auto-confirm', 'Auto confirm yes to all prompts.', false);
         cmd.option('--debug', 'Run cli in debug mode, outputting verbose logs.', false);
+        // Default to true so that interactive mode is enabled by default
+        cmd.option('--no-interactive', 'Disable interactive prompts for missing arguments.', true);
```
TBonnin: not sure I understand why --no-interactive is true by default if we want the interactive mode enabled by default

packages/cli/lib/index.ts
```diff
@@ -112,9 +130,19 @@ program
     .option('--ai [claude|cursor...]', 'Optional: Setup AI agent instructions files. Supported: claude code, cursor', [])
     .option('--copy', 'Optional: Only copy files, will not npm install or pre-compile', false)
     .action(async function (this: Command) {
-        const { debug, ai, copy } = this.opts<GlobalOptions & { ai: string[]; copy: boolean }>();
+        const { debug, ai, copy, interactive } = this.opts<GlobalOptions & { ai: string[]; copy: boolean }>();
+        let [projectPath] = this.args;
```
TBonnin, line 136: where is projecPath being used?

packages/cli/lib/index.ts
```diff
@@ -150,16 +178,54 @@ program
...
+        } else {
+            if (sync) functionType = 'sync';
+            else if (action) functionType = 'action';
+            else if (onEvent) functionType = 'on-event';
+        }
+
+        if (!functionType || !integration || !name) {
+            console.error(chalk.red('Error: Missing required arguments. Use --sync, --action, or --on-event, and provide an integration and a name.'));
+            this.help();
+        }
```
TBonnin: I feel like extracting this logic in its own file/class with functions and early returns will reduce the branching, make it easier to read and more testable as well as reusable in the different commands. Something like that:

```
export class Ensure {
    constructor(private readonly interactive: boolean)

    private async ensure<T>(
      currentValue: T | undefined,
      promptFn: () => Promise<T>,
      errorMessage: string
  ): Promise<T> {
      if (currentValue) return currentValue;
      if (!interactive) throw new MissingArgumentError(errorMessage);
      try {
      return await promptFn();
      catch {
           // show error and process exit
      }
  }

  public async functionType(
      sync: boolean,
      action: boolean,
      onEvent: boolean,
  ): Promise<FunctionType> {
      if (sync) return 'sync';
      if (action) return 'action';
      if (onEvent) return 'on-event';

      if (!this.interactive) {
          throw new MissingArgumentError('Must specify --sync, --action, or --on-event');
      }

      return await promptForFunctionType();
  }

  public async integration(
      current: string | undefined,
      context: { fullPath?: string; isNangoFolder: boolean; isZeroYaml:
  boolean; debug?: boolean }
  ): Promise<string> {
      return ensure(
          current,
          () => promptForIntegrationName(context),
          'Integration name is required'
      );
  }

  public async functionName(
      current: string | undefined,
      functionType: FunctionType,
  ): Promise<string> {
      return ensure(
          current,
          () => promptForFunctionName(functionType),
          'Function name is required'
      );
  }

  export async function environment(
      current: string | undefined,
      interactive: boolean
  ): Promise<string> {
      return ensure(
          current,
          interactive,
          () => promptForEnvironment(),
          'Environment is required'
      );
  }

  public async function(
      current: string | undefined,
      availableFunctions: { name: string; type: string }[]
  ): Promise<string> {
      return ensure(
          current,
          () => promptForFunctionToRun(availableFunctions),
          'Function name is required'
      );
  }

  public async connection(
      current: string | undefined,
      environment: string
  ): Promise<string> {
      return ensure(
          current,
          () => promptForConnection(environment),
          'Connection ID is required'
      );
  }

  public async projectPath(
      current: string | undefined,
  ): Promise<string> {
      return ensure(
          current,
          () => promptForProjectPath(),
          'Project path is required'
      );
  }
```

so it can be used like this 
```
const ensure = new Ensure(interactive)
const functionType = await ensure.functionType(sync, action,
  onEvent);
integration = await ensure.integration(integration, { ... });
name = await ensure.functionName(name, functionType);
```

wdyt?

packages/cli/lib/index.ts
```diff
@@ -245,6 +314,37 @@ program
...
+                    } else {
+                        console.error(chalk.red('Could not build function definitions to select from.'));
+                        process.exit(1);
+                    }
+                }
+                if (!connectionId) {
+                    connectionId = await promptForConnection(environment);
+                }
+            } catch (err: any) {
+                console.error(chalk.red(err.isTtyError ? "Prompt couldn't be rendered in the current environment" : err.message));
```
TBonnin: can we add some info about how users could fix the issue?

packages/cli/package.json
```diff
@@ -61,8 +61,9 @@
         "figlet": "1.8.2",
         "glob": "11.1.0",
         "import-meta-resolve": "4.1.0",
-        "jscodeshift": "17.3.0",
+        "inquirer": "^13.1.0",
```
TBonnin: we always pin to a specific version. We have all collectively been burned too many time by breaking changes (even in patch version)

packages/cli/lib/services/interactive.service.ts
agusayerza: Checked the API and there is no endpoint to list environments yet, maybe I missed it? I talked with Bastien of this being a fast-follow-up

packages/cli/lib/services/interactive.service.ts
agusayerza: Done. I moved the parsing logic out and just pass the list to the prompt function now

packages/cli/lib/index.ts
agusayerza: `--no` prefix on commander sets the flag to false when you pass it instead of true. Added a better description to the comment to explain that weird behaviour

packages/cli/lib/index.ts
agusayerza: Good catch. It's being used now to resolve the absolute path, missed it

packages/cli/lib/index.ts
agusayerza: Great suggestion, implemented it

packages/cli/package.json
agusayerza: Good call, pinned it

packages/cli/lib/index.ts
agusayerza: Added a hint to use `--no-interactive` if the prompt fails (I would expect this to happen in CI/non-TTY).

packages/cli/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "nango",
-    "version": "0.69.20",
+    "version": "0.70.0",
```
agusayerza: Thoughts on if this should be a minor or a patch? Sticking to strict semver its a minor, but open to change it if anyone thinks its a patch

packages/cli/lib/services/interactive.service.ts
agusayerza: Just checked the code, endpoint exists but not documented or exposed on the SDK. My bad. I still think this is an improvement to be done as a fast-follow-up

packages/cli/lib/errors.ts
```diff
@@ -0,0 +1,6 @@
+export class MissingArgumentError extends Error {
```
marcindobry: There is also `packages/cli/lib/utils/errors.ts`. Can we unify them?

packages/cli/lib/index.ts
```diff
@@ -16,12 +16,14 @@ import { nangoConfigFile } from '@nangohq/nango-yaml';
 
 import { initAI } from './ai/init.js';
 import { generate, getVersionOutput, tscWatch } from './cli.js';
+import { MissingArgumentError } from './errors.js';
```
marcindobry: ooc: Why do we need to import `.js` files here?

packages/cli/lib/index.ts
```diff
@@ -43,15 +45,26 @@ class NangoCommand extends Command {
...
         cmd.hook('preAction', async function (this: Command, actionCommand: Command) {
-            const { debug } = actionCommand.opts<GlobalOptions>();
-            printDebug('Debug mode enabled', debug);
-            if (debug && fs.existsSync('.env')) {
-                printDebug('.env file detected and loaded', debug);
+            const opts = actionCommand.opts<GlobalOptions>();
+
+            // opts.interactive is true by default (from the option default), or false if --no-interactive is passed.
+            // We also disable it if we are in a CI environment.
+            opts.interactive = opts.interactive && !isCI;
```
marcindobry, line 65: When would this be used in a CI environment?

I'm not sure it's a great idea to have the completely opposite behavior of a flag in CI instead of outside of it. Would it make more sense to raise an error in CI environment and make sure people specify --no-interactive instead? This would be a breaking change, though, hence the question above :P

agusayerza: Some customers use the CLI today on CI envs to test and deploy nango. If we force them to adopt the `--no-interactive` flag we would break their CI pipelines. At the same time, we want interactive mode to be the default use case for people using the CLI on the terminal, so this gets both things done. A warning for CI envs that do not provide the `--no-interactive` flag might be in order though. Wdyt?

packages/cli/lib/errors.ts
agusayerza: Missed that file, will merge them

packages/cli/lib/index.ts
agusayerza: ESM modules require file extension on imports. Addtionally, my understanding is that typescript doesn't change file extensions when transpiling to javascript, so we need to reference the `js` files.

Might be wrong and there is another reason why this is done, but I think this is the case

packages/cli/lib/index.ts
marcindobry: Yup I think a warning would be great.

marcindobry (approved):

packages/cli/package.json
TBonnin: is it a breaking change? aka: does it modify the behavior of existing commands? if yes, minor otherwise patch

TBonnin: version isn't managed by modifying the package.json directly. The release workflow takes care of it

packages/cli/package.json
```diff
@@ -83,6 +84,7 @@
         "@types/commander": "2.12.5",
         "@types/ejs": "3.1.5",
         "@types/figlet": "1.5.6",
+        "@types/inquirer": "^9.0.7",
```
TBonnin: pin, pin, pin

packages/cli/lib/services/interactive.service.ts
TBonnin, line 1: does interactive and ensure belongs to the `services` folder? You are gonna ask what's a `service` in the context of the CLI and to be honest I am not sure I have an good answer :/

TBonnin (approved):
lgtm. Just the package version and dependency pinning and the rest is good to go I think

packages/cli/package.json
agusayerza: 🤦

packages/cli/lib/services/interactive.service.ts
agusayerza: I just followed suit, but it doesn't make any sense. Thought of it as just where the main logic leaves, but it is a bit a weird naming scheme if you ask me

TBonnin: you don't need to change anything here. Let's keep it in mind though when/if we refactor the CLI ;)

PR #5176, opened 2025-12-19
feat(webapp): redesign integrations page

kaposke:
> Deployed to development for testing (check if it's still there!) Server deployment is having problems, so if you're having bugs saving auth settings, checkout to the branch and test locally

Huge redesign of the integrations page. This includes:
- List of functions (now a table with tabs)
- Individual functions page
  - New way of rendering inputs and outputs
- Settings page (with all the coverage for every auth type)

Video walkthrough:
https://www.loom.com/share/d84c46fb8f6742a1b159d4688ffe5bf5

It's a lot of code so I expect to have a few very visible mistakes, or duplicated code here and there. Don't hesitate to point anything out.

---

The functions tab now groups entries with inline enable/disable confirmations and deep links into a dedicated detail view that renders expandable JSON schemas for inputs and outputs, while the settings tab uses reusable inline editors with per-field confirmation flows to update each authentication mode in place.

<details>

• packages/webapp/src/pages/Integrations/providerConfigKey
• packages/webapp/src/components-v2
• packages/server/lib/controllers/v1/integrations/providerConfigKey/patchIntegration.ts
• packages/server/lib/routes.private.ts
• packages/types/lib/integration/api.ts
• packages/webapp/src/index.css
• packages/webapp/package.json

</details>

---


packages/server/lib/controllers/v1/integrations/providerConfigKey/patchIntegration.ts
kaposke, line 1: Had to make things optional because we now do per-field saves. Not a single form for these anymore. Curious if there's any unforeseen consequence.

packages/webapp/src/pages/Integrations/providerConfigKey/Functions/Tab.tsx
```diff
@@ -0,0 +1,158 @@
...
+        </Navigation>
+    );
+};
+
+const GroupedFunctionsTable: React.FC<{
+    groupedFunctions: Record<string, NangoSyncConfig[]>;
+    onFunctionClick?: (func: NangoSyncConfig) => void;
+    integration: ApiIntegration;
+}> = ({ groupedFunctions, onFunctionClick, integration }) => {
+    return (
```
TBonnin, line 105: the hover and group header background color seems to be the same. Making it hard to differentiate when hovering with the mouse

Image

packages/webapp/src/pages/Integrations/components/jsonSchema/JsonSchema.tsx
```diff
@@ -0,0 +1,174 @@
...
+import type { JSONSchema7, JSONSchema7Type } from 'json-schema';
+
+export const JsonSchemaTopLevelObject: React.FC<{ schema: JSONSchema7 }> = ({ schema }) => {
+    if (!isObjectSchema(schema)) {
+        throw new Error('Expected object schema.');
+    }
+
+    return (
+        <div className="flex flex-col gap-1.5">
+            {Object.entries(schema.properties || {}).map(([name, property]) => (
```
TBonnin, line 19: should we only show the N first top level properties in case a schema has tons of them?

packages/server/lib/controllers/v1/integrations/providerConfigKey/patchIntegration.ts
```diff
@@ -147,22 +147,48 @@ export const patchIntegration = asyncWrapper<PatchIntegration>(async (req, res)
...
+            if (body.clientId !== undefined) {
+                integration.oauth_client_id = body.clientId;
+            }
+            if (body.clientSecret !== undefined) {
+                integration.oauth_client_secret = body.clientSecret;
+            }
+            if (body.appLink !== undefined) {
+                integration.app_link = body.appLink;
+            }
+            if (body.appId !== undefined || body.privateKey !== undefined) {
```
TBonnin: why the `if` combine with the spread operator below with the same condition

packages/server/lib/controllers/v1/integrations/providerConfigKey/patchIntegration.ts
TBonnin: can a partial but invalid state be processed and stored successfully?

packages/webapp/src/pages/Integrations/components/jsonSchema/JsonSchema.tsx
kaposke: Sure. I'm curious to see a case where it gets crazy enough though

kaposke: I'd say it looks Ok until level 7 (this is in it's minimum width). I'll cap it there.
image

packages/server/lib/controllers/v1/integrations/providerConfigKey/patchIntegration.ts
kaposke: Removed

packages/webapp/src/pages/Integrations/providerConfigKey/Functions/Tab.tsx
kaposke: good catch, it had the wrong hover color

packages/server/lib/controllers/v1/integrations/providerConfigKey/patchIntegration.ts
kaposke: Hmm not sure I got what you mean

packages/webapp/src/pages/Integrations/components/jsonSchema/JsonSchema.tsx
TBonnin: I'll cap it at 3-4. I feel like everything more complicated isn't going to be read in the browser anyway

marcindobry:
Food for thought:

If the action output only has 1 schema, would it make sense to just hide that schema name at the top? Otherwise it looks out-of-place and it's not clear what it means. If there were 2, then clicking between them would make it clear what they are.

image

marcindobry:
It's hard to show on video, but when I click the "Enabled" toggle on the actions page, turning on shows a spinner indicating it is loading, but turning off doesn't have any feedback that stuff is happening in the background and it the toggle does turn off after a moment, but it feels a bit more as if the UI was "frozen". Should we add a spinner to both transitions while the request is in-flight?

https://www.loom.com/share/934dfc1f835f448eb79460465504d45c

packages/server/lib/controllers/v1/integrations/providerConfigKey/patchIntegration.ts
marcindobry: Technically for a PATCH endpoint that's fine to only send partial updates.

marcindobry (approved):
Pretty sure I missed some things but hopefully nothing major 🙈

PR #5177, opened 2025-12-19
feat(webapp): new components & pre-requisites for integration page redesign

kaposke:
I'm extracting new components and updates to existing components from my integrations page redesign to this PR, in an attempt to make both PRs a bit more palatable. 

It's hard to review components without seeing their intended behavior, so I've made a quick loom to show them in action:
https://www.loom.com/share/d11c79a22f43420f99a05ec7c4b72d81
> It should be easier to review after watching this.

---

**Introduce reusable navigation/edit inputs and backend API status polling**

Large UI refactor introduces reusable navigation primitives (`Navigation`, `Tabs`), editable field components (`EditableInput`, `ScopesInput`), and confirmation dialog utilities used by environment settings and billing pages. A new `/api-status/:service` server endpoint plus `useApiStatus` hook polls API Down Watch using a server-side key, while shared form controls (`InputGroup`, `Input`, `Textarea`, `CopyButton`, `Dot`) gain filled-state styling and forwardRef support.

<details>

• Added `EditableInput`, `ScopesInput`, `ConfirmDialog`, and upgraded `Navigation`/`Tabs` components with controlled/uncontrolled support and horizontal styling options.
• Introduced hooks `useHashNavigation`, `usePathNavigation`, `useConfirmDialog`, and `useApiStatus` to synchronize UI state with URL and poll backend API status.
• Refactored environment settings and billing pages to rely on new navigation hooks/components; moved API Down Watch polling to new backend endpoint `/api-status/:service` with env parsing for `API_DOWN_WATCH_API_KEY`.
• Updated shared UI primitives (`InputGroup`, `InputGroupInput`, `InputGroupTextarea`, `Input`, `Textarea`, `Label`, `CopyButton`, `Dot`) for filled/error state styling and ref forwarding.

</details>

<details>

• `packages/webapp/src/components-v2`
• `packages/webapp/src/hooks`
• `packages/webapp/src/pages/Environment/Settings/Show.tsx`
• `packages/webapp/src/pages/Team/Billing`
• `packages/server/lib/controllers/v1/apiStatus/getApiStatus.ts`
• `packages/server/lib/routes.private.ts`
• `packages/types/lib`

</details>

---


packages/webapp/src/hooks/useConfirmDialog.tsx
```diff
@@ -0,0 +1,53 @@
+import { useCallback, useRef, useState } from 'react';
+
+import { ConfirmDialog } from '@/components-v2/ConfirmDialog';
+
+import type { ConfirmDialogOptions } from '@/components-v2/ConfirmDialog';
+
+export const useConfirmDialog = () => {
```
marcindobry, line 7: Can you show me an example of how this is used? I can't wrap my head around it yet

packages/webapp/src/components-v2/ConfirmDialog.tsx
```diff
@@ -0,0 +1,76 @@
...
+    title: string;
+    description: string;
+    confirmButtonText?: string;
+    cancelButtonText?: string;
+    confirmVariant?: 'primary' | 'destructive' | 'secondary' | 'tertiary';
+    icon?: React.ReactNode;
+    onConfirm: () => void | Promise<void>;
+}
+
+interface ConfirmDialogProps extends ConfirmDialogOptions {
```
marcindobry, line 16: Do they have to be separate for the hook to make sense?

packages/types/lib/web/env.ts
```diff
@@ -1,6 +1,7 @@
 export interface WindowEnv {
     apiUrl: string;
     apiDownWatchPublicKey: string;
+    apiDownWatchApiKey: string;
```
marcindobry: Does this mean we're going to be pushing an API key into the browser? Could people then abuse it and rack up a big bill for us (among other things)? No other thing here is a secret, they're all public keys :/

packages/webapp/src/components-v2/ui/input-group.tsx
```diff
@@ -23,11 +23,13 @@ function InputGroup({ className, ...props }: React.ComponentProps<'div'>) {
...
-                // Focus state.
-                'has-[[data-slot=input-group-control]:focus-visible]:shadow-focus has-[[data-slot=input-group-control]:focus-visible]:border-border-default',
+                // Focus state
+                'has-[[data-slot=input-group-control]:focus-visible]:focus-default has-[[data-slot=input-group-control]:focus-visible]:border-border-default',
+                // Filled state - different border when input has text (works for both controlled and uncontrolled inputs)
+                'has-[[data-slot=input-group-control][data-filled=true]:not(:disabled)]:border-border-strong',
 
                 // Error state.
-                'has-[[data-slot][aria-invalid=true]]:focus-error has-[[data-slot][aria-invalid=true]]:border-feedback-error-border',
+                'has-[[data-slot][aria-invalid=true]]:!focus-error has-[[data-slot][aria-invalid=true]]:!border-feedback-error-border',
```
marcindobry, line 32: Can we avoid having a `!` here or is it basically impossible?

marcindobry (commented):
Biggest concern is around pushing a secret key into the browser. I don't think we should do that and instead e.g. have a proxy endpoint through Nango.

packages/types/lib/web/env.ts
kaposke: Yeah I think you're right. Will migrate to an endpoint.

packages/webapp/src/components-v2/ui/input-group.tsx
kaposke: I think it would be avoidable if we also negated the aria-invalid for the regular focus state. This wouldn't work without the ! because of specificity I think. I might be able to find a cleaner solution with JS instead of these tailwind tricks.

packages/webapp/src/hooks/useConfirmDialog.tsx
kaposke: Yes. This is a bit hard to understand indeed, but usage is simple:
```tsx
function MyComponent() {
    const {confirm, DialogComponent} = useConfirmDialog();

    const openDialog = async () => {
        const confirmed = await confirm({
            title: 'Dialog Title',
            description: 'Dialog Description',
            onConfirm: () => {
                console.log('Confirmed');
            }
        });
    }

    return (
        <Button onClick={openDialog}>
            Open Dialog
        </Button>
        {DialogComponent}
    )
}
```

packages/webapp/src/components-v2/ConfirmDialog.tsx
kaposke: Not necessarily, just thought it fit better in a component file

packages/webapp/src/components-v2/ui/input-group.tsx
kaposke: Update: couldn't do it easily. I'd push this to the future if that's ok

packages/webapp/src/hooks/useConfirmDialog.tsx
marcindobry: I'm not 100% sold on the design but I'm sure you had a lot more to think about it. However, what do you think about at least making `useConfirmDialog` return a `DialogComponent` that's a function so that we can embed it as `<DialogComponent />` instead?

marcindobry (approved):

PR #5183, opened 2025-12-19
feat(cli): Unified mock format and migration support

agusayerza:
This PR introduces a unified structure for test mocks, consolidating scattered JSON files into a single `.test.json` file per test suite.

### Unified Mock Format
- Updated `nango dryrun --save` to generate the new unified mock format
- Implemented `ResponseCollector` to aggregate API calls during dry runs
- Updated `NangoActionMock` and `NangoSyncMock` in `nango/test` to support the new format

### Migration Path
- Running `MIGRATE_MOCKS=2026-01 npm test` records legacy mock usage and generates the new unified file automatically
- Migration captures **all** mock files for each endpoint (not just those accessed during the test run)
- Handles both hash-based directory format and name-based file format

### Pagination Bug Fix
The legacy test utilities had a bug where pagination would stop after the first page, causing:
- Tests to falsely pass (only testing page 1)
- Mock files with incorrect request hashes
- Missing mock files for subsequent pages

**Fixes implemented:**
- **Fuzzy matching**: When exact hash lookup fails, matches by comparing `requestIdentity.params` directly
- **Best-effort recovery**: Migration recovers mocks where possible, with clear errors when not

### Documentation
- Updated testing docs to explain the pagination bug and migration behavior
- Clear error messages guide users to re-record with `nango dryrun <sync> <connection> --save`

---

**Note:** Follow-up needed in integration-templates repo to update `vitest.setup.ts` to import from `nango/test` instead of using local mock implementations.

[NAN-4463](https://linear.app/nango/issue/NAN-4237/cli-interactivity-and-cleanup)

## Testing Instructions

1. Run `nango dryrun <name> <connectionId> --save` and verify a single `.test.json` is created
2. Run `npm test` to verify tests pass with the new format
3. (Optional) In a project with legacy mocks, run `MIGRATE_MOCKS=2026-01 npm test` and verify the new mock file is generated
4. (Optional) Test pagination: verify paginated endpoints capture all pages in the migrated file

---

The unified fixtures capture complete request and response metadata—including connection context and proxied traffic—so the CLI helpers can replay dry runs, manage legacy-to-unified migration inline, and actively close the long-standing pagination gap.

<details>

• `packages/cli/lib/services/response-collector.service.ts`
• `packages/cli/lib/services/dryrun.service.ts`
• `packages/cli/lib/testMocks/utils.ts`
• `docs/implementation-guides/building-integrations/testing.mdx`
• `packages/cli/package.json`

</details>

---


packages/cli/lib/services/response-collector.service.ts
```diff
@@ -0,0 +1,306 @@
...
+
+            const endpointKey = `/${endpoint}`;
+            const existingEntry = mockData.api[method][endpointKey];
+
+            const params = Object.fromEntries(call.requestIdentity.params);
+            const headers = Object.fromEntries(call.requestIdentity.headers);
+
+            let requestData = call.requestIdentity.data;
+            if (typeof requestData === 'string') {
+                requestData = JSON.parse(requestData);
```
TBonnin: can the parsing throw?

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -17,7 +16,7 @@ import { ActionError, BASE_VARIANT, InvalidActionInputSDKError, InvalidActionOut
 import { parse } from './config.service.js';
 import { DiagnosticsMonitor, formatDiagnostics } from './diagnostics-monitor.service.js';
 import { loadSchemaJson } from './model.service.js';
-import * as responseSaver from './response-saver.service.js';
```
TBonnin, line 20: can the 'response-saver.service.ts` file be deleted?

packages/cli/lib/services/response-collector.service.ts
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    private ensureDirectoryExists(directoryName: string): void {
+        if (!fs.existsSync(directoryName)) {
+            fs.mkdirSync(directoryName, { recursive: true });
+        }
+    }
+
+    private computeRequestIdentity(config: AxiosRequestConfig): { requestIdentity: RequestIdentity; requestIdentityHash: string } {
+        const method = config.method?.toUpperCase() || 'GET';
```
TBonnin: previous logic was using lowercase. Isn't using uppercase changing the identity hash and making it incompatible?

agusayerza: You are right, although it wouldn't really matter as this is for new test generations, and when parsing the request at test runtime, we still LowerCase it. Regardless, I changed it back to use lowercase as there is no reason to modify our behaviour.

packages/cli/lib/services/response-collector.service.ts
agusayerza: It might, if the request isn't json. I have wrapped it with a try/catch

packages/cli/lib/services/dryrun.service.ts
agusayerza: Done

packages/cli/lib/services/response-collector.service.ts
```diff
@@ -0,0 +1,315 @@
...
+export class ResponseCollector {
+    private apiCalls: CachedRequest[] = [];
+    private connectionInfo?: { connectionId: string; provider: string };
+    private metadata?: Metadata;
+    private batchSaveData: Record<string, unknown[]> = {};
+    private batchDeleteData: Record<string, unknown[]> = {};
+
+    public onAxiosRequestFulfilled(response: AxiosResponse, connectionId: string): AxiosResponse {
+        // Handle getConnection/getMetadata calls
+        if (response.request.path.includes(`/connections/${connectionId}`)) {
```
TBonnin, line 77: the string doesn't match the comment.

packages/cli/lib/services/response-collector.service.ts
```diff
@@ -0,0 +1,315 @@
...
+
+            const params = Object.fromEntries(call.requestIdentity.params);
+            const headers = Object.fromEntries(call.requestIdentity.headers);
+
+            let requestData = call.requestIdentity.data;
+            if (typeof requestData === 'string') {
+                try {
+                    requestData = JSON.parse(requestData);
+                } catch {
+                    // ignore
```
TBonnin, line 183: should we output a warning here?

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -435,95 +418,46 @@ export class DryRunService {
...
-                    if (confirmation) {
-                        displayBatch();
-                    } else {
-                        break;
-                    }
+            const nangoInstance = results.response?.nango;
+            if (nangoInstance instanceof NangoSyncCLI) {
+                const logMessages = nangoInstance.logMessages;
+                if (logMessages && logMessages.messages.length > 0) {
+                    // ... (rest of the logging logic)
```
TBonnin, line 436: is that a TODO?

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -435,95 +418,46 @@ export class DryRunService {
...
+            if (options.saveResponses) {
+                const testFilePath = `${this.fullPath}/${providerConfigKey}/tests/${syncName}.test.json`;
+                responseCollector.saveUnifiedMock({
+                    filePath: testFilePath,
+                    input: normalizedInput,
+                    output: results.response?.output,
+                    stubbedMetadata: stubbedMetadata
+                });
+                console.log(chalk.green(`\n✅ Mocks saved to ${testFilePath}`));
+            }
```
TBonnin: the response is now saved at the end compared to at the beginning before. could this change of behavior be problematic and cause partial saving.

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -7,7 +7,6 @@ import * as url from 'url';
 
 import { AxiosError } from 'axios';
 import chalk from 'chalk';
-import promptly from 'promptly';
```
TBonnin, line 10: is promptly used anywhere else? if not let's remove it from package.json

packages/cli/lib/testMocks/utils.ts
```diff
@@ -2,49 +2,457 @@ import * as crypto from 'crypto';
...
+
+                if (matchedMock) {
+                    return matchedMock;
+                }
+            } else {
+                return apiMock;
+            }
+        }
+
+        throw new Error(`No mock found for ${method.toUpperCase()} ${endpoint} (normalized: ${normalizedEndpoint}) with hash ${requestIdentityHash}`);
```
TBonnin: looks like we can simplify this piece of code with some early return and utility function.
```
const apiMock = this.mockData.api?.[method.toUpperCase()]?.[normalizedEndpoint] 
            || this.mockData.api?.[method.toUpperCase()]?.[`/${normalizedEndpoint}`];
if (!apiMock) { throw ... } ;

if (!Array.isArray(apiMock)) return apiMock;

const getFieldsMatch = (expected, actual, opts: { caseSensitive: bool } = {caseSensitive: true}) {
  return Object.entries(expected).every(([key, value]) => {
    const actualField = actual.find(([k]) => 
      opts.caseSensitive ? k.toLowerCase() === key.toLowerCase() : k === key
    );
    return actualField && String(actualField[1]) === String(value);
  });
}

// algo: hash > params > headers > data
const matchedMock = apiMock.find((mock) => {

  if (mock.hash === requestIdentityHash) return true;
  
  if (!mock.request) return false;
  
  if (mock.request.params && 
      !fieldsMatch(mock.request.params, identity.requestIdentity.params, { caseSensitive: false})) {
    return false;
  }
  
  if (mock.request.headers && 
      !fieldsMatch(mock.request.headers, identity.requestIdentity.headers)) {
    return false;
  }
  
  if (mock.request.data !== undefined) {
    const expectedDataIdentity = computeDataIdentity({ data: mock.request.data });
    if (expectedDataIdentity !== identity.requestIdentity.data) {
      return false;
    }
  }
  
  return true;
});

return matchedMock;
```

docs/implementation-guides/building-integrations/testing.mdx
```diff
@@ -285,18 +314,36 @@ describe('github create-ticket tests', () => {
...
 - Tests work **offline**
 
+### Migrating from the old format
+
+If you have tests using the old multi-file mock format, you can automatically migrate them to the new unified format.
+
+Set the `MIGRATE_MOCKS` environment variable to `true` and run your tests:
+
+```bash
+MIGRATE_MOCKS=true npm test
```
TBonnin: what about something like `MOCKS_MIGRATION=2026-01` or similar. In case we need to migrate again in the future :)

packages/cli/lib/testMocks/utils.ts
```diff
@@ -209,55 +543,210 @@ class NangoActionMock implements Omit<NangoActionBase, 'nango'> {
...
+            }
+
+            if (offsetCalculationMethod === 'per-page') {
+                offset++;
+            } else {
+                offset += responseData.length;
+            }
+        }
+    }
+
```
TBonnin, line 1002: why cannot we re-used the pagination logic?

packages/cli/lib/testMocks/utils.ts
```diff
@@ -209,55 +543,210 @@ class NangoActionMock implements Omit<NangoActionBase, 'nango'> {
...
+                args.endpoint = url.pathname + url.search;
+            }
+
+            args.params = {};
+        }
+    }
+
+    private getNextPageLinkFromBodyOrHeaders(linkPagination: LinkPagination, response: MockResponse, paginationConfig: Pagination) {
+        if (linkPagination.link_rel_in_response_header) {
+            const linkHeader = parseLinksHeader(response.headers['link']);
```
TBonnin, line 933: do we need an extra dependency for parsing the link header?

docs/implementation-guides/building-integrations/testing.mdx
```diff
@@ -285,18 +314,36 @@ describe('github create-ticket tests', () => {
...
+```
+
+This will:
+1. Run your tests using the old mock files.
+2. Intercept all mock data accessed during the test run.
+3. Save the data into a new `.test.json` file.
+4. The old mock directory can then be safely deleted.
```
TBonnin: can we name the old mock directory to make it super clear what can be deleted?

TBonnin (commented):
have you tested the performance with a huge test.json file?

docs/implementation-guides/building-integrations/testing.mdx
agusayerza: Done

packages/cli/lib/testMocks/utils.ts
agusayerza: Its already being used in the runners, I think it makes sense to use the same dependency to keep it consistent. This bugged pagination logic is slated to be removed once we migrate all tests on `integration-templates`

docs/implementation-guides/building-integrations/testing.mdx
agusayerza: Added, much better

packages/cli/lib/testMocks/utils.ts
agusayerza: Done

packages/cli/lib/services/dryrun.service.ts
agusayerza: It is still being used

packages/cli/lib/services/dryrun.service.ts
agusayerza: AI doing AI stuff. Good catch

packages/cli/lib/services/response-collector.service.ts
agusayerza: Added a warning, this shouldn't happen but better to be safe than sorry

agusayerza:
> have you tested the performance with a huge test.json file?

Yes, actually I tested it on the whole `integration-templates` repo. I don't think there are HUGE files, but its the most real use case scenario possible. It wasn't a timed test, but for the most part runs last roughly the same.

TBonnin (approved):
Thank you @agusayerza

PR #5188, opened 2026-01-05
fix(proxy): Proxy JSON validation from remote

thw-nango:
- On failure, emit HTTP status 502 (Bad Gateway) instead of 500.
 - In error message, mention that this is in parsing upstream data.
 - Work for non-json payloads.
 - Avoid logging SyntaxError to stderr in tests.

---

The update refactors the response handler to separate JSON validation from delivery, adjusts metrics and logging so failures are recorded before returning, and guards against downstream send errors being misclassified as upstream parse issues.

<details>

• `packages/server/lib/controllers/proxy/allProxy.ts`
• `packages/server/lib/controllers/proxy/allProxy.unit.test.ts`

</details>

---

### DIFF
--- a/packages/server/lib/controllers/proxy/allProxy.ts
+++ b/packages/server/lib/controllers/proxy/allProxy.ts
@@ -345,29 +345,34 @@ export async function handleResponse({ res, responseStream, logCtx }: { res: Res
             return;
         }
 
-        try {
-            if (isJsonResponse) {
-                // Validate JSON structure without re-serializing to avoid JSON.parse limitations (ex: precision loss with big integers).
-                // TODO: consider removing validation and forwarding upstream response as-is (even if invalid JSON) to avoid performance overhead.
-                try {
-                    JSON.parse(Buffer.concat(responseData).toString());
-                    res.setHeader('Content-Type', 'application/json');
-                } catch (err) {
-                    res.writeHead(502, { 'Content-Type': 'application/json' });
-                    res.end(JSON.stringify({ error: 'Failed to parse JSON response from upstream service' }));
-                    void logCtx.error('Failed to parse JSON response from upstream service', { error: err });
-                    await logCtx.failed();
-                    metrics.increment(metrics.Types.PROXY_FAILURE);
-                    return;
-                }
+        if (isJsonResponse) {
+            // Validate JSON structure without re-serializing to avoid JSON.parse limitations (ex: precision loss with big integers).
+            // TODO: consider removing validation and forwarding upstream response as-is (even if invalid JSON) to avoid performance overhead.
+            try {
+                JSON.parse(Buffer.concat(responseData).toString());
+            } catch (err) {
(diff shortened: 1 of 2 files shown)


TBonnin (commented):
A more meaningful improvement would be not to parse the json. It is a change of behavior (that's why I didn't remove it when I touched this code in December) but I don't really see why our proxy logic should validate the json instead of simply passing it through. I'll vote for biting the bullet and remove the json parsing/validation

thw-nango:
I agree strongly. How about we merge this PR as a rollback target, in case we break someone by removing the JSON validation?

This PR fixes a spurious error log to stderr in tests, as well as wrongly reporting exceptions in `res.send`, `metrics.increment` or `logCtx.success` as JSON parsing errors.

TBonnin:
> I agree strongly. How about we merge this PR as a rollback target, in case we break someone by removing the JSON validation?
> 
> This PR fixes a spurious error log to stderr in tests, as well as wrongly reporting exceptions in `res.send`, `metrics.increment` or `logCtx.success` as JSON parsing errors.

I am fine with merging this one. Just one comment about not catching `res.send` though. It is unlikely to throw but could in theory

thw-nango:
Great catch! PTAL at my latest commit. I catch res.send failures, log them and make sure to only increment our outgoing size and success metrics on delivery.

packages/server/lib/controllers/proxy/allProxy.ts
```diff
@@ -345,28 +345,34 @@ export async function handleResponse({ res, responseStream, logCtx }: { res: Res
...
             await logCtx.failed();
             metrics.increment(metrics.Types.PROXY_FAILURE);
-        } finally {
-            metrics.increment(metrics.Types.PROXY_OUTGOING_PAYLOAD_SIZE_BYTES, responseLen, { accountId: logCtx.accountId });
+            return;
         }
+
+        await logCtx.success();
+        metrics.increment(metrics.Types.PROXY_SUCCESS);
+        metrics.increment(metrics.Types.PROXY_OUTGOING_PAYLOAD_SIZE_BYTES, responseLen, { accountId: logCtx.accountId });
```
TBonnin, line 375: `PROXY_OUTGOING_PAYLOAD_SIZE_BYTES` won't be sent anymore when errors are caught

thw-nango: Yes. The rationale is that if we fail validating `responseData`, the outgoing payload will be an error (of a different, smaller size). If we fail `send`ing the payload, we don't know how many bytes made it to the client, probably zero. Whether this is correct depends on how the metric is defined. What do you think?

TBonnin: You are right. That's fine, the metric is only used for monitoring purpose so we can revisit its definition and only account for successful send

TBonnin (approved):

PR #5198, opened 2026-01-06
feat(persist): add records auto-pruning

TBonnin:
We want to start removing records data and finally enforce the 'cache' nature of the records data store (aka: it is not a customer-facing database but a temporary location where records are saved before the customers replicate them into their own system). 
It will help with data privacy compliance and accountability as well as scalability and storage. 
The first step is to prune the record (ie: empty the payload) when a record isn't being updated for 30 days. Pruning a record allows Nango features like changes detection to work while not storing any customer data anymore for stale records (other than the record id). Customers would have 30 days after a records is updated by a sync to replicate it.

This PR is introducing a background daemon that prunes the payload of records not updated in that last 30 days.
The daemon runs in dryrun mode for now, not pruning any records yet until after the grace period communicated to customers.
The daemon runs in a tight loop. For each iteration a candidate connection/model is being selected and up to 1000 records for it are pruned. Each loop processing is kept small (max: 1 connection / 1000 records, random partition) in order to have a minimal impact on the db.
I am not sure yet that the tick interval and limit values would be enough to keep up (or if they are too aggressive). They might need to be tweaked once running in production.

 

<details>

• `packages/persist` background daemons
• `packages/records` service (`deleteRecords`, new `autoPruningCandidate`)
• `packages/utils` utilities (`cancellableDaemon`, metrics enum, env parsing)`
• Integration tests for records deletion/pruning

</details>

---

---

Supporting updates add a shared cancellable daemon utility with environment-configurable intervals and limits, extend deleteRecords’ dry-run handling, and broaden integration coverage to validate dry-run execution and candidate selection.

---

### DIFF
--- a/packages/persist/lib/daemons/autopruning.daemon.ts
+++ b/packages/persist/lib/daemons/autopruning.daemon.ts
@@ -11,11 +11,11 @@ import { logger } from '../logger.js';
  * Auto-pruning daemon
  * This daemon runs in a loop and prunes old records from the database based on the defined stale period
  * each tick, a candidate connection/model is selected for pruning
- * If a candidate is found, records older than the stale period are deleted
- * Only a limited number of records are deleted per tick to avoid overwhelming the database
+ * If a candidate is found, records older than the stale period are pruned
+ * Only a limited number of records are pruned per tick to avoid overwhelming the database
  * We rely on the daemon to run periodically on each persist instance to continue pruning old records
  */
-export function autoPruningDaemon(): { abort: () => Promise<void> } {
+export function autoPruningDaemon(): Awaited<ReturnType<typeof cancellableDaemon>> {
     return cancellableDaemon({
         tickIntervalMs: envs.PERSIST_AUTO_PRUNING_INTERVAL_MS,
         tick: async (): Promise<void> => {
@@ -51,23 +51,25 @@ export function autoPruningDaemon(): { abort: () => Promise<void> } {
                 });
                 if (res.isErr()) {
                     span.addTags({ error: res.error.message }).finish();
-                    logger.error(`[Auto-pruning] error deleting records: ${res.error.message}`);
+                    logger.error(`[Auto-pruning] error pruning records: ${res.error.message}`);
                     return;
                 }
                 // lag: how far behind are we from the desired pruning point
                 // high lag means we are not keeping up with pruning
                 let lagMs: number | null = null;
-                const cursorSort = Cursor.from(candidate.value.cursor)?.sort;
-                if (cursorSort) {
-                    const cursorDate = new Date(cursorSort);
-                    lagMs = Date.now() - cursorDate.getTime() - envs.PERSIST_AUTO_PRUNING_STALE_AFTER_MS;
+                if (res.value.lastCursor) {
+                    const cursorSort = Cursor.from(res.value.lastCursor)?.sort;
+                    if (cursorSort) {
+                        const cursorDate = new Date(cursorSort);
+                        lagMs = Date.now() - cursorDate.getTime() - envs.PERSIST_AUTO_PRUNING_STALE_AFTER_MS;
+                    }
                 }
                 span.addTags({
-                    deleted: res.value.count,
+                    pruned: res.value.count,
                     ...(lagMs ? { lagMs } : {})
                 }).finish();
             }
-            span.addTags({ deleted: 0, candidate: 'none_found' }).finish();
+            span.addTags({ pruned: 0, candidate: 'none_found' }).finish();
         }
     });
 }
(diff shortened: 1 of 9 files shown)


packages/utils/lib/daemon.ts
```diff
@@ -0,0 +1,27 @@
...
+    };
+    const done = loop();
+
+    const abort = async () => {
+        ac.abort();
+        return await done;
+    };
+
+    return { abort };
+}
```
TBonnin, line 27: Made it a helper function I will reuse in the next PR that will be hard deleting records for syncs not run in the last 6 months

packages/utils/lib/environment/parse.ts
```diff
@@ -54,6 +54,12 @@ export const ENVS = z.object({
 
     // Persist
     PERSIST_SERVICE_URL: z.url().optional(),
+    PERSIST_AUTO_PRUNING_INTERVAL_MS: z.coerce.number().optional().default(5_000), // set to 0 to disable
+    PERSIST_AUTO_PRUNING_LIMIT: z.coerce.number().optional().default(1_000),
```
TBonnin, line 58: as mentioned in the PR message, not sure yet if those values are too or not aggressive enough. Will tweak once in prod. Not that each instance of persist (8 as I write this message) will be running the daemon. Each of them will select a random partition and potentially prune records

packages/records/lib/models/records.ts
```diff
@@ -829,25 +842,32 @@ export async function deleteRecords({
                 if (lastDeletedRecord) {
                     lastCursor = Cursor.new({ id: lastDeletedRecord.id, last_modified_at: lastDeletedRecord.updated_at });
                 }
+
+                // break if the returned page is not full
+                if (paginatedRecords < toDelete) {
+                    break;
+                }
```
TBonnin, line 849: early break was not necessary because the records where updated and the query where conditions would ensure same records are not processed again but we need it for dryRun mode where records are not udpated. Without it the while loop would never exit, querying the same records for ever

packages/persist/lib/daemons/autopruning.daemon.ts
```diff
@@ -0,0 +1,73 @@
...
+                    return;
+                }
+                // lag: how far behind are we from the desired pruning point
+                // high lag means we are not keeping up with pruning
+                let lagMs: number | null = null;
+                const cursorSort = Cursor.from(candidate.value.cursor)?.sort;
+                if (cursorSort) {
+                    const cursorDate = new Date(cursorSort);
+                    lagMs = Date.now() - cursorDate.getTime() - envs.PERSIST_AUTO_PRUNING_STALE_AFTER_MS;
+                }
```
TBonnin, line 66: once we caught up with historic data the lag should be quite small. If not it would indicate the daemon isn't able to keep up
React

packages/persist/lib/daemons/autopruning.daemon.ts
```diff
@@ -0,0 +1,73 @@
...
+                span.addTags({
+                    deleted: res.value.count,
+                    ...(lagMs ? { lagMs } : {})
+                }).finish();
```
kaposke, line 70: Smart to track this ✨

packages/records/lib/models/records.ts
```diff
@@ -1081,6 +1101,60 @@ export async function deleteCount(
...
+        cursor: string;
+    } | null>
+> {
+    // Pick a random partition
+    const partition = Math.floor(Math.random() * 256);
+
+    try {
+        // Find a record that is stale in that partition
+        const [candidate] = await db
+            .from(`${RECORDS_TABLE}_p${partition}`)
```
kaposke, line 1128: Not familiar with how this is setup, are there 256 fixed partitions?

packages/utils/lib/telemetry/metrics.ts
```diff
@@ -96,7 +96,9 @@ export enum Types {
 
     USAGE_IS_CAPPED = 'nango.capping.isCapped',
 
-    PUBSUB_PUBLISH = 'nango.pubsub.publish'
+    PUBSUB_PUBLISH = 'nango.pubsub.publish',
+
+    PERSIST_RECORDS_AUTO_PRUNING = 'nango.persist.daemon.autopruning'
```
kaposke: Don't think you used this

packages/utils/lib/daemon.ts
```diff
@@ -0,0 +1,27 @@
+import { setTimeout } from 'node:timers/promises';
+
+export function cancellableDaemon<T>({ tick, tickIntervalMs }: { tick: () => Promise<T>; tickIntervalMs: number }): {
```
kaposke, line 3: Clever

kaposke (approved):

packages/utils/lib/telemetry/metrics.ts
TBonnin: indeed.  good catch

packages/records/lib/models/records.ts
TBonnin: yep https://github.com/NangoHQ/nango/blob/e972926e7e88fa344b13a8f0247b4c3c837765c8/packages/records/lib/db/migrations/20240327151027_create_records_table.ts#L4

PR #5205, opened 2026-01-07
fix(auth): correct token refresh for providers using providerClient

hassan254-prog:
- Fix token refresh for providers using `providerClient` with `config_overrides`.

---

The refresh flow now temporarily applies the override client credentials before invoking the provider client and reapplies those overrides to the returned OAuth2 credentials so downstream consumers continue receiving the correct metadata.

<details>

• Injects override `client_id` and `client_secret` into `providerConfig` prior to invoking `providerClient.refreshToken`.
• Reattaches the override `client_id` and `client_secret` onto the refreshed `OAuth2Credentials.config_override` payload.

</details>

<details>

• packages/shared/lib/services/connection.service.ts

</details>

---

### DIFF
--- a/packages/shared/lib/services/connection.service.ts
+++ b/packages/shared/lib/services/connection.service.ts
@@ -1415,7 +1415,7 @@ class ConnectionService {
     > {
         if (providerClient.shouldUseProviderClient(providerConfig.provider)) {
             const credentials = connection.credentials as OAuth2Credentials;
-            if (credentials.config_override && credentials.config_override.client_id && credentials.config_override.client_secret) {
+            if (credentials.config_override?.client_id && credentials.config_override?.client_secret) {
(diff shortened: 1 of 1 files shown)


packages/shared/lib/services/connection.service.ts
```diff
@@ -1414,9 +1414,24 @@ class ConnectionService {
         >
     > {
         if (providerClient.shouldUseProviderClient(providerConfig.provider)) {
+            const credentials = connection.credentials as OAuth2Credentials;
+            if (credentials.config_override && credentials.config_override.client_id && credentials.config_override.client_secret) {
```
TBonnin: you can use optional chaining to skip the first conditional check

TBonnin (approved):

PR #5217, opened 2026-01-09
feat(docs): add docs for records retention policies

TBonnin:
- Adding details in `docs/guides/platform/security.mdx` about the recently introduced records data retention policies. 
- mentioned retention policies and link to new doc in various guides (implement a sync, webhook, listRecords, ...)

---

The update introduces a dedicated “Synced records retention” subsection in the security guide that spells out the 30-day payload pruning and 60-day hard-deletion timelines, their impact on Sync functions, recommended handling practices, and escalation options for longer retention windows, while propagating the same retention notice across the syncs overview, implementation guide, webhooks guide, and the GET /records reference to underline the resulting data availability constraints.

---


docs/guides/platform/retention-policies.mdx
```diff
@@ -0,0 +1,104 @@
...
+### 1. Automated payload pruning (30 days)
+
+If a record has not been updated for **30 days**, Nango will automatically empty its payload.
+
+**Impact:**
+After pruning, the record's data fields are no longer retrievable from the Nango cache via the API or SDK. However, Nango can still track the record and detect future changes. This pruning does not impact sync execution or delta-detection capabilities.
+
+### 2. Automated hard deletion (60 days)
+
+If a sync has not executed for **60 days**, all records belonging to that sync will be permanently deleted.
```
agusayerza: nit: This will also remove the hashes so it will break delta-detection, right? Should we be more clear about that? I understand that this is what we are saying on the impact section, but maybe we should use the same wording as above?

agusayerza (approved):
LGTM, stamping so we can merge if desired, although my understanding is that we are waiting on Bastien

docs/guides/platform/retention-policies.mdx
```diff
@@ -0,0 +1,104 @@
...
+title: 'Data Retention Policies'
+sidebarTitle: 'Data Retention'
+description: 'Understand how Nango manages data retention'
+---
+
+Nango automatically manages data retention to ensure that sensitive user data is not stored indefinitely when it's no longer needed, helping you meet modern data compliance and security expectations.
+
+# Synced records
+
+Data retention policies for records stored in the Nango records cache.
```
bastienbeurier: IMO we should add a tiny bit more context here, because some people might come here when they're only exploring Managed auth. 

We could add: "This only applies if you are using Functions of type Sync, e.g. for replicating data or receive trigger notifications from an external API."

docs/guides/platform/retention-policies.mdx
```diff
@@ -0,0 +1,104 @@
+---
+title: 'Data Retention Policies'
+sidebarTitle: 'Data Retention'
+description: 'Understand how Nango manages data retention'
+---
+
+Nango automatically manages data retention to ensure that sensitive user data is not stored indefinitely when it's no longer needed, helping you meet modern data compliance and security expectations.
+
```
bastienbeurier: We already have a "Data retention" section in the "Securirty" page: https://nango.dev/docs/guides/platform/security#retention-after-deletion

IMO we should add this information there, alongside the data retention information regarding connections & API credentials.

bastienbeurier (changes requested):

docs/guides/platform/retention-policies.mdx
TBonnin: I would do it the other way around and move the connection retention info into this page and reference it from the security page, consolidating everything related to data retention in one dedicated place since it is becoming a important topic that goes beyond security. wdyt?

docs/guides/platform/retention-policies.mdx
TBonnin: Good point. The assumption is the sync isn't running anymore so there is not much to break. but yes let me clarify hard deletion impacts Nango ability to detect changes

docs/guides/platform/retention-policies.mdx
bastienbeurier: I'm doing a rework of the docs to simplify our navigation and reduce the number of top-level pages, so ideally I'd prefer keeping this under "Security" (even if I agree it goes beyond). If we realize it's hurting discoverability too much, we can easily separate it again if that works.

TBonnin: Moved everything to security.mdx. @bastienbeurier Feel free to tweak the content and merge when you think it is ready

bastienbeurier (approved):

PR #5250, opened 2026-01-16
feat(cli): fetch environments dynamically from server

agusayerza:
Description: This PR updates the CLI to dynamically fetch available environments from the server when prompting the user, instead of relying solely on hardcoded defaults.
Changes:
- Added GET /api/v1/environments endpoint to the server to list environments for the current account.
- Updated CLI utils.ts to include getEnvironments function.
- Updated interactive.service.ts to use getEnvironments to populate the environment selection prompt.
- Updated ensure.service.ts to integrate with the updated prompt logic.

Another alternative would have been to use the GetMeta endpoint to get the environments, but decided to add an endpoint specifically to get them instead.

[NAN-4556](https://linear.app/nango/issue/NAN-4556/get-envs-from-api-when-doing-a-deploy-from-cli)

---

The CLI now logs warnings and falls back to the default environment list when the API request fails or returns no environments, and a debug flag surfaces the request diagnostics during the interactive prompts.

---

### DIFF
--- a/packages/cli/lib/index.ts
+++ b/packages/cli/lib/index.ts
@@ -305,7 +305,7 @@ program
 
         try {
             const ensure = new Ensure(interactive);
-            environment = await ensure.environment(environment);
+            environment = await ensure.environment(environment, debug);
 
             const definitions = await buildDefinitions({ fullPath, debug });
             if (definitions.isOk()) {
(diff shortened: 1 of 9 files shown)


kaposke (approved):
I see it's using a `NANGO_SECRET_KEY` env var to authorize. Is that a recent addition? Is that just any environment key?

agusayerza:
> I see it's using a `NANGO_SECRET_KEY` env var to authorize. Is that a recent addition? Is that just any environment key?

Its just any env key. This is also used by other commands such as `dryrun`

PR #5259, opened 2026-01-19
fix(server): Fix constraints around large environment variables

marcindobry:
Our API allowed for environment variables with keys up-to 256 characters, and values up-to 4000 characters. However, our DB only allowed 255 characters each. This introduces a migration and adds test to match what the API declared constraints.

---

It adds integration coverage at the controller and service layers to confirm that boundary-length payloads are accepted while invalid submissions are rejected, and updates the API type definition so the env query parameter is required.

<details>

• packages/database/lib/migrations/20260116165950_increase_environment_variables_column_sizes.cjs
• packages/server/lib/controllers/v1/environment/variables/postVariables.integration.test.ts
• packages/shared/lib/services/environment.service.integration.test.ts
• packages/types/lib/environment/variable/api.ts

</details>

---

### DIFF
--- a/packages/types/lib/environment/variable/api.ts
+++ b/packages/types/lib/environment/variable/api.ts
@@ -5,6 +5,7 @@ export type ApiEnvironmentVariable = Pick<DBEnvironmentVariable, 'name' | 'value
 export type PostEnvironmentVariables = Endpoint<{
     Method: 'POST';
     Path: '/api/v1/environments/variables';
+    Querystring: { env: string };
     Body: { variables: { name: string; value: string }[] };
     Success: {
         success: boolean;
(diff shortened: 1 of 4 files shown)


packages/types/lib/environment/variable/api.ts
```diff
@@ -5,6 +5,7 @@ export type ApiEnvironmentVariable = Pick<DBEnvironmentVariable, 'name' | 'value
 export type PostEnvironmentVariables = Endpoint<{
     Method: 'POST';
     Path: '/api/v1/environments/variables';
+    Querystring: { env: string };
```
marcindobry, line 8: So apparently we are sending this in the POST, and from what I understand the code expects it too https://github.com/NangoHQ/nango/blob/f7f5367d7a8f69f62240aec515adfe7ce7ffb781/packages/server/lib/controllers/v1/environment/variables/postVariables.ts#L17

image

kaposke: That's something I'd like to improve in v2 (lots of endpoints don't need environment).

kaposke (approved):

PR #5268, opened 2026-01-20
fix(action-payloads): remove allow listed variable

khaliqgant:
---

**Enforce 2MB action payload limit for all non-enterprise accounts**

Removes the action payload allowlist so that the 2 MB response size limit applies uniformly to all non-enterprise deployments. Also drops the server-side logging path that previously warned allowlisted tenants about large payload usage.

<details>

• Removed `actionAllowListCustomers` usage from `postTriggerAction` controller and related logging logic
• Deleted `actionAllowListCustomers` export from `packages/utils/lib/environment/detection.ts`
• Simplified runner payload size enforcement in `packages/runner/lib/exec.ts` to gate solely on `isEnterprise`

</details>

<details>

• `packages/server/lib/controllers/action/postTriggerAction.ts`
• `packages/utils/lib/environment/detection.ts`
• `packages/runner/lib/exec.ts`

</details>

---

### DIFF
--- a/packages/utils/lib/environment/detection.ts
+++ b/packages/utils/lib/environment/detection.ts
@@ -33,5 +33,3 @@ export const flagEnforceCLIVersion = process.env['FLAG_ENFORCE_CLI_VERSION'] ===
 export const flags = {
     hasAdminCapabilities: Boolean(process.env['NANGO_ADMIN_UUID'])
 };
-
-export const actionAllowListCustomers = [0, 662, 1760, 1920, 4530, 5166, 7157, 7359, 7696, 2981, 6254];
--- a/packages/runner/lib/exec.ts
+++ b/packages/runner/lib/exec.ts
@@ -11,7 +11,7 @@ import * as unzipper from 'unzipper';
 import * as zod from 'zod';
 
 import { ActionError, ExecutionError, SDKError } from '@nangohq/runner-sdk';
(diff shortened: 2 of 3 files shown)


packages/utils/lib/environment/detection.ts
```diff
@@ -33,5 +33,3 @@ export const flagEnforceCLIVersion = process.env['FLAG_ENFORCE_CLI_VERSION'] ===
 export const flags = {
     hasAdminCapabilities: Boolean(process.env['NANGO_ADMIN_UUID'])
 };
-
-export const actionAllowListCustomers = [0, 662, 1760, 1920, 4530, 5166, 7157, 7359, 7696, 2981, 6254];
```
rossmcewan, line 37: I don't think this is a good thing having production customer account numbers in our code base. This should rather be an environment variable (or another kind of setting).

khaliqgant: Agree, but they're being removed in this case.

rossmcewan (commented):
--

packages/utils/lib/environment/detection.ts
rossmcewan: Ignore my comments. I'm dozing. Is this already in the code base?

khaliqgant: 😂, yes. Already in the codebase and is being actively removed since it was a temporary allowance hence the hard coded ints

rossmcewan (approved):

PR #5271, opened 2026-01-20
feat(multi-api-secret): Create api_secrets table with 1:N relation to environments.

thw-nango:
Every time an environment secret is created or modified, we also create or modify its mirror image in api_secrets.
This means that for now, the relation remains 1:1.
This lays the foundation for a transition away from a secret-environment correspondence, to a 1 environment N secrets model.
In a second step, we will enable multiple keys per environment.

---

**Introduce mirrored `api_secrets` store for environment keys**

The PR adds an `api_secrets` table and wires environment secret lifecycle events to maintain a mirrored default secret per environment. Secret creation, rotation, activation, and database encryption flows now run transactional updates that keep `_nango_environments` and `api_secrets` in sync, with new helper methods in `encryption.manager` plus integration coverage validating the mirror stays current and decryptable.

<details>

• Create migration `packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs` defining the `api_secrets` table, indexes, default backfill from `_nango_environments`, and cascading deletes.
• Add `DBAPISecret` type and new helper methods `decryptAPISecret`, `upsertDefaultAPISecret`, and `getDefaultAPISecret` to `packages/shared/lib/utils/encryption.manager.ts`.
• Wrap environment creation and secret activation in transactions that update `_nango_environments` and call `encryptionManager.upsertDefaultAPISecret` in `packages/shared/lib/services/environment.service.ts`.
• Invoke `upsertDefaultAPISecret` during `encryptDatabase()` to keep mirrored secrets aligned when encryption gets enabled post-migration.
• Extend integration tests in `packages/shared/lib/services/environment.service.integration.test.ts` to assert mirrored default secret hashing and decryptability before and after rotations.

</details>

<details>

• packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
• packages/shared/lib/services/environment.service.ts
• packages/shared/lib/utils/encryption.manager.ts
• packages/shared/lib/services/environment.service.integration.test.ts
• packages/types/lib/environment/db.ts

</details>

---


packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
```diff
@@ -0,0 +1,48 @@
...
+            select
+                id                              as environment_id,
+                'default'                       as name,
+                coalesce(secret_key, '')        as secret,
+                coalesce(secret_key_iv, '')     as iv,
+                coalesce(secret_key_tag, '')    as tag,
+                coalesce(secret_key_hashed, '') as hashed,
+                true                            as is_default
+            from
+                _nango_environments;
```
TBonnin: we use UPPERCASE for SQL keywords. It makes it easier to parse

packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
```diff
@@ -0,0 +1,48 @@
...
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`
+        create table if not exists api_secrets (
+            id             serial primary key,
+            environment_id integer not null references _nango_environments(id),
+            name           varchar(255) not null,
```
TBonnin: let's call it display_name to make it clearer it is meant for human consumption

packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
```diff
@@ -0,0 +1,48 @@
+exports.config = { transaction: false };
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`
+        create table if not exists api_secrets (
```
TBonnin: are we ever gonna need some other types of secrets? If yes would it makes sense to call this table `secrets` so we can extend it later?

packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
```diff
@@ -0,0 +1,48 @@
...
+            updated_at     timestamptz not null default CURRENT_TIMESTAMP
+        );
+
+        create index if not exists api_secrets_environment_id on api_secrets (environment_id);
+
+        create unique index if not exists api_secrets_one_default_per_environment
+            on api_secrets (environment_id)
+            where is_default = true;
+
+        insert into api_secrets (environment_id, name, secret, iv, tag, hashed, is_default)
```
TBonnin: can we make this subquery idempotent to avoid partial success this the migration isn't running in a transaction

packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
```diff
@@ -0,0 +1,48 @@
...
+ */
+exports.up = async function (knex) {
+    await knex.raw(`
+        create table if not exists api_secrets (
+            id             serial primary key,
+            environment_id integer not null references _nango_environments(id),
+            name           varchar(255) not null,
+            secret         varchar(255) not null,
+            iv             varchar(255) not null,
+            tag            varchar(255) not null,
```
TBonnin: are iv and tag always present? I recall in dev for instance secrets are not always encrypted?

packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
```diff
@@ -0,0 +1,48 @@
...
+            hashed         varchar(255) not null,
+            is_default     boolean not null default false,
```
TBonnin: what's the meaning of a `default` secret?

packages/shared/lib/services/environment.service.ts
```diff
@@ -379,6 +378,28 @@ class EnvironmentService {
...
+                name: 'default',
+                secret: env.secret_key,
+                iv: env.secret_key_iv ?? '',
+                tag: env.secret_key_tag ?? '',
+                hashed: env.secret_key_hashed!,
+                is_default: true,
+                updated_at: new Date()
+            })
+            .onConflict(trx.raw('(environment_id) where is_default = true'))
+            .merge(['secret', 'iv', 'tag', 'hashed', 'updated_at']);
```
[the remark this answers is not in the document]
thw-nango: Done

packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
thw-nango: Done.

packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
thw-nango: The longer term plan is to get rid of secrets in the DB entirely. Only keyholes should remain in end-state. Plus one recovery secret in an external KMS.

packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
thw-nango: In Postgres, non-transactional statements run in autocommit mode. In that mode, every individual statement is run in an implicit transaction; it either succeeds fully or not at all. In our case, if this statement fails (e.g. no disk space left), the migration fails and down() is used for rollback, deleting the entire table.

packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
thw-nango: I added a comment. It is the secret used by anything that needs "the environment's secret" e.g. runners. As we migrate away from storing secrets, it will become obsolete as well.

packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
thw-nango: Correct, in the initial commit I used a WHERE clause to filter out unencrypted secrets. In the commit after that I included them by coalescence. IV & AuthTag are epsilon for unencrypted secrets. It would be worth a (group) discussion if we want to continue supporting unencrypted databases as we revamp our security story. The special-casing of plaintext credentials is already tedious and will only get worse.

packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
thw-nango: Done.

packages/shared/lib/services/environment.service.ts
```diff
@@ -287,27 +291,21 @@ class EnvironmentService {
...
+            secret_key: environment.pending_secret_key!,
+            secret_key_iv: environment.pending_secret_key_iv!,
+            secret_key_tag: environment.pending_secret_key_tag!,
+            secret_key_hashed: await hashSecretKey(decrypted.pending_secret_key!),
+            pending_secret_key: null,
+            pending_secret_key_iv: null,
+            pending_secret_key_tag: null
+        };
+        await db.knex.transaction(async (trx) => {
+            await trx<DBEnvironment>(TABLE).where({ id }).update(update);
```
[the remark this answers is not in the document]
thw-nango: The previous code was not checking for update. It was an unnecessary second re-fetch from DB.

packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
TBonnin: knex `down` isn't automatically called when a migration query fails. With transaction disabled it means the database can be left in a partial state. That's why we usually prefer to write SQL in a way that executing it twice is idempotent

packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
TBonnin: do we need the concept of default secret then? if it is gonna become obsolete can we get rid of it right now?

packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
```diff
@@ -0,0 +1,52 @@
...
+            updated_at     TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP
+        );
+
+        CREATE INDEX IF NOT EXISTS api_secrets_environment_id ON api_secrets (environment_id);
+
+        CREATE UNIQUE INDEX IF NOT EXISTS api_secrets_one_default_per_environment
+            ON api_secrets (environment_id)
+            WHERE is_default = true;
+
+        INSERT INTO api_secrets (environment_id, display_name, secret, iv, tag, hashed, is_default)
```
marcindobry, line 31: Related to this conversation https://github.com/NangoHQ/nango/pull/5271#discussion_r2709405694

I think we need to add ON CONFLICT DO NOTHING to the INSERT to make it idempotent if it fails and we result in a partial completion state.

marcindobry (approved):
Apart from 1 comment LGTM as first step towards having 1:N environments-secrets

packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
thw-nango: In autocommit mode (non-transaction), since this is one statement, it is never going to be left in a partial state. The statement is run in an implicit transaction. It either succeeds or fails fully. That's one of the great things about Postgres.

packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
thw-nango: In autocommit mode (non-transaction), since this is one statement, it is never going to be left in a partial state. The statement is run in an implicit transaction. It either succeeds or fails fully. That's one of the great things about Postgres.

I have added the ON CONFLICT DO NOTHING clause nevertheless.

packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
thw-nango: Sorry, I just saw this now. Since Github resolves these comments when I reply, I do not get any notifications afterwards. Still getting used to the tooling. Yes, we need the default for now to use it wherever any code uses `environment.secret` (basically everywhere where it is assumed there is 1 secret, like in runners). But I agree that we should get rid of it ASAP in step 2.

TBonnin: we could keep things simple and use the most recent key as default.

packages/database/lib/migrations/20260119161100_add_api_secrets_table.cjs
TBonnin: I am not talking about postgres leaving tables in partial state. I am talking about the migration being partially applied since `transaction: false`. Here the non idempotent query is the last one so there is no risk of inserting twice but in general we try to make migrations idempotent. This is automatically handled when migration runs in a transaction but when it is disabled we need to be extra careful and handle scenarios where a migration fails

PR #5274, opened 2026-01-20
feat: Connection and connect session tags

marcindobry:
⚠️ The PR seems big but most of the new lines are actually just tests. ⚠️ 

Add tags to connect sessions and connections. There are multiple ways these tags get set
- Existing endUser/organization data on connect sessions/connections is translated into tags
- Can be set as a separate field when creating connect sessions / reconnecting. These will get passed on to a new connection that gets created during the connect session
- Can be updated via the API (e.g. PATCH /connections/:connectionId)

They are then visible in
- The connect session / connection APIs, although they are undocumented
- Various auth webhooks e.g. connection-created (if they are set). This is also undocumented

This is a 3rd attempt at building an attribution mechanism between connect sessions and connections and we want to give it to customers who weren't able to use the previous 2 to first give us some feedback before we document it fully.

I left a few comments with questions:
- should we normalize tags to lower case?
- is 64 tags per connection too many?

Further reading: https://linear.app/nango/project/connection-tags-9bf95766e842/overview

---

The implementation also normalizes tag keys to lowercase with validation safeguards and extends the connection listing endpoints so tags can be queried and filtered alongside their inclusion in auth webhooks.

<details>

• packages/server/lib/controllers/connection/*
• packages/server/lib/controllers/connect/*
• packages/server/lib/controllers/auth/*
• packages/shared/lib/services/connection.service.ts
• packages/shared/lib/services/connectSession.service.ts
• packages/shared/lib/services/tags/*
• packages/types/lib/*
• packages/webhooks/lib/*

</details>

---

### DIFF
--- a/packages/server/lib/controllers/auth/postApiKey.ts
+++ b/packages/server/lib/controllers/auth/postApiKey.ts
@@ -168,7 +168,8 @@ export const postPublicApiKeyAuthorization = asyncWrapper<PostPublicApiKeyAuthor
             connectionConfig,
             metadata: {},
             config,
-            environment
+            environment,
+            tags: connectSession?.tags
         });
         if (!updatedConnection) {
             res.status(500).send({ error: { code: 'server_error', message: 'failed to create connection' } });
--- a/packages/server/lib/controllers/connection/connectionId/patchConnection.integration.test.ts
+++ b/packages/server/lib/controllers/connection/connectionId/patchConnection.integration.test.ts
@@ -1,0 +1,258 @@
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+
+import db from '@nangohq/database';
+import { seeders } from '@nangohq/shared';
+
+import { isError, isSuccess, runServer, shouldBeProtected } from '../../../utils/tests.js';
+
+import type { DBConnection } from '@nangohq/types';
+
+let api: Awaited<ReturnType<typeof runServer>>;
+
+const endpoint = '/connections/:connectionId';
+
+describe(`PATCH ${endpoint}`, () => {
+    beforeAll(async () => {
+        api = await runServer();
+    });
+    afterAll(() => {
+        api.server.close();
+    });
+
+    it('should be protected', async () => {
+        const res = await api.fetch(endpoint, {
+            method: 'PATCH',
+            params: { connectionId: 'test' },
+            query: { provider_config_key: 'github' },
+            body: {}
+        });
+
+        shouldBeProtected(res);
+    });
+
+    it('should return 400 for unknown provider', async () => {
+        const { env } = await seeders.seedAccountEnvAndUser();
+
+        const res = await api.fetch(endpoint, {
+            method: 'PATCH',
+            token: env.secret_key,
+            params: { connectionId: 'test' },
+            query: { provider_config_key: 'unknown' },
+            body: {}
+        });
+
+        isError(res.json);
+        expect(res.json).toMatchObject({
+            error: { code: 'unknown_provider_config' }
+        });
+        expect(res.res.status).toBe(400);
+    });
+
+    it('should return 404 for unknown connection', async () => {
+        const { env } = await seeders.seedAccountEnvAndUser();
+        await seeders.createConfigSeed(env, 'github', 'github');
+
+        const res = await api.fetch(endpoint, {
+            method: 'PATCH',
+            token: env.secret_key,
+            params: { connectionId: 'unknown' },
+            query: { provider_config_key: 'github' },
+            body: {}
+        });
+
+        isError(res.json);
+        expect(res.json).toStrictEqual<typeof res.json>({
+            error: { code: 'not_found', message: 'Failed to find connection' }
+        });
+        expect(res.res.status).toBe(404);
+    });
+
+    describe('end_user', () => {
+        it('should update end_user only', async () => {
+            const { env } = await seeders.seedAccountEnvAndUser();
+            await seeders.createConfigSeed(env, 'github', 'github');
+            const conn = await seeders.createConnectionSeed({ env, provider: 'github' });
+
+            const res = await api.fetch(endpoint, {
+                method: 'PATCH',
+                token: env.secret_key,
+                params: { connectionId: conn.connection_id },
+                query: { provider_config_key: 'github' },
+                body: {
+                    end_user: { id: 'user-123', email: 'test@example.com', display_name: 'Test User' }
+                }
+            });
+
+            isSuccess(res.json);
+            expect(res.json).toStrictEqual({ success: true });
+
+            const updatedConn = await db.knex.select('*').from<DBConnection>('_nango_connections').where({ id: conn.id }).first();
+            expect(updatedConn?.end_user_id).not.toBeNull();
+        });
+    });
+
+    describe('tags', () => {
+        it('should update tags only', async () => {
+            const { env } = await seeders.seedAccountEnvAndUser();
+            await seeders.createConfigSeed(env, 'github', 'github');
+            const conn = await seeders.createConnectionSeed({ env, provider: 'github' });
+
+            const res = await api.fetch(endpoint, {
+                method: 'PATCH',
+                token: env.secret_key,
+                params: { connectionId: conn.connection_id },
+                query: { provider_config_key: 'github' },
+                body: { tags: { projectId: '123', orgId: '456' } }
+            });
+
+            isSuccess(res.json);
+            expect(res.json).toStrictEqual({ success: true });
+
+            // Keys are normalized to lowercase by connectionTagsSchema
+            const updatedConn = await db.knex.select('*').from<DBConnection>('_nango_connections').where({ id: conn.id }).first();
+            expect(updatedConn?.tags).toStrictEqual({ projectid: '123', orgid: '456' });
+        });
+
+        it('should update both tags and end_user', async () => {
+            const { env } = await seeders.seedAccountEnvAndUser();
+            await seeders.createConfigSeed(env, 'github', 'github');
+            const conn = await seeders.createConnectionSeed({ env, provider: 'github' });
+
+            const res = await api.fetch(endpoint, {
+                method: 'PATCH',
+                token: env.secret_key,
+                params: { connectionId: conn.connection_id },
+                query: { provider_config_key: 'github' },
+                body: {
+                    tags: { env: 'production' },
+                    end_user: { id: 'user-456' }
+                }
+            });
+
+            isSuccess(res.json);
+            expect(res.json).toStrictEqual({ success: true });
+
+            const updatedConn = await db.knex.select('*').from<DBConnection>('_nango_connections').where({ id: conn.id }).first();
+            // Tags are merged: end_user generated tags + explicit tags (explicit tags take priority)
+            expect(updatedConn?.tags).toStrictEqual({ end_user_id: 'user-456', env: 'production' });
+            expect(updatedConn?.end_user_id).not.toBeNull();
+        });
+
+        it('should fail with invalid tag key format', async () => {
+            const { env } = await seeders.seedAccountEnvAndUser();
+            await seeders.createConfigSeed(env, 'github', 'github');
+            const conn = await seeders.createConnectionSeed({ env, provider: 'github' });
+
+            const res = await api.fetch(endpoint, {
+                method: 'PATCH',
+                token: env.secret_key,
+                params: { connectionId: conn.connection_id },
+                query: { provider_config_key: 'github' },
+                body: {
+                    tags: { '123invalid': 'value' }
+                }
+            });
+
+            isError(res.json);
+            expect(res.json).toMatchObject({
+                error: { code: 'invalid_body' }
+            });
+            expect(res.res.status).toBe(400);
+        });
+
+        it('should allow tag values with spaces', async () => {
+            const { env } = await seeders.seedAccountEnvAndUser();
+            await seeders.createConfigSeed(env, 'github', 'github');
+            const conn = await seeders.createConnectionSeed({ env, provider: 'github' });
+
+            const res = await api.fetch(endpoint, {
+                method: 'PATCH',
+                token: env.secret_key,
+                params: { connectionId: conn.connection_id },
+                query: { provider_config_key: 'github' },
+                body: {
+                    tags: { key: 'value with spaces' }
+                }
+            });
+
+            isSuccess(res.json);
+            expect(res.json).toStrictEqual({ success: true });
+
+            const updatedConn = await db.knex.select('*').from<DBConnection>('_nango_connections').where({ id: conn.id }).first();
+            expect(updatedConn?.tags).toStrictEqual({ key: 'value with spaces' });
+        });
+
+        it('should replace existing tags completely', async () => {
+            const { env } = await seeders.seedAccountEnvAndUser();
+            await seeders.createConfigSeed(env, 'github', 'github');
+            const conn = await seeders.createConnectionSeed({ env, provider: 'github' });
+
+            await db
+                .knex('_nango_connections')
+                .where({ id: conn.id })
+                .update({ tags: { old: 'value', keep: 'this' } });
+
+            // Update with new tags (should replace, not merge)
+            const res = await api.fetch(endpoint, {
+                method: 'PATCH',
+                token: env.secret_key,
+                params: { connectionId: conn.connection_id },
+                query: { provider_config_key: 'github' },
+                body: { tags: { new: 'value' } }
+            });
+
+            isSuccess(res.json);
+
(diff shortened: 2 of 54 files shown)


packages/server/lib/controllers/connection/connectionId/patchConnection.integration.test.ts
```diff
@@ -0,0 +1,268 @@
+import { afterAll, beforeAll, describe, expect, it } from 'vitest';
```
marcindobry, line 1: Since there were tests missing for patchConnection, I added some basic ones in addition to tag ones

packages/shared/lib/services/tags/utils.ts
```diff
@@ -0,0 +1,67 @@
...
+const TAG_VALUE_MAX_LENGTH = 200;
+const TAG_MAX_COUNT = 64;
+const TAG_KEY_PATTERN = /^[a-zA-Z][a-zA-Z0-9_\-:./]*$/;
+const TAG_VALUE_PATTERN = /^[a-zA-Z0-9_\-:./]*$/;
+
+/**
+ * Normalizes tag keys and values to lowercase.
+ * Returns an error if normalization would create duplicate keys.
+ */
+export function normalizeTags(tags: Record<string, string>): { success: true; tags: Tags } | { success: false; error: string } {
```
marcindobry: I'd like an opinion here. I wanted to follow what Datadog does (I expect devs to be used to that) and have all tag keys and values be lowercase and convert the inputs to lowercase if they are not. However, I'm not sure if either of these is a good idea (both enforcing lowercase only, and doing auto-conversion)

packages/server/lib/helpers/validation.ts
```diff
@@ -148,7 +148,22 @@ export const connectionCredentialsGithubAppSchema = z.strictObject({
...
+    })
+    .max(64, { message: 'Tag keys must be at most 64 characters' });
+const connectionTagsValueSchema = z
+    .string()
+    .regex(/^[a-zA-Z0-9_\-:./]*$/, { message: 'Tag values must contain only alphanumerics, underscores, hyphens, colons, periods, or slashes' })
+    .max(200, { message: 'Tag values must be at most 200 characters' });
+
 export const connectionTagsSchema = z
+    .record(connectionTagsKeySchema, connectionTagsValueSchema)
+    .refine((v) => Object.keys(v).length <= 64, { message: 'Tags cannot contain more than 64 keys' });
```
marcindobry: Is this too generous? 🤔

packages/server/lib/controllers/auth/postApiKey.ts
```diff
@@ -207,6 +208,7 @@ export const postPublicApiKeyAuthorization = asyncWrapper<PostPublicApiKeyAuthor
 
         if (isConnectSession) {
             await syncEndUserToConnection(db.knex, { connectSession, connection: updatedConnection.connection, account, environment });
+            await syncTagsToConnection(db.knex, { connectSession, connection: updatedConnection.connection });
```
TBonnin: never been a huge fan of the previous naming, especially since sync is a pretty loaded term in Nango but also because it is a extra level of indirection on top of updateConnectionTags that doesn't bring much value.
Why not call updateConnectionTags with an extra allowEmpty params if necessary?
`await updateConnectionTags(db, { connection: updatedConnection.connection, tags: connection.tags, allowEmpty:false });`

packages/shared/lib/services/tags/utils.ts
```diff
@@ -0,0 +1,62 @@
+import type { Tags } from '@nangohq/types';
+
+const TAG_KEY_MAX_LENGTH = 64;
+const TAG_VALUE_MAX_LENGTH = 200;
+const TAG_MAX_COUNT = 64;
```
TBonnin: I would be less generous and start with 10 or 15.

packages/shared/lib/services/tags/utils.ts
TBonnin: I like the idea of tag keys being case insensitive. It makes things simpler. It would need to be documented and return clear errors though. 
I don't think values should be normalized though. I don't think Datadog lowercase the tag values.
Regardless of the normalization, search functionality should also normalize the input.

packages/shared/lib/services/tags/utils.ts
```diff
@@ -0,0 +1,62 @@
...
+const TAG_KEY_PATTERN = /^[a-zA-Z][a-zA-Z0-9_\-:./]*$/;
+const TAG_VALUE_PATTERN = /^[a-zA-Z0-9_\-:./]*$/;
+
+/**
+ * Normalizes tag keys and values to lowercase.
+ * Returns an error if normalization would create duplicate keys.
+ */
+export function normalizeTags(tags: Record<string, string>): { success: true; tags: Tags } | { success: false; error: string } {
```
TBonnin: Let's use zod for the validation instead of trying to re-invent the wheel.

packages/server/lib/controllers/connection/connectionId/patchConnection.ts
```diff
@@ -74,5 +75,12 @@ export const patchPublicConnection = asyncWrapper<PatchPublicConnection>(async (
         });
     }
 
+    if (body.tags) {
+        await updateConnectionTags(db.knex, {
+            connection,
+            tags: body.tags
+        });
+    }
```
TBonnin, line 91: Do I understand correctly that all tags are updated and we don't support appending/merging one or more tags to the existing ones? 
I can definitely see customers wanting to do add some tags after the fact without touching the existing ones.

packages/shared/lib/services/tags/utils.ts
```diff
@@ -0,0 +1,62 @@
+import type { Tags } from '@nangohq/types';
+
+const TAG_KEY_MAX_LENGTH = 64;
+const TAG_VALUE_MAX_LENGTH = 200;
+const TAG_MAX_COUNT = 64;
+const TAG_KEY_PATTERN = /^[a-zA-Z][a-zA-Z0-9_\-:./]*$/;
+const TAG_VALUE_PATTERN = /^[a-zA-Z0-9_\-:./]*$/;
```
TBonnin: no space?

TBonnin (commented):
we probably want to add support for tags search in the `GET /connnections` and `GET /connections/:id` endpoints. Can be done in next PR though? https://github.com/NangoHQ/nango/blob/4f1b8a4d4fd3975f6d6af0ce00b2de52cea596e1/packages/server/lib/controllers/connection/connectionId/getConnection.ts#L73

packages/shared/lib/services/tags/utils.ts
marcindobry: > I don't think Datadog lowercase the tag values.

From what I've read they do, but after thinking about it I don't think we should because they could be IDs that are case sensitive so I'll definitely change that.

packages/server/lib/controllers/auth/postApiKey.ts
marcindobry: I changed it to be part of the upsert. `syncEndUserToConnection` does some additional stuff (like upserting an endUser in a separate table), but tags are perfect fine to be part of the upsert with the rest of the data.

packages/server/lib/controllers/connection/connectionId/patchConnection.ts
marcindobry: We'd need a separate set of endpoints for it, don't think a PATCH Connection is appropriate for such manipulations. It would be difficult to distinguish what specific payloads mean, e.g.: sending a single tag - does  this mean add that tag to the existing ones, or delete the existing ones? If the former, then how do you even delete existing tags?

For now having it here still allows all these operations, it just requires doing a GET Connection first, figuring out the final tag state and then sending a PATCH Connection.

packages/shared/lib/services/tags/utils.ts
marcindobry: I'm not opposed to adding it but do you think it would be valuable?

marcindobry:
@TBonnin re tag search parameter in `GET Connections`: 

This is actually a gap I was meant to implement but forgot so going to push changes for it as part of this PR. However, it wouldn't be as part of the search parameter but a separate query parameter `tags`. This is something we briefly discussed at the retreat (the format of `tags` query parameter). You can read more about it here https://linear.app/nango/project/connection-tags-9bf95766e842/overview#give-the-ability-to-set-b86ac889

I actually want to eventually deprecate the search parameter. It isn't well documented how it works, and after tags are adopted as the replacement for end_user, the search parameter will be completely obsolete anyways, because it only searches for end_user fields. I wrote more about it here https://linear.app/nango/project/connection-tags-9bf95766e842/overview#deprecate-search-parameter-in-get-28961acb

packages/server/lib/controllers/connection/connectionId/patchConnection.ts
```diff
@@ -60,6 +61,10 @@ export const patchPublicConnection = asyncWrapper<PatchPublicConnection>(async (
 
     const connection = connectionRes.response;
 
+    // Generate tags from end_user (similar to postSessions.ts and postReconnect.ts)
+    const endUserTags = body.end_user ? buildTagsFromEndUser(body.end_user, null) : {};
+    const mergedTags = { ...endUserTags, ...body.tags };
```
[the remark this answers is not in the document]
marcindobry: No, the user might not be using tags at all yet, and we want the underlying tags to be updated even if just the end_user data changes. If they start using tags, we will recommend not to use endUser as this is a replacement. This behavior will be well documented.

packages/shared/lib/services/connection.service.ts
```diff
@@ -101,14 +102,16 @@ class ConnectionService {
...
+        metadata,
+        tags
     }: {
         connectionId: string;
         providerConfigKey: string;
         parsedRawCredentials: AllAuthCredentials;
         connectionConfig?: ConnectionConfig;
         environmentId: number;
         metadata?: Metadata | null;
+        tags: Tags;
```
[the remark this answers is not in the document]
marcindobry: This is fair, I'll address this.

TBonnin:
> @TBonnin re tag search parameter in `GET Connections`:
> 
> This is actually a gap I was meant to implement but forgot so going to push changes for it as part of this PR

Fine with me. Be aware that one of the main use-case for `end users` and therefore `tags` is the capability to search for connection(s). Without it the feature should be considered incomplete and we should refrain from announcing it until the search is implemented.

packages/shared/lib/services/tags/utils.ts
TBonnin: are we not migrating `end_user.display_name`?

TBonnin:
The documentation must also be updated but maybe you are planning to do it in next PR

marcindobry:
> Fine with me. Be aware that one of the main use-case for end users and therefore tags is the capability to search for connection(s). Without it the feature should be considered incomplete and we should refrain from announcing it until the search is implemented.

This is implemented in this PR now.

> The documentation must also be updated but maybe you are planning to do it in next PR

Yup I'll do it as a follow-up. Didn't want to inflate the diff numbers in this one

packages/server/lib/controllers/connection/getConnections.ts
```diff
@@ -16,11 +16,40 @@ const validationQuery = z
...
+            const value = trimmed.slice(colonIndex + 1);
+            if (key in tags) {
+                throw new DuplicateTagKeyError(key);
+            }
+            tags[key] = value;
+        }
+    }
+    return tags;
+}
+
```
TBonnin: let's validate using zod to stay consistant

packages/shared/lib/services/connection.service.ts
```diff
@@ -765,6 +786,11 @@ class ConnectionService {
                     subQuery.join('_nango_configs', '_nango_connections.config_id', '_nango_configs.id').whereIn('_nango_configs.unique_key', integrationIds);
                 }
 
+                // Filter by tags (JSONB containment using GIN index)
+                if (tags && Object.keys(tags).length > 0) {
+                    subQuery.whereRaw('_nango_connections.tags @> ?', [JSON.stringify(tags)]);
```
TBonnin: I am pretty sure pg does it automatically but we could type cast with `?::jsonb` to avoid edge-cases

packages/server/lib/controllers/connection/getConnections.ts
```diff
@@ -16,11 +16,40 @@ const validationQuery = z
...
 
+class DuplicateTagKeyError extends Error {
+    constructor(public readonly key: string) {
+        super(`Duplicate tag key: ${key}`);
+        this.name = 'DuplicateTagKeyError';
+    }
+}
+
+/**
+ * Parse tags query parameter from "key:value,key2:value2" format to Record<string, string>
```
TBonnin: Have you considered other options:
- repeated keys with dots: `GET /connections?tags.env=prod&tags.region=us`
- brackets: GET /connections?tags[env]=prod&tags[region]=us

marcindobry: Yes, [I looked into it](https://linear.app/nango/project/connection-tags-9bf95766e842/overview#give-the-ability-to-set-b86ac889) but there is no single industry standard how to do this stuff, they all have pros and cons. Even Datadog has a different way of doing it depending which API you use :/ I just made the call that this is what I prefer (shorter and easier to parse for a human when debugging). 

Stuff like this is partly the reason why I wanted to wait with documenting this feature until we see some real-world usage with some early-access customers who this would unblock to migrate away from the public key approach of creating connections.

packages/shared/lib/services/tags/schema.ts
```diff
@@ -0,0 +1,58 @@
...
+            seen.add(normalized);
+        }
+    })
+    .transform((tags) => {
+        const normalized: Record<string, string> = {};
+        for (const [key, value] of Object.entries(tags)) {
+            normalized[key.toLowerCase()] = value;
+        }
+        return normalized;
+    });
```
TBonnin, line 65: from the spec:
> The email tag needs to be a valid email, and the domain tag needs to be a valid URL

packages/server/lib/services/connectSession.service.ts
```diff
@@ -98,10 +103,26 @@ export async function createConnectSession(
             | 'overrides'
             | 'endUser'
             | 'endUserId'
+            | 'tags'
         >,
         'connectionId'
     >
 ): Promise<Result<ConnectSession, ConnectSessionError>> {
+    let normalizedTags: Tags = {};
+    if (tags && Object.keys(tags).length > 0) {
```
TBonnin: the safe parse should be able to handle this scenario. All of the validation logic should live in the same place imho

packages/shared/lib/services/tags/schema.ts
```diff
@@ -0,0 +1,62 @@
...
+                }
+            }
+        }
+    })
+    .transform((tags) => {
+        const normalized: Record<string, string> = {};
+        for (const [key, value] of Object.entries(tags)) {
+            normalized[key.toLowerCase()] = value;
+        }
+        return normalized;
```
TBonnin, line 64: should the transform be done first. then there is no need to do it again in the `refine`?

packages/shared/lib/services/tags/schema.ts
```diff
@@ -0,0 +1,62 @@
...
+export const TAG_VALUE_MAX_LENGTH = 200;
+
+const connectionTagsKeySchema = z
+    .string()
+    .regex(/^[a-zA-Z][a-zA-Z0-9_\-./]*$/, {
+        message: 'Tag keys must start with a letter and contain only alphanumerics, underscores, hyphens, periods, or slashes'
+    })
+    .max(TAG_KEY_MAX_LENGTH, { message: `Tag keys must be at most ${TAG_KEY_MAX_LENGTH} characters` });
+
+const connectionTagsValueSchema = z.string().max(TAG_VALUE_MAX_LENGTH, { message: `Tag values must be at most ${TAG_VALUE_MAX_LENGTH} characters` });
```
TBonnin: should we disallow empty string?

TBonnin (approved):

packages/shared/lib/services/tags/schema.ts
marcindobry: No, the refine needs to be done first because the transform can silently allow duplicate tags that are not duplicates before lowercase is applied, e.g. `projectid` and `projectId`

packages/shared/lib/services/tags/schema.ts
marcindobry: We don't have a domain field anywhere, we just parse it out of the end user email here for this component https://github.com/NangoHQ/nango/blob/998eda94a2d8d5643d612589c4444f4a11a69193/packages/webapp/src/components/AvatarCustom.tsx#L141

marcindobry: We can add domain later as new functionality

PR #5276, opened 2026-01-20
fix: unknown bearer token should return a 401

TBonnin:
unknown_user_account is not a known error code https://github.com/NangoHQ/nango/blob/85976c3ce6451660f17ef5f6847b3e833fccf5ad/packages/shared/lib/utils/error.ts#L51
hence the API returning a 500 status code instead of 401

 

---

Secret and public key authentication failures now reuse the existing unknown_account error so the error manager responds with a 401, and an integration test against the /providers endpoint confirms the status code and payload for unknown bearer tokens.

<details>

• Replaced `unknown_user_account` with `unknown_account` in `packages/server/lib/middleware/access.middleware.ts` for secret and public key validation failures.
• Added an integration test in `packages/server/lib/routes.integration.test.ts` asserting a `401` with code `unknown_account` when an unrecognized bearer token is used.

</details>

<details>

• `packages/server/lib/middleware/access.middleware.ts`
• `packages/server/lib/routes.integration.test.ts`

</details>

---

### DIFF
--- a/packages/server/lib/routes.integration.test.ts
+++ b/packages/server/lib/routes.integration.test.ts
@@ -58,4 +58,22 @@ describe('route', () => {
             });
         });
     });
+
+    describe('Authenticated endpoints', () => {
+        it('should return 401 if unknown bearer token', async () => {
+            const res = await fetch(`${api.url}/providers`, {
+                method: 'GET',
+                headers: { Authorization: `Bearer 00000000-0000-4000-8000-000000000000` }
+            });
+
+            expect(res.status).toBe(401);
+            expect(await res.json()).toStrictEqual({
+                error: {
+                    message: 'Authentication failed. The provided authorization header does not match any account.',
+                    code: 'unknown_account',
+                    payload: {}
+                }
+            });
+        });
+    });
 });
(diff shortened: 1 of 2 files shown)


packages/server/lib/routes.integration.test.ts
```diff
@@ -58,4 +58,22 @@ describe('route', () => {
...
+        it('should return 401 if unknown bearer token', async () => {
+            const res = await fetch(`${api.url}/providers`, {
+                method: 'GET',
+                headers: { Authorization: `Bearer 00000000-0000-4000-8000-000000000000` }
+            });
+
+            expect(res.status).toBe(401);
+            expect(await res.json()).toStrictEqual({
+                error: {
+                    message: 'Authentication failed. The provided authorization header does not match any account.',
```
thw-nango: Testing the exact error message string is arguably not necessary, makes changing it later harder, and could lead to consumers relying on string matching in their error handling, instead of using the stable `code` identifier.

kaposke (approved):

packages/server/lib/routes.integration.test.ts
TBonnin: Testing it or not doesn't change the fact consumers can rely on it. 
Checking the exact message is indeed unnecessary though, removing it

PR #5283, opened 2026-01-21
feat(github-app-oauth): GitHub app when adding a repo doesnt return correctly

khaliqgant:
---

**Handle GitHub App callback without auth code for existing installations**

Refactors the GitHub App OAuth callback to handle cases where GitHub returns without an authorization code (e.g., when adding repositories to an existing installation). A new `finishGithubAppConnectionWithoutTokenExchange` helper encapsulates the connection completion logic so existing installations can be finalized via stored app credentials while keeping logging, validation, and websocket notifications consistent.

<details>

• Added helper `finishGithubAppConnectionWithoutTokenExchange` that reloads the connect session, finishes the connection via `connectionService.getAppCredentialsAndFinishConnection`, syncs end users, and triggers websocket notifications without needing a token exchange.
• Extended the CUSTOM-mode callback branch to look up existing connections via `connectionService.findConnectionsByMultipleConnectionConfigValues` using `installation_id` and skip the token exchange when a matching provider config is found.
• Reused the new helper for the prior `setup_action=update` path so both update flows now share identical error handling, logging, and notification logic.

</details>

<details>

• packages/server/lib/controllers/oauth.controller.ts

</details>

---


packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -1193,81 +1299,58 @@ class OAuthController {
...
-                    {
-                        connection: res.connection,
-                        environment,
-                        account,
-                        auth_mode: 'APP',
-                        operation: res.operation,
-                        endUser: connectSession?.connectSession.endUser ?? undefined
-                    },
-                    account,
+                await this.finishGithubAppConnectionWithoutTokenExchange({
```
TBonnin, line 1328: should this function be called `finishAppConnectionWithoutTokenExchange`. it doesn't seem to be absolutely unique to github app from the signature

khaliqgant: It is only used by github app so think the name is appropriate until we want to make it generalizable

TBonnin (approved):

PR #5293, opened 2026-01-22
fix: don't allow empty value edits in integration settings

kaposke:
In integration settings, when trying to clear a field that had a value, saving would fail in the backend. That's because our validation schemas, despite being optional (can be omitted), isn't nullable (values can't be cleared), and requires a string with minimum length of 1. Example:
```ts
export const integrationAuthTypeOAuthSchema = z
    .object({
        authType: z.enum(['OAUTH1', 'OAUTH2', 'TBA']),
        clientId: z.string().min(1).max(255).optional(),
        clientSecret: z.string().min(1).optional(),
        scopes: scopesSchema
    })
    .strict();
```

Ideally, I wanted to allow the users to clear fields by:
- Making the backend validation schemas `.nullable()`.
- Sending any empty strings as `null` from the frontend.

However, making these types nullable has all sorts of effects throughout the code-base (lots of things break).
To avoid unforeseen side-effects, I'm left with 2 options:
1. Removing the `.min()` validation and allowing empty strings to be saved.
   > Saving empty strings could still have unforeseen side-effects.
2. Simply don't allow empty strings to be saved. One can create an integration with empty values initially (those are saved as `null` by `POST /integrations`), but he can't edit a field to be empty later (because `PATCH /integrations` doesn't allow it).
   > This is the same behavior as the one we had before redesigning. It's safer and the user abilities are unchanged from before the redesign.
   
So this PR is solution 2.

---

The update formalizes this behavior by adding a shared validateNotEmpty helper that is applied across the integration credential inputs and by tightening the server controller to skip decoding custom private keys unless the required fields are present.

<details>

• Added `validateNotEmpty` utility and applied it to credential inputs in `InstallPluginSettings.tsx`, `McpGenericSettings.tsx`, `CustomAuthSettings.tsx`, `OAuthSettings.tsx`, `AppAuthSettings.tsx`, and `GeneralSettings.tsx`
• Adjusted `AppPrivateKeyInput.tsx` validation to rely on format checking while still failing empty input via the existing prefix/suffix check
• Updated `packages/server/lib/controllers/v1/integrations/providerConfigKey/getIntegration.ts` to decode custom `private_key` only when both `private_key` and `app_id` exist

</details>

<details>

• packages/webapp/src/pages/Integrations/providerConfigKey/Settings/components
• packages/webapp/src/pages/Integrations/utils.ts
• packages/server/lib/controllers/v1/integrations/providerConfigKey/getIntegration.ts

</details>

---

### DIFF
--- a/packages/server/lib/controllers/v1/integrations/providerConfigKey/getIntegration.ts
+++ b/packages/server/lib/controllers/v1/integrations/providerConfigKey/getIntegration.ts
@@ -55,8 +55,8 @@ export const getIntegration = asyncWrapper<GetIntegration>(async (req, res) => {
         webhookSecret = crypto.createHash('sha256').update(hash).digest('hex');
     }
 
-    if (provider.auth_mode === 'CUSTOM' && integration.custom) {
-        integration.custom['private_key'] = Buffer.from(integration.custom['private_key'] as string, 'base64').toString('ascii');
+    if (provider.auth_mode === 'CUSTOM' && integration.custom?.['private_key'] && integration.custom?.['app_id']) {
+        integration.custom['private_key'] = Buffer.from(integration.custom['private_key'], 'base64').toString('ascii');
         const hash = `${integration.custom['app_id']}${integration.custom['private_key']}${integration.app_link}`;
         webhookSecret = crypto.createHash('sha256').update(hash).digest('hex');
     }
--- a/packages/webapp/src/pages/Integrations/providerConfigKey/Settings/components/CustomAuthSettings.tsx
+++ b/packages/webapp/src/pages/Integrations/providerConfigKey/Settings/components/CustomAuthSettings.tsx
@@ -11,7 +11,7 @@ import { Label } from '@/components-v2/ui/label';
 import { useConfirmDialog } from '@/hooks/useConfirmDialog';
 import { apiPatchIntegration } from '@/hooks/useIntegration';
 import { useToast } from '@/hooks/useToast';
-import { validateUrl } from '@/pages/Integrations/utils';
+import { validateNotEmpty, validateUrl } from '@/pages/Integrations/utils';
 import { useStore } from '@/store';
 import { defaultCallback } from '@/utils/utils';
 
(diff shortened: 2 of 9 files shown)


packages/server/lib/controllers/v1/integrations/providerConfigKey/getIntegration.ts
```diff
@@ -55,8 +55,8 @@ export const getIntegration = asyncWrapper<GetIntegration>(async (req, res) => {
         webhookSecret = crypto.createHash('sha256').update(hash).digest('hex');
     }
 
-    if (provider.auth_mode === 'CUSTOM' && integration.custom) {
-        integration.custom['private_key'] = Buffer.from(integration.custom['private_key'] as string, 'base64').toString('ascii');
+    if (provider.auth_mode === 'CUSTOM' && integration.custom?.['private_key'] && integration.custom?.['app_id']) {
```
kaposke, line 58: This is unrelated, but one can now (after the new create integration form step) create an integration without private key and app Id, so we can't assume these fields exist (`integration.custom` can be `{}`)

packages/webapp/src/pages/Integrations/providerConfigKey/Settings/components/CustomAuthSettings.tsx
```diff
@@ -107,7 +107,12 @@ export const CustomAuthSettings: React.FC<{ data: GetIntegration['Success']['dat
             <div className="flex flex-col gap-2">
                 <Label htmlFor="client_id">Client ID</Label>
                 <>
-                    <EditableInput initialValue={integration.oauth_client_id || ''} onSave={handleClientIdSave} onEditingChange={setIsEditingClientId} />
+                    <EditableInput
+                        initialValue={integration.oauth_client_id || ''}
+                        onSave={handleClientIdSave}
+                        onEditingChange={setIsEditingClientId}
+                        validate={validateNotEmpty}
```
marcindobry, line 114: I don't have a working version to play around with, but does validation only run on change, or on form submission too? If it's the latter, would any form submission fail if any of these fields is empty anyways (meaning that no other modification can be done, even to non-empty fields)?

kaposke: These are "single-field forms". You edit and submit a single field at a time (see integration settings page).
Validation is done onChange and on submission.
Did I understand the question correctly?

marcindobry (approved):

PR #5296, opened 2026-01-23
fix(deps): Undepend from vulnerable `qs` lib version

thw-nango:
We depend on a 'high' vulnerability in `qs` via `webflow-api`.

I updated webflow-api to its latest version in `shared` and `scripts`.

This fixes the 'high' severity vulnerability and leaves us with:

> 15 vulnerabilities (5 low, 7 moderate, 3 high)

This was done via `npm i webflow-api@latest` in `shared/` and in `scripts/`.

No files were manually edited.

Unit and integration tests ran.

---

The dependency bump pins `webflow-api` at version 3.2.1 and includes a refreshed root lockfile so the patched `qs` range and its related transitive updates—such as `crypto-browserify`, `js-base64`, and `node-fetch`—are captured.

---

### DIFF
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -49,7 +49,7 @@
         "simple-oauth2": "5.1.0",
         "undici": "6.21.2",
         "uuid": "9.0.0",
-        "webflow-api": "^3.2.1",
+        "webflow-api": "3.2.1",
         "xml-crypto": "6.1.2",
         "zod": "4.0.5"
     },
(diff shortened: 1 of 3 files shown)


packages/shared/package.json
```diff
@@ -48,8 +49,8 @@
         "simple-oauth2": "5.1.0",
         "undici": "6.21.2",
         "uuid": "9.0.0",
+        "webflow-api": "^3.2.1",
```
marcindobry: Can we still pin the version to 3.2.1?

marcindobry (changes requested):

packages/shared/package.json
thw-nango: Done.

marcindobry (approved):

TBonnin (approved):

PR #5312, opened 2026-01-26
chore: deprecate sync type

TBonnin:
I am not planning to merge this PR before the replacement for sync type (ie: checkpoint) is ready but I wanted to have as many eyes as possible on the incoming changes and make sure I haven't forgotten any other places where the concept of sync type was exposed. cc @khaliqgant @kaposke 

---

**Deprecate legacy `syncType` references across SDK, CLI, and docs**

Adds deprecation annotations to all TypeScript surface areas that still expose `syncType`/`SyncJobsType`, ensuring consumers know the field is slated for removal while runtime behavior stays unchanged. The CLI now injects a default `sync_type: 'full'` when scripts omit it, the Integrations UI no longer renders the badge, and documentation/OpenAPI spec call out the deprecation (though several JSON examples now contain invalid inline comments).

<details>

• Tagged `syncType`, `SyncJobsType`, and related DTO fields across `packages/types`, runner SDK, node client, and webhook payload contracts with `@deprecated` JSDoc comments.
• Updated `packages/cli/lib/zeroYaml/definitions.ts` so generated zero YAML definitions default `sync_type` to `'full'` when not explicitly provided, matching legacy behavior.
• Removed the `Sync type` badge from `packages/webapp/src/pages/Integrations/providerConfigKey/Functions/One.tsx` so the UI no longer surfaces the deprecated concept.
• Documentation (implementation guides, SDK reference, functions reference, OpenAPI spec) now explicitly labels `syncType` as deprecated.

</details>

<details>

• Documentation JSON snippets now include `// DEPRECATED`, which breaks copy/paste usage.
• Defaulting unspecified `sync_type` values to `'full'` in CLI output could hide incremental configurations unless authors consciously override.

</details>

---


packages/shared/lib/models/Sync.ts
```diff
@@ -8,6 +8,7 @@ export enum SyncStatus {
     ERROR = 'ERROR'
 }
 
+/** deprecated **/
```
marcindobry: Are you also deprecating the other things in the enum?

packages/types/lib/sync/index.ts
```diff
@@ -1,5 +1,6 @@
 export type SyncStatus = 'RUNNING' | 'PAUSED' | 'STOPPED' | 'SUCCESS' | 'ERROR';
 
+/** @deprecated **/
```
marcindobry, line 5: ditto: there are non-sync types in this enum

TBonnin: `SyncJobsType` types are internal only. The deprecated tag is more an indication for us to not build more features on top of it. Those types must be kept though for backward compatibility :(

docs/implementation-guides/use-cases/syncs/realtime-syncs.mdx
```diff
@@ -61,8 +61,6 @@ export default createSync({
 });
 ```
 
-After processing an external webhook, Nango sends [a sync finished webhook](/implementation-guides/platform/webhooks-from-nango#sync-webhooks) to your app with `"syncType": "WEBHOOK"`.
```
kaposke, line 64: Is it not the case anymore?

kaposke (approved):
I think you got it covered. Only instance I see is `packages/webapp/src/types.ts`, which I just realized is on the frontend and should probably be refactored anyway.

docs/implementation-guides/use-cases/syncs/realtime-syncs.mdx
TBonnin: It is still gonna be there for backward compatibility. just not documented

PR #5314, opened 2026-01-27
feat(multi-api-secret): [BIG] Deprecate environment's secret_key.

thw-nango:
I apologize for the size of the change. It's the smallest cohesive unit of work I managed to clump apart from the rest of the codebase.

In this change, we:
 - Rename all 'secret_key' and 'pending_secret_key' columns in the `_nango_environments` table to 'deprecated_*'. In preparation for their ultimate removal.
 - Switch to using `api_secrets` instead. Also for pending secrets during secret rotation.
 - Pipe the used secret through auth middleware and into `res` local implicits.
 - Introduce a secret service + integration tests.
 - Clean up account service, environment service and secret service to expose tighter, cohesive postconditions.
 - Make better use of transactions and read/write segregation therein.

In my last change, we enforced a 1:1 correspondence between environments and (default) secrets.

With this change, we add an optional second non-default secret per environment. It is analogous to what column `_nango_environments.pending_secret_key` was: A secret in mid-rotation.

We preserve the fields `secret_key` and `pending_secret_key` in `DBEnvironment` for now, as they are used all throughout the codebase. Instead of fetching them from the env table, we set them by hand.

---

Environment retrieval now hydrates the legacy `DBEnvironment.secret_key` fields from decrypted entries in `api_secrets`, rotations and promotions flow entirely through the new secret service, account/context lookups join against `api_secrets` while caching PBKDF2 hashes, and enabling encryption re-hashes legacy secrets while expanded Result helpers provide safer optional-value handling.

---


packages/database/lib/migrations/20260127102000_deprecate_secret_key_in_env.cjs
```diff
@@ -0,0 +1,32 @@
...
+    await knex.raw(`
+        ALTER TABLE _nango_environments RENAME COLUMN deprecated_secret_key             TO secret_key;
+        ALTER TABLE _nango_environments RENAME COLUMN deprecated_secret_key_iv          TO secret_key_iv;
+        ALTER TABLE _nango_environments RENAME COLUMN deprecated_secret_key_tag         TO secret_key_tag;
+        ALTER TABLE _nango_environments RENAME COLUMN deprecated_secret_key_hashed      TO secret_key_hashed;
+        ALTER TABLE _nango_environments RENAME COLUMN deprecated_pending_secret_key     TO pending_secret_key;
+        ALTER TABLE _nango_environments RENAME COLUMN deprecated_pending_secret_key_iv  TO pending_secret_key_iv;
+        ALTER TABLE _nango_environments RENAME COLUMN deprecated_pending_secret_key_tag TO pending_secret_key_tag;
+    `);
+};
```
TBonnin: do we need to rename? can we stop using those columns and then remove them all together?

If the answer is 'yes we must rename', we are gonna need to do it in a two-steps deploy because some of the live instances are gonna expect either the old or new names because of the rollout strategy.

packages/server/lib/webhook/gmail-webhook-routing.ts
```diff
@@ -64,7 +64,7 @@ export async function validate(integration: IntegrationConfig, headers: Record<s
             return false;
         }
 
-        const environment = await environmentService.getRawById(integration.environment_id);
+        const environment = await environmentService.getById(integration.environment_id);
```
TBonnin, line 67: why this change?

packages/shared/lib/services/environment.service.ts
```diff
@@ -377,14 +390,40 @@ class EnvironmentService {
...
-        return key;
+    private async setAllSecrets(trx: Knex, envs: DBEnvironment[]) {
+        // Precondition: `envs` contains no duplicates.
+
+        // Note: For now, exactly one default secret per environment exists
+        // and zero or one non-default secret: The pending secret (during rotation).
+        const envByID = new Map(envs.map((env) => [env.id, env]));
+        const allSecrets = await secretService.getAllSecretsForAllEnvs(trx, envByID.keys().toArray());
+        for (const [envId, secrets] of allSecrets) {
+            const env = envByID.get(envId)!;
```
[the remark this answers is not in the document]
thw-nango: Fixed in new commit.

packages/persist/lib/middleware/auth.middleware.ts
```diff
@@ -51,6 +52,7 @@ export const authMiddleware = async (req: Request, res: Response<any, AuthLocals
 
         res.locals['account'] = accountContext.account;
         res.locals['environment'] = accountContext.environment;
+        res.locals['secret'] = accountContext.secret;
```
TBonnin: why do we need to store the secret in memory?

packages/shared/lib/services/environment.service.ts
```diff
@@ -43,19 +44,20 @@ class EnvironmentService {
...
-            const result = await db.knex.select('*').from<DBEnvironment>(TABLE).where({ id, deleted: false });
-
-            if (result == null || result.length == 0 || result[0] == null) {
+        return await db.readOnly.transaction(async (trx) => {
+            const env = await this.getByIdWithoutSecrets(trx, id);
+            if (!env) {
                 return null;
             }
+            await this.setSecretsOnEnv(trx, env);
+            return env;
```
TBonnin, line 53: why not a single query with a join instead of 2 queries in a transaction?

packages/shared/lib/services/environment.service.integration.test.ts
```diff
@@ -47,48 +42,51 @@ describe('Environment service', () => {
             webhook_url_secondary: null,
             otlp_settings: null,
             deleted: false,
-            deleted_at: null
+            deleted_at: null,
+            // Deprecated fields:
```
TBonnin: let's use the jsdoc format `@deprecated`

packages/shared/lib/services/secret.service.ts
```diff
@@ -0,0 +1,138 @@
...
+import encryptionManager, { pbkdf2 } from '../utils/encryption.manager.js';
+import { NangoError } from '../utils/error.js';
+
+import type { DBAPISecret } from '@nangohq/types';
+import type { Knex } from 'knex';
+
+const API_SECRETS_TABLE = 'api_secrets';
+
+class SecretService {
+    public async getDefaultSecretForEnv(trx: Knex, envId: number): Promise<DBAPISecret> {
```
TBonnin: we prefer to return the explicit `Result` type when a function can fail to force caller to handle error

packages/shared/lib/services/secret.service.ts
```diff
@@ -0,0 +1,138 @@
...
+            iv: '',
+            tag: '',
+            hashed: await this.hashSecret(plainText),
+            is_default: isDefault
+        } satisfies Partial<DBAPISecret>;
+
+        const encrypted = encryptionManager.encryptAPISecret(secret);
+
+        const rows = await trx<DBAPISecret>(API_SECRETS_TABLE).insert(encrypted).returning('*');
+        if (!rows || rows.length === 0) {
```
TBonnin: you can do `if(rows?[0])`. It should also save you for using `!` below

packages/shared/lib/services/secret.service.ts
```diff
@@ -0,0 +1,138 @@
...
+                });
+            await trx<DBAPISecret>(API_SECRETS_TABLE)
+                .where({
+                    id: secretId,
+                    environment_id: secret.environment_id,
+                    is_default: false
+                })
+                .update({
+                    is_default: true,
+                    updated_at: new Date()
```
TBonnin: we haven't been consistent at all but let's use `knex.fn.now()` and use pg as the source of truth for times

packages/shared/lib/services/environment.service.ts
```diff
@@ -70,34 +72,32 @@ class EnvironmentService {
...
-
-        const encryptedEnvironment = await encryptionManager.encryptEnvironment({
-            ...environment,
-            secret_key_hashed: await hashSecretKey(environment.secret_key)
-        });
-
-        await trx.transaction(async (trx) => {
-            await trx<DBEnvironment>(TABLE).where({ id: environment.id }).update(encryptedEnvironment);
-            await encryptionManager.upsertDefaultAPISecret(trx, encryptedEnvironment);
+        return trx.transaction(async (trx) => {
```
TBonnin, line 86: trx can already be a transaction. I wonder if we want to come up with a solution to avoid nesting transaction

packages/server/lib/formatters/environment.ts
```diff
@@ -2,15 +2,7 @@ import { getGlobalOAuthCallbackUrl } from '@nangohq/shared';
...
-export function environmentToApi({
-    secret_key_iv,
-    secret_key_tag,
-    secret_key_hashed,
-    pending_secret_key_iv,
-    pending_secret_key_tag,
-    pending_public_key,
-    ...env
-}: DBEnvironment): ApiEnvironment {
+export function environmentToApi(env: DBEnvironment): ApiEnvironment {
```
agusayerza, line 5: If we are renaming the fields we should still strip the `deprecated_*` fields from the API response

packages/shared/lib/services/secret.service.integration.test.ts
```diff
@@ -0,0 +1,96 @@
...
+        const env = await newEnv();
+        await expect(
+            secretService.createSecret(db.knex, {
+                environmentId: env.id,
+                isDefault: true
+            })
+        ).rejects.toThrow();
+    });
+
+    it('can create multiple non-default secrets', async () => {
```
agusayerza: Shouldn't we enforce that only one non-default secret exists at any given time?

packages/database/lib/migrations/20260127102000_deprecate_secret_key_in_env.cjs
marcindobry: I agree with @TBonnin here, the value in renaming is questionable and isn't as simple to pull off. We can just... not use them, communicate that via a comment for now and we are a small enough team that there if anyone decides to use them in some upcoming PR, we can quickly catch that. I'm pretty sure we could also just alias the fields in Typescript with deprecated_ prefix and it should achieve the same goal.

packages/server/lib/formatters/environment.ts
marcindobry: Yup, I think this is used by the public API so it would be a breaking change.

packages/shared/lib/services/environment.service.ts
```diff
@@ -234,20 +245,26 @@ class EnvironmentService {
...
+    async rotateSecretKey(envId: number): Promise<string | null> {
+        const created = await db.knex.transaction(async (trx) => {
+            const environment = await this.getByIdWithoutSecrets(trx, envId);
+            if (!environment) {
+                trx.rollback();
+                return null;
+            }
+            // Note: For now, we enforce the invariant that only one non-default API secret
+            // can exist at a time: the 'pending' secret, during rotation.
+            await trx<DBAPISecret>('api_secrets').delete().where({
```
marcindobry, line 261: Does this throw an exception if it fails? If not, then would we need to handle it before we create a new secret?

packages/shared/lib/services/secret.service.integration.test.ts
marcindobry: Yup I agree that enforcement should live at the service level and not just an assumption baked in some parts of the code

packages/database/lib/migrations/20260127102000_deprecate_secret_key_in_env.cjs
thw-nango: That's a very good observation. Let's strategize. If we keep the columns intact, we have to make sure that we manually delete the properties from any object obtained from a database query on `_nango_environments`, to make sure that no code anywhere depends on these properties anymore. It would mean adding more code, but we can delete it together with the columns in the next PR. WDYT?

packages/server/lib/webhook/gmail-webhook-routing.ts
thw-nango: `environmentService` had an (implicit) invariant that environments are always returned unencrypted from its public methods. `getRawById` was a private method used in `environmentService` that returned a still encrypted environment. At some point it was marked public for testing database encryption.

This webhook took advantage of that and used it as public API, even though it's not.

packages/persist/lib/middleware/auth.middleware.ts
thw-nango: Great catch, we can actually exclude this from this PR. The only code that needs it was excluded as well. Will do.

packages/shared/lib/services/environment.service.ts
thw-nango: Good question! 

In `environmentService`, there's 5 different queries that retrieve environments (sometimes one, sometimes many) with different strategies.

I considered adding the two joins to each of them:
 - Advantages: single query per retrieval.
 - Disadvantages: many more lines of repeating code.

IMHO adding the joins would equate to a premature optimization. This is more readable. WDYT?

packages/shared/lib/services/environment.service.ts
thw-nango: Yeah, it would be nice to have a `ensureTransaction` func that only starts a tx if necessary. Lucky for us, Postgres doesn't actually support nested transactions (none of the knex-supported DBs do).

With Postgres, knex wil emit `savepoint`s, in this scenario. Which is nice, because it allows the tx to flow with the code: If I `throw` from the nested block, the parent has the option to catch and continue -- only the inner savepoint will be rolled back.

With `ensureTransaction`, the only 100% safe option is to never catch. Doing so can lead to some very unintuitive behavior (inner statements survive the throw).

packages/shared/lib/services/environment.service.ts
thw-nango: Yes, all knex queries throw when they fail. The 'curious' results handling we see throughout comes from either AI or somebody who didn't read the manual.

packages/shared/lib/utils/encryption.manager.ts
```diff
@@ -265,18 +228,14 @@ export class EncryptionManager extends Encryption {
...
-
-            environment = await this.encryptEnvironment(environment);
-            await db.knex.transaction(async (txr) => {
-                await txr<DBEnvironment>(`_nango_environments`).where({ id: environment.id }).update(environment);
-                await this.upsertDefaultAPISecret(txr, environment);
-            });
+            const encrypted = this.encryptAPISecret(secret);
+            encrypted.updated_at = new Date();
+            await db.knex<DBAPISecret>(`api_secrets`).where({ id: secret.id }).update(encrypted);
         }
```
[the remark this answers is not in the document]
thw-nango: Done.

packages/persist/lib/middleware/auth.middleware.ts
thw-nango: Done.

packages/shared/lib/services/secret.service.ts
thw-nango: Done (here and further down).

packages/shared/lib/services/secret.service.ts
thw-nango: Done.

packages/persist/lib/middleware/auth.middleware.ts
```diff
@@ -3,12 +3,13 @@ import tracer from 'dd-trace';
...
 import { flagHasPlan, stringifyError, tagTraceUser } from '@nangohq/utils';
 
-import type { DBEnvironment, DBPlan, DBTeam } from '@nangohq/types';
+import type { DBAPISecret, DBEnvironment, DBPlan, DBTeam } from '@nangohq/types';
 import type { NextFunction, Request, Response } from 'express';
 
 export interface AuthLocals {
     account: DBTeam;
     environment: DBEnvironment;
+    secret: DBAPISecret;
```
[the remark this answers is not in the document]
thw-nango: Done.

packages/server/lib/middleware/access.middleware.ts
```diff
@@ -23,7 +23,7 @@ import { connectSessionTokenPrefix, connectSessionTokenSchema } from '../helpers
 import * as connectSessionService from '../services/connectSession.service.js';
 
 import type { RequestLocals } from '../utils/express.js';
-import type { ConnectSession, DBEnvironment, DBPlan, DBTeam, InternalEndUser } from '@nangohq/types';
+import type { ConnectSession, DBAPISecret, DBEnvironment, DBPlan, DBTeam, InternalEndUser } from '@nangohq/types';
```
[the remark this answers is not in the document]
thw-nango: Done.

packages/shared/lib/services/secret.service.ts
thw-nango: I am with you. I would love to only use `Result` everywhere.

This particular scenario lends itself poorly to the idea:
 - It's DB code: an exception rolls back the ambient tx. If we returned `Result` we would need to be very careful calling `trx.rollback()` at all possible failure points (but never twice because that itself would throw an exception).
 - Knex itself throws when its queries fail. We would have to wrap everything in try-catch blocks.
 - In our own `throw` below, we signal an invariant violation: This is not a recoverable condition. We want it to be loud and abortive, never swallowed.

It's the same distinction that Rust & Go make between returning an error, and `panic`. But our predicament is worse, since most JS/TS libs throw exceptions, and we build on 2640 of them (`npm ls --all --parseable | wc -l`).

packages/database/lib/migrations/20260127102000_deprecate_secret_key_in_env.cjs
TBonnin: > to make sure that no code anywhere depends on these properties anymore.

We are not really solving the issue by renaming those columns though. We still need to ensure that no code relies on the old column anyway. Renaming will make breakage smore obvious but the objective is not to break anything.
I understand it is challenging since `environment.service` is one of the oldest piece of code in the codebase but I don't think we can avoid it.

packages/server/lib/webhook/gmail-webhook-routing.ts
TBonnin: why do we need an unencrypted environment here is more my question

packages/shared/lib/services/environment.service.ts
TBonnin: > This is more readable
That's debatable ;)

Good point though about the 5 differents queries. Would it make sense to take the time to consolidate all of them (and include the joins) 
```
private function selectEnvironmentBaseQuery(db: knex.Knex): Knex.QueryBuilder<...> {
   return db.knex
      .from<DBEnvironment>(TABLE)
      .join(SECRETS_TABLE, ...)
      .leftJoin(...)
}
```
Then it can be used in all differents `get` functions

packages/shared/lib/services/environment.service.ts
TBonnin: an `ensureTransaction` wrapper would be nice. Let's add it to the `utils` package.

> the only 100% safe option is to never catch

I am not sure I understand

packages/shared/lib/services/environment.service.ts
TBonnin: > The 'curious' results handling we see throughout comes from either AI or somebody who didn't read the manual.

Can you explain what you mean here?

packages/server/lib/formatters/environment.ts
TBonnin: I don't think the secret related fields where returned by the API anyway

packages/shared/lib/services/account.service.ts
```diff
@@ -136,6 +130,7 @@ class AccountService {
             const result = await trx.from<DBTeam>(`_nango_accounts`).insert({ name: teamName, found_us: foundUs }).returning('*');
 
             if (!result[0]) {
+                trx.rollback();
```
TBonnin, line 133: I am not sure why we have a transaction here since it is a single query.

packages/shared/lib/services/secret.service.ts
TBonnin: > Knex itself throws when its queries fail. We would have to wrap everything in try-catch blocks.

Exception would need to be try-catched somewhere anyway. Our principle is to make it happen as early as possible and make the error explicit so the caller is forced to handle or forward the error.

> In our own throw below, we signal an invariant violation: This is not a recoverable condition. We want it to be loud and abortive, never swallowed.

I am not proposing to swallow the error but to make it explicit (vs implicit exception)

> It's DB code: an exception rolls back the ambient tx. If we returned Result we would need to be very careful calling trx.rollback() at all possible failure points (but never twice because that itself would throw an exception).

It it true that one must be careful when handling failed Result in transaction. But between 
- a world where all errors are implicit and can be swallowed 
- a world where everything must return a Result and errors must handled explicitly but one need to be careful when using transactions
I am pretty convinced the second one will lead to less bugs overtime.

Ideally this problem would be solved by a wrapper to make knex transactions play well with Result.

packages/server/lib/webhook/gmail-webhook-routing.ts
thw-nango: We don't. Never did. No idea why they called this here.

packages/shared/lib/services/environment.service.ts
thw-nango:
```
function a(db: Knex) {
    return ensureTransaction(db, trx => {
        try {
            return b(trx);
        } catch (e) {
            // Problem: b's write is still in tx.
            return trx.readQuery(...);
        }
    });
}

function b(db: Knex) {
    return ensureTransaction(db, trx => {
        db.writeQuery(...); // <- The problematic write.
        return somethingThatThrows();
    })
}
```

Without savepoints, the exception thrown in `b` can't undo `b`'s effects when bubbling up the call stack. This'll lead to two unexpected consequences:
 - In `a`'s catch branch, `b`'s write is still in the tx, and it will be applied when `a` returns.
 - `a` will observe `b`s write in its own reads.

packages/shared/lib/services/environment.service.ts
thw-nango: I mean [these kinds of checks](https://github.com/NangoHQ/nango/blob/f6621ee4cae9d1cdd7ed4c4c73bbbe9ef7c7a7b1/packages/shared/lib/services/environment.service.ts#L75-L77) that you can find all over the legacy code.

In the linked case, `result` can never be `null` (nor anything that coerces to `null` via `==`). The elements of `result` can also never be `null` either. The result is guaranteed to be an array. The only check that makes sense here is the `length` check, which is still done with loose comparison `==`. What _can_ happen, tho, is an exception. And that's not handled.

The exact checks performed vary throughout the codebase.

packages/shared/lib/services/secret.service.ts
thw-nango: I'm thinking what if we had a utility function to catch exceptions and turn them into `Result<T, E>`, e.g something like:
```
async function catchToResult(f : async () => T) : Promise<Result<T>> {
    try {
        return Ok(f())
    } catch (e) {
        return Err(e)
    }
}
```
This would allow us to wrap our transactional code blocks such that exceptions can still be relied on to roll back, but the return type is a `Result`:
```
async function myDBFunc(trx: Knex) : Promise<Result<DBType>> {
    return catchToResult(async () => {
        ...
    })
}
```
The reason is that I'm afraid that manually rolling back in every failure branch is very easy to get wrong and can have durable consequences.

TBonnin: > The reason is that I'm afraid that manually rolling back in every failure branch is very easy to get wrong and can have durable consequences.

The way we currently do it is re-throwing when `isErr` and trycatch at the outer level. Still error prone I agree.
I was therefore more thinking about a transaction specific wrapper to make Knex and Result play better together. Something like
```ts
async function withTransaction<T, E extends Error>(
  knex: Knex,
  fn: (trx: Knex.Transaction) => Promise<Result<T, E>>
): Promise<Result<T, E>> {
  try {
    const value = await knex.transaction(async (trx) => {
      const result = await fn(trx);

      if (result.isErr()) {
        throw result.error;
      }

      return result.value;
    });

    return Ok(value);
  } catch (e) {
    return Err(e);
  }
}

// and instead of using
db.transaction(async (trx) => { ... } )
// we use 
withTransaction(db, (trx) => { ... })

// could even be combine with the `ensureTransaction` concept we talked about earlier this week
```

packages/shared/lib/services/environment.service.ts
TBonnin: indeed those impossible or redundant checks should be removed. 
Feel free to fix when you encounter them

packages/shared/lib/services/secret.service.ts
thw-nango: I spent some time on this today. As I was "resultifying" my code I noticed some problems with the implementation of `Result`:

Wrong [return type on `Left.Map`](https://github.com/NangoHQ/nango/blob/7152d1e7891aee1c1812d54b35dbe5885d01939b/packages/types/lib/result.ts#L10):
```
map<U>(fn: (value: T) => U): Result<T, E>;
```
should be
```
map<U>(fn: (value: T) => U): Result<U, E>; // <- U is the new value type, not T.
```
But actually, as the [implementor correctly noted](https://github.com/NangoHQ/nango/blob/7152d1e7891aee1c1812d54b35dbe5885d01939b/packages/utils/lib/result.ts#L11-L17), `fn` can throw. So really it should be:
```
map<U>(fn: (value: T) => U): Result<U, Error>; // Widest Error type because fn can throw.
```
(And since every `map`/`mapError` op will widen the error type to `Error`, it begs the question whether it makes sense to keep an `E` in the signature at all.)

However, there's already callers depending on the broken type signatures. This is looking more like a cleanup task to me. WDYT?

thw-nango: I have implemented a half-fix, half feature commit. Check it out.

Still, this needs to be cleaned up across the codebase.

Edit: I actually broke the [Orchestrator client](https://github.com/NangoHQ/nango/blob/d45f08ede0ae205816a4d26960918fe97873574a/packages/orchestrator/lib/clients/client.ts#L68-L72) with this, because it violates the `Result` API by doing:
```
return Err({
    name: ...,
    message: ...,
    payload: ...,
});
```
Nineteen times. I don't know how this code ever passed type checking. But it's in there and I guess we need to either clean that up too or support it as a special case. Headache.

Edit 2: OK, I see how it passed type-checking. There's at least 3 different "standard" definitions of `Error`: ES2022's interface, ES5's interface and the builtin Error class. I guess the OG author's intention was to target specifically ES5's interface. So let's do that explicitly.

Edit3: Fixed with explicit ES5-targeting.

packages/shared/lib/services/account.service.ts
thw-nango: It's not a single query. See `createDefaultEnvironments` and `createPlan` immediately below.

packages/shared/lib/services/secret.service.integration.test.ts
thw-nango: Done.

packages/database/lib/migrations/20260127102000_deprecate_secret_key_in_env.cjs
thw-nango: Done.

packages/shared/lib/services/environment.service.integration.test.ts
thw-nango: Obsolete, because we decided against renaming.

packages/server/lib/formatters/environment.ts
thw-nango: Done.

packages/shared/lib/services/account.service.ts
```diff
@@ -264,8 +273,8 @@ class AccountService {
         if ('secretKey' in opts) {
             // Hashing is slow by design so it's very slow to recompute this hash all the time
             // We keep the hash in-memory to not compromise on security if the db leak
-            hash = hashLocalCache.get(opts.secretKey) || (await hashSecretKey(opts.secretKey));
-            q.where('secret_key_hashed', hash);
+            const hash = hashLocalCache.get(opts.secretKey) || (await secretService.hashSecret(opts.secretKey)).unwrap();
```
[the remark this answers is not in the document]
thw-nango: Done.

packages/shared/lib/services/environment.service.ts
thw-nango: I gave it a go and ran into the realization that it's not that simple. Going this route we would need to do all the same logic we do in `getAccountContext` (as it exists in this PR) for each of the 5 queries, meaning:
```
baseQuery().select(
    db.knex.raw('row_to_json(_nango_environments.*) as environment'),
    db.knex.raw('row_to_json(default_secret.*) as default_secret'),
    db.knex.raw('row_to_json(pending_secret.*) as pending_secret'),
)
```
Then type-annotate every specific case, pick the right fields out, decrypt the secrets manually, unset the deprecated fields, overwrite with the decrypted ones. It's actually much, much more code.

packages/shared/lib/services/account.service.ts
```diff
@@ -264,8 +273,8 @@ class AccountService {
         if ('secretKey' in opts) {
             // Hashing is slow by design so it's very slow to recompute this hash all the time
             // We keep the hash in-memory to not compromise on security if the db leak
-            hash = hashLocalCache.get(opts.secretKey) || (await hashSecretKey(opts.secretKey));
-            q.where('secret_key_hashed', hash);
+            const hashed = hashLocalCache.get(opts.secretKey) || (await secretService.hashSecret(opts.secretKey)).unwrap();
+            q.where('default_secret.hashed', hashed);
```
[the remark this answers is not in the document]
thw-nango: Wrong interpretation of the rotation mechanic.

packages/shared/lib/services/environment.service.ts
TBonnin: ok. let's not do this then

packages/shared/lib/services/account.service.ts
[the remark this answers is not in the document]
TBonnin: another of our conventions: we refrain to use unwrap outside of tests in favor or more verbose but more explicit error handling

packages/shared/lib/services/secret.service.ts
TBonnin: good point. Let's make another PR though since it is unrelated to this environment secrets change

TBonnin (approved):
2 last comments about:
- not using unwrap outside of test. This one I would like you to address it
- split the Result fix into another PR since it is not related to those changes. Up to you though

packages/shared/lib/services/secret.service.ts
thw-nango: Dear reviewer 😆 you insisted on this, here. As PR author, I serve at your approval. The `Result` changes are half the code volume of this PR now. I delivered what was requested. We both knew it was only tangentially related. But we pushed on. Going back and separating everything again is gonna equate to a lot of wasted cycles.

thw-nango:
> * not using unwrap outside of test. This one I would like you to address it.

Where I used unwrap is where exceptions were previously thrown. For example in:
```
async rotateSecretKey(envId: number): Promise<string | null>
async revertSecretKey(envId: number): Promise<string | null>
async getById(id: number): Promise<DBEnvironment | null>
```
Without also changing these method's signature, the other alternative is to swallow the error and return `null`.
I agree that we should change the signature of all of these methods to also return `Result`s. But not in this PR.
Let it rip as it has always ripped. And let's clean up legacy in a concerted effort over the next few PRs.
WDYS?

TBonnin:
> > * not using unwrap outside of test. This one I would like you to address it.
> 
> Where I used unwrap is where exceptions were previously thrown. For example in:
> 
> ```
> async rotateSecretKey(envId: number): Promise<string | null>
> async revertSecretKey(envId: number): Promise<string | null>
> async getById(id: number): Promise<DBEnvironment | null>
> ```
> 
> Without also changing these method's signature, the other alternative is to swallow the error and return `null`. I agree that we should change the signature of all of these methods to also return `Result`s. But not in this PR. Let it rip as it has always ripped. And let's clean up legacy in a concerted effort over the next few PRs. WDYS?

I am only talking about the usage of unwrap() in non-test code. I am fine not fixing the signature of existing methods in the PR. Please explicitly throw the error with `if(res.isErr()) throw ...` instead of using `unwrap()`. thank you

PR #5316, opened 2026-01-27
feat(checkpoint): add checkpoint data model

TBonnin:
It will be possible for customers to get/save checkpoint explicitely in function to allow customers to implement "incremental" syncs, track progress, etc...

See inlined comments to discuss design decision.

Note: code isn't wired anywhere yet. The goal of this PR is to validate the checkpoints data model

 

---

**Checkpoint persistence model and service layer**

Adds a dedicated `checkpoints` persistence layer (migration, types, service helpers) so future function executions can save and resume progress per environment/connection/key with optimistic locking semantics. Includes comprehensive validation plus unit and integration coverage, but the tables/services are not yet wired into runtime flows.

<details>

• Created migration `packages/database/lib/migrations/20260127120000_create_checkpoints_table.cjs` for the `checkpoints` table with `(environment_id, connection_id, key)` uniqueness, soft-delete support, version counter, and a concurrent `varchar_pattern_ops` index to accelerate prefix cleanup queries.
• Defined checkpoint type contracts in `packages/types/lib/checkpoint/db.ts` and re-exported them so other packages can consume `Checkpoint`, `CheckpointValue`, and `DBCheckpoint`.
• Implemented `validateCheckpoint`, `getCheckpoint`, `upsertCheckpoint`, `deleteCheckpoint`, and `hardDeleteCheckpoints` in `packages/shared/lib/services/checkpoints/checkpoints.ts`, enforcing flat payloads, size limits, and optimistic locking/versioning.
• Added extensive unit tests for validation rules plus integration tests covering upsert/update/resurrection, soft delete, hard delete (with prefix + escaping), and database interactions.
• Re-exported the checkpoint service via `packages/shared/lib/index.ts` to make the helpers available to other modules.

</details>

<details>

• `checkpoints` database table and migration
• `packages/shared/lib/services/checkpoints` service + tests
• `packages/types` exports
• `@nangohq/shared` public surface area

</details>

---

### DIFF
--- a/packages/database/lib/migrations/20260127120000_create_checkpoints_table.cjs
+++ b/packages/database/lib/migrations/20260127120000_create_checkpoints_table.cjs
@@ -1,0 +1,37 @@
+const CHECKPOINTS_TABLE = 'checkpoints';
+
+exports.config = { transaction: false };
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`
+        CREATE TABLE IF NOT EXISTS ${CHECKPOINTS_TABLE} (
+            id SERIAL PRIMARY KEY,
+            environment_id INTEGER NOT NULL REFERENCES _nango_environments(id) ON DELETE CASCADE,
+            key VARCHAR(255) NOT NULL,
+            checkpoint JSONB NOT NULL DEFAULT '{}',
+            version INTEGER NOT NULL DEFAULT 1,
+            created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
+            updated_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
+            deleted_at TIMESTAMP WITH TIME ZONE DEFAULT NULL,
+            UNIQUE(environment_id, key)
+        )
+    `);
+
+    // Index to optimize prefix searches on the 'key' column
+    // Note: The UNIQUE(environment_id, key) constraint already creates an index for exact lookups
+    // This is useful for queries that look for keys starting with a certain prefix
+    // e.g., WHERE key LIKE 'prefix%'
+    await knex.raw(`
+        CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_checkpoints_key_prefix
+            ON ${CHECKPOINTS_TABLE}
+            USING BTREE (environment_id, key varchar_pattern_ops)
+    `);
+};
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.down = async function () {};
--- a/packages/database/lib/migrations/20260127120000_create_checkpoints_table.cjs
+++ b/packages/database/lib/migrations/20260127120000_create_checkpoints_table.cjs
@@ -10,24 +10,25 @@ exports.up = async function (knex) {
         CREATE TABLE IF NOT EXISTS ${CHECKPOINTS_TABLE} (
             id SERIAL PRIMARY KEY,
             environment_id INTEGER NOT NULL REFERENCES _nango_environments(id) ON DELETE CASCADE,
+            connection_id INTEGER NOT NULL REFERENCES _nango_connections(id) ON DELETE CASCADE,
             key VARCHAR(255) NOT NULL,
             checkpoint JSONB NOT NULL DEFAULT '{}',
             version INTEGER NOT NULL DEFAULT 1,
             created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
             updated_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
             deleted_at TIMESTAMP WITH TIME ZONE DEFAULT NULL,
-            UNIQUE(environment_id, key)
+            UNIQUE(environment_id, connection_id, key)
         )
     `);
 
     // Index to optimize prefix searches on the 'key' column
-    // Note: The UNIQUE(environment_id, key) constraint already creates an index for exact lookups
+    // Note: The UNIQUE(environment_id, connection_id, key) constraint already creates an index for exact lookups
     // This is useful for queries that look for keys starting with a certain prefix
     // e.g., WHERE key LIKE 'prefix%'
     await knex.raw(`
         CREATE INDEX CONCURRENTLY IF NOT EXISTS idx_checkpoints_key_prefix
             ON ${CHECKPOINTS_TABLE}
-            USING BTREE (environment_id, key varchar_pattern_ops)
+            USING BTREE (environment_id, connection_id, key varchar_pattern_ops)
     `);
 };
 
--- a/packages/shared/lib/services/checkpoints/checkpoints.ts
+++ b/packages/shared/lib/services/checkpoints/checkpoints.ts
@@ -39,14 +39,13 @@ export function validateCheckpoint(data: unknown): Result<Checkpoint> {
 
 /**
  * Get a checkpoint by environment ID and key.
- * Only returns non-deleted checkpoints.
  * @param environmentId - The environment ID the checkpoint belongs to.
  * @param key - The key of the checkpoint.
  * @returns The checkpoint if found, or null if not found.
  */
 export async function getCheckpoint(db: Knex, { environmentId, key }: { environmentId: number; key: string }): Promise<Result<DBCheckpoint | null>> {
     try {
-        const result = await db.from<DBCheckpoint>(TABLE).where({ environment_id: environmentId, key }).whereNull('deleted_at').first();
+        const result = await db.from<DBCheckpoint>(TABLE).where({ environment_id: environmentId, key }).first();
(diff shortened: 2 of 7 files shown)


packages/database/lib/migrations/20260127120000_create_checkpoints_table.cjs
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`
+        CREATE TABLE IF NOT EXISTS ${CHECKPOINTS_TABLE} (
+            id SERIAL PRIMARY KEY,
+            environment_id INTEGER NOT NULL REFERENCES _nango_environments(id) ON DELETE CASCADE,
+            key VARCHAR(255) NOT NULL,
```
TBonnin, line 14: I went for an extensible approach with a key string. We could start with `sync_id` but a key will allow us to add support for actions or other kind of functions. It would also allow for named checkpoint, etc...

packages/database/lib/migrations/20260127120000_create_checkpoints_table.cjs
```diff
@@ -0,0 +1,37 @@
...
+            checkpoint JSONB NOT NULL DEFAULT '{}',
+            version INTEGER NOT NULL DEFAULT 1,
```
TBonnin, line 16: version is used for optimistic locking. saving checkpoint would be done from the runner over the network and even the orchestrator guarantee that only one execution of a sync is being run at a time I wanted to have a mechanism to prevent race conditions built into checkpoints so its use-case can be extended without having to re-evaluate the assumption about how functions are executed

packages/shared/lib/services/checkpoints/checkpoints.ts
```diff
@@ -0,0 +1,167 @@
...
+const MAX_FIELDS = 16;
+
+/**
+ * Zod schema for checkpoint validation.
+ * - Keys: max 255 characters
+ * - String values: max 255 characters
+ * - Date values: converted to ISO string
+ * - Max 16 fields
+ */
+const checkpointValueSchema = z.union([z.string().max(MAX_STRING_VALUE_LENGTH), z.number(), z.boolean(), z.date().transform((d) => d.toISOString())]);
```
TBonnin, line 22: I went for a very constrained format at first (16 keys max, string/number/date/boolean values only, no nesting). 
We can be more permissive later based on how customers are gonna use checkpoints but it is easier to go from less permissive to more permissive than to restrict usage.

packages/shared/lib/services/checkpoints/checkpoints.ts
```diff
@@ -0,0 +1,167 @@
...
+        }
+        return Ok(undefined);
+    } catch (err) {
+        return Err(new Error('failed_to_delete_checkpoint', { cause: err }));
+    }
+}
+
+/**
+ * Hard delete all checkpoints matching a key prefix (e.g., all checkpoints for a connection).
+ * This does not use optimistic locking as it's typically used for cleanup operations.
```
TBonnin, line 153: The idea is this function would only be used by the background cleanup process that is hard-deleting connection after 30 days.

packages/types/lib/checkpoint/db.ts
```diff
@@ -0,0 +1,40 @@
...
+ *
+ * @example Key formats:
+ * - `connection:123:config:456` - checkpoint for a specific sync/action per connection
+ */
+export interface DBCheckpoint extends Timestamps {
+    id: number;
+    environment_id: number;
+    key: string;
+    checkpoint: Checkpoint;
+    version: number;
```
TBonnin, line 39: Only the internal `Checkpoint` data will be exposed to customers in functions. The version dance will be handled transparently by the `runner-sdk`

packages/shared/lib/services/checkpoints/checkpoints.ts
```diff
@@ -0,0 +1,167 @@
...
+            INSERT INTO ${TABLE} (environment_id, key, checkpoint, version, created_at, updated_at)
+            VALUES (?, ?, ?::jsonb, 1, NOW(), NOW())
+            ON CONFLICT (environment_id, key)
+            DO UPDATE SET
+                checkpoint = EXCLUDED.checkpoint,
+                version = ${TABLE}.version + 1,
+                deleted_at = NULL,
+                updated_at = NOW()
+            WHERE
+                ${TABLE}.version = ?
```
agusayerza, line 104: Not sure if this will be important, but a soft delete blocks the key from being reused.

Let's say we have a checkpoint on version 99 and we soft delete it.
We set the `deleted_at` flag. Then the user attempts to re-use the same key and calls `saveCheckpoint({state: newState})` 
I assume the key will resolve to the same value, something to do with the integration/action running.

1. The `INSERT` conflicts
2. The `UPDATE` fails because version (99) != expectedVersion (null).

We can obviously handle this case before calling `upsertCheckpoint`, but we would need to check againts soft deleted rows, which seems a bit weird if you ask me

I think we can handle this at a db level:
```suggestion
                ${TABLE}.version = ?
            OR
                (
                    ${TABLE}.deleted_at IS NOT NULL
                AND  ? IS NULL)
```

packages/database/lib/migrations/20260127120000_create_checkpoints_table.cjs
marcindobry: I'm a bit concerned about this being too flexible and then us actually encoding extra data into a single string and having to parse it. For example, for named checkpoints we'd need to prefix/suffix the name with e.g. sync_id so that there are no conflicts when people name checkpoints the same thing in multiple functions. For now it should be fine though.

marcindobry (approved):

packages/shared/lib/services/checkpoints/checkpoints.ts
```diff
@@ -0,0 +1,167 @@
...
+    try {
+        // Escape special characters for SQL LIKE query
+        // % and _ are wildcards, \ is the escape character
+        const escapedPrefix = keyPrefix.replace(/[%_\\]/g, '\\$&');
+
+        const count = await db
+            .from<DBCheckpoint>(TABLE)
+            .where('environment_id', environmentId)
+            .whereRaw("key LIKE ? ESCAPE '\\'", [`${escapedPrefix}%`])
+            .delete();
```
kaposke: Solid :)

packages/database/lib/migrations/20260127120000_create_checkpoints_table.cjs
kaposke: Is there a case where it wouldn't be unique per function and connection?

kaposke (commented):
Code looks great. Only open question is about the key being super flexible.

packages/types/lib/checkpoint/db.ts
```diff
@@ -0,0 +1,40 @@
+import type { Timestamps } from '../db.js';
+
+/**
+ * Allowed types for checkpoint values.
+ * Only flat key-value structures are supported (no nested objects or arrays).
+ * Date values are automatically converted to ISO strings when stored.
+ */
+export type CheckpointValue = string | number | boolean | Date;
```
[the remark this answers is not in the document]
agusayerza: I think the bot makes a pretty compeling argument

```
await nango.saveCheckpoint({ last_updated: new Date() }); 

// User retrieves it
const state = await nango.getCheckpoint();
if (state.last_updated.getTime() < ...) // Fails as last_updated is now a string
```

packages/database/lib/migrations/20260127120000_create_checkpoints_table.cjs
TBonnin: > Is there a case where it wouldn't be unique per function and connection

For now no but a function can have different shapes and be referenced by different kind of identifiers, hence the flexible string.
The connection part though could be an invariant and move outside of the key. Let me to that, I think it is a constraint we can live with, I don't really see a world where checkpoints are not bound to a specific connection.

> I'm a bit concerned about this being too flexible and then us actually encoding extra data into a single string and having to parse it

The goal is not to parse it but to use it as a reference key.

The only use-case for not querying a single key is to for cleanup and it still doesn't mean parsing but only prefix matching.

packages/types/lib/checkpoint/db.ts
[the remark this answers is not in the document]
TBonnin: That's a good point indeed.
My idea was to enforce types in the `runner-sdk` where we actually know the type of the internal Checkpoint.

packages/shared/lib/services/checkpoints/checkpoints.ts
TBonnin: The plan is to have the `runner-sdk` functions `get/SaveCheckpoints` take care of the version handling transparently and always pass the version regardless of the deleted status of the checkpoint. So a soft deleted checkpoint can be resurrected.
That make me realize though we need to return the soft deleted checkpoint. I will fix.
Thank you for the pointer

TBonnin:
pushed a commit to make `connection_id` an explicit field of `checkpoints`. Let me know if you have any additional concerns. 
This code isn't wired yet so I can address anything comment in next PRs

kaposke (approved):

PR #5322, opened 2026-01-28
feat(jobs): Broadcast task cancellation to all active runners

agusayerza:
Cancellation only targeted a single runner, which failed when a task was still running on a different active runner (e.g., after key rotation).
Fan out abort requests to all active runners for the team’s routingId and succeed if any runner confirms the abort.

[NAN-4519](https://linear.app/nango/issue/NAN-4519/task-cancellation-must-broadcast-to-all-active-runners)

- Added test cases for new functionality
- Want to attempt to test it on an actual run

---

**Broadcast abort requests to every active runner**

This PR changes task cancellation so that abort requests are fanned out to every active runner registered under a team’s routing ID, ensuring cancellation succeeds even if execution has shifted to another node during rollouts. It introduces a shared `getRunners` helper that consults fleet state (or falls back to the legacy single-runner lookup) plus a fleet-level `getNodesByRoutingId` API to enumerate applicable nodes, and updates the runtime adapter and abort operation to treat success as soon as any runner confirms the abort.

<details>

• Added `getRunnerIdForTeam` and new `getRunners` helper in `packages/jobs/lib/runner/runner.ts` that queries the fleet for `RUNNING`/`OUTDATED` nodes (with remote/default fallbacks) and returns multiple `FleetRunner` clients.
• Extended `packages/fleet/lib/fleet.ts` with a `getNodesByRoutingId` method (including `NodeState` typing) to retrieve nodes matching a routing ID and optional state filter.
• Updated `packages/jobs/lib/execution/operations/abort.ts` and `packages/jobs/lib/runtime/runner.adapter.ts` to broadcast aborts via `Promise.allSettled`, treating the operation as successful if any runner fulfills with `true`, and added explanatory comments.
• Introduced targeted unit tests in `packages/jobs/lib/runner/runner.unit.test.ts` and `packages/jobs/lib/execution/operations/abort.unit.test.ts` covering multi-runner selection, fallbacks, and success/error cases.

</details>

<details>

• packages/jobs/lib/runner/runner.ts
• packages/jobs/lib/execution/operations/abort.ts
• packages/jobs/lib/runtime/runner.adapter.ts
• packages/fleet/lib/fleet.ts
• packages/jobs/lib/runner/runner.unit.test.ts
• packages/jobs/lib/execution/operations/abort.unit.test.ts

</details>

---

### DIFF
--- a/packages/fleet/lib/fleet.ts
+++ b/packages/fleet/lib/fleet.ts
@@ -16,7 +16,7 @@ import { withPgLock } from './utils/locking.js';
 
 import type { ImageVerifier } from './image-verifier/verifier.js';
 import type { NodeProvider } from './node-providers/node_provider.js';
-import type { Node, NodeConfigOverride } from './types.js';
+import type { Node, NodeConfigOverride, NodeState } from './types.js';
 import type { Deployment, ImageType, RoutingId } from '@nangohq/types';
 import type { Result } from '@nangohq/utils';
 import type knex from 'knex';
@@ -157,6 +157,30 @@ export class Fleet {
             return recurse(supervisor, start);
         };
         return recurse(this.supervisor, new Date());
+    }
+
+    public async getNodesByRoutingId({
+        routingId,
+        states = ['PENDING', 'STARTING', 'RUNNING', 'OUTDATED']
+    }: {
+        routingId: RoutingId;
+        states?: [NodeState, ...NodeState[]];
+    }): Promise<Result<Node[]>> {
+        const search = await nodes.search(this.dbClient.db, {
+            states,
+            routingId
+        });
+        if (search.isErr()) {
+            return Err(search.error);
+        }
+
+        const byState = search.value.get(routingId);
+        if (!byState) {
+            return Ok([]);
+        }
+
+        const nodesByState = Object.values(byState);
+        return Ok(nodesByState.flat());
     }
 
     public async registerNode({ nodeId, url }: { nodeId: number; url: string }): Promise<Result<Node>> {
--- a/packages/jobs/lib/execution/operations/abort.ts
+++ b/packages/jobs/lib/execution/operations/abort.ts
@@ -5,7 +5,7 @@ import { Err, Ok } from '@nangohq/utils';
 import { orchestratorClient } from '../../clients.js';
 import { envs } from '../../env.js';
 import { logger } from '../../logger.js';
-import { getRunner } from '../../runner/runner.js';
+import { getRunners } from '../../runner/runner.js';
 
 import type { TaskAbort } from '@nangohq/nango-orchestrator';
 import type { Result } from '@nangohq/utils';
@@ -38,12 +38,16 @@ export async function abortTask(task: TaskAbort): Promise<Result<void>> {
 export async function abortTaskWithId({ taskId, teamId }: { taskId: string; teamId: number }): Promise<Result<void>> {
     try {
         await setAbortFlag(taskId);
-        const runner = await getRunner(teamId);
-        if (runner.isErr()) {
-            return Err(runner.error);
+        // Broadcast abort to all runners as a task might still be running on a different active runner during/after rollouts (e.g., after key rotation).
+        const runners = await getRunners(teamId);
+        if (runners.isErr()) {
+            return Err(runners.error);
         }
-        const isAborted = await runner.value.client.abort.mutate({ taskId });
-        if (!isAborted) {
+
+        const results = await Promise.allSettled(runners.value.map((runner) => runner.client.abort.mutate({ taskId })));
+        const didAbort = results.some((result) => result.status === 'fulfilled' && result.value);
+
+        if (!didAbort) {
             return Err(`Error aborting script for task: ${taskId}`);
         }
         return Ok(undefined);
--- a/packages/jobs/lib/execution/operations/abort.ts
+++ b/packages/jobs/lib/execution/operations/abort.ts
@@ -38,7 +38,7 @@ export async function abortTask(task: TaskAbort): Promise<Result<void>> {
 export async function abortTaskWithId({ taskId, teamId }: { taskId: string; teamId: number }): Promise<Result<void>> {
     try {
         await setAbortFlag(taskId);
-        // Broadcast abort to all runners as a task might still be running on a different active runner during/after rollouts (e.g., after key rotation).
+        // Broadcast abort to all runners as a task might still be running on a different active runner during/after rollouts (e.g. deploying a new runner version).
         const runners = await getRunners(teamId);
         if (runners.isErr()) {
             return Err(runners.error);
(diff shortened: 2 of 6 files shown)


packages/fleet/lib/fleet.ts
```diff
@@ -26,6 +26,8 @@ const defaultDbUrl =
     envs.NANGO_DATABASE_URL ||
     `postgres://${encodeURIComponent(envs.NANGO_DB_USER)}:${encodeURIComponent(envs.NANGO_DB_PASSWORD)}@${envs.NANGO_DB_HOST}:${envs.NANGO_DB_PORT}/${envs.NANGO_DB_NAME}?application_name=${envs.NANGO_DB_APPLICATION_NAME}${envs.NANGO_DB_SSL ? '&sslmode=no-verify' : ''}`;
 
+const defaultNodeSearchStates: [NodeState, ...NodeState[]] = ['PENDING', 'STARTING', 'RUNNING', 'OUTDATED'];
```
TBonnin: looks like this cons is only used once. We can probably inline it

packages/fleet/lib/fleet.ts
```diff
@@ -159,6 +161,30 @@ export class Fleet {
...
+            return Err(search.error);
+        }
+
+        const byState = search.value.get(routingId);
+        if (!byState) {
+            return Ok([]);
+        }
+
+        const nodesByState = Object.values(byState);
+        return Ok(nodesByState.flat());
```
TBonnin, line 183: you can probably do a one liner but not sure it is more readable
```
return Ok(
   Object
       .values(search.value.get(routingId) || {})
       .flat()
);
```

packages/jobs/lib/execution/operations/abort.ts
```diff
@@ -38,12 +38,15 @@ export async function abortTask(task: TaskAbort): Promise<Result<void>> {
 export async function abortTaskWithId({ taskId, teamId }: { taskId: string; teamId: number }): Promise<Result<void>> {
     try {
         await setAbortFlag(taskId);
-        const runner = await getRunner(teamId);
-        if (runner.isErr()) {
-            return Err(runner.error);
+        const runners = await getRunners(teamId);
```
TBonnin, line 42: let's add a little comment to mention that abort tasks must be sent to all runners because we don't know which one is actually running the task during/after rollout

packages/jobs/lib/runner/runner.ts
```diff
@@ -45,6 +45,35 @@ export async function getRunner(teamId: number): Promise<Result<Runner>> {
     }
 }
 
+export async function getRunners(teamId: number): Promise<Result<Runner[]>> {
+    try {
+        const runnerId = isProd ? getRunnerId(`${teamId}`) : getRunnerId('default');
```
TBonnin: we can maybe consolidate this into a private function to use it here and in getRunner above

packages/jobs/lib/runner/runner.ts
```diff
@@ -45,6 +45,35 @@ export async function getRunner(teamId: number): Promise<Result<Runner>> {
...
+        if (envs.RUNNER_TYPE === 'REMOTE') {
+            const runner = await getOrStartRunner(runnerId).catch(() => getOrStartRunner(getRunnerId('default')));
+            return Ok([runner]);
+        }
+
+        const runnersFleet = getDefaultFleet();
```
TBonnin, line 59: here we assume the function is running on the default fleet and/or that execution will never be split across fleet.
Is it a fair assumption @rossmcewan ?

packages/jobs/lib/runner/runner.ts
```diff
@@ -45,6 +45,35 @@ export async function getRunner(teamId: number): Promise<Result<Runner>> {
...
+        if (nodes.isErr()) {
+            return Err(nodes.error);
+        }
+
+        const runners = nodes.value.filter((node) => node.url).map((node) => new FleetRunner(runnerId, node.url as string));
+        if (runners.length > 0) {
+            return Ok(runners);
+        }
+
+        const runner = await getOrStartRunner(runnerId).catch(() => getOrStartRunner(getRunnerId('default')));
```
TBonnin, line 73: do we need to default to the default runner in this case?

packages/jobs/lib/runner/runner.unit.test.ts
```diff
@@ -0,0 +1,74 @@
...
+            { id: 1, url: 'http://runner-a' },
+            { id: 2, url: 'http://runner-b' }
+        ] as any[];
+
+        (getDefaultFleet as unknown as { mockReturnValue: (value: any) => void }).mockReturnValue({
+            getNodesByRoutingId: vi.fn().mockResolvedValue(Ok(nodes)),
+            getRunningNode: vi.fn()
+        });
+
+        const result = await getRunners(123);
```
TBonnin, line 53: you'll need to unwrap the result

packages/jobs/lib/runtime/runner.adapter.ts
```diff
@@ -9,11 +9,12 @@ import type { Result } from '@nangohq/utils';
...
-            const runner = await getRunner(params.nangoProps.team.id);
-            if (runner.isErr()) {
-                return Err(runner.error);
+            const runners = await getRunners(params.nangoProps.team.id);
+            if (runners.isErr()) {
+                return Err(runners.error);
             }
-            const isAborted = await runner.value.client.abort.mutate({ taskId: params.taskId });
+            const results = await Promise.allSettled(runners.value.map((runner) => runner.client.abort.mutate({ taskId: params.taskId })));
+            const isAborted = results.some((result) => result.status === 'fulfilled' && result.value);
```
TBonnin: nit: this same variable was called `didAbort` in abort.ts

packages/jobs/lib/execution/operations/abort.ts
```diff
@@ -38,6 +38,7 @@ export async function abortTask(task: TaskAbort): Promise<Result<void>> {
 export async function abortTaskWithId({ taskId, teamId }: { taskId: string; teamId: number }): Promise<Result<void>> {
     try {
         await setAbortFlag(taskId);
+        // Broadcast abort to all runners as a task might still be running on a different active runner during/after rollouts (e.g., after key rotation).
```
TBonnin: Most common cause of rollout is us deploying a new version of the runner.

packages/fleet/lib/fleet.ts
agusayerza: I'd keep the current version. It’s a bit more readable with the guard and intermediate variables IMO

packages/jobs/lib/runner/runner.ts
agusayerza: This is the same assumption already used by getRunner/invoke. If we ever split execution across fleets, we’ll need a fleet‑aware lookup here. For now I would say it’s consistent with existing behavior.

packages/jobs/lib/runner/runner.ts
agusayerza: Kept the default fallback to prevent any regressions, as `getRunner` had it in place. Happy to remove if that is something we don't want.

TBonnin (approved):

PR #5337, opened 2026-01-29
fix: npm audit and install with node22/npm10

TBonnin:
node25/npm11 seems to resolve dependencies differently from node v22 resulting in a different package-lock.json.
This commit is the result of running npm audit fix and install with `node: v22.21.1` and `npm: 10.9.4`. Node v22 being the currently supported Nango version

---

It also captures the dependency updates—locking `hono` to 4.11.7, `tar` to 7.5.7—and records the npm 10 peer-only markers on the platform-specific `@rollup/rollup-*` artifacts so the lockfile mirrors what that runtime produces.

<details>

• Bumped locked dependency `node_modules/hono` from version `4.11.5` to `4.11.7` with new tarball URL and integrity hash.
• Updated `node_modules/tar` from version `7.5.6` to `7.5.7` to capture the npm audit fix.
• Added `peer": true` markers for each platform-specific `@rollup/rollup-*` entry to mirror npm 10’s metadata output.

</details>

---


thw-nango:
IIUC this fix will only work if all of us to install the same (major/release) version of `node` and `npm`. Otherwise the next PR modifying `package-lock.json` has the potential to undo these changes silently. Is that correct?

TBonnin:
> IIUC this fix will only work if all of us to install the same (major/release) version of `node` and `npm`. Otherwise the next PR modifying `package-lock.json` has the potential to undo these changes silently. Is that correct?

we should indeed enforce the use of the version defined in `.nvmrc`. Let me see what can be done

Edit: so enforcing node version can be done https://github.com/NangoHQ/nango/pull/5337/commits/216a777502121070da817f31a3a25c83e742b1b5 but we run the tests on newer versions so CI fails. I'll have to come back to that next week.

Edit 2: forcing the node version when installing isn't the right approach I think. Let's try to understand why there is a decrepency between running npm install on 22 vs 25. 

In the meantime let's merge the fix and figure out how to prevent someone or the publish script to break the package-lock.json silently

kaposke (approved):

marcindobry (approved):

PR #5341, opened 2026-01-30
feat(tags): Match the validation of end_user tags to new tags

marcindobry:
Make sure end_user.tags have the same validation on almost all levels as the new tags so that we don't have issues during the mapping for customers who haven't transitioned to the new tags yet and might populate invalid data in the future.

The only difference really is that we won't lowercase the end user tags to keep existing behavior intact.

 

---

**Align `end_user.tags` validation with connection tag rules**

This PR centralizes tag validation logic so that legacy `end_user.tags` accept only the same key/ value patterns as the newer connection tags. A reusable helper now enforces key count limits, case-insensitive duplicate detection, and `end_user_email` formatting, and the server-side end-user schema reuses the shared key schema while deliberately omitting tag lowercasing to maintain historical behavior.

<details>

• Exported `connectionTagsKeySchema` and introduced `validateCaseInsensitiveTagKeys` in `packages/shared/lib/services/tags/schema.ts`, replacing the deprecated `.superRefine` block with a `.check` that consumes the helper.
• Re-exported the new schema elements from `packages/shared/lib/services/tags.service.ts` for downstream consumers.
• Updated `packages/server/lib/helpers/validation.ts` so `connectionEndUserTagsSchema` uses `connectionTagsKeySchema` plus the shared validation helper, ensuring API-level checks mirror the connection tag rules without altering key casing.

</details>

<details>

• `packages/shared/lib/services/tags/schema.ts`
• `packages/shared/lib/services/tags.service.ts`
• `packages/server/lib/helpers/validation.ts`

</details>

---

### DIFF
--- a/packages/shared/lib/services/tags/schema.ts
+++ b/packages/shared/lib/services/tags/schema.ts
@@ -4,7 +4,7 @@ export const TAG_MAX_COUNT = 10;
 export const TAG_KEY_MAX_LENGTH = 64;
 export const TAG_VALUE_MAX_LENGTH = 255;
 
-const connectionTagsKeySchema = z
+export const connectionTagsKeySchema = z
     .string()
     .regex(/^[a-zA-Z][a-zA-Z0-9_\-./]*$/, {
         message: 'Tag keys must start with a letter and contain only alphanumerics, underscores, hyphens, periods, or slashes'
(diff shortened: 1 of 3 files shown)


packages/shared/lib/services/tags/schema.ts
```diff
@@ -16,44 +16,47 @@ const connectionTagsValueSchema = z
...
-            if (normalized === 'end_user_email') {
-                const emailResult = z.string().email().min(5).safeParse(value);
-                if (!emailResult.success) {
-                    ctx.addIssue({
-                        code: z.ZodIssueCode.custom,
-                        message: 'Tag "end_user_email" must be a valid email'
-                    });
-                }
-            }
+    .check((payload) => {
```
marcindobry, line 53: `.superRefine` is deprecated, this is the recommended method to use instead

TBonnin (approved):

PR #5342, opened 2026-01-30
feat(tags): Add backfill script/migration for end user tags -> connection tags

marcindobry:
This query will update `114 282` rows in production. I will run it asynchronously using the migrate script in our cloud, and let knex automatically migrate connections for on-prem/self-hosted deployments.

There are a few assumptions in this query to make it simpler, but I did run the queries in prod DB to verify we don't have to care about them
- It doesn't truncate keys longer than 64 characters or values longer than 250 characters (which would be invalid new tags) because there are no existing keys or values that long
- It doesn't validate that it won't create more than 10 tags because there are no end users with that many (max is 4)
- It doesn't validate invalid key format because all existing end user tags are valid

---

**Backfill connection tags and harden tag sanitization**

Introduces a reusable SQL CTE pipeline and operational tooling to backfill `_nango_connections.tags` from existing end-user metadata while aligning runtime tag generation with the same constraints. Adds a non-transactional Knex migration, an operational `migrate.ts` runner with optional migration marking, a local seeding helper, and an end-to-end integration test that exercises truncation, invalid keys, tag precedence, and connections lacking end users. Updates `buildTagsFromEndUser` to normalize keys/values, enforce length/count limits, emit structured telemetry about dropped or truncated tags, and fall back to base tags when custom data is invalid so online ingestion matches the migration behavior.

<details>

• Added SQL helper `packages/database/lib/migration-helpers/backfillConnectionTagsSql.cjs` that normalizes base tags, sanitizes end-user tag entries, enforces 10-key limits, merges with existing connection tags, and returns summary metrics for operational visibility.
• Registered non-transactional migration `packages/database/lib/migrations/20260128120000_backfill_connection_tags.cjs` plus `scripts/one-off/connection-tags-migration/migrate.ts` (with `--mark-migration`) and companion README/seed tooling for cloud and self-hosted deployments.
• Created integration test `packages/database/lib/backfillConnectionTags.integration.test.ts` with inline seed helpers covering special characters, precedence, truncation, invalid formats, over-limit keys, and connections lacking end users.
• Expanded `packages/shared/lib/services/endUser.service.ts` to truncate/validate tags, log structured telemetry, drop custom tags when exceeding limits, and fall back gracefully if custom tags fail schema validation, updating unit tests accordingly.

</details>

<details>

• `packages/database/lib/migration-helpers/backfillConnectionTagsSql.cjs`
• `packages/database/lib/migrations/20260128120000_backfill_connection_tags.cjs`
• `scripts/one-off/connection-tags-migration/*`
• `packages/database/lib/backfillConnectionTags.integration.test.ts`
• `packages/shared/lib/services/endUser.service.ts` and unit tests

</details>

---

### DIFF
--- a/packages/database/lib/backfillConnectionTags.integration.test.ts
+++ b/packages/database/lib/backfillConnectionTags.integration.test.ts
@@ -1,0 +1,220 @@
+import { randomUUID } from 'node:crypto';
+import { createRequire } from 'node:module';
+
+import { beforeAll, describe, expect, it } from 'vitest';
+
+import db, { multipleMigrations } from './index.js';
+
+const require = createRequire(import.meta.url);
+const { buildConnectionTagsBackfillUpdateSql } = require('./migration-helpers/backfillConnectionTagsSql.cjs');
+
+// Inline seeds to avoid circular dependency with @nangohq/shared
+
+interface TestAccount {
+    id: number;
+    name: string;
+    uuid: string;
+}
+
+interface TestEnvironment {
+    id: number;
+    account_id: number;
+    name: string;
+}
+
+interface TestConfig {
+    id: number;
+    environment_id: number;
+    unique_key: string;
+    provider: string;
+}
+
+interface TestEndUser {
+    id: number;
+    end_user_id: string;
+    account_id: number;
+    environment_id: number;
+    email: string;
+    display_name: string | null;
+    organization_id: string | null;
+    organization_display_name: string | null;
+    tags: Record<string, string> | null;
+}
+
+interface TestConnection {
+    id: number;
+    environment_id: number;
+    provider_config_key: string;
+    connection_id: string;
+    end_user_id: number | null;
+    tags: Record<string, string>;
+}
+
+async function createTestAccount(): Promise<TestAccount> {
+    const name = `test-account-${randomUUID()}`;
+    const [account] = await db.knex.insert({ name }).into('_nango_accounts').returning('*');
+    return account;
+}
+
+async function createTestEnvironment(accountId: number): Promise<TestEnvironment> {
+    const [env] = await db.knex.insert({ account_id: accountId, name: 'dev' }).into('_nango_environments').returning('*');
+    return env;
+}
+
+async function createTestConfig(environmentId: number, uniqueKey: string, provider: string): Promise<TestConfig> {
+    const [config] = await db.knex
+        .insert({
+            environment_id: environmentId,
+            unique_key: uniqueKey,
+            provider: provider
+        })
+        .into('_nango_configs')
+        .returning('*');
+    return config;
+}
+
+async function createTestEndUser({
+    accountId,
+    environmentId,
+    endUserId,
+    email,
+    displayName,
+    organization,
+    tags
+}: {
+    accountId: number;
+    environmentId: number;
+    endUserId: string;
+    email: string;
+    displayName?: string | null;
+    organization?: { organizationId: string; displayName?: string | null } | null;
+    tags?: Record<string, string>;
+}): Promise<TestEndUser> {
+    const [endUser] = await db.knex
+        .insert({
+            end_user_id: endUserId,
+            account_id: accountId,
+            environment_id: environmentId,
+            email: email,
+            display_name: displayName ?? null,
+            organization_id: organization?.organizationId ?? null,
+            organization_display_name: organization?.displayName ?? null,
+            tags: tags ?? null
+        })
+        .into('end_users')
+        .returning('*');
+    return endUser;
+}
+
+async function createTestConnection({
+    environmentId,
+    configId,
+    providerConfigKey,
+    endUser,
+    tags
+}: {
+    environmentId: number;
+    configId: number;
+    providerConfigKey: string;
+    endUser?: TestEndUser;
+    tags?: Record<string, string>;
+}): Promise<TestConnection> {
+    const connectionId = `conn-${randomUUID()}`;
+    const [connection] = await db.knex
+        .insert({
+            environment_id: environmentId,
+            config_id: configId,
+            provider_config_key: providerConfigKey,
+            connection_id: connectionId,
+            credentials: {},
+            connection_config: {},
+            end_user_id: endUser?.id ?? null,
+            tags: tags ?? {}
+        })
+        .into('_nango_connections')
+        .returning('*');
+    return connection;
+}
+
+async function getConnectionTags(connectionId: number) {
+    const record = await db.knex.select('tags').from('_nango_connections').where({ id: connectionId }).first();
+    return record?.tags ?? null;
+}
+
+describe('Connection tags backfill', () => {
+    beforeAll(async () => {
+        await multipleMigrations();
+    });
+
+    it('backfills tags with edge cases', async () => {
+        const account = await createTestAccount();
+        const env = await createTestEnvironment(account.id);
+        const config = await createTestConfig(env.id, 'github', 'github');
+
+        // Test case 1: Special characters in all fields
+        const specialEndUser = await createTestEndUser({
+            accountId: account.id,
+            environmentId: env.id,
+            endUserId: 'user id/1?=+&',
+            email: 'user+test@example.com',
+            displayName: 'Jane Doe (R&D)',
+            organization: { organizationId: 'org:alpha/1', displayName: 'Org & Co' },
+            tags: {
+                project: 'R&D / alpha',
+                team: 'core team',
+                ip: '2001:db8::1'
+            }
+        });
+        const specialConn = await createTestConnection({
+            environmentId: env.id,
+            configId: config.id,
+            providerConfigKey: 'github',
+            endUser: specialEndUser,
+            tags: {}
+        });
+        const expectedSpecialTags = {
+            end_user_id: 'user id/1?=+&',
+            end_user_email: 'user+test@example.com',
+            end_user_display_name: 'Jane Doe (R&D)',
+            organization_id: 'org:alpha/1',
+            organization_display_name: 'Org & Co',
+            project: 'R&D / alpha',
+            team: 'core team',
+            ip: '2001:db8::1'
+        };
+
+        // Test case 2: Existing tags should take precedence over generated tags
+        const overrideEndUser = await createTestEndUser({
+            accountId: account.id,
+            environmentId: env.id,
+            endUserId: 'user-override',
+            email: 'original@example.com'
+        });
+        const overrideConn = await createTestConnection({
+            environmentId: env.id,
+            configId: config.id,
+            providerConfigKey: 'github',
+            endUser: overrideEndUser,
+            tags: { end_user_email: 'override@example.com', custom: 'keep' }
+        });
+        const expectedOverrideTags = {
+            end_user_id: 'user-override',
+            end_user_email: 'override@example.com',
+            custom: 'keep'
+        };
+
+        // Test case 3: Connection without end_user should not be affected
+        const noEndUserConn = await createTestConnection({
+            environmentId: env.id,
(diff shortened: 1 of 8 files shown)


scripts/one-off/connection-tags-migration/seed-local.ts
```diff
@@ -0,0 +1,132 @@
+import db from '../../../packages/database/lib/index.js';
```
marcindobry, line 1: This script for local testing, once it's all migrated, it's safe to delete

scripts/one-off/connection-tags-migration/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Run
+
+Dry run (runs the query in a transaction and then rolls back):
+
+```bash
+npx tsx scripts/one-off/connection-tags-migration/migrate.ts --dry-run
+```
+
+Execute the update:
```
marcindobry, line 7: This version is for testing locally

scripts/one-off/connection-tags-migration/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+```bash
+npx tsx scripts/one-off/connection-tags-migration/migrate.ts
+```
+
+Execute the update and mark the migration as applied (cloud manual workflow):
+
+```bash
+npx tsx scripts/one-off/connection-tags-migration/migrate.ts --mark-migration
```
marcindobry, line 16: This is what I'll run in prod

scripts/one-off/connection-tags-migration/migrate.ts
```diff
@@ -0,0 +1,97 @@
+import { createRequire } from 'node:module';
+
+import { KnexDatabase } from '../../../packages/database/lib/index.js';
+
+const require = createRequire(import.meta.url);
+const { buildConnectionTagsBackfillUpdateSql } = require('../../../packages/database/lib/migration-helpers/backfillConnectionTagsSql.cjs');
+
+interface ScriptOptions {
+    dryRun: boolean;
+    markMigration: boolean;
```
TBonnin, line 9: nit: I would have called it `commitMigration`

scripts/one-off/connection-tags-migration/migrate.ts
```diff
@@ -0,0 +1,97 @@
...
+    const migrationName = '20260128120000_backfill_connection_tags.cjs';
+    const migrationTable = '_nango_auth_migrations';
+
+    const existing = await knex.withSchema(schemaName).select('name').from(migrationTable).where({ name: migrationName }).first();
+    if (existing) {
+        console.log(`Migration already marked: ${migrationName}`);
+        return;
+    }
+
+    const batchResult = await knex.withSchema(schemaName).from(migrationTable).max('batch as max').first();
```
TBonnin, line 37: don't we want the `last`

scripts/one-off/connection-tags-migration/migrate.ts
```diff
@@ -0,0 +1,97 @@
...
+    const maxBatch = Number(batchResult?.max ?? 0);
```
TBonnin: I would just fail if batchResult.max is not defined. That shouldn't happen in prod

scripts/one-off/connection-tags-migration/migrate.ts
```diff
@@ -0,0 +1,97 @@
...
+    const nextBatch = Number.isNaN(maxBatch) ? 1 : maxBatch + 1;
```
TBonnin: same thing here if maxBatch isn't a number something is wrong

packages/database/lib/backfillConnectionTags.integration.test.ts
```diff
@@ -0,0 +1,220 @@
...
+
+async function getConnectionTags(connectionId: number) {
+    const record = await db.knex.select('tags').from('_nango_connections').where({ id: connectionId }).first();
+    return record?.tags ?? null;
+}
+
+describe('Connection tags backfill', () => {
+    beforeAll(async () => {
+        await multipleMigrations();
+    });
```
TBonnin, line 154: should we cleanup all the tests data  or it's unnecessary?

packages/database/lib/migration-helpers/backfillConnectionTagsSql.cjs
```diff
@@ -0,0 +1,71 @@
...
+                    'organization_display_name', organization_display_name
+                )
+            )
+            || COALESCE(
+                (
+                    SELECT jsonb_object_agg(key, value)
+                    FROM jsonb_each_text(COALESCE(end_user_tags::jsonb, '{}'::jsonb))
+                ),
+                '{}'::jsonb
+            )
```
TBonnin: can this simply be `|| COALESCE(end_user_tags, '{}'::jsonb)`? isn't end_user_tags already jsonb?

scripts/one-off/connection-tags-migration/migrate.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    try {
+        if (options.dryRun) {
+            const trx = await knex.transaction();
+            try {
+                const updateResult = await trx.raw(updateSql);
+                const updatedCount = updateResult.rowCount ?? 0;
+                console.log(`Dry run updated rows (rolled back): ${updatedCount}`);
+            } finally {
+                await trx.rollback();
```
TBonnin: not sure I understand the `finally { rollback }`

scripts/one-off/connection-tags-migration/migrate.ts
```diff
@@ -0,0 +1,97 @@
...
+            }
+            return;
+        }
+
+        const updateResult = await knex.raw(updateSql);
+        const updatedCount = updateResult.rowCount ?? 0;
+        console.log(`Updated rows: ${updatedCount}`);
+
+        if (options.markMigration) {
+            await markMigrationApplied(knex, schemaName);
```
TBonnin: should applying the migration be part of the transaction?

TBonnin (commented):
> It doesn't truncate keys longer than 64 characters or values longer than 250 characters (which would be invalid new tags) because there are no existing keys or values that long
    It doesn't validate that it won't create more than 10 tags because there are no end users with that many (max is 4)
    It doesn't validate invalid key format because all existing end user tags are valid

That's true for cloud but what if self-hosted customers have invalid data, what would happen?

marcindobry:
> That's true for cloud but what if self-hosted customers have invalid data, what would happen?

That is an excellent point that I didn't consider. I'll need to rework this a bit to take that into account. What do you think about:
- too long tag keys: truncate
- too long tag values: truncate
- too many tag keys: only use the ones generated from end_user/organization fields directly and skip end_user.tags
- invalid key format: skip

Honestly, I could be convinced into any approach to handle each of these cases, don't have strong opinions yet on any of them.

scripts/one-off/connection-tags-migration/migrate.ts
marcindobry: If it's max, does it matter?

scripts/one-off/connection-tags-migration/migrate.ts
marcindobry: This for a dry-run, it rolls back the updates. Now thinking about it, the operation is idempotent and if it populates the new column incorrectly, it can be reran after implementing a fix so maybe it's not worth having this.

TBonnin:
>     * too long tag keys: truncate
> 
>     * too long tag values: truncate
> 
>     * too many tag keys: only use the ones generated from end_user/organization fields directly and skip end_user.tags
> 
>     * invalid key format: skip

sounds reasonable.
Would it be possible to output the truncated/skipped entries in the logs by any chance?

scripts/one-off/connection-tags-migration/migrate.ts
TBonnin: indeed 🙈

packages/database/lib/migration-helpers/backfillConnectionTagsSql.cjs
```diff
@@ -14,56 +14,137 @@ WITH source AS (
...
+        kv.key AS raw_key,
+        kv.value AS raw_value,
+        left(lower(kv.key), 64) AS normalized_key,
+        NULLIF(left(kv.value, 255), '') AS normalized_value,
+        (length(kv.key) > 64) AS key_too_long,
+        (length(kv.value) > 255) AS value_too_long,
+        (left(lower(kv.key), 64) ~ '^[a-z][a-z0-9_./-]*$') AS key_valid,
+        (NULLIF(left(kv.value, 255), '') IS NOT NULL) AS value_valid
+    FROM base AS b
+    CROSS JOIN LATERAL jsonb_each_text(b.end_user_tags_object) AS kv(key, value)
```
marcindobry, line 45: This creates a row per end_user.tags entry so that we can then sanitize and count a problem with each of these entries

packages/database/lib/migration-helpers/backfillConnectionTagsSql.cjs
```diff
@@ -14,56 +14,137 @@ WITH source AS (
...
+    SELECT
+        b.id,
+        b.existing_tags,
+        b.base_tags,
         COALESCE(
-            (
-                SELECT jsonb_object_agg(lower(key), value)
-                FROM jsonb_each_text(raw_tags)
-            ),
+            jsonb_object_agg(e.normalized_key, e.normalized_value) FILTER (WHERE e.key_valid AND e.value_valid),
```
marcindobry, line 53: This combines all the sanitized tag rows back into a single row

packages/database/lib/migration-helpers/backfillConnectionTagsSql.cjs
```diff
@@ -14,56 +14,137 @@ WITH source AS (
...
+        b.base_tags
+ ),
+bounded AS (
+    SELECT
+        s.id,
+        s.existing_tags,
+        s.base_tags,
+        s.end_user_tags_sanitized,
+        s.candidate_key_count,
+        CASE
```
marcindobry, line 72: This just adds a flag we can count if we are above >10 keys

packages/database/lib/migration-helpers/backfillConnectionTagsSql.cjs
```diff
@@ -14,56 +14,137 @@ WITH source AS (
...
+            WHEN s.candidate_key_count > 10 THEN true
+            ELSE false
+        END AS dropped_end_user_tags_due_to_max_count,
+        CASE
```
marcindobry, line 76: This actually drops the custom end user tags if we are above 10 keys and only uses the base generated tags

packages/database/lib/migration-helpers/backfillConnectionTagsSql.cjs
```diff
@@ -14,56 +14,137 @@ WITH source AS (
...
+        FROM sanitized_end_user_tags
+    ) AS s
+ ),
 merged AS (
     SELECT
         id,
-        existing_tags,
-        generated_tags,
+        dropped_end_user_tags_due_to_max_count,
         (generated_tags || existing_tags) AS merged_tags
```
marcindobry, line 94: This does the final merge

packages/shared/lib/services/endUser.service.ts
```diff
@@ -319,41 +319,145 @@ export function buildTagsFromEndUser(
     endUser: ConnectSessionInput['end_user'] | null | undefined,
     organization: ConnectSessionInput['organization'] | null | undefined
 ): Tags {
-    const generatedTags: Tags = {};
+    const MAX_LOG_KEYS = 200;
```
marcindobry, line 322: This is in case somehow on-premise customers have created a billion tag keys, we don't want to log all of them

TBonnin (approved):

PR #5354, opened 2026-02-02
refactor(api-secrets): stop using env.secret_key in tests, use secret instead

thw-nango:
I split this out from a larger change set.

This PR rewires all integration tests in `package/server/lib/controllers` to stop using `env.secret_key` and use their default secret instead.

This is a voluminous but repetitive PR: It's the same edit applied over and over.

This is in preparation for the next PR that will remove the fields `secret_key` and `pending_secret_key` (and siblings) entirely from the environment object.

Notes:
 - The first commit is the bulk of the work.
 - I accidentally committed a file that I didn't want to, so I reverted it in the second commit.
 - Contrapositively, I missed one file, so I added it in the third commit.

---

Seeder helpers and shared utilities now return or fetch the seeded secret directly—using secretService when only an environment reference exists—and the common getConnectSessionToken helper was updated to accept a raw secret string so every test follows the same secret-handling path.

---

### DIFF
--- a/packages/server/lib/utils/tests.ts
+++ b/packages/server/lib/utils/tests.ts
@@ -173,7 +173,7 @@ export async function runServer(): Promise<{ server: Server; url: string; fetch:
 /**
  * Get connect session token
  * @param api
- * @param env
+ * @param secret
  * @returns connect session token
  * @throws Error if no connect session token
  * @example const token = await getConnectSessionToken(api, env);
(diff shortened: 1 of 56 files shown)


packages/server/lib/utils/tests.ts
TBonnin: `@param secret`

TBonnin (approved):

packages/server/lib/utils/tests.ts
thw-nango: Done.

PR #5356, opened 2026-02-02
fix(cli): accept user args when running the dryrun command

hassan254-prog:

- Fix CLI to accepts users args when running nango dryrun

---

The dryrun command now reads the environment, input, lastSyncDate, variant, metadata, diagnostics, and saveResponses flags directly from Command.opts(), passes them through to DryRunService.run, and clarifies in the --environment help text that an interactive prompt appears when the flag is omitted.

<details>

• Changed the `--environment` option description to indicate prompting instead of defaulting to `dev`.
• Ensured `input`, `lastSyncDate`, `variant`, `metadata`, `diagnostics`, and `saveResponses` are destructured from `this.opts()` alongside other flags.
• Passed the newly captured options through to `DryRunService.run` so user inputs (including `--diagnostics`) reach the service layer.

</details>

<details>

• `packages/cli/lib/index.ts`

</details>

---

### DIFF
--- a/packages/cli/lib/index.ts
+++ b/packages/cli/lib/index.ts
@@ -266,7 +266,7 @@ program
     .description('Dry run the sync|action process to help with debugging against an existing connection in cloud.')
     .argument('[name]', 'The name of the sync or action to run.')
     .argument('[connection_id]', 'The ID of the connection to use.')
-    .option('-e, --environment [environment]', 'The Nango environment, defaults to dev.', 'dev')
+    .option('-e, --environment [environment]', 'The Nango environment. If not provided, you will be prompted to select one.')
     .option(
         '-l, --lastSyncDate [lastSyncDate]',
         'Optional (for syncs only): last sync date to retrieve records greater than this date. The format is any string that can be successfully parsed by `new Date()` in JavaScript'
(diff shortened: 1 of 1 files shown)


packages/cli/lib/index.ts
```diff
@@ -291,11 +291,11 @@ program
     .option('--save, --save-responses', 'Optional: Save all dry run responses to a tests/mocks directory to be used alongside unit tests', false)
     .option('--diagnostics', 'Optional: Display performance diagnostics including memory usage and CPU metrics', false)
     .action(async function (this: Command) {
-        const { autoConfirm, debug, interactive, integrationId, validation, saveResponses } = this.opts();
+        const { autoConfirm, debug, interactive, integrationId, validation, saveResponses, input, lastSyncDate, variant, metadata, diagnostics } = this.opts();
         const shouldValidate = validation || saveResponses;
         const fullPath = process.cwd();
         let [name, connectionId] = this.args;
-        let { e: environment } = this.opts();
+        let { environment } = this.opts();
```
marcindobry, line 298: Why does this have to be separate from line 294?

packages/cli/lib/index.ts
```diff
@@ -360,7 +360,12 @@ program
             connectionId,
             optionalEnvironment: environment,
             optionalProviderConfigKey: integrationId,
-            saveResponses
+            saveResponses,
+            input,
+            lastSyncDate,
+            variant,
+            metadata,
+            diagnostics
```
TBonnin, line 368: was it ever working?

TBonnin (approved):

packages/cli/lib/index.ts
khaliqgant: Definitely was working at one point. Unsure where exactly it broke though.

packages/cli/lib/index.ts
hassan254-prog: `environment` is mutable and reassigned later, while the rest aren't,  keeping it separate makes the mutability explicit

packages/cli/lib/index.ts
TBonnin: we used to pass `this.opts` https://github.com/NangoHQ/nango/commit/22e17a6d2dcca56a7a760cd6cce1a09bd7d2a258#diff-e9a1201e7374a8633c118e7688a0e2ba48f80a178f3205e3c068580ecae18f43L274

PR #5362, opened 2026-02-02
refactor(runner): consolidate persist client logic

TBonnin:
Consolidating all the logic to make requests from runner to persist into a single client that is using persist typed endpoint definitions.

There is 2 ways to make calls to persist from runner:
1. the legacy one (the one using axios)
2. the most recent option that is using persist typed endpoint definitions.

This PR is refactoring persist client in runner to consolidate all the logic in a single file using persist endpoint definition defined in `@nangohq/types`. The persist client is now using default node fetch

 

---

It also updates the shared HTTP helper, runner SDK, monitoring flows, environment parsing, and accompanying unit tests so that every runner execution path now relies on the consolidated fetch-based client.

<details>

• `packages/runner/lib/clients/persist.ts`
• `packages/runner/lib/clients/http.ts`
• `packages/runner/lib/monitor.ts`
• `packages/runner/lib/sdk/sdk.unit.test.ts`
• `packages/runner/lib/index.ts`

</details>

---


packages/runner/lib/sdk/persist.ts
```diff
@@ -1,350 +0,0 @@
...
-    MergingStrategy,
-    PostRecordsSuccess,
-    PutRecordsSuccess
-} from '@nangohq/types';
-import type { Result } from '@nangohq/utils';
-import type { AxiosInstance, AxiosRequestConfig, AxiosResponse } from 'axios';
-import type { Socket } from 'net';
-import type { RequestOptions } from 'node:https';
-
-export class MaxLifetimeAgent extends Agent {
```
TBonnin, line 26: `MaxLifetimeAgent` was introduced because previous cloud provider was implementing sticky connection load balancing. I think this problem doesn't exist anymore with aws LB

packages/runner/lib/clients/http.ts
```diff
@@ -1,45 +1,96 @@
...
+                return res;
+            } catch (err) {
+                if (shouldRetry(err)) {
+                    throw err;
+                }
 
-        return response;
+                // Non-retryable error
+                return new Response(JSON.stringify({ error: stringifyError(err) }), {
+                    status: 503,
```
thw-nango: 502 Bad Gateway is more common in these scenarios.

packages/runner/lib/clients/persist.ts
```diff
@@ -1,32 +1,302 @@
...
-        const resp = await httpFetch(`${this.baseUrl}/environment/${environmentId}/log`, {
-            method: 'POST',
+    private async fetch<R>({
+        method,
+        url,
+        data,
+        params
+    }: {
+        method: string;
+        url: string;
```
thw-nango: Param `url` corresponds to the URL's `path` component only. The Client builds the URL. Ensure this string is empty or starts with a `/`.

packages/runner/lib/clients/persist.ts
```diff
@@ -1,32 +1,302 @@
...
+                responseData.error &&
+                typeof responseData.error === 'object' &&
+                'message' in responseData.error
+                    ? String(responseData.error.message)
+                    : JSON.stringify(responseData);
+            return Err(new Error(message));
         }
-        return Ok(undefined as PostLog['Success']);
+
+        const responseData = await response.json().catch(() => undefined);
```
thw-nango: JSON parse error is swallowed here.

packages/runner/lib/clients/persist.ts
```diff
@@ -1,32 +1,302 @@
...
+    }
+
+    public async postLog({ environmentId, data }: { environmentId: number; data: string }): Promise<Result<void>> {
+        const res = await this.fetch<void>({
+            method: 'POST',
+            url: `/environment/${environmentId}/log`,
+            data: JSON.parse(data)
+        });
+        if (res.isErr()) {
+            return Err(new Error(`Failed to log: ${res.error.message}`));
```
thw-nango: Message suggests `logger.log` failed. Maybe something like "Failed to persist log entry" would be clearer.

packages/runner/lib/clients/persist.ts
```diff
@@ -1,32 +1,302 @@
...
+            body: data ? JSON.stringify(data) : null,
+            userAgent: this.userAgent
         });
-        if (!resp.ok) {
-            return Err(`postLog_failed`);
+
+        if (!response.ok) {
+            const responseData = await response.json().catch(() => ({}));
+            logger.error(`${method} ${url} failed: errorCode=${response.status} response='${JSON.stringify(responseData)}'`);
+            const message =
```
marcindobry: Could we break this line up into several checks and/or some function? It is quite a headache to parse as a human 😅

packages/runner/lib/clients/persist.ts
```diff
@@ -1,32 +1,302 @@
...
+                    : JSON.stringify(responseData);
+            return Err(new Error(message));
         }
-        return Ok(undefined as PostLog['Success']);
+
+        const responseData = await response.json().catch(() => undefined);
+        return Ok(responseData as R);
+    }
+
+    public async postLog({ environmentId, data }: { environmentId: number; data: string }): Promise<Result<void>> {
```
marcindobry, line 88: Is this the best place for a logging method to live? It isn't scoped to a connection like everything else persist-related, and actions could also want to use this.

packages/runner/lib/clients/persist.ts
TBonnin: good point. let's call it path then instead of url

packages/runner/lib/clients/persist.ts
TBonnin: We are talking about Nango Logs here. Not all endpoints in persist are scoped to a connection: https://github.com/NangoHQ/nango/blob/785a715e8fa0244a6f09cd9565b39cfafe4b4e1f/packages/types/lib/persist/api.ts#L13

thw-nango (approved):

PR #5366, opened 2026-02-03
fix(oauth): Validate state session ID in Oauth2 callback

thw-nango:
CVE-2023-36019 is out of the bag. We are affected by a similar vulnerability.

To mitigate, we validate `state` against an OOB cookie carried by the OAuth2 flow initiator.

Consider this change high-risk, as some OAuth2 providers are known to not implement the specification correctly. That could lead to broken providers in Nango.

You can ignore the first commit in this PR, as it turned out to be tautological.

---

The controller now mints a secure, HTTP-only `oauth2-${session.id}` cookie whenever it issues an authorization URL, the callback handler refuses token exchanges until that cookie is presented and then clears it, and cookie parsing is confined to the `/oauth/callback` route so the rest of the public router remains unaffected.

---


packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -1086,17 +1096,28 @@ class OAuthController {
...
+        if (req.cookies[stateCookie] !== '1') {
+            const err = new Error('Invalid state parameter in OAuth callback');
             errorManager.report(err, { source: ErrorSourceEnum.PLATFORM, operation: LogActionEnum.AUTH });
             authHtml({ res, error: err.message });
             return;
         }
+        res.clearCookie(stateCookie, {
+            secure: req.secure,
+            httpOnly: true
+        });
```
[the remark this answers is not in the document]
thw-nango: Done.

packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -1310,6 +1319,21 @@ class OAuthController {
...
         const setupAction = req.query['setup_action'] as string | undefined;
 
+        const stateCookie = `oauth2-${session.id}`;
+        if (req.cookies[stateCookie] !== '1') {
+            void logCtx.error('Invalid state parameter in OAuth2 callback', {
+                session,
+                query: req.query
+            });
+            await logCtx.failed();
+            res.redirect('https://http.cat/403');
```
TBonnin: can we not redirect to a page with a cat?

packages/server/lib/server.ts
```diff
@@ -50,7 +50,7 @@ process.on('uncaughtException', (err) => {
 const app = express();
 app.set('query parser', 'extended');
 app.disable('x-powered-by');
-app.set('trust proxy', 1);
```
TBonnin, line 53: why this change?

TBonnin (commented):
Have you tested this change? I don't think the public API currently support cookies.

Also what about the other oauth2 flows (mcp, oauth2cc, outbound, ...)? Are they also vulnerable?

packages/server/lib/routes.public.ts
```diff
@@ -100,6 +101,7 @@ publicAPI.use(
 );
 publicAPI.use(bodyParser.raw({ type: 'text/xml', limit: bodyLimit }));
 publicAPI.use(express.urlencoded({ extended: true, limit: bodyLimit }));
+publicAPI.use(cookieParser());
```
TBonnin: let's only enable cookies for the oauth endpoints, not all API

hassan254-prog: I have updated this with the latest [commit](https://github.com/NangoHQ/nango/pull/5366/commits/26ffe51e271f1a902dd34269b87d8b941c404a7e).

hassan254-prog:
> Have you tested this change? I don't think the public API currently support cookies.
> 
> Also what about the other oauth2 flows (mcp, oauth2cc, outbound, ...)? Are they also vulnerable?

I have tested this with several `OAUTH2` providers. This change applies only to `OAUTH2`; other auth modes such as `CUSTOM`, `APP`, `OAUTH1`, and `OUTBOUND`, are also affected and will require a separate pr (decided to split this) as they also rely on provider callbacks.

TBonnin (approved):

TBonnin:
> > Have you tested this change? I don't think the public API currently support cookies.
> > Also what about the other oauth2 flows (mcp, oauth2cc, outbound, ...)? Are they also vulnerable?
> 
> I have tested this with several `OAUTH2` providers. This change applies only to `OAUTH2`; other auth modes such as `CUSTOM`, `APP`, `OAUTH1`, and `OUTBOUND`, are also affected and will require a separate pr (decided to split this) as they also rely on provider callbacks.

Thank you for checking. Makes sense to split the PRs. It would be good to address the other auth modes relatively quickly

PR #5373, opened 2026-02-03
fix(server): filter connections with errors before paginating

kaposke:
Error filtering was happening after pagination

---

**Filter errored connections before pagination**

The `withError` filtering in `connectionService.listConnections` now happens inside the `filtered_connections` CTE so pagination operates on prefiltered IDs. The change also removes redundant post-join filtering and adds an integration test that reproduces the original issue by mixing errored and non-errored connections.

<details>

• Applied `whereExists`/`whereNotExists` against `ACTIVE_LOG_TABLE` inside `filtered_connections` when `withError` is `true` or `false`, ensuring the ID set is filtered before `orderBy`, `limit`, and `offset` are applied.
• Removed the downstream `whereNull`/`whereNotNull` checks on `active_logs_agg.connection_id` because error filtering now occurs upstream.
• Added integration test `should filter errored connections before pagination` to verify a `withError: true` request returns the first five errored connections even when underlying rows alternate between errored and non-errored.

</details>

<details>

• `packages/shared/lib/services/connection.service.integration.test.ts`

</details>

---

### DIFF
--- a/packages/shared/lib/services/connection.service.ts
+++ b/packages/shared/lib/services/connection.service.ts
@@ -815,7 +815,7 @@ class ConnectionService {
                 if (withError === false) {
                     // Only connections without active logs
                     subQuery.whereNotExists(function () {
-                        this.select('*')
+                        this.select('1')
                             .from(ACTIVE_LOG_TABLE)
                             .whereRaw(`${ACTIVE_LOG_TABLE}.connection_id = _nango_connections.id`)
                             .where(`${ACTIVE_LOG_TABLE}.active`, true);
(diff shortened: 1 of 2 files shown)


packages/shared/lib/services/connection.service.ts
```diff
@@ -812,6 +812,24 @@ class ConnectionService {
                     }
                 }
 
+                if (withError === false) {
+                    // Only connections without active logs
+                    subQuery.whereNotExists(function () {
+                        this.select('*')
```
kaposke: Went with subqueries here because aggregating for every connection that might not have it at all seemed more costly? This should make good use of the existing index, but lmk if you would do it differently.

TBonnin: you can use `select(1)` to signal intent (aka: the row content doesn't matter but its presence does)

TBonnin (approved):

PR #5376, opened 2026-02-04
fix(cli): disable ora spinner in non-interactive environments

### LINKED ISSUE #5207: Ora via `nango compile` is causing an issue with CI/CD tooling.
We're using [pre-commit](<https://pre-commit.com>) to validate our nango-integrations folder. We just moved over to the new non-yaml implementation and replaced our previous `nango generate` with `nango compile` now our pre-commit check for nango never finishes.

Tracking the issue: when I comment out all the [ora](<https://www.npmjs.com/package/ora>) references in: [https://github.com/NangoHQ/nango/blob/master/packages/cli/lib/zeroYaml/compile.ts](<https://github.com/NangoHQ/nango/blob/master/packages/cli/lib/zeroYaml/compile.ts>) the compile step completes with no issues when triggered by `pre-commit`

It looks like `ora` is only used within this single file; which is the reason why all other nango commands work fine with the pre-commit hooks.

I've searched the ora issues for any mention of this problem and haven't found anything yet.<br>I'm putting together a reproduction repo with `pre-commit` and `ora`. And will provide it here for your further investigation.

agusayerza:
## Issue
`nango compile` hangs indefinitely when run via pre-commit hooks due to an upstream bug in ora's `stdin-discarder` dependency (#5207). This is blocking a customer's CI/CD workflow.

## Root cause
The bug was fixed in `stdin-discarder@0.3.1`, but `ora@9.1.0` (latest) still depends on `stdin-discarder@^0.2.2` and hasn't released a patched version yet.

## Fix
Added a `createSpinner()` wrapper that returns a no-op spinner when running in non-interactive environments (CI, pre-commit hooks, piped output, or when `--no-interactive` flag is passed). The real ora spinner is only used when both `interactive=true` AND `process.stdout.isTTY` is true.

This workaround should be removed once `ora` releases a patched version

I have opened a PR to actually bump the `stdin-discarder` version on the `ora` repo (https://github.com/sindresorhus/ora/pull/251) 

Closes #5207

---

**Add resilient spinner wrapper and propagate non-interactive handling across CLI**

Introduces a centralized `Spinner` utility that wraps `ora`, falls back to text-based logging when either stdin/stdout is non-TTY, and is wired through compile, dev, deploy, migrations, init, clone, and test generation flows. Global CLI options now auto-disable interactivity in CI environments (`isCI` or `--no-interactive`), ensuring commands invoked from hooks or piped executions no longer hang, while also updating `ora` to `9.2.0` and related dependencies (e.g., `stdin-discarder@0.3.1`).

<details>

• Created `packages/cli/lib/utils/spinner.ts` implementing a Proxy-based `Spinner` class that mirrors the `Ora` interface and prints textual status when `process.stdin` or `stdout` is non-TTY.
• Replaced direct `ora` usage across CLI commands (`compile`, `dev`, `deploy`, zero-yaml migrations, init, clone, tests, etc.) with the new `Spinner`, adding `interactive` parameters where needed.
• Updated `packages/cli/lib/index.ts` global option handling to default to interactive mode, automatically disable it under CI, and pass the flag through to all commands.
• Bumped `ora` to `9.2.0` in `packages/cli/package.json` and lockfile, pulling in the fixed `stdin-discarder@0.3.1` plus new transitive dependencies.

</details>

<details>

• packages/cli/lib/utils/spinner.ts
• packages/cli/lib/zeroYaml/* (compile, dev, deploy, migrations, init, check)
• packages/cli/lib/services/{clone,test}.service.ts
• packages/cli/lib/index.ts
• packages/cli/package.json & package-lock.json

</details>

---

### DIFF
--- a/packages/cli/lib/zeroYaml/compile.ts
+++ b/packages/cli/lib/zeroYaml/compile.ts
@@ -4,12 +4,12 @@ import path from 'node:path';
 import * as babel from '@babel/core';
 import chalk from 'chalk';
 import { build } from 'esbuild';
-import ora from 'ora';
 import { serializeError } from 'serialize-error';
 import ts from 'typescript';
 
 import { generateAdditionalExports } from '../services/model.service.js';
 import { Err, Ok } from '../utils/result.js';
+import { Spinner } from '../utils/spinner.js';
 import { printDebug } from '../utils.js';
 import { allowedPackages, importRegex, npmPackageRegex, tsconfig, tsconfigString } from './constants.js';
 import { buildDefinitions } from './definitions.js';
@@ -26,8 +26,17 @@ import type { Result } from '@nangohq/types';
  * - Compile the code to .cjs
  * - Rebuild nango.yaml in memory
  */
-export async function compileAll({ fullPath, debug }: { fullPath: string; debug: boolean }): Promise<Result<boolean>> {
-    let spinner = ora({ text: 'Typechecking' }).start();
+export async function compileAll({
+    fullPath,
+    debug,
+    interactive = true
+}: {
+    fullPath: string;
+    debug: boolean;
+    interactive?: boolean;
+}): Promise<Result<boolean>> {
+    const spinnerFactory = new Spinner({ interactive });
+    let spinner = spinnerFactory.start('Typechecking');
 
     try {
         // Read the index.ts content
@@ -60,7 +69,7 @@ export async function compileAll({ fullPath, debug }: { fullPath: string; debug:
 
         // Build the entry points
         const text = `Building ${entryPoints.length} file(s)`;
-        spinner = ora({ text }).start();
+        spinner = spinnerFactory.start(text);
         printDebug('Building', debug);
         for (const entryPoint of entryPoints) {
             const entryPointFullPath = path.join(fullPath, entryPoint);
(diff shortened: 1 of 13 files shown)


packages/cli/lib/zeroYaml/compile.ts
```diff
@@ -17,6 +17,41 @@ import { CompileError, ReadableError, badExportCompilerError, fileErrorToText, t
...
+            return this;
+        },
+        fail(text?: string) {
+            if (text) {
+                console.log(`✗ ${text}`);
+            }
+            return this;
+        }
+    } as Ora;
+}
```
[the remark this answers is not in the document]
agusayerza: Fixing its own introduced bug .... smh

packages/cli/lib/zeroYaml/compile.ts
```diff
@@ -17,6 +17,37 @@ import { CompileError, ReadableError, badExportCompilerError, fileErrorToText, t
...
+            }
+            return this;
+        },
+        fail(text?: string) {
+            if (text) {
+                console.log(`✗ ${text}`);
+            }
+            return this;
+        }
+    } as Ora;
```
TBonnin: isn't Ora supporting more methods? Would it makes sense to wrap it into a `Proxy` to catch any undefined methods and avoid runtime crash?

TBonnin (approved):

packages/cli/lib/zeroYaml/compile.ts
agusayerza: Yes, that is what I'm actually doing locally, as well as replacing other ora usages on other CLI commands

TBonnin: I see you went into supported all known methods. I would have done something simpler just to avoid crash without supporting everything we don't use (yet). ex:

```
const safeSpinner = new Proxy(new Spinner(), {
  get(target, prop) {
    if (typeof target[prop] === 'function') return target[prop].bind(target);
    if (prop in target) return target[prop];
    return (...args) => {
      console.warn(`Method "${prop}" not found, called with:`, args);
      return undefined;
    };
  }
});
```

PR #5383, opened 2026-02-04
feat(integrations): add support for salesforce(jwt)

hassan254-prog:

- Add support for salesforce(jwt)

---

**Salesforce JWT integration across runtime, UI, and docs**

Introduces a first-class `salesforce-jwt` provider by defining its auth spec in `packages/providers/providers.yaml`, wiring up provider-client introspection for both OAuth2 and JWT-based Salesforce tokens, and teaching `connection.service` to treat `token_expires_in_ms = 0` as a non-expiring credential (with a safeguarding unit test). The PR also adds customer-facing documentation (quickstart, connect guide, generated tooling snippets) plus a nav entry and logo asset so the new integration is visible in the product and docs.

<details>

• Added a full `salesforce-jwt` provider definition (JWT signature parameters, credential schema, proxy base URL, docs links) and nav entry so it can be configured from Nango
• Updated `ProviderClient.introspectedTokenExpired` to reuse Salesforce introspection for both OAuth2 and `TWO_STEP` (JWT) connections, including validation for required client credentials
• Adjusted `connection.service.parseRawCredentials` to treat `token_expires_in_ms = 0` as “no expiry” and covered the behavior with `connection.service.unit.test.ts`
• Documented the new flow via `docs/api-integrations/salesforce-jwt*.mdx`, generated tooling/use-case snippets, and referenced assets such as the template logo

</details>

<details>

• `packages/providers/providers.yaml`
• `packages/shared/lib/clients/provider.client.ts`
• `packages/shared/lib/services/connection.service.unit.test.ts`
• `docs/api-integrations/salesforce-jwt*.mdx`
• `docs/snippets/generated/salesforce-jwt/*`
• `docs/docs.json`
• `packages/webapp/public/images/template-logos/salesforce-jwt.svg`

</details>

---


packages/webapp/public/images/template-logos/salesforce-jwt.svg
```diff
@@ -0,0 +1 @@
+salesforce.svg
```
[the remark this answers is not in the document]
hassan254-prog: this is a symbolic link

packages/providers/providers.yaml
```diff
@@ -12944,6 +12944,73 @@ salesforce:
...
+            example: 3MVG9dAEux2v1sLsg0e52***************CJpGsXgqg8E
+            description: The Consumer Key from your Salesforce Connected App
+            doc_section: '#step-4-get-consumer-key-and-consumer-secret'
+        clientSecret:
+            type: string
+            title: Consumer Secret
+            example: 954F73F4C234***************9D069B8FCF83B5
+            pattern: '^[A-F0-9]{64}$'
+            description: The Consumer Secret from your Salesforce Connected App
+            doc_section: '#step-4-get-consumer-key-and-consumer-secret'
```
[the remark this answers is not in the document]
hassan254-prog: I Tested this with upto 7 different connected apps consumer secrets and they all had this pattern.

packages/shared/lib/clients/provider.client.ts
```diff
@@ -199,23 +200,29 @@ class ProviderClient {
     }
 
     public async introspectedTokenExpired(config: ProviderConfig, connection: DBConnectionDecrypted): Promise<boolean> {
-        if (connection.credentials.type !== 'OAUTH2') {
+        const { credentials } = connection;
+        const isOAuth2 = credentials.type === 'OAUTH2';
+        const isTwoStepSalesforceJwt = credentials.type === 'TWO_STEP';
```
TBonnin: should the variable called `isTwoStep` ? Right now the only provider for this flow is Salesforce but it might not stay true in the future

packages/shared/lib/clients/provider.client.ts
```diff
@@ -199,23 +200,29 @@ class ProviderClient {
...
             case 'salesforce-sandbox':
             case 'salesforce-experience-cloud':
-                return this.introspectedSalesforceTokenExpired(
-                    credentials.access_token,
-                    config.oauth_client_id,
-                    config.oauth_client_secret,
-                    oauthConnection.connection_config as Record<string, string>
-                );
+            case 'salesforce-jwt':
+                return this.introspectedSalesforceTokenExpired(accessToken, clientId, clientSecret, connectionConfig);
```
TBonnin, line 229: this also applies to `salesforce-experience-cloud` now?

packages/shared/lib/services/connection.service.ts
```diff
@@ -1013,8 +1013,8 @@ class ConnectionService {
                         }
                         expiresAt = new Date(Date.now() + durationMs);
                     }
-                } else if (template.token_expires_in_ms) {
-                    expiresAt = new Date(Date.now() + template.token_expires_in_ms);
+                } else if (template.token_expires_in_ms != null) {
```
TBonnin, line 1029: can `token_expires_in_ms` ever be 0 though? It would mean expiresAt would be immediate

hassan254-prog: I was aiming for a simpler approach to avoid adding another field in the provider’s config. Where we rely on `token_expires_in_ms`, which is interpreted as follows: if the value is greater than 0, we set an `expireAt` time on our end; if it’s 0, we assume the token has no expiration, like with Salesforce, where this value is not provided and we always need to introspect the token instead.

packages/shared/lib/clients/provider.client.ts
hassan254-prog: Introspection has always been performed on all 3 salesforce providers prior to this change, including `salesforce-experience-cloud`.

packages/shared/lib/clients/provider.client.ts
hassan254-prog: I agree, updated with the latest [commit](https://github.com/NangoHQ/nango/pull/5383/commits/33a6f21ef441e959485e9739a5f151a83493a689).

TBonnin (approved):

packages/shared/lib/services/connection.service.ts
TBonnin: ok. let's add a comment so we remember that `template.token_expires_in_ms = 0 => no expiresAt`and maybe a test to make us pause if we change this logic one day

PR #5387, opened 2026-02-04
feat: migration to remove the unique name constraint from `providers_shared`

hassan254-prog:

- Migration to remove the unique name constraint from providers_shared. We now want to allow multiple credentials per provider to support credential rotation in cases where we hit limits.

---

The migration specifically drops the auto-generated providers_shared_credentials_name_key constraint via a non-transactional run and omits a rollback implementation, so reinstating the uniqueness check would require a follow-up migration.

<details>

• `packages/database/lib/migrations/20260204200310_remove_providers_shared_credentials_name_unique.cjs`

</details>

---


marcindobry (approved):

packages/database/lib/migrations/20260204200310_remove_providers_shared_credentials_name_unique.cjs
```diff
@@ -0,0 +1,15 @@
+const SHARED_CREDENTIALS_TABLE = 'providers_shared_credentials';
+
+exports.config = { transaction: false };
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`ALTER TABLE ${SHARED_CREDENTIALS_TABLE} DROP CONSTRAINT IF EXISTS providers_shared_credentials_name_key;`);
```
marcindobry, line 9: How confident are you that this is the name of the constraint everywhere? I can't find it being referenced in our codebase anywhere.

marcindobry: How will the rotation workflow look like? Maybe instead of dropping the constraint altogether, we add a column to the uniqueness. For example, if we add "version" and "is_active" columns, and make sure that the next version that's populated is always active version + 1, then we could just add version to the uniqueness key.

marcindobry (commented):
Taking back the approval after thinking a bit more about this, would like to discuss a few things

marcindobry (changes requested):
Requesting changes is the only way of taking back the approval :(

packages/database/lib/migrations/20260204200310_remove_providers_shared_credentials_name_unique.cjs
hassan254-prog: The constraint  name `providers_shared_credentials_name_key` is not explicitly defined in our code, it’s automatically generated by pg when we declared the column-level [unique constraint ](https://github.com/NangoHQ/nango/blob/6af9f1e248b79b23a335bd2708e0bd024171f240/packages/database/lib/migrations/20250728111830_create_providers_shared_credentials_table.cjs#L12) without specifying the name. pg uses the [format](https://www.pgtutorial.com/postgresql-tutorial/postgresql-unique-constraint) `{table_name}_{column_name}_key`, to create this constraint.

hassan254-prog: I was going for a more simpler approach of using the created date to determine if we have new credentials provisioned for the provider, allowing us to pick these for newer integrations.

marcindobry (approved):

PR #5393, opened 2026-02-05
feat(connection tags): Make end_user optional when tags are present

marcindobry:
---

**Allow tag-only connect sessions (creation, fetch, reconnect)**

Extends connect session handling so that new sessions (and reconnect sessions) can be created without supplying `end_user` data as long as top-level `tags` are present. The GET endpoint now returns nullable `endUser`, enforces a defensive guard that either an `endUser` or `tags` exist on the session, and type definitions plus integration coverage were updated to reflect the new payload union and nullable response.

<details>

• Introduced `bodySchemaWithTagsNoEndUser` in `packages/server/lib/controllers/connect/postSessions.ts` and routing logic that validates tag-only payloads before calling `generateSession`.
• Updated `packages/server/lib/controllers/connect/getSession.ts` to treat `endUser` as nullable, add a guard that rejects sessions lacking both `endUser` and `tags`, and return `endUser: null` when appropriate.
• Relaxed `packages/server/lib/controllers/connect/postReconnect.ts` so reconnect sessions no longer require `connection.end_user_id`, persisting `endUserId: null` when tags-only flows are used.
• Extended shared types in `packages/types/lib/connect/api.ts` with `PostConnectSessionsBody` union and nullable `ConnectSessionOutput['endUser']`.
• Added integration tests covering tag-only create, fetch, and reconnect scenarios (including empty tag objects) to ensure the new flows work end-to-end.

</details>

<details>

• packages/server/lib/controllers/connect/postSessions.ts
• packages/server/lib/controllers/connect/getSession.ts
• packages/server/lib/controllers/connect/postReconnect.ts
• packages/types/lib/connect/api.ts
• connect session integration tests

</details>

---

### DIFF
--- a/packages/server/lib/controllers/connect/getSession.ts
+++ b/packages/server/lib/controllers/connect/getSession.ts
@@ -22,7 +22,7 @@ export const getConnectSession = asyncWrapper<GetConnectSession>(async (req, res
 
     const { connectSession, account, environment, plan } = res.locals;
 
-    let endUser: InternalEndUser;
+    let endUser: InternalEndUser | null = null;
     if (connectSession.endUserId) {
         const getEndUser = await endUserService.getEndUser(db.knex, {
             id: connectSession.endUserId,
@@ -37,8 +37,14 @@ export const getConnectSession = asyncWrapper<GetConnectSession>(async (req, res
         endUser = getEndUser.value;
     } else if (connectSession.endUser) {
         endUser = connectSession.endUser;
-    } else {
-        res.status(404).send({ error: { code: 'not_found', message: 'End user not found' } });
+    }
+
+    // Defensive check: end_user can only be skipped when tags were used.
+    // This is validated by zod at session creation time, but we keep an explicit guard
+    // here to avoid regressions if the validation rules change.
+    const tagsArePresent = connectSession.tags !== undefined && connectSession.tags !== null;
+    if (!endUser && !tagsArePresent) {
+        res.status(400).send({ error: { code: 'invalid_body', message: 'end_user is required unless tags are provided' } });
         return;
     }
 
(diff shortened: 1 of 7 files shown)


packages/server/lib/controllers/connect/postReconnect.ts
```diff
@@ -59,19 +59,14 @@ export const postConnectSessionsReconnect = asyncWrapper<PostPublicConnectSessio
             };
         }
 
-        if (!connection.end_user_id) {
```
marcindobry, line 62: This checks whether a connection has been created using connect sessions or the public key method. An end_user_id present is used as a proxy to check whether it was created using a connect session, so this assumption won't be true anymore.

It is probably the most risky change here, but here is my thought process:

Connections won't have an end_user_id either if people are using tags, or using public key. If people are using tags, then it's fine to skip this check during reconnecting.

If people are using public key, they wouldn't have been able to reconnect before so they most likely don't have this flow implemented. New customers are also unable to use public key since it's been deprecated - this is only available for legacy customers.

packages/server/lib/controllers/connect/getSession.ts
```diff
@@ -37,8 +37,14 @@ export const getConnectSession = asyncWrapper<GetConnectSession>(async (req, res
...
     } else if (connectSession.endUser) {
         endUser = connectSession.endUser;
-    } else {
-        res.status(404).send({ error: { code: 'not_found', message: 'End user not found' } });
+    }
+
+    // Defensive check: end_user can only be skipped when tags were used.
+    // This is validated by zod at session creation time, but we keep an explicit guard
+    // here to avoid regressions if the validation rules change.
+    const tagsArePresent = connectSession.tags !== undefined && connectSession.tags !== null;
```
TBonnin, line 45: `!!connectSession.tags` would do the same no? 

Also do we need to check if tags has at least one key?

packages/server/lib/controllers/connect/postReconnect.ts
```diff
@@ -110,7 +105,7 @@ export const postConnectSessionsReconnect = asyncWrapper<PostPublicConnectSessio
 
         // create connect session
         const createConnectSession = await connectSessionService.createConnectSession(trx, {
-            endUserId: endUser.id,
+            endUserId: endUser ? endUser.id : null,
```
TBonnin: `endUser?.id || null` if you don't want to use ternary operator

packages/server/lib/controllers/connect/postSessions.ts
```diff
@@ -66,16 +73,32 @@ export const postConnectSessions = asyncWrapper<PostConnectSessions>(async (req,
...
+    if (val.success) {
+        const body: PostConnectSessions['Body'] = val.data;
+        await generateSession(res, body, plan);
         return;
     }
 
-    const { plan } = res.locals;
+    const bodyIsObject = req.body && typeof req.body === 'object' && !Array.isArray(req.body);
+    const hasTopLevelTags = bodyIsObject && Object.prototype.hasOwnProperty.call(req.body, 'tags');
+    const hasEndUser = bodyIsObject && Object.prototype.hasOwnProperty.call(req.body, 'end_user');
```
TBonnin: `in` is shorter than `hasOwnProperty`

```
const isObject = body && typeof body === 'object' && !Array.isArray(body); 

const hasTags = isObject && 'tags' in body; 

const hasEndUser = isObject && 'end_user' in body;
```

TBonnin (approved):

packages/server/lib/controllers/connect/getSession.ts
marcindobry: I decided against checking if tags has at least one key. Basically, even if it's `{}` but not `undefined/null` it is a signal that someone is using tags deliberately.

PR #5398, opened 2026-02-05
feat: add checkpoint support to functions-sdk

TBonnin:
To be used as follow when writing a function. Similar to the metadata one where the shape of the checkpoint is defined as a zod type which is enforced by the compiler when calling `getCheckpoint/saveCheckpoint`. 
Ex:
```ts
const sync = createSync({
    description: 'My super sync',
    ...
    checkpoint: z.object({
        lastPage: z.number(),
    }),
    exec: async (nango) => {
        const cp = await nango.getCheckpoint();
        for (const entries of nango.paginate({ endpoint: `/entries?page=${cp ? cp.lastPage : 0}` })) {
            const records = ... // process entries
            await nango.batchSave(records, 'MyModel');
            await nango.saveCheckpoint({ lastPage: entries.data.page, });
        }
   ...
}

```

Note: internals are not implemented yet. The objective of this PR is to settle on the interface cc @bastienbeurier 

---

**Add typed checkpoint support to Functions SDK & runner persistence**

This PR introduces first-class, typed checkpoint support across the Functions SDK, runner, CLI, and orchestration layers. Script factories (`createSync`, `createAction`, `createOnEvent`) now accept a `checkpoint` zod schema whose inferred type flows through `NangoActionBase`/`NangoSyncBase`, exposing strongly-typed `getCheckpoint`, `saveCheckpoint`, and `clearCheckpoint` helpers. On the runner side, a new `Checkpointing` helper coordinates per-connection persistence with optimistic locking, while orchestrator flows clear checkpoints on full re-runs and the CLI stubs the API for dry-run scenarios.

Backend services and utilities were updated to validate checkpoint payloads, support forced deletions, expose checkpoint key helpers, and include telemetry/reporting updates. Type declarations, snapshots, and unit tests were refreshed to reflect the expanded generics and API surface.

<details>

• Extended `createSync`/`createAction`/`createOnEvent` props to accept `checkpoint` schemas and thread the generic into `NangoActionBase`/`NangoSyncBase`, adding typed checkpoint methods and docstrings.
• Implemented `packages/runner/lib/sdk/checkpointing.ts` with optimistic locking, checkpoint validation, and state tracking, wiring it into `NangoActionRunner`/`NangoSyncRunner` plus telemetry allowlists.
• Updated persistence services (`deleteCheckpoint`, tests) to allow forced soft-deletes, validate payload shapes, and expose `getCheckpointKey` via `@nangohq/utils` for consistent key generation.
• Orchestrator `runSyncCommand` now clears checkpoints during `RUN_FULL`, and CLI dry-run SDKs stub checkpoint methods to maintain API compatibility.
• Regenerated SDK type declarations and CLI snapshots to include the new generics and methods.

</details>

<details>

• `packages/runner-sdk/lib/scripts.ts`
• `packages/runner-sdk/lib/action.ts`
• `packages/runner-sdk/lib/sync.ts`
• `packages/runner/lib/sdk/checkpointing.ts`
• `packages/shared/lib/services/checkpoints/*`
• `packages/utils/lib/checkpoint/checkpoint.ts`
• `packages/shared/lib/clients/orchestrator.ts`
• `packages/cli/lib/services/sdk.ts`
• `packages/runner-sdk/models.d.ts`

</details>

---

### DIFF
--- a/packages/cli/lib/services/sdk.ts
+++ b/packages/cli/lib/services/sdk.ts
@@ -135,6 +135,22 @@ export class NangoActionCLI extends NangoActionBase {
         // Not applicable to CLI
     }
 
+    // eslint-disable-next-line @typescript-eslint/require-await
+    public override async getCheckpoint(): Promise<null> {
+        // Not applicable to CLI
+        return null;
+    }
+
+    // eslint-disable-next-line @typescript-eslint/require-await
+    public override async saveCheckpoint(): Promise<void> {
+        // Not applicable to CLI
+    }
+
+    // eslint-disable-next-line @typescript-eslint/require-await
+    public override async clearCheckpoint(): Promise<void> {
+        // Not applicable to CLI
+    }
+
     protected showLoggerLevelWarning = showLoggerLevelWarning();
 }
 
@@ -172,6 +188,9 @@ export class NangoSyncCLI extends NangoSyncBase {
     tryAcquireLock = NangoActionCLI['prototype']['tryAcquireLock'];
     releaseLock = NangoActionCLI['prototype']['releaseLock'];
     releaseAllLocks = NangoActionCLI['prototype']['releaseAllLocks'];
+    getCheckpoint = NangoActionCLI['prototype']['getCheckpoint'];
+    saveCheckpoint = NangoActionCLI['prototype']['saveCheckpoint'];
+    clearCheckpoint = NangoActionCLI['prototype']['clearCheckpoint'];
 
     protected showLoggerLevelWarning = showLoggerLevelWarning();
 
--- a/packages/cli/lib/services/sdk.ts
+++ b/packages/cli/lib/services/sdk.ts
@@ -136,9 +136,9 @@ export class NangoActionCLI extends NangoActionBase {
     }
 
     // eslint-disable-next-line @typescript-eslint/require-await
-    public override async getCheckpoint(): Promise<null> {
+    public override async getCheckpoint<T = never>(): Promise<T> {
         // Not applicable to CLI
-        return null;
+        return null as T;
     }
 
     // eslint-disable-next-line @typescript-eslint/require-await
--- a/packages/runner-sdk/lib/action.ts
+++ b/packages/runner-sdk/lib/action.ts
@@ -482,7 +482,7 @@ export abstract class NangoActionBase<
      * const checkpoint = await nango.getCheckpoint();
      * ```
      * */
-    public abstract getCheckpoint(): Promise<TCheckpointInferred | null>;
+    public abstract getCheckpoint<T = TCheckpointInferred>(): Promise<T>;
 
     /**
      * Save a checkpoint object that can be used to store progress or state to resume from.
@@ -496,7 +496,7 @@ export abstract class NangoActionBase<
      * });
      * ```
      * */
-    public abstract saveCheckpoint(checkpoint: TCheckpointInferred): Promise<void>;
+    public abstract saveCheckpoint<T = TCheckpointInferred>(checkpoint: T): Promise<void>;
 
     /**
      * Clear the function checkpoint
--- a/packages/runner-sdk/lib/action.ts
+++ b/packages/runner-sdk/lib/action.ts
@@ -482,7 +482,7 @@ export abstract class NangoActionBase<
      * const checkpoint = await nango.getCheckpoint();
      * ```
      * */
-    public abstract getCheckpoint<T = TCheckpointInferred>(): Promise<T>;
+    public abstract getCheckpoint(): Promise<TCheckpointInferred | null>;
 
     /**
      * Save a checkpoint object that can be used to store progress or state to resume from.
@@ -496,7 +496,7 @@ export abstract class NangoActionBase<
      * });
      * ```
      * */
-    public abstract saveCheckpoint<T = TCheckpointInferred>(checkpoint: T): Promise<void>;
+    public abstract saveCheckpoint(checkpoint: TCheckpointInferred): Promise<void>;
 
     /**
      * Clear the function checkpoint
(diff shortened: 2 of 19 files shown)


packages/cli/lib/services/sdk.ts
```diff
@@ -135,6 +135,22 @@ export class NangoActionCLI extends NangoActionBase {
...
+    // eslint-disable-next-line @typescript-eslint/require-await
+    public override async saveCheckpoint(): Promise<void> {
+        // Not applicable to CLI
+    }
+
+    // eslint-disable-next-line @typescript-eslint/require-await
+    public override async clearCheckpoint(): Promise<void> {
+        // Not applicable to CLI
+    }
+
```
TBonnin, line 153: Not applicable to CLI but I will add in a follow-up PR an argument to the CLI `dryrun` command so developer can stub the checkpoint and test its checkpoint related logic, equivalent to what we are currently doing with metadata.

bastienbeurier:
On the proposed interface: 
- Should we have a simplified notation for date-based checkpoints (I think 99% of cases)
- Should we force specifying an additional timestamp for non-date-based checkpoints, so the UI/API can report how far we've synced

marcindobry (commented):
I like the interface, it's pretty simple. I think the date-based datapoints difficulties can be solved by good documentation and examples since we just store whatever the customer wants us to and they handle input/output of serialization/deserialization.

Are we going to introduce some limits to what can be stored in a checkpoint?

TBonnin:
> On the proposed interface:
> 
>     * Should we have a simplified notation for date-based checkpoints (I think 99% of cases)

Do you mean allowing the checkpoint to be a Date directly instead of an object? Yes we could I think, but I feel like the benefit is pretty small for the added complexity on our end.

```
checkpoint: z.object({ until: z.date() }),
exec: async (nango) => {
        const cp = await nango.getCheckpoint();
        await nango.get ({ endpoint: `/entries?until=${until ? cp.until : 0}` }))
        await nango.saveCheckpoint({ until: new Date(...) });
        ...

// vs

checkpoint: z.date(),
exec: async (nango) => {
        const until = await nango.getCheckpoint();
        await nango.get ({ endpoint: `/entries?until=${until ? until : 0}` }))
        await nango.saveCheckpoint({ until: new Date(...) });
        ...
``` 


>     * Should we force specifying an additional timestamp for non-date-based checkpoints, so the UI/API can report how far we've synced

What would this additional timestamp value be? Date.now? if yes we already store the checkpoint `updatedAt` internally

TBonnin:
> Are we going to introduce some limits to what can be stored in a checkpoint?

- values can only be number, boolean, date, string (max TBD)
- no nesting

packages/runner-sdk/lib/scripts.ts
```diff
@@ -112,6 +112,20 @@ export interface CreateSyncProps<TModels extends Record<string, ZodModel>, TMeta
      */
     metadata?: TMetadata;
 
+    /**
+     * The checkpoint schema for storing sync progress and resume state.
+     * Values must be primitives (string, number, boolean, or Date).
```
kaposke: Should we be clear(er) that it should be an object with those primitives?
Eg: "Must be an object with primitives (string, number, boolean, or Date)."

packages/runner-sdk/lib/scripts.ts
```diff
@@ -251,6 +273,20 @@ export interface CreateActionProps<TInput extends z.ZodTypeAny, TOutput extends
      */
     metadata?: TMetadata;
 
+    /**
+     * The checkpoint schema for storing action progress and resume state.
+     * Values must be primitives (string, number, boolean, or Date).
```
kaposke: Same as other comment (?)

packages/runner-sdk/lib/scripts.ts
kaposke: Also, I'm not sure it's obvious to a new-ish customer that this is per-connection. Might be worth being explicit (similar to metadata)

TBonnin: good point about being scoped per connection. I will add it

packages/runner/lib/sdk/checkpointing.ts
```diff
@@ -0,0 +1,118 @@
...
+} | null;
+
+export class Checkpointing {
+    private persistClient: PersistClient;
+    private environmentId: number;
+    private nangoConnectionId: number;
+    private key: string;
+
+    public from: CheckpointState = null; // checkpoint at first checkpointing call
+    public last: CheckpointState = null; // last saved checkpoint
```
kaposke, line 19: Do these need to be public?

packages/runner-sdk/lib/action.ts
```diff
@@ -471,4 +473,33 @@ export abstract class NangoActionBase<
      * Release all locks acquired during the execution of a script.
      */
     public abstract releaseAllLocks(): Promise<void>;
+
+    /**
+     * Retrieve the last saved checkpoint
```
kaposke, line 478: And I wonder if it's also relevant to mention the connection scope for these.
With onboarding in mind, new customers might not have the clear picture that a sync/action is scoped to a connection. Every little bit of reinforcement helps solidify the concept.

kaposke, line 478: And I wonder if it's also relevant to mention the connection scope for these.
With onboarding in mind, new customers might not have the clear picture that a sync/action is scoped to a connection. Every little bit of reinforcement helps solidify the concept.
CC @bastienbeurier

kaposke (approved):

marcindobry (approved):

packages/runner/lib/sdk/checkpointing.ts
TBonnin: right now no but in the future the checkpoint states will be returned in the output of the function to be forwarded in the webhook. I just anticipated it ;)

bastienbeurier (approved):

TBonnin:
@marcindobry @kaposke I have added the last missing piece to the checkpoint logic (aka: deleting the checkpoint when triggering a full sync) can you have a look at the the added commit https://github.com/NangoHQ/nango/pull/5398/changes/ab75bdd309ea405970899ff27955967357e71fcc

marcindobry:
:shipit:

PR #5412, opened 2026-02-09
fix(persist): set qs array limit to 100

TBonnin:
qs, the query parsing library persist is using, has a default array limit size of 20, which means any array parameters with more than 20 elements will be parsed as an indexed object, which break the zod validation of getRecordIds query

 

---

**Configure Persist query parser to allow larger arrays**

The PR replaces the default `express` extended query parser with an explicit `qs` parser configured to accept arrays up to length 100. It also adds `qs` as an explicit dependency so the custom parser is available at runtime.

<details>

• Import `qs` in `packages/persist/lib/server.ts` and replace `server.set('query parser', 'extended')` with a custom parser that calls `qs.parse(str, { arrayLimit: 100 })`
• Add the `qs@6.14.1` dependency in `packages/persist/package.json` and lockfile

</details>

<details>

• `packages/persist/lib/server.ts`
• `packages/persist/package.json`
• `package-lock.json`

</details>

---

### DIFF
--- a/packages/persist/package.json
+++ b/packages/persist/package.json
@@ -26,6 +26,7 @@
         "@nangohq/pubsub": "file:../pubsub",
         "dd-trace": "5.52.0",
         "express": "5.1.0",
+        "qs": "6.14.1",
         "zod": "4.0.5"
     },
     "devDependencies": {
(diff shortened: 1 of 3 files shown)


packages/persist/package.json
```diff
@@ -26,6 +26,7 @@
         "@nangohq/pubsub": "file:../pubsub",
         "dd-trace": "5.52.0",
         "express": "5.1.0",
+        "qs": "6.14.1",
```
TBonnin, line 29: qs wasn't explicitly defined as a persist dependency :(

hassan254-prog (approved):

PR #5426, opened 2026-02-10
feat(webapp): New referral screen

agusayerza:
# PR Summary

**Problem:** We needed a dedicated "How did you hear about Nango?" step after signup (per design), shown only to the first user of a workspace (not invited users), and to stop collecting this on the signup form itself.

- Removed the inline "How did you hear about Nango?" field from the signup form (`SignupForm.tsx`). Signup no longer sends or requires `foundUs`.
- Added a post-signup step at `/onboarding/hear-about-us` with the design: one question, 7 option buttons (e.g. "My team is already using Nango", "Search engine", "LLM search", etc.) and a "Skip" link.
- When it's shown: Only for the first user of a workspace: after Google OAuth signup (redirect when no invite) and after email verification (when `showHearAboutUs` is returned from verify endpoint). Invited users are never redirected there.
- Persistence: Choice (or `"skipped"`) is stored in `_nango_accounts.found_us` via `POST /api/v1/account/onboarding/hear-about-us`. Skip is stored as `"skipped"` so the step is not shown again.
- Analytics: PostHog event `signup_hear_about` with `source` (including `"skipped"` when they skip).
- Other: Route order and PrivateRoute updated so `/onboarding/hear-about-us` is not matched as `/:env` and doesn't 404.
- 
[Design](https://www.figma.com/design/a4t7gaUATMqUcWedowFPFd/%F0%9F%90%BA-Design-Sytem---Product?node-id=6672-25758&m=dev)

Screenshot 2026-02-10 at 10 42 16 AM

---

## Issue ticket number and link (if applicable)

NAN-4614

---

## Testing instructions

1. **Email signup (first user):** Sign up with email → get verification link from DB (e.g. `SELECT 'http://localhost:3000/signup/verification/' || email_verification_token FROM _nango_users WHERE email_verified = false AND email_verification_token IS NOT NULL ORDER BY created_at DESC LIMIT 1`) → open link → expect "Email verified" and redirect to `/onboarding/hear-about-us` → choose an option or Skip → redirect to `/`. Confirm `found_us` (or `skipped`) in `_nango_accounts`.
2. **Google signup (first user):** Sign up with Google (no invite) → after callback, expect redirect to `/onboarding/hear-about-us` → submit or Skip → redirect to `/`.
3. **Invited user:** Sign up via invite link (or Google with invite) → expect no redirect to hear-about-us; land on normal app.

---

Server-side support adds dedicated GET and POST onboarding controllers, their route wiring, and access-middleware exemptions, alongside shared account-service helpers that the signup and verification flows call to determine eligibility and persist the first user’s selection exactly once.

<details>

• `PrivateRoute` now special-cases `location.pathname.startsWith('/onboarding/hear-about-us')`; ensure this does not inadvertently allow other onboarding subpaths to skip env validation if more routes are added later.
• No explicit error feedback is shown on the hear-about-us page if POST fails (user is redirected only on 200); consider communicating failures to reduce silent drops.

</details>

---


packages/webapp/src/layout/DefaultLayout.tsx
```diff
@@ -5,9 +5,9 @@ interface DefaultLayoutI {
 export default function DefaultLayout({ children }: DefaultLayoutI) {
```
agusayerza, line 5: Updating `DefaultLayout` to match new design. Checked and did minor fixes to other pages using this layout so everything looks the same or better. Will follow up with a PR to update all those to match the new design as well.

packages/webapp/src/layout/DefaultLayout.tsx
```diff
@@ -5,9 +5,9 @@ interface DefaultLayoutI {
...
     return (
         <div className="flex min-h-full flex-col justify-center py-10 sm:px-6 lg:px-8">
-            <div className="bg-dark-800 border border-[#141417] shadow-sm shadow-zinc-900 p-12 mx-auto px-16">
-                <div className="">
-                    <img className="mx-auto h-14 w-auto" src="/logo-dark.svg" alt="Nango" />
+            <div className="flex flex-col bg-bg-elevated border border-border-disabled p-[60px] mx-auto">
+                <div className="pb-[20px]">
```
agusayerza, line 9: @kaposke I followed design exactly, this made it so I needed to use some static padding values. Do we have a workaround? `p-{x}` didn't quite match the spacing on the design. I think we should prefer those over static `px` values, but opening it up for discussion.

packages/webapp/src/pages/Onboarding/HearAboutUs.tsx
```diff
@@ -0,0 +1,98 @@
...
+        setSubmitting(false);
+        if (res.status === 200) {
+            navigate('/', { replace: true });
+        }
+    };
+
+    if (loading) {
+        return (
+            <DefaultLayout>
+                <div className="mt-4 flex flex-col justify-center items-center min-h-[200px]" />
```
agusayerza: Note-to-self: add a spinner

packages/server/lib/controllers/v1/account/onboarding/getHearAboutUs.ts
```diff
@@ -0,0 +1,25 @@
...
+    const emptyQuery = requireEmptyQuery(req, { withEnv: false });
+    if (emptyQuery) {
+        res.status(400).send({ error: { code: 'invalid_query_params', errors: zodErrorToHTTP(emptyQuery.error) } });
+        return;
+    }
+
+    const account = res.locals.account;
+    const count = await userService.countUsers(account.id);
+    const hasNotSetFoundUs = account.found_us === null || account.found_us === '';
+    const showHearAboutUs = hasNotSetFoundUs && count === 1;
```
TBonnin: you can spread the needed account fields and simplify the condition I think
```
const { id, found_us } = res.locals.account;
const onlyOneUser = await userService.countUsers(account.id) === 1;
const showHearAboutUs = !found_us && onlyOneUser;
```

packages/server/lib/controllers/v1/account/validateEmailAndLogin.ts
```diff
@@ -67,13 +68,21 @@ export const validateEmailAndLogin = asyncWrapper<ValidateEmailAndLogin>(async (
 
     await userService.verifyUserEmail(user.id);
 
+    let showHearAboutUs = false;
+    const account = await accountService.getAccountById(db.knex, user.account_id);
+    if (account) {
+        const count = await userService.countUsers(account.id);
+        const hasNotSetFoundUs = account.found_us === null || account.found_us === '';
+        showHearAboutUs = hasNotSetFoundUs && count === 1;
```
TBonnin: what about a function in account service since we have the same duplicated logic?

packages/shared/lib/services/account.service.ts
```diff
@@ -79,6 +79,10 @@ class AccountService {
         await db.knex.update({ name, updated_at: new Date() }).from<DBTeam>(`_nango_accounts`).where({ id });
     }
 
+    async updateFoundUs(accountId: number, foundUs: string): Promise<void> {
+        await db.knex.update({ found_us: foundUs, updated_at: new Date() }).from<DBTeam>(`_nango_accounts`).where({ id: accountId });
+    }
```
TBonnin, line 92: I would be tempted to modify the editAccount function above to support updating foundUs. Otherwise we are gonna end up with one function per editable field

agusayerza: Yes, makes sense

packages/webapp/src/pages/Onboarding/HearAboutUs.tsx
```diff
@@ -0,0 +1,98 @@
...
+                // navigate('/', { replace: true });
+                // return;
+            }
+            const data = (await res.json()) as GetOnboardingHearAboutUs['Success'];
+            if (!data.data.showHearAboutUs) {
+                console.log(data);
+                // navigate('/', { replace: true });
+                // return;
+            }
+            setLoading(false);
```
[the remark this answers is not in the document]
agusayerza: Had it commented to test it 🤦

packages/shared/lib/services/account.service.ts
```diff
@@ -75,8 +76,21 @@ class AccountService {
...
-    async editAccount({ name, id }: { name: string; id: number }): Promise<void> {
-        await db.knex.update({ name, updated_at: new Date() }).from<DBTeam>(`_nango_accounts`).where({ id });
+    async updateAccount({ id, name, foundUs }: { id: number; name?: string; foundUs?: string }): Promise<void> {
+        const updates: Partial<DBTeam> & { updated_at: Date } = { updated_at: new Date() };
+        if (name !== undefined) {
+            updates.name = name;
+        }
+        if (foundUs !== undefined) {
+            updates.found_us = foundUs;
+        }
```
TBonnin: Using the spreading operator is more concise
```
const updates: Partial<DBTeam> & { updated_at: Date } = {
   ...(name ? { name } : {})
   ...(foundUs ? { found_us: foundUs } : {})
   updated_at: new Date()
}
```

TBonnin (commented):
I will refer to @kaposke for the webapp changes but rest lgtm

TBonnin (approved):

packages/server/lib/controllers/v1/account/onboarding/postHearAboutUs.ts
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    const account = res.locals.account;
+    const hasSetFoundUs = !(account.found_us === null || account.found_us === '');
+    if (hasSetFoundUs) {
+        res.status(403).send({ error: { code: 'forbidden', message: 'Not allowed to set hear-about-us for this account.' } });
+        return;
+    }
+
+    await accountService.updateAccount({ id: account.id, foundUs: val.data.source });
```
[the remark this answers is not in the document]
kaposke: Account is the workspace.

packages/server/lib/middleware/access.middleware.ts
```diff
@@ -30,7 +30,16 @@ import type { NextFunction, Request, Response } from 'express';
...
-const ignoreEnvPaths = ['/api/v1/environments', '/api/v1/meta', '/api/v1/user', '/api/v1/user/name', '/api/v1/signin', '/api/v1/invite/:id'];
+const ignoreEnvPaths = [
+    '/api/v1/environments',
+    '/api/v1/meta',
+    '/api/v1/user',
+    '/api/v1/user/name',
+    '/api/v1/signin',
+    '/api/v1/invite/:id',
+    '/api/v1/account/onboarding/hear-about-us',
+    '/account/onboarding/hear-about-us'
```
kaposke: Why the one without the prefix?

packages/webapp/src/components/PrivateRoute.tsx
```diff
@@ -38,6 +38,12 @@ export const PrivateRoute: React.FC = () => {
             return;
         }
 
+        // Skip env validation to avoid 404 for paths under /onboarding/hear-about-us
+        if (location.pathname.startsWith('/onboarding/hear-about-us')) {
+            setReady(true);
+            return;
+        }
+
```
kaposke, line 46: Not sure I understood this one

packages/webapp/src/layout/DefaultLayout.tsx
kaposke: Sheet, I'm also modifying this. If it works ok for now, don't stress too much about it. I can handle it in my redesign.

packages/webapp/src/pages/Onboarding/HearAboutUs.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+export const HearAboutUs: React.FC = () => {
+    const navigate = useNavigate();
+    const analyticsTrack = useAnalyticsTrack();
+    const [loading, setLoading] = useState(true);
+    const [submitting, setSubmitting] = useState(false);
+
+    useEffect(() => {
+        const check = async () => {
+            const res = await apiFetch('/api/v1/account/onboarding/hear-about-us');
```
kaposke, line 42: I know this is a case where it's a simple fetch, with no caching needed, but in general I've been trying to stick to using tanstack-query and have things in hooks. It simplifies loading states at the very least. The auth pages are doing it this way though, and I'm refactoring them as well, so it' ok as well to merge this and I refactor it later with my changes.

kaposke (approved):

packages/webapp/src/components/PrivateRoute.tsx
agusayerza: The `/onboarding/hear-about-us` route lives under a PrivateRoute but doesn't need an environment selected. Without this skip, `PrivateRoute` would try to validate the alleged `:env` from the URL path, fail to find one, and show a 404. This bypasses that env validation for this specific path.

packages/webapp/src/pages/Onboarding/HearAboutUs.tsx
agusayerza: Leaving it as-is right now

PR #5435, opened 2026-02-11
chore: Fix release github action breaking package-lock.json

agusayerza:
## Root cause

The release workflow (`publish.yaml`) runs on `ubuntu-latest` (Linux x64). The old `publish.mjs` called `npm i` **8 times** during the publish sequence (once after each package publish). On Linux, npm strips platform-specific optional dependencies (e.g. `lightningcss-darwin-arm64`, `lightningcss-win32-x64-msvc`, etc.) from `package-lock.json` because they fail the `EBADPLATFORM` check. Then `gitrelease.mjs` commits this corrupted lockfile to master.

This is a long-standing npm bug: [npm/cli#4828](https://github.com/npm/cli/issues/4828) (opened April 2022) and [npm/cli#7961](https://github.com/npm/cli/issues/7961) (regression in npm 10.3.0+).

A fix was merged in [npm/cli#8184](https://github.com/npm/cli/pull/8184) and shipped in npm 11.3.0 (April 2025), but **it does not fully resolve the issue**. Users continue to report the bug on npm [11.5.2](https://github.com/npm/cli/issues/4828#issuecomment-2861929046), [11.6.0](https://github.com/npm/cli/issues/4828#issuecomment-2914428131), and later. Our release that broke ([bca40e6d](https://github.com/NangoHQ/nango/commit/bca40e6d075fc320019a7276b1e444fa8a8fdc44), v0.69.31) was already using npm 11.5.1. There is **no flag** (`--include=optional`, `--package-lock-only`, `--ignore-scripts`, etc.) that prevents the pruning.

I have tested the logic on this PR to verify it works on dryruns. We will need to verify the actual fix once this is merged on master.

## Fix

Replace all `npm i` and `npm version` calls in `publish.mjs` with direct JSON file manipulation:

- **`bumpWorkspacePackageVersion()`** — writes version directly to `packages/*/package.json`
- **`bumpReference()`** — updates cross-workspace dependency references in `package.json` files
- **`bumpRootVersion()`** — writes version to root `package.json`
- **`bumpLockfileVersions()`** — updates version entries in `package-lock.json` for all published packages

npm never modifies the lockfile during the release, so platform-specific optional dependencies are preserved.

## Ghost packages cleanup

Found 5 extraneous registry-resolved entries in `package-lock.json` for workspace packages (`@nangohq/types`, `@nangohq/node`, `@nangohq/providers`). These were nested under `packages/cli/node_modules/` and `packages/webapp/node_modules/` with `"extraneous": true`.

These are artifacts from the old publish flow: the old script would publish a package to npm, then run `npm i` — npm would pull the just-published version from the registry and nest it inside other workspace packages instead of using the workspace link. Each release accumulated more of these ghost entries.

Since we no longer call `npm i` during publish, these entries won't be recreated. They've been removed from the lockfile.

---

The change also introduces a curated publishedPackageNames set so every workspace package receives deterministic JSON-based version rewrites without depending on npm install/version.

<details>

• `bumpLockfileVersions` only rewrites dependency values that match the strict `versionRegex`, so caret/range specifiers or git refs will not be updated and may drift.
• Any workspace omitted from `publishedPackageNames` will never receive version updates in the lockfile, potentially causing publish inconsistencies.
• Concurrent release runs could race because package versions are written before `npm publish`; a second run might see already-bumped versions and behave unpredictably.

</details>

---

### DIFF
--- a/scripts/publish.mjs
+++ b/scripts/publish.mjs
@@ -7,6 +7,16 @@ const argv = minimist(process.argv.slice(2));
 const nextVersion = argv.version;
 const skipCli = argv['skip-cli'] === true;
 const dryRun = argv['dry-run'] === true;
+const publishedPackageNames = new Set([
+    '@nangohq/types',
+    '@nangohq/nango-yaml',
+    '@nangohq/providers',
+    '@nangohq/node',
+    '@nangohq/runner-sdk',
+    '@nangohq/frontend',
+    'nango'
+]);
+const lockfileDependencyFields = ['dependencies', 'devDependencies', 'peerDependencies', 'optionalDependencies'];
 
 if (!nextVersion) {
     echo`${chalk.red(`Please specify a version: "node publish.mjs 0.0.1"`)}`;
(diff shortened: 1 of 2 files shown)


scripts/publish.mjs
```diff
@@ -119,6 +129,36 @@ async function npmPublish(packageName) {
...
+    if (content.version !== nextVersion) {
+        content.version = nextVersion;
+        await fs.writeFile(rootPackageJson, `${JSON.stringify(content, null, 4)}\n`);
+        echo(chalk.grey(`  ${figures.tick} Bumped root package version in ${rootPackageJson}`));
+    }
+}
+
 async function bumpReference(packageName) {
     const packagesJson = await glob('packages/*/package.json');
     // We don't use npm install, because it behaves incoherently with workspaces and different terminals
```
agusayerza: Funny you say so... 😆

scripts/publish.mjs
```diff
@@ -119,6 +129,36 @@ async function npmPublish(packageName) {
...
+        if (content.name !== packageName) {
+            continue;
+        }
+
+        if (content.version !== nextVersion) {
+            content.version = nextVersion;
+            await fs.writeFile(packageJson, `${JSON.stringify(content, null, 4)}\n`);
+            echo(chalk.grey(`  ${figures.tick} Bumped ${packageName} version in ${packageJson}`));
+        }
+        return;
```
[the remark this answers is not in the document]
agusayerza: Missed this one when squashing all the commits

scripts/publish.mjs
```diff
@@ -135,8 +177,47 @@ async function bumpReference(packageName) {
     }
 }
 
-async function npmInstall() {
-    await spinner('npm install', async () => {
-        await $`npm i`;
+async function bumpLockfileVersions() {
+    await spinner('update package-lock versions', async () => {
+        const lockfilePath = 'package-lock.json';
+        const lock = JSON.parse((await fs.readFile(lockfilePath)).toString());
```
TBonnin: maybe a try/catch in case the file is actually broken

scripts/publish.mjs
```diff
@@ -110,15 +120,47 @@ async function npmPublish(packageName) {
...
+        await $`npm publish --access public --provenance -w "${packageName}"`;
+
         echo(chalk.green(`${figures.tick} Published ${packageName}      `));
     });
 }
 
+async function bumpWorkspacePackageVersion(packageName) {
+    const packagesJson = await glob('packages/*/package.json');
+    for (const packageJson of packagesJson) {
+        const content = JSON.parse((await fs.readFile(packageJson)).toString());
```
TBonnin: maybe try/catch if file is corrupted and parsing fails

TBonnin (approved):
Thank you for doing this.
Can we add a comment section on top of the bump functions to clarify why we don't do npm install/version with link to npm issue tracker and details about the issue, so we remember in 6months when one of us want to bring back npm install

PR #5439, opened 2026-02-11
feat: add checkpoint to sync status response

TBonnin:
Example:
```
{
  "syncs": [
    {
      "id": "6ec3be19-e91e-403d-b209-5299af4019a2",
      "name": "syncA",
      ...
      "checkpoint": {
        "lastPage": 3
      }
    },
    {
      "id": "4a08b79b-9501-418a-8d98-d0302fb37f99",
      "name": "syncB",
      ...
      "checkpoint": null
    }
  ]
}
```

---

**Expose sync checkpoints in `syncStatus` responses**

Adds checkpoint data to the sync-status pipeline so clients can inspect the last persisted checkpoint for each sync. The sync manager now loads the checkpoint record for every sync, surfaces it in `ReportedSyncJobStatus`, and propagates the field through the Node SDK types and public documentation (reference docs and OpenAPI spec). Soft-deleted checkpoints are filtered out so consumers only see active metadata.

<details>

• Load per-sync checkpoints in `packages/shared/lib/services/sync/manager.service.ts` and include `checkpoint: checkpoint.value && !checkpoint.value.deleted_at ? checkpoint.value.checkpoint : null` in the returned status payload.
• Extend `ReportedSyncJobStatus` in `packages/types/lib/sync/index.ts` and Node SDK `SyncStatus` in `packages/node-client/lib/types.ts` with an optional `checkpoint: Checkpoint | null` field imported from `@nangohq/types`.
• Document the new response property in `docs/reference/sdks/node.mdx` and `docs/spec.yaml`, describing the checkpoint object in both the SDK reference and OpenAPI schema.

</details>

<details>

• `syncStatus` now fails entirely if `getCheckpoint` errors for any sync, which may regress behavior compared to returning partial statuses.

</details>

---

### DIFF
--- a/docs/spec.yaml
+++ b/docs/spec.yaml
@@ -1805,6 +1805,11 @@ paths:
                                                 recordCount:
                                                     type: object
                                                     description: Total count of records for each model synced by the sync
+                                                checkpoint:
+                                                    anyOf:
+                                                        - type: object
+                                                        - type: 'null'
+                                                    description: Last checkpoint saved during sync execution. Null if no checkpoint exists.
                 '400':
                     $ref: '#/components/responses/BadRequest'
                 '401':
--- a/packages/shared/lib/services/sync/manager.service.ts
+++ b/packages/shared/lib/services/sync/manager.service.ts
@@ -564,7 +564,7 @@ export class SyncManagerService {
             latestResult: latestJob?.result,
             latestExecutionStatus: latestJob?.status,
             recordCount,
-            checkpoint: checkpoint.value?.checkpoint || null,
+            checkpoint: checkpoint.value && !checkpoint.value.deleted_at ? checkpoint.value.checkpoint : null,
             ...(includeJobStatus ? { jobStatus: latestJob?.status as SyncStatus } : {})
         };
     }
--- a/packages/shared/lib/services/sync/manager.service.ts
+++ b/packages/shared/lib/services/sync/manager.service.ts
@@ -1,4 +1,5 @@
-import { getLogger, stringifyError } from '@nangohq/utils';
+import db from '@nangohq/database';
+import { getCheckpointKey, getLogger, stringifyError } from '@nangohq/utils';
 
 import connectionService from '../connection.service.js';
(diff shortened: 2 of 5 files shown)


packages/shared/lib/services/sync/manager.service.ts
```diff
@@ -538,6 +540,17 @@ export class SyncManagerService {
                 {} as Record<string, number>
             ) || {};
 
+        const checkpoint = await getCheckpoint(db.knex, {
```
agusayerza, line 543: getCheckpoint db query isn't filtering for deleted checkpoints. Is this intentional?

docs/spec.yaml
```diff
@@ -1805,6 +1805,11 @@ paths:
                                                 recordCount:
                                                     type: object
                                                     description: Total count of records for each model synced by the sync
+                                                checkpoint:
+                                                    anyOf:
+                                                        - type: object
```
agusayerza, line 1810:
```suggestion
                                                        - type: object
                                                           additionalProperties:
                                                             oneOf:
                                                               - type: string
                                                               - type: number
                                                               - type: boolean
```

TBonnin: I follow the other object fields like records, which we don't type. Happy to add it though what do you think?

packages/shared/lib/services/sync/manager.service.ts
TBonnin: `getCheckpoint` is intentionally returning soft-deleted checkpoint so runner-sdk can still get version to undelete it if needed. 
What's not intentional is the status logic should discard the checkpoint if soft-deleted. Good catch. I will fix

TBonnin: [ab1c737](https://github.com/NangoHQ/nango/pull/5439/commits/ab1c7378cbaf70a73ba012e187cbd6ffe332bde6)

marcindobry (approved):

docs/spec.yaml
agusayerza: This is a super small nit tbh, so no strong oppinion. I would add it just as I think the more we document the better and its quite a small change

agusayerza (approved):

PR #5440, opened 2026-02-11
feat(cli): add --checkpoint option to dryrun command

TBonnin:
adding dryrun option `--checkpoint` to be able to test checkpointing in function (similar to the already existig --metadata option)
The object passed via this option would become the initial checkpoint when dry-running the function

```
// Example
nango dryrun mySync myConn --checkpoint '{"lastModifiedAt": "2026-02-11T22:04:30.803Z"}'
```

---

The CLI now accepts checkpoint payloads via inline JSON or @file inputs, runs them through the shared JSON parser and central checkpoint validator, threads the seeded state through the sync/action CLIs so dry runs exercise get/save/clear checkpoint logic, and documents the workflow while aligning the SDK’s checkpoint types—removing Date values—around the new validator.

<details>

• Removing `Date` from `CheckpointValue` is a breaking change for any existing checkpoints that stored Date instances—migrations or release notes may be required
• CLI still surfaces the raw `Invalid checkpoint: ...` message from Zod, which may be noisy for common mistakes (arrays, null)

</details>

---

### DIFF
--- a/packages/cli/lib/services/dryrun.service.ts
+++ b/packages/cli/lib/services/dryrun.service.ts
@@ -289,73 +289,32 @@ export class DryRunService {
         let normalizedInput;
 
         if (actionInput) {
-            if (actionInput.startsWith('@') && actionInput.endsWith('.json')) {
-                const fileContents = readFile(actionInput);
-                if (!fileContents) {
-                    console.log(chalk.red('The file could not be read. Please make sure it exists.'));
-                    return;
-                }
-                try {
-                    normalizedInput = JSON.parse(fileContents);
-                } catch {
-                    console.log(chalk.red('There was an issue parsing the action input file. Please make sure it is valid JSON.'));
-                    return;
-                }
-            } else {
-                try {
-                    normalizedInput = JSON.parse(actionInput);
-                } catch {
-                    throw new Error('Failed to parse --input');
-                }
+            const result = parseJsonArg(actionInput, 'input');
+            if (!result.ok) {
+                console.log(chalk.red(result.message));
+                return;
             }
+            normalizedInput = result.value;
         }
 
         if (rawStubbedMetadata) {
-            if (rawStubbedMetadata.startsWith('@') && rawStubbedMetadata.endsWith('.json')) {
-                const fileContents = readFile(rawStubbedMetadata);
-                if (!fileContents) {
-                    console.log(chalk.red('The metadata file could not be read. Please make sure it exists.'));
-                    return;
-                }
-                try {
-                    stubbedMetadata = JSON.parse(fileContents);
-                } catch {
-                    console.log(chalk.red('There was an issue parsing the metadata file. Please make sure it is valid JSON.'));
-                    return;
-                }
-            } else {
-                try {
-                    stubbedMetadata = JSON.parse(rawStubbedMetadata);
-                } catch {
-                    throw new Error('fail to parse --metadata');
-                }
+            const result = parseJsonArg(rawStubbedMetadata, 'metadata');
+            if (!result.ok) {
+                console.log(chalk.red(result.message));
+                return;
             }
+            stubbedMetadata = result.value;
         }
 
         let stubbedCheckpoint: Checkpoint | undefined = undefined;
 
         if (rawStubbedCheckpoint) {
-            let parsed: Checkpoint;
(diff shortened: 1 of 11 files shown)


packages/runner-sdk/lib/checkpoint.ts
```diff
@@ -0,0 +1,22 @@
+import * as z from 'zod';
+
+import type { Checkpoint } from '@nangohq/types';
+
+/*
+ * The checkpoint can contain date strings.
+ * This function recursively checks the checkpoint object and converts any string that is a valid datetime to a Date object.
+ */
+export function validateCheckpoint(checkpoint: Checkpoint): Checkpoint {
```
marcindobry: nit: This function doesn't really do any validation but actually mostly just transforms date strings into dates. Could we rename it to e.g. `parseDates`?

packages/runner-sdk/lib/checkpoint.ts
```diff
@@ -0,0 +1,22 @@
...
+    const validated: Checkpoint = {};
+    for (const [key, value] of Object.entries(checkpoint)) {
+        if (typeof value === 'string') {
+            const parsed = z.string().datetime().safeParse(value);
```
marcindobry: A suggestion for a drive-by fix: .datetime() is deprecated

```
@deprecated — Use z.iso.datetime() instead.
```

packages/runner-sdk/lib/checkpoint.ts
marcindobry: What would happen if the user actually decides to operate on datetime strings in their sync code? Can this lead to subtle bugs e.g. when after [saving a checkpoint](https://github.com/NangoHQ/nango/blob/b2aa364ca7d429314bcbe4c6bb842ff63ad3b57e/packages/runner/lib/sdk/checkpointing.ts#L65) `this.last` suddenly has a different datatype than what was passed in?

TBonnin: I have reworked the function to also validate that values are only `string | number | boolean | date`. The name is more fitting now aab199896

Regarding the automatic parsing of iso string into date: You are right, that's definitely a limitation of storing the dates as json, we loose the ability to differentiate between real date and date as string. 
Unless we encode checkpoint data differently, we either:
- remove `Date` as a possible type and let customers take care of the encoding/decoding. Then no problem with storing the data in json since it would only be a string, boolean or number but more work for customers.
- automatically parse iso string into date, making it impossible to manipulate checkpoint value that are pure string in iso format (ie: they must dates)
What do you think @marcindobry ?

marcindobry: If we aren't going to encode the datatype then I don't think we should allow dates, I can imagine this kind of logic will lead to a bunch of unexpected behaviors in the future. However, at this point I think it's also a product decision. Maybe we should actually encode the datatype to make date support less problematic?

packages/cli/lib/services/dryrun.service.ts
```diff
@@ -313,6 +332,32 @@ export class DryRunService {
...
+                if (!fileContents) {
+                    console.log(chalk.red('The checkpoint file could not be read. Please make sure it exists.'));
+                    return;
+                }
+                try {
+                    parsed = JSON.parse(fileContents);
+                } catch {
+                    console.log(chalk.red('There was an issue parsing the checkpoint file. Please make sure it is valid JSON.'));
+                    return;
+                }
```
agusayerza: Can we extract to a function and reuse? action, metadata and now checkpoints do this

packages/runner-sdk/lib/checkpoint.ts
TBonnin: the simplest is to not allow `Date` type. While encoding the datatype could be done in the runner, it would be very awkward with the CLI --metadata. Let's start simple and remove `date` as acceptable checkpoint field type. Most of the date checkpoint are gonna be used as string in url query string params or POST json body so in reality it might not be an issue

It is a breaking change but I just checked and nobody but me is using checkpoint in prod.

TBonnin: https://github.com/NangoHQ/nango/pull/5440/changes/ba93e51723f7d7f837bd4e5849aa05d358b7972e

TBonnin:
@marcindobry @agusayerza can I have a review of the last 2 commits. Thank you

agusayerza (approved):

PR #5443, opened 2026-02-12
fix(cli): Fix nango generate:tests to work with the unified mocks

marcindobry:
Also fix a few other issues with our CLI
- generated tests should use PaginationService from runner-sdk, not from lib
- When looking for mock responses, the wrong case was used for key lookup (uppercase vs lowercase)

---

**Align CLI test generation with unified mocks and pagination service**

PR updates the CLI test generation flow so it can consume the new unified mock artifacts produced by `nango dryrun --save`. It introduces a helper to probe multiple mock locations (unified files near tests, legacy colocated mocks) and logs the attempted paths when running with `--debug`, ensuring generated sync/action tests point `Nango*Mock` to the correct directory.
Additionally, the test mocks now import `PaginationService` from the public `@nangohq/runner-sdk` entry point and keep HTTP method keys in their original case to match the serialized mock files. CLI templates and unit tests were adjusted to emit `dirname: __dirname`, and the `dryrun` command description for `--save/--save-responses` now refers to the unified mock location.

<details>

• Added `firstExistingPath` in `packages/cli/lib/services/test.service.ts` to probe unified (`tests/*.test.json`) and legacy (`mocks/<name>`) mock locations before generating tests.
• Updated `packages/cli/lib/templates/*-test-template.ejs` to set `dirname: __dirname`, with corresponding unit assertions in `packages/cli/lib/services/test.service.unit.cli-test.ts`.
• Switched `packages/cli/lib/testMocks/utils.ts` to import `PaginationService` from `@nangohq/runner-sdk` and to preserve original HTTP method casing when reading/recording mocks.
• Improved debug messaging when mocks are missing by listing all attempted paths for syncs/actions.
• Tidied `packages/cli/lib/index.ts` CLI options by removing duplicate `--save-responses` and clarifying that `--save/--save-responses` writes unified mock files.

</details>

<details>

• When both `tests/<name>.test.json` and `tests/<integration>-<name>.test.json` exist, the CLI silently takes the first match; confirm this ordering is intentional and documented for users who maintain both.

</details>

---

### DIFF
--- a/packages/cli/lib/index.ts
+++ b/packages/cli/lib/index.ts
@@ -286,9 +286,8 @@ program
         '--integration-id [integrationId]',
         'Optional: The integration id to use for the dryrun. If not provided, the integration id will be retrieved from the nango.yaml file. This is useful using nested directories and script names are repeated'
     )
-    .option('--save-responses', 'Optional: Save all dry run responses to a tests/mocks directory to be used alongside unit tests', false)
     .option('--validate, --validation', 'Optional: Enforce input, output and records validation', false)
-    .option('--save, --save-responses', 'Optional: Save all dry run responses to a tests/mocks directory to be used alongside unit tests', false)
+    .option('--save, --save-responses', 'Optional: Save all dry run responses to <integration>/tests/<name>.test.json for unit tests', false)
     .option('--diagnostics', 'Optional: Display performance diagnostics including memory usage and CPU metrics', false)
     .action(async function (this: Command) {
         const { autoConfirm, debug, interactive, integrationId, validation, saveResponses, input, lastSyncDate, variant, metadata, diagnostics } = this.opts();
--- a/packages/cli/lib/services/test.service.ts
+++ b/packages/cli/lib/services/test.service.ts
@@ -163,6 +163,16 @@ async function pathExists(path: string): Promise<boolean> {
     }
 }
 
+async function firstExistingPath(paths: string[]): Promise<string | null> {
+    for (const p of paths) {
+        if (await pathExists(p)) {
+            return p;
+        }
+    }
+
+    return null;
+}
+
 async function findUpFilename(filename: string, fromDir: string): Promise<string | null> {
     let currentDir = fromDir;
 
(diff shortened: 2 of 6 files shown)


packages/cli/lib/testMocks/utils.ts
```diff
@@ -559,7 +558,7 @@ class RecordingFixtureProvider implements FixtureProvider {
     async getCachedResponse(identity: ConfigIdentity) {
         const data = await this.delegate.getCachedResponse(identity);
 
-        const method = identity.method.toUpperCase();
+        const method = identity.method;
```
marcindobry, line 561: `method` is always lowercase here.

packages/cli/lib/testMocks/utils.ts
```diff
@@ -6,7 +6,7 @@ import parseLinksHeader from 'parse-link-header';
 import { vi } from 'vitest';
 
 import { getProvider } from '@nangohq/providers';
-import paginateService from '@nangohq/runner-sdk/lib/paginate.service.js';
+import { PaginationService } from '@nangohq/runner-sdk';
```
marcindobry, line 9: At the point of using the CLI, there is no `lib` folder in node_modules.

packages/cli/lib/index.ts
```diff
@@ -286,9 +286,8 @@ program
         '--integration-id [integrationId]',
         'Optional: The integration id to use for the dryrun. If not provided, the integration id will be retrieved from the nango.yaml file. This is useful using nested directories and script names are repeated'
     )
-    .option('--save-responses', 'Optional: Save all dry run responses to a tests/mocks directory to be used alongside unit tests', false)
```
marcindobry, line 289: This was a duplicate entry

packages/cli/lib/services/test.service.ts
```diff
@@ -536,9 +546,17 @@ export async function generateTests({
                     continue;
                 }
                 const sync = syncs[currentSyncName];
-                const mockPath = path.resolve(absolutePath, `${integration}/mocks/${currentSyncName}`);
-
-                if (await pathExists(mockPath)) {
+                const mockCandidates = [
+                    // New unified mocks format (saved by `nango dryrun --save`)
```
marcindobry, line 550: This is in case people had mocks already saved from dryrun and wanted to generate tests. @agusayerza do you think it's reasonable to care about these cases or should we just assume that usually these 2 things are done in tandem and we can only care about the new mock directory?

agusayerza (approved):

packages/cli/lib/services/test.service.ts
agusayerza: In theory we discourage this from happening, but I think adding support makes sense right now. We can move away from this and enforce it later on (at least lets migrate our own templates repo before enforcing so we know we are at a stable place)

PR #5445, opened 2026-02-12
fix(autoprune): records for deleted connections were not pruned

TBonnin:
We were fetching the records connection in order to get the environment.
However deleted connection weren't returned and autopruning was bailing out without pruning the records.
This commit, instead of making an extra query to fetch the connection, is directly getting the environment by joining with the records count table.

 

---

The candidate selection query now enforces the presence of a matching record_counts row, surfacing an error when it is missing, and the daemon threads the supplied environment ID through its plan checks and pruning requests so these cleanups continue to work even after the connection itself has been removed.

<details>

• New `Err` return when `record_counts` is missing will stop pruning for that partition; ensure this is desired and that the error is actionable.

</details>

---

### DIFF
--- a/packages/shared/lib/services/connection.service.ts
+++ b/packages/shared/lib/services/connection.service.ts
@@ -1828,15 +1828,6 @@ class ConnectionService {
             .limit(limit);
     }
 
-    async hardDeleteByIntegration({ integrationId, limit }: { integrationId: number; limit: number }): Promise<number> {
-        return await db.knex
-            .from<DBConnection>('_nango_connections')
-            .whereIn('id', function (sub) {
-                sub.select('id').from<DBConnection>('_nango_connections').where('config_id', integrationId).limit(limit);
-            })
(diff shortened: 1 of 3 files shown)


packages/shared/lib/services/connection.service.ts
```diff
@@ -1828,15 +1828,6 @@ class ConnectionService {
...
 
-    async hardDeleteByIntegration({ integrationId, limit }: { integrationId: number; limit: number }): Promise<number> {
-        return await db.knex
-            .from<DBConnection>('_nango_connections')
-            .whereIn('id', function (sub) {
-                sub.select('id').from<DBConnection>('_nango_connections').where('config_id', integrationId).limit(limit);
-            })
-            .delete();
-    }
-
```
TBonnin, line 1839: unrelated but found out this function isn't used anymore

kaposke (approved):

PR #5454, opened 2026-02-13
fix: lower autoprune limit

TBonnin:
autoprune query is timing out a lot when trying to prune records. Lowering the limit (1000 -> 200) while I am looking into ways to speed up the query

---

The change specifically updates the default `PERSIST_AUTO_PRUNING_LIMIT` value in `packages/utils/lib/environment/parse.ts`, and it keeps the rest of the configuration unchanged so the timeout mitigation comes solely from the smaller batch size.

<details>

• Lower limit could let stale data accumulate faster if autoprune frequency is not adjusted in environments with large datasets.

</details>

---


TBonnin:
closing. I will override the default via infra env var.

PR #5467, opened 2026-02-17
fix(cli): Fix hash calculation in unified mocks when pagination took place

marcindobry:
`nango dryrun --save` records (and hashes) unified mocks from the final request URL by storing the path as endpoint and the query string as params, but link pagination at runtime rewrites subsequent requests by putting the query into endpoint and clearing params. That representation mismatch makes the computed request identity/hash differ, so the test runner can’t find the saved mock for paginated requests.

---

**Align unified mock hashing with response collector**

Updates the CLI mock utilities so unified mock identities are built the same way as the response collector, addressing failures where paginated requests rewrote query strings into the endpoint and broke hash matching. The change introduces unified-specific URL/param/header normalization logic, switches proxy requests to choose the correct identity builder per fixture type, and exports the response collector’s header filter so both paths stay consistent. A regression test now verifies that a mock recorded for `'/foo'` with params still matches a request made to `'/foo?a=1&b=2'`.

<details>

• Replaced the single `computeConfigIdentity` with `computeUnifiedConfigIdentity` and `computeLegacyConfigIdentity`, selecting between them in `NangoActionMock.proxyData()` based on `fixtureProvider.isUnifiedMocks()`.
• Added unified-specific helpers (`parseUnifiedEndpointAsUrl`, `applyParamsToUrl`, `normalizeHeadersForUnifiedIdentity`) to merge params into URLs, trim `/proxy/` prefixes, and filter headers identically to the response collector.
• Exported `FILTER_HEADERS` from `packages/cli/lib/services/response-collector.service.ts` and introduced `FILTER_HEADERS_LEGACY` to keep legacy behavior unchanged.
• Extended `packages/cli/lib/testMocks/utils.unified.unit.cli-test.ts` with a Vitest case ensuring mocks recorded with params match requests whose query is embedded in the endpoint.

</details>

<details>

• `applyParamsToUrl` replaces array params by joining with commas, which may diverge from APIs expecting repeated query keys.
• If both the endpoint string and `params` contain the same key, the final value depends on insertion order; reviewers should confirm this matches proxy construction.
• `normalizeHeadersForUnifiedIdentity` drops duplicate headers after lower-casing; ensure no scenarios rely on distinct casing or repeated headers.

</details>

---

### DIFF
--- a/packages/cli/lib/testMocks/utils.ts
+++ b/packages/cli/lib/testMocks/utils.ts
@@ -8,6 +8,8 @@ import { vi } from 'vitest';
 import { getProvider } from '@nangohq/providers';
 import { PaginationService } from '@nangohq/runner-sdk';
 
+import { FILTER_HEADERS as FILTER_HEADERS_UNIFIED } from '../services/response-collector.service.js';
+
 import type { CursorPagination, LinkPagination, OffsetCalculationMethod, OffsetPagination, Pagination, UserProvidedProxyConfiguration } from '@nangohq/types';
 import type { AxiosResponse } from 'axios';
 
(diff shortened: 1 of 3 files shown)


packages/cli/lib/testMocks/utils.ts
```diff
@@ -1061,7 +1062,36 @@ class NangoSyncMock extends NangoActionMock {
...
+    'user-agent',
+    'nango-proxy-user-agent',
+    'accept-encoding',
+    'retries',
+    'host',
+    'connection-id',
+    'provider-config-key'
+];
+
+// Keep in sync with packages/cli/lib/services/response-collector.service.ts FILTER_HEADERS
```
agusayerza: nit: should we export and reuse here then?

agusayerza (approved):

PR #5469, opened 2026-02-17
fix(webapp): fix 404 after login when default env is missing

hassan254-prog:

- fix 404 after login when env is missing. If the current env is in the env list, redirect to it, If not, redirect to `prod` if it exists, otherwise to the first env in the list as a last fallback.

---

The root redirect now explicitly checks the requested environment against the available list, loading the configured production environment name and falling back to that or the first available option before routing so users always land on a valid environment.

<details>

• Import `PROD_ENVIRONMENT_NAME` into `packages/webapp/src/pages/Root.tsx`
• When `location.pathname === '/'`, compute `targetEnv` by checking the requested `env`, falling back to prod, then to the first available environment
• Use the computed `targetEnv` in the `navigate` call to avoid 404s caused by stale defaults

</details>

<details>

• If `meta.environments` is an empty array, the fallback reuses the missing `env`, which might still yield a 404.

</details>

---

### DIFF
--- a/packages/webapp/src/pages/Root.tsx
+++ b/packages/webapp/src/pages/Root.tsx
@@ -1,6 +1,7 @@
 import { useEffect } from 'react';
 import { useLocation, useNavigate } from 'react-router-dom';
 
+import { PROD_ENVIRONMENT_NAME } from '../constants';
 import { useMeta } from '../hooks/useMeta';
 import { useStore } from '../store';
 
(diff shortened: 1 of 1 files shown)


marcindobry (approved):

marcindobry (changes requested):
Can we default to dev instead?

hassan254-prog:
> Can we default to dev instead?

`dev` can be renamed or deleted, `prod` is the only env that cannot be modified.

marcindobry (approved):

agusayerza (approved):

PR #5482, opened 2026-02-18
feat(integrations): okta-cc support

agusayerza:
### Problem
Nango had no support for OAUTH2_CC providers that authenticate using `private_key_jwt` (RFC 7523) instead of a client secret. This blocked adding Okta's Management API via Client Credentials, which exclusively uses `private_key_jwt` for its API Services app type.

Several gaps needed to be closed:
- `token_request_auth_method` only accepted `basic` and `custom` — `private_key_jwt` was not a valid value
- `getOauthClientCredentials` had no `private_key_jwt` branch
- The Connect UI always rendered `client_secret` as a required field for OAUTH2_CC, even when it is irrelevant
- The frontend SDK routed to the OAUTH2_CC auth endpoint only when `client_secret` was present as a key in the credentials object — so hiding it from the form caused a silent failure with no error shown to the user
- The backend's OAUTH2_CC guard required `client_secret` and had no fallback for `client_private_key`
- Scopes for OAUTH2_CC were not being forwarded from the integration config into `connectionConfig` when not explicitly passed

### Solution

**`private_key_jwt` auth method support:**
- `packages/shared/lib/services/connection.service.ts` — Added `private_key_jwt` branch in `getOauthClientCredentials`: parses the private JWK, exports it to PEM, and builds a signed RS256 JWT client assertion (`client_assertion_type: urn:ietf:params:oauth:client-assertion-type:jwt-bearer`). Made `client_secret` optional in the signature with explicit guards for `basic` and `custom`. `client_private_key` is now also filtered from logs.
- `packages/shared/lib/auth/jwt.ts` — Exported `signJWT` so it can be used outside the module.
- `packages/types/lib/providers/provider.ts` — Moved `token_request_auth_method` to `BaseProvider` (was only on `ProviderOAuth2`) and added `private_key_jwt` as a valid value.
- `scripts/validation/providers/schema.json` — Added `private_key_jwt` to the `token_request_auth_method` enum.

**Connect UI fix:**
- `packages/connect-ui/src/views/Go.tsx` — Hides `client_secret` and marks it optional in the Zod schema when `token_request_auth_method === 'private_key_jwt'`.

**Frontend SDK fix:**
- `packages/frontend/lib/index.ts` — Updated both OAUTH2_CC detection checks (`convertCredentialsToConfig` and `customAuth`) to also match when `client_private_key` is present, since `private_key_jwt` flows don't include `client_secret` in the credentials object at all.
- `packages/frontend/lib/types.ts` — Made `client_secret` optional in the frontend `OAuth2ClientCredentials` interface.

**Backend fix:**
- `packages/server/lib/controllers/oauth.controller.ts` — Relaxed the `missing_client_secret` guard to pass when `client_private_key` is present. Also forwards `oauth_scopes` from the integration config into `connectionConfig` when not already set.

[NAN-4803](https://nango.dev/NAN-4803)

---

Documentation and catalog assets for the new `okta-cc` integration were migrated into the `/api-integrations` structure, including generated snippets and a setup guide so teams can adopt the Okta Management API template end‑to‑end.

<details>

• Auto-populating `connectionConfig['oauth_scopes']` whenever the integration defines scopes may surprise teams that relied on an empty scope list to request all defaults—confirm desired behavior or gate by provider.
• `parsedCreds.client_secret` now stores `''` when no secret is provided; downstream consumers expecting `undefined` may misinterpret an empty string.

</details>

---

### DIFF
--- a/packages/server/lib/controllers/oauth.controller.ts
+++ b/packages/server/lib/controllers/oauth.controller.ts
@@ -345,13 +345,13 @@ class OAuthController {
             return;
         }
 
-        if (!body.client_secret) {
+        if (!body.client_secret && !body.client_private_key) {
             errorManager.errRes(res, 'missing_client_secret');
 
             return;
         }
 
-        const { client_id, client_secret, client_certificate, client_private_key }: Record<string, string> = body;
+        const { client_id, client_secret, client_certificate, client_private_key }: Record<string, string | undefined> = body;
 
         if (isConnectSession && receivedConnectionId) {
             errorRestrictConnectionId(res);
@@ -443,6 +443,10 @@ class OAuthController {
                 if (defaults?.connectionConfig) {
                     Object.assign(connectionConfig, defaults.connectionConfig);
                 }
+            }
+
+            if (config.oauth_scopes && !connectionConfig['oauth_scopes']) {
+                connectionConfig['oauth_scopes'] = config.oauth_scopes;
             }
 
             if (missesInterpolationParam(tokenUrl, connectionConfig)) {
--- a/packages/shared/lib/services/connection.service.ts
+++ b/packages/shared/lib/services/connection.service.ts
@@ -1231,7 +1231,7 @@ class ConnectionService {
                     jti: uuidv4()
                 },
                 secretOrPrivateKey: privKeyPem,
-                options: { algorithm: 'RS256', header: { kid } }
+                options: { algorithm: 'RS256', keyid: kid }
             });
             params.append('client_id', client_id);
             params.append('client_assertion_type', 'urn:ietf:params:oauth:client-assertion-type:jwt-bearer');
--- a/packages/shared/lib/services/connection.service.ts
+++ b/packages/shared/lib/services/connection.service.ts
@@ -1203,10 +1203,22 @@ class ConnectionService {
             if (!client_private_key) {
                 throw new NangoError('missing client_private_key');
             }
-            const privateJwk = JSON.parse(client_private_key) as { kid: string; [key: string]: unknown };
-            const kid = privateJwk.kid;
-            // Export the JWK to PEM so signJWT can use it.
-            const privKeyPem = createPrivateKey({ key: privateJwk, format: 'jwk' }).export({ type: 'pkcs8', format: 'pem' }) as string;
+            let privKeyPem: string;
+            let kid: string;
+
+            try {
+                const privateJwk = JSON.parse(client_private_key) as { kid: string; [key: string]: unknown };
+                kid = privateJwk.kid;
+                privKeyPem = createPrivateKey({
+                    key: privateJwk,
+                    format: 'jwk'
+                }).export({
+                    type: 'pkcs8',
+                    format: 'pem'
+                }) as string;
+            } catch (err) {
+                throw new NangoError('invalid_client_private_key_format', { cause: err });
+            }
             const now = Math.floor(Date.now() / 1000);
 
             const assertion = jwtClient.signJWT({
--- a/packages/shared/lib/services/connection.service.ts
+++ b/packages/shared/lib/services/connection.service.ts
@@ -1190,18 +1190,18 @@ class ConnectionService {
 
         if (provider.token_request_auth_method === 'basic') {
             if (!client_secret) {
-                throw new NangoError('missing client_secret');
+                throw new NangoError('missing_client_secret');
             }
             headers['Authorization'] = 'Basic ' + Buffer.from(client_id + ':' + client_secret).toString('base64');
         } else if (provider.token_request_auth_method === 'custom') {
             if (!client_secret) {
-                throw new NangoError('missing client_secret');
+                throw new NangoError('missing_client_secret');
             }
             params.append('username', client_id);
             params.append('password', client_secret);
         } else if (provider.token_request_auth_method === 'private_key_jwt') {
             if (!client_private_key) {
-                throw new NangoError('missing client_private_key');
+                throw new NangoError('missing_client_private_key');
             }
             let privKeyPem: string;
             let kid: string;
@@ -1223,8 +1223,6 @@ class ConnectionService {
 
             const assertion = jwtClient.signJWT({
                 payload: {
-                    alg: 'RS256',
-                    kid: kid,
                     iss: client_id,
                     sub: client_id,
                     aud: url.toString(),
@@ -1240,7 +1238,9 @@ class ConnectionService {
             params.append('client_assertion', assertion);
         } else {
             params.append('client_id', client_id);
-            params.append('client_secret', client_secret ?? '');
+            if (client_secret) {
+                params.append('client_secret', client_secret);
+            }
         }
 
         if (tokenParams) {
--- a/packages/shared/lib/services/connection.service.ts
+++ b/packages/shared/lib/services/connection.service.ts
@@ -1,3 +1,5 @@
+import { createPrivateKey } from 'crypto';
+
 import { XMLBuilder, XMLParser } from 'fast-xml-parser';
 import ms from 'ms';
 import { Agent } from 'undici';
@@ -1160,7 +1162,7 @@ class ConnectionService {
     }: {
         provider: ProviderOAuth2;
         client_id: string;
-        client_secret: string;
+        client_secret?: string | undefined;
         connectionConfig: ConnectionConfig;
         logCtx: LogContextStateless;
         client_certificate?: string | undefined;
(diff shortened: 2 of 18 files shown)


packages/shared/lib/services/connection.service.ts
```diff
@@ -1187,13 +1189,45 @@ class ConnectionService {
...
+                    sub: client_id,
+                    aud: url.toString(),
+                    iat: now,
+                    exp: now + 300,
+                    jti: uuidv4()
+                },
+                secretOrPrivateKey: privKeyPem,
+                options: { algorithm: 'RS256', header: { kid } }
+            });
+            logger.info('assertion: ' + assertion);
```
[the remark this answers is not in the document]
agusayerza: Woops, debugging log. Removing

packages/shared/lib/services/connection.service.ts
```diff
@@ -1187,13 +1189,46 @@ class ConnectionService {
...
+            if (!client_private_key) {
+                throw new NangoError('missing client_private_key');
+            }
+            const privateJwk = JSON.parse(client_private_key) as { kid: string; [key: string]: unknown };
+            const kid = privateJwk.kid;
+            // Export the JWK to PEM so signJWT can use it.
+            const privKeyPem = createPrivateKey({ key: privateJwk, format: 'jwk' }).export({ type: 'pkcs8', format: 'pem' }) as string;
+            const now = Math.floor(Date.now() / 1000);
+
+            const assertion = jwtClient.signJWT({
```
hassan254-prog, line 1224: To simplify this, maybe you could use `TWO_STEP` with `protocol: RSA`, you can have a look at `google-service-account` as an example on how to build the `headers` and the `payload`.

agusayerza: My concern with using `TWO_STEP` here is that Okta CC is a standard OAuth2 Client Credentials flow. `private_key_jwt` is just the client authentication method used within that flow. The JWT isn't really a step in the auth process, it's how the client proves its identity in a single standard token request.

Correct me if I’m wrong, but `TWO_STEP` feels like it was designed for non-standard flows that need multiple sequential HTTP calls, like `google-service-account` which is a custom JWT exchange. If we modeled Okta CC as `TWO_STEP`, we'd lose the OAuth2 semantics and make it harder to understand at a glance that this is just a standard OAuth2 provider.

hassan254-prog: `TWO_STEP` was designed to support such complex configs that require complex interpolation like client assertion generation, without having to make changes to the other auth-modes. Keeping this complex interpolation logic centralized helps simplify the overall codebase. Keeping this within `TWO_STEP` also reduces the likelihood of future changes to `oauth-cc` if we encounter another provider with a different approach to how they want their client assertion generated.
It’s also still possible to make a single HTTP call using `TWO_STEP`.

agusayerza: I hear you on the design intent, but I think using `TWO_STEP` here would be the wrong abstraction. Okta CC is a standard OAuth2 Client Credentials flow by definition: `private_key_jwt` is just the client authentication method within that single token request. There's no second step, it's one HTTP call that happens to require a signed JWT as the client assertion.

Modeling it as `TWO_STEP` would make it harder to understand at a glance what this provider actually does, and would lose the OAuth2 semantics (OAuth2ClientCredentials type, standard CC refresh path, scope handling, etc.) for something that is, unambiguously, an OAuth2 CC flow.

On the centralization argument, I think the right place to centralize `private_key_jwt` logic is within `OAUTH2_CC` itself, since it's a recognized client authentication method for OAuth2 ([Check RFC 7523, using JWT as Client authentication](https://datatracker.ietf.org/doc/html/rfc7523#page-5)). If more providers need it, they can use the same `token_request_auth_method: private_key_jwt` flag without any changes.

packages/shared/lib/services/connection.service.ts
```diff
@@ -1187,13 +1189,58 @@ class ConnectionService {
...
+                }) as string;
+            } catch (err) {
+                throw new NangoError('invalid_client_private_key_format', { cause: err });
+            }
+            const now = Math.floor(Date.now() / 1000);
+
+            const assertion = jwtClient.signJWT({
+                payload: {
+                    alg: 'RS256',
+                    kid: kid,
```
TBonnin: aren't `alg` and `kid` headers?

packages/shared/lib/services/connection.service.ts
```diff
@@ -1187,13 +1189,58 @@ class ConnectionService {
         headers['Content-Type'] = bodyFormat === 'json' ? 'application/json' : 'application/x-www-form-urlencoded';
 
         if (provider.token_request_auth_method === 'basic') {
+            if (!client_secret) {
+                throw new NangoError('missing client_secret');
+            }
             headers['Authorization'] = 'Basic ' + Buffer.from(client_id + ':' + client_secret).toString('base64');
         } else if (provider.token_request_auth_method === 'custom') {
+            if (!client_secret) {
+                throw new NangoError('missing client_secret');
```
TBonnin: let's stick to snake case like below ex: `invalid_client_private_key_format`

packages/shared/lib/services/connection.service.ts
```diff
@@ -1187,13 +1189,58 @@ class ConnectionService {
...
+                secretOrPrivateKey: privKeyPem,
+                options: { algorithm: 'RS256', header: { kid } }
+            });
+            params.append('client_id', client_id);
+            params.append('client_assertion_type', 'urn:ietf:params:oauth:client-assertion-type:jwt-bearer');
+            params.append('client_assertion', assertion);
         } else {
             params.append('client_id', client_id);
-            params.append('client_secret', client_secret);
+            params.append('client_secret', client_secret ?? '');
```
TBonnin: would it make sense not to append client_secret if undefined instead of defaulting to a empty string?

packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -446,6 +446,10 @@ class OAuthController {
                 }
             }
 
+            if (config.oauth_scopes && !connectionConfig['oauth_scopes']) {
+                connectionConfig['oauth_scopes'] = config.oauth_scopes;
+            }
```
TBonnin, line 450: is this potentially a breaking change?

packages/shared/lib/services/connection.service.ts
agusayerza: Good catch. Missed this pesky line on the [docs](https://developer.okta.com/docs/guides/implement-oauth-for-okta-serviceapp/main/#create-and-sign-the-jwt) (`The alg parameter goes in the JWT header rather than a claim in the payload of the body.`)

packages/shared/lib/services/connection.service.ts
agusayerza: Done

packages/shared/lib/services/connection.service.ts
```diff
@@ -1187,13 +1189,58 @@ class ConnectionService {
...
+                payload: {
+                    iss: client_id,
+                    sub: client_id,
+                    aud: url.toString(),
+                    iat: now,
+                    exp: now + 300,
+                    jti: uuidv4()
+                },
+                secretOrPrivateKey: privKeyPem,
+                options: { algorithm: 'RS256', header: { kid } }
```
TBonnin: isn't `alg` a required header too? https://www.rfc-editor.org/rfc/rfc7515.html#section-4.1

packages/server/lib/controllers/oauth.controller.ts
agusayerza: The CC token request only reads scopes from `connectionConfig['oauth_scopes']` , so integration-level scopes were silently ignored for all CC integrations, this line fixes that inconsistency.

The condition (`config.oauth_scopes` must be explicitly set AND `connectionConfig['oauth_scopes']` not already provided) makes it safe: it won't affect any CC integration where the customer hasn't deliberately configured scopes at the integration level. Happy to scope it to `private_key_jwt` only if you'd prefer to be conservative, but it feels like a broader fix rather than a breaking change to me. WDYT?

packages/shared/lib/services/connection.service.ts
agusayerza: [Set by the lib](https://github.com/auth0/node-jsonwebtoken/blob/v9.0.3/sign.js#L98) when sending the `algorithm: 'RS256'` option

agusayerza: Could also use the same to set the kid tbh

agusayerza: Updated for consistency

TBonnin (approved):

packages/webapp/public/images/template-logos/okta-cc.svg
```diff
@@ -0,0 +1 @@
+okta.svg
```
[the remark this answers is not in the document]
agusayerza: Rejected. Its a symlink

PR #5496, opened 2026-02-20
fix(webapp): integrations cache invalidation

kaposke:
This PR migrates all the integration related hooks and methods to use tanstack-query, which we have been defaulting to. Mutation queries also invalidate cache for their related resources, so everything stays up to date with every update.

Fixes these problems reported by Robin https://nangohq.slack.com/archives/C06MTTJHHL1/p1771537801407719

---

It also propagates the TanStack Query contracts through connection creation and settings flows so that list, detail, flow, and other dependent screens remain synchronized without manual cache clearing.

<details>

• `CriticalErrorAlert` now hides underlying API error details on the integration show page, reducing debuggability compared to the prior `ErrorPageComponent`.
• `CreateConnectionSelector` still uses SWR cache flushing for connections while relying on TanStack invalidations for integrations, which could lead to inconsistent refresh behavior.
• Detail (`useGetIntegration`) and flow (`useGetIntegrationFlows`) queries no longer refetch on intervals, so long-lived pages may show stale data unless another mutation invalidates them.
• Most toast messages now use a generic fallback string and do not expose server-provided error messages, potentially obscuring actionable feedback.

</details>

---

### DIFF
--- a/packages/webapp/src/hooks/useIntegration.tsx
+++ b/packages/webapp/src/hooks/useIntegration.tsx
@@ -1,4 +1,4 @@
-import { useQuery } from '@tanstack/react-query';
+import { useMutation, useQuery, useQueryClient } from '@tanstack/react-query';
 import useSWR from 'swr';
 
 import { APIError, apiFetch, swrFetcher } from '../utils/api';
@@ -50,16 +50,25 @@ export async function apiPostIntegration(env: string, body: PostIntegration['Bod
     };
 }
 
-export async function apiPatchIntegration(env: string, integrationId: string, body: PatchIntegration['Body']) {
-    const res = await apiFetch(`/api/v1/integrations/${integrationId}?env=${env}`, {
-        method: 'PATCH',
-        body: JSON.stringify(body)
+export function usePatchIntegration(env: string, integrationId: string) {
+    const queryClient = useQueryClient();
+    return useMutation<PatchIntegration['Success'], APIError, PatchIntegration['Body']>({
+        mutationFn: async (body) => {
+            const res = await apiFetch(`/api/v1/integrations/${integrationId}?env=${env}`, {
+                method: 'PATCH',
+                body: JSON.stringify(body)
+            });
+            const json = (await res.json()) as PatchIntegration['Reply'];
+            if (!res.ok || 'error' in json) {
+                throw new APIError({ res, json });
+            }
+            return json;
+        },
+        onSuccess: () => {
+            void queryClient.invalidateQueries({ queryKey: ['integrations', env, integrationId] });
+            void queryClient.invalidateQueries({ queryKey: ['integrations', env] });
+        }
     });
-
-    return {
-        res,
-        json: (await res.json()) as PatchIntegration['Reply']
-    };
 }
 
 export async function apiDeleteIntegration(env: string, integrationId: string) {
--- a/packages/webapp/src/hooks/useIntegration.tsx
+++ b/packages/webapp/src/hooks/useIntegration.tsx
@@ -38,16 +38,24 @@ export function useGetIntegration(env: string, integrationId: string) {
     });
 }
 
-export async function apiPostIntegration(env: string, body: PostIntegration['Body']) {
-    const res = await apiFetch(`/api/v1/integrations?env=${env}`, {
-        method: 'POST',
-        body: JSON.stringify(body)
+export function usePostIntegration(env: string) {
+    const queryClient = useQueryClient();
+    return useMutation<PostIntegration['Success'], APIError, PostIntegration['Body']>({
+        mutationFn: async (body) => {
+            const res = await apiFetch(`/api/v1/integrations?env=${env}`, {
+                method: 'POST',
+                body: JSON.stringify(body)
+            });
+            const json = (await res.json()) as PostIntegration['Reply'];
+            if (!res.ok || 'error' in json) {
+                throw new APIError({ res, json });
+            }
+            return json;
+        },
+        onSuccess: () => {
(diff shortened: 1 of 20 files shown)


packages/webapp/src/hooks/useIntegration.tsx
```diff
@@ -1,97 +1,118 @@
...
+            });
+            const json = (await res.json()) as PostIntegration['Reply'];
+            if (!res.ok || 'error' in json) {
+                throw new APIError({ res, json });
+            }
+            return json;
+        },
+        onSuccess: () => {
+            void queryClient.invalidateQueries({ queryKey: ['integrations', env] });
+        }
```
kaposke, line 54: This is the sauce in mutations that invalidate the keys and forces stale data to revalidate.

packages/webapp/src/pages/Integrations/providerConfigKey/Settings/components/AppAuthSettings.tsx
```diff
@@ -18,20 +18,21 @@ export const AppAuthSettings: React.FC<{ data: GetIntegration['Success']['data']
...
+        try {
+            await patchIntegration({
+                authType: template.auth_mode as Extract<typeof template.auth_mode, 'APP'>,
+                ...field
+            });
             toast({ title: 'Successfully updated', variant: 'success' });
+        } catch {
+            const message = 'Failed to update, an error occurred';
+            toast({ title: message, variant: 'error' });
+            throw new Error(message);
```
TBonnin, line 35: all but `McpOAuthSettings` https://github.com/NangoHQ/nango/pull/5496/changes#diff-11a1c20f3c9b987835bdf9fcaf69408e9ce864ec7db23fd8fdfdc6d3b3ade1c1R55 are swallowing the caught error. Is it on purpose?

packages/webapp/src/pages/Integrations/providerConfigKey/Functions/One.tsx
```diff
@@ -31,10 +31,12 @@ export const FunctionsOne: React.FC = () => {
     const { providerConfigKey, functionName } = useParams();
 
     const env = useStore((state) => state.env);
-    const { data: integrationData, loading: integrationLoading } = useGetIntegration(env, providerConfigKey!);
-    const { data, loading: flowsLoading } = useGetIntegrationFlows(env, providerConfigKey!);
+    const { data: integrationResponse, isPending: integrationLoading } = useGetIntegration(env, providerConfigKey!);
```
TBonnin: we seems to use `isLoading` name in most places. Can we be consistent?

packages/webapp/src/hooks/useIntegration.tsx
```diff
@@ -1,101 +1,112 @@
...
+        queryKey: ['integrations', env, integrationId],
+        queryFn: async (): Promise<GetIntegration['Success']> => {
+            const res = await apiFetch(`/api/v1/integrations/${integrationId}?env=${env}`, { method: 'GET' });
+            const json = (await res.json()) as GetIntegration['Reply'];
+            if (!res.ok || 'error' in json) {
+                throw new APIError({ res, json });
+            }
+            return json;
+        },
+        enabled: Boolean(env && integrationId)
```
TBonnin, line 34: no refetchInterval?

packages/webapp/src/pages/Integrations/providerConfigKey/Show.tsx
```diff
@@ -20,15 +20,17 @@ import { useStore } from '@/store';
...
-    const { data, loading: loadingIntegration, error } = useGetIntegration(env, providerConfigKey!);
 
-    if (error) {
-        return <ErrorPageComponent title="Integration" error={error} />;
+    const { environmentAndAccount, loading: loadingEnvironment, error: environmentError } = useEnvironment(env);
+    const { data, isPending: loadingIntegration, error: integrationError } = useGetIntegration(env, providerConfigKey!);
+    const integration = data?.data;
+
+    if (integrationError || environmentError) {
+        return <CriticalErrorAlert message="Something went wrong while loading the integration" />;
```
TBonnin, line 30: we use to show the error before instead of a generic one.

TBonnin (approved):

packages/webapp/src/pages/Integrations/providerConfigKey/Settings/components/AppAuthSettings.tsx
kaposke: They re-throw because `EditableInput` and `ScopesInput` handle errors (don't stop editing, show error, cancel loading, etc). Not sure if that answers your question

packages/webapp/src/hooks/useIntegration.tsx
kaposke: I don't think it's necessary? Even the one on list integrations is only there because I kept old behavior, but not sure it's something that needs or is expected to auto-refresh

packages/webapp/src/pages/Integrations/providerConfigKey/Show.tsx
kaposke: Sorry, did not understand what you mean

TBonnin: sorry. I edited my comment. I mean that before this change we use to show the error message instead of a generic message

kaposke: Oh yeah. Well, inside that ErrorComponent it only shows the error if it's the generic error message with the id. Do you think it's useful?

PR #5498, opened 2026-02-21
feat(webapp): redesign toasts

kaposke:
> This depends on some changes from my other PR, so it's queued after it

I was bothered so I redesigned the toasts. I couldn't find the design in figma, but I remember it being similar to the Alerts. We can iterate later but this is a huge improvement.

Used `sonner` (shadcn now recommends this, and I found it super easy to work with). 
I managed to keep te interface exactly the same, so nothing broke and we get new toasts :)

All variations:
image

When there are multiple, they collapse unless you hover (this example is exaggerated, max is 3):
image

---

Toast notifications now flow through a dedicated components-v2/Toast built on the alert visual system, with a streamlined useToast wrapper around sonnerToast.custom, alert variants renamed to error/error-secondary and applied in consumers like CriticalErrorAlert, the Auth tab, and the application shell, while the legacy react-toastify container remains temporarily mounted for older flows.

<details>

• Replaced the reducer-based implementation in `packages/webapp/src/hooks/useToast.tsx` with a minimal wrapper around `sonnerToast.custom`, exposing only the `toast` helper.
• Deleted the shadcn `Toaster` component and mounted `sonner`'s `<Toaster />` in `packages/webapp/src/App.tsx`, keeping `<ToastContainer />` as a stopgap for legacy flows.
• Added `packages/webapp/src/components-v2/Toast.tsx`, rendering alert-styled toasts with variant-specific icons and dismiss buttons wired to `sonnerToast.dismiss`.
• Retrofitted `packages/webapp/src/components-v2/ui/alert.tsx` to rename `destructive` variants to `error`, adjust grid behavior, and support the toast layout; updated downstream usages (e.g., `CriticalErrorAlert`, `AuthTab`) accordingly.
• Introduced the `sonner` dependency in `packages/webapp/package.json` / `package-lock.json`.

</details>

<details>

• `useToast` no longer exposes `dismiss` or state, so any outstanding consumer relying on those helpers will break.
• Components still using `variant="destructive"` or `'destructive-secondary'` will now render incorrectly because those variants were removed.
• Keeping both `sonner`'s `Toaster` and `react-toastify`'s `ToastContainer` could cause duplicate notifications if the same event is routed through both systems.

</details>

---


packages/webapp/src/components-v2/Toast.tsx
```diff
@@ -0,0 +1,42 @@
...
+        <Alert variant={variant} className="w-[350px]">
+            {icon}
+            {title && description ? (
+                <>
+                    <AlertTitle className="text-body-medium-semi">{title}</AlertTitle>
+                    <AlertDescription className="text-body-medium-regular">{description}</AlertDescription>
+                </>
+            ) : (
+                (title || description) && <AlertDescription className="text-body-medium-regular">{title || description}</AlertDescription>
+            )}
```
TBonnin, line 33: this is a bit awkward. can we simply do
```
  {title && <AlertTitle className="text-body-medium-semi">{title}</AlertTitle>}
  {description && <AlertDescription className="text-body-medium-regular">{description}</AlertDescription>}
```
and let the css selectors deal with it

TBonnin (approved):

packages/webapp/src/components-v2/Toast.tsx
kaposke: I actually specifically want title to be rendered as a description when it's by itself. I find that a standalone description looks more correct than a standalone title. Doing it this way makes it so the current interface used everywhere (`toast({ title: "Something went wrong", variant: "error })`) can look still look right.

PR #5503, opened 2026-02-23
fix(lambda): add lambda destination for failures

rossmcewan:
---

**Add Lambda async failure destination processing via SQS listener**

This PR introduces a Lambda async failure handling pipeline by configuring Lambda failure destinations to SQS, adding a new SQS event listener, and a Lambda invocation processor that parses failure payloads and routes them into existing error handling. It also wires the new processor into the jobs service lifecycle and adds support for the `LAMBDA_FAILURE_DESTINATION` environment variable.

Supporting changes include centralizing `nangoProps` validation into a shared schema, extending error typing for Lambda runtime failures, improving Lambda cancellation to set abort flags, and adding unit tests for the SQS listener and invocation processor. A new dependency on `@aws-sdk/client-sqs` is added.

<details>

• Added `SqsEventListener` and `EventListener` interface to consume SQS failure messages with abort handling in `packages/jobs/lib/events/sqs.listener.ts` and `packages/jobs/lib/events/listener.ts`
• Implemented `LambdaInvocationsProcessor` to parse Lambda failure payloads and call `handleError` in `packages/jobs/lib/invocations/lambda.processor.ts`
• Configured Lambda async failure destinations via `PutFunctionEventInvokeConfigCommand` in `packages/jobs/lib/runner/lambda.ts` and added `LAMBDA_FAILURE_DESTINATION` env parsing in `packages/utils/lib/environment/parse.ts`
• Centralized `nangoPropsSchema` in `packages/jobs/lib/schemas/nango-props.ts` and reused it in `packages/jobs/lib/routes/tasks/putTask.ts`
• Updated Lambda cancellation to set abort flags via `setAbortFlag` in `packages/jobs/lib/runtime/lambda.adapter.ts` and exported `setAbortFlag` in `packages/jobs/lib/execution/operations/abort.ts`
• Extended runtime error types for Lambda failures in `packages/shared/lib/utils/error.ts`
• Added unit tests in `packages/jobs/lib/events/sqs.listener.unit.test.ts` and `packages/jobs/lib/invocations/lambda.processor.unit.test.ts`

</details>

<details>

• SQS messages are deleted even when `onMessage` throws, which can drop failures and prevent retries in `packages/jobs/lib/events/sqs.listener.ts`
• `invocationsProcessor.start()` is fire-and-forget; listener startup failures could be unhandled unless explicitly caught in `packages/jobs/lib/app.ts`
• Lambda failure type mapping relies on string matching (`signal: killed`, `Task timed out`), which could misclassify other failure modes

</details>

---

### DIFF
--- a/packages/jobs/lib/app.ts
+++ b/packages/jobs/lib/app.ts
@@ -36,6 +36,7 @@ try {
     const srv = server.listen(port);
     logger.info(`🚀 service ready at http://localhost:${port}`);
     const processor = new Processor(orchestratorUrl);
+    const invocationsProcessor = new LambdaInvocationsProcessor();
 
     // We are using a setTimeout because we don't want overlapping setInterval if the DB is down
     let healthCheck: NodeJS.Timeout | undefined;
@@ -77,6 +78,7 @@ try {
             await db.knex.destroy();
             await db.readOnly.destroy();
             await destroyKvstore();
+            await invocationsProcessor.stop();
 
             console.info('Closed');
 
@@ -105,7 +107,6 @@ try {
 
     processor.start();
 
-    const invocationsProcessor = new LambdaInvocationsProcessor();
     await invocationsProcessor.start();
 
     void otlp.register(getOtlpRoutes);
--- a/packages/jobs/lib/app.ts
+++ b/packages/jobs/lib/app.ts
@@ -8,6 +8,7 @@ import { getOtlpRoutes } from '@nangohq/shared';
 import { getLogger, initSentry, once, report, stringifyError } from '@nangohq/utils';
 
 import { envs } from './env.js';
+import { InvocationsProcessor } from './invocations/processor.js';
 import { Processor } from './processor/processor.js';
 import { getDefaultFleet, startFleets, stopFleets } from './runtime/runtimes.js';
 import { server } from './server.js';
@@ -103,6 +104,9 @@ try {
     await startFleets();
 
     processor.start();
+
+    const invocationsProcessor = new InvocationsProcessor();
+    await invocationsProcessor.start();
 
     void otlp.register(getOtlpRoutes);
 } catch (err) {
--- a/packages/jobs/lib/app.ts
+++ b/packages/jobs/lib/app.ts
@@ -8,7 +8,7 @@ import { getOtlpRoutes } from '@nangohq/shared';
 import { getLogger, initSentry, once, report, stringifyError } from '@nangohq/utils';
 
 import { envs } from './env.js';
-import { InvocationsProcessor } from './invocations/processor.js';
+import { LambdaInvocationsProcessor } from './invocations/lambda.processor.js';
 import { Processor } from './processor/processor.js';
 import { getDefaultFleet, startFleets, stopFleets } from './runtime/runtimes.js';
 import { server } from './server.js';
@@ -105,7 +105,7 @@ try {
 
     processor.start();
 
-    const invocationsProcessor = new InvocationsProcessor();
+    const invocationsProcessor = new LambdaInvocationsProcessor();
     await invocationsProcessor.start();
 
     void otlp.register(getOtlpRoutes);
--- a/packages/jobs/lib/app.ts
+++ b/packages/jobs/lib/app.ts
@@ -107,7 +107,7 @@ try {
 
     processor.start();
 
-    await invocationsProcessor.start();
+    invocationsProcessor.start();
 
     void otlp.register(getOtlpRoutes);
 } catch (err) {
--- a/packages/jobs/lib/events/sqs.listener.ts
+++ b/packages/jobs/lib/events/sqs.listener.ts
@@ -22,6 +22,7 @@ const getRegion = (): string => {
 
 export class SqsEventListener implements EventListener {
     private readonly client: SQSClient;
+    private abortController: AbortController | null = null;
 
     constructor() {
         this.client = new SQSClient({ region: getRegion() });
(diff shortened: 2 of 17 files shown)


kaposke (approved):
lgtm

packages/jobs/lib/events/sqs.listener.ts
```diff
@@ -0,0 +1,119 @@
...
+}
+
+const getRegion = (): string => {
+    const env = typeof process !== 'undefined' ? process.env['LAMBDA_REGION'] : undefined;
+    return env ?? 'us-west-2';
+};
+
+export class SqsEventListener implements EventListener {
+    private readonly client: SQSClient;
+    private abortController: AbortController | null = null;
```
TBonnin: we can instantiate the abort controller here instead of in `listen`? Is it bringing anything that it can be null?

packages/jobs/lib/events/sqs.listener.ts
```diff
@@ -0,0 +1,119 @@
...
+                    });
+                    await new Promise((r) => setTimeout(r, 1000));
+                }
+            }
+        } finally {
+            this.abortController = null;
+        }
+    }
+
+    private async getQueueUrl(queueName: string): Promise<string | null> {
```
TBonnin: can we use Result instead of null and leave the error handling to the caller so it is all in one place

packages/jobs/lib/events/sqs.listener.ts
```diff
@@ -0,0 +1,119 @@
...
+                            });
+                        }
+
+                        await this.client.send(
+                            new DeleteMessageCommand({
+                                QueueUrl: queueUrl,
+                                ReceiptHandle: msg.ReceiptHandle
+                            }),
+                            { abortSignal: signal }
+                        );
```
TBonnin, line 81: messages that throw won't be deleted. Can they end up blocking the queue?

packages/jobs/lib/events/sqs.listener.ts
```diff
@@ -0,0 +1,119 @@
...
+                    }
+                } catch (err) {
+                    if (err instanceof Error && err.name === 'AbortError') {
```
TBonnin: is this error coming from the aws sdk?

packages/jobs/lib/invocations/lambda.processor.ts
```diff
@@ -0,0 +1,66 @@
...
+import { SqsEventListener } from '../events/sqs.listener.js';
+import { handleError } from '../execution/operations/handler.js';
+import { nangoPropsSchema } from '../schemas/nango-props.js';
+
+import type { EventListener, QueueMessage } from '../events/listener.js';
+import type { NangoProps } from '@nangohq/types';
+
+function lambdaErrorTypeFromMessage(errorMessage: string): 'function_runtime_out_of_memory' | 'function_runtime_timed_out' | 'function_runtime_other' {
+    if (errorMessage.includes('signal: killed')) return 'function_runtime_out_of_memory';
+    if (errorMessage.includes('Task timed out')) return 'function_runtime_timed_out';
```
TBonnin, line 14: are the errors exposed by AWS lambda js sdk? Would be better to get it from there instead of relying on hardcoded messages

packages/jobs/lib/invocations/lambda.processor.ts
```diff
@@ -0,0 +1,66 @@
...
+                }),
+                responsePayload: z.object({
+                    errorMessage: z.string()
+                }),
+                requestPayload: z.object({
+                    taskId: z.string(),
+                    nangoProps: nangoPropsSchema
+                })
+            })
+            .parse(JSON.parse(message.body));
```
TBonnin: zod schema creation can become expensive quite fast. Best practice is to create the schema once as a local variable

packages/shared/lib/utils/error.ts
```diff
@@ -534,6 +534,21 @@ export class NangoError extends NangoInternalError {
                 this.message = 'Environment does not have a default API secret';
                 break;
 
+            case 'function_runtime_out_of_memory':
+                this.status = 500;
+                this.message = 'The function runtime ran out of memory';
```
TBonnin, line 539: would be nice to show how much memory was provisioned.

packages/jobs/lib/app.ts
```diff
@@ -104,6 +107,8 @@ try {
 
     processor.start();
 
+    await invocationsProcessor.start();
```
TBonnin: isn't this blocking since start call listen which is a while(true) loop?

packages/utils/lib/environment/parse.ts
```diff
@@ -417,6 +417,7 @@ export const ENVS = z.object({
     LAMBDA_FUNCTION_ALIAS: z.string().optional().default('latest'),
     LAMBDA_PROVISIONED_CONCURRENCY: z.coerce.number().optional().default(1),
     LAMBDA_PROVISIONED_CONCURRENCY_SCALING_TARGET: z.coerce.number().optional().default(0.7),
+    LAMBDA_FAILURE_DESTINATION: z.url().optional(),
```
TBonnin: isn't it an ARN?

packages/jobs/lib/invocations/lambda.processor.ts
```diff
@@ -0,0 +1,66 @@
...
+    if (errorMessage.includes('signal: killed')) return 'function_runtime_out_of_memory';
+    if (errorMessage.includes('Task timed out')) return 'function_runtime_timed_out';
+    return 'function_runtime_other';
+}
+
+export class LambdaInvocationsProcessor {
+    private eventListener: EventListener;
+
+    constructor() {
+        this.eventListener = new SqsEventListener();
```
TBonnin: should it only be instantiated in start when `LAMBDA_FAILURE_DESTINATION` is set

packages/jobs/lib/events/sqs.listener.ts
```diff
@@ -0,0 +1,119 @@
...
+
+export class SqsEventListener implements EventListener {
+    private readonly client: SQSClient;
+    private abortController: AbortController | null = null;
+
+    constructor() {
+        this.client = new SQSClient({ region: getRegion() });
+    }
+
+    async listen(queue: string, onMessage?: (message: QueueMessage) => void | Promise<void>): Promise<void> {
```
TBonnin: slightly different signature compared to the interface

packages/jobs/lib/events/sqs.listener.ts
```diff
@@ -0,0 +1,119 @@
...
+            });
+            return null;
+        }
+    }
+
+    async stop(): Promise<void> {
+        if (this.abortController) {
+            this.abortController.abort();
+        }
+        this.client.destroy();
```
TBonnin, line 122: should we wait for in-flight messages to be processed before destroying?

packages/jobs/lib/events/sqs.listener.ts
rossmcewan: Just to be clear where you mean - messages that throw in the `onMessage` handler or where the delete fails?

packages/jobs/lib/events/sqs.listener.ts
rossmcewan: If the abort controller is used then any unsettled promises are rejected with an error with name 'AbortError'

packages/jobs/lib/invocations/lambda.processor.ts
rossmcewan: I wish they were. That's why I'm parsing the messages. The payload received by SQS doesn't have defined error types.

packages/shared/lib/utils/error.ts
rossmcewan: The message doesn't include that information (AFAIK). We would know it from the function though - but would have to do some work to either parse it out of the function name or look it up.

packages/jobs/lib/app.ts
rossmcewan: Oof! Yes!! Fixed.

packages/jobs/lib/invocations/lambda.processor.ts
rossmcewan: Changed.

packages/jobs/lib/events/sqs.listener.ts
rossmcewan: I don't think so, at least not how its written now. The receiver and deleter both take the abort signal and will reject calls with an AbortError. If we want to process all in flight, will probably have to change that. The message handlers should be idempotent to receive the message again when the listener restarts. 

Can discuss if you disagree.

packages/jobs/lib/events/sqs.listener.ts
TBonnin: yep that's what I meant

packages/jobs/lib/invocations/lambda.processor.ts
TBonnin: :( are we sure for instance that `signal: killed` only means OOM failure? If not maybe the error should be a generic `function_runtime_killed` and the error message to mention potential OOM issue. I am worried about exposing the wrong cause of the issue to customers

TBonnin (approved):

packages/jobs/lib/invocations/lambda.processor.ts
rossmcewan: From the research I did, this is the primary cause of `signal: killed`.  I couldn't find any examples from other causes.

packages/jobs/lib/events/sqs.listener.ts
rossmcewan: I have added a separate try catch around the delete.

- If the handler fails, we delete the message anyway - we catch the error, report it, and move on. The task manager will mark these as failed anyway, just not immediately and we may lose the state for this one invocation. We could add a metric here and monitor it.
- If the delete fails, the message will get processed again on the next iteration (handling should be idempotent and in this case will likely fail with an invalid state transition - will still get deleted though)

PR #5505, opened 2026-02-23
fix(server): early exit on failing refresh token connections

agusayerza:
When a connection refresh fails, subsequent refresh attempts are skipped for 30 seconds after the last failure. This prevents high-frequency syncs and action calls from hammering a failing connection on every request. After 30 seconds, the refresh is attempted again normally. Force refreshes (manual) always bypass this cooldown.

During the cooldown window, the proxy returns HTTP 424 with error code connection_refresh_backoff instead of counting it as a PROXY_FAILURE, keeping metrics clean.

- Why 30 seconds?
Short enough to recover quickly from transient provider outages, while still dramatically reducing the blast radius of a burst failure. A 30s cooldown reduces ~60k daily attempts down to ~3k.

**Context: INC-88**
A customer had a single connection generating ~60k failed refresh attempts in a day. Their sync was retrying the refresh on every request after receiving a failed auth webhook. With this change, only ~3k of those attempts will actually hit the provider, and those will be returned as 424s rather than counted as server errors in metrics.

What changed:
  - `connection_refresh_backoff`: new error type for the cooldown early-exit, distinct from the existing `connection_refresh_exhausted` (which fires after 4 consecutive days of failures)
  - `REFRESH_FAILURE_COOLDOWN_MS`: configurable constant for the cooldown window (currently 30s)
  - Proxy returns 424 with `connection_refresh_backoff` code and skips `PROXY_FAILURE` metric increment during the cooldown window

NAN-4634
INC-88

---

Supporting timing constants were standardized to milliseconds, and the rollout is backed by integration and unit tests that validate cooldown expiry, refresh exhaustion handling, and the instant-refresh exceptions to ensure the backoff behaves reliably.

<details>

• Services other than the proxy that call `refreshOrTestCredentials` may need explicit handling for `connection_refresh_backoff` to avoid surfacing generic 5xxs.
• Cooldown responses skip `metrics.Types.PROXY_FAILURE`; ensure alternative metrics capture frequency of backoffs for observability.

</details>

---

### DIFF
--- a/packages/server/lib/controllers/proxy/allProxy.ts
+++ b/packages/server/lib/controllers/proxy/allProxy.ts
@@ -147,13 +147,13 @@ export const allPublicProxy = asyncWrapper<AllPublicProxy>(async (req, res, next
         });
         if (credentialResponse.isErr()) {
             const err = credentialResponse.error;
+            void logCtx.error('Failed to get connection credentials', { error: err });
+            await logCtx.failed();
             if (err.type === 'connection_refresh_backoff') {
                 res.status(err.status).send({ error: { code: err.type, message: err.message } });
             } else {
-                void logCtx.error('Failed to get connection credentials', { error: err });
-                await logCtx.failed();
                 metrics.increment(metrics.Types.PROXY_FAILURE);
-                res.status(400).send({ error: { code: 'server_error', message: `Failed to get connection credentials: '${err.message}'` } });
+                res.status(err.status).send({ error: { code: 'server_error', message: `Failed to get connection credentials: '${err.message}'` } });
             }
             return;
         }
--- a/packages/server/lib/controllers/proxy/allProxy.ts
+++ b/packages/server/lib/controllers/proxy/allProxy.ts
@@ -146,12 +146,15 @@ export const allPublicProxy = asyncWrapper<AllPublicProxy>(async (req, res, next
             onRefreshFailed: connectionRefreshFailed
         });
         if (credentialResponse.isErr()) {
-            void logCtx.error('Failed to get connection credentials', { error: credentialResponse.error });
-            await logCtx.failed();
-            metrics.increment(metrics.Types.PROXY_FAILURE);
-            res.status(400).send({
-                error: { code: 'server_error', message: `Failed to get connection credentials: '${credentialResponse.error.message}'` }
-            });
+            const err = credentialResponse.error;
+            if (err.type === 'connection_refresh_backoff') {
+                res.status(err.status).send({ error: { code: err.type, message: err.message } });
+            } else {
+                void logCtx.error('Failed to get connection credentials', { error: err });
+                await logCtx.failed();
+                metrics.increment(metrics.Types.PROXY_FAILURE);
+                res.status(400).send({ error: { code: 'server_error', message: `Failed to get connection credentials: '${err.message}'` } });
+            }
             return;
         }
 
--- a/packages/shared/lib/services/connections/credentials/refresh.integration.test.ts
+++ b/packages/shared/lib/services/connections/credentials/refresh.integration.test.ts
@@ -290,4 +290,119 @@ describe('refreshOrTestCredentials', () => {
         expect(decryptedUpdatedConnection.last_fetched_at?.getTime()).toBeGreaterThan(decryptedConnection.last_fetched_at!.getTime());
         expect(decryptedUpdatedConnection.updated_at?.getTime()).toBeGreaterThan(decryptedConnection.updated_at.getTime());
     });
+
+    it('should return early with connection_refresh_exhausted if refresh_exhausted is set', async () => {
+        const { env, account } = await seedAccountEnvAndUser();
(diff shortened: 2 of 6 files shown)


packages/server/lib/controllers/proxy/allProxy.ts
```diff
@@ -146,12 +146,15 @@ export const allPublicProxy = asyncWrapper<AllPublicProxy>(async (req, res, next
...
-            metrics.increment(metrics.Types.PROXY_FAILURE);
-            res.status(400).send({
-                error: { code: 'server_error', message: `Failed to get connection credentials: '${credentialResponse.error.message}'` }
-            });
+            const err = credentialResponse.error;
+            if (err.type === 'connection_refresh_backoff') {
+                res.status(err.status).send({ error: { code: err.type, message: err.message } });
+            } else {
+                void logCtx.error('Failed to get connection credentials', { error: err });
+                await logCtx.failed();
```
agusayerza: We should probably still log, right?

TBonnin: I'd think so

packages/shared/lib/services/connections/credentials/refresh.ts
```diff
@@ -9,7 +9,7 @@ import providerClient from '../../../clients/provider.client.js';
 import { NangoError } from '../../../utils/error.js';
 import { isTokenExpired } from '../../../utils/utils.js';
 import connectionService from '../../connection.service.js';
-import { REFRESH_MARGIN_S, getExpiresAtFromCredentials } from '../utils.js';
+import { REFRESH_FAILURE_COOLDOWN_MS, REFRESH_MARGIN_S, getExpiresAtFromCredentials } from '../utils.js';
```
TBonnin: can we have all duration in ms?

packages/shared/lib/services/connections/credentials/refresh.integration.test.ts
```diff
@@ -290,4 +290,119 @@ describe('refreshOrTestCredentials', () => {
...
+            onRefreshSuccess: onSuccess,
+            connectionTestHook: onTest,
+            logContextGetter: logContextGetter
+        });
+
+        // Should have attempted the refresh (and failed via test hook), not returned early
+        expect(res.isErr()).toBe(true);
+        expect(onFailed).toHaveBeenCalled();
+        expect(onTest).toHaveBeenCalled();
+    });
```
TBonnin, line 412: can we test if the following refresh outside of the cooldown period is going through?

packages/server/lib/controllers/proxy/allProxy.ts
```diff
@@ -146,12 +146,15 @@ export const allPublicProxy = asyncWrapper<AllPublicProxy>(async (req, res, next
...
+                metrics.increment(metrics.Types.PROXY_FAILURE);
+                res.status(400).send({ error: { code: 'server_error', message: `Failed to get connection credentials: '${err.message}'` } });
+            }
```
TBonnin, line 157: can we handle all errors the same way (aka: using the error status)

TBonnin (approved):

packages/shared/lib/services/connections/credentials/refresh.integration.test.ts
```diff
@@ -290,4 +291,200 @@ describe('refreshOrTestCredentials', () => {
...
+
+        // Simulate cooldown expiry by backdating last_refresh_failure
+        const connectionForBackdate = encryptionManager.decryptConnection((await connectionService.getConnectionById(connection.id))!);
+        if (!connectionForBackdate) {
+            throw new Error('Failed to decrypt connection');
+        }
+        await connectionService.updateConnection({
+            ...connectionForBackdate,
+            last_refresh_failure: new Date(Date.now() - REFRESH_FAILURE_COOLDOWN_MS - 1000)
+        });
```
kaposke: I feel like this would be a good place for vitest timers (I think `vi.setSystemTime(date)`). Don't think we've used it anywhere but it should be the best way to test time-related stuff by actually advancing in time.

TBonnin: we use vitest timers in a few places. ex: https://github.com/NangoHQ/nango/blob/ba93e51723f7d7f837bd4e5849aa05d358b7972e/packages/account-usage/lib/usage.integration.test.ts#L132

packages/shared/lib/services/connections/credentials/refresh.integration.test.ts
```diff
@@ -416,75 +416,75 @@ describe('refreshOrTestCredentials', () => {
...
+                onRefreshSuccess: onSuccess3,
+                connectionTestHook: onTest3,
+                logContextGetter
+            });
+            expect(res3.isOk()).toBe(true);
+            expect(onTest3).toHaveBeenCalled();
+            expect(onSuccess3).toHaveBeenCalled();
+        } finally {
+            vi.useRealTimers();
         }
```
TBonnin, line 446: let's move `useFakeTimers/useRealTimers` so we don't have to try/finally

PR #5513, opened 2026-02-24
chore(runner): expose error.cause in http client

TBonnin:
Expose error cause when http request from runner is failing.
Right now we only report the parent error (`fetch failed`) which doesn't give any details about why the request failed

BEFORE
Screenshot 2026-02-24 at 13 56 40

AFTER
Screenshot 2026-02-24 at 13 55 12

---

It introduces an optional cause-aware serialization path through stringifyError so the runner’s HTTP client and report logger surface nested failure details in responses and logs, with accompanying tests ensuring the behavior.

<details>

• Extended `stringifyError` signature to accept `{ cause?: boolean }` and only include the `cause` field when requested.
• Updated `report` in `packages/utils/lib/errors.ts` to log errors with both stack traces and causes when Sentry is disabled.
• Modified `packages/runner/lib/clients/http.ts` to include serialized `error.cause` in 502 responses returned after HTTP failures.
• Added a unit test ensuring `stringifyError` includes `cause` when the new option is enabled.

</details>

<details>

• `error.cause` may contain large or sensitive structures, so exposing it externally could require additional filtering.

</details>

---

### DIFF
--- a/packages/utils/lib/errors.ts
+++ b/packages/utils/lib/errors.ts
@@ -26,9 +26,9 @@ export function errorToObject(err: unknown): ErrorObject {
 /**
  * Transform any Error or primitive to a string
  */
-export function stringifyError(err: unknown, opts?: { pretty?: boolean; stack?: boolean }) {
+export function stringifyError(err: unknown, opts?: { pretty?: boolean; stack?: boolean; cause?: boolean }): string {
     const serialized = serializeError(err);
-    const allowedErrorProperties = ['name', 'message', 'provider_error_payload', ...(opts?.stack ? ['stack', 'cause'] : [])];
+    const allowedErrorProperties = ['name', 'message', 'provider_error_payload', ...(opts?.stack ? ['stack'] : []), ...(opts?.cause ? ['cause'] : [])];
 
     const enriched: Record<string, unknown> = {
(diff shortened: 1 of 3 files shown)


packages/utils/lib/errors.ts
```diff
@@ -26,9 +26,9 @@ export function errorToObject(err: unknown): ErrorObject {
 /**
  * Transform any Error or primitive to a string
  */
-export function stringifyError(err: unknown, opts?: { pretty?: boolean; stack?: boolean }) {
+export function stringifyError(err: unknown, opts?: { pretty?: boolean; stack?: boolean; cause?: boolean }): string {
     const serialized = serializeError(err);
-    const allowedErrorProperties = ['name', 'message', 'provider_error_payload', ...(opts?.stack ? ['stack', 'cause'] : [])];
+    const allowedErrorProperties = ['name', 'message', 'provider_error_payload', ...(opts?.stack ? ['stack'] : []), ...(opts?.cause ? ['cause'] : [])];
```
TBonnin, line 31: splitting `stack` and `cause` options because it is not super intuitive that `cause` would be included when `stack=true`

agusayerza (approved):

PR #5516, opened 2026-02-24
feat(server): conditional token refresh

hassan254-prog:

- Conditional token refresh allows tokens to be refreshed on demand, i.e., when a specific error is encountered. This can be defined in the proxy configuration or in the provider's configuration, ensuring that the next call is retried after the token is refreshed.

---

**Add conditional refresh-token retries in proxy flow**

This PR introduces configurable conditional token refresh for proxy requests, allowing retries to trigger a credential refresh based on specific HTTP status codes. The new `refreshTokenOn` option flows from provider metadata and SDK configuration through proxy config building, retry logic, and the server controller to call credential refresh before retrying, with expanded unit tests and documentation updates.

<details>

• Added `refreshTokenOn` support to proxy configuration types and builder in `packages/types/lib/proxy/api.ts` and `packages/shared/lib/services/proxy/utils.ts`
• Implemented refresh-triggered retries with `onRefreshToken`, `MAX_REFRESH_TOKEN_ATTEMPTS`, and retry budget adjustments in `packages/shared/lib/services/proxy/request.ts`
• Updated proxy controller to parse `refresh-token-on` header and invoke `refreshOrTestCredentials` before retrying in `packages/server/lib/controllers/proxy/allProxy.ts`
• Centralized refreshable auth modes in `packages/shared/lib/constants.ts` and used them in `packages/shared/lib/services/connections/credentials/refresh.ts` and `packages/shared/lib/services/proxy/retry.ts`
• Added tests for refresh-triggered retries and caps in `packages/shared/lib/services/proxy/request.unit.test.ts` and `packages/shared/lib/services/proxy/retry.unit.test.ts`, plus docs updates in `docs/spec.yaml`, `docs/reference/functions.mdx`, and `docs/reference/sdks/node.mdx`

</details>

<details>

• Refresh failures from `refreshOrTestCredentials` are currently ignored in `packages/server/lib/controllers/proxy/allProxy.ts`, potentially causing wasted retries with stale credentials.
• Documentation grammar for “status codes that triggers” appears in multiple docs sections and may need correction.

</details>

---

### DIFF
--- a/packages/shared/lib/services/proxy/request.ts
+++ b/packages/shared/lib/services/proxy/request.ts
@@ -82,8 +82,9 @@ export class ProxyRequest {
     public async request(): Promise<Result<AxiosResponse>> {
         let refreshTokenAttempts = 0;
 
-        const minRetriesForRefresh = this.config.refreshTokenOn?.length && this.onRefreshToken ? MAX_REFRESH_TOKEN_ATTEMPTS + 1 : 0;
-        const maxRetries = Math.max(this.config.retries ?? 0, minRetriesForRefresh);
+        // when refresh-on-error is enabled we need enough attempt slots: 1 initial try + up to MAX_REFRESH_TOKEN_ATTEMPTS refresh-and-retry cycles
+        const minAttemptSlotsForRefresh = this.config.refreshTokenOn?.length && this.onRefreshToken ? MAX_REFRESH_TOKEN_ATTEMPTS + 1 : 0;
+        const maxRetries = Math.max(this.config.retries ?? 0, minAttemptSlotsForRefresh);
 
         try {
             const response = await retryFlexible<Promise<AxiosResponse>>(
--- a/packages/shared/lib/services/proxy/request.ts
+++ b/packages/shared/lib/services/proxy/request.ts
@@ -10,10 +10,13 @@ import type { ApplicationConstructedProxyConfiguration, ConnectionForProxy, Inte
 import type { Result, RetryAttemptArgument } from '@nangohq/utils';
 import type { AxiosRequestConfig, AxiosResponse } from 'axios';
 
+const MAX_REFRESH_TOKEN_ATTEMPTS = 2;
+
 interface Props {
     proxyConfig: ApplicationConstructedProxyConfiguration;
     logger: (msg: MessageRowInsert) => MaybePromise<void>;
     onError?: (args: { err: unknown; max: number; attempt: number; retry: RetryReason }) => RetryReason;
+    onRefreshToken?: () => Promise<void>;
     getConnection: () => MaybePromise<ConnectionForProxy>;
     getIntegrationConfig: () => MaybePromise<IntegrationConfigForProxy>;
 }
@@ -45,6 +48,11 @@ export class ProxyRequest {
     onError?: Props['onError'];
 
     /**
+     * Called when retry reason is refresh_token; use to refresh credentials before the next attempt.
+     */
+    onRefreshToken?: Props['onRefreshToken'];
+
+    /**
      * Build at each iteration
      */
     axiosConfig?: AxiosRequestConfig;
@@ -63,6 +71,7 @@ export class ProxyRequest {
         this.config = props.proxyConfig;
         this.logger = props.logger;
         this.onError = props.onError;
+        this.onRefreshToken = props.onRefreshToken;
         this.getConnection = props.getConnection;
         this.getIntegrationConfig = props.getIntegrationConfig;
     }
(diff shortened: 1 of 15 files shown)


packages/shared/lib/services/proxy/retry.ts
```diff
@@ -3,6 +3,8 @@ import get from 'lodash-es/get.js';
 
 import { networkError } from '@nangohq/utils';
 
+import { REFRESHABLE_AUTH_MODES } from '../../constants.js';
```
marcindobry, line 6: Does this need to have a .js suffix? Also, once we are in double `..` territory, it's worth just importing it with an absolute path

packages/shared/lib/services/proxy/request.ts
```diff
@@ -71,6 +80,11 @@ export class ProxyRequest {
      * Send a request to the third-party with retry.
      */
     public async request(): Promise<Result<AxiosResponse>> {
+        let refreshTokenAttempts = 0;
+
+        const minRetriesForRefresh = this.config.refreshTokenOn?.length && this.onRefreshToken ? MAX_REFRESH_TOKEN_ATTEMPTS + 1 : 0;
+        const maxRetries = Math.max(this.config.retries ?? 0, minRetriesForRefresh);
```
marcindobry: Could you explain this block? Why are we assigning a MAX+1 to a variable with minimum in the name, and then assigning a min to a max again? They all have retry and/or refresh in the name, it's confusing :/

TBonnin:
@hassan254-prog I haven't follow the discussion about this feature. Can you explain why do we need it?

packages/shared/lib/services/proxy/retry.ts
hassan254-prog: Yes, it does, when using `node16` module resolution, relative imports must include the file extension, using an absolute path would mean we will need to add the package `shared` as its own dependency.

packages/shared/lib/services/proxy/request.ts
hassan254-prog: I’ve updated this with clearer [naming](https://github.com/NangoHQ/nango/pull/5516/commits/bbe9922c2803e10a60dda7579ec7d0bd2aa88fbc). The `+1` ensures there are enough attempt slots to handle the two token refreshes.

hassan254-prog:
> @hassan254-prog I haven't follow the discussion about this feature. Can you explain why do we need it?

We are addressing a performance issue with salesforce-related provider's proxy calls. Currently, every time we call `getConnection` to retrieve tokens for proxy calls, we perform an introspection call to salesforce to check if the token is still valid. This adds unnecessary overhead. 
With the new approach, we only refresh the token when a specific error code is encountered, rather than on every call. The pr to change salesforce configuration will come after this is merged.

marcindobry (approved):
LGTM but this seems like it could break something by accident and would love someone with more knowledge about the code to also have a look. I haven't worked with the proxy and retry logic yet

TBonnin:
> > @hassan254-prog I haven't follow the discussion about this feature. Can you explain why do we need it?
> 
> We are addressing a performance issue with salesforce-related provider's proxy calls. Currently, every time we call `getConnection` to retrieve tokens for proxy calls, we perform an introspection call to salesforce to check if the token is still valid. This adds unnecessary overhead. With the new approach, we only refresh the token when a specific error code is encountered, rather than on every call. The pr to change salesforce configuration will come after this is merged.

why are we always introspecting and not checking the `credentials.expires_at` we have stored in the db?

hassan254-prog:
> > > @hassan254-prog I haven't follow the discussion about this feature. Can you explain why do we need it?
> > 
> > 
> > We are addressing a performance issue with salesforce-related provider's proxy calls. Currently, every time we call `getConnection` to retrieve tokens for proxy calls, we perform an introspection call to salesforce to check if the token is still valid. This adds unnecessary overhead. With the new approach, we only refresh the token when a specific error code is encountered, rather than on every call. The pr to change salesforce configuration will come after this is merged.
> 
> why are we always introspecting and not checking the `credentials.expires_at` we have stored in the db?

No, we don’t have that info stored for salesforce, they don't return it. Their token also doesn’t include expiry details, since it includes opaque strings and not a regular jwt, which is why we need to introspect it to determine whether it has expired.

TBonnin:
> No, we don’t have that info stored for salesforce, they don't return it. Their token also doesn’t include expiry details, since it includes opaque strings and not a regular jwt, which is why we need to introspect it to determine whether it has expired.

I can see plenty of salesforce connections with a set `credentials_expires_at`? What are those?

hassan254-prog:
> From our logic, when we upsert an Oauth2 connection, the [getExpiresAtFromCredentials](https://github.com/NangoHQ/nango/blob/master/packages/shared/lib/services/connection.service.ts#L131) defaults to `now + 1d` if there is no `expires_at` value in the credentials. 

Can we set the real `expiresAt` when introspecting the token so we can rely on it to decide when to refresh or not?

hassan254-prog:
> > From our logic, when we upsert an Oauth2 connection, the [getExpiresAtFromCredentials](https://github.com/NangoHQ/nango/blob/master/packages/shared/lib/services/connection.service.ts#L131) defaults to `now + 1d` if there is no `expires_at` value in the credentials.
> 
> Can we set the real `expiresAt` when introspecting the token so we can rely on it to decide when to refresh or not?

That would work, but according to the current implementation, we can't always rely on the introspection endpoint to always return a 200 success status. This is also why we encountered a bug a few weeks ago with the error handling in the [catch](https://github.com/NangoHQ/nango/blob/master/packages/shared/lib/clients/provider.client.ts#L1193) block, which was supposed to return `true` to refresh the token.

PR #5517, opened 2026-02-24
feat: make track deletes work across execution

TBonnin:
The track deletes features currently assume the entire dataset is being fetched/saved within a single execution. 
Which means it isn't compatible with checkpoints and fetching/saving across multiple executions. 
To solve this problem, this PR is introducing a way to explicitly define the start and the end of the track deletes interval, storing the starting point in a special checkpoint that survive across executions, instead of assuming the start is always the beginning of the current execution

cc @bastienbeurier to agree on the naming `trackDeletesStart/trackDeletesEnd`

ex:
```
exec: async (nango) => {
        await nango.trackDeletesStart('MyModel');

        ...
        await nango.batchSave([...], 'MyModel');
        ...

        const deleted = await nango.trackDeletesEnd('MyModel');
}
```

---

The runner stores a per-model delete-window checkpoint keyed off the sync checkpoint and, when the window closes, uses it to invoke deletion of outdated records before clearing the checkpoint.

<details>

• Added `trackDeletesStart`/`trackDeletesEnd` to `NangoSyncBase` and `NangoSyncRunner`, with per-model checkpoint keys and `deleteOutdatedRecords` using stored `syncJobId`
• Refactored `Checkpointing` in `packages/runner/lib/sdk/checkpointing.ts` to manage per-key state and accept `key` arguments for checkpoint operations
• Expanded CLI parser/compiler validations to track `trackDeletes` calls per model and enforce ordering around `batchSave`
• Updated docs and examples to use `trackDeletesStart`/`trackDeletesEnd` and mark `deleteRecordsFromPreviousExecutions` as deprecated
• Reset behavior in `packages/shared/lib/clients/orchestrator.ts` now hard-deletes all checkpoints with a key prefix

</details>

<details>

• A run that calls `trackDeletesStart` but exits before `trackDeletesEnd` will leave the delete-window checkpoint in place for future runs
• The per-key `stateByKey` map in `Checkpointing` is not pruned, which could grow in long-lived processes with many dynamic keys
• Full reset now returns an error if `hardDeleteCheckpoints` fails, which could block user-initiated resets

</details>

---

### DIFF
--- a/docs/implementation-guides/use-cases/syncs/deletion-detection.mdx
+++ b/docs/implementation-guides/use-cases/syncs/deletion-detection.mdx
@@ -84,17 +84,13 @@ export default createSync({
 
 Syncs that fetch all records on every run can automatically detect deletions.
 
-Nango can therefore detect removals by computing the diff between two consecutive result sets. Enable this behaviour by calling the `deleteRecordsFromPreviousExecutions` function. ([full reference](/reference/functions#detect-deletions-automatically)).
-
-<Note>
-  `deleteRecordsFromPreviousExecutions` does not work with checkpoint-based syncs because fetching the data incrementally prevents performing a diff and automatically detecting deletions.
-</Note>
+Nango detects removals by computing the diff between what existed before `trackDeletesStart` and what was saved between `trackDeletesStart` and `trackDeletesEnd`. ([full reference](/reference/functions#detect-deletions-automatically)).
 
 ### Example sync with deletion detection
 
 ```ts
-import { createSync } from 'nango';
-import * as z from 'zod';
+import { createSync } from ‘nango’;
+import * as z from ‘zod’;
 
 const TicketSchema = z.object({
   id: z.string(),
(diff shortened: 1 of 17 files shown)


docs/implementation-guides/use-cases/syncs/deletion-detection.mdx
```diff
@@ -103,55 +99,57 @@ const TicketSchema = z.object({
...
 
 <Warning>
-**Be careful with exception handling when using** `deleteRecordsFromPreviousExecutions`
+**Be careful with exception handling when using** `trackDeletesStart`/`trackDeletesEnd`
 
 Nango only performs deletion detection (the “diff”) if a sync run completes successfully without any uncaught exceptions.
 
-If you’re using `deleteRecordsFromPreviousExecutions`, exception handling becomes critical:
-- If your sync doesn’t fetch the full dataset, but still call `deleteRecordsFromPreviousExecutions` (e.g. you catch and swallow an exception), Nango will attempt the diff on an incomplete dataset.
+Exception handling is critical:
```
marcindobry, line 138: Could we add an example how to handle e.g. a 500 in between the track calls?

packages/cli/lib/services/parser.service.ts
```diff
@@ -264,6 +281,33 @@ class ParserService {
...
+            if (trackDeletesStartLines.length === 0) {
+                console.log(chalk.red(`trackDeletesEnd is called but trackDeletesStart is never called in "${filePath}".`));
+                usedCorrectly = false;
+            }
+            if (batchingRecordsLines.length > 0 && batchingRecordsLines.some((line) => line > Math.min(...trackDeletesEndLines))) {
+                console.log(
+                    chalk.red(`trackDeletesEnd should be called after all batching records functions in "${filePath}:${Math.max(...trackDeletesEndLines)}".`)
+                );
+                usedCorrectly = false;
+            }
```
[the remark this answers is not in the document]
marcindobry: This is fair, we should track the lines per key

packages/runner/lib/sdk/sdk.ts
```diff
@@ -522,6 +528,40 @@ export class NangoSyncRunner extends NangoSyncBase<never, never, ZodCheckpoint>
         return res.value;
     }
 
+    public async trackDeletesStart(model: string): Promise<void> {
+        this.throwIfAborted();
+        const key = `${this.checkpointKey}:trackDeletes:${model}`;
```
marcindobry: nit: Could we have the key generation be a separate function?

marcindobry (changes requested):

packages/cli/lib/services/parser.service.ts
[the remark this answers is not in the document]
TBonnin: good point. I'll fix

docs/implementation-guides/use-cases/syncs/deletion-detection.mdx
TBonnin: not sure I follow

packages/cli/lib/services/parser.service.ts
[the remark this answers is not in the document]
TBonnin: fixed

docs/implementation-guides/use-cases/syncs/deletion-detection.mdx
marcindobry: This says that exception handling is critical and describes a scenario where an exception occurs, but to me it'd be useful to show a small piece of code that would show the issue and how it should be handled correctly / recovered from. An example would be if an API call during pagination throws a 500.

TBonnin: it is explained in plain-english below:

> - Let exceptions bubble up and interrupt the run.
> - If you’re using `try/catch`, re-throw exceptions that indicate incomplete data.

Code examples would be less succinct imho. ex:
``` ts
// let exceptions bubble up and interrupt the run
await nango.trackDeletesStart('MyModel')
for(const page = await nango.paginate(...)) { // <--- let any exception interrupt the execution.
   ...
   await nango.batchSave(...)
}
await nango.trackDeletesEnds('MyModel')

// or if you’re using `try/catch`, re-throw exceptions that indicate incomplete data
await nango.trackDeletesStart('MyModel')
try {
    for(const page = await nango.paginate(...)) {
      ...
      await nango.batchSave(...)
   }
catch(err) {
   ...
   throw ... // rethrow
}
await nango.trackDeletesEnds('MyModel')
```

packages/runner/lib/sdk/sdk.ts
```diff
@@ -522,6 +528,44 @@ export class NangoSyncRunner extends NangoSyncBase<never, never, ZodCheckpoint>
...
+        }
+
+        const res = await this.persistClient.deleteOutdatedRecords({
+            model: this.modelFullName(model),
+            environmentId: this.environmentId,
+            nangoConnectionId: this.nangoConnectionId!,
+            syncId: this.syncId!,
+            syncJobId: stored.syncJobId,
+            activityLogId: this.activityLogId
+        });
```
kaposke, line 559: It took me a long time to grasp what's hapenning tbh.
So the trick is persisting the `syncJobId` of the first execution? And then I'm guessing `deleteOutdatedRecords` deletes all records before the given `syncJobId`? 
So it's not precisely "any records before calling 'trackDeletesStart'", but "any records before the sync run that called 'trackDeletesStart'"? Should it be based on a timestamp to match the precise moment when `start` is called?

TBonnin: > So the trick is persisting the syncJobId of the first execution? And then I'm guessing deleteOutdatedRecords deletes all records before the given syncJobId?

You are right. The behavior of `deleteOutdatedRecords` didn't change. It was already marking as deleted all records with `syncJobId < currentSyncJobId`, assuming that all records will be saved in a single execution. This assumption might not be true anymore when we start interrupting functions, hence storing the syncJobId of the first execution instead of assuming it is the current one.

> So it's not precisely "any records before calling 'trackDeletesStart'", but "any records before the sync run that called 'trackDeletesStart'"? Should it be based on a timestamp to match the precise moment when start is called?

Yep but since the compiler ensures that trackDeletesStart/End are called before/after batchSave it is equivalent

marcindobry (approved):

packages/runner/lib/sdk/sdk.ts
kaposke: I see, that makes sense.

kaposke (approved):

PR #5519, opened 2026-02-25
fix(cli): smart filtering in dryrun interactive mode

agusayerza:
## Summary

- When `--integration-id` is provided, the interactive function picker and connection picker are now filtered to that integration only
- When only a `connectionId` is provided (no `--integration-id`), the CLI looks it up via the API and filters the function picker to the matching integrations. Even when the same connection ID exists across multiple integrations, we narrow to that subset rather than showing everything
- After a function is selected interactively with no integration context, the CLI identifies which integration owns it (unambiguous case) and uses that to filter the connection picker

## How it works

The interactive flow now builds an `integrationFilter: string[]` before showing any prompt:

1. If `--integration-id` is given: filter = `[integrationId]`
2. Else if `connectionId` arg is given: call `listConnections({ connectionId })` and use all returned `provider_config_key` values as the filter (empty array = no filter applied)
3. Function picker is filtered to the integration(s) in the filter set
4. Once the function name is known, derive the owning integration from local definitions (respecting the filter, so same-named functions in other integrations are ignored)
5. Connection picker is filtered to that single integration

## Test plan

- [x] `nango dryrun --integration-id workday`: function and connection pickers show only workday entries
- [x] `nango dryrun` with no args: select a function and then connection picker filtered to that function's integration
- [x] `nango dryrun my-sync` pick integration if unknown, connection picker filtered to that function's integration
- [x] `nango dryrun my-sync my-conn-id -e {env}`: no prompts, runs directly
- [x] Unit tests: `npm run test:unit -- packages/cli`

Before:
Screenshot 2026-02-24 at 11 32 57 PM

After:
Screenshot 2026-02-24 at 11 32 00 PM

---

It also adds early validation for missing integrations or scripts, centralizes host resolution, standardizes error logging, and safely skips environment fetching when `NANGO_SECRET_KEY` is absent.

<details>

• `inferIntegrationsFromConnectionId` swallows API errors and returns an empty array, which can silently fall back to unfiltered prompts.
• If `connectionId` exists across multiple integrations, ensure `integrationFilter` is consistently applied to both function and connection prompts.

</details>

---

### DIFF
--- a/packages/cli/lib/index.ts
+++ b/packages/cli/lib/index.ts
@@ -345,17 +345,31 @@ program
             environment = await ensure.environment(environment, debug);
 
             const definitions = await buildDefinitions({ fullPath, debug });
-            if (definitions.isOk()) {
-                const functions = definitions.value.integrations
-                    .flatMap((i) => [...i.syncs, ...i.actions])
-                    .map((f) => ({ name: f.name, type: f.type as string }));
-                name = await ensure.function(name, functions);
-            } else {
+            if (definitions.isErr()) {
                 console.error(chalk.red('Could not build function definitions to select from.'));
                 process.exit(1);
             }
 
-            connectionId = await ensure.connection(connectionId, environment);
+            let integrationFilter: string[] | undefined = integrationId ? [integrationId] : undefined;
+            if (connectionId && !integrationFilter) {
+                // integration id not provided, try to infer it from the connection id
+                const inferred = await ensure.inferIntegrations(connectionId, environment);
+                if (inferred.length > 0) {
+                    integrationFilter = inferred;
+                }
+            }
+
+            // Show functions for which a connection with the given id is valid
+            const functions = definitions.value.integrations
+                .filter((i) => !integrationFilter || integrationFilter.includes(i.providerConfigKey))
+                .flatMap((i) => [...i.syncs, ...i.actions])
+                .map((f) => ({ name: f.name, type: f.type as string }));
+            name = await ensure.function(name, functions);
+
+            const selectedIntegrationId = definitions.value.integrations.find((i) =>
+                [...i.syncs, ...i.actions].some((s) => s.name === name)
+            )?.providerConfigKey;
+            connectionId = await ensure.connection(connectionId, environment, selectedIntegrationId);
         } catch (err: any) {
             console.error(chalk.red(err.message));
             if (err instanceof MissingArgumentError) {
--- a/packages/cli/lib/index.ts
+++ b/packages/cli/lib/index.ts
@@ -25,6 +25,7 @@ import { generate as generateDocs } from './services/docs.service.js';
 import { DryRunService } from './services/dryrun.service.js';
 import { Ensure } from './services/ensure.service.js';
 import { create } from './services/function-create.service.js';
+import { inferIntegrationsFromConnectionId } from './services/interactive.service.js';
 import { directoryMigration, endpointMigration, v1toV2Migration } from './services/migration.service.js';
 import { generateTests } from './services/test.service.js';
 import verificationService from './services/verification.service.js';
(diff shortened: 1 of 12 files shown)


packages/cli/lib/services/ensure.service.ts
```diff
@@ -61,12 +63,38 @@ export class Ensure {
...
+        return this.ensure(current, () => promptForConnection(environment, integrationId), 'Connection ID is required');
+    }
+
+    public async integrationForScript(scriptName: string, integrations: string[]): Promise<string> {
+        return this.ensure(undefined, () => promptForIntegration(integrations), `Multiple integrations have "${scriptName}". Please use --integration-id`);
     }
 
-    public async connection(current: string | undefined, environment: string): Promise<string> {
-        return this.ensure(current, () => promptForConnection(environment), 'Connection ID is required');
+    public async inferIntegrations(connectionId: string, environment: string): Promise<string[]> {
```
marcindobry: Do we need this function call if it's just calling `inferIntegrationsFromConnectionId` with the same arguments and same return type? Can't we call `inferIntegrationsFromConnectionId` directly wherever this is used?

marcindobry (approved):

packages/cli/lib/services/ensure.service.ts
agusayerza: You are right, unnecessary. Missed it during the refactor.

packages/cli/lib/index.ts
```diff
@@ -345,17 +347,61 @@ program
...
+                if (!allScriptNames.has(name)) {
+                    const hint = integrationFilter ? ` in integration "${integrationFilter.join(', ')}"` : '';
+                    throw new Error(`No script named "${name}" found${hint}`);
+                }
+            }
+
+            // Show functions for which a connection with the given id is valid
+            const functions = filteredIntegrations.flatMap((i) =>
+                [...i.syncs, ...i.actions].map((f) => ({ name: f.name, type: f.type as string, integration: i.providerConfigKey }))
+            );
```
TBonnin, line 393: destructuring `i` and `f` might make the function slightly more readable:
```
const functions = filteredIntegrations.flatMap(({ syncs, actions, providerConfigKey }) =>
  [...syncs, ...actions].map(({ name, type }) => ({
    name,
    type: type as string,
    integration: providerConfigKey,
  }))
);
```

packages/cli/lib/utils.ts
```diff
@@ -227,7 +235,7 @@ export async function getEnvironments(debug = false): Promise<GetEnvironments['S
         const res = await http.get(url, { headers });
         return res.data as GetEnvironments['Success'];
     } catch (err) {
-        console.log(`❌ ${err instanceof AxiosError ? err.response?.data.error : JSON.stringify(err, ['message'])}`);
+        console.log(`❌ ${JSON.stringify(err)}`);
```
TBonnin: are we losing some information about the error here? Using `stringifyError` might be better

packages/cli/lib/services/ensure.service.ts
```diff
@@ -61,12 +62,34 @@ export class Ensure {
...
+            return await promptForFunctionToRun(availableFunctions);
+        } catch (err: any) {
+            if (err.isTtyError) {
+                throw new Error(
+                    "Prompt couldn't be rendered in the current environment. Please use the --no-interactive flag and pass all required arguments."
+                );
+            }
+            if (err.name === 'ExitPromptError') {
+                console.log('Interactive prompt cancelled.');
+                process.exit(0);
```
TBonnin, line 81: why process.exit instead of letting the error being thrown?

packages/cli/lib/index.ts
```diff
@@ -345,17 +347,61 @@ program
...
             }
 
-            connectionId = await ensure.connection(connectionId, environment);
+            let integrationFilter: string[] | undefined = integrationId ? [integrationId] : undefined;
+            if (connectionId && !integrationFilter) {
+                // integration id not provided, try to infer it from the connection id
+                const inferred = await inferIntegrationsFromConnectionId(connectionId, environment);
+                if (inferred.length > 0) {
+                    integrationFilter = inferred;
+                }
```
TBonnin, line 363: would a printDebug be useful here if `inferred` is empty to make it easier to debug?

TBonnin (approved):

packages/cli/lib/services/ensure.service.ts
agusayerza: `ExitPromptError` is the error thrown by Inquirer when a `Ctrl+C` interruption is detected. We handle the error so that they don't see a stack trace when correctly exiting the CLI tool

PR #5541, opened 2026-02-27
fix(webapp): invalidate flows when toggling function

kaposke:
I migrated useGetIntegrationFlows to useQuery, but I missed this place that mutated something in flows and didn't invalidate it properly (it was still invalidating swr keys).

---

It also shifts flow enable/disable and pre-built deploy behavior in the UI to React Query mutation hooks, improving loading and API error handling while ensuring flow data is refreshed after toggling.

<details>

• Replaced SWR `mutate` and manual API calls in `packages/webapp/src/pages/Integrations/components/FunctionSwitch.tsx` with React Query mutations from `packages/webapp/src/hooks/useFlow.tsx`
• Added `usePreBuiltDeployFlow`, `useFlowEnable`, and `useFlowDisable` hooks that wrap `apiFetch` and invalidate `['integrations', env, integrationId, 'flows']` on success
• Updated enable/disable logic to use `APIError`-based error handling and shared request body creation

</details>

<details>

• Only `['integrations', env, integrationId, 'flows']` is invalidated; if other UI relies on broader integration queries, those views may remain stale.

</details>

---

### DIFF
--- a/packages/webapp/src/hooks/useFlow.tsx
+++ b/packages/webapp/src/hooks/useFlow.tsx
@@ -1,17 +1,27 @@
-import { apiFetch } from '../utils/api';
+import { useMutation, useQueryClient } from '@tanstack/react-query';
+
+import { APIError, apiFetch } from '../utils/api';
 
 import type { PatchFlowDisable, PatchFlowEnable, PatchFlowFrequency, PostPreBuiltDeploy, PutUpgradePreBuiltFlow } from '@nangohq/types';
 
-export async function apiPreBuiltDeployFlow(env: string, body: PostPreBuiltDeploy['Body']) {
-    const res = await apiFetch(`/api/v1/flows/pre-built/deploy/?env=${env}`, {
-        method: 'POST',
-        body: JSON.stringify(body)
+export function usePreBuiltDeployFlow(env: string, integrationId: string) {
+    const queryClient = useQueryClient();
+    return useMutation<PostPreBuiltDeploy['Success'], APIError, PostPreBuiltDeploy['Body']>({
+        mutationFn: async (body) => {
+            const res = await apiFetch(`/api/v1/flows/pre-built/deploy/?env=${env}`, {
+                method: 'POST',
+                body: JSON.stringify(body)
+            });
+            const json = (await res.json()) as PostPreBuiltDeploy['Reply'];
+            if (!res.ok || 'error' in json) {
+                throw new APIError({ res, json });
+            }
+            return json;
+        },
+        onSuccess: () => {
+            void queryClient.invalidateQueries({ queryKey: ['integrations', env, integrationId, 'flows'] });
+        }
     });
-
-    return {
-        res,
-        json: (await res.json()) as PostPreBuiltDeploy['Reply']
-    };
 }
 
 export async function apiPreBuiltUpgrade(env: string, body: PutUpgradePreBuiltFlow['Body']) {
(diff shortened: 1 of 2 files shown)


packages/webapp/src/hooks/useFlow.tsx
```diff
@@ -1,17 +1,27 @@
...
+                body: JSON.stringify(body)
+            });
+            const json = (await res.json()) as PostPreBuiltDeploy['Reply'];
+            if (!res.ok || 'error' in json) {
+                throw new APIError({ res, json });
+            }
+            return json;
+        },
+        onSuccess: () => {
+            void queryClient.invalidateQueries({ queryKey: ['integrations', env, integrationId, 'flows'] });
```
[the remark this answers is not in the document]
kaposke: I don't think there's data in flows that affect the integrations queries. Flows are obtained in a separate fetch. I would prefer to default to not invalidating them unless we find good reason to.

packages/webapp/src/hooks/useFlow.tsx
```diff
@@ -1,17 +1,27 @@
...
+        }
     });
-
-    return {
-        res,
-        json: (await res.json()) as PostPreBuiltDeploy['Reply']
-    };
 }
 
 export async function apiPreBuiltUpgrade(env: string, body: PutUpgradePreBuiltFlow['Body']) {
```
TBonnin, line 27: should other function be refactored to use the same pattern instead of returning the fetch response directly?

packages/webapp/src/pages/Integrations/components/FunctionSwitch.tsx
```diff
@@ -59,73 +62,58 @@ export const FunctionSwitch: React.FC<{
...
+                    provider: integration.provider,
+                    providerConfigKey: integration.unique_key,
+                    type: flow.type,
+                    scriptName: flow.name
+                }
+            });
             toast({ title: `Disabled successfully`, variant: 'success' });
-            await mutate((key) => typeof key === 'string' && key.startsWith('/api/v1/integrations'));
+        } catch {
+            toast({ title: 'An unexpected error occurred', variant: 'error' });
```
TBonnin, line 115: should we check for `APIErrror` here?

packages/webapp/src/pages/Integrations/components/FunctionSwitch.tsx
```diff
@@ -59,73 +62,58 @@ export const FunctionSwitch: React.FC<{
...
-                toast({ title: 'An unexpected error occurred', variant: 'error' });
+                await deployFlow(body);
             }
-        } else {
             toast({ title: `Enabled successfully`, variant: 'success' });
-            await mutate((key) => typeof key === 'string' && key.startsWith('/api/v1/integrations'));
-
             if (plan && plan.auto_idle && !plan.trial_end_at) {
-                await mutateEnv();
+                void mutateEnv();
```
TBonnin, line 84: is not awaiting on purpose?

TBonnin (approved):
is it possible to add a test to ensure the flow is correctly invalidated?

packages/webapp/src/pages/Integrations/components/FunctionSwitch.tsx
kaposke: Not done because none of the possible errors are user-friendly (like the `resource_capped` error in the other one).

packages/webapp/src/pages/Integrations/components/FunctionSwitch.tsx
kaposke: ai tweaked this and I missed it, but it should have no consequence. Will revert to be safe though

packages/webapp/src/hooks/useFlow.tsx
kaposke: This function is actually not used anywhere. I considered deleting it but I second guessed because I would lose track of this endpoint's existence. But I might as well tbh.

kaposke: Oh this is for bringing a template up to date to the most recent version. Might want to add this ability back at some point? Not sure. @bastienbeurier ideas?

PR #5551, opened 2026-03-02
fix(proxy): properly send error responses

hassan254-prog:

- Previously, `res.writeHead()` was called before piping the error stream to `res`, which immediately flushed headers to the client before all chunks were collected. This would at times cause clients to receive incomplete error responses.

- The fix removes the premature `res.writeHead()` call and instead collects all chunks first, then sends status, headers, and body atomically via` res` status set and send.

---

It also makes the error handler exportable and adds broader test coverage for proxy error scenarios, including streamed error bodies.

<details>

• If `error.response.data` is not a readable stream, attaching `on` handlers could throw or silently skip response; consider a stream type guard.

</details>

---


packages/server/lib/controllers/proxy/allProxy.ts
TBonnin, line 432: unrelated but we should handle `error` event, not just data and end. ex:
```
 errorData.on('error', (err) => {
      void logCtx.error('Error reading upstream error stream', ...);
      res.status(500).send();
  });
```

TBonnin: also unrelated to your change but we are shadowing errorData here, which make reading the code harder. an we rename it `parsedData` or similar?

packages/server/lib/controllers/proxy/allProxy.ts
```diff
@@ -422,12 +416,9 @@ function handleErrorResponse({
             callback(null, chunk);
         }
     });
```
TBonnin: I don't think we need this transform anymore since it is not piped into res anymore. no?

TBonnin (approved):

packages/server/lib/controllers/proxy/allProxy.ts
hassan254-prog: thanks, updated with the latest commit.

packages/server/lib/controllers/proxy/allProxy.ts
hassan254-prog: yes, we don’t need it, I've updated it with the latest commit.

packages/server/lib/controllers/proxy/allProxy.unit.test.ts
```diff
@@ -156,3 +159,153 @@ describe('handleResponse', () => {
...
+
+        handleErrorResponse({ res, error: axiosError, logCtx: mockLogCtx });
+
+        await new Promise<void>((resolve) => {
+            if (sendFn.mock.calls.length > 0) {
+                resolve();
+            } else {
+                sendFn.mockImplementationOnce(() => resolve());
+            }
+        });
```
TBonnin, line 295: can we use `stream.once('end', resolve)`. It would be simpler if yes

TBonnin (approved):

PR #5553, opened 2026-03-03
feat(integrations): add support for avanan

hassan254-prog:

- add support for avanan

---

**Add `avanan` integration with enhanced interpolation utilities**

This PR introduces the new `avanan` provider integration using `TWO_STEP` auth, including provider configuration, connection/credential schemas, proxy/token header templates, docs pages, and a logo asset. It also expands string interpolation utilities to support formatted `now` values, `sha256Hex` hashing, stable per-request `random`/`now` values, and `endpoint` interpolation, and wires those into proxy headers and token header generation.

Additional updates include safer handling of `newCredentials.raw` during refresh, new unit tests covering interpolation and proxy header stability, documentation navigation updates, and the addition of `dayjs` to support date formatting.

---


packages/providers/providers.yaml
```diff
@@ -1575,6 +1575,59 @@ auvik:
...
+    token_response:
+        token: ''
+    token_expires_in_ms: 86400000
+    proxy:
+        base_url: https://${connectionConfig.subdomain}.avanan.net
+        headers:
+            x-av-app-id: ${credentials.appId}
+            x-av-token: ${accessToken}
+            x-av-req-id: ${random}
+            x-av-date: '${now | date: "%Y-%m-%dT%H:%M:%S"}'
```
hassan254-prog: This will allow for specific date format interpolation, avanan doesn’t accept dates in zulu format.

packages/shared/lib/services/proxy/utils.ts
```diff
@@ -393,6 +394,22 @@ export function buildProxyHeaders({
 
     // Custom headers handling
     if ('proxy' in config.provider && 'headers' in config.provider.proxy) {
+        const headerValues = Object.values(config.provider.proxy.headers).filter((v): v is string => typeof v === 'string');
```
hassan254-prog, line 402: this will ensure that the dates and random numbers we generate remain consistent across all headers.

docs/api-integrations/avanan/connect.mdx
```diff
@@ -0,0 +1,49 @@
...
+| Europe | `smart-api-production-1-eu` |
+| Canada | `smart-api-production-1-ca` |
+| Australia | `smart-api-production-5-ap` |
+| United Kingdom (UK) | `smart-api-production-1-euw2` |
+| United Arab Emirates (UAE) | `smart-api-production-1-mec1` |
+| India | `smart-api-production-1-aps1` |
+
+#### Step 2: Getting Your Client ID and Secret Key
+
+1. Contact [Avanan Support](https://www.checkpoint.com/harmony/email-security/t) or your Avanan representative.
```
TBonnin: I think the link is truncated

packages/providers/providers.yaml
TBonnin: is that confirmed. the docs I found only have example with zulu format

hassan254-prog: Yes, this is confirmed and tested, if you use a zulu format date, this is the error you get.
Screenshot 2026-03-05 at 23 36 12

TBonnin: I propose we refuse to support API that don't support timezone. Life is already hard enough

packages/shared/lib/services/proxy/utils.ts
```diff
@@ -393,6 +394,22 @@ export function buildProxyHeaders({
...
+        const needsStableRandom = headerValues.some((v) => v.includes('${random}'));
+        const needsStableNow = headerValues.some((v) => v.includes('${now}') || v.includes('now | date:'));
```
TBonnin: I would prefer if the format is `{$now}` or ${now:FORMAT} making it clear is always now. (FORMAT being optional, defaults to zula time if not specified). I am not sure the piping syntax with `| date` is very understandable

packages/shared/lib/services/proxy/utils.ts
```diff
@@ -393,6 +394,22 @@ export function buildProxyHeaders({
...
+        const stableReplacers: Record<string, unknown> = {};
+
+        if (needsStableRandom || needsStableNow) {
```
TBonnin: this top if isn't necessesary

packages/shared/lib/utils/utils.ts
```diff
@@ -522,3 +547,24 @@ function removeEmptyValues(obj: Record<string, any>): Record<string, any> {
...
+    const ms = date.getUTCMilliseconds();
+    return format
+        .replace(/%Y/g, pad(year, 4))
+        .replace(/%m/g, pad(month, 2))
+        .replace(/%d/g, pad(day, 2))
+        .replace(/%H/g, pad(hours, 2))
+        .replace(/%M/g, pad(minutes, 2))
+        .replace(/%S/g, pad(seconds, 2))
+        .replace(/%L/g, pad(ms, 3));
+}
```
TBonnin, line 577: are we not already using a date library that provide this functionality?

packages/shared/lib/services/connection.service.ts
```diff
@@ -1382,6 +1382,18 @@ class ConnectionService {
...
+            const needsStableNow = headerValues.some((v) => v.includes('${now}') || v.includes('now | date:'));
+            // some providers may use now or random values which they expect these to be the same allthrough
+            if (needsStableRandom || needsStableNow) {
+                if (needsStableRandom) {
+                    dynamicCredentials['random'] = uuidv4();
+                }
+                if (needsStableNow) {
+                    dynamicCredentials['now'] = new Date().toISOString();
+                }
+            }
```
TBonnin: this logic is similar to the one in proxy utils. can we consolidate?

hassan254-prog: Yes, I have updated this with the latest [commit](https://github.com/NangoHQ/nango/pull/5553/commits/e093e6afd487f6040fb37594c57fde7f7fe5bff4).

packages/shared/lib/utils/utils.ts
hassan254-prog: I have updated this with the latest [commit](https://github.com/NangoHQ/nango/pull/5553/commits/e093e6afd487f6040fb37594c57fde7f7fe5bff4) to use `dayjs`.

packages/shared/lib/services/proxy/utils.ts
hassan254-prog: Yes, I agree. I’ve updated this to a more readable format in the latest [commit](https://github.com/NangoHQ/nango/pull/5553/commits/e093e6afd487f6040fb37594c57fde7f7fe5bff4).

packages/providers/providers.yaml
hassan254-prog: same, or APIs that don’t adhere to their documented specs.

packages/shared/lib/utils/utils.ts
```diff
@@ -326,6 +326,21 @@ export function interpolateObject(obj: Record<string, any>, dynamicValues: Recor
...
+
+    if (needsStableRandom) {
+        stableReplacers['random'] = crypto.randomUUID();
+    }
+    if (needsStableNow) {
+        stableReplacers['now'] = new Date().toISOString();
+    }
+
+    return stableReplacers;
+}
```
TBonnin, line 336: what about the more concise:
```
export function getStableInterpolationReplacers(values: string[]): Record<string, string> {
    const has = (pattern: string) => values.some((v) => v.includes(pattern));

    return {
        ...(has('${random}') && { random: crypto.randomUUID() }),
        ...(has('${now}') || has('${now:') && { now: new Date().toISOString() }),
    };
}
```

packages/shared/lib/utils/utils.ts
```diff
@@ -255,6 +265,11 @@ export function interpolateStringFromObject(str: string, replacers: Record<strin
         return Buffer.from(resolvedInner).toString('base64');
     });
 
+    str = str.replace(/\${sha256Hex\((.*?)\)}/g, (_, inner) => {
```
TBonnin: same logic as above. can we consolidate?

packages/shared/lib/utils/utils.ts
```diff
@@ -212,12 +216,18 @@ export function interpolateString(str: string, replacers: Record<string, any>):
         return getFingerprint(resolvedInner);
     });
 
+    str = str.replace(/\${sha256Hex\((.*?)\)}/g, (_, inner) => {
```
TBonnin: does this work with nested value like the avanan header? 
Can we add a test with a pattern with base64 nested into sha256Hex?

packages/shared/lib/services/connection.service.ts
```diff
@@ -1382,6 +1383,15 @@ class ConnectionService {
         const headers: Record<string, any> | string = {};
 
         if (tokenHeaders) {
+            const headerValues = Object.values(tokenHeaders).filter((v): v is string => typeof v === 'string');
+            const stableReplacers = getStableInterpolationReplacers(headerValues);
+            // some providers may use now or random values which they expect these to be the same allthrough
```
TBonnin: can you point to an example in this comment instead of using some?

packages/shared/lib/services/connection.service.ts
```diff
@@ -1392,6 +1402,10 @@ class ConnectionService {
                     headers[key] = strippedValue;
                 }
             }
+            if (stableReplacers['random'] || stableReplacers['now']) {
+                delete dynamicCredentials['random'];
+                delete dynamicCredentials['now'];
+            }
```
TBonnin: would it make more sense to pass the replacers to the interpolate functions as an optional parameter instead of mutating the credentials object?

hassan254-prog: Updated with the latest [commit](https://github.com/NangoHQ/nango/pull/5553/commits/03a29f15e4f639a4d6b943a2cd1d0fc8fff3d9b0).

packages/shared/lib/utils/utils.ts
hassan254-prog: Yes, I have added tests for this pattern.

packages/shared/lib/utils/utils.ts
hassan254-prog: Updated with the latest [commit](https://github.com/NangoHQ/nango/pull/5553/commits/03a29f15e4f639a4d6b943a2cd1d0fc8fff3d9b0).

TBonnin (approved):

PR #5555, opened 2026-03-03
feat(runner-sdk): add execution lifecycle with interrupt and kill timeouts

TBonnin:
AWS lambda runtime has a 15mins time limit. We want to avoid as much as possible for the AWS Lambda to timeout in order to prevent race conditions and undefined behaviors. (Ex: lambda times out after the exec function has returned but before or while the runner is sending back the output)

This commit extends the cancellation mechanism and introduces the concept of lifecycle timeouts:
- hard timeout checked on every runner-sdk public function (same as checking the abort signal)
- soft timeout on `saveCheckpoint` so we can treat it as a partial success that needs immediate rescheduling.

The idea will be to set the lifecycle when using Lambda runtime to something like 12mins for the soft timeout and 14mins for the hard one, allowing the function to gracefully terminate if hitting a `saveCheckpoint` or at least halt before the AWS lambda time limit.
It doesn't fully prevent any AWS Lambda timeout (ex: super long processing without calling any `nango.` function or unresponsive API) but it would greatly reduce the likelihood of the AWS Lambda to reach its 15mins limit.

The lifecycle isn't set yet so those timeouts don't trigger for now. 
In order to enable the soft timeout we need to be able to treat it as a partial success that requires a new task to be immediately scheduled to continue syncing.

### DIFF
--- a/packages/runner-sdk/lib/action.ts
+++ b/packages/runner-sdk/lib/action.ts
@@ -1,6 +1,6 @@
 import { getProvider } from '@nangohq/providers';
 
-import { AbortedSDKError, ActionError, UnknownProviderSDKError } from './errors.js';
+import { ActionError, ExecutionAbortedSDKError, ExecutionInterruptedSDKError, ExecutionTimeoutSDKError, UnknownProviderSDKError } from './errors.js';
 import paginateService from './paginate.service.js';
 
 import type { ZodCheckpoint, ZodMetadata } from './types.js';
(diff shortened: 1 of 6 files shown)


packages/runner-sdk/lib/action.ts
```diff
@@ -94,7 +99,6 @@ export abstract class NangoActionBase<
         this.activityLogId = config.activityLogId;
         this.scriptType = config.scriptType;
         this.isCLI = config.isCLI;
-        this.startTime = Date.now();
```
TBonnin, line 97: not used

packages/runner/lib/sdk/sdk.ts
```diff
@@ -305,26 +306,36 @@ export class NangoActionRunner extends NangoActionBase<never, ZodCheckpoint> {
...
     public override async tryAcquireLock(props: { key: string; ttlMs: number }): Promise<boolean> {
+        this.throwIfAbortedOrKilled();
         return this.locking.tryAcquireLock(props);
     }
 
     public override async releaseLock(props: { key: string }): Promise<boolean> {
-        return this.locking.releaseLock(props);
+        const res = await this.locking.releaseLock(props);
+        this.throwIfAbortedOrKilled();
+        return res;
```
TBonnin, line 316: releasing the lock first. Locks have a ttl so not a big deal but cleaner

rossmcewan (approved):

PR #5559, opened 2026-03-04
fix(slack): use environment id for slack connection id with legacy fallback

pfreixes:
Renaming an environment broke Slack alerts because the connection id was built using the environment name. Switch to environment id (stable) for new installations and add a fallback to the legacy name-based id for existing ones.

Tests run in local:
- [x] Test that new slack alerts are resilient to an environment name change
- [x] Test that legacy slack alerts are still working when they did not suffer an environment name change
- [x] Test that once legacy slack alerts once renamed they work

---

The update also applies the dual-lookup behavior across server-side Slack channel resolution, notification sending, and webapp disconnect flows, with `connectSlack` now constructing IDs from `envId`, and it expands API typing plus adds integration tests for the current-environment endpoint to validate Slack channel resolution for both new and legacy formats.

<details>

• Legacy disconnect retry in `slackDisconnect` only triggers on `404`; other failure statuses may leave legacy connections undeleted.

</details>

---

### DIFF
--- a/packages/server/lib/controllers/v1/environment/getEnvironment.integration.test.ts
+++ b/packages/server/lib/controllers/v1/environment/getEnvironment.integration.test.ts
@@ -64,7 +64,15 @@ describe(`GET ${route}`, () => {
     });
 
     describe('slack_notifications_channel', () => {
+        let adminProdEnv: Awaited<ReturnType<typeof seeders.createEnvironmentSeed>>;
         let originalAdminUUID: string | undefined;
+
+        beforeAll(async () => {
+            const { account: adminAccount } = await seeders.seedAccountEnvAndUser();
+            adminProdEnv = await seeders.createEnvironmentSeed(adminAccount.id, 'prod');
+            await seeders.createConfigSeed(adminProdEnv, 'slack', 'slack');
+            (envs as any).NANGO_ADMIN_UUID = adminAccount.uuid;
+        });
 
         beforeEach(() => {
             originalAdminUUID = envs.NANGO_ADMIN_UUID;
@@ -75,10 +83,6 @@ describe(`GET ${route}`, () => {
         });
 
         it('should return channel using new ID-based connection', async () => {
-            const { account: adminAccount } = await seeders.seedAccountEnvAndUser();
-            const adminProdEnv = await seeders.createEnvironmentSeed(adminAccount.id, 'prod');
-            await seeders.createConfigSeed(adminProdEnv, 'slack', 'slack');
-
             const { account: customerAccount, env: customerEnv, user } = await seeders.seedAccountEnvAndUser();
             await db.knex('_nango_environments').where({ id: customerEnv.id }).update({ slack_notifications: true });
 
@@ -89,7 +93,6 @@ describe(`GET ${route}`, () => {
                 connectionConfig: { 'incoming_webhook.channel': '#new-format-alerts' }
             });
 
-            (envs as any).NANGO_ADMIN_UUID = adminAccount.uuid;
             const session = await authenticateUser(api, user);
             const res = await api.fetch(route, {
                 method: 'GET',
@@ -104,10 +107,6 @@ describe(`GET ${route}`, () => {
         });
 
         it('should return channel using legacy name-based connection', async () => {
-            const { account: adminAccount } = await seeders.seedAccountEnvAndUser();
-            const adminProdEnv = await seeders.createEnvironmentSeed(adminAccount.id, 'prod');
-            await seeders.createConfigSeed(adminProdEnv, 'slack', 'slack');
-
             const { account: customerAccount, env: customerEnv, user } = await seeders.seedAccountEnvAndUser();
             await db.knex('_nango_environments').where({ id: customerEnv.id }).update({ slack_notifications: true });
 
@@ -118,7 +117,6 @@ describe(`GET ${route}`, () => {
                 connectionConfig: { 'incoming_webhook.channel': '#legacy-format-alerts' }
             });
 
-            (envs as any).NANGO_ADMIN_UUID = adminAccount.uuid;
             const session = await authenticateUser(api, user);
             const res = await api.fetch(route, {
                 method: 'GET',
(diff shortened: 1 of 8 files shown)


packages/shared/lib/services/notification/slack.service.ts
```diff
@@ -597,14 +598,21 @@ export class SlackService {
...
         }
 
-        const slackConnectionId = generateSlackConnectionId(account.uuid, environment.name);
+        // Try new ID-based connection ID first, fall back to legacy name-based for existing installations
+        let slackConnectionResult = await connectionService.getConnection(
+            generateSlackConnectionId(account.uuid, environment.id),
+            this.integrationKey,
+            adminRes.environment.id
+        );
+        if (!slackConnectionResult.success && slackConnectionResult.error?.type === 'unknown_connection') {
```
TBonnin: isn't `!slackConnectionResult.success ` redundant?

rossmcewan (approved):

TBonnin (approved):
lgtm. I don't think this endpoint has any integration tests yet. Might be a good time to add some :)

packages/shared/lib/services/notification/slack.service.ts
pfreixes: Yeps good point! let me remove it!

pfreixes:
@TBonnin @rossmcewan thanks for the review, Ive added some new integration tests and thanks to that Ive caught a [bug](https://github.com/NangoHQ/nango/pull/5559/changes#diff-ad8b585fc85ce1810dd5bce8dcb9dad794ca9e04475c9c5bec767d7f5b2849acR59) that I was not aware of!

A new review would be very appreciate it!

packages/server/lib/controllers/v1/environment/getEnvironment.ts
```diff
@@ -45,16 +46,23 @@ export const getEnvironment = asyncWrapper<GetEnvironment>(async (req, res) => {
...
         if (info) {
-            const connectionConfig = await connectionService.getConnectionConfig({
+            // Try new ID-based connection ID first, fall back to legacy name-based for existing installations
+            let connectionConfig = await connectionService.getConnectionConfig({
                 provider_config_key: integrationId,
                 environment_id: info.environmentId,
-                connection_id: connectionId
+                connection_id: generateSlackConnectionId(account.uuid, environment.id)
             });
+            if (!connectionConfig['incoming_webhook.channel']) {
```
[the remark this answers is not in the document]
pfreixes: good catch can be null! since its a nullable column

packages/server/lib/controllers/v1/environment/getEnvironment.integration.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    describe('slack_notifications_channel', () => {
+        let originalAdminUUID: string | undefined;
+
+        beforeEach(() => {
+            originalAdminUUID = envs.NANGO_ADMIN_UUID;
+        });
+
+        afterEach(() => {
+            (envs as any).NANGO_ADMIN_UUID = originalAdminUUID;
+        });
```
TBonnin, line 83: I would create the admin account and slack config in beforeAll so it doesn't need to be done in each test.

packages/server/lib/controllers/v1/environment/getEnvironment.ts
```diff
@@ -45,16 +46,23 @@ export const getEnvironment = asyncWrapper<GetEnvironment>(async (req, res) => {
...
         if (info) {
-            const connectionConfig = await connectionService.getConnectionConfig({
+            // Try new ID-based connection ID first, fall back to legacy name-based for existing installations
+            let connectionConfig = await connectionService.getConnectionConfig({
                 provider_config_key: integrationId,
                 environment_id: info.environmentId,
-                connection_id: connectionId
+                connection_id: generateSlackConnectionId(account.uuid, environment.id)
             });
+            if (!connectionConfig || !connectionConfig['incoming_webhook.channel']) {
```
TBonnin, line 59: why `!connectionConfig['incoming_webhook.channel']`

packages/shared/lib/services/notification/slack.service.ts
```diff
@@ -597,14 +598,21 @@ export class SlackService {
...
         }
 
-        const slackConnectionId = generateSlackConnectionId(account.uuid, environment.name);
+        // Try new ID-based connection ID first, fall back to legacy name-based for existing installations
+        let slackConnectionResult = await connectionService.getConnection(
+            generateSlackConnectionId(account.uuid, environment.id),
+            this.integrationKey,
+            adminRes.environment.id
+        );
+        if (slackConnectionResult.error?.type === 'unknown_connection') {
```
TBonnin, line 607: can we be consistent and use the same fallback condition here and in getEnvironment

packages/server/lib/controllers/v1/environment/getEnvironment.ts
pfreixes: bc getConnectionConfig can either return an empty object or a null, see [this](https://github.com/NangoHQ/nango/blob/4f03a5016b3b2c11fbfff566a229ca5d6db02dcf/packages/shared/lib/services/connection.service.ts#L512-L516).  From my understanding, I could be wrong here since my JS is a bit rusty,  when there is no record in the DB we return an empty object, but when there is a record we just return the `connection_config` column which can be null.

pfreixes: Indeed this was the original pattern, see [this](https://github.com/NangoHQ/nango/pull/5559/changes#diff-ad8b585fc85ce1810dd5bce8dcb9dad794ca9e04475c9c5bec767d7f5b2849acR66)

packages/shared/lib/services/notification/slack.service.ts
pfreixes: Let me know if Im missing something here, but both are not semantically equivalent. While this one just makes a look up to the db for checking if a connection exists for enabling the dual look up (originally there was an expectation that this function should always return a record and should be attributed to the connection slack one, otherwise this would mean that the dataset was corrupted or there was a bug - which is what is happening Today) the getEnvironment one uses the `getConnectionConfig` which is kinda different.

We could try to use in both cases `connectionService.getConnection`, but this is what we already had in place before. Also remember that we will remove this dual look up.

packages/server/lib/controllers/v1/environment/getEnvironment.integration.test.ts
pfreixes: Good point done [here](https://github.com/NangoHQ/nango/pull/5559/changes/14fb655f331e16af3cbe7f75e2f556a98efc4f7b).

TBonnin (approved):

PR #5561, opened 2026-03-04
feat(server): support flow-level json-schema

kaposke:
For context, each sync_config (flow) is saved in DB with a `models_json_schema` column representing it's used models.
The `POST /sync/deploy` (and `POST /sync/deploy/internal`) endpoint takes a list of flows, and a single top-level json-schema object with all the schemas used by the flows. This PR adds the option to pass a `models_json_schema` at a flow level, and marks the top-level json-schema as deprecated.

---

The deploy flow now prefers the per-flow `models_json_schema`, falling back to filtering the aggregated schema for legacy requests, with validation to ensure required model definitions are present; test coverage was also expanded for both deploy endpoints and deploy service behavior.

<details>

• Validation requires `models_json_schema.definitions` to include `input` when `input` is a string; confirm this matches intended behavior for all syncs.
• Legacy `jsonSchema` filtering assumes `definitions` structure; non-standard schema formats may still be rejected.

</details>

---

### DIFF
--- a/packages/shared/lib/services/sync/config/deploy.service.ts
+++ b/packages/shared/lib/services/sync/config/deploy.service.ts
@@ -2,13 +2,13 @@ import db, { dbNamespace } from '@nangohq/database';
 import { nangoConfigFile } from '@nangohq/nango-yaml';
 import { env, filterJsonSchemaForModels } from '@nangohq/utils';
 
-import configService from '../../config.service.js';
-import remoteFileService from '../../file/remote.service.js';
-import { getSyncsByProviderConfigKey } from '../sync.service.js';
 import { getSyncAndActionConfigByParams, getSyncAndActionConfigsBySyncNameAndConfigId, increment } from './config.service.js';
 import { NangoError } from '../../../utils/error.js';
+import configService from '../../config.service.js';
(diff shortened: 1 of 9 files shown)


packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -327,10 +329,11 @@ async function compileDeployInfo({
         }
     }
 
-    let models_json_schema: JSONSchema7 | null = null;
-    if (jsonSchema) {
+    let models_json_schema: JSONSchema7 | null = flow.models_json_schema ?? null;
+    if (!models_json_schema && aggregatedJsonSchema) {
```
TBonnin, line 333: is the CLI ever gonna send both?

TBonnin (commented):
i like half of the PR is tests 🏆

packages/shared/lib/services/sync/config/deploy.service.ts
kaposke: No. This code is for backwards compatibility with older CLI versions. Not really expecting both to be there, but expecting one of them to be there.

TBonnin (approved):

PR #5565, opened 2026-03-04
feat: expose function checkpoints (from/to) in Nango logs

TBonnin:
Runner is now returning from/to checkpoints which is exposed in the logs operation metadata to make it visible for customers for monitoring/debugging purpose

Screenshot 2026-03-04 at 14 54 07

`from` is the checkpoint value at the beginning of the execution (aka: last saved checkpoint of the previous execution)
`to` is the last saved checkpoints of this execution

Note: in a followup PR, the checkpoints info will be added to the sync webhook

 

---

It also propagates these checkpoint ranges through execution results into the jobs service, updating the jobs API contract and task payloads so handlers capture checkpoints on both success and error paths for end-to-end logging and metadata.

<details>

• If `getCheckpoint` is never called, `from` may remain `null` and logs may show an incomplete range; confirm that is acceptable for monitoring expectations.

</details>

---

### DIFF
--- a/packages/jobs/lib/invocations/lambda.processor.ts
+++ b/packages/jobs/lib/invocations/lambda.processor.ts
@@ -65,7 +65,8 @@ export class LambdaInvocationsProcessor {
                     status: parsedMessage.responseContext.statusCode
                 },
                 telemetryBag: { customLogs: 0, proxyCalls: 0, durationMs: 0, memoryGb: 0 },
-                functionRuntime: 'lambda'
+                functionRuntime: 'lambda',
+                checkpoints: null
             });
         }
     }
--- a/packages/jobs/lib/execution/action.ts
+++ b/packages/jobs/lib/execution/action.ts
@@ -28,6 +28,7 @@ import type { LogContext } from '@nangohq/logs';
 import type { OrchestratorTask, TaskAction } from '@nangohq/nango-orchestrator';
 import type { Config } from '@nangohq/shared';
 import type {
+    CheckpointRange,
     ConnectionJobs,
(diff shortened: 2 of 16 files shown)


packages/jobs/lib/invocations/lambda.processor.ts
```diff
@@ -65,7 +65,8 @@ export class LambdaInvocationsProcessor {
                     status: parsedMessage.responseContext.statusCode
                 },
                 telemetryBag: { customLogs: 0, proxyCalls: 0, durationMs: 0, memoryGb: 0 },
-                functionRuntime: 'lambda'
+                functionRuntime: 'lambda',
+                checkpoints: null
```
TBonnin, line 69: Unfortunately we are losing the checkpoints when the lambda OOM or times out. (same as telemetryBag).
Not a deal breaker imho for now but annoying nonetheless

rossmcewan: If this is something we want to fix we could keep an execution log of some sort that we could interrogate at this point. Example, when a checkpoint is stored we could record something using the AWS request id (or task id). If not a deal breaker right now maybe not pursuing in the short term, but if it becomes an issue, we have options.

rossmcewan (approved):

packages/jobs/lib/invocations/lambda.processor.ts
TBonnin: yes, logs/proxy/checkpoints are all touching the backend so we could keep track of them instead of relying on the runner to report them. Much more complicated though

PR #5572, opened 2026-03-05
feat(cli): build and send json-schema per function

kaposke:
> Function aka "flow", "sync_config"

We currently generate a "global" json schema with all the models in the integration folders and save it to `schema.json`. When deploying / generating docs / dry-running, we load that big schema and use it as the source of truth. We send the full json-schema for deploying, and let the server filter the schemas for each function (aka. sync_config, flow).
Furthermore, we take a very convoluted path to generate `schema.json`:
- Convert zod schemas to `NangoModel`
- Use `NangoModel` to generate `schema.ts`
- Use `ts-to-json-schema-generator` to generate `schema.json` from `schema.ts`

In my previous PR, I made the server accept function-level json-schemas. Each function can now be sent with it's own json-schema, as opposed to a single top-level json-schema in `POST /sync/deploy`.

This PR:
- Keeps the generation of `schema.ts` and `schema.json` AS IS (adds a deprecation notice comment on `schema.ts`)
- Updates zod versions in the `cli` and `runner-sdk` packages to v4.3.6 (currently latest)
- Also updates zod version in `cli/example/package.json`. This is what will make our customers packages be updated as soon as a cli command is run.
- Leverages `z.toJsonSchema` (introduced in zod v4) to generate json-schemas directly from the zod schemas.
- Adds the json-schemas to each function in the payload.
- Stops sending top-level json-schema with all models
- Modifies dry-run and generate:docs to use these function-level json-schemas instead of `schema.json`.
- These new json-schemas are generated and kept in-memory. They don't generate any artifacts. Every run of `deploy`, `dry-run` and `generate:docs` will re-generate them in-memory for use. The source is the compiled `cjs` files, and not the typescript files themselves.

---

It also adjusts parsing and validation types to allow optional per-function json_schema fields and updates related tests/snapshots to match the new schema format expectations.

---


packages/server/lib/controllers/sync/deploy/validation.ts
```diff
@@ -84,7 +84,6 @@ export const flowConfig = z
         webhookSubscriptions: z.array(z.string().max(255)).optional(),
         models_json_schema: z
             .object({
-                $schema: z.literal('http://json-schema.org/draft-07/schema#'),
```
kaposke, line 87: Felt like this is unnecessary and will increase storage with redundancy.

packages/cli/lib/zeroYaml/json-schema.ts
```diff
@@ -0,0 +1,53 @@
+import { z } from 'zod';
+
+import type { JSONSchema7 } from 'json-schema';
+
+function zodSchemaToJsonSchema(schema: z.ZodTypeAny): JSONSchema7 | null {
+    if (schema.constructor.name === 'ZodVoid') {
```
TBonnin: would it be less fragile to use `instanceOf` instead of `.name ===` or at least use the `_zod` private api?

packages/cli/lib/zeroYaml/json-schema.ts
```diff
@@ -0,0 +1,53 @@
...
+    delete jsonSchema['$schema'];
+
+    return jsonSchema;
+}
+
+/**
+ * Converts a map of named Zod schemas into a single JSON Schema document
+ * with a `definitions` block containing each model.
+ *
+ * Zod schemas that resolve to `void` are silently skipped.
```
TBonnin: that's not what the code above is saying

kaposke: Good catch, changed the code and forgot comment

packages/cli/lib/zeroYaml/json-schema.ts
kaposke: indeed. I got this from old zod to NangoModel code but instaceof seems to work just fine

packages/server/lib/controllers/sync/deploy/validation.ts
TBonnin: is it part of the spec? Are we ever gonna need to handle different draft version? if answers are no, we can omit it I think

packages/cli/lib/services/docs.service.ts
```diff
@@ -37,11 +36,6 @@ export async function generate({
 
     await fs.mkdir(writePath, { recursive: true });
 
-    const jsonSchema = loadSchemaJson({ fullPath: absolutePath });
-    if (!jsonSchema) {
-        return false;
```
TBonnin, line 42: we used to fail fast before if json schema couldn't be loaded. Is that ok to change this behavior?

packages/cli/lib/zeroYaml/definitions.ts
```diff
@@ -278,3 +272,65 @@ function postValidation(parsed: NangoYamlParsed): Result<void> {
...
+    }
+    return models;
+}
+
+/**
+ * Builds legacy NangoModels for an action.
+ * To remove when we deprecate `schema.ts` and `schema.json` files.
+ */
+export function buildNangoModelsForAction(
+    params: CreateActionResponse<z.ZodTypeAny, z.ZodTypeAny, z.ZodObject>,
```
TBonnin: CreateActionResponse has now 4 generic params

packages/cli/lib/zeroYaml/json-schema.ts
```diff
@@ -0,0 +1,52 @@
+import { z } from 'zod';
+
+import type { JSONSchema7 } from 'json-schema';
+
+function zodSchemaToJsonSchema(schema: z.ZodTypeAny): JSONSchema7 | null {
+    if (schema instanceof z.ZodVoid) {
+        return { type: 'null' };
+    }
```
TBonnin, line 8: should void just be omitted?

TBonnin (approved):

packages/cli/lib/zeroYaml/json-schema.ts
kaposke: Kept to keep current behavior

packages/cli/lib/services/docs.service.ts
kaposke: We don't load from a file anymore, it's generated in memory from zod. So it can't fail the same way

PR #5578, opened 2026-03-05
fix(orchestrator): enforce group max concurrency when dequeuing with wildcard pattern

TBonnin:
The running count subquery in the dequeue query used the raw groupKeyPattern (with `*`) instead of the SQL LIKE pattern (with `%`), causing it to never match any rows. 
Running count was then always 0 and max concurrency was never enforced when dequeueing with a wildcard group key pattern.

 

---

This update also ensures wildcard patterns are consistently converted from `*` to SQL `%` across all `LIKE` clauses in `dequeue`, adds integration validation for wildcard group concurrency, and adjusts default environment concurrency limits for sync, action, and webhook workloads.

<details>

• `groupKeyPattern` characters that are SQL wildcards (e.g., `_` or `%`) are not escaped, which could expand matching beyond intended `*` behavior.

</details>

---

### DIFF
--- a/packages/utils/lib/environment/parse.ts
+++ b/packages/utils/lib/environment/parse.ts
@@ -123,9 +123,9 @@ export const ENVS = z.object({
                 maxConcurrency: 50
             }
         ]),
-    SYNC_ENVIRONMENT_MAX_CONCURRENCY: z.coerce.number().optional().default(100),
-    ACTION_ENVIRONMENT_MAX_CONCURRENCY: z.coerce.number().optional().default(100),
-    WEBHOOK_ENVIRONMENT_MAX_CONCURRENCY: z.coerce.number().optional().default(50),
+    SYNC_ENVIRONMENT_MAX_CONCURRENCY: z.coerce.number().optional().default(200),
+    ACTION_ENVIRONMENT_MAX_CONCURRENCY: z.coerce.number().optional().default(200),
+    WEBHOOK_ENVIRONMENT_MAX_CONCURRENCY: z.coerce.number().optional().default(200),
     ON_EVENT_ENVIRONMENT_MAX_CONCURRENCY: z.coerce.number().optional().default(50),
 
     // Runner
--- a/packages/scheduler/lib/models/tasks.integration.test.ts
+++ b/packages/scheduler/lib/models/tasks.integration.test.ts
@@ -126,7 +126,7 @@ describe('Task', () => {
         const dequeued = (await tasks.dequeue(db, { groupKeyPattern: task.groupKey, limit: 1 })).unwrap();
         expect(dequeued).toHaveLength(0);
     });
-    it('should be dequeued according to group max concurrency ', async () => {
+    it('should be dequeued according to group max concurrency', async () => {
         const groupKey = nanoid();
         const groupMaxConcurrency = 2;
         const t0 = await createTask(db, { groupKey, groupMaxConcurrency });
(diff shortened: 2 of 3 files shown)


packages/utils/lib/environment/parse.ts
```diff
@@ -123,9 +123,9 @@ export const ENVS = z.object({
                 maxConcurrency: 50
             }
         ]),
-    SYNC_ENVIRONMENT_MAX_CONCURRENCY: z.coerce.number().optional().default(100),
-    ACTION_ENVIRONMENT_MAX_CONCURRENCY: z.coerce.number().optional().default(100),
-    WEBHOOK_ENVIRONMENT_MAX_CONCURRENCY: z.coerce.number().optional().default(50),
```
TBonnin, line 128: those numbers were ok a year ago but now our biggest customers have higher concurrency

pfreixes (approved):
Good finding, and good to see that we have this first throttling mechanism, some comments/questions:

Do we have a way for knowing when tasks are piled up at our side? so once we reach max concurrency for a specific customer I understand that we will stop dequeueing until there is room, IMO but be ideal to have a signal here (as we have Today with the lambda metrics) i.e message age, num pending tasks. Without this will be difficult to evaluate also the values experimented as limits vs what our customers are asking

With this change we are gonna be partially/totally moving the back-pressure system from SQS to our DB, right?, so we will pile up tasks in our DB as records rather than having messages piling up in SQS waiting for more resources (Lambda autoscale), so wondering if we should eventually have also a limit for the max tasks that a customer can have waiting/stored in the DB, or is this something that right now does not concern us?

Also and maybe as a future discussion, this change will defeat/limit somehow the capacity of Lambda for growing its pool execution for attending the current demand, if we limit the concurrency this might end up being also the ceiling of the max concurrency achieved by the Lambda. Still when the demand is coming from multiple customers the Lambda should be able to keep scaling.

TBonnin:
> Good finding, and good to see that we have this first throttling mechanism, some comments/questions:
> 
> Do we have a way for knowing when tasks are piled up at our side? so once we reach max concurrency for a specific customer I understand that we will stop dequeueing until there is room, IMO but be ideal to have a signal here (as we have Today with the lambda metrics) i.e message age, num pending tasks. Without this will be difficult to evaluate also the values experimented as limits vs what our customers are asking

All this info lives in the orchestrator db but there are no graph views or anything. Something to add to the list of features for the infamous admin portal. :)
 
> With this change we are gonna be partially/totally moving the back-pressure system from SQS to our DB, right?, so we will pile up tasks in our DB as records rather than having messages piling up in SQS waiting for more resources (Lambda autoscale), so wondering if we should eventually have also a limit for the max tasks that a customer can have waiting/stored in the DB, or is this something that right now does not concern us?

It would indeed be good to have a max queue size safeguard limit (similar to the 6h invocation queue threshold in Lambda). I'll create a ticket. A bit annoying because orchestrator design was until now about "insert quickly, dequeue later". 
 
> Also and maybe as a future discussion, this change will defeat/limit somehow the capacity of Lambda for growing its pool execution for attending the current demand, if we limit the concurrency this might end up being also the ceiling of the max concurrency achieved by the Lambda. Still when the demand is coming from multiple customers the Lambda should be able to keep scaling.

Yesterday incident shows that Lambda doesn't scale up magically so having a safeguard mechanism to avoid splash effect and noisy neighbors is necessary to protect the systems upstream.

PR #5580, opened 2026-03-06
fix(runner): distributed conflict tracking

rossmcewan:
---

**Distributed conflict tracking for runner sync tasks via KV store**

This PR replaces in-memory-only conflict detection in the runner with KV-backed tracking to avoid concurrent sync conflicts across distributed runners. `RunnerMonitor` now writes conflict keys to a `KVStore` with TTL, refreshes them during heartbeats, and removes them on completion; start flow now awaits tracking and treats key set failures as the conflict signal.

It adds configuration for conflict tracking mode and heartbeat TTL multiplier, introduces the `@nangohq/kvstore` dependency, and includes unit tests verifying `track`/`untrack` behavior for sync tasks.

---

### DIFF
--- a/packages/runner/lib/monitor.ts
+++ b/packages/runner/lib/monitor.ts
@@ -53,11 +53,12 @@ export class RunnerMonitor {
         }
     }
 
-    async untrack(nangoProps: NangoProps, taskId: string): Promise<void> {
-        this.tracked.delete(taskId);
-        if (this.conflictTracking.functionTypes.includes(nangoProps.scriptType)) {
+    async untrack(taskId: string): Promise<void> {
+        const nangoProps = this.tracked.get(taskId)?.nangoProps;
+        if (nangoProps && this.conflictTracking.functionTypes.includes(nangoProps.scriptType)) {
             await this.conflictTracking.tracker.delete(`function:${nangoProps.scriptType}:${nangoProps.syncId}`);
         }
+        this.tracked.delete(taskId);
     }
 
     resetIdleMaxDurationMs(): void {
--- a/packages/runner/lib/monitor.ts
+++ b/packages/runner/lib/monitor.ts
@@ -51,11 +51,12 @@ export class RunnerMonitor {
             try {
                 await this.conflictTracking.tracker.set(this.generateConflictKey(nangoProps), '1', {
                     canOverride: opts.refresh,
-                    ttlMs: envs.RUNNER_HEARTBEAT_INTERVAL_MS * envs.RUNNER_HEARTBEAT_INTERVAL_MULTIPLIER
+                    ttlMs: envs.RUNNER_HEARTBEAT_INTERVAL_MS * envs.RUNNER_SYNC_CONFLICT_HEARTBEAT_INTERVAL_MULTIPLIER
                 });
             } catch (err) {
+                logger.error('Failed to track sync for conflicts', { error: err });
                 if (err instanceof Error && err.message.includes('set_key_already_exists')) {
-                    throw new Error('conflicting_sync');
+                    throw new Error('Conflicting sync detected');
                 }
                 throw err;
             }
@@ -159,15 +160,6 @@ export class RunnerMonitor {
             }
             this.checkIdle(nextTimeout);
         }, timeoutMs);
-    }
-
-    async hasConflictingSync(newTask: NangoProps): Promise<boolean> {
-        if (newTask.scriptType == 'sync') {
-            const key = `function:${newTask.scriptType}:${newTask.syncId}`;
-            const exists = await this.conflictTracking.tracker.exists(key);
-            return exists;
-        }
-        return false;
     }
 }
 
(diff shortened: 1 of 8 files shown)


packages/runner/lib/monitor.ts
```diff
@@ -39,15 +42,22 @@ export class RunnerMonitor {
...
 
-    track(nangoProps: NangoProps, taskId: string): void {
+    async track(nangoProps: NangoProps, taskId: string): Promise<void> {
         this.lastIdleTrackingDate = Date.now();
         this.tracked.set(taskId, { nangoProps });
         if (!this.persistClient) {
             this.persistClient = new PersistClient({ secretKey: nangoProps.secretKey });
         }
+        if (this.conflictTracking.functionTypes.includes(nangoProps.scriptType)) {
+            await this.conflictTracking.tracker.set(`function:${nangoProps.scriptType}:${nangoProps.syncId}`, '1');
```
pfreixes: Unless Im missing something here we have the chance to have data races between the exists call an the set, this was most likely not happening in the in-memory (you would tell me otherwise) due to the non interruption of the event loop.

So there is a chance that we call exists multiple times before we have set the key value, for doing an atomic set if does not exist we would require to use NX

```
SET lock:job123 worker1 NX
```

Also what are the consequences of leaving keys orphan? if we never call the `untrack` (for example the process dies) the key will remain there forever, should we add a TTL?

pfreixes (commented):
Left some comments

packages/runner/lib/monitor.ts
```diff
@@ -22,8 +24,9 @@ export class RunnerMonitor {
     private memoryInterval: NodeJS.Timeout | null = null;
     private runnerAccountId: string | null = null;
 
-    constructor({ runnerId }: { runnerId: number }) {
+    constructor({ runnerId, conflictTracking }: { runnerId: number; conflictTracking: { tracker: KVStore; functionTypes: ScriptType[] } }) {
```
TBonnin: everything but syncs can be called concurrently. So I don't think we will ever need to detect conflict for other types without a huge refactoring of how Nango works. I would remove the functionTypes options and just hardcode `sync` for simplicity

packages/runner/lib/monitor.ts
TBonnin: valid points. The set/check isn't atomic so prone to race conditions.
Adding a long TTL (even a week) would be nice for things to autoclean

TBonnin (changes requested):

packages/runner/lib/monitor.ts
```diff
@@ -39,15 +42,25 @@ export class RunnerMonitor {
...
         if (!this.persistClient) {
             this.persistClient = new PersistClient({ secretKey: nangoProps.secretKey });
         }
     }
 
-    untrack(taskId: string): void {
+    async untrack(taskId: string): Promise<void> {
+        const nangoProps = this.tracked.get(taskId)?.nangoProps;
+        if (nangoProps && nangoProps.scriptType == 'sync') {
+            await this.conflictTracking.tracker.delete(`function:${nangoProps.scriptType}:${nangoProps.syncId}`);
```
TBonnin: what about adding the environmentId in the key. Not required functionally speaking since syncId is unique but for debugging purpose

packages/runner/lib/monitor.ts
```diff
@@ -39,15 +42,25 @@ export class RunnerMonitor {
...
+        }
         this.tracked.delete(taskId);
     }
```
TBonnin, line 85: track and untrack must catch tracker exceptions. Letting them bubble up is more complicated to understand and change the behavior between REDIS and IN_MEMORY trackers

TBonnin (approved):
Approving so you can deploy on Monday but the exceptions handling in track/untrack must be addressed

packages/runner/lib/monitor.ts
TBonnin: We can actually catch the `set_key_already_exists` and even remove the `hasConflictingSync` function and have the `set` failure act as the detection mechanism

packages/runner/lib/monitor.ts
```diff
@@ -39,15 +42,25 @@ export class RunnerMonitor {
         }
     }
 
-    track(nangoProps: NangoProps, taskId: string): void {
+    async track(nangoProps: NangoProps, taskId: string): Promise<void> {
+        if (nangoProps.scriptType == 'sync') {
+            await this.conflictTracking.tracker.set(`function:${nangoProps.scriptType}:${nangoProps.syncId}`, '1', {
+                canOverride: false,
+                ttlMs: 1000 * 60 * 60 * 24 * 7 // 1 week
```
TBonnin: actually the ttl might be an issue. if the function crashes before untracking, it would lead to the sync not executing for a week since the key was never cleaned up. With the in-memory store when the function crashes the tracked items are wiped (rightly so since the runner crashed and the tracked functions aren't running anymore) which doesn't happen with the shared kvstore.
A potential solution would be to set a TTL slightly longer than the heartbeat interval and "renew" the key on heartbeat. At least if it crashes, the conflict will be resolve after maximum the heartbeat interval

TBonnin (commented):
Made another comment about the TTL and lack of automatic cleanup when crashing. Let's talk on Monday

PR #5581, opened 2026-03-06
feat(playground): NAN-4846: Create playground

agusayerza:
Screenshot 2026-04-07 at 10 25 18 AM
Screenshot 2026-04-07 at 10 25 27 AM
Screenshot 2026-04-07 at 10 25 44 AM

---

**Add `Playground` End-to-End Feature with Persisted State, Execution Lifecycle, and Permission Gating**

This PR introduces a major new `Playground` experience in the web app, accessible from `packages/webapp/src/components-v2/AppHeader.tsx` and rendered in `packages/webapp/src/layout/DashboardLayout.tsx` via a right-side `Sheet`. Users can select an integration, connection, and function (`action` or `sync`), provide schema-derived inputs, trigger runs, view execution status/results, cancel sync runs, and jump to filtered logs. The implementation includes new feature modules (`Playground` components, orchestration hook, utilities, and shared `Combobox`) plus updates to shared UI primitives and styling utilities.

State management was split into a dedicated persisted Zustand store in `packages/webapp/src/store/playground.ts` with `sessionStorage` persistence and selective partialization for reattach behavior. The run lifecycle is handled in `packages/webapp/src/components-v2/Playground/usePlayground.ts` with input parsing/validation, operation discovery, status polling, terminal result handling, and best-effort cancellation. The PR also adds production authorization checks via `packages/authz/lib/permissions.ts` (`canUseProdPlayground`) and broad test coverage for schema parsing, parsing/validation helpers, store behavior/persistence, and logs URL generation.

---

### DIFF
--- a/packages/server/lib/controllers/v1/trigger/postTriggerFunction.integration.test.ts
+++ b/packages/server/lib/controllers/v1/trigger/postTriggerFunction.integration.test.ts
@@ -3,7 +3,7 @@ import { afterAll, beforeAll, describe, expect, it, vi } from 'vitest';
 import { envs } from '@nangohq/logs';
 import { seeders, syncManager } from '@nangohq/shared';
 
-import { authenticateUser, isError, runServer } from '../../../utils/tests.js';
+import { authenticateUser, isError, runServer, shouldBeProtected } from '../../../utils/tests.js';
 
 let api: Awaited<ReturnType<typeof runServer>>;
 
@@ -16,24 +16,26 @@ const mockRunSyncCommand = vi.spyOn(syncManager, 'runSyncCommand').mockResolvedV
 });
 
 describe(`POST ${endpoint}`, () => {
+    let logsEnabled: boolean;
+
     beforeAll(async () => {
         api = await runServer();
+        logsEnabled = envs.NANGO_LOGS_ENABLED;
         envs.NANGO_LOGS_ENABLED = false;
     });
     afterAll(() => {
+        envs.NANGO_LOGS_ENABLED = logsEnabled;
         api.server.close();
     });
 
-    it('should require session auth', async () => {
+    it('should be protected', async () => {
         const res = await api.fetch(endpoint, {
             method: 'POST',
             query: { env: 'dev' },
             body: { type: 'sync', function_name: 'sync1', provider_config_key: 'test-key', connection_id: 'conn1' }
         });
 
-        isError(res.json);
-        expect(res.res.status).toBe(401);
-        expect(res.json).toStrictEqual({ error: { code: 'unauthorized' } });
+        shouldBeProtected(res);
     });
 
     it('should require the env query param', async () => {
@@ -49,7 +51,9 @@ describe(`POST ${endpoint}`, () => {
         });
 
         isError(res.json);
-        expect(res.res.status).toBe(400);
+        // sessionAuth middleware validates env and returns 401 for missing/invalid env
+        expect(res.res.status).toBe(401);
+        expect(res.json.error.code).toBe('invalid_env');
     });
 
     describe('body validation', () => {
--- a/packages/server/lib/controllers/v1/trigger/postTriggerFunction.integration.test.ts
+++ b/packages/server/lib/controllers/v1/trigger/postTriggerFunction.integration.test.ts
@@ -1,0 +1,166 @@
+import { afterAll, beforeAll, describe, expect, it, vi } from 'vitest';
+
+import { envs } from '@nangohq/logs';
+import { seeders, syncManager } from '@nangohq/shared';
+
+import { authenticateUser, isError, runServer } from '../../../utils/tests.js';
+
+let api: Awaited<ReturnType<typeof runServer>>;
+
+const endpoint = '/api/v1/trigger/function';
+
+const mockRunSyncCommand = vi.spyOn(syncManager, 'runSyncCommand').mockResolvedValue({
+    success: true,
+    response: true,
+    error: null
+});
+
+describe(`POST ${endpoint}`, () => {
+    beforeAll(async () => {
+        api = await runServer();
+        envs.NANGO_LOGS_ENABLED = false;
+    });
+    afterAll(() => {
+        api.server.close();
+    });
+
+    it('should require session auth', async () => {
+        const res = await api.fetch(endpoint, {
+            method: 'POST',
+            query: { env: 'dev' },
+            body: { type: 'sync', function_name: 'sync1', provider_config_key: 'test-key', connection_id: 'conn1' }
+        });
+
+        isError(res.json);
+        expect(res.res.status).toBe(401);
+        expect(res.json).toStrictEqual({ error: { code: 'unauthorized' } });
+    });
+
+    it('should require the env query param', async () => {
+        const { user } = await seeders.seedAccountEnvAndUser();
+        const session = await authenticateUser(api, user);
+
+        const res = await api.fetch(endpoint, {
+            method: 'POST',
+            // @ts-expect-error intentionally omitting required query
+            query: {},
+            body: { type: 'sync', function_name: 'sync1', provider_config_key: 'test-key', connection_id: 'conn1' },
+            session
+        });
+
+        isError(res.json);
+        expect(res.res.status).toBe(400);
+    });
+
+    describe('body validation', () => {
+        it('should return 400 when type is missing', async () => {
+            const { user } = await seeders.seedAccountEnvAndUser();
+            const session = await authenticateUser(api, user);
+
+            const res = await api.fetch(endpoint, {
+                method: 'POST',
+                query: { env: 'dev' },
+                // @ts-expect-error intentionally invalid body
+                body: { function_name: 'sync1', provider_config_key: 'test-key', connection_id: 'conn1' },
+                session
+            });
+
+            isError(res.json);
+            expect(res.res.status).toBe(400);
+            expect(res.json.error.code).toBe('invalid_body');
+        });
+
+        it('should return 400 when type is invalid', async () => {
+            const { user } = await seeders.seedAccountEnvAndUser();
+            const session = await authenticateUser(api, user);
+
+            const res = await api.fetch(endpoint, {
+                method: 'POST',
+                query: { env: 'dev' },
+                // @ts-expect-error intentionally invalid body
+                body: { type: 'webhook', function_name: 'sync1', provider_config_key: 'test-key', connection_id: 'conn1' },
+                session
+            });
+
+            isError(res.json);
+            expect(res.res.status).toBe(400);
+            expect(res.json.error.code).toBe('invalid_body');
+        });
+
+        it('should return 400 when function_name is missing', async () => {
+            const { user } = await seeders.seedAccountEnvAndUser();
+            const session = await authenticateUser(api, user);
+
+            const res = await api.fetch(endpoint, {
+                method: 'POST',
+                query: { env: 'dev' },
+                // @ts-expect-error intentionally invalid body
+                body: { type: 'sync', provider_config_key: 'test-key', connection_id: 'conn1' },
+                session
+            });
+
+            isError(res.json);
+            expect(res.res.status).toBe(400);
+            expect(res.json.error.code).toBe('invalid_body');
+        });
+
+        it('should return 400 when provider_config_key is missing', async () => {
+            const { user } = await seeders.seedAccountEnvAndUser();
+            const session = await authenticateUser(api, user);
+
+            const res = await api.fetch(endpoint, {
+                method: 'POST',
+                query: { env: 'dev' },
+                // @ts-expect-error intentionally invalid body
+                body: { type: 'sync', function_name: 'sync1', connection_id: 'conn1' },
+                session
+            });
+
+            isError(res.json);
+            expect(res.res.status).toBe(400);
+            expect(res.json.error.code).toBe('invalid_body');
+        });
+    });
+
+    describe('sync', () => {
+        it('should call syncManager.runSyncCommand with RUN and correct identifiers', async () => {
+            const { user } = await seeders.seedAccountEnvAndUser();
+            const session = await authenticateUser(api, user);
+
+            const res = await api.fetch(endpoint, {
+                method: 'POST',
+                query: { env: 'dev' },
+                body: { type: 'sync', function_name: 'my-sync', provider_config_key: 'test-key', connection_id: 'conn1' },
+                session
+            });
+
+            expect(res.res.status).toBe(200);
+            expect(mockRunSyncCommand).toHaveBeenCalledWith(
+                expect.objectContaining({
+                    command: 'RUN',
+                    providerConfigKey: 'test-key',
+                    connectionId: 'conn1',
+                    syncIdentifiers: [{ syncName: 'my-sync', syncVariant: 'base' }]
+                })
+            );
+        });
+    });
+
+    describe('action', () => {
+        it('should return 400 for unknown connection', async () => {
+            const { user } = await seeders.seedAccountEnvAndUser();
+            const session = await authenticateUser(api, user);
+
+            const res = await api.fetch(endpoint, {
+                method: 'POST',
+                query: { env: 'dev' },
+                body: { type: 'action', function_name: 'my-action', provider_config_key: 'test-key', connection_id: 'nonexistent' },
+                session
+            });
+
+            isError(res.json);
+            expect(res.res.status).toBe(400);
+            expect(res.json.error.code).toBe('unknown_connection');
+        });
+    });
+});
--- a/packages/server/lib/controllers/v1/trigger/postTriggerFunction.ts
+++ b/packages/server/lib/controllers/v1/trigger/postTriggerFunction.ts
@@ -4,7 +4,7 @@ import * as z from 'zod';
 import { OtlpSpan, defaultOperationExpiration, logContextGetter } from '@nangohq/logs';
 import { records as recordsService } from '@nangohq/records';
 import { SyncCommand, configService, connectionService, errorManager, getSyncConfigRaw, syncManager } from '@nangohq/shared';
-import { metrics, zodErrorToHTTP } from '@nangohq/utils';
+import { metrics, truncateJson, zodErrorToHTTP } from '@nangohq/utils';
 
 import { envs } from '../../../env.js';
 import { connectionIdSchema, providerConfigKeySchema, syncNameSchema } from '../../../helpers/validation.js';
@@ -88,7 +88,7 @@ export const postTriggerFunction = asyncWrapper<PostInternalTriggerFunction>(asy
                         integration: { id: provider.id!, name: connection.provider_config_key, provider: provider.provider },
                         connection: { id: connection.id, name: connection.connection_id },
                         syncConfig: { id: syncConfig.id, name: syncConfig.sync_name },
-                        meta: { input }
+                        meta: truncateJson({ input })
                     }
                 );
                 logCtx.attachSpan(new OtlpSpan(logCtx.operation));
--- a/packages/server/lib/controllers/v1/trigger/postTriggerFunction.ts
+++ b/packages/server/lib/controllers/v1/trigger/postTriggerFunction.ts
@@ -1,19 +1,17 @@
 import tracer from 'dd-trace';
 import * as z from 'zod';
 
-import { OtlpSpan, defaultOperationExpiration, logContextGetter } from '@nangohq/logs';
+import { logContextGetter } from '@nangohq/logs';
 import { records as recordsService } from '@nangohq/records';
-import { SyncCommand, configService, connectionService, errorManager, getSyncConfigRaw, syncManager } from '@nangohq/shared';
-import { metrics, truncateJson, zodErrorToHTTP } from '@nangohq/utils';
+import { SyncCommand, errorManager, syncManager } from '@nangohq/shared';
+import { getHeaders, metrics, redactHeaders, zodErrorToHTTP } from '@nangohq/utils';
 
-import { envs } from '../../../env.js';
 import { connectionIdSchema, providerConfigKeySchema, syncNameSchema } from '../../../helpers/validation.js';
-import { pubsub } from '../../../pubsub.js';
 import { asyncWrapper } from '../../../utils/asyncWrapper.js';
 import { getOrchestrator } from '../../../utils/utils.js';
+import { runAction } from '../../action/runAction.js';
 import { normalizeSyncParams } from '../../sync/helpers.js';
 
-import type { LogContextOrigin } from '@nangohq/logs';
 import type { PostInternalTriggerFunction } from '@nangohq/types';
 
 const bodySchema = z.discriminatedUnion('type', [
@@ -50,93 +48,32 @@ export const postTriggerFunction = asyncWrapper<PostInternalTriggerFunction>(asy
         metrics.increment(metrics.Types.ACTION_INCOMING_PAYLOAD_SIZE_BYTES, req.rawBody ? Buffer.byteLength(req.rawBody) : 0, { accountId: account.id });
 
         await tracer.trace<Promise<void>>('server.trigger.action', async (span) => {
-            let logCtx: LogContextOrigin | undefined;
-            try {
-                const { success, response: connection } = await connectionService.getConnection(connection_id, provider_config_key, environment.id);
-                if (!success || !connection) {
-                    res.status(400).send({ error: { code: 'unknown_connection', message: 'Failed to find connection' } });
-                    return;
+            const logCtx = await runAction({
+                account: account,
+                environment: environment,
+                connectionId: connection_id,
+                providerConfigKey: provider_config_key,
+                actionName: function_name,
+                input,
+                isAsync: false,
+                retryMax: 0,
+                res,
+                span
+            });
+
+            const reqHeaders = getHeaders(req.headers);
+            const responseHeaders = getHeaders(res.getHeaders());
+            await logCtx?.enrichOperation({
+                request: {
+                    url: `${req.protocol}://${req.get('host')}${req.originalUrl}`,
+                    method: req.method,
+                    headers: redactHeaders({ headers: reqHeaders })
+                },
+                response: {
+                    code: res.statusCode,
+                    headers: redactHeaders({ headers: responseHeaders })
                 }
-
-                const provider = await configService.getProviderConfig(provider_config_key, environment.id);
-                if (!provider) {
-                    res.status(400).json({ error: { code: 'unknown_provider', message: 'Failed to find provider' } });
-                    return;
-                }
-
-                const syncConfig = await getSyncConfigRaw({ environmentId: environment.id, config_id: provider.id!, name: function_name, isAction: true });
-                if (!syncConfig) {
-                    res.status(404).json({ error: { code: 'not_found', message: 'Action not found' } });
-                    return;
-                }
-
-                if (!syncConfig.enabled) {
-                    res.status(404).json({ error: { code: 'disabled_resource', message: 'The action is disabled' } });
-                    return;
-                }
-
-                span.setTag('nango.actionName', function_name)
-                    .setTag('nango.connectionId', connection_id)
-                    .setTag('nango.environmentId', environment.id)
-                    .setTag('nango.providerConfigKey', provider_config_key);
-
-                logCtx = await logContextGetter.create(
-                    { operation: { type: 'action', action: 'run' }, expiresAt: defaultOperationExpiration.action() },
-                    {
-                        account,
-                        environment,
-                        integration: { id: provider.id!, name: connection.provider_config_key, provider: provider.provider },
-                        connection: { id: connection.id, name: connection.connection_id },
-                        syncConfig: { id: syncConfig.id, name: syncConfig.sync_name },
-                        meta: truncateJson({ input })
-                    }
-                );
-                logCtx.attachSpan(new OtlpSpan(logCtx.operation));
-
-                const actionResponse = await orchestrator.triggerAction({
-                    accountId: account.id,
-                    connection,
-                    actionName: function_name,
-                    input,
-                    async: false,
-                    retryMax: 0,
-                    maxConcurrency: envs.ACTION_ENVIRONMENT_MAX_CONCURRENCY,
-                    logCtx
-                });
-
-                if (actionResponse.isOk()) {
-                    const responseData = 'statusUrl' in actionResponse.value ? actionResponse.value : actionResponse.value.data;
-                    res.status(200).json(responseData);
-
-                    void pubsub.publisher.publish({
-                        subject: 'usage',
-                        type: 'usage.actions',
-                        idempotencyKey: logCtx.id,
-                        payload: {
-                            value: 1,
-                            properties: {
-                                accountId: account.id,
-                                connectionId: connection.connection_id,
-                                environmentId: environment.id,
-                                environmentName: environment.name,
-                                integrationId: provider_config_key,
-                                actionName: function_name
-                            }
-                        }
-                    });
-                } else {
-                    span.setTag('nango.error', actionResponse.error);
-                    await logCtx.failed();
-                    errorManager.errResFromNangoErr(res, actionResponse.error);
-                }
-            } catch (err) {
-                span.setTag('nango.error', err);
-                if (logCtx) {
-                    void logCtx.error('Failed to run action', { error: err });
-                    await logCtx.failed();
-                }
-                res.status(500).send({ error: { code: 'internal_server_error', message: 'Failed to run action' } });
-            }
+            });
         });
     } else {
         const syncIdentifiers = normalizeSyncParams([function_name]);
@@ -144,7 +81,7 @@ export const postTriggerFunction = asyncWrapper<PostInternalTriggerFunction>(asy
         const { success, error } = await syncManager.runSyncCommand({
             recordsService,
             orchestrator,
-            environment,
+            environment: environment,
             providerConfigKey: provider_config_key,
             syncIdentifiers,
             command: SyncCommand.RUN,
--- a/packages/server/lib/controllers/v1/trigger/postTriggerFunction.ts
+++ b/packages/server/lib/controllers/v1/trigger/postTriggerFunction.ts
@@ -1,0 +1,163 @@
+import tracer from 'dd-trace';
+import * as z from 'zod';
+
+import { OtlpSpan, defaultOperationExpiration, logContextGetter } from '@nangohq/logs';
+import { records as recordsService } from '@nangohq/records';
+import { SyncCommand, configService, connectionService, errorManager, getSyncConfigRaw, syncManager } from '@nangohq/shared';
+import { metrics, zodErrorToHTTP } from '@nangohq/utils';
+
+import { envs } from '../../../env.js';
+import { connectionIdSchema, providerConfigKeySchema, syncNameSchema } from '../../../helpers/validation.js';
+import { pubsub } from '../../../pubsub.js';
+import { asyncWrapper } from '../../../utils/asyncWrapper.js';
+import { getOrchestrator } from '../../../utils/utils.js';
+import { normalizeSyncParams } from '../../sync/helpers.js';
+
+import type { LogContextOrigin } from '@nangohq/logs';
+import type { PostInternalTriggerFunction } from '@nangohq/types';
+
+const bodySchema = z.discriminatedUnion('type', [
+    z.strictObject({
+        type: z.literal('action'),
+        function_name: syncNameSchema,
+        provider_config_key: providerConfigKeySchema,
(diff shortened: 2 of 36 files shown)


packages/webapp/src/components-v2/Playground/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+                        'flex flex-col items-start gap-[10px]',
+                        'w-[537px] max-w-none sm:max-w-none',
+                        'data-[state=open]:fade-in-0 data-[state=closed]:fade-out-0',
+                        '[&>button]:hidden'
+                    )}
+                >
+                    {/* Header */}
+                    <div className="flex w-full items-start justify-between pb-8">
+                        <div className="min-w-0">
+                            <h2 className="text-text-primary text-heading-medium font-medium text-[20px] pb-2">Playground</h2>
```
agusayerza: Lots of static values `text-[20px]` throughout the PR as we did a best effort approach to match design until the new DS is adopted

packages/webapp/src/components-v2/Playground/index.tsx
```diff
@@ -0,0 +1,165 @@
...
+                    side="right"
+                    overlayClassName="hidden"
+                    insetTop={108}
+                    insetBottom={44}
+                    insetRight={24}
+                    onInteractOutside={(e) => e.preventDefault()}
+                    onPointerDownOutside={(e) => e.preventDefault()}
+                    onFocusOutside={(e) => e.preventDefault()}
+                    className={cn(
+                        'dark:bg-bg-elevated text-text-primary rounded-lg border border-border-muted shadow-lg p-6',
```
agusayerza: `bg-bg-elevated` didn't match the design so went with `dark:bg-bg-elevated` which did. Wonder if its an underlying issue?

packages/webapp/src/components-v2/Playground/usePlaygroundRun.ts
```diff
@@ -0,0 +1,307 @@
...
+            };
+
+            // If the trigger failed immediately (non-2xx), skip polling and surface the error right away.
+            if (!response.ok) {
+                setPlaygroundResult({ success: false, data: triggerData, durationMs: triggerDurationMs });
+                return;
+            }
+
+            // Poll until we find the matching operation in logs.
+            let operation = null as SearchOperations['Success']['data'][number] | null;
```
agusayerza: This is a best effort approach to find the triggered operation. Follow up diff that actually changes how the trigger endpoint works to return the actual operation coming soon tm.

packages/webapp/src/components-v2/ui/sheet.tsx
```diff
@@ -39,28 +39,70 @@ function SheetContent({
     className,
     children,
     side = 'right',
+    overlayClassName,
+    insetTop,
+    insetBottom,
+    insetLeft,
+    insetRight,
```
agusayerza, line 46: Is this the best way of doing it? I think right now as the sheet is only used on the playground we could have left them as static values

packages/webapp/src/components-v2/AppHeader.tsx
```diff
@@ -1,14 +1,36 @@
...
+                            playgroundOpen
+                                ? '[background:linear-gradient(var(--color-bg-surface),var(--color-bg-surface))_padding-box,linear-gradient(90deg,var(--color-brand-500)_0%,var(--color-brand-500)_100%)_border-box]'
+                                : '[background:linear-gradient(var(--color-bg-surface),var(--color-bg-surface))_padding-box,linear-gradient(90deg,var(--color-brand-500)_0%,var(--color-border-default)_100%)_border-box]'
```
agusayerza: The design for this button go way out of standard and that is by design

packages/webapp/src/components-v2/Playground/usePlaygroundRun.ts
```diff
@@ -0,0 +1,347 @@
...
+    ]);
+
+    const handleCancel = useCallback(async () => {
+        // Abort local polling immediately
+        abortRef.current?.abort();
+        setPlaygroundPendingOperationId(null);
+        setPlaygroundRunning(false);
+        setPlaygroundResult(null);
+
+        // For syncs, also cancel the backend operation (best-effort)
```
agusayerza: This will be improved once the trigger response is improved on. Follow up PR

packages/webapp/src/store.ts
```diff
@@ -19,39 +46,176 @@ interface State {
...
+    setPlaygroundInputValue: (name: string, value: string) => void;
+    setPlaygroundResult: (result: PlaygroundResult | null) => void;
+    setPlaygroundPendingOperationId: (operationId: string | null) => void;
+    setPlaygroundRunning: (running: boolean) => void;
+    setPlaygroundInputErrors: (errors: Record<string, string>) => void;
+    setPlaygroundInputError: (name: string, message: string) => void;
+    clearPlaygroundInputError: (name: string) => void;
+    setPlaygroundConnectionSearch: (search: string) => void;
+    resetPlayground: () => void;
+    setPlaygroundState: (value: PlaygroundState) => void;
```
marcindobry: Could we extract all the playground store stuff into a separate store, with separate state interface etc.? There is a guide about this for zustand - [how to modularize the stores](https://zustand.docs.pmnd.rs/learn/guides/slices-pattern), but I think that maybe even have a completely separate ones not to pollute the type of the global one could also make sense. It is only used to pass around a bunch of state within some components, it's not really referenced globally across the rest of the app

agusayerza: Good call. I've moved all playground state into a separate `usePlaygroundStore`

packages/webapp/src/store/playground.ts
```diff
@@ -0,0 +1,120 @@
...
+        }),
+        {
+            name: LocalStorageKeys.Playground,
+            partialize: (s) => ({
+                isOpen: false,
+                integration: s.integration,
+                connection: s.connection,
+                function: s.function,
+                functionType: s.functionType,
+                inputValues: s.inputValues,
```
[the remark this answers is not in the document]
agusayerza: I think the concern is valid but this is a UX trade-off. We want to store the inputs so we can recover the state. I think the right balance will be moving it to session storage

packages/webapp/src/components-v2/Playground/usePlaygroundReattach.ts
```diff
@@ -0,0 +1,167 @@
...
+                    if (err instanceof Error && err.name === 'AbortError') {
+                        // Sheet closed or component unmounted. Leave pendingOperationId in
+                        // place so reattach fires again when the sheet reopens.
+                    } else {
+                        setPlaygroundPendingOperationId(null);
+                        setPlaygroundResult({ success: false, data: { error: 'Failed to re-attach to operation' }, durationMs: 0 });
+                    }
+                } finally {
+                    reattachingRef.current = false;
+                    usePlaygroundStore.getState().setRunning(false);
```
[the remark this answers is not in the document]
agusayerza: The brief flicker is real but harmless, as the by setting it to `false` the listner will kick off and detect that the playground is open and there is an operation id setting back to `true`. Without setting it to `false` we risk getting stuck

TBonnin (commented):
Super cool.
I'll @kaposke review the code but I have tried it locally:

1.
Black integration icon isn't visible in the dropdown
Image

unless I hover 
Image

2.
The playground button border has a weird gradient. Not sure it is on purpose but my first reaction was to think something was wrong. (I am on Firefox)
Image

3.
Clicking on Docs or Logs opens the dedicated page, closing the playground and breaking the flow imho. 

4.
The panel, being only ~1/3 of the screen, is pretty small on my laptop. I wish it was wider.

TBonnin:
other comment is more for @bastienbeurier:
The name `Playground` let me think it is all play and fun but since I can use real connections and real actions it can have destructive or mutative consequences if used in prod environments.

packages/webapp/src/components-v2/Playground/index.tsx
kaposke: It's because `SheetContent` in `sheet.tsx` has: `bg-white dark:bg-neutral-950`. Instead of overriding styles here, you should re-style the shadcn component directly (modify `sheet.tsx` with these styles here).

packages/webapp/src/components-v2/Playground/index.tsx
kaposke: `text-heading-medium` doesn't exist, right?

packages/webapp/src/components-v2/Playground/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+                        'w-[537px] max-w-none sm:max-w-none',
+                        'data-[state=open]:fade-in-0 data-[state=closed]:fade-out-0',
+                        '[&>button]:hidden'
+                    )}
+                >
+                    {/* Header */}
+                    <div className="flex w-full items-start justify-between pb-8">
+                        <div className="min-w-0">
+                            <h2 className="text-text-primary text-heading-medium font-medium text-[20px] pb-2">Playground</h2>
+                            <p className="text-body-regular-medium text-text-secondary text-body-medium-medium text-[14px] font-400 line-height-[160%]">
```
kaposke: `text-body-regular-medium` doesn't exist. `text-body-medium-medium` does but looks like you're overriding it's values.

packages/webapp/src/components-v2/Playground/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+                                Quickly run any function.
+                            </p>
+                        </div>
+                        <Button variant="ghost" size="icon" className="size-7 mt-0.5" onClick={() => setPlaygroundOpen(false)} aria-label="Close playground">
+                            <X className="size-4" />
```
kaposke: Buttons handle svg sizes, so no need for defining the size here.
```suggestion
                            <X />
```

packages/webapp/src/components-v2/Playground/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+                    overlayClassName="hidden"
+                    insetTop={108}
+                    insetBottom={44}
+                    insetRight={24}
+                    onInteractOutside={(e) => e.preventDefault()}
+                    onPointerDownOutside={(e) => e.preventDefault()}
+                    onFocusOutside={(e) => e.preventDefault()}
+                    className={cn(
+                        'dark:bg-bg-elevated text-text-primary rounded-lg border border-border-muted shadow-lg p-6',
+                        'flex flex-col items-start gap-[10px]',
```
kaposke: Try to stick with tailwind values when it's reasonable. 1 tailwind unit = 4px
```suggestion
                        'flex flex-col items-start gap-2.5',
```

packages/webapp/src/components-v2/Playground/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+                >
+                    {/* Header */}
+                    <div className="flex w-full items-start justify-between pb-8">
+                        <div className="min-w-0">
+                            <h2 className="text-text-primary text-heading-medium font-medium text-[20px] pb-2">Playground</h2>
+                            <p className="text-body-regular-medium text-text-secondary text-body-medium-medium text-[14px] font-400 line-height-[160%]">
+                                Quickly run any function.
+                            </p>
+                        </div>
+                        <Button variant="ghost" size="icon" className="size-7 mt-0.5" onClick={() => setPlaygroundOpen(false)} aria-label="Close playground">
```
kaposke: I would just use it unstyled (as discussed in person)
```suggestion
                        <Button variant="ghost" size="icon" onClick={() => setPlaygroundOpen(false)} aria-label="Close playground">
```

packages/webapp/src/components-v2/Playground/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+                    className={cn(
+                        'dark:bg-bg-elevated text-text-primary rounded-lg border border-border-muted shadow-lg p-6',
+                        'flex flex-col items-start gap-[10px]',
+                        'w-[537px] max-w-none sm:max-w-none',
+                        'data-[state=open]:fade-in-0 data-[state=closed]:fade-out-0',
+                        '[&>button]:hidden'
+                    )}
+                >
+                    {/* Header */}
+                    <div className="flex w-full items-start justify-between pb-8">
```
kaposke: Avoid using padding and margins to handle spacing. Use parent instead to handle spacing instead.
```suggestion
                    <div className="flex w-full items-start justify-between">
```

packages/webapp/src/components-v2/Playground/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+                        <div className="min-w-0">
+                            <h2 className="text-text-primary text-heading-medium font-medium text-[20px] pb-2">Playground</h2>
```
kaposke: Same comment for spacing. Have the parent with `flex flex-col gap-2` or `spacing-y-2`. I usually stick with flex because I've had weird issues with `spacing-y`.
```suggestion
                            <h2 className="text-text-primary text-heading-medium font-medium text-[20px]">Playground</h2>
```

packages/webapp/src/components-v2/Playground/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+                                queryEnv={queryEnv}
+                                isSync={isSync}
+                                inputFields={inputFields}
+                                inputErrors={inputErrors}
+                                clearInputError={clearInputError}
+                            />
+                        )}
+
+                        {/* Run controls */}
+                        <div className="pt-1 flex gap-2">
```
kaposke: Looks like the spacing is already correctly handled by the parent.
```suggestion
                        <div className="flex gap-2">
```

packages/webapp/src/components-v2/Playground/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+                            {running ? (
+                                <>
+                                    <Button variant="primary" disabled loading={true} size="sm">
+                                        Running
+                                    </Button>
+                                    {isSync && (
+                                        <Button variant="destructive" size="sm" onClick={handleCancel}>
+                                            <X className="size-4" />
```
kaposke: One more instance of not needing to define icon styles in buttons. Will stop commenting on this as you get the gist of it.
```suggestion
                                            <X />
```

packages/webapp/src/components-v2/Playground/index.tsx
```diff
@@ -0,0 +1,168 @@
...
+                )}
+                onClick={() => setPlaygroundOpen(false)}
+            />
+            <Sheet open={playgroundOpen} onOpenChange={setPlaygroundOpen} modal={false}>
+                <SheetContent
+                    side="right"
+                    overlayClassName="hidden"
+                    insetTop={108}
+                    insetBottom={44}
+                    insetRight={24}
```
kaposke: As discussed in person:

```suggestion
                    insetTop={88}
                    insetBottom={24}
                    insetRight={24}
```

+ remove min-height in SheetContent

packages/webapp/src/components-v2/Playground/PlaygroundInputs.tsx
```diff
@@ -0,0 +1,147 @@
...
+    const connectionDetailsQuery = useConnection(
+        { env: queryEnv, provider_config_key: playgroundIntegration || '' },
+        { connectionId: playgroundConnection || '' }
+    );
+    const connectionMetadata = connectionDetailsQuery.data?.connection?.metadata ?? null;
+
+    if (isSync) {
+        return (
+            <div className="grid grid-cols-[110px_1fr] gap-x-4">
+                <label className="text-text-primary text-label-large">Metadata</label>
```
kaposke, line 38: As discussed in person as well, worth creating these non-existing text classes.

packages/webapp/src/components-v2/Playground/PlaygroundInputs.tsx
```diff
@@ -0,0 +1,147 @@
...
+                <div className="min-w-0 flex flex-col gap-3">
+                    <Alert variant="info" className="px-3 py-2" actionsBelow>
+                        <Info className="size-4" />
```
kaposke: Icon size is handled by Alert
```suggestion
                        <Info  />
```

packages/webapp/src/components-v2/Playground/PlaygroundInputs.tsx
```diff
@@ -0,0 +1,147 @@
...
+                                    : undefined;
+                            const isObjectValue = rawValue !== null && rawValue !== undefined && typeof rawValue === 'object';
+                            return (
+                                <div key={field.name} className="flex flex-col gap-1">
+                                    <div className="flex items-center justify-between gap-3">
+                                        <span className="text-text-primary text-body-medium-medium">
+                                            {field.name}
+                                            {field.required && <span className="text-feedback-error-fg text-body-medium-medium">*</span>}
+                                        </span>
+                                        <Badge variant="gray" size="xs" className="capitalize text-system-label-small">
```
kaposke: I don't think it's supposed to be capitalized. Also, consider extracting the `CatalogBadge` component from the integrations folder. Here's the component in use:
https://github.com/nangohq/nango/blob/agus/NAN-4846/playground/packages/webapp/src/pages/Integrations/components/jsonSchema/JsonSchema.tsx#L82

It has "light" and "dark" variants because of how we invert colors when nesting objects in the function page, but you'll only use one variant for your use-case.

packages/webapp/src/components-v2/Playground/PlaygroundInputs.tsx
```diff
@@ -0,0 +1,147 @@
...
+            <label className="text-text-primary text-label-large">Inputs</label>
+            <div className="min-w-0 flex flex-col gap-3">
+                {inputFields.map((field) => (
+                    <div key={field.name} className="flex flex-col gap-1">
+                        <div className="flex items-center justify-between gap-3">
+                            <span className="text-text-primary text-body-medium-medium">
+                                {field.name}
+                                {field.required && <span className="text-feedback-error-fg text-body-medium-medium">*</span>}
+                            </span>
+                            <Badge variant="gray" size="xs" className="capitalize font-mono shrink-0">
```
kaposke: Same comment on using the existing badge and not capitalizing

packages/webapp/src/components-v2/Playground/PlaygroundResult.tsx
```diff
@@ -0,0 +1,96 @@
...
+            <div className="flex flex-col gap-3">
+                <p className="text-text-primary text-body-small-semi">Results</p>
+
+                <Alert variant={result.state === 'waiting' || result.state === 'running' ? 'info' : result.success ? 'success' : 'error'} className="px-3 py-2">
+                    {result.state === 'waiting' || result.state === 'running' ? (
+                        <Info className="size-4" />
+                    ) : result.success ? (
+                        <CheckCircle2 className="size-4" />
+                    ) : (
+                        <XCircle className="size-4" />
```
kaposke: No need to style the svgs

packages/webapp/src/components-v2/Playground/PlaygroundResult.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+    let resultJson = '';
+    try {
+        resultJson = JSON.stringify(result.data, null, 2);
+    } catch {
+        resultJson = String(result.data);
+    }
+    if (resultJson.length >= JSON_DISPLAY_LIMIT) {
+        resultJson = 'Result too large to display';
+    }
```
kaposke: NIT: do it in a useMemo so it's not happening on re-renders

packages/webapp/src/components-v2/Playground/PlaygroundSelectors.tsx
```diff
@@ -0,0 +1,234 @@
...
+    const connectionSearch = usePlaygroundStore((s) => s.connectionSearch);
+    const setPlaygroundOpen = usePlaygroundStore((s) => s.setOpen);
+    const setPlaygroundIntegration = usePlaygroundStore((s) => s.setIntegration);
+    const setPlaygroundConnection = usePlaygroundStore((s) => s.setConnection);
+    const setPlaygroundFunction = usePlaygroundStore((s) => s.setFunction);
+    const setPlaygroundResult = usePlaygroundStore((s) => s.setResult);
+    const setPlaygroundInputErrors = usePlaygroundStore((s) => s.setInputErrors);
+    const setPlaygroundConnectionSearch = usePlaygroundStore((s) => s.setConnectionSearch);
+    const setPlaygroundPendingOperationId = usePlaygroundStore((s) => s.setPendingOperationId);
+    const setPlaygroundRunning = usePlaygroundStore((s) => s.setRunning);
```
kaposke, line 37: Honest question, is it possible to just:
```ts
const { playgroundIntegration, playgroundConnection, .... } = usePlaygroundStore((s) => s);
```
?

packages/webapp/src/components-v2/Playground/PlaygroundSelectors.tsx
```diff
@@ -0,0 +1,234 @@
...
+    const [debouncedConnectionSearch, setDebouncedConnectionSearch] = useState('');
+
+    useEffect(() => {
+        if (!connectionSearch) {
+            setDebouncedConnectionSearch('');
+            return;
+        }
+        const t = window.setTimeout(() => setDebouncedConnectionSearch(connectionSearch), 250);
+        return () => window.clearTimeout(t);
+    }, [connectionSearch]);
```
kaposke: [for all the debounced searches here]
We have a library called `react-use` that has some useful hooks. One of them is `useDebounce`. Here's an example of it being used in searching connections in the connection list:
https://github.com/nangohq/nango/blob/agus/NAN-4846/playground/packages/webapp/src/pages/Connection/List.tsx#L147-L153

packages/webapp/src/components-v2/Playground/PlaygroundSelectors.tsx
agusayerza: Full disclouse, this is claude talking: individual selectors are better as they will not trigger re-render un unrelated changes.

Now me: On this case I don't think it matters regardless as any/all changes on the playground state would trigger a re-render regardless and the impact is minimal

packages/webapp/src/components-v2/Playground/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+                                            {isSync && (
+                                                <Button variant="destructive" size="sm" onClick={handleCancel}>
+                                                    <X />
+                                                    Cancel run
+                                                </Button>
+                                            )}
+                                        </>
+                                    ) : result ? (
+                                        <Button variant="primary" size="sm" onClick={handleRun} disabled={!canRun}>
+                                            <RotateCcw className="size-4" />
```
kaposke:
```suggestion
                                            <RotateCcw />
```

packages/webapp/src/components-v2/Playground/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+                                            Run again
+                                        </Button>
+                                    ) : (
+                                        <Button variant="primary" size="sm" onClick={handleRun} disabled={!canRun}>
+                                            <Play className="size-4" />
```
kaposke:
```suggestion
                                            <Play />
```

packages/webapp/src/components-v2/Playground/PlaygroundInputs.tsx
```diff
@@ -0,0 +1,143 @@
...
+                            )}
+                            <AlertButton asChild variant="info">
+                                <a
+                                    href="https://nango.dev/docs/implementation-guides/use-cases/customer-configuration"
+                                    target="_blank"
+                                    rel="noopener noreferrer"
+                                >
+                                    Docs <ExternalLink className="size-4" />
+                                </a>
+                            </AlertButton>
```
kaposke: Surprisingly, we have AlertButtonLink. Wyt?

packages/webapp/src/components-v2/Playground/usePlaygroundRun.ts
```diff
@@ -0,0 +1,386 @@
...
+        }
+    }
+    if ((field.type === 'number' || field.type === 'integer') && typeof value === 'number') {
+        if (field.minimum != null && value < field.minimum) return `Must be ≥ ${field.minimum}`;
+        if (field.maximum != null && value > field.maximum) return `Must be ≤ ${field.maximum}`;
+        if (field.exclusiveMinimum != null && value <= field.exclusiveMinimum) return `Must be > ${field.exclusiveMinimum}`;
+        if (field.exclusiveMaximum != null && value >= field.exclusiveMaximum) return `Must be < ${field.exclusiveMaximum}`;
+    }
+    return null;
+}
```
kaposke: zod v4 has `.fromJsonSchema` so in the future we could simplify validation. But ok for now as updating to a new major is often not as trivial and the PR is big enough

packages/webapp/src/components-v2/Playground/index.tsx
```diff
@@ -0,0 +1,169 @@
...
+    const clearPlaygroundInputError = usePlaygroundStore((s) => s.clearInputError);
+
+    const location = useLocation();
+
+    useEffect(() => {
+        // Auto-close the sheet when the route changes
+        if (playgroundOpen) {
+            setPlaygroundOpen(false);
+        }
+    }, [location.pathname]);
```
kaposke, line 45: nit: Add eslint ignore deps comment

packages/webapp/src/components-v2/Playground/usePlaygroundReattach.ts
```diff
@@ -0,0 +1,204 @@
...
+ * When the playground opens and a pendingOperationId is stored, re-attaches to
+ * that operation and polls until it reaches a terminal state. Sets running=true
+ * during the wait so the UI shows the running state just as if the run were live.
+ *
+ * Polling backs off from 1.5s to 10s after the first 30s to reduce server load
+ * for long-running syncs.
+ *
+ * Called lazily — only activates when the sheet is open.
+ */
+export function usePlaygroundReattach() {
```
kaposke: As discussed in person, I don't think having 2 separate hooks is necessary. However, if you do merge them, do refactor and simplify the code. High level thoughts would be:
- Extract json-schema validation
- Extract `findOperation`
- Maybe use `useQuery` and manage polling with it (I think it could simplify the code but I can't see the full picture, I would try asking an agent and see how it looks)

In short, keep hook very readable, extract most logic.

packages/webapp/src/components-v2/AppHeader.tsx
```diff
@@ -1,14 +1,36 @@
...
+                        )}
+                    >
+                        <Box className="size-4" />
```
kaposke:
```suggestion
                        <Box />
```

packages/webapp/src/components-v2/Playground/PlaygroundSelectors.tsx
```diff
@@ -0,0 +1,227 @@
...
+                            type="button"
+                            variant="secondary"
+                            size="sm"
+                            className="h-auto rounded-full bg-btn-secondary-bg px-2 py-1 text-body-small-regular gap-0.5 justify-center items-center text-text-primary"
+                            onClick={() => {
+                                setPlaygroundOpen(false);
+                                navigate(`/${env}/integrations/create`);
+                            }}
+                        >
+                            <Plus className="size-4" /> Add
```
kaposke:
```suggestion
                            <Plus /> Add
```

packages/webapp/src/components-v2/Playground/PlaygroundSelectors.tsx
```diff
@@ -0,0 +1,227 @@
...
+                            type="button"
+                            variant="secondary"
+                            size="sm"
+                            className="h-auto rounded-full bg-btn-secondary-bg px-2 py-1 text-body-small-regular gap-0.5 justify-center items-center text-text-primary"
+                            onClick={() => {
+                                setPlaygroundOpen(false);
+                                navigate(`/${env}/connections/create${playgroundIntegration ? `?integration_id=${playgroundIntegration}` : ''}`);
+                            }}
+                        >
+                            <Plus className="size-4" /> Add
```
kaposke:
```suggestion
                            <Plus /> Add
```

packages/webapp/src/components-v2/Playground/PlaygroundSelectors.tsx
```diff
@@ -0,0 +1,227 @@
...
+                                type="button"
+                                variant="secondary"
+                                size="sm"
+                                className="h-auto rounded-full bg-btn-secondary-bg px-2 py-1 text-body-small-regular gap-0.5 justify-center items-center text-text-primary"
+                                onClick={() => {
+                                    setPlaygroundOpen(false);
+                                    navigate(`/${env}/integrations/${playgroundIntegration}`);
+                                }}
+                            >
+                                Activate <ExternalLink className="size-4" />
```
kaposke:
```suggestion
                                Activate <ExternalLink />
```

packages/webapp/src/utils/logs.ts
```diff
@@ -3,31 +3,45 @@ import { addMilliseconds, startOfDay, subDays, subHours, subMinutes } from 'date
...
 
-    usp.set('live', 'false');
+    setParam(usp, 'operationId', options.operationId != null ? String(options.operationId) : undefined);
+    setParam(usp, 'search', options.search);
+    setParam(usp, 'limit', options.limit);
+    setParam(usp, 'states', options.states);
+    setParam(usp, 'types', options.types);
+    setParam(usp, 'integrations', options.integrations);
+    setParam(usp, 'connections', options.connections);
+    setParam(usp, 'syncs', options.syncs);
```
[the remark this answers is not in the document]
agusayerza: OOS for this PR

packages/webapp/src/components-v2/Playground/usePlayground.ts
```diff
@@ -0,0 +1,232 @@
...
+    // --- useQuery: poll operation status when pendingOperationId is set ---
+    const { data: operationData } = useQuery({
+        queryKey: ['playground-operation', env, pendingOperationId],
+        queryFn: () => fetchOperation(pendingOperationId!, env),
+        enabled: !!pendingOperationId && isOpen,
+        refetchInterval: (query) => {
+            const state = query.state.data?.state;
+            return state === 'running' || state === 'waiting' ? STATUS_POLL_INTERVAL_MS : false;
+        }
+    });
```
kaposke, line 63: Turned out good!

kaposke:
@codex review

packages/webapp/src/components-v2/ui/dropdown-menu.tsx
```diff
@@ -72,7 +72,7 @@ function DropdownMenuCheckboxItem({ className, children, checked, ...props }: Re
         >
             <span className="pointer-events-none absolute left-2 flex size-3.5 items-center justify-center">
                 <DropdownMenuPrimitive.ItemIndicator>
-                    <CheckIcon className="size-4" />
+                    <CheckIcon />
```
kaposke: The DropdownPrimitive handles svg sizes?

packages/webapp/src/components-v2/ui/select.tsx
```diff
@@ -112,7 +112,7 @@ function SelectScrollUpButton({ className, ...props }: React.ComponentProps<type
             className={cn('flex cursor-default items-center justify-center py-1', className)}
             {...props}
         >
-            <ChevronUpIcon className="size-4" />
+            <ChevronUpIcon />
```
kaposke: Same, does this SelectPrimitive handle svg sizes?

packages/webapp/src/components-v2/AppHeader.tsx
```diff
@@ -1,14 +1,52 @@
...
+    const { data: envData } = useEnvironment(env);
+    const environment = envData?.environmentAndAccount?.environment;
+    const { can } = usePermissions();
+    const canUsePlayground = envData != null && (can(permissions.canUseProdPlayground) || !environment?.is_production);
+
     return (
         <header className="h-16 px-10 pl-2 py-2.5 items-center flex justify-between shrink-0 gap-1.5">
             <Breadcrumbs />
             <div className="flex gap-1.5 justify-end">
+                <PermissionGate condition={canUsePlayground}>
```
kaposke, line 28: Codex seems to be right. This is a prod env, with a support role. I could open the playground and send the request, but it failed with invalid api key (which is expected).
Image

packages/webapp/src/components-v2/ui/dropdown-menu.tsx
agusayerza: Reverted, it was inherinting size from the span, but even then it was wrong as it was slightly smaller

packages/webapp/src/components-v2/ui/select.tsx
agusayerza: Same

agusayerza:
@codex review

packages/webapp/src/components-v2/Playground/usePlayground.ts
```diff
@@ -0,0 +1,232 @@
...
+                setResult({ success: false, data: triggerData, durationMs: triggerDurationMs });
+                setRunning(false);
+                return;
+            }
+
+            // Poll until we find the matching operation in logs.
+            const findDeadlineMs = playgroundFunctionType === 'sync' ? 15_000 : 5_000;
+            const findStart = Date.now();
+            let operation = null as Awaited<ReturnType<typeof findOperation>>;
+            while (Date.now() - findStart < findDeadlineMs) {
```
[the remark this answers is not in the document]
agusayerza: Real but not a huge problem. Actions wait up to 5s for the log lookup, but they still succeed correctly. The optimization (skip polling entirely for actions) would be a nice-to-have but is a bigger refactor of the polling flow, introducing more friction to an already big friction PR

packages/server/lib/controllers/v1/trigger/postTriggerFunction.integration.test.ts
```diff
@@ -0,0 +1,166 @@
...
+    success: true,
+    response: true,
+    error: null
+});
+
+describe(`POST ${endpoint}`, () => {
+    beforeAll(async () => {
+        api = await runServer();
+        envs.NANGO_LOGS_ENABLED = false;
+    });
```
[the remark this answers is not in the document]
agusayerza: 2-liner, might as well fix it

packages/webapp/src/components-v2/Playground/usePlayground.ts
```diff
@@ -0,0 +1,232 @@
...
+
+        // For syncs, also cancel the backend operation (best-effort)
+        if (playgroundFunctionType === 'sync' && playgroundIntegration && playgroundConnection && playgroundFunction) {
+            try {
+                const res = await apiFetch(
+                    `/api/v1/sync?env=${env}&connection_id=${encodeURIComponent(playgroundConnection)}&provider_config_key=${encodeURIComponent(playgroundIntegration)}`
+                );
+                if (res.ok) {
+                    const syncs = (await res.json()) as SyncResponse[];
+                    const sync = syncs.find((s) => s.name === playgroundFunction);
```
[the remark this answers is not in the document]
agusayerza: Ignoring for now, same stand as above

packages/server/lib/controllers/v1/trigger/postTriggerFunction.ts
```diff
@@ -0,0 +1,163 @@
...
+    const valBody = bodySchema.safeParse(req.body);
+    if (!valBody.success) {
+        res.status(400).send({ error: { code: 'invalid_body', errors: zodErrorToHTTP(valBody.error) } });
+        return;
+    }
+
+    const { account, environment } = res.locals;
+    const { type, function_name, provider_config_key, connection_id } = valBody.data;
+
+    if (type === 'action') {
```
TBonnin, line 45: should we call `PostTriggerAction` handler here (or consolidate the logic) because it is basically duplication of almost the same code

TBonnin: or even reuse the trigger endpoints themselves from the frontend?

agusayerza: Extracted the logic and reusing now. I was doubting on doing it originally as I wanted to reduce the blast radius on existing features as much as possible, but the refactor to extract to share logic is simple enough

packages/webapp/src/components-v2/Playground/playground.utils.ts
```diff
@@ -0,0 +1,226 @@
...
+    functionType: 'action' | 'sync';
+    integration: string;
+    connection: string;
+    functionName: string;
+}
+
+export async function findOperation(params: FindOperationParams, signal: AbortSignal): Promise<SearchOperations['Success']['data'][number] | null> {
+    const { env, triggerStartTime, functionType, integration, connection, functionName } = params;
+    const from = new Date(triggerStartTime - 60_000).toISOString();
+    const to = new Date().toISOString();
```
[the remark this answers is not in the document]
agusayerza: Will be fixed with the followup on trigger returning the actual op id

kaposke (approved):

PR #5590, opened 2026-03-09
fix(cli): ignore axios default urlencoded Content-Type in unified mocks

marcindobry:
Axios injects Content-Type: application/x-www-form-urlencoded for body-less POST/PUT/PATCH requests. Unified mock recording hashes the actual axios request config, but test playback hashes the user-provided proxy config (which often omits this header), causing requestIdentityHash mismatches and strict header matching failures ("No mock found ...").

Treat the urlencoded content-type as an ignorable axios default in both recording (ResponseCollector) and unified identity normalization, and relax unified request-header matching to consider it optional.

Backwards compatibility: existing .test.json fixtures that already include the urlencoded content-type (and hashes computed with it) still match via request matching, while new recordings omit the header so identities stay stable.

Add a regression test covering the body-less POST scenario.

 

---

This also introduces a shared helper to consistently detect axios default content types so recording, identity normalization, and header matching stay aligned.

---

### DIFF
--- a/packages/cli/lib/services/response-collector.service.ts
+++ b/packages/cli/lib/services/response-collector.service.ts
@@ -65,6 +65,15 @@ export const FILTER_HEADERS: string[] = [
     'retry-on'
 ];
 
+export function isAxiosDefaultContentTypeForMockIdentity(headerKey: string, headerValue: unknown): boolean {
+    if (headerKey.toLowerCase() !== 'content-type') {
+        return false;
+    }
+
+    const normalizedValue = String(headerValue).toLowerCase();
+    return normalizedValue === 'application/json' || normalizedValue === 'undefined' || normalizedValue.startsWith('application/x-www-form-urlencoded');
+}
+
 export class ResponseCollector {
     private apiCalls: CachedRequest[] = [];
     private connectionInfo?: { connectionId: string; provider: string };
@@ -245,8 +254,10 @@ export class ResponseCollector {
                     // Skip filtered
                     if (FILTER_HEADERS.includes(lowerKey)) return false;
 
-                    // Skip application/json
-                    if (lowerKey === 'content-type' && (value.toLowerCase() === 'application/json' || value === 'undefined')) return false;
+                    // Skip default content-types injected by axios.
+                    if (isAxiosDefaultContentTypeForMockIdentity(lowerKey, value)) {
+                        return false;
+                    }
 
                     return true;
                 });
--- a/packages/cli/lib/testMocks/utils.ts
+++ b/packages/cli/lib/testMocks/utils.ts
@@ -8,7 +8,7 @@ import { vi } from 'vitest';
 import { getProvider } from '@nangohq/providers';
 import { PaginationService } from '@nangohq/runner-sdk';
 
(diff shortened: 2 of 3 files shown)


packages/cli/lib/testMocks/utils.ts
```diff
@@ -427,22 +427,42 @@ class UnifiedFixtureProvider implements FixtureProvider {
                         }
                     }
 
-                    const mockHeaderCount = mock.request.headers ? Object.keys(mock.request.headers).length : 0;
-                    const requestHeaderCount = identity.requestIdentity.headers.length;
+                    const normalizeHeaderEntries = (entries: [string, unknown][]): Map<string, string> => {
```
marcindobry, line 430: This part is mostly for backwards compatibility.

Because we have already filtered `application/json` and `undefined` on both sides (recording and replaying), the hashes wouldn't have included them. However, someone might have figured out that the only way for the empty-body POST mocks to work is to manually pass in `application/x-www-form-urlencoded`. In that case, since we now filter the header on replay, the hash won't match, so we need to filter out this manually added header before we do the header comparison.

packages/cli/lib/services/response-collector.service.ts
```diff
@@ -245,8 +245,14 @@ export class ResponseCollector {
                     // Skip filtered
                     if (FILTER_HEADERS.includes(lowerKey)) return false;
 
-                    // Skip application/json
-                    if (lowerKey === 'content-type' && (value.toLowerCase() === 'application/json' || value === 'undefined')) return false;
+                    // Skip default content-types injected by axios.
+                    const lowerValue = value.toLowerCase();
+                    if (
+                        lowerKey === 'content-type' &&
+                        (lowerValue === 'application/json' || lowerValue === 'undefined' || lowerValue.startsWith('application/x-www-form-urlencoded'))
```
TBonnin: should this be using the `isAxiosDefaultFormContentType` helper?

packages/cli/lib/testMocks/utils.ts
```diff
@@ -1260,9 +1280,13 @@ function normalizeHeadersForUnifiedIdentity(headers: UserProvidedProxyConfigurat
...
 
         const value = String(rawValue);
+        const lowerValue = value.toLowerCase();
 
         // Match ResponseCollector behavior for axios defaults.
-        if (lowerKey === 'content-type' && (value.toLowerCase() === 'application/json' || value === 'undefined')) {
+        if (
+            lowerKey === 'content-type' &&
+            (lowerValue === 'application/json' || lowerValue === 'undefined' || lowerValue.startsWith('application/x-www-form-urlencoded'))
+        ) {
```
TBonnin: can we consolidate this logic that is also used in ResponseCollector?

TBonnin (approved):

packages/cli/lib/services/response-collector.service.ts
marcindobry: In retrospect `isAxiosDefaultFormContentType` is a bad name because it's only for the x-www-form-urlencoded. I'll rename that and make a new one that checks for all the headers

PR #5601, opened 2026-03-11
fix(lambda): support large payloads in Lambda

rossmcewan:
---

**Support oversized Lambda payloads via S3 offload**

This PR adds a large-payload handling path for Lambda invocations by conditionally offloading `code` and `codeParams` to S3 when the serialized payload exceeds the configured size threshold. The `LambdaRuntimeAdapter` now uploads gzipped payload artifacts to S3, includes S3 references in the invoke payload, and enforces a hard payload size limit.

On the Lambda runner side, the request schema accepts either inline payloads or S3 references, retrieves and gunzips S3 content, and attempts to delete `codeParams` objects after execution. Environment parsing and Lambda function configuration are updated to pass new payload-related env vars, and unit tests are added to cover both inline and S3 paths.

---


packages/jobs/lib/runner/lambda.ts
```diff
@@ -160,23 +160,30 @@ class Lambda {
...
-                NANGO_CUSTOMER_REDIS_URL: envs.NANGO_CUSTOMER_REDIS_URL || envs.NANGO_REDIS_URL || '',
-                NANGO_TELEMETRY_SDK: String(envs.NANGO_TELEMETRY_SDK),
-                DD_ENV: envs.DD_ENV || '',
-                DD_SITE: envs.DD_SITE || '',
-                DD_PROFILING_ENABLED: String(node.isProfilingEnabled),
-                DD_APM_TRACING_ENABLED: String(node.isTracingEnabled),
-                DD_TRACE_ENABLED: String(node.isTracingEnabled || node.isProfilingEnabled),
-                DD_API_KEY_SECRET_ARN: envs.DD_API_KEY_SECRET_ARN || ''
-            }
+            Variables: variables
```
TBonnin: the rest of the codebase, as well as the rest of your PR, uses the spread operator to deal with such scenario instead of mutating temporary variables
```
return {
   ...(envs.LAMBDA_PAYLOAD_MAX_SIZE_BYTES && { LAMBDA_PAYLOAD_MAX_SIZE_BYTES: String(envs.LAMBDA_PAYLOAD_MAX_SIZE_BYTES) })
}
```

packages/jobs/lib/runtime/lambda.adapter.ts
```diff
@@ -34,17 +48,115 @@ export class LambdaRuntimeAdapter implements RuntimeAdapter {
         };
     }
 
+    private getPayloadsBucket(): string {
```
TBonnin: let's follow convention and use `Result<string>` to make it explicit this function can fail

packages/jobs/lib/runtime/lambda.adapter.ts
```diff
@@ -34,17 +48,115 @@ export class LambdaRuntimeAdapter implements RuntimeAdapter {
...
+    }): Promise<S3ObjectRef> {
+        if (params.skipIfExists) {
+            try {
+                const head = await s3.send(new HeadObjectCommand({ Bucket: params.bucket, Key: params.key }));
+                return {
+                    kind: 's3' as const,
+                    bucket: params.bucket,
+                    key: params.key,
+                    ...(head.VersionId !== undefined && { versionId: head.VersionId }),
+                    ...(head.ETag !== undefined && { etag: head.ETag.replace(/"/g, '') })
```
TBonnin:
```suggestion
                    ...(head.VersionId && { versionId: head.VersionId }),
                    ...(head.ETag && { etag: head.ETag.replace(/"/g, '') })
```
unless you want to have it set when empty or null

packages/jobs/lib/runtime/lambda.adapter.ts
```diff
@@ -34,17 +48,115 @@ export class LambdaRuntimeAdapter implements RuntimeAdapter {
...
+            ...(put.VersionId !== undefined && { versionId: put.VersionId }),
+            ...(put.ETag !== undefined && { etag: put.ETag.replace(/"/g, '') })
+        };
+    }
+
+    private async uploadCode(params: { taskId: string; nangoProps: NangoProps; code: string }): Promise<S3ObjectRef> {
+        const bucket = this.getPayloadsBucket();
+        const body = gzipSync(Buffer.from(params.code, 'utf8'));
+        const hash = createHash('sha256').update(params.code).digest('hex');
+        const key = `environments/${params.nangoProps.environmentId}/function-code/${hash}.cjs.gz`;
```
TBonnin: would be nice to have the account as the key prefix. Would make it easier for debugging

packages/lambda-runner/lib/index.ts
```diff
@@ -46,6 +51,59 @@ function getNangoHost() {
...
+                    Bucket: request.codeParamsRef.bucket,
+                    Key: request.codeParamsRef.key
+                })
+            );
+        }
+    } catch (err) {
+        logger.error('Error deleting code params', { error: err });
+        //ignore error - the payload will be deleted by the bucket lifecycle policy
+    }
+}
```
TBonnin, line 86: can we just rely on the bucket lifecycle policy and avoid explicit deletion?

packages/lambda-runner/lib/schemas.ts
```diff
@@ -83,6 +83,24 @@ export const nangoPropsSchema = zod.object({
...
+        .object({
+            kind: zod.literal('s3'),
+            bucket: zod.string(),
+            key: zod.string(),
+            versionId: zod.string().optional(),
+            etag: zod.string().optional()
+        })
+        .optional(),
     nangoProps: nangoPropsSchema
 });
```
TBonnin, line 95: Refs and non refs parameters are exclusive. no?
```
const s3RefSchema = zod.object({
    kind: zod.literal('s3'),
    bucket: zod.string(),
    key: zod.string(),
    versionId: zod.string().optional(),
    etag: zod.string().optional()
});

const inlineCodeSchema = zod.object({
    code: zod.string(),
    codeParams: zod.any().optional()
});

const refCodeSchema = zod.object({
    codeRef: s3RefSchema,
    codeParamsRef: s3RefSchema.optional()
});

export const requestSchema = zod
    .object({
        taskId: zod.string(),
        nangoProps: nangoPropsSchema
    })
    .and(zod.union([inlineCodeSchema, refCodeSchema]));
```

packages/lambda-runner/lib/index.ts
```diff
@@ -111,29 +169,12 @@ export const handler = async (event: zod.infer<typeof requestSchema>, context: C
...
-                status: 500
-            },
-            telemetryBag: {
-                customLogs: 0,
-                proxyCalls: 0,
-                durationMs: Date.now() - startTime,
-                memoryGb: Number(context.memoryLimitInMB) / 1024
-            },
-            functionRuntime: 'lambda'
-        });
```
TBonnin, line 131: why remove the task update?

packages/lambda-runner/lib/index.ts
```diff
@@ -94,8 +152,8 @@ export const handler = async (event: zod.infer<typeof requestSchema>, context: C
     try {
         const payload = {
             nangoProps: { ...nangoProps, host: getNangoHost() },
-            code: request.code,
-            codeParams: request.codeParams,
+            code: await getCode(request),
+            codeParams: await getCodeParams(request),
```
TBonnin: can those function be called concurrently with `Promise.all`?

packages/jobs/lib/runtime/lambda.adapter.ts
```diff
@@ -34,17 +48,115 @@ export class LambdaRuntimeAdapter implements RuntimeAdapter {
...
+        return {
+            kind: 's3' as const,
+            bucket: params.bucket,
+            key: params.key,
+            ...(put.VersionId !== undefined && { versionId: put.VersionId }),
+            ...(put.ETag !== undefined && { etag: put.ETag.replace(/"/g, '') })
+        };
+    }
+
+    private async uploadCode(params: { taskId: string; nangoProps: NangoProps; code: string }): Promise<S3ObjectRef> {
```
TBonnin: taskId is unused. no?

packages/jobs/lib/runtime/lambda.adapter.ts
```diff
@@ -34,17 +48,115 @@ export class LambdaRuntimeAdapter implements RuntimeAdapter {
...
+        if (Buffer.byteLength(payloadString, 'utf-8') > envs.LAMBDA_PAYLOAD_MAX_SIZE_BYTES && envs.LAMBDA_PAYLOADS_BUCKET_NAME) {
+            const [codeRef, codeParamsRef] = await Promise.all([this.uploadCode(params), this.uploadCodeParams(params)]);
+            return JSON.stringify({
+                taskId: params.taskId,
+                nangoProps: params.nangoProps,
+                codeRef: codeRef,
+                codeParamsRef: codeParamsRef
+            });
+        } else {
+            return JSON.stringify(payload);
```
TBonnin: let's reuse payloadString instead of stringifying it again

packages/jobs/lib/runtime/lambda.adapter.ts
```diff
@@ -34,17 +48,115 @@ export class LambdaRuntimeAdapter implements RuntimeAdapter {
...
+            }
+        }
+        const put = await s3.send(
+            new PutObjectCommand({
+                Bucket: params.bucket,
+                Key: params.key,
+                Body: params.body,
+                ContentType: params.contentType,
+                ...(params.contentEncoding && { ContentEncoding: params.contentEncoding }),
+                Tagging: params.tagging
```
TBonnin: tagging can be undefined too. Shouldn't be spread conditionally?

packages/jobs/lib/runtime/lambda.adapter.ts
```diff
@@ -34,17 +48,115 @@ export class LambdaRuntimeAdapter implements RuntimeAdapter {
...
+
+    private async preparePayload(params: { taskId: string; nangoProps: NangoProps; code: string; codeParams: object }): Promise<string> {
+        const payload = {
+            taskId: params.taskId,
+            nangoProps: params.nangoProps,
+            code: params.code,
+            codeParams: params.codeParams
+        };
+        const payloadString = JSON.stringify(payload);
+        if (Buffer.byteLength(payloadString, 'utf-8') > envs.LAMBDA_PAYLOAD_MAX_SIZE_BYTES && envs.LAMBDA_PAYLOADS_BUCKET_NAME) {
```
TBonnin: we should have a hard limit though before somebody tries to submit a payload in GBs :)

packages/lambda-runner/lib/index.ts
rossmcewan: I would rather get rid of it as soon as possible tbh. There is potentially sensitive or PII in these payloads.

packages/lambda-runner/lib/index.ts
rossmcewan: The Lambda destinations will take care of any of these failures. Thinking about it a bit more though, it's probably cleaner to keep it here and keep the destinations for truly unmanaged failures. I'll put it back.

packages/jobs/lib/runtime/lambda.adapter.ts
rossmcewan: Indeed. But this can be a follow up, imo. We can control it from the API.

packages/jobs/lib/runtime/lambda.adapter.ts
```diff
@@ -34,17 +48,123 @@ export class LambdaRuntimeAdapter implements RuntimeAdapter {
...
 
+    private getPayloadsBucket(): Result<string> {
+        const bucket = envs.LAMBDA_PAYLOADS_BUCKET_NAME;
+        if (!bucket) {
+            return Err(new Error('LAMBDA_PAYLOADS_BUCKET_NAME is not set'));
+        }
+        return Ok(bucket);
+    }
+
+    private async uploadToS3(params: {
```
pfreixes, line 59: 💅 should we have some metrics around this function for knowing latencies and specific errors related to `uploadToS3`?

packages/jobs/lib/runtime/lambda.adapter.ts
pfreixes: Yea agree on having a limit, and I would say that a hard limit in the code would be also more than a nice to have so nobody can overwrite this to a crazy value.

Also we need to consider the cost in terms of CPU when gziping or calculating the sha of objects of hundreds of MBs, which will for sure stop the event loop of Node.js

If current limitation was 1MB, should we start with a reasonable limit of lets say < 100MB?

packages/jobs/lib/runtime/lambda.adapter.ts
```diff
@@ -34,17 +48,123 @@ export class LambdaRuntimeAdapter implements RuntimeAdapter {
...
+        key: string;
+        body: Buffer;
+        contentType: string;
+        contentEncoding?: 'gzip';
+        skipIfExists?: boolean;
+        tagging?: string;
+    }): Promise<S3ObjectRef> {
+        if (params.skipIfExists) {
+            try {
+                const head = await s3.send(new HeadObjectCommand({ Bucket: params.bucket, Key: params.key }));
```
pfreixes, line 70: That is a good optimization! but it can lead to some data races when using algo the lifecycle [rule](https://github.com/NangoHQ/nango-infra/pull/90/changes#diff-1bb737394511c3a65b017c19a595abc86064a5b9b69ef8d097fe867eca4de031R224)

If Im not mistaken this can happen:
- Our runtime does not uplaoad the code (same for params) bc they are already in S3
- The runtime send an async event
-  S3 purges the object
- Tha lambda can not recover the object bc was sremoved

I know that its un unlikely event, but it can happen

If we want we could mitigate this by checking the `Last-Modified` field returned by the HEAD and skip the upload if for example if we have still 1 day ahead before the payload is removed

packages/jobs/lib/runtime/lambda.adapter.ts
```diff
@@ -34,17 +48,123 @@ export class LambdaRuntimeAdapter implements RuntimeAdapter {
...
+            bucket,
+            key,
+            body,
+            contentType: 'application/json',
+            contentEncoding: 'gzip',
+            tagging: `type=function-params`
+        });
+    }
+
+    private async preparePayload(params: { taskId: string; nangoProps: NangoProps; code: string; codeParams: object }): Promise<string> {
```
pfreixes: 💅 should we return an object and make the serialization [here](https://github.com/NangoHQ/nango/pull/5601/changes#diff-27e2c59743d02c31316615939870bf6253a632f8b5a496b213f8599d999abfc1R167)?

packages/lambda-runner/lib/index.ts
```diff
@@ -92,10 +150,11 @@ export const handler = async (event: zod.infer<typeof requestSchema>, context: C
         }
     }, heartbeatIntervalMs);
     try {
+        const [code, codeParams] = await Promise.all([getCode(request), getCodeParams(request)]);
```
pfreixes, line 153: If Im not mistaken we are returning empty strings for some error code paths - like [this](https://github.com/NangoHQ/nango/pull/5601/changes#diff-0bc180e6141929bdbf130a68a9fafecbcc7796cc83d2dbe2af381a27f29dfae0R54) - , should we just return in cases like this? otherwise we will try to execute something that is broken

Side discussion, how are we  observing here errors like this? do we have DD enabled?

packages/lambda-runner/lib/index.ts
```diff
@@ -134,6 +193,7 @@ export const handler = async (event: zod.infer<typeof requestSchema>, context: C
         if (pass.lock) {
             await gate.exit(pass.lock);
         }
+        await deleteCodeParams(request);
```
pfreixes, line 199: If we are eager to delete the code params always, should we avoid using the HEAD strategy and create a unique key for this object?

Otherwise there is a risk that we could be deleting an object related to a code param that might need to be attended by another lambda

pfreixes (commented):
Left some comments, there are couple of data races that could be concerning unless Im missing something!

packages/lambda-runner/lib/index.ts
rossmcewan: Yes, we have full observability through Datadog of Lambda.

packages/lambda-runner/lib/index.ts
rossmcewan: We don't use the HEAD strategy for code params, only for code. They are unique per task.

packages/lambda-runner/lib/index.ts
rossmcewan: Maybe with empty code we should rather throw an error.

packages/jobs/lib/runtime/lambda.adapter.ts
rossmcewan: I can put a safety net of 100mb for now.

packages/jobs/lib/runtime/lambda.adapter.ts
rossmcewan: Hmm. That is a good point. We would have to inject the lifecycle policy days into the function (it's currently 30). Will make that change.

packages/jobs/lib/runtime/lambda.adapter.ts
rossmcewan: I'm not sure what you mean by this?

packages/jobs/lib/runtime/lambda.adapter.ts
rossmcewan: We do have this in our traces. Not sure metrics will be meaningful yet as we only have a couple of examples where large payloads are even necessary.
I would prefer to add a follow up task for this.

TBonnin (approved):

packages/jobs/lib/runtime/lambda.adapter.ts
pfreixes: no problemo, just consider the 💅 as "nice to have" but feel free to either push back/skip/make it later

packages/lambda-runner/lib/index.ts
pfreixes: you are right 🤦

pfreixes (approved):

PR #5602, opened 2026-03-11
feat(webhooks): add webhook support for ms client credentials

hassan254-prog:

- add webhook support for ms client credentials

---

**Add Microsoft client-credentials webhook routing with allowlisted query params**

This PR introduces a new webhook routing handler for Microsoft client credentials, including handling `validationToken` subscription validation, filtering notifications by `clientState`, and forwarding only validated notifications. It also threads optional query parameters through the webhook controller, manager, and handler signature with a per-provider allowlist to preserve the default deny posture.

Provider metadata and validation schema are updated to support `webhook_allowed_query_params`, and the Microsoft OAuth2 client-credentials provider is configured to use the new routing script, allow user-defined secrets, and allowlist `validationToken`.

---

### DIFF
--- a/packages/server/lib/controllers/webhook/environmentUuid/postWebhook.ts
+++ b/packages/server/lib/controllers/webhook/environmentUuid/postWebhook.ts
@@ -3,7 +3,7 @@ import * as z from 'zod';
 
 import db from '@nangohq/database';
 import { logContextGetter } from '@nangohq/logs';
-import { accountService, configService, getPlan } from '@nangohq/shared';
+import { accountService, configService, getPlan, getProvider } from '@nangohq/shared';
 import { flagHasPlan, metrics, zodErrorToHTTP } from '@nangohq/utils';
 
 import { providerConfigKeySchema } from '../../../helpers/validation.js';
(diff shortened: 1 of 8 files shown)


packages/server/lib/controllers/webhook/environmentUuid/postWebhook.ts
```diff
@@ -13,6 +13,15 @@ import { routeWebhook } from '../../../webhook/webhook.manager.js';
...
 import type { DBPlan, PostPublicWebhook } from '@nangohq/types';
 
+/**
+ * Names of query parameters that are forwarded to webhook routing.
+ * Only these parameters are forwarded; all others are ignored for security reasons.
+ * By default, query parameters are not passed through: some providers send
+ * unverifiable parameters, which are ignored to maintain security,
+ * since they generally cannot be included in webhook signatures.
+ */
+const ALLOWED_WEBHOOK_QUERY_PARAMS = ['validationToken'];
```
agusayerza: IMO`ALLOWED_WEBHOOK_QUERY_PARAMS` shouldn't be a global decision for all webhook providers. I would make this a per-provider setting instead (e.g. `webhook_allowed_query_params: ['validationToken']` on the provider in `providers.yaml`, defaulting to `[]`)? That keeps the current default-deny security posture, but makes exceptions explicit per provider and avoids growing a global allowlist in the controller as we add more webhook integrations.

hassan254-prog: I have updated this with the latest [commit](https://github.com/NangoHQ/nango/pull/5602/commits/f35fb420985ba568261837c6f7d97b961b40ffef).

agusayerza (approved):

PR #5604, opened 2026-03-11
feat(cli): drop `nango.yaml` support in CLI commands

kaposke:
Removes `nango.yaml` ramifications and validates for zero-yaml in all relevant commands. Deletes now unused code (might have missed some spots, but I will likely find more as I continue this)

Also minor renaming of a couple functions.

---

It also routes all key CLI commands through zero‑yaml validation and compilation, simplifies dryrun to rely solely on zero‑yaml definitions, removes remaining legacy/hidden CLI workflows tied to nango.yaml, and tweaks initialization behavior to focus on zero‑yaml projects.

---


packages/cli/lib/services/dryrun.service.ts
```diff
@@ -669,18 +643,13 @@ export class DryRunService {
...
-                } else {
-                    const results = await scriptExports.default(nango);
-                    return { success: true, error: null, response: { output: results, nango } };
+                const syncPayload = scriptExports.default as unknown as nangoScript.CreateSyncResponse<any, any>;
+                if (syncPayload.type !== 'sync') {
+                    throw new Error('Incorrect script loaded for sync');
                 }
+
+                await syncPayload.exec(nango as any);
+                return { success: true, response: { output: undefined, nango }, error: null };
```
TBonnin, line 652: `output: undefined` ?

packages/cli/lib/services/verification.service.ts
```diff
@@ -2,114 +2,10 @@ import fs from 'fs';
...
-
-        if (missingFiles.length > 0) {
-            console.log(chalk.red(`The following functions are missing a corresponding .ts file: ${missingFiles.join(', ')}`));
-            throw new Error('Script missing .ts files');
-        }
-
-        return true;
-    }
-
     public async preCheck({
```
TBonnin: cam `preCheck` be made private now?

TBonnin (approved):

packages/cli/lib/services/dryrun.service.ts
kaposke: While it previously did
```ts
const results = await payload.exec(nango as any);
return { success: true, response: { output: results, nango }, error: null };
```
Results was of type `void` for syncs.

marcindobry (approved):

packages/cli/lib/index.ts
```diff
@@ -173,8 +167,8 @@ program
             }
         };
 
-        const check = await verificationService.preCheck({ fullPath: absolutePath, debug });
-        if (check.hasNangoYaml || check.isZeroYaml) {
+        const precheck = await verificationService.ensureZeroYaml({ fullPath: absolutePath, debug });
```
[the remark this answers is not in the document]
kaposke: Whoops, rolling back

PR #5607, opened 2026-03-11
feat: add variants_per_sync_max to plan

TBonnin:
so we can bump the number of allowed variants per customer. 
The commit also get rid of `has_sync_variants` that can now be controlled by setting `variants_per_sync_max` to 0 in order to disable the feature. Note: we recently agree to open variants to all customers not just paying ones (ie: all get 100/sync by default)
Migration to remove the has_sync_variants column will be done in followup PR

 

---

**Add `variants_per_sync_max` plan flag and enforce per-sync variant limits**

This PR introduces a new `variants_per_sync_max` plan flag, adds a database column with a default of 100, and uses it to cap variants per sync in the `postSyncVariant` controller. It also removes the legacy `has_sync_variants` checks and references, relying on `variants_per_sync_max` (including `0` to disable) instead.

---

### DIFF
--- a/packages/database/lib/migrations/20260311120000_plans_add_variants_per_sync_max.cjs
+++ b/packages/database/lib/migrations/20260311120000_plans_add_variants_per_sync_max.cjs
@@ -1,0 +1,13 @@
+exports.config = { transaction: true };
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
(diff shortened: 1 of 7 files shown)


kaposke (approved):
> (disabling variants) can now be controlled by setting `variants_per_sync_max` to 0 in order to disable the feature

Should it be 1 for the `base` variant, and 0 for unlimited?

PR #5614, opened 2026-03-12
fix(deploy): detect deleted syncs when deploying

hassan254-prog:
This PR fixes two issues.

1. Deploying a single integration did not remove deleted syncs/actions
When deploying a single integration with the flag `--integration-id`, the deploy process incorrectly treated it as `singleDeployMode`, which skipped reconciliation entirely. As a result, functions that were removed are not deleted.
singleDeployMode now only applies when deploying a single named function (--sync or --action), not when deploying an integration by ID.

2. Flow lookup mismatch between syncs and actions
Fixes a bug where the flow lookup did not account for flow type, causing deletions to be missed when a sync and action shared the same script name, but one of them was left in the config.

---

It also standardizes deploy scoping by replacing the boolean flag with a string-based deploy mode across CLI packaging, API validation, controllers, and shared services (with backward-compatible mapping), and scopes on-event script diffing by deploy mode, including an integration fallback when no scripts are provided, with accompanying test updates.

---


TBonnin (commented):
can you describer the problem? thank you

hassan254-prog:
> can you describer the problem? thank you

Updated the pr description, I was waiting for AI to finish reviewing first, It's a habit of mine when working on docs prs that it can easily handle, can't do that in draft mode 😅 .

packages/cli/lib/zeroYaml/deploy.ts
```diff
@@ -158,7 +158,9 @@ async function createPackage({
 
     const postData: CLIDeployFlowConfig[] = [];
     const onEventScriptsByProvider: OnEventScriptsByProvider[] | undefined = optionalActionName || optionalSyncName ? undefined : []; // only load on-event scripts if we're not deploying a single sync or action
-    const singleDeployMode = Boolean(optionalSyncName || optionalActionName || optionalIntegrationId);
+    const hasSingleScript = Boolean(optionalSyncName || optionalActionName);
+    const singleDeployMode = hasSingleScript;
+    const deployedProviderConfigKeys = optionalIntegrationId && !hasSingleScript ? [optionalIntegrationId] : undefined;
```
TBonnin: why an array if it is only a single `integrationId`?

packages/server/lib/controllers/sync/deploy/validation.ts
```diff
@@ -152,6 +152,7 @@ const commonValidation = z
         reconcile: z.boolean(),
         debug: z.boolean(),
         singleDeployMode: z.boolean().optional().default(false),
+        deployedProviderConfigKeys: z.array(z.string()).optional(),
```
TBonnin: what about replacing the `singleDeployMode: boolean` argument with a `mode: 'all' | 'single' | 'integration'?

packages/server/lib/controllers/sync/deploy/validation.ts
```diff
@@ -151,7 +151,8 @@ const commonValidation = z
         jsonSchema: jsonSchema.optional(),
         reconcile: z.boolean(),
         debug: z.boolean(),
-        singleDeployMode: z.boolean().optional().default(false),
+        deployMode: z.enum(['all', 'single', 'integration']).optional().default('all'),
```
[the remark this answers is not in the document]
hassan254-prog: We only deploy functions from the cli, and the necessary changes have been made to that too.

packages/server/lib/controllers/sync/deploy/validation.ts
hassan254-prog: Yes, I have updated this with the latest [commit](https://github.com/NangoHQ/nango/pull/5614/commits/8862042eb9630196d745f543c89db063244b8263).

packages/server/lib/controllers/sync/deploy/postConfirmation.ts
```diff
@@ -34,7 +34,8 @@ export const postDeployConfirmation = asyncWrapper<PostDeployConfirmation>(async
         flows: body.flowConfigs,
         performAction: false,
         debug: body.debug,
-        singleDeployMode: body.singleDeployMode,
```
TBonnin, line 37: we need to keep `singleDeployMode` and deprecate it for CLI that haven't updated yet

packages/cli/lib/zeroYaml/deploy.ts
hassan254-prog: Thanks for flagging that. Since it’s only a single integrationId, an array isn’t necessary. I’ve updated this in the latest [commit](https://github.com/NangoHQ/nango/pull/5614/commits/8862042eb9630196d745f543c89db063244b8263).

packages/cli/lib/zeroYaml/deploy.ts
```diff
@@ -158,7 +158,9 @@ async function createPackage({
 
     const postData: CLIDeployFlowConfig[] = [];
     const onEventScriptsByProvider: OnEventScriptsByProvider[] | undefined = optionalActionName || optionalSyncName ? undefined : []; // only load on-event scripts if we're not deploying a single sync or action
-    const singleDeployMode = Boolean(optionalSyncName || optionalActionName || optionalIntegrationId);
+    const hasSingleScript = Boolean(optionalSyncName || optionalActionName);
+    const deployMode: 'all' | 'single' | 'integration' = hasSingleScript ? 'single' : optionalIntegrationId ? 'integration' : 'all';
+    const deployedProviderConfigKey = optionalIntegrationId && !hasSingleScript ? optionalIntegrationId : undefined;
```
TBonnin: could the flows sent to deploy only contains the ones for the selected integrations so we don't have to track the integrationId separately?

packages/server/lib/controllers/sync/deploy/postConfirmation.ts
hassan254-prog: updated with the latest [commit](https://github.com/NangoHQ/nango/pull/5614/commits/982353b9a725d639869ad12e1515f608b3381055), the validation now handles the `singleDeployMode` → `deployMode` backcompat transform directly.

packages/cli/lib/zeroYaml/deploy.ts
hassan254-prog: I agree, updated this logic with the latest [commit](https://github.com/NangoHQ/nango/pull/5614/commits/982353b9a725d639869ad12e1515f608b3381055), no need for tracking them separately.

packages/shared/lib/services/on-event-scripts.service.ts
```diff
@@ -186,10 +186,16 @@ export const onEventScriptService = {
...
+        // Filter existing scripts to only include those from providers being deployed when not in all-providers deploy mode.
+        // For 'integration' mode, infer the scoped provider from the incoming scripts (or fall back to deployingProviders[0])
+        // so stale scripts are detected even when no scripts are incoming.
+        const integrationScopeKey = deployMode === 'integration' ? (onEventScriptsByProvider[0]?.providerConfigKey ?? deployingProviders[0]) : undefined;
+        const relevantExistingScripts =
+            deployMode === 'integration' && integrationScopeKey
+                ? existingScripts.filter((script) => script.providerConfigKey === integrationScopeKey)
+                : deployMode === 'single'
+                  ? existingScripts.filter((script) => deployingProviders.includes(script.providerConfigKey))
+                  : existingScripts;
```
TBonnin: a lookup map would be easier to read
```
const filters: Record<'integration' | 'single' | 'all', (s: Script) => boolean> = {
    integration: (s) => s.providerConfigKey === integrationScopeKey,
    single: (s) => deployingProviders.includes(s.providerConfigKey),
    all: () => true,
};

const relevant = existingScripts.filter(filters[deployMode]);
```

packages/shared/lib/services/sync/sync.service.ts
```diff
@@ -483,9 +483,16 @@ export const getAndReconcileDifferences = async ({
     const deletedActions: SlimAction[] = [];
     const deletedModels: string[] = [];
 
-    if (!singleDeployMode) {
+    if (deployMode !== 'single') {
+        const integrationScopeKey = deployMode === 'integration' ? flows[0]?.providerConfigKey : undefined;
```
TBonnin, line 487: could it be have a problem if flows has not entries? Then the new `if` below won't ever be entered and syncs not skipped

packages/server/lib/controllers/sync/deploy/postConfirmation.ts
```diff
@@ -48,7 +48,7 @@ export const postDeployConfirmation = asyncWrapper<PostDeployConfirmation>(async
         const diff = await onEventScriptService.diffChanges({
             environmentId,
             onEventScriptsByProvider: body.onEventScriptsByProvider,
-            singleDeployMode: body.singleDeployMode || false,
+            deployMode: body.deployMode ?? 'all',
```
TBonnin, line 51: do we need the fallback?

TBonnin (approved):

packages/shared/lib/services/sync/sync.service.ts
hassan254-prog: The cli validates this before making any call, if the integration has no flows, it fails with `No functions to deploy` when deploying, so flows will never be empty when `deployMode === 'integration'`.

packages/server/lib/controllers/sync/deploy/postConfirmation.ts
hassan254-prog: Yes, its required because `exactOptionalPropertyTypes: true` prevents passing undefined to an optional property.

packages/shared/lib/services/on-event-scripts.service.ts
hassan254-prog: updated with the latest [commit](https://github.com/NangoHQ/nango/pull/5614/commits/3b0756376b267eeec9353e3882ba8851c7bf0348).

packages/shared/lib/services/sync/sync.service.ts
TBonnin: ok. given how simple it is to check for empty flows, I would still have a guard server side. up to you though

hassan254-prog: Makes sense to add the guard, updated with the latest commit.

PR #5624, opened 2026-03-12
docs: add stop listening on connection deletion and real-time sync concept for gmail and google calendar webhooks

agusayerza:
docs: add stop listening on connection deletion and real-time sync concept for gmail and google calendar webhooks

 

---

**Update Gmail and Google Calendar webhook documentation with lifecycle and real-time sync guidance**

This PR revises the Gmail and Google Calendar webhook docs to clarify setup wording, add lifecycle cleanup steps, and document real-time sync options. It introduces examples for automated channel/watch creation, renewal flows with metadata storage, and pre-connection deletion cleanup, plus notes on connection matching and webhook-driven sync behavior.

---

### DIFF
--- a/docs/api-integrations/google-calendar/webhooks.mdx
+++ b/docs/api-integrations/google-calendar/webhooks.mdx
@@ -27,6 +27,55 @@ Copy the webhook URL from your Google Calendar integration page in the Nango das
 
 Create a notification channel for the calendar you want to watch by sending a `POST` request to the appropriate `watch` endpoint. See the [API reference](https://developers.google.com/workspace/calendar/api/guides/push#create-notification-channels) for the exact URI.
 
+You can automate creating the channel for new connections with a [post-connection-creation script](/implementation-guides/use-cases/implement-event-handler):
+
+```typescript
+import { createOnEvent, ProxyConfiguration } from 'nango';
+import { randomUUID } from 'crypto';
+import z from 'zod';
+
+export default createOnEvent({
+    event: 'post-connection-creation',
+    description: 'Create a Google Calendar notification channel for events on the primary calendar',
+    metadata: z.object({
+        googleCalendarChannelId: z.string().optional(),
+        googleCalendarResourceId: z.string().optional(),
+    }),
+    exec: async (nango) => {
+        const calResponse = await nango.get({
+            baseUrlOverride: 'https://www.googleapis.com/calendar',
+            endpoint: '/v3/users/me/calendarList/primary' 
+        });
+
+        const calendarEmail = calResponse.data.id; 
+        const channelId = randomUUID();
+        const expiration = Date.now() + 7 * 24 * 60 * 60 * 1000; // 7 days in ms
+        const webhookUrl = await nango.getWebhookURL();
+
+        const config: ProxyConfiguration = {
+            baseUrlOverride: 'https://www.googleapis.com/calendar',
+            endpoint: `/v3/calendars/${encodeURIComponent(calendarEmail)}/events/watch`,
+            data: {
+                id: channelId,
+                type: 'webhook',
+                // Webhook URL can be found on https://app.nango.dev/dev/integrations/google-calendar (your integration settings page)
+                address: webhookUrl,
+                expiration: expiration,
+            },
+        };
+
+        const response = await nango.post(config);
+
+        // Store channel info so it can be used to stop notifications later (step 4)
+        await nango.updateMetadata({
+            googleCalendarChannelId: channelId,
+            googleCalendarResourceId: response.data.resourceId,
+        });
+    }
+});
+```
+
+You can also do this manually:
 Example: watch the **events** collection on a calendar:
 
 ```bash
@@ -36,7 +85,7 @@ curl -X POST "https://www.googleapis.com/calendar/v3/calendars/<CALENDAR_ID>/eve
   -d '{
     "id": "01234567-89ab-cdef-0123456789ab",
     "type": "web_hook",
-    "address": "https://api.nango.dev/webhook/...",
+    "address": "<REPLACE_WITH_NANGO_WEBHOOK_URL>",
     "expiration": 1426325213000
   }'
 ```
@@ -49,36 +98,6 @@ Replace:
 - **id** — A unique string (e.g. UUID) identifying this channel; max 64 characters. It is echoed in `X-Goog-Channel-ID` on every notification.
 - **expiration** — (Optional) Unix timestamp in **milliseconds** when the channel should stop sending notifications. If omitted, Google applies a default; channels must be renewed before they expire.
 
-You can automate creating the channel for new connections with a [post-connection-creation script](/implementation-guides/use-cases/implement-event-handler):
-
-```typescript
-import { createOnEvent, ProxyConfiguration } from 'nango';
-import { randomUUID } from 'crypto';
-
-export default createOnEvent({
-    event: 'post-connection-creation',
-    description: 'Create a Google Calendar notification channel for events on the primary calendar',
-    exec: async (nango) => {
-        const channelId = randomUUID();
-        const expiration = Date.now() + 7 * 24 * 60 * 60 * 1000; // 7 days in ms
-
-        const config: ProxyConfiguration = {
-            baseUrlOverride: 'https://www.googleapis.com/calendar',
-            endpoint: '/v3/calendars/primary/events/watch',
-            data: {
-                id: channelId,
-                type: 'web_hook',
-                address: 'https://api.nango.dev/webhook/...',
-                expiration,
-            },
-        };
-
-        const response = await nango.post(config);
-        await nango.log(response.data);
-    }
-});
-```
-
 ### 3. Renew the notification channel
 
 Google does **not** renew channels automatically. When a channel is close to its expiration, you must create a new channel by calling the `watch` endpoint again with a **new** unique `id`. See [Renew notification channels](https://developers.google.com/workspace/calendar/api/guides/push#renew-notification-channels).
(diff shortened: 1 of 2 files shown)


docs/api-integrations/google-calendar/webhooks.mdx
```diff
@@ -191,11 +197,49 @@ curl -X PATCH "https://api.nango.dev/connection/metadata" \
 To do this in bulk, iterate over the [list connections](/reference/api/connection/list) response and [update the metadata](/reference/api/connection/update-metadata) for each one.
 
 <Warning>
-If Nango cannot match the incoming webhook to a connection (because neither `connection_config` nor `metadata` contains the email address), the webhook will still be received but won't include a `connectionId` in the forwarded payload.
+If Nango cannot match the incoming webhook to a connection (because it can't find a matching hash/email in `connection_config` or `metadata`), the webhook will still be received but won't include a `connectionId` in the forwarded payload.
 </Warning>
 
 ## Stop notifications
```
rguldener: I think stop notification should be part of the setup flow (one of the numbered steps), or do you see a scenario where stopping webhooks on connection deletion is not desirable?

We had issues in the past with people only cursorily reading these guides and missing important steps that were not very clearly part of the required setup.

docs/api-integrations/google-calendar/webhooks.mdx
```diff
@@ -191,11 +197,49 @@ curl -X PATCH "https://api.nango.dev/connection/metadata" \
...
+
+```typescript
+import { createOnEvent, ProxyConfiguration } from 'nango';
+
+export default createOnEvent({
+    event: 'pre-connection-deletion',
+    description: 'Stop Google Calendar notification channel before connection deletion',
+    exec: async (nango) => {
+        const metadata = await nango.getMetadata();
+        const channelId = metadata['googleCalendarChannelId'];
```
rguldener: I don't see this information being stored to connection metadata anywhere in the setup guide. If it's required here to unsubscribe, storing it should be part of the setup IMO.

docs/api-integrations/google-calendar/webhooks.mdx
```diff
@@ -191,11 +197,49 @@ curl -X PATCH "https://api.nango.dev/connection/metadata" \
...
+            await nango.post(config);
+        } catch (err) {
+            // Avoid blocking connection deletion if stop fails.
+            await nango.log(`Failed to stop Calendar channel: ${String(err)}`);
+        }
+    }
+});
+```
+
 To stop receiving notifications before the channel expires, call the [channels.stop](https://developers.google.com/workspace/calendar/api/guides/push#stop-notifications) method with the channel `id` and `resourceId` returned when you created the channel:
```
rguldener: 1. Is this paragraph still necessary?
2. if it is, currently it looks like `channelId` and `resourceId` are not being stored anywhere, so presumably they wouldn't be available?

(nango logs clear after 14 days)

docs/api-integrations/google-mail/webhooks.mdx
rguldener: Nit: Instead of "Gmail push notification" can we call it "Gmail webhooks"?
Most devs working on the web expect webhooks, IMO this would help with discoverability and understanding.

rguldener, line 40: Most Nango customers setting up gmail don't use google cloud a lot and probably don't have the gcloud CLI setup & authenticated.
If there is an easy alternative (ideally a resource from Google we can link) would be cool to have an alternative. Otherwise, totally ok to leave as is, IMO.

rguldener: Where is this `metadata['topicName']` being set?
This is a new connection, it's highly unlikely any metadata is present when this function runs. Or am I missing something?

If I understand correct, the topic is a user-chosen value in the pub-sub setup. Probably they should just hard-code their value here?

docs/api-integrations/google-calendar/webhooks.mdx
rguldener: Looks like the gmail guide already does this

docs/api-integrations/google-calendar/webhooks.mdx
rguldener: I don't think this is how we should use the endpoints config. @agusayerza can explain you what its for

rguldener (changes requested):
Found a number of issues (also in the previous version), but let me know if I misunderstood something!

@rbwest Please also thoroughly review

docs/api-integrations/google-calendar/webhooks.mdx
```diff
@@ -27,6 +27,54 @@ Copy the webhook URL from your Google Calendar integration page in the Nango das
...
+        const calendarEmail = calResponse.data.id; 
+        const channelId = randomUUID();
+        const webhookUrl = await nango.getWebhookURL();
+
+        const config: ProxyConfiguration = {
+            baseUrlOverride: 'https://www.googleapis.com/calendar',
+            endpoint: `/v3/calendars/${encodeURIComponent(calendarEmail)}/events/watch`,
+            data: {
+                id: channelId,
+                type: 'webhook',
```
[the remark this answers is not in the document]
agusayerza: Both `webhook` and `web_hook` work

docs/api-integrations/google-calendar/webhooks.mdx
agusayerza: Added as part of the setup flow

docs/api-integrations/google-calendar/webhooks.mdx
agusayerza: Missed

docs/api-integrations/google-mail/webhooks.mdx
agusayerza: Added instructions to do it from the cloud dashboard instead. They need to be on GCP to be able to have Client Credentials for the API, but you are right more often than not they won't have the cli

agusayerza: Removed the nango.getMetadata() call and replaced it with a hardcoded placeholder + comment telling users to set their own topic name from step 1. Applied the same fix in the renew sync script.

docs/api-integrations/google-calendar/webhooks.mdx
agusayerza: Fixed. Changed the endpoints path from `/v3/calendars/primary/events/watch` (which looked like the external Google API path) to `/google-calendar/watch-renewal`. Applied the same fix in the gmail renew sync (`/gmail/watch-renewal`).

docs/api-integrations/google-calendar/webhooks.mdx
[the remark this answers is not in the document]
agusayerza: https://developers.google.com/workspace/calendar/api/v3/reference/events/watch#request-body

docs/api-integrations/google-calendar/webhooks.mdx
rbwest: Suppose this should say webhooks instead of push notifications to match the updated language on the Gmail side per Robin's note.

```suggestion
description: 'Learn how to set up Google Calendar webhooks with notification channels and Nango'
---

This guide shows you how to receive real-time Google Calendar webhooks in Nango using [notification channels](https://developers.google.com/workspace/calendar/api/guides/push#create-notification-channels). Unlike Gmail, Google Calendar does **not** require Google Cloud Pub/Sub—you configure a webhook URL directly with the Calendar API.

## How it works

The Google Calendar API uses notification channels to deliver webhooks. The flow is:
```

docs/api-integrations/google-calendar/webhooks.mdx
```diff
@@ -36,7 +85,7 @@ curl -X POST "https://www.googleapis.com/calendar/v3/calendars/<CALENDAR_ID>/eve
   -d '{
     "id": "01234567-89ab-cdef-0123456789ab",
     "type": "web_hook",
-    "address": "https://api.nango.dev/webhook/...",
+    "address": "<REPLACE_WITH_NANGO_WEBHOOK_URL>",
     "expiration": 1426325213000
```
rbwest: Should probably make it more clear this needs to be replaced as well.

docs/api-integrations/google-calendar/webhooks.mdx
```diff
@@ -49,75 +98,102 @@ Replace:
...
+export default createSync({
+    description: 'Renew the Google Calendar events watch channel before it expires',
+    models: {},
+    metadata: z.object({
+        googleCalendarChannelId: z.string().optional(),
+        googleCalendarResourceId: z.string().optional(),
+    }),
+    endpoints: [{ method: 'GET', path: '/google-calendar/watch-renewal' }],
+    syncType: 'full',
+    frequency: '1d',
```
rbwest, line 121: Should we bump this so people don't do it daily by default copy/pasting? Default watch lasts 7 days from my cursory inspection, so maybe every 5-6 days?

docs/api-integrations/google-mail/webhooks.mdx
```diff
@@ -93,26 +103,25 @@ Google [recommends](https://developers.google.com/workspace/gmail/api/guides/pus
...
     description: 'Calls the watch endpoint daily to keep a Gmail webhook subscription active',
-    models: { Empty },
+    models: { Empty: z.object({}) },
     endpoints: [{
         method: 'GET',
-        path: '/watch',
+        path: '/gmail/watch-renewal',
     }],
     syncType: 'full',
     frequency: '1d',
```
rbwest, line 116: Same notes as above on frequency: https://github.com/NangoHQ/nango/pull/5624/changes#r2933286448

rbwest (changes requested):

docs/api-integrations/google-calendar/webhooks.mdx
agusayerza: Say that it was 5d, but If we fail for some reason, then the watch will expire before we re-attempt. At the same time, renewing every day without stopping the old watch will make us have to handle multiple triggers for the same event and its a little bit more taxing on the customer bill (minimal though). Once again I think it comes down to what the customers what, so maybe adding a note about it?
Also, now that I write about it, I think we should actually stop the old watch after successful renewal

docs/api-integrations/google-calendar/webhooks.mdx
agusayerza: Good catch

docs/api-integrations/google-calendar/webhooks.mdx
rbwest: Makes sense. Resolving!

docs/api-integrations/google-mail/webhooks.mdx
rbwest: See https://github.com/NangoHQ/nango/pull/5624#discussion_r2933337804

rbwest (approved):

rguldener (approved):
Only did a quick cursory check, looks good now!

PR #5627, opened 2026-03-12
refactor(cli): light refac on build and deploy 

kaposke:
Some light renaming and comment that IMO makes this a bit more understandable. Ideally I would rename `NangoYamlParsed` and so on, but it still has a lot has dependents. I also went a bit further but went back on some changes. Trying to reduce risk of breaking something while still improving this a bit.

Plan on also moving some files around, but that would make the PR review confusing. Will hold off for now.

---

**Rename definition parsing helper and clarify CLI packaging output**

This PR performs light refactoring across the CLI to rename `buildDefinitions` to `parseIntegrationDefinitions`, updating all call sites in deploy, compile, dry-run, test generation, and command routing. It also improves naming for deploy packaging by renaming `createPackage` to `createDeployConfirmationPackage` and adds clarifying comments.

Additionally, `generateAdditionalExports` now logs the `nango.json` export path when debug is enabled, and the generated `schema.ts` export includes a deprecation notice in its header comment.

---

### DIFF
--- a/packages/cli/lib/services/model.service.ts
+++ b/packages/cli/lib/services/model.service.ts
@@ -217,7 +217,17 @@ export function generateAdditionalExports({ fullPath, parsed, debug }: { fullPat
         printDebug(`Generated export ${pathJSON}`);
     }
 
-    fs.writeFileSync(path.join(exportPath, 'nango.json'), JSON.stringify(parsed.integrations, null, 2));
+    const nangoJsonPath = path.join(exportPath, 'nango.json');
+    fs.writeFileSync(nangoJsonPath, JSON.stringify(parsed.integrations, null, 2));
+    if (debug) {
+        printDebug(`Generated export ${nangoJsonPath}`);
+    }
+
+    console.log(
+        chalk.yellow(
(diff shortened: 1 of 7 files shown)


packages/cli/lib/services/model.service.ts
```diff
@@ -217,7 +217,17 @@ export function generateAdditionalExports({ fullPath, parsed, debug }: { fullPat
...
+    const nangoJsonPath = path.join(exportPath, 'nango.json');
+    fs.writeFileSync(nangoJsonPath, JSON.stringify(parsed.integrations, null, 2));
+    if (debug) {
+        printDebug(`Generated export ${nangoJsonPath}`);
+    }
+
+    console.log(
+        chalk.yellow(
+            'The generated .nango/schema.ts and .nango/schema.json are deprecated and will stop being generated in future versions. For access to function types, please export types directly from your code. You can leverage `zod.infer` to generate types from your zod schemas. See the official zod documentation: https://zod.dev/basics?id=inferring-types'
+        )
```
kaposke: This is still up for debate (and not really a refac, but a small feat). I was suggested to have a shorter message and link to a migration guide, but tbh I feel like this message basically conveys everything I would want on a guide (it's that simple).

TBonnin: that might be a bit noisy to show it on every compile/deploy. I don't have a better idea though

TBonnin (approved):

PR #5629, opened 2026-03-13
feat(database): add role column to users/invitations and is_production to environments

pfreixes:
Add schema migrations for the permissions system (NAN-2184):
- Add `role` column (VARCHAR(50), default 'administrator') to _nango_users
- Add `role` column (VARCHAR(50), default 'administrator') to _nango_invited_users
- Add `is_production` column (BOOLEAN, default false) to _nango_environments

We will doing first this migration which should be a `noop` from the point of view of our application, once this one is in place we will follow up with the following changes

- [PR](https://github.com/NangoHQ/nango/pull/5631) for add logic in our app for tagging all new created `prod` environments as `is_production` as `True`
- [PR](https://github.com/NangoHQ/nango/pull/5632) for adding a new migration for updating all existing `prod` environments as `is_production`  as `True`

This will allow us to not have any dat race and keeping the all proper `prod` environments tagged as they should be

After that we will proceed with more PRs for supporting the RBAC system

 

---

It also updates the environment service integration test expectations to include the new default value for `is_production`.

---

### DIFF
--- a/packages/database/lib/migrations/20260313120000_add_role_to_users.cjs
+++ b/packages/database/lib/migrations/20260313120000_add_role_to_users.cjs
@@ -1,0 +1,21 @@
+exports.config = { transaction: false };
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`
+        ALTER TABLE _nango_users
+        ADD COLUMN IF NOT EXISTS role VARCHAR(50) NOT NULL DEFAULT 'administrator'
+    `);
+};
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.down = async function (knex) {
+    await knex.raw(`
+        ALTER TABLE _nango_users
+        DROP COLUMN IF EXISTS role
+    `);
+};
--- a/packages/database/lib/migrations/20260313120000_add_role_to_users.cjs
+++ b/packages/database/lib/migrations/20260313120000_add_role_to_users.cjs
@@ -8,14 +8,11 @@ exports.up = async function (knex) {
         ALTER TABLE _nango_users
         ADD COLUMN IF NOT EXISTS role VARCHAR(50) NOT NULL DEFAULT 'administrator'
     `);
-};
 
-/**
- * @param {import('knex').Knex} knex
- */
-exports.down = async function (knex) {
     await knex.raw(`
         ALTER TABLE _nango_users
-        DROP COLUMN IF EXISTS role
+        ADD CONSTRAINT check_users_role CHECK (role IN ('administrator', 'production_support', 'development_full_access'))
     `);
 };
+
+exports.down = function () {};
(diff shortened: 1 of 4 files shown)


packages/database/lib/migrations/20260313120000_add_role_to_users.cjs
```diff
@@ -0,0 +1,21 @@
+exports.config = { transaction: false };
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`
+        ALTER TABLE _nango_users
+        ADD COLUMN IF NOT EXISTS role VARCHAR(50) NOT NULL DEFAULT 'administrator'
```
marcindobry, line 9: nit: Would it be safer for the default to be the least powerful role?

marcindobry (approved):

packages/database/lib/migrations/20260313120000_add_role_to_users.cjs
pfreixes: The idea is that we do not break the current product, RBAC will be used only for a small subset of users, so most of the current ones will keep using and expecting same level of privilegies.

Als this allows us to deploy all changes in separated PRs, and having for now all users getting the same level of permissions will break nothing.

The latest PR will allow the users to either change the role of an existing user or when inviting set a different role, which by default will be for now `administrator`

From there and at some point we can change this default value

packages/shared/lib/services/environment.service.integration.test.ts
```diff
@@ -29,6 +29,7 @@ describe('Environment service', () => {
             hmac_enabled: false,
             hmac_key: null,
             id: expect.any(Number),
+            is_production: false,
```
pfreixes, line 32: Seems that our API returns all fields from the environment, but expecting that our UI will just not make any usage of this field.

Once we have the three PRs merged that were pointed into the description of this PR, this field the `is_production` will be totally trustable (but not yet used in anyway by our frontend or backend)

/cc @TBonnin

packages/database/lib/migrations/20260313120000_add_role_to_users.cjs
```diff
@@ -0,0 +1,21 @@
...
+};
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.down = async function (knex) {
+    await knex.raw(`
+        ALTER TABLE _nango_users
+        DROP COLUMN IF EXISTS role
+    `);
```
TBonnin, line 11: we don't have any migration automatic revert process so we gave up pretending and stop writing down function

TBonnin (approved):
we are going all-in with application layer validation (no db level constraints, FK references). Are we fully ok with that?

packages/database/lib/migrations/20260313120000_add_role_to_users.cjs
pfreixes: Ill remove it!

pfreixes:
> we are going all-in with application layer validation (no db level constraints, FK references). Are we fully ok with that?

Good point, Ill go for at least adding for now a check which only allows the three allowed roles, so any bug in our application does not leave our DB in an inconsistency state.

PR #5634, opened 2026-03-13
fix(utils): extract provider error fields from response data directly

hassan254-prog:

- Widen this to extract provider error fields directly from the response data instead of reading them from the error object.

---

**Adjust `stringifyError` provider payload extraction to use `response.data` and expand test coverage**

This PR updates `stringifyError` to extract whitelisted provider error fields directly from `response.data`, including support for string responses, top-level fields, nested `responseData.error` objects, and precedence where top-level fields override nested ones. It also preserves a Boom payload fallback when no axios data is present and filters out object-valued fields unless `null`.

Tests in `packages/utils/lib/errors.unit.test.ts` are expanded to cover string payloads, top-level and nested extraction, override behavior, invalid response shapes, and axios-vs-Boom priority.

---


packages/utils/lib/errors.ts
```diff
@@ -37,16 +37,19 @@ export function stringifyError(err: unknown, opts?: { pretty?: boolean; stack?:
...
-            if (responseData.error && typeof responseData.error === 'object') {
+            if (typeof responseData === 'object' && responseData !== null) {
                 const filteredError: Record<string, unknown> = {};
                 for (const field of PROVIDER_ERROR_MESSAGE_FIELDS) {
-                    if (field in responseData.error) {
-                        filteredError[field] = responseData.error[field];
+                    if (!(field in responseData)) {
+                        continue;
+                    }
+                    const value = (responseData as Record<string, unknown>)[field];
```
TBonnin, line 65: we are changing the semantic of the logic, not looking at `responseData.error` anymore but `responseData`. Anything in `.error` will be skipped since in the line below object value are filtered out. 
Are we not breaking current behavior?

packages/utils/lib/errors.ts
```diff
@@ -37,16 +37,19 @@ export function stringifyError(err: unknown, opts?: { pretty?: boolean; stack?:
...
+                    if (typeof value !== 'object' || value === null) {
```
TBonnin, line 66: why `value === null`?

packages/utils/lib/errors.ts
hassan254-prog: Yes, we need to extract the error from the response and not specifically from the error object. Maybe we can perform an object traversal if an error object exists, or return the entire object if the field is white labelled. wdyt?

packages/utils/lib/errors.ts
hassan254-prog: without `value === null`, `null` values would be incorrectly treated as objects and filtered out, but we still need to preserve them when they are present.

packages/utils/lib/errors.ts
TBonnin: I don't fully understand the problem we are trying to solve. is `.error` not relevant anymore? Do we still need to stringified the content of `.error`? Why do we now need to stringify responseData directly and not `.error` anymore?

hassan254-prog: The original logic was written to look inside `responseData.error` as a nested object. The change widens the scope, instead of only extracting whitelisted fields from within `.error`, we now extract them from the entire `responseData ` directly. `.error` is still whitelisted and still gets extracted, just treated as a string field alongside the others rather than a container to dig into. We can still stringify whitelisted fields even if they are objects; open question though:- do we want to filter the fields inside a nested whitelisted object, or just return the entire object as-is?

packages/utils/lib/errors.ts
TBonnin: is it useful to keep null field though?

packages/utils/lib/errors.ts
TBonnin: This PR is introducing a breaking change

Before this change we had
```
response: {
   data: {
      error: { message: 'Nested error message', code: 401 },
      error_description: 'Top-level description'
    }
}
// used to become
{
   message: 'Nested error message'
}
// now it would be
{
   error_description: 'Top-level description''
}
```

What is motivating this change? Is it because of some API not wrapping error details in a error field? 
It would be good to add more details about the 'why' this PR exists. Otherwise I can only comment on the code and from what I see it introduce breaking changes and error message wrapped in an error field will be discarded.

packages/utils/lib/errors.ts
hassan254-prog: Yes, if its present in the whitelisted fields, we should still keep it.

packages/utils/lib/errors.ts
hassan254-prog: some providers don’t wrap their error details inside the `.error` object. Instead, they include the details directly in the response body. We need to widen our filter to inspect the entire response data, not just the `.error` object, so that all error messages which are white listed are captured correctly.

TBonnin: Ok but we are not gonna capture the fields for all the providers that are wrapping the error in an error object anymore. I personally don't think this is ok. Feel free to convince me otherwise.

hassan254-prog: Updated the logic to preserve the existing error handling, ensuring no breaking changes are introduced with the latest [commit](https://github.com/NangoHQ/nango/pull/5634/commits/9bf1699fe0f57bc9a03111794e02d1a0806ab48c), but now top-level primitive fields override nested `error.*` fields if present.

TBonnin (approved):

PR #5636, opened 2026-03-13
feat(webflow-sync): add force update flag and monthly scheduled run

hassan254-prog:

- Add a force update flag and a monthly scheduled run to ensure any providers added in the past but missing from the docs are listed.

---

**Add manual force update input and monthly schedule to Webflow sync**

This PR updates the Webflow sync workflow to support a manual `force_update` input and a monthly scheduled run, enabling periodic full refreshes. It also wires a `FORCE_UPDATE` environment variable into `scripts/webflow-api-sync.ts` to bypass diff checks when forced.

---

### DIFF
--- a/.github/workflows/webflow-sync.yaml
+++ b/.github/workflows/webflow-sync.yaml
@@ -9,6 +9,14 @@ on:
             - docs/integrations/all/*.md
             - packages/shared/flows.yaml
     workflow_dispatch:
+        inputs:
+            force_update:
+                description: 'Force update all providers regardless of changes'
+                required: false
+                default: 'false'
+                type: boolean
+    schedule:
+        - cron: '0 0 1 * *'
 
 concurrency:
     group: webflow-sync
@@ -31,3 +39,4 @@ jobs:
             - run: npx tsx scripts/webflow-api-sync.ts
               env:
                   WEBFLOW_CMS_API_TOKEN: ${{ secrets.WEBFLOW_CMS_API_TOKEN }}
+                  FORCE_UPDATE: ${{ github.event_name == 'schedule' || inputs.force_update == 'true' }}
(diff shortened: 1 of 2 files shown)


.github/workflows/webflow-sync.yaml
```diff
@@ -9,6 +9,14 @@ on:
...
             - packages/shared/flows.yaml
     workflow_dispatch:
+        inputs:
+            force_update:
+                description: 'Force update all providers regardless of changes'
+                required: false
+                default: 'false'
+                type: boolean
+    schedule:
+        - cron: '0 0 1 * *'
```
TBonnin, line 19: isn't it not running on master push? Why the need for a cron job?

hassan254-prog: Yes, it does, but some integrations are missing, possibly due to silent action failures when pushing to master that don’t block the merge. The cron job ensures that no integrations are missed, even if failures occur during publishing.

TBonnin: sounds like a band aid. Can we confirm there is indeed silent actions failures and fix those?

hassan254-prog: Yes, [this](https://github.com/NangoHQ/nango/actions/runs/20833599951/job/59853533207) had been silently failing for some time, which caused the list in master to differ from the integration list in webflow. Even after the [fix](https://github.com/NangoHQ/nango/pull/5192) was pushed, the missing integrations were never updated, because we only update the one present in the pr.

TBonnin: still not a huge fan of not fixing the underlying issue but up to you

TBonnin (approved):

PR #5644, opened 2026-03-16
feat(server): NAN-2184 role-based authorization system

pfreixes:
Introduce a server-side authorization layer gated by FLAG_AUTH_ROLES_ENABLED (default: false). When disabled, all permissions are permissive — no behavioral change. When enabled, three roles are enforced: administrator (full access), production_support (read prod, deny writes), and development_full_access (full non-prod, no prod access).

We will enable the `FLAG_AUTH_ROLES_ENABLED` in a different PR, which should have no consequences since all users - the old ones an new ones - are administartors

Check this [one](https://linear.app/nango/issue/NAN-2184/rbac#comment-be303bc8) for the overall plan.

 

---

It also adds a structured permission evaluator with route-level enforcement and exposes computed permissions to the UI, propagates roles through user creation and invitation flows, and updates environment metadata and production credential/secret redaction behavior, along with related configuration and test updates.

---


packages/server/lib/authz/middleware.ts
```diff
@@ -0,0 +1,59 @@
...
+    const allowed = authorize(method, routePath, role, res.locals);
+
+    if (!allowed) {
+        res.status(403).json({ error: { code: 'forbidden', message: 'You do not have permission to perform this action' } });
+        return;
+    }
+
+    // Populate authz locals for Category 3 (service-layer) enforcement
+    const isProduction = res.locals['environment']?.is_production ?? false;
+    res.locals['authz'] = {
```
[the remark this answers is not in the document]
pfreixes: Fixed in this [commit](https://github.com/NangoHQ/nango/pull/5644/changes/960055a1710841d00e0f84b50fb22073aa8c3c92)

packages/shared/lib/services/connection.service.ts
```diff
@@ -460,7 +462,13 @@ class ConnectionService {
             return Err('failed_to_fetch_connection');
         }
 
-        return Ok({ connection: encryptionManager.decryptConnection(result.connection), end_user: result.end_user });
+        // decryptConnection is always needed: it also converts date strings to Date objects.
+        const connection = encryptionManager.decryptConnection(result.connection);
+        if (!includeCredentials) {
+            connection.credentials = {};
```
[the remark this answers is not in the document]
pfreixes: Fixed in this [commit](https://github.com/NangoHQ/nango/pull/5644/changes/960055a1710841d00e0f84b50fb22073aa8c3c92)

packages/server/lib/controllers/v1/user/getUser.ts
```diff
@@ -12,7 +13,8 @@ export const getUser = asyncWrapper<GetUser, never>((req, res) => {
         return;
     }
 
+    const user = res.locals['user'] as DBUser;
     res.status(200).send({
-        data: userToAPI(res.locals['user'])
+        data: { ...userToAPI(user), role: user.role, permissions: buildPermissions(user.role) }
```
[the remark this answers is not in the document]
pfreixes: When the feature flag is disabled we just return all permissions for the UI as `true`, see [this](https://github.com/NangoHQ/nango/pull/5644/changes/BASE..960055a1710841d00e0f84b50fb22073aa8c3c92#diff-70b788bea3be3e0a261b6e986b731a82355c64e7dc9213c6030c80e729e86ba7R36)

Also the users are always created with one role (by default administrator) since we do not allow a null value

packages/server/lib/authz/authorize.ts
```diff
@@ -0,0 +1,15 @@
+import { evaluator } from './evaluator.js';
+import { resolveRoute } from './resolvers.js';
+
+import type { RequestLocals } from '../utils/express.js';
+import type { Role } from '@nangohq/types';
+
+export function authorize(method: string, path: string, role: Role, locals: RequestLocals): boolean {
+    if (role === 'administrator') return true;
```
rossmcewan: Shouldn't this be left up to the evaluator?

packages/server/lib/authz/types.ts
```diff
@@ -0,0 +1,11 @@
+import type { Role } from '@nangohq/types';
+
+export interface Permission {
+    action: string; // 'create' | 'delete' | 'write' | 'read' | '*'
+    resource: string; // 'environment' | 'integration' | 'team' | '*'
+    isProduction: boolean | null; // true = production env, false = non-production, null = not environment-scoped
+}
+
+export interface PermissionEvaluator {
+    evaluate(subject: { role: Role }, permission: Permission): boolean;
```
rossmcewan: Should this not return a `Promise<boolean>` or `Promise<Result<boolean>>` to support async behavior - i.e. future-proof the contract.
This has a downstream affect on the callers that would also need to be async (i.e. middleware, etc.).

packages/server/lib/controllers/v1/account/managed/getCallback.ts
```diff
@@ -117,7 +117,8 @@ export const getManagedCallback = asyncWrapper<GetManagedCallback>(async (req, r
             email: authorizedUser.email,
             name,
             account_id: account.id,
-            email_verified: true
+            email_verified: true,
+            role: invitation ? invitation.role : 'administrator'
```
rossmcewan: Maybe default role could be an env var

packages/server/lib/controllers/v1/account/signup.ts
```diff
@@ -109,7 +111,8 @@ export const signup = asyncWrapper<PostSignup>(async (req, res) => {
         hashed_password: hashedPassword,
         salt,
         account_id: account.id,
-        email_verified: token ? true : false
+        email_verified: token ? true : false,
+        role: token ? invitationRole : 'administrator'
```
rossmcewan: Same here. Use env var for default role.

packages/server/lib/controllers/v1/environment/patchEnvironment.ts
```diff
@@ -50,6 +51,20 @@ export const patchEnvironment = asyncWrapper<PatchEnvironment>(async (req, res)
 
         data.name = body.name;
     }
+    if (typeof body.is_production !== 'undefined') {
```
rossmcewan, line 56: Why the `typeof`? Would `body.is_production !== undefined` also work?

packages/server/lib/authz/evaluator.ts
```diff
@@ -0,0 +1,34 @@
...
+function matchesAction(rule: string, actual: string): boolean {
+    return rule === '*' || rule === actual;
+}
+
+function matchesResource(rule: string, actual: string): boolean {
+    return rule === '*' || rule === actual;
+}
+
+function matchesProduction(rule: boolean | null, actual: boolean | null): boolean {
+    if (rule === null && actual === null) return true; // both non-environment-scoped
```
TBonnin: is this condition not covered by `return rule === actual` below

packages/server/lib/authz/middleware.ts
```diff
@@ -0,0 +1,62 @@
...
+            canReadCredentials: true,
+            canReadProdSecrets: true,
+            canAccessProdEnvironments: true,
+            canToggleIsProduction: true
+        };
+        next();
+        return;
+    }
+
+    const role: Role = user.role ?? 'administrator';
```
TBonnin: is the administrator fallback temporary? can it be dangerous?

packages/server/lib/controllers/v1/environment/patchEnvironment.ts
pfreixes: Just followed same pattern used fore other keys, see above.

the `body.is_production !== undefined` would throw an exception if body would be null while the `typeof` will not if Im not mistaken, we could use `body?.is_production` but for the sake of using the existing pattern

packages/server/lib/controllers/v1/connections/connectionId/getConnection.ts
```diff
@@ -90,6 +94,13 @@ export const getConnection = asyncWrapper<GetConnection>(async (req, res) => {
     if (credentialResponse.isOk()) {
         connection = credentialResponse.value;
     }
+
+    // Strip credentials from response if the user doesn't have permission to view them
+    const canReadCredentials = authz?.canReadCredentials ?? true;
```
TBonnin: should the default be `false`?

packages/server/lib/authz/resolvers.ts
```diff
@@ -0,0 +1,76 @@
...
+    ['GET', '/connections', (l) => ({ action: 'read', resource: 'connection', isProduction: isProd(l) })],
+    ['GET', '/connections/count', (l) => ({ action: 'read', resource: 'connection', isProduction: isProd(l) })],
+    ['GET', '/connections/:connectionId', (l) => ({ action: 'read', resource: 'connection', isProduction: isProd(l) })],
+    ['GET', '/environments/current', (l) => ({ action: 'read', resource: 'environment', isProduction: isProd(l) })],
+    ['GET', '/logs/operations/:id', (l) => ({ action: 'read', resource: 'log', isProduction: isProd(l) })],
+    ['POST', '/logs/operations', (l) => ({ action: 'read', resource: 'log', isProduction: isProd(l) })],
+    ['POST', '/logs/messages', (l) => ({ action: 'read', resource: 'log', isProduction: isProd(l) })],
+    ['POST', '/logs/filters', (l) => ({ action: 'read', resource: 'log', isProduction: isProd(l) })],
+    ['POST', '/logs/insights', (l) => ({ action: 'read', resource: 'log', isProduction: isProd(l) })]
+]);
```
TBonnin: Having a separate list of endpoints means we are gonna need to  keep it in sync with `routes.private.ts`. I would prefer to use a helper and have the route defined in a single place with permissions attached

packages/server/lib/controllers/v1/environment/getEnvironment.ts
```diff
@@ -39,6 +39,12 @@ export const getEnvironment = asyncWrapper<GetEnvironment>(async (req, res) => {
         }
     }
 
+    // Remove secret key if user doesn't have permission to read production secrets
+    if (!(authz?.canReadProdSecrets ?? true)) {
+        delete (environment as any).secret_key;
+        delete (environment as any).pending_secret_key;
+    }
```
TBonnin, line 48: would it make more sense to unset those in the middleware that is setting `res.locals.environment` instead of doing it at the endpoint level?

packages/server/lib/controllers/v1/integrations/providerConfigKey/getIntegration.ts
```diff
@@ -61,10 +61,15 @@ export const getIntegration = asyncWrapper<GetIntegration>(async (req, res) => {
         webhookSecret = crypto.createHash('sha256').update(hash).digest('hex');
     }
 
+    const includeCredentials = authz?.canReadCredentials ?? true;
     const count = await connectionService.countConnections({ environmentId: environment.id, providerConfigKey: params.providerConfigKey });
+    const apiIntegration = integrationToApi(integration);
+    if (!includeCredentials) {
+        delete (apiIntegration as any).oauth_client_secret;
+    }
```
TBonnin: `integrationToApi` might be a better placer to remove attributes. Then we don't have to add the logic in several endpoints that return integrations?

packages/server/lib/authz/types.ts
```diff
@@ -0,0 +1,11 @@
+import type { Role } from '@nangohq/types';
+
+export interface Permission {
+    action: string; // 'create' | 'delete' | 'write' | 'read' | '*'
+    resource: string; // 'environment' | 'integration' | 'team' | '*'
```
TBonnin: can we literal unions instead of untyped strings?

packages/server/lib/authz/types.ts
```diff
@@ -0,0 +1,11 @@
...
+    isProduction: boolean | null; // true = production env, false = non-production, null = not environment-scoped
```
TBonnin: the `is_production` concept is already falling apart with a null discriminant that require to read this comment to understand

Can we replace `isProduction` by a `Scope: production | non-production | global` attribute that would be self-explanatory and doesn't hide the semantic in the comment/implementation

packages/server/lib/authz/deny-map.ts
```diff
@@ -0,0 +1,57 @@
...
+        { action: 'delete', resource: 'invite', isProduction: null },
+        { action: 'write', resource: 'connect_ui_settings', isProduction: null },
+        { action: '*', resource: 'billing', isProduction: null },
+        { action: 'write', resource: 'plan', isProduction: null },
+        { action: 'write', resource: 'environment_production_flag', isProduction: null },
+        { action: 'create', resource: 'environment', isProduction: null },
+        // all production access (read + write)
+        { action: '*', resource: '*', isProduction: true }
+    ]
+};
```
TBonnin, line 37: can we avoid duplication and simplify adding new rules by defining constants 
```
  const ADMIN_ONLY = [
      { action: 'write',  resource: 'team', scope: 'global' },
      { action: 'delete', resource: 'team_member',  scope: 'global' },
      { action: 'write',  resource: 'invite',  scope: 'global' },
      { action: 'delete', resource: 'invite',  scope: 'global' },
      { action: 'write',  resource: 'connect_ui_settings', scope: 'global' },
      { action: '*',      resource: 'billing', scope: 'global' },
      { action: 'write',  resource: 'plan', scope: 'global' },
      { action: 'write',  resource: 'environment_production_flag',scope: 'global' },
      { action: 'create', resource: 'environment',  scope: 'global' },
  ] as const satisfies Permission[];

  const PROD_WRITES = [
      { action: 'write',  resource: 'integration', scope: 'production' },
      { action: 'delete', resource: 'integration',  scope: 'production' },
      { action: 'write',  resource: 'connection',   scope: 'production' },
      { action: 'delete', resource: 'connection',   scope: 'production' },
      { action: 'write',  resource: 'flow', scope: 'production' },
      { action: 'delete', resource: 'environment',  scope: 'production' },
      { action: 'write',  resource: 'environment',  scope: 'production' },
      { action: 'write',  resource: 'environment_key', scope: 'production' },
      { action: 'write',  resource: 'environment_variable',  scope: 'production' },
      { action: 'write',  resource: 'webhook',  scope: 'production' },
  ] as const satisfies Permission[];

  const PROD_SECRETS = [
      { action: 'read', resource: 'secret_key',             scope: 'production' },
      { action: 'read', resource: 'connection_credential',  scope: 'production' },
  ] as const satisfies Permission[];

  export const ROLE_DENY_MAP: Record<Role, Permission[]> = {
      administrator: [],
      production_support: [...ADMIN_ONLY, ...PROD_WRITES, ...PROD_SECRETS],
      development_full_access: [..ADMIN_ONLY, { action: '*', resource: '*', scope: 'production' }],
  };
```

packages/server/lib/authz/authorize.ts
pfreixes: Good call, done here https://github.com/NangoHQ/nango/pull/5644/changes/e26dcfbf6f28b71ab49c34dea417611b19f6d411

packages/server/lib/authz/types.ts
pfreixes: Good call, done here https://github.com/NangoHQ/nango/pull/5644/changes/e26dcfbf6f28b71ab49c34dea417611b19f6d411

packages/server/lib/controllers/v1/account/managed/getCallback.ts
pfreixes: Good call done here https://github.com/NangoHQ/nango/pull/5644/changes/e26dcfbf6f28b71ab49c34dea417611b19f6d411 also defaulting to `administrator` when there is no env variable

packages/server/lib/controllers/v1/account/signup.ts
pfreixes: Good call done here https://github.com/NangoHQ/nango/pull/5644/changes/e26dcfbf6f28b71ab49c34dea417611b19f6d411 also defaulting to `administrator` when there is no env variable

packages/server/lib/authz/permissions.ts
```diff
@@ -0,0 +1,39 @@
...
+    canWriteProdConnections: { action: 'write', resource: 'connection', isProduction: true },
+    canWriteProdFlows: { action: 'write', resource: 'flow', isProduction: true },
+    canWriteProdEnvironment: { action: 'write', resource: 'environment', isProduction: true },
+    canWriteProdEnvironmentKeys: { action: 'write', resource: 'environment_key', isProduction: true },
+    canDeleteProdEnvironment: { action: 'delete', resource: 'environment', isProduction: true },
+
+    // production secrets/credentials
+    canReadProdSecretKey: { action: 'read', resource: 'secret_key', isProduction: true },
+    canReadProdConnectionCredentials: { action: 'read', resource: 'connection_credential', isProduction: true }
+};
```
TBonnin: I am not sure what the CAPABILITIES map is for. Is it to give a application-friendly name to a permissions. Can the name be defined in a single place alongside the permissions or at least have permissions defined in a single place and reference in the deny map and the capabilities map?

packages/server/lib/controllers/v1/integrations/providerConfigKey/getIntegration.ts
```diff
@@ -61,10 +61,15 @@ export const getIntegration = asyncWrapper<GetIntegration>(async (req, res) => {
         webhookSecret = crypto.createHash('sha256').update(hash).digest('hex');
     }
 
+    const includeCredentials = authz?.canReadCredentials ?? true;
```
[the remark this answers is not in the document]
pfreixes: Its a sha256 derived value and I understand that `prod_support` role users will want to copy the webhook url. So we should keep returning this one

/cc @TBonnin @bastienbeurier

packages/server/lib/authz/permissions.ts
pfreixes: CAPABILITIES is being used by the frontend for knowing if the current user has permissions for making all the described actions with the goal of deciding to hide or not certain parts of the UI. It basically describes what permissions are required for each resource and they are computed against the rules that we have defined.

We could move all of this logic under `getUser.ts` which is basically the endpoint that returns this materialization.

Thoughts?

packages/server/lib/authz/middleware.ts
pfreixes: Yeps, good point, we do not need fallback at all here since the migrations made sure that all users have a role assigned and by default we are assigning `administrator`, removed the fallback [here](https://github.com/NangoHQ/nango/pull/5644/changes/d7b458dc3bb5e20d4e66843d7a78f56734461683)

packages/server/lib/authz/resolvers.ts
pfreixes: This is something that I was also thinking about, good point. Ive tried to put all of them together using a helper as you suggested, take a look to this [one](https://github.com/NangoHQ/nango/pull/5644/changes/d7b458dc3bb5e20d4e66843d7a78f56734461683)

packages/server/lib/controllers/v1/integrations/providerConfigKey/getIntegration.ts
pfreixes: Yeps good catch, moved there and also added four new integration tests which caught the same bug as you, check [this](https://github.com/NangoHQ/nango/pull/5644/changes/d7b458dc3bb5e20d4e66843d7a78f56734461683).

packages/server/lib/authz/types.ts
pfreixes: Yeps, sorry did not know that we could do that with TS, done [here](https://github.com/NangoHQ/nango/pull/5644/changes/d7b458dc3bb5e20d4e66843d7a78f56734461683)!

packages/server/lib/authz/types.ts
pfreixes: Good approach, changed this to what you suggested [here](https://github.com/NangoHQ/nango/pull/5644/changes/d7b458dc3bb5e20d4e66843d7a78f56734461683).

packages/server/lib/authz/deny-map.ts
pfreixes: Yeps, done [here](https://github.com/NangoHQ/nango/pull/5644/changes/d7b458dc3bb5e20d4e66843d7a78f56734461683).

packages/server/lib/controllers/v1/environment/getEnvironment.ts
pfreixes: I think I would prefer to create a formatter and move it there so we are consistent with what we do with integrations and connections so we keep a consistent pattern, WDYT?

packages/server/lib/controllers/v1/connections/connectionId/getConnection.ts
pfreixes: when permissions are not enabled we default to `true` for all of the predicates, this follows the same rule as the auth evaluator which basically returns true for accessing any route.

packages/server/lib/authz/evaluator.ts
pfreixes: this should be addressed too

packages/server/lib/authz/deny-map.ts
TBonnin: I actually think we can go further and define each permissions individually so they can be reference in the code. Ex: 

```
// permissions.ts
import type { Permission } from './types.js';
export const permissions = {
    canManageTeam: { action: 'write',  resource: 'team', scope: 'global' },
    ...
}

// deny-map.ts
import { permissions as p } from './permissions.js';

const ADMIN_ONLY: Permission[] = [
    p.canManageTeam, 
    p.canDeleteTeamMember,
    ...
];

const PROD_WRITES: Permission[] = [
    p.canWriteIntegrationProduction, 
    ...
];

const PROD_SECRETS: Permission[] = [
    p.canReadSecretKey, 
    p.canReadConnectionCredential,
];

export const ROLE_DENY_MAP: Record<Role, Permission[]> = {
    administrator:           [],
    production_support:      [...ADMIN_ONLY, ...PROD_WRITES, ...PROD_SECRETS],
    development_full_access: [...ADMIN_ONLY, { action: '*', resource: '*', scope: 'production' }],
};

```

The advantage is the individual permission can be referenced in the rest of the code. For instance, no need for yet another translation layer like `AuthzLocals` for in-app enforcement. We can store `role` in  `locals` and then call `resolve({ locals['role'], permission.canReadSecretKey}) at runtime instead of mapping every single possible permissions in the middleware

packages/server/lib/authz/permissions.ts
TBonnin: We are basically building an allow list for the frontend. Instead of repeating the permission in this file, can we send the permission directly to the frontend by constructing an allow list like:

```
type AllowedPermissions = Partial<Record<Resource, Partial<Record<Scope, Action[]>>>>;

function resolvePermissions(role: Role): AllowedPermissions {
    const denyList = ROLE_DENY_MAP[role];
    const allowed = ALL_PERMISSIONS.filter(p => !isDenied(p, denyList));

    return allowed.reduce<AllowedPermissions>((acc, { action, resource, scope }) => {
        acc[resource]        ??= {};
        acc[resource][scope] ??= [];
        acc[resource][scope].push(action);
        return acc;
    }, {});
}
```
then the frontend can use the allow list like this 
```
import type { AllowedPermissions, Resource, Scope, Action } from '@nangohq/types';

export function can(
    permissions: AllowedPermissions,
    permission: Permission
): boolean {
    return permissions[resource]?.[scope]?.includes(action) ?? false;
}

// then
const {  user } =  useQuery<GetUser['Success']>('/api/v1/user', { data: user }, { revalidate: false });
<Button disabled={!can(user.permissions, {'write', 'integration', 'production'})}>Create Integration</Button>

```

packages/server/lib/authz/types.ts
```diff
@@ -0,0 +1,34 @@
...
+    | 'connection_credential'
+    | 'log'
+    | '*';
+export type Scope = 'production' | 'non-production' | 'global';
+
+export interface Permission {
+    action: Action;
+    resource: Resource;
+    scope: Scope;
+}
```
TBonnin, line 30: this type doesn't prevent impossible combinations. Can we derived the Permission type from the set of valid permissions:
```
const ALL_PERMISSIONS = [...ADMIN_ONLY, ...PROD_WRITES, ...PROD_SECRETS] as const;
export type Permission = typeof ALL_PERMISSIONS[number];
```

packages/server/lib/routes.private.ts
```diff
@@ -131,6 +140,25 @@ web.use(
...
+): void {
+    const m = method.toLowerCase() as 'get' | 'post' | 'put' | 'patch' | 'delete';
+    web.route(path)[m](...webAuth, handler);
+    registerPermission(method, path, permission);
+}
+
+function openRoute<L extends Record<string, any>>(method: HttpMethod, path: string, handler: RequestHandler<any, any, any, any, L>): void {
+    const m = method.toLowerCase() as 'get' | 'post' | 'put' | 'patch' | 'delete';
+    web.route(path)[m](...webAuth, handler);
+}
```
TBonnin: would it be simpler as a normal express middleware.
```
function can(permission: Permission): RequestHandler {
    return (req, res, next) => {
        // resolve permission against req.user
        ...
        next();
    };
}

// then
web.route('/meta').get(webAuth, can(permission.CanDoSomething),  getMeta)
...
```

packages/server/lib/authz/types.ts
```diff
@@ -0,0 +1,34 @@
+import type { Role } from '@nangohq/types';
+
+export type Action = 'create' | 'delete' | 'write' | 'read' | '*';
```
TBonnin: is `write` encapsulate create and delete. should it be `create | delete | update | read | *`

packages/server/lib/authz/deny-map.ts
pfreixes: Yea this also helps us to merge the capabilities and the permissions into one single entity, done [here](https://github.com/NangoHQ/nango/pull/5644/changes/7a7a21e107986ca98676c21297078cb083493935).

packages/server/lib/authz/types.ts
pfreixes: We have the wildcard one which covers any action,operation for production, we should also explicitly enumerate them for making this work.

Thoughts?

packages/server/lib/routes.private.ts
pfreixes: great suggestion, I like this idea. Done

packages/server/lib/authz/types.ts
pfreixes: `write` no, write is only about update. Let me rename it.

pfreixes: done

packages/server/lib/controllers/v1/integrations/getIntegrations.ts
```diff
@@ -15,6 +17,7 @@ export const getIntegrations = asyncWrapper<GetIntegrations>(async (req, res) =>
     }
 
     const { environment } = res.locals;
+    const includeCredentials = environment.is_production ? await resolve(res.locals, permissions.canReadProdConnectionCredentials) : true;
```
TBonnin: can be simplified with `environment.is_production || await resolve(res.locals, permissions.canReadProdConnectionCredentials)`

packages/server/lib/formatters/connection.ts
```diff
@@ -34,15 +34,15 @@ export function connectionSimpleToApi({
...
+export function connectionFullToApi(connection: DBConnectionDecrypted, options?: { includeCredentials?: boolean }): ApiConnectionFull {
     return {
         id: connection.id,
         config_id: connection.config_id,
         environment_id: connection.environment_id,
         connection_id: connection.connection_id,
         provider_config_key: connection.provider_config_key,
         connection_config: connection.connection_config,
-        credentials: connection.credentials,
+        credentials: options?.includeCredentials === false ? ({} as DBConnectionDecrypted['credentials']) : connection.credentials,
```
TBonnin: i would simplify and reverse the condition : `options?.includeCredentials ? connection.credentials : {}`

packages/server/lib/authz/index.ts
```diff
@@ -0,0 +1,6 @@
+export { can, envScope } from './middleware.js';
+export { evaluator } from './evaluator.js';
+export { permissions } from './permissions.js';
+export { buildPermissions, resolve } from './resolve.js';
+export { ROLE_DENY_MAP } from './deny-map.js';
+export type { Permission, PermissionEvaluator } from './types.js';
```
TBonnin: not sure an index is super useful here since it doesn't expose anything that is already exported and just add a level of indirection

packages/server/lib/routes.private.ts
```diff
@@ -156,86 +158,221 @@ if (flagHasManagedAuth) {
...
-web.route('/environments').post(webAuth, postEnvironment);
-web.route('/environments/').patch(webAuth, patchEnvironment);
-web.route('/environments/').delete(webAuth, deleteEnvironment);
-web.route('/environments/current').get(webAuth, getEnvironment);
-web.route('/environments/webhook').patch(webAuth, patchWebhook);
-web.route('/environments/variables').post(webAuth, postEnvironmentVariables);
+web.route('/environments').post(webAuth, can(p.canCreateEnvironment), postEnvironment);
+web.route('/environments/').patch(
+    webAuth,
+    can((l) => ({ action: 'update', resource: 'environment', scope: envScope(l) })),
```
TBonnin: we could alter the signature of can to simplify the way it is called
```
export function cane(permission: Permission | (Omit<Permission, 'scope'> & { scopedBy: (locals: RequestLocals) => Scope })): RequestHandler {
...
        const allowed = await evaluator.evaluate(user.role, {
            action: permission.action,
            resource: permission.resource,
            scope: 'scopedBy' in permission ? permission.scopedBy(res.locals as RequestLocals) : permission.scope;
        });

...
    };
}

// then called as
web.route('/plans/trial/extension').post(
   webAuth, 
   can(p.canChangePlan), 
   postPlanExtendTrial
);
// or
web.route('/environments/variables').post(
    webAuth,
    can({ action: 'update', resource: 'environment_variable', scopedBy: envScope })),
    postEnvironmentVariables
);

```

packages/server/lib/authz/resolve.ts
```diff
@@ -0,0 +1,30 @@
...
+    if (!flags.hasAuthRoles) return true;
+    const user = locals.user;
+    if (!user) return true;
+    return evaluator.evaluate(user.role, permission);
+}
+
+export async function buildPermissions(role: Role): Promise<AllowedPermissions> {
+    const result: AllowedPermissions = {};
+    for (const perm of Object.values(permissions)) {
+        const allowed = flags.hasAuthRoles ? await evaluator.evaluate(role, perm) : true;
```
TBonnin: const allowed = flags.hasAuthRoles || await evaluator.evaluate(role, perm);

packages/server/lib/authz/resolve.ts
```diff
@@ -0,0 +1,30 @@
...
+        if (!allowed) continue;
+        const byScope = result[perm.resource] ?? (result[perm.resource] = {});
+        const actions = byScope[perm.scope] ?? (byScope[perm.scope] = []);
```
TBonnin: nit: not a huge fan of this very terse onliner with ?? and assignation.
I even find the verbose option more readable
```
if (!result[perm.resource]) {
  result[perm.resource] = {};
}
const byScope = result[perm.resource];
if (!byScope[perm.scope]) {
  byScope[perm.scope] = [];
}
```

packages/types/lib/user/api.ts
```diff
@@ -38,6 +38,15 @@ export interface ApiUser {
     gettingStartedClosed: boolean;
 }
 
+export type AllowedPermissions = Partial<
+    Record<string, Partial<Record<'production' | 'non-production' | 'global', ('create' | 'read' | 'update' | 'delete' | '*')[]>>>
+>;
+
+export type ApiUserWithPermissions = ApiUser & {
+    role: string;
```
TBonnin: should it be of type `Role`

packages/server/lib/authz/permissions.ts
```diff
@@ -0,0 +1,31 @@
+import type { Permission } from './types.js';
+
+export const permissions = {
+    // team management (global scope)
+    canManageTeam: { action: 'update', resource: 'team', scope: 'global' },
+    canRemoveTeamMember: { action: 'delete', resource: 'team_member', scope: 'global' },
+    canInviteMember: { action: 'update', resource: 'invite', scope: 'global' },
```
TBonnin: shouldn't the action be 'create'?

packages/server/lib/routes.private.ts
```diff
@@ -156,86 +158,221 @@ if (flagHasManagedAuth) {
...
+);
 web.route('/connections/admin/:connectionId').delete(webAuth, connectionController.deleteAdminConnection.bind(connectionController));
 
+// User
 web.route('/user').get(webAuth, getUser);
 web.route('/user').patch(webAuth, patchUser);
 web.route('/user/password').put(webAuth, putUserPassword);
 
+// Sync / Flows
 web.route('/sync').get(webAuth, syncController.getSyncsByParams.bind(syncController));
```
TBonnin: should this endpoint be behind permission. Otherwise non-production role will be able to call it. no?

TBonnin (approved):

packages/shared/lib/services/user.service.ts
```diff
@@ -125,7 +127,8 @@ class UserService {
                 account_id,
                 email_verified,
                 email_verification_token: email_verified ? null : uuid.v4(),
-                email_verification_token_expires_at: email_verified ? null : expires_at
+                email_verification_token_expires_at: email_verified ? null : expires_at,
+                role
```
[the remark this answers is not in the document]
pfreixes: this was already done!

packages/shared/lib/services/invitations.ts
```diff
@@ -47,7 +49,8 @@ export async function inviteEmail({
             account_id: accountId,
             invited_by: invitedByUserId,
             token,
-            expires_at
+            expires_at,
+            role
```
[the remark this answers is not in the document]
pfreixes: same as above its already done

packages/server/lib/routes.private.ts
pfreixes: good point, missed this one in the original document. Protected for not allowing to get the syncs for production envs for none production role

packages/server/lib/authz/permissions.ts
pfreixes: yeps, done

packages/types/lib/user/api.ts
pfreixes: yeps, done

packages/server/lib/authz/resolve.ts
pfreixes: done

packages/server/lib/authz/resolve.ts
pfreixes: done

packages/server/lib/routes.private.ts
pfreixes: done

packages/server/lib/authz/index.ts
pfreixes: done

packages/server/lib/formatters/connection.ts
pfreixes: done

packages/server/lib/controllers/v1/integrations/getIntegrations.ts
pfreixes: done

packages/webapp/src/pages/Account/EmailVerified.tsx
```diff
@@ -60,7 +60,7 @@ export const EmailVerified: React.FC = () => {
                 });
 
                 signin(user);
-                await mutate<GetUser['Success']>('/api/v1/user', { data: user }, { revalidate: false });
+                await mutate<GetUser['Success']>('/api/v1/user', { data: user as GetUser['Success']['data'] }, { revalidate: false });
```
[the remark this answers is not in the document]
pfreixes: We are not yet making usage of the permissions in the frontend so this if needs to be addressed can be done in the frontend PR

PR #5648, opened 2026-03-16
fix(refreshToken): return Redis error on lock timeout during token refresh

hassan254-prog:

- Previously, any exception thrown inside `refreshCredentialsIfNeeded` was wrapped as `refresh_token_external_error` (400), including [internal Redis lock](https://github.com/NangoHQ/nango/blob/14637d0448cf8ed2d22a43d764e76b016676d265/packages/shared/lib/services/connections/credentials/refresh.ts#L426) acquisition timeouts. This caused infra failures to appear as OAuth provider errors, making it impossible for callers to distinguish the two. This change detects lock timeout errors and maps them to a new `refresh_token_lock_timeout` error type with a 503 status.

---

It also introduces a typed lock-timeout error in the kvstore layer, updates NangoError metadata for the new error type, and adds a unit test to validate the lock-timeout mapping behavior.

---

### DIFF
--- a/packages/kvstore/lib/Locking.ts
+++ b/packages/kvstore/lib/Locking.ts
@@ -6,6 +6,18 @@ export interface Lock {
     readonly key: string;
 }
 
+export class LockTimeoutError extends Error {
+    public readonly key: string;
+    public readonly timeoutMs: number;
+
+    constructor(key: string, timeoutMs: number) {
+        super(`Acquiring lock for key: ${key} timed out after ${timeoutMs}ms`);
+        this.name = this.constructor.name;
+        this.key = key;
+        this.timeoutMs = timeoutMs;
+    }
+}
+
 export class Locking {
     private store: KVStore;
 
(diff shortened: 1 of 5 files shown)


TBonnin (changes requested):
discussed on Slack. 
For reference the refresh logic is here https://github.com/NangoHQ/nango/blob/9b1c463f8e6f5deba40fd9d7190f5e90062e72da/packages/shared/lib/services/connections/credentials/refresh.ts?plain=1#L424-L450 

If a refresh lock cannot be acquired and times out (aka: refresh was already in-flight) Nango tries to retrieve the connection (in case it was refreshed in parallel). If the connection hasn't been refreshed, Nango fails with a 400 error which means that the connection couldn't be refreshed. I don't think returning an internal error when Nango couldn't refresh the credentials with the provider is appropriate.

hassan254-prog:
@TBonnin the problem is that we don’t differentiate between credentials failing to refresh due to an external failure or due to the lock, since the error is propagated to the outer try/catch when it occurs.

TBonnin:
> @TBonnin the problem is that we don’t differentiate between credentials failing to refresh due to an external failure or due to the lock, since the error is propagated to the outer try/catch when it occurs.

What I am trying to say, and apparently not doing a very good job at it, is that the lock timing out and the credentials not to be refreshed isn't a locking problem. It means:
- even after waiting this request didn't refresh the connection (lock expired)
- no other request did refresh the connection while waiting. (connection check after timeout expired).

We indeed don't differentiate between the reason why the connection wasn't refreshed. It could be an error or a timeout from the provider. Assuming it is always a timeout (which is what this PR is doing) isn't more appropriate than assuming it is always a failure.

PR #5649, opened 2026-03-16
fix(lambda): node routing overrides

rossmcewan:
---

**Add plan-level Lambda routing overrides and routing context propagation**

This PR introduces plan-scoped Lambda routing overrides by adding a `fleet_node_routing_override` column on `plans` and propagating a new `RoutingContext` through job execution, runtime selection, and Lambda invocation. Routing IDs now use a configurable prefix plus a t‑shirt size derived from `LAMBDA_DEFAULT_MEMORY_MB`.

It also moves Lambda defaults to environment-driven values, enforces AWS resource caps when creating Lambda functions, relaxes Lambda ARN validation to support custom prefixes, and updates plan typing, seed data, and unit tests to cover the new routing behavior.

---

### DIFF
--- a/packages/database/lib/migrations/20260316194000_add_node_routing_override.cjs
+++ b/packages/database/lib/migrations/20260316194000_add_node_routing_override.cjs
@@ -6,7 +6,7 @@ exports.config = { transaction: true };
 exports.up = async function (knex) {
     await knex.raw(`
         ALTER TABLE plans
-        ADD COLUMN IF NOT EXISTS node_routing_override varchar(255) DEFAULT NULL;
+        ADD COLUMN IF NOT EXISTS fleet_node_routing_override varchar(255) DEFAULT NULL;
     `);
 };
 
--- a/packages/database/lib/migrations/20260316194000_add_node_routing_override.cjs
+++ b/packages/database/lib/migrations/20260316194000_add_node_routing_override.cjs
@@ -1,0 +1,13 @@
+exports.config = { transaction: true };
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`
+        ALTER TABLE plans
+        ADD COLUMN IF NOT EXISTS node_routing_override varchar(255) DEFAULT NULL;
+    `);
+};
+
+exports.down = async function () {};
--- a/packages/jobs/lib/runner/lambda.ts
+++ b/packages/jobs/lib/runner/lambda.ts
@@ -219,7 +219,7 @@ export const lambdaNodeProvider: NodeProvider = {
     defaultNodeConfig: {
         cpuMilli: 500,
         memoryMb: 512,
-        storageMb: 10240,
+        storageMb: 512,
         isTracingEnabled: false,
         isProfilingEnabled: false,
         idleMaxDurationMs: 0,
--- a/packages/jobs/lib/runner/lambda.ts
+++ b/packages/jobs/lib/runner/lambda.ts
@@ -20,7 +20,6 @@ import { Err, Ok, getLogger } from '@nangohq/utils';
 
 import { envs } from '../env.js';
 import { registerWithFleet } from '../runtime/runtimes.js';
-import { getSizeFromRoutingId } from '../utils/lambda.js';
 
 import type { Environment } from '@aws-sdk/client-lambda';
 import type { Node, NodeProvider } from '@nangohq/fleet';
@@ -33,10 +32,6 @@ const applicationAutoScalingClient = new ApplicationAutoScalingClient();
 
 export function getFunctionName(node: Node): string {
     return `${node.routingId}-${node.id}`;
-}
-
-function getSize(node: Node): number {
-    return getSizeFromRoutingId(node.routingId);
 }
 
 export function getFunctionQualifier(node: Node): string {
@@ -72,8 +67,11 @@ class Lambda {
                     },
                     PackageType: 'Image',
                     Architectures: [envs.LAMBDA_ARCHITECTURE],
-                    MemorySize: getSize(node),
-                    Timeout: node.executionTimeoutSecs || envs.LAMBDA_EXECUTION_TIMEOUT_SECS,
+                    MemorySize: Math.min(10240, node.memoryMb), //max 10GB allowed by Lambda
+                    EphemeralStorage: {
+                        Size: Math.min(10240, node.storageMb) //max 10GB allowed by Lambda
+                    },
+                    Timeout: Math.min(900, node.executionTimeoutSecs || envs.LAMBDA_EXECUTION_TIMEOUT_SECS), //max 15 minutes allowed by Lambda
                     Environment: this.getEnvironmentVariables(node),
                     VpcConfig: {
                         SubnetIds: envs.LAMBDA_SUBNET_IDS,
@@ -209,7 +207,7 @@ class Lambda {
     }
 
     async verifyUrl(_url: string): Promise<Result<void>> {
-        const regex = /^arn:aws:lambda:.*:function:nango-runner-function-\d+-[a-f0-9-]+:.*$/;
+        const regex = /^arn:aws:lambda:.*:function:.*/;
         if (!regex.test(_url)) {
             return Err('Invalid URL');
         }
@@ -221,12 +219,12 @@ export const lambdaNodeProvider: NodeProvider = {
     defaultNodeConfig: {
         cpuMilli: 500,
         memoryMb: 512,
-        storageMb: 20000,
+        storageMb: 10240,
         isTracingEnabled: false,
         isProfilingEnabled: false,
         idleMaxDurationMs: 0,
-        executionTimeoutSecs: 900,
-        provisionedConcurrency: 1,
+        executionTimeoutSecs: envs.LAMBDA_EXECUTION_TIMEOUT_SECS,
+        provisionedConcurrency: envs.LAMBDA_PROVISIONED_CONCURRENCY,
         replicas: 1
     },
     start: async (node: Node) => {
--- a/packages/jobs/lib/runner/lambda.ts
+++ b/packages/jobs/lib/runner/lambda.ts
@@ -123,8 +123,8 @@ class Lambda {
                         ServiceNamespace: 'lambda',
                         ScalableDimension: 'lambda:function:ProvisionedConcurrency',
                         ResourceId: resourceId,
-                        MinCapacity: node.provisionedConcurrency || envs.LAMBDA_PROVISIONED_CONCURRENCY,
-                        MaxCapacity: (node.provisionedConcurrency || envs.LAMBDA_PROVISIONED_CONCURRENCY) * 10
+                        MinCapacity: node.provisionedConcurrency || envs.LAMBDA_DEFAULT_PROVISIONED_CONCURRENCY,
+                        MaxCapacity: (node.provisionedConcurrency || envs.LAMBDA_DEFAULT_PROVISIONED_CONCURRENCY) * 10
                     })
                 );
                 await applicationAutoScalingClient.send(
@@ -224,7 +224,7 @@ export const lambdaNodeProvider: NodeProvider = {
         isProfilingEnabled: false,
         idleMaxDurationMs: 0,
         executionTimeoutSecs: envs.LAMBDA_EXECUTION_TIMEOUT_SECS,
-        provisionedConcurrency: envs.LAMBDA_PROVISIONED_CONCURRENCY,
+        provisionedConcurrency: envs.LAMBDA_DEFAULT_PROVISIONED_CONCURRENCY,
         replicas: 1
     },
     start: async (node: Node) => {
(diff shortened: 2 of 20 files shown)


packages/jobs/lib/runner/lambda.ts
```diff
@@ -209,7 +207,7 @@ class Lambda {
     }
 
     async verifyUrl(_url: string): Promise<Result<void>> {
-        const regex = /^arn:aws:lambda:.*:function:nango-runner-function-\d+-[a-f0-9-]+:.*$/;
+        const regex = /^arn:aws:lambda:.*:function:.*/;
```
rossmcewan, line 210: Prefixes can be overridden from `plan` so simplify the verification.

packages/database/lib/migrations/20260316194000_add_node_routing_override.cjs
```diff
@@ -0,0 +1,13 @@
+exports.config = { transaction: true };
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`
+        ALTER TABLE plans
+        ADD COLUMN IF NOT EXISTS node_routing_override varchar(255) DEFAULT NULL;
```
TBonnin: not sure the name `node_routing_override` is conveying this is about the runners/functions. What about `functions_routing_override`?
Should it actually be more `functions_routing_prefix`?

packages/utils/lib/environment/parse.ts
```diff
@@ -382,7 +382,7 @@ export const ENVS = z.object({
 
     // Lambda
     LAMBDA_ENABLED: z.stringbool().optional().default(false),
-    LAMBDA_DEFAULT_SIZE: z.coerce.number().default(512),
+    LAMBDA_PREFIX: z.string().optional().default('nango-function'),
```
TBonnin: `LAMBDA_DEFAULT_PREFIX`?

packages/jobs/lib/utils/lambda.ts
```diff
@@ -1,20 +1,23 @@
...
+    return envs.LAMBDA_DEFAULT_MEMORY_MB;
 }
 
-export function getSizeFromRoutingId(name: string): number {
-    const parts = name.split('-');
-    const size = Number.parseInt((parts.length && parts[parts.length - 1]) || '');
-    if (Number.isNaN(size)) return envs.LAMBDA_DEFAULT_SIZE;
-    return size;
+export function getRoutingId(params: { nangoProps: NangoProps; runtimeContext?: RuntimeContext | undefined }): string {
+    const size = getSizeFromProps(params.nangoProps);
```
TBonnin: `memeoryMb` so we know what we are talking about. Same below in `getTShirtSize`

packages/jobs/lib/utils/lambda.ts
```diff
@@ -1,20 +1,23 @@
...
 }
 
-export function getRoutingId(nangoProps: NangoProps): string {
-    const size = getSizeFromProps(nangoProps);
-    return `nango-runner-function-${size}`;
+function getTShirtSize(size: number): string {
+    if (size <= 1024) return 'S';
+    if (size <= 2048) return 'M';
+    if (size <= 4096) return 'L';
+    if (size <= 8192) return 'XL';
```
TBonnin: is it not strict `<` otherwise 1024Mb would be a S. no?

packages/jobs/lib/runtime/adapter.ts
```diff
@@ -1,7 +1,13 @@
...
 import type { Result } from '@nangohq/utils';
 
 export interface RuntimeAdapter {
-    invoke(params: { taskId: string; nangoProps: NangoProps; code: string; codeParams: object }): Promise<Result<boolean>>;
+    invoke(params: {
+        taskId: string;
+        nangoProps: NangoProps;
+        code: string;
+        codeParams: object;
+        runtimeContext?: RuntimeContext | undefined;
```
TBonnin: what if we rename `RuntimeContext` into `RoutingContext` since the info it contains is used for routing purpose (routing to lambda or runner, routing to specific lambdas, ...)

packages/jobs/lib/runner/lambda.ts
TBonnin: should we enforce a fixed prefix like `nango-functions-` that is also applied in `getRoutingId`?

TBonnin (approved):

packages/utils/lib/environment/parse.ts
```diff
@@ -416,6 +416,8 @@ export const ENVS = z.object({
     LAMBDA_ARCHITECTURE: z.enum(['arm64', 'x86_64']).optional().default('arm64'),
     LAMBDA_CREATE_TIMEOUT_SECS: z.coerce.number().optional().default(120),
     LAMBDA_EXECUTION_TIMEOUT_SECS: z.coerce.number().optional().default(900),
+    LAMBDA_DEFAULT_MEMORY_MB: z.coerce.number().optional().default(512),
+    LAMBDA_DEFAULT_STORAGE_MB: z.coerce.number().optional().default(512),
```
TBonnin, line 420: it doesn't matter but it was set to `20_000` before this PR

packages/shared/lib/services/plans/definitions.ts
```diff
@@ -31,7 +31,8 @@ export const freePlan: PlanDefinition = {
         sync_function_runtime: 'lambda',
         action_function_runtime: 'lambda',
         webhook_function_runtime: 'lambda',
-        on_event_function_runtime: 'lambda'
+        on_event_function_runtime: 'lambda',
+        node_routing_override: null
```
TBonnin: you don't need to specify it here since the default value is null anyway

packages/shared/lib/services/plans/plans.ts
```diff
@@ -291,6 +291,10 @@ export function mergeFlags({ currentPlan, newPlanDefinition }: { currentPlan: DB
                 }
                 break;
             }
+            case 'node_routing_override': {
+                overrides[key] = currentPlan[key] !== newPlanDefinition.flags[key] ? newPlanDefinition.flags[key] : currentPlan[key];
```
TBonnin: this means the new plan value will override the current one, which would set it to null for all plans definition were it is not set and cancel the override when upgrading. That's not what we want I think

TBonnin (changes requested):

packages/jobs/lib/runtime/lambda.adapter.ts
```diff
@@ -34,8 +34,8 @@ interface S3ObjectRef {
 export class LambdaRuntimeAdapter implements RuntimeAdapter {
     constructor(private readonly fleet: Fleet) {}
 
-    private async getFunction(nangoProps: NangoProps): Promise<LambdaFunction> {
-        const routingId = getRoutingId(nangoProps);
+    private async getFunction(params: { nangoProps: NangoProps; runtimeContext?: RuntimeContext | undefined }): Promise<LambdaFunction> {
+        const routingId = getRoutingId(params);
```
TBonnin, line 38: are you planning to manually add a lambda node with the new naming scheme to avoid created it inline when the first function is being dispatched to it?

packages/jobs/lib/utils/lambda.ts
```diff
@@ -1,20 +1,23 @@
...
-    const parts = name.split('-');
-    const size = Number.parseInt((parts.length && parts[parts.length - 1]) || '');
-    if (Number.isNaN(size)) return envs.LAMBDA_DEFAULT_SIZE;
-    return size;
+export function getRoutingId(params: { nangoProps: NangoProps; runtimeContext?: RuntimeContext | undefined }): string {
+    const size = getSizeFromProps(params.nangoProps);
+    const tshirtSize = getTShirtSize(size);
+    const prefix = params.runtimeContext?.plan?.node_routing_override || envs.LAMBDA_PREFIX;
+    return `${prefix}-${tshirtSize}`;
 }
```
TBonnin, line 9: we could add some tests for this function.

packages/database/lib/migrations/20260316194000_add_node_routing_override.cjs
rossmcewan: I think `node_routing_override` is more appropriate as that is what it is doing in the semantics of the fleet. The codebase generally refers to nodes and the routing_id is a property of a node.

packages/jobs/lib/utils/lambda.ts
rossmcewan: I will collapse this all into t-shirt size because it's doubtful the customer will be able to influence specific memory sizes in MB but rather just t-shirt sizes anyway.

packages/jobs/lib/runtime/adapter.ts
rossmcewan: Yep. Much better name.

packages/utils/lib/environment/parse.ts
rossmcewan: It was, but the value that was used was 512. The max it can be in Lambda is 10GB. This is now used so the value needs to be correct.

packages/shared/lib/services/plans/plans.ts
rossmcewan: Hmm. This might be a problem for the function runtime overrides too right? It would go back to default if we had overridden it?

packages/jobs/lib/runtime/lambda.adapter.ts
rossmcewan: Yes, I think that's the only way without split PRs that could conditionally control the rollout. I have changed the default prefix to `nango-runner-function` so that URL validation still works.

packages/jobs/lib/runner/lambda.ts
rossmcewan: I don't think it's worth it tbh. This is less for security rather than ensuring the ARN is valid, imo.

packages/jobs/lib/runtime/lambda.adapter.ts
rossmcewan: Update: changing default prefix wasn't sufficient as the old validation also expected a number. This has been manually resolved so we have a function ready to go.

packages/database/lib/migrations/20260316194000_add_node_routing_override.cjs
TBonnin: In the context of `fleet` the terminology `node` make sense. When looking at plans though one must be aware that `node` here mean fleet node or runner node. Not obvious imho. 
What about `fleet_node_routing_override`? A bit long but more explicit

packages/jobs/lib/utils/lambda.ts
```diff
@@ -1,20 +1,18 @@
...
-export function getRoutingId(nangoProps: NangoProps): string {
-    const size = getSizeFromProps(nangoProps);
-    return `nango-runner-function-${size}`;
+function getTShirtSize(_nangoProps: NangoProps): string {
+    const memoryMb = envs.LAMBDA_DEFAULT_MEMORY_MB;
+    if (memoryMb < 1024) return 'S';
+    if (memoryMb < 2048) return 'M';
+    if (memoryMb < 4096) return 'L';
+    if (memoryMb < 8192) return 'XL';
+    return 'XXL';
```
TBonnin, line 19: I am confused here. NangoProps argument isn't being used and we are just mapping the env var to a size. Why not having the default prefix contains the default size then no need for this function and getRoutingId becomes 
```ts
export function getRoutingId( routingContext?: RoutingContext | undefined }): string {
    return params.routingContext?.plan?.node_routing_override || envs.LAMBDA_DEFAULT_ROUTING_ID; // with LAMBDA_DEFAULT_ROUTING_ID = 'nango-function-lambda-S'
   // or if you want to keep prefix and size separate default
   `${envs.LAMBDA_DEFAULT_PREFIX}-${envs.LAMBDA_DEFAULT_SIZE}`
}
```

TBonnin (commented):
Made a last comment about `getRoutingId`. Rest looks good thank you

packages/jobs/lib/utils/lambda.ts
rossmcewan: - The routing override doesn't include the size so they have to be separate
- The next change I'll be working on will be supporting the size selection (probably from connection tags)
- Size is already set because it used when defining the default properties for the node - so translating that to a t-shirt size makes more sense otherwise we need to maintain both (or do a reverse translation)
- As for the unused nangoProps and having the `getTshirtSize` function - I think that supports what we want to have when we do the actual size selection

packages/database/lib/migrations/20260316194000_add_node_routing_override.cjs
rossmcewan: I don't mind `fleet_node_routing_override`

packages/jobs/lib/utils/lambda.ts
rossmcewan: I've added comments to make it clearer why it's structured like that.

TBonnin (approved):

PR #5653, opened 2026-03-16
feat: detect features in CLI and pass to /deploy endpoint  

TBonnin:
For now we only detect the use of checkpoints in function code. 

This information will then be used to dynamically dispatch checkpointed syncs to Lambda

Notes:
- built-in templates are supported yet (aka: no features detected for them)
- nango YAML is not supported

---

**Add CLI feature detection and persist `features` through deploy pipeline**

This PR introduces feature detection in the CLI build pipeline (currently tracking checkpoint usage) and propagates a new `features` array through parsed definitions, deploy payloads, and persisted sync configs. It adds database support for storing features, updates shared types and validation to include the field, and ensures templates and YAML parsing default to empty feature lists where detection is not supported.

Tests and fixtures are updated to cover feature detection and to include `features` in expected configs, sync configs, and webhook/job contexts, ensuring the new field is consistently handled across services.

---

### DIFF
--- a/packages/cli/lib/zeroYaml/compile.ts
+++ b/packages/cli/lib/zeroYaml/compile.ts
@@ -16,7 +16,7 @@ import { Spinner } from '../utils/spinner.js';
 import { printDebug } from '../utils.js';
 
 // import type { BabelErrorType } from './constants.js';
-import type { Result } from '@nangohq/types';
+import type { Feature, Result } from '@nangohq/types';
 
 /**
  * This function is used to compile the code in the integration.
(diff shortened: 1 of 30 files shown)


packages/cli/lib/zeroYaml/compile.ts
```diff
@@ -406,6 +406,29 @@ export function tsToJsPath(filePath: string) {
...
+        });
+        const features: Feature[] = [];
+        if (bag.checkpointsLines.length > 0) {
+            features.push('checkpoints');
+        }
+        return Ok(features);
+    } catch (err) {
+        return Err(new Error('failed_to_detect_features', { cause: err }));
+    }
+}
```
TBonnin, line 430: I went for traversing the code again instead of piggy backing on the compilation step because it would have required to entirely refactor how compilation outcome is handled and returned.
Since this is pure ast traversing, the performance impact is low.

packages/cli/lib/zeroYaml/deploy.ts
```diff
@@ -203,6 +204,15 @@ async function createPackage({
                     return Err(new Error(`No script files found for "${sync.name}"`));
                 }
 
+                const entryPoint = path.join(fullPath, providerConfigKey, 'syncs', `${sync.name}.ts`);
+                const features = await detectFeatures({ entryPoint });
+                if (features.isErr()) {
```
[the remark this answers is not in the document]
TBonnin: on purpose for now. features will only be used internally and false negative are acceptable for now. It will result in syncs not being candidate for lambdas which would is already the fallback option

packages/cli/lib/zeroYaml/deploy.ts
```diff
@@ -248,6 +256,15 @@ async function createDeployConfirmationPackage({
...
                 }
 
+                const entryPoint = path.join(fullPath, providerConfigKey, 'actions', `${action.name}.ts`);
+                const features = await detectFeatures({ entryPoint });
+                if (features.isErr()) {
+                    console.error(
+                        chalk.red(`Error detecting features for "${action.name}"`),
+                        features.error instanceof Error ? features.error.message : features.error
+                    );
+                }
```
kaposke: This could also potentially be done in `parseSync`/`parseAction` in [packages/cli/lib/zeroYaml/definitions.ts](https://github.com/nangohq/nango/blob/master/packages/cli/lib/zeroYaml/definitions.ts). "Features" would have to be added to the `ParseNangoSync` and `ParseNangoAction` types, and you would access them here with `action.features`.
However, it would also make it into `nango.json`, which might not be intentional. So no strong opinion tbh, just pointing it out

kaposke: Actually, having it in `nango.json` also means having it in `flows.zero.yaml` which we use to deploy preBuilt templates. So this might be a requirement.

TBonnin: good idea. let me look into it

TBonnin: moved the features detection to parsing 7bf3cbf29a4be249401cfe28428264862414b207
For now not, built-in templates aren't supported

kaposke (approved):

PR #5654, opened 2026-03-17
fix(connect-ui): pass authorization params during reconnect

hassan254-prog:

- pass `authorization-params` during reconnect, same as from [postSession](https://github.com/NangoHQ/nango/blob/9b19b1734df07a157929385c747d1f76f308f72d/packages/server/lib/controllers/connect/postSessions.ts#L204).

---

It also updates flow schema handling by making the JSON schema optional and guarding filtering when a top-level schema is absent, with integration tests adjusted to reflect the new deploy and schema behavior.

---


agusayerza (approved):

packages/server/lib/controllers/sync/deploy/postDeploy.integration.test.ts
```diff
@@ -256,26 +256,45 @@ describe(`POST ${endpoint}`, () => {
                 token: secret.secret,
                 body: {
                     debug: false,
-                    jsonSchema: {
-                        $comment: '',
-                        $schema: 'http://json-schema.org/draft-07/schema#',
-                        definitions: {}
-                    },
-                    flowConfigs: [],
+                    flowConfigs: [
```
hassan254-prog, line 259: Previously this test deployed with empty `flowConfigs` and `reconcile: true`, expecting all syncs to be deleted from the DB, which is not the case since we can't deploy empty flows from the [cli](https://github.com/NangoHQ/nango/blob/9b19b1734df07a157929385c747d1f76f308f72d/packages/cli/lib/zeroYaml/deploy.ts#L272).

TBonnin (approved):

PR #5656, opened 2026-03-17
fix(query): `deleteOutdatedRecords` transaction handling

agusayerza:
## Fix `deleteOutdatedRecords` transaction handling

Fixes NAN-5045

### Problem

We've been seeing a major increase in persist errors ("Failed to save record" / "Failed to delete record") rooted in DB timeouts, with spikes correlating to high DB stress (CPU >85%, block read times 10-20min, IO:DataFileRead wait events).

The main culprit identified via Datadog APM is the `DELETE .../outdated` endpoint (`deleteOutdatedRecords`): **57.8k requests, 4.43s p95 latency spiking to 24s, 0.4% error rate**

The issue is structural:

1. **Single transaction wrapping all batches**: for a large customer with 100k outdated records, that's 20 batches of 5000 rows all within one transaction. The DB connection is held for the entire 4-24s duration, and all index updates (4 indexes per row) generate sustained IO/WAL pressure that competes with normal sync operations.
2. **Large batch size (5000)** — amplifies all of the above per batch.

Other things to consider:
1. **No advisory lock**: both `upsert()` and `deleteRecords()` acquire `pg_advisory_xact_lock`, this function did not. As we are just removing old records, Im not sure if this is necessary tbh
2. **No deadlock retry**: copied `upsert()` which retries on PostgreSQL deadlock errors (`40P01`), a deadlock here would fail the entire operation (once again, not critical and not something I've seen happen but it would fail the sync)

### Fix

**`packages/records/lib/models/records.ts` `deleteOutdatedRecords`:**

- Move the transaction inside the batch loop: each batch of records is now its own independent transaction that acquires the lock, updates rows, updates counts, and commits. This releases the DB connection and all locks between batches, allowing other queries to interleave naturally.
- Add `pg_advisory_xact_lock` using the same `newLockId(connectionId, model)` as `upsert()` and `deleteRecords()`, ensuring clean serialization.
- Add deadlock retry (3 attempts, 500ms delay) matching the `upsert()` pattern.
- Reduce default batch size from 5000 to 1000.

---

This PR updates `deleteOutdatedRecords` to process each batch in its own transaction, reducing long-lived locks and connection usage during large deletions. It also adds per-batch advisory locking and deadlock retry handling while lowering the default batch size.

---

### DIFF
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -891,7 +891,7 @@ export async function deleteOutdatedRecords({
     connectionId,
     model,
     generation,
-    batchSize = 5000
+    batchSize = 1000
 }: {
     environmentId: number;
     connectionId: number;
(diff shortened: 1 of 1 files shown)


packages/records/lib/models/records.ts
```diff
@@ -906,67 +906,89 @@ export async function deleteOutdatedRecords({
...
-                            .from(RECORDS_TABLE)
+        const deletedIds: string[] = [];
+        let hasMore = true;
+        while (hasMore) {
+            const batchResult = await retry(
+                () => {
+                    return db.transaction(async (trx) => {
+                        const now = trx.fn.now(6);
+                        // Lock to prevent concurrent modifications with upserts and deletes
+                        await trx.raw(`SELECT pg_advisory_xact_lock(?) as lock_records_outdated`, [newLockId(connectionId, model)]);
```
agusayerza, line 923: Unsure if the lock is really needed to delete stale records

packages/records/lib/models/records.ts
```diff
@@ -905,73 +905,98 @@ export async function deleteOutdatedRecords({
...
+                {
+                    maxAttempts: 3,
+                    delayMs: 500,
+                    retryOnError: (err) => {
+                        if ('code' in err) {
+                            const errorCode = (err as { code: string }).code;
+                            return errorCode === '40P01'; // deadlock_detected
+                        }
+                        return false;
+                    }
```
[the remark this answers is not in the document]
agusayerza: This is the same code as on `upsert`

packages/records/lib/models/records.ts
```diff
@@ -905,73 +905,98 @@ export async function deleteOutdatedRecords({
...
+        }
+        return Ok(deletedIds);
     } catch (err) {
         const e = new Error(`Failed to mark previous generation records as deleted for connection ${connectionId}, model ${model}, generation ${generation}`, {
             cause: err
         });
         span.setTag('error', err);
+        if (deletedIds.length > 0) {
+            logger.error(`${e.message}. Partial progress: ${deletedIds.length} records deleted before failure.`);
+            return Ok(deletedIds);
```
TBonnin: that's problematic. Sync will be considered successful and webhook sent to customer accordingly even though some of the records that are outdated haven't been deleted.

packages/records/lib/models/records.ts
```diff
@@ -905,73 +905,98 @@ export async function deleteOutdatedRecords({
...
+
+        if (partition) {
+            span.setTag('nango.partition', partition);
+        }
+        return Ok(deletedIds);
     } catch (err) {
         const e = new Error(`Failed to mark previous generation records as deleted for connection ${connectionId}, model ${model}, generation ${generation}`, {
             cause: err
         });
         span.setTag('error', err);
```
TBonnin: I realize here we are passing `err` instead of `e`

TBonnin (approved):
deletion is not atomic anymore. Is it worth adding a comment to explain the tradeoff?

PR #5657, opened 2026-03-17
feat: add ability to download deployed functions in debug mode

kaposke:
- Migrates old download endpoint to new format.
- Changes route and simplifies parameters.
- It only works for deployed functions
- Conditionally render download icon besides toggle switch when in debug mode (when impersonating, and function is deployed).
- It downloads the typescript file, not the bundled js.

image

---

It introduces a new v1 GET handler that validates parameters and resolves the provider config key before zipping artifacts, and adds a download button with client-side helper and basic toast feedback.

---

### DIFF
--- a/packages/server/lib/controllers/v1/flow/getDownload.ts
+++ b/packages/server/lib/controllers/v1/flow/getDownload.ts
@@ -5,7 +5,7 @@ import { requireEmptyQuery, zodErrorToHTTP } from '@nangohq/utils';
 
 import { asyncWrapper } from '../../../utils/asyncWrapper.js';
 
-import type { PostFlowDownload } from '@nangohq/types';
+import type { GetFlowDownload } from '@nangohq/types';
 
 export const validationParams = z
     .object({
@@ -13,7 +13,7 @@ export const validationParams = z
     })
     .strict();
 
-export const postFlowDownload = asyncWrapper<PostFlowDownload>(async (req, res) => {
+export const getFlowDownload = asyncWrapper<GetFlowDownload>(async (req, res) => {
     const emptyQuery = requireEmptyQuery(req, { withEnv: true });
     if (emptyQuery) {
         res.status(400).send({ error: { code: 'invalid_query_params', errors: zodErrorToHTTP(emptyQuery.error) } });
--- a/packages/server/lib/routes.private.ts
+++ b/packages/server/lib/routes.private.ts
@@ -248,10 +248,10 @@ web.route('/sync/command').post(
 );
 web.route('/flows/pre-built/deploy').post(webAuth, can({ action: 'update', resource: 'flow', scopedBy: envScope }), postPreBuiltDeploy);
 web.route('/flows/pre-built/upgrade').put(webAuth, can({ action: 'update', resource: 'flow', scopedBy: envScope }), putUpgradePreBuilt);
-web.route('/flow/download').post(webAuth, flowController.downloadFlow.bind(flowController));
 web.route('/flows/:id/disable').patch(webAuth, can({ action: 'update', resource: 'flow', scopedBy: envScope }), patchFlowDisable);
 web.route('/flows/:id/enable').patch(webAuth, can({ action: 'update', resource: 'flow', scopedBy: envScope }), patchFlowEnable);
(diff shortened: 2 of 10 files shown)


packages/server/lib/controllers/v1/flow/getDownload.ts
```diff
@@ -0,0 +1,47 @@
...
+import { configService, getSyncConfigById, remoteFileService } from '@nangohq/shared';
+import { requireEmptyQuery, zodErrorToHTTP } from '@nangohq/utils';
+
+import { asyncWrapper } from '../../../utils/asyncWrapper.js';
+
+import type { PostFlowDownload } from '@nangohq/types';
+
+export const validationParams = z
+    .object({
+        id: z.coerce.number().positive()
```
TBonnin, line 12: what about `syncConfigId`? it is not immediately clear which id we are talking about here

packages/shared/lib/services/config.service.ts
```diff
@@ -34,6 +34,16 @@ class ConfigService {
         return result.id;
     }
 
+    async getProviderConfigKeyById(environment_id: number, id: number): Promise<string | null> {
```
TBonnin: let's have all new function return a `Result`

TBonnin (commented):
> It downloads the typescript file, not the bundled js.

I would really prefer to get the bundled js. Most of the involved function are using some imported logic. Only having the ts entrypoint will be useless in those cases

packages/server/lib/controllers/v1/flow/getDownload.ts
kaposke: Should probably leave it as id to keep consistency with other flow routes (`/flows/:id/download`). I'll rename it when destructuring thought to make it more clear

TBonnin (approved):

packages/server/lib/routes.private.ts
```diff
@@ -216,10 +217,10 @@ web.route('/sync').get(webAuth, syncController.getSyncsByParams.bind(syncControl
 web.route('/sync/command').post(webAuth, syncController.syncCommand.bind(syncController));
 web.route('/flows/pre-built/deploy').post(webAuth, postPreBuiltDeploy);
 web.route('/flows/pre-built/upgrade').put(webAuth, putUpgradePreBuilt);
-web.route('/flow/download').post(webAuth, flowController.downloadFlow.bind(flowController));
 web.route('/flows/:id/disable').patch(webAuth, patchFlowDisable);
 web.route('/flows/:id/enable').patch(webAuth, patchFlowEnable);
 web.route('/flows/:id/frequency').patch(webAuth, patchFlowFrequency);
+web.route('/flows/:id/download').post(webAuth, postFlowDownload);
```
TBonnin: do you know why it was a POST. Looks more like a GET to me?

packages/webapp/src/pages/Integrations/providerConfigKey/Functions/One.tsx
```diff
@@ -73,6 +75,13 @@ export const FunctionsOne: React.FC = () => {
 
     const isLoading = integrationLoading || flowsLoading;
 
+    const downloadCode = useCallback(async () => {
+        if (!func || !func.enabled || !func.id) {
+            return;
+        }
+        await apiFlowDownload(env, { id: func.id }, func.name);
```
TBonnin: what happen if `apiFlowDownload` fails?

packages/server/lib/routes.private.ts
kaposke: You're right. It previously took some info in the body. But after my simplification it should definitely be a GET

packages/webapp/src/pages/Integrations/providerConfigKey/Functions/One.tsx
kaposke: I didn't try to polish the UX too much since this is internal. But I can probably add some basic error handling

kaposke: Added some basic feedback, much better

marcindobry (approved):

PR #5663, opened 2026-03-18
feat(permissions): NAN-2184 implements team/invite role managment

pfreixes:
This PR implements some of the leftovers that were missing for RBAC adding all required backend changes for implementing  the team/invite role management.
 

---

**Add team role management and role-aware invitations**

This PR introduces backend support for team role management and role-aware invitations. It adds a new `PATCH /api/v1/team/users/:id` endpoint for updating user roles with validation and self-demotion protection, and extends invitation creation to accept an optional role that is persisted and applied when the invite is accepted.

It also centralizes role definitions via a shared `roles` constant, updates permissions/deny maps to include team member updates, wires the new route, and adds integration/authz tests covering role updates, invite roles, and access restrictions for non-admin roles.

---

### DIFF
--- a/packages/server/lib/controllers/v1/invite/acceptInvite.ts
+++ b/packages/server/lib/controllers/v1/invite/acceptInvite.ts
@@ -37,7 +37,7 @@ export const acceptInvite = asyncWrapper<AcceptInvite>(async (req, res) => {
     }
 
     await acceptInvitation(data.id);
-    const updated = await userService.update({ id: user.id, account_id: invitation.account_id });
+    const updated = await userService.update({ id: user.id, account_id: invitation.account_id, role: invitation.role });
     if (!updated) {
         res.status(500).send({ error: { code: 'server_error', message: 'failed to update user team' } });
         return;
--- a/packages/utils/lib/roles.ts
+++ b/packages/utils/lib/roles.ts
@@ -1,0 +1,3 @@
+import type { Role } from '@nangohq/types';
+
+export const roles = ['administrator', 'production_support', 'development_full_access'] as const satisfies readonly Role[];
--- a/packages/server/lib/authz/authz.integration.test.ts
+++ b/packages/server/lib/authz/authz.integration.test.ts
@@ -171,6 +171,67 @@ describe('authz integration', () => {
             expect(res.res.status).toBe(403);
         });
 
+        it('should deny DELETE invite', async () => {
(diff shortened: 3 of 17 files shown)


packages/server/lib/controllers/v1/invite/acceptInvite.ts
```diff
@@ -37,7 +37,7 @@ export const acceptInvite = asyncWrapper<AcceptInvite>(async (req, res) => {
     }
 
     await acceptInvitation(data.id);
-    const updated = await userService.update({ id: user.id, account_id: invitation.account_id });
+    const updated = await userService.update({ id: user.id, account_id: invitation.account_id, role: invitation.role });
```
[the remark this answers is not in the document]
pfreixes: The [column](https://github.com/NangoHQ/nango/blob/master/packages/database/lib/migrations/20260313120001_add_role_to_invitations.cjs#L9) is not nullable and it defaults to `administrator` Today if no value is provided.

kaposke (approved):

packages/utils/lib/roles.ts
```diff
@@ -0,0 +1,3 @@
+import type { Role } from '@nangohq/types';
+
+export const roles = ['administrator', 'production_support', 'development_full_access'] as const satisfies readonly Role[];
```
TBonnin, line 3: why do you need this?

can we use simply `satisfies Role[]` in the zod schema above?

pfreixes: We are using it in three [1] different places, so thought that would be worth it to not repeat the same all the time and have one unique place where we declare this array.

I would have liked to use directly the type declared in our types, but if Im not mistaken the usage `emitDeclarationOnly: true` was preventing to have the array of values in runtime.

If you have any suggestion here I would like to see how we can do this easier!

[1] https://github.com/NangoHQ/nango/pull/5663/changes#diff-fa0765d02a3b777c95489cbd89f0e9143f46092f97b687416ad59197a9706c18R22
[2] https://github.com/NangoHQ/nango/pull/5663/changes#diff-c717768bc7782d9d97bd7773ee8fd16837d49f4d27842511b768c5bc3435c279R16
[3] https://github.com/NangoHQ/nango/pull/5663/changes#diff-68112fbb99bcefed6f0baec93c60069a23049c80b9b1fdceb936b79d8853acc3R18

PR #5668, opened 2026-03-18
feat(proxy): add a flag in the proxyConfig to enable forwarding header on redirect

hassan254-prog:

- This feature was originally added to support dayforce, which required authorization headers to be forwarded during redirects. We now provide a flag in `proxyConfig` that allows customers to disable header forwarding for redirects.

- Use case: In hibob, when users request file downloads, the API returns a HiBob download URL that issues a HTTP redirect to an amazon s3 signed url. Passing the Authorization header to this redirected URL results in an error, as signed urls are pre-authorized and do not accept additional authorization headers.

---

**Add `forwardHeadersOnRedirect` proxy flag with metrics and provider support**

This PR adds a configurable `forwardHeadersOnRedirect` option to proxy configuration, allowing callers to disable forwarding headers when `axios` follows redirects while preserving the existing default behavior. The flag is wired through public proxy headers, internal proxy configuration construction, `axios` setup, and type definitions, with a new `PROXY_REDIRECT` metric emitted on redirects.

Provider schema validation and provider configs are extended to support a provider-level default (`forward_headers_on_redirect`), and SDK/CLI surfaces are updated to send/ignore the new header. Unit tests and snapshots are adjusted to reflect the new configuration flow and `axios` behavior.

---

### DIFF
--- a/packages/node-client/lib/index.ts
+++ b/packages/node-client/lib/index.ts
@@ -1090,7 +1090,17 @@ export class Nango {
 
         validateProxyConfiguration(config);
 
-        const { providerConfigKey, connectionId, method, retries, headers: customHeaders, baseUrlOverride, decompress, retryOn } = config;
+        const {
+            providerConfigKey,
+            connectionId,
+            method,
+            retries,
+            headers: customHeaders,
+            baseUrlOverride,
+            decompress,
+            retryOn,
+            disableHeaderForwarding
+        } = config;
 
         let url = `${this.serverUrl}/proxy${config.endpoint[0] === '/' ? '' : '/'}${config.endpoint}`;
 
@@ -1124,6 +1134,10 @@ export class Nango {
 
         if (retryOn) {
             headers['Retry-On'] = retryOn.join(',');
+        }
+
+        if (disableHeaderForwarding) {
+            headers['Disable-Header-Forwarding'] = true;
         }
 
         const options: AxiosRequestConfig = {
--- a/packages/node-client/lib/index.ts
+++ b/packages/node-client/lib/index.ts
@@ -1137,7 +1137,7 @@ export class Nango {
         }
 
         if (enableHeaderForwarding !== undefined) {
-            headers['Enable-Header-Forwarding'] = enableHeaderForwarding;
+            headers['Enable-Header-Forwarding'] = String(enableHeaderForwarding);
         }
 
         const options: AxiosRequestConfig = {
--- a/packages/node-client/lib/index.ts
+++ b/packages/node-client/lib/index.ts
@@ -1099,7 +1099,7 @@ export class Nango {
             baseUrlOverride,
             decompress,
             retryOn,
-            disableHeaderForwarding
+            enableHeaderForwarding
         } = config;
 
         let url = `${this.serverUrl}/proxy${config.endpoint[0] === '/' ? '' : '/'}${config.endpoint}`;
@@ -1136,8 +1136,8 @@ export class Nango {
             headers['Retry-On'] = retryOn.join(',');
         }
 
-        if (disableHeaderForwarding) {
-            headers['Disable-Header-Forwarding'] = true;
+        if (enableHeaderForwarding) {
+            headers['Enable-Header-Forwarding'] = true;
         }
 
         const options: AxiosRequestConfig = {
--- a/packages/node-client/lib/index.ts
+++ b/packages/node-client/lib/index.ts
@@ -1136,8 +1136,8 @@ export class Nango {
             headers['Retry-On'] = retryOn.join(',');
         }
 
-        if (enableHeaderForwarding) {
-            headers['Enable-Header-Forwarding'] = true;
+        if (enableHeaderForwarding !== undefined) {
+            headers['Enable-Header-Forwarding'] = enableHeaderForwarding;
         }
 
         const options: AxiosRequestConfig = {
--- a/packages/node-client/lib/index.ts
+++ b/packages/node-client/lib/index.ts
@@ -1099,7 +1099,7 @@ export class Nango {
             baseUrlOverride,
             decompress,
             retryOn,
-            enableHeaderForwarding
+            forwardHeadersOnRedirect
         } = config;
 
         let url = `${this.serverUrl}/proxy${config.endpoint[0] === '/' ? '' : '/'}${config.endpoint}`;
@@ -1136,8 +1136,8 @@ export class Nango {
             headers['Retry-On'] = retryOn.join(',');
         }
 
-        if (enableHeaderForwarding !== undefined) {
-            headers['Enable-Header-Forwarding'] = String(enableHeaderForwarding);
+        if (forwardHeadersOnRedirect !== undefined) {
+            headers['Forward-Headers-On-Redirect'] = forwardHeadersOnRedirect;
         }
 
         const options: AxiosRequestConfig = {
--- a/packages/server/lib/controllers/proxy/allProxy.ts
+++ b/packages/server/lib/controllers/proxy/allProxy.ts
@@ -47,7 +47,8 @@ const schemaHeaders = z.object({
         .string()
         .regex(/^\d+(,\d+)*$/)
         .optional(),
-    'enable-header-forwarding': z.enum(['true', 'false']).default('false'),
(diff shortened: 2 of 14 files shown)


TBonnin:
> This feature was originally added to support dayforce, which required authorization headers to be forwarded during redirects

Is it not a security risk to always forward headers. Should it only be enabled when domain is the same?

hassan254-prog:
> > This feature was originally added to support dayforce, which required authorization headers to be forwarded during redirects
> 
> Is it not a security risk to always forward headers. Should it only be enabled when domain is the same?



> > This feature was originally added to support dayforce, which required authorization headers to be forwarded during redirects
> 
> Is it not a security risk to always forward headers. Should it only be enabled when domain is the same?

Yes it is, should we disable for all and only enable it for dayforce and ensure this only works with same hostname? I feel like this would introduce a breaking change if we have other providers relying on the same, and dayforce not honoring same domain redirect, wdyt?

TBonnin:
it is indeed a breaking change. Do we know what kind of redirect dayforce is doing? From a security point of view, it is a bit dangerous to blindly forward the headers. That's why I imagine S3 is discarding requests with unknown headers.

marcindobry (changes requested):
This might sound nit-picky, but IMO a nicer "interface" would be to call it `enableHeaderForwarding` instead.

Header forwarding should be an opt-in functionality, by default it should be disabled, so it's weird that to opt in, it is required to pass in `false` as a flag override

hassan254-prog:
> it is indeed a breaking change. Do we know what kind of redirect dayforce is doing? From a security point of view, it is a bit dangerous to blindly forward the headers. That's why I imagine S3 is discarding requests with unknown headers.

@TBonnin  The docs mention that dayforce issues a HTTP redirect for every request for both [authentication](https://help.dayforce.com/r/documents/Dayforce-RESTful-Web-Services-Developer-Guide/Authentication) and api [calls](https://help.dayforce.com/r/documents/Dayforce-RESTful-Web-Services-Developer-Guide/Determine-Which-URL-to-Use), but they don’t specify the type of redirect or the target urls. I agree with @marcindobry  we can disable automatic forwarding and enable it only for dayforce within the provider configurations, since it’s the only one we are aware of. I’ve updated this in the latest [commit](https://github.com/NangoHQ/nango/pull/5668/commits/18be483c420e58e585bafde9691c94073993faa7).

packages/node-client/lib/index.ts
```diff
@@ -1126,6 +1136,10 @@ export class Nango {
             headers['Retry-On'] = retryOn.join(',');
         }
 
+        if (enableHeaderForwarding !== undefined) {
+            headers['Enable-Header-Forwarding'] = enableHeaderForwarding;
```
TBonnin: does it need to be a `String()`. maybe it is automatically encoded by axios but better to be explicit?

packages/shared/lib/services/proxy/utils.ts
```diff
@@ -71,16 +71,20 @@ export function getAxiosConfiguration({
...
         method: proxyConfig.method,
         url,
-        headers,
-        beforeRedirect: (options: Record<string, any>) => {
+        headers
+    };
+
+    const shouldForward = proxyConfig.enableHeaderForwarding ?? proxyConfig.provider.proxy?.enable_header_forwarding;
+    if (shouldForward) {
+        axiosConfig.beforeRedirect = (options: Record<string, any>) => {
```
TBonnin: that's a breaking change. Do we need to warn some customers?

packages/server/lib/controllers/proxy/allProxy.ts
```diff
@@ -47,6 +47,7 @@ const schemaHeaders = z.object({
         .string()
         .regex(/^\d+(,\d+)*$/)
         .optional(),
+    'enable-header-forwarding': z.enum(['true', 'false']).optional(),
```
TBonnin: can we default to false so we don't have to handle `undefined` after validation?

packages/node-client/lib/index.ts
hassan254-prog: Agree, updated with the latest [commit](https://github.com/NangoHQ/nango/pull/5668/commits/86dcb3849af68d9ccd26f897559f8b56e478616f).

packages/server/lib/controllers/proxy/allProxy.ts
hassan254-prog: Updated with the latest [commit](https://github.com/NangoHQ/nango/pull/5668/commits/86dcb3849af68d9ccd26f897559f8b56e478616f).

TBonnin (approved):
I don't know if any customer would expect the header forwarding to work by default but from the code point of view it lgtm

packages/shared/lib/services/proxy/utils.ts
hassan254-prog: Maybe you can advise us on how you would like us to handle this, or how you would have tackled it.

TBonnin: I don't know much about the background of the forward headers feature. How many of our customers rely on it? if it is a low number we can try to talk to them and ask then to explicitly request headers forwarding before we merge this change. Otherwise we might need to make an announcement, give customers a few weeks of grace period before we merge this breaking change. Last resort option is to release the breaking change and react/support customers affected by it. That's a product decision @bastienbeurier , not a technical one

bastienbeurier: I don't understand the scope of this breaking change. Is there any way to measure who would be affected?

hassan254-prog: There’s no way to determine how many customers are currently using this. This was originally added specifically for dayforce, since it requires forwarding headers during redirects. However, other providers might also rely on this behavior.
To avoid introducing a breaking change, I propose we default `shouldForward` to `true`. At the same time, we can add a metric to track usage over a period of time and identify which providers rely on header forwarding during redirects. Once we have a complete list, we can explicitly enable this behavior in the provider configs and then remove the default true fallback. 

Currently I only know of dayforce who perform redirects for their api calls.

marcindobry (approved):

packages/shared/lib/services/proxy/utils.ts
TBonnin: not ideal but ok. let's default to `true` if the option is not set in order to avoid breaking existing customers. Customers who need to make sure headers are not forwarded on redirects can explicitly set it to false.
@hassan254-prog is the headers forwarding only concerns redirect? if yes being explicit about that in the naming would be better I think: `forwardHeadersOnRedirect: boolean` instead of `enableHeaderForwarding`

hassan254-prog: Done. It now defaults to `true`, which matches the previous behavior. I’ve also added metrics for now and updated the naming, since only headers are being forwarded.

packages/shared/lib/services/proxy/utils.ts
```diff
@@ -71,8 +71,14 @@ export function getAxiosConfiguration({
...
         url,
-        headers,
-        beforeRedirect: (options: Record<string, any>) => {
+        headers
+    };
+
+    // TODO: change default to false after removing the metric below
+    const shouldForward = proxyConfig.forwardHeadersOnRedirect ?? proxyConfig.provider.proxy?.forward_headers_on_redirect ?? true;
+    axiosConfig.beforeRedirect = (options: Record<string, any>) => {
+        metrics.increment(metrics.Types.PROXY_REDIRECT, 1, { provider: proxyConfig.providerName, forwardHeadersOnRedirect: String(shouldForward) });
```
TBonnin: I am not sure what we are trying to meter here. Isn't shouldForward always gonna be true unless customer specify `forwardHeadersOnRedirect=false`?

hassan254-prog: The `beforeRedirect` callback is only invoked by axios when the upstream server responds with a redirect. So the metric only increments when a redirect actually happens, it's completely silent for normal requests that get a direct response, this will help us identify providers that perform redirects so that we can now configure them properly in the providers config when we restore the default false behaviour.

TBonnin: ok. please put a reminder to review the data in a few weeks and remove the metric. Thank you

PR #5676, opened 2026-03-19
feat(syncs): Expose listRecords for use in syncs

rbwest:
listRecords needs to be exposed in sync code to allow for [two-layer sync for MobileFirst](https://linear.app/nango/issue/NAN-4939/themobilefirstco-attio-real-time-syncing) to reduce sync load

[NAN-5073: Expose listRecords for use in syncs](https://linear.app/nango/issue/NAN-5073/expose-listrecords-for-use-in-syncs)

In a sample integrations repo:
- Point towards local copy of CLI/runner SDK
- Create sync that uses newly exposed nango.listRecords function
- Compile and dry run sync

---

**Expose `listRecords` async iterator in sync runners**

This PR introduces a new `listRecords` async iterator to sync execution paths, enabling scripts to iterate persisted records page-by-page. The change is wired through the runner SDK, CLI sync runner, type definitions, telemetry allowlists, and persist client request handling, with abort checks during iteration and batching.

Supporting updates include a shared `fetchRecordsPage` helper for `getRecordsByIds`, optional `limit` propagation to the persist API, parser/compile validations, mock updates, and expanded unit tests covering abort behavior and multi-page cursor flows.

---


packages/cli/lib/services/parser.service.ts
```diff
@@ -89,11 +90,12 @@ class ParserService {
         const callsProxy = ['proxy', 'get', 'post', 'put', 'patch', 'delete'];
         const callsBatchingRecords = ['batchSave', 'batchDelete', 'batchUpdate'];
         const callsReferencingModelsToCheck = callsBatchingRecords.concat(
-            'setMergingStrategy',
             'deleteRecordsFromPreviousExecutions',
-            'trackDeletesStart',
+            'getRecordsByIds',
+            'listRecords',
```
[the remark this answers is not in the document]
rbwest: Changed the call signature to have model be the final argument like all other functions. Will update validation if the team requests it instead. Assuming the model is the final argument does feel a bit hacky 🤷

packages/runner/lib/clients/persist.ts
```diff
@@ -282,21 +282,24 @@ export class PersistClient {
...
     }): Promise<Result<GetRecordsSuccess>> {
         const res = await this.fetch<GetRecordsSuccess>({
             method: 'GET',
             path: `/environment/${environmentId}/connection/${nangoConnectionId}/records`,
             params: {
                 model,
                 ...(cursor && { cursor }),
-                ...(externalIds && { externalIds })
+                ...(externalIds && { externalIds }),
+                ...(limit && { limit: String(limit) })
```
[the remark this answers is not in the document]
rbwest: Updated

packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
```diff
@@ -481,6 +481,11 @@ export declare class NangoSync<TCheckpoint = Checkpoint> extends NangoAction {
     trackDeletesStart(model: string): Promise<void>;
     trackDeletesEnd(model: string): Promise<{ deletedKeys: string[] }>;
     getRecordsByIds<K = string | number, T = any>(ids: K[], model: string): Promise<Map<K, T>>;
+    listRecords<T extends object = any>(
+        cursor: string | undefined,
+        limit: number | undefined,
+        model: string
```
TBonnin: I'd pass an object like the node client equivalent function.
`limit` should have a default value when not provided. Maybe quite a small number to discourage any heavy processing

packages/runner/lib/sdk/sdk.ts
```diff
@@ -613,6 +613,31 @@ export class NangoSyncRunner extends NangoSyncBase<never, never, ZodCheckpoint>
...
+            cursor: options?.cursor,
+            externalIds: options?.externalIds,
+            limit: options?.limit
+        });
+
+        if (res.isErr()) {
+            throw res.error;
+        }
+
+        const raw = res.unwrap();
```
TBonnin: unwrap is not to be used outside of tests.
since you have an isErr() guard you can get the value with `res.value`.

packages/runner/lib/sdk/sdk.ts
```diff
@@ -626,33 +651,46 @@ export class NangoSyncRunner extends NangoSyncBase<never, never, ZodCheckpoint>
...
-            });
-
-            if (res.isErr()) {
-                throw res.error;
+            const pageOptions: { cursor?: string; externalIds: string[] } = {
+                externalIds: Array.from(externalIdMap.keys())
+            };
+            if (cursor !== undefined) {
+                pageOptions.cursor = cursor;
             }
```
TBonnin: you can do 
```
const pageOptions = {
   externalIds: Array.from(externalIdMap.keys()),
   ...(cursor ? { cursor } : {} )
}
```

packages/runner/lib/sdk/sdk.ts
```diff
@@ -626,33 +651,46 @@ export class NangoSyncRunner extends NangoSyncBase<never, never, ZodCheckpoint>
...
 
-            const { nextCursor, records } = res.unwrap();
-            cursor = nextCursor;
+            const { records, next_cursor } = await this.fetchRecordsPage<T>(model, pageOptions);
+            cursor = next_cursor ?? undefined;
 
             for (const record of records) {
-                const stringId = String(record.id);
+                const stringId = String((record as Record<string, unknown>)['id']);
```
TBonnin: do you need the casting?

packages/runner/lib/sdk/sdk.ts
```diff
@@ -626,33 +651,46 @@ export class NangoSyncRunner extends NangoSyncBase<never, never, ZodCheckpoint>
...
+    ): Promise<{ records: T[]; next_cursor: string | null }> {
+        this.throwIfAbortedOrKilled();
+
+        const pageOptions: { cursor?: string; limit?: number } = {};
+        if (cursor !== undefined) {
+            pageOptions.cursor = cursor;
+        }
+        if (limit !== undefined) {
+            pageOptions.limit = limit;
+        }
```
TBonnin, line 678: same comment about using the spread operator.

packages/cli/lib/services/parser.service.ts
```diff
@@ -54,27 +54,28 @@ class ParserService {
...
+            'get',
+            'getConnection',
+            'getEnvironmentVariables',
             'getFieldMapping',
-            'setFieldMapping',
             'getMetadata',
-            'setMetadata',
-            'proxy',
-            'get',
+            'listRecords', // 1. Allow await nango.listRecords() at compile time
```
TBonnin: I think listRecords needs to be added to `disallowedActionCalls.` below

packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
```diff
@@ -481,6 +481,11 @@ export declare class NangoSync<TCheckpoint = Checkpoint> extends NangoAction {
...
+    ): Promise<{ records: T[]; next_cursor: string | null }>;
```
TBonnin: slightly unrelated question. Should we deprecate `getRecordsByIds` and make `ids` and optional field of listRecords. Since they are basically the same thing under the hood.

packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
rbwest: I originally had it that way, but there was a check in `parser.service.ts` that looks for the model arg in the last position per this [AI note:](https://github.com/NangoHQ/nango/pull/5676/changes/BASE..14a44afb292b41db89aa7242e7393b84dd53217f#r2962323618)

`const modelArg = args[args.length - 1]; `

Want me to revert the signature to: 

`public async listRecords<T = any>(
        model: string,
        options?: {
            cursor?: string;
            limit?:  number;
        }
)`

And update the parser to allow this?

packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
rbwest: Yeah, I think that made sense, but wasn't sure what the process for deprecating looks like. Is there any action we want to take now, like an @deprecated tag on the function?

packages/cli/lib/services/parser.service.ts
rbwest: Updated, check now!

packages/runner/lib/sdk/sdk.ts
rbwest: Updated, check now!

packages/runner/lib/sdk/sdk.ts
rbwest: It was because records were getting type `T` from the cast in `fetchRecordsPage` instead of carrying the type `NangoRecord` from the call to the persist client.

I've adjusted the types to be more explicit (from `T[]` to `NangoRecord<T>[]` and `T = any` to `T=Record<string, any>`) so it's no longer required. Let me know if I'm missing some issue with the new types though.

packages/runner/lib/sdk/sdk.ts
rbwest: Updated, check now!

packages/runner/lib/sdk/sdk.ts
rbwest: Updated, check now!

packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
TBonnin: optional cursor/limit before the model looks weird. I think we should modify the parser to allow optional params after the model.

packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
TBonnin: yes a @deprecated tag to start with. don't worry in this PR you can do it later

packages/cli/lib/services/parser.service.ts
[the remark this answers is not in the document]
marcindobry: I must admit it's a weird convention. Also, it's the only required argument out of the 3 which feels weird to have to potentially have to call with (undefined, undefined, 'FooModel'). I'd prefer for it to be the first argument

packages/runner/lib/sdk/sdk.unit.test.ts
```diff
@@ -534,3 +534,52 @@ describe('getRecordsById', () => {
...
+    it('should return records and next_cursor from a single page', async () => {
+        const records = [
+            { id: '1', name: 'a' },
+            { id: '2', name: 'b' }
+        ];
+        const mockPersistClient = new PersistClient({ secretKey: '***' });
+        mockPersistClient.getRecords = vi.fn().mockResolvedValueOnce(Ok({ records, nextCursor: null }));
+
+        const nango = new NangoSyncRunner({ ...nangoProps }, { persistClient: mockPersistClient, locks });
+        const result = await nango.listRecords(undefined, undefined, 'SomeModel');
```
kaposke: Yeah I'll have to agree with the other comments, this signature doesn't look good

packages/cli/lib/services/parser.service.ts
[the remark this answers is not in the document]
rbwest: Per @kaposke, this class is actually deprecated and only used for the zero yaml migration script. I've reverted the call signature back to the orginal (`model: string, options?: {...}`), added an @deprecated comment to this file, and skipped updating the model arg validation.

packages/runner/lib/sdk/sdk.ts
```diff
@@ -626,33 +651,53 @@ export class NangoSyncRunner extends NangoSyncBase<never, never, ZodCheckpoint>
...
+    public async *listRecords<T extends Record<string, any>>(
+        model: string,
+        options?: {
+            cursor?: string;
+        }
+    ): AsyncGenerator<NangoRecord<T>> {
+        this.throwIfAbortedOrKilled();
+
+        let cursor: string | undefined = options?.cursor ?? undefined;
+        do {
```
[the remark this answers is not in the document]
rbwest: Updated.

packages/runner/lib/sdk/sdk.unit.test.ts
rbwest: Updated.

packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
```diff
@@ -481,11 +482,10 @@ export declare class NangoSync<TCheckpoint = Checkpoint> extends NangoAction {
...
     trackDeletesEnd(model: string): Promise<{ deletedKeys: string[] }>;
     getRecordsByIds<K = string | number, T = any>(ids: K[], model: string): Promise<Map<K, T>>;
-    listRecords<T extends object = any>(
-        cursor: string | undefined,
-        limit: number | undefined,
-        model: string
-    ): Promise<{ records: T[]; next_cursor: string | null }>;
+    listRecords<T extends Record<string, any> = Record<string, any>>(
+        model: string,
+        options?: { cursor?: string; }
```
TBonnin: let's keep the limit options

packages/runner/lib/sdk/sdk.ts
```diff
@@ -671,19 +671,31 @@ export class NangoSyncRunner extends NangoSyncBase<never, never, ZodCheckpoint>
...
+            cursor?: string;
+        }
+    ): AsyncGenerator<NangoRecord<T>> {
         this.throwIfAbortedOrKilled();
 
-        const pageOptions: { cursor?: string; limit?: number } = {
-            ...(cursor ? { cursor } : {}),
-            ...(limit ? { limit } : {})
-        };
+        let cursor: string | undefined = options?.cursor ?? undefined;
```
TBonnin: not sure I understand what this line is achieving?

packages/runner/lib/sdk/sdk.ts
```diff
@@ -671,19 +671,31 @@ export class NangoSyncRunner extends NangoSyncBase<never, never, ZodCheckpoint>
...
+        do {
+            const pageOptions: { cursor?: string } = {
+                ...(cursor ? { cursor } : {})
+            };
+            const { records, next_cursor } = await this.fetchRecordsPage<T>(model, pageOptions);
+
+            for (const record of records) {
+                yield record;
+            }
```
TBonnin, line 690: should the function generate records or page of records? wdyt?
how is `listRecords` going to be used in the webhook debouncing use case for instance?

packages/runner/lib/sdk/sdk.ts
```diff
@@ -671,19 +671,31 @@ export class NangoSyncRunner extends NangoSyncBase<never, never, ZodCheckpoint>
...
+
+            if (!next_cursor) {
+                break;
+            }
```
TBonnin: is this break necessary. isn't the while condition be enough to exit the loop?

packages/cli/lib/services/__snapshots__/model.service.unit.test.ts.snap
rbwest: Hmm... I was going back and forth on this yesterday and removed it after discussing with Marcin and Agustin. 

Can you explain how you would expect the limit to apply when returning an iterator? If there is some limit on the number of rows the iterator will yield, then we need to provide a nextCursor for the user to know to pickup the remaining records. If we return nextCursor, the user will have to paginate via cursor to get all records. At that point, what's the point having an iterator vs just paginating like it was originally?

packages/runner/lib/sdk/sdk.ts
rbwest: I thought returning a single record at a time made it easier to use the iterator. The webhook debouncing use case will want to get all records, de-dupe ids into a map, then fetch the data for every unique id in the map. I think this could work fine returning individual records or pages, but individual records felt cleaner.

packages/runner/lib/sdk/sdk.ts
rbwest: Just storing cursor from the param. `?? undefined` was a rogue autocomplete I didn't catch I think. Removed, thanks!

packages/runner/lib/sdk/sdk.ts
rbwest: Yes, sorry. It's a relic from the original implementation of the loop. Changed it from while(true)...break to a do...while and forgot to remove. Fixed now.

TBonnin (approved):

packages/runner/lib/sdk/sdk.ts
```diff
@@ -613,9 +614,31 @@ export class NangoSyncRunner extends NangoSyncBase<never, never, ZodCheckpoint>
...
+        });
 
+        if (res.isErr()) {
+            throw res.error;
+        }
+
+        return { records: res.value.records as NangoRecord<T>[], next_cursor: res.value.nextCursor ?? null };
+    }
+
+    public async getRecordsByIds<K = string | number, T extends Record<string, any> = Record<string, any>>(ids: K[], model: string): Promise<Map<K, T>> {
```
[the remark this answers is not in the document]
rbwest: Updated.

packages/runner-sdk/models.d.ts
```diff
@@ -438,6 +438,12 @@ export declare class NangoSync<TCheckpoint = Checkpoint> extends NangoAction {
     trackDeletesStart(model: string): Promise<void>;
     trackDeletesEnd(model: string): Promise<{ deletedKeys: string[] }>;
     getRecordsByIds<K = string | number, T = any>(ids: K[], model: string): Promise<Map<K, T>>;
+    listRecords<T extends Record<string, any> = Record<string, any>>(
+        model: string,
+        options?: {
+            cursor?: string;
+        }
+    ): AsyncGenerator<NangoRecord<T>>;
```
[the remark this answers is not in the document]
rbwest: Fixed.

PR #5677, opened 2026-03-19
fix(orch): enforce max amount of queued tasks

TBonnin:
Enforcing an upper bound limit of tasks in CREATED tasks for a given group in order to protect the orchestrator from a sudden surge of tasks creation. 
We assume that if the cap is reached something else is going wrong and that tasks cannot be dequeued and processed fast enough. Hence discarding the tasks instead of cascading failures.
This is only a safeguard mechanism, not a strict limit.

 

---

The safeguard is configurable via an environment variable and is enforced during `tasks.create`, returning an error when the per-group `CREATED` task cap is reached.

---

### DIFF
--- a/packages/utils/lib/environment/parse.ts
+++ b/packages/utils/lib/environment/parse.ts
@@ -80,6 +80,7 @@ export const ENVS = z.object({
     ORCHESTRATOR_CLEANING_OLDER_THAN_DAYS: z.coerce.number().optional().default(5),
     ORCHESTRATOR_SCHEDULING_TICK_INTERVAL_MS: z.coerce.number().optional().default(100),
     ORCHESTRATOR_TASK_CREATED_EVENT_DEBOUNCE_MS: z.coerce.number().optional().default(100),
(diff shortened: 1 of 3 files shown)


packages/utils/lib/environment/parse.ts
```diff
@@ -82,6 +80,7 @@ export const ENVS = z.object({
     ORCHESTRATOR_CLEANING_OLDER_THAN_DAYS: z.coerce.number().optional().default(5),
     ORCHESTRATOR_SCHEDULING_TICK_INTERVAL_MS: z.coerce.number().optional().default(100),
     ORCHESTRATOR_TASK_CREATED_EVENT_DEBOUNCE_MS: z.coerce.number().optional().default(100),
+    ORCHESTRATOR_TASK_CREATED_PER_GROUP_COUNT_MAX: z.coerce.number().optional().default(10_000),
```
TBonnin, line 83: picked 10_000 arbitrarily. This is 20x the current concurrency limit. Let me know what you think?

rossmcewan (approved):

PR #5678, opened 2026-03-19
fix(sync): improve error handling and operation logging for sync commands

hassan254-prog:

**Fix some few issues with the various sync commands**
- Silent crash on unknown integration
- ConnectionId missing from operation details in the nango logs
- Incomplete error types 
- Swallowed error codes that returned `server_error` instead of forwarding the actual typed error
- Operation log contexts were never closed in error paths, leaving them stuck in a running state

---

It also formalizes typed error responses via the shared error handler and updates the sync API typings to align with the new error behavior.

---

### DIFF
--- a/packages/shared/lib/services/sync/manager.service.ts
+++ b/packages/shared/lib/services/sync/manager.service.ts
@@ -272,27 +272,30 @@ export class SyncManagerService {
         deleteRecords?: boolean;
     }): Promise<ServiceResponse<boolean>> {
         const provider = await configService.getProviderConfig(providerConfigKey, environment.id); // Todo: pass provider as argument as it's most likely already loaded
-        if (!provider) {
+        if (!provider || !provider.id) {
             return { success: false, error: new NangoError('unknown_provider_config'), response: false };
         }
 
         const account = (await accountService.getAccountFromEnvironment(environment.id))!; // Todo: pass account as argument as it's most likely already loaded
 
-        const logCtx = await logContextGetter.create(
-            { operation: { type: 'sync', action: syncCommandToOperation[command] } },
-            { account, environment, integration: { id: provider.id!, name: provider.unique_key, provider: provider.provider } }
-        );
+        let logCtx: Awaited<ReturnType<LogContextGetter['create']>>;
 
         if (connectionId) {
             const { success, error, response: connection } = await connectionService.getConnection(connectionId, providerConfigKey, environment.id);
 
             if (!success || !connection) {
-                void logCtx.error(error?.message || 'Connection not found', { error });
-                await logCtx.failed();
                 return { success: false, error, response: false };
             }
 
-            await logCtx.enrichOperation({ connectionId: connection.id, connectionName: connection.connection_id });
+            logCtx = await logContextGetter.create(
+                { operation: { type: 'sync', action: syncCommandToOperation[command] } },
+                {
+                    account,
+                    environment,
+                    integration: { id: provider.id, name: provider.unique_key, provider: provider.provider },
+                    connection: { id: connection.id, name: connection.connection_id }
+                }
+            );
 
             let syncs = syncIdentifiers;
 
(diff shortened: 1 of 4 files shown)


packages/shared/lib/services/sync/manager.service.ts
```diff
@@ -272,20 +272,28 @@ export class SyncManagerService {
...
         if (connectionId) {
             const { success, error, response: connection } = await connectionService.getConnection(connectionId, providerConfigKey, environment.id);
 
             if (!success || !connection) {
+                void logCtx.error(error?.message || 'Connection not found', { error });
+                await logCtx.failed();
                 return { success: false, error, response: false };
             }
 
+            await logCtx.enrichOperation({ connectionId: connection.id, connectionName: connection.connection_id });
```
pfreixes: Should we just add the connection id in the create step? not sure how much value provides here the connection name, Im just saying this bc this will create a new update to ES if Im not mistaken

pfreixes: Also seems that we have a function which makes two branches when the connection id is provided and when not, so we could do two different creation steps, and the one that uses the connection might first fetch if the connection exists, if it does not we just return the error without creating any logging at all.

This could be feasible if we do not need to have any tracing at all for these situations

pfreixes (commented):
Good catches!

Added only one comment! But might be worth it to have the expertise from @TBonnin or @rossmcewan for making sure that Im not missing anything here

packages/shared/lib/services/sync/manager.service.ts
TBonnin: can we just create the operation here, only after connection was retrieved?

packages/shared/lib/services/sync/manager.service.ts
```diff
@@ -299,7 +307,9 @@ export class SyncManagerService {
             for (const { syncName, syncVariant } of syncs) {
                 const sync = await getSync({ connectionId: connection.id, name: syncName, variant: syncVariant });
                 if (!sync) {
-                    throw new Error(`Sync "${syncName}" doesn't exists.`); // Todo: return this error instead of throwing
+                    void logCtx.error(`Sync "${syncName}" doesn't exist.`);
```
TBonnin: let's add syncVariant to the error

packages/shared/lib/services/sync/manager.service.ts
```diff
@@ -320,7 +330,8 @@ export class SyncManagerService {
 
             if (!syncs || syncs.length === 0) {
                 const error = new NangoError('no_syncs_found');
-
+                void logCtx.error('No syncs found for the provided params', { error });
```
TBonnin: `provided params` is vague. Can we explicitly log the syncIdentifiers?

packages/shared/lib/services/sync/manager.service.ts
```diff
@@ -272,20 +272,28 @@ export class SyncManagerService {
...
+        if (!provider) {
+            return { success: false, error: new NangoError('unknown_provider_config'), response: false };
+        }
+
         const account = (await accountService.getAccountFromEnvironment(environment.id))!; // Todo: pass account as argument as it's most likely already loaded
 
         const logCtx = await logContextGetter.create(
             { operation: { type: 'sync', action: syncCommandToOperation[command] } },
-            { account, environment, integration: { id: provider!.id!, name: provider!.unique_key, provider: provider!.provider } }
+            { account, environment, integration: { id: provider.id!, name: provider.unique_key, provider: provider.provider } }
```
TBonnin: we shouldn't need the `!` since there is an added `if` guard above

TBonnin (approved):

TBonnin (approved):
lgtm

packages/shared/lib/services/sync/manager.service.ts
hassan254-prog: Updated with the latest [commit](https://github.com/NangoHQ/nango/pull/5678/commits/660435d46d8620cb77fd67bd3350f71ac8296f03).

packages/shared/lib/services/sync/manager.service.ts
hassan254-prog: Updated with the latest [commit](https://github.com/NangoHQ/nango/pull/5678/commits/660435d46d8620cb77fd67bd3350f71ac8296f03).

packages/shared/lib/services/sync/manager.service.ts
hassan254-prog: Thanks 🙏 , I have updated this with the latest [commit](https://github.com/NangoHQ/nango/pull/5678/commits/660435d46d8620cb77fd67bd3350f71ac8296f03).

PR #5685, opened 2026-03-20
fix(connect-ui): restore client secret field for providers that use OAUTH2_CC and define custom credentials

hassan254-prog:

-  restore client secret field for providers that use OAUTH2_CC and define custom credentials i.e `adp`

---

This also clarifies that the form logic only makes `client_secret` optional for `private_key_jwt` providers, restoring the field for custom-credential `OAUTH2_CC` providers rather than hiding it when custom credentials omit `client_secret`.

---

### DIFF
--- a/packages/connect-ui/src/views/Go.tsx
+++ b/packages/connect-ui/src/views/Go.tsx
@@ -149,10 +149,7 @@ export const Go: React.FC = () => {
             if ((name === 'client_certificate' || name === 'client_private_key') && provider.require_client_certificate !== true) {
                 continue;
             }
-            if (
-                name === 'client_secret' &&
-                (provider.token_request_auth_method === 'private_key_jwt' || (provider.credentials && !('client_secret' in provider.credentials)))
-            ) {
+            if (name === 'client_secret' && provider.token_request_auth_method === 'private_key_jwt') {
                 baseForm.shape['client_secret'] = z.string().optional();
                 continue;
             }


packages/connect-ui/src/views/Go.tsx
```diff
@@ -149,10 +149,7 @@ export const Go: React.FC = () => {
             if ((name === 'client_certificate' || name === 'client_private_key') && provider.require_client_certificate !== true) {
                 continue;
             }
-            if (
-                name === 'client_secret' &&
-                (provider.token_request_auth_method === 'private_key_jwt' || (provider.credentials && !('client_secret' in provider.credentials)))
-            ) {
+            if (name === 'client_secret' && provider.token_request_auth_method === 'private_key_jwt') {
```
agusayerza, line 152: This would force the field to show up for integrations that don't need the `client_secret`, such as `okta-cc`. I think the right approach would be to add `client_secret` to adp's credential definition on `providers.yaml` instead.

I would also add test for this kind of conditional rendering (missed this on my initial PR)

agusayerza (changes requested):

packages/connect-ui/src/views/Go.tsx
hassan254-prog: This will skip okta-cc because it has the prop `token_request_auth_method === 'private_key_jwt'`. The [default](https://github.com/NangoHQ/nango/blob/e4797f9dfbda1dc989e30dd20f9d833dbc1ea73a/packages/connect-ui/src/views/Go.tsx#L75) behaviour is for all OAUTH2_CC to have the client_id and client_secret fields present, even if they are not listed in the `credentials.` within the providers.yaml since we have most providers without these.

agusayerza: You are right, my bad

agusayerza (approved):

PR #5688, opened 2026-03-20
chore: batch insert tasks

TBonnin:
Optimizing orchestrator to create tasks in batch. Useful when the scheduling daemon needs to create hundreds of tasks.

---

**Batch task creation with capped group tracking and scheduler updates**

This PR updates task creation to use a batch API for the scheduler and scheduler daemon, enabling bulk inserts with per-group caps and returning both created tasks and capped group keys. The scheduling flow now builds an array of task properties, calls the batch `tasks.create` API, logs capped group keys, and records the scheduled count via trace tags.

---

### DIFF
--- a/packages/scheduler/lib/daemons/scheduling/scheduling.daemon.ts
+++ b/packages/scheduler/lib/daemons/scheduling/scheduling.daemon.ts
@@ -56,7 +56,7 @@ export class SchedulingDaemon extends SchedulerDaemon {
                         if (getDueSchedules.isErr()) {
                             throw new Error(`Failed to get due schedules: ${stringifyError(getDueSchedules.error)}`);
                         } else {
-                            await tracer.trace('scheduler.scheduling.tasks_creation', async () => {
+                            await tracer.trace('scheduler.scheduling.tasks_creation', async (span) => {
                                 const now = new Date();
                                 const taskProps = getDueSchedules.value.map((schedule) => ({
                                     scheduleId: schedule.id,
@@ -94,6 +94,10 @@ export class SchedulingDaemon extends SchedulerDaemon {
                                 if (scheduleRes.isErr()) {
                                     logger.error(`Error updating schedules with last scheduled task: ${stringifyError(scheduleRes.error)}`);
                                 }
+
+                                span.addTags({
+                                    'scheduling.scheduled': scheduledTasks.length
+                                });
 
                                 // Notify the listeners about the scheduled tasks
                                 scheduledTasks.forEach((task) => this.onScheduling(task));
--- a/packages/scheduler/lib/daemons/scheduling/scheduling.daemon.ts
+++ b/packages/scheduler/lib/daemons/scheduling/scheduling.daemon.ts
@@ -76,13 +76,12 @@ export class SchedulingDaemon extends SchedulerDaemon {
                                 if (createRes.isErr()) {
                                     throw new Error(`Failed to schedule tasks: ${stringifyError(createRes.error)}`);
                                 }
-                                const skipped = taskProps.length - createRes.value.length;
-                                if (skipped > 0) {
-                                    logger.warn(`${skipped} task(s) were not scheduled.`);
+                                if (createRes.value.cappedGroupKeys.length > 0) {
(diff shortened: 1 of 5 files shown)


packages/scheduler/lib/daemons/scheduling/scheduling.daemon.ts
```diff
@@ -58,37 +58,35 @@ export class SchedulingDaemon extends SchedulerDaemon {
...
+                                    heartbeatTimeoutSecs: schedule.heartbeatTimeoutSecs,
+                                    ownerKey: null
+                                }));
+                                const createRes = await tasks.create(trx, taskProps);
+                                if (createRes.isErr()) {
+                                    throw new Error(`Failed to schedule tasks: ${stringifyError(createRes.error)}`);
+                                }
+                                const skipped = taskProps.length - createRes.value.length;
+                                if (skipped > 0) {
+                                    logger.warn(`${skipped} task(s) were not scheduled.`);
```
TBonnin: we are losing error message at the individual task level with this approach but I think it is worth the performance gain

packages/scheduler/lib/models/tasks.ts
```diff
@@ -123,38 +124,65 @@ export const DbTask = {
...
-        if (!inserted?.[0]) {
-            return Err(new Error(`Error: no task '${taskProps.name}' created`));
+        const now = new Date();
+        const toInsertPerGroup = new Map<string, Task[]>();
+        for (const props of taskProps) {
+            if (!toInsertPerGroup.has(props.groupKey)) {
+                toInsertPerGroup.set(props.groupKey, []);
+            }
+            const group = toInsertPerGroup.get(props.groupKey)!;
+            if (group.length < opts.groupTaskCap - (sizes.value.get(props.groupKey) ?? 0)) {
```
pfreixes, line 150: We could still log here that some tasks were discarded due to reaching the max cap for a specific group key

packages/scheduler/lib/daemons/scheduling/scheduling.daemon.ts
pfreixes: This ones would be specifically related to group keys that reached the max right? we can use the log that I was proposing [here](https://github.com/NangoHQ/nango/pull/5688/changes#r2974408869)

pfreixes: Failures at the middle of the  batch would go over this other [path](https://github.com/NangoHQ/nango/pull/5688/changes#diff-1bb91dcd1167214809d8945aff95d5f28a5bfc01fab0dd126460c694892934b1R77) right? But these would be errors accessing to the DB which not sure if we require to have detail at task level

pfreixes (approved):
Left one comment, up to you

packages/scheduler/lib/daemons/scheduling/scheduling.daemon.ts
TBonnin: the function returned a `Result` so query failure are captured by the error. What's not captured by the return type is the capped group. Trying to keep the task module single purpose (tasks) and not have additional side-effects (including logging).
Let's explicitely return the capped groups though and let the caller do the logging

packages/scheduler/lib/daemons/scheduling/scheduling.daemon.ts
```diff
@@ -56,39 +56,36 @@ export class SchedulingDaemon extends SchedulerDaemon {
...
+                                    groupMaxConcurrency: envs.SYNC_ENVIRONMENT_MAX_CONCURRENCY,
+                                    retryCount: 0,
+                                    retryMax: schedule.retryMax,
+                                    createdToStartedTimeoutSecs: schedule.createdToStartedTimeoutSecs,
+                                    startedToCompletedTimeoutSecs: schedule.startedToCompletedTimeoutSecs,
+                                    heartbeatTimeoutSecs: schedule.heartbeatTimeoutSecs,
+                                    ownerKey: null
+                                }));
+                                const createRes = await tasks.create(trx, taskProps);
+                                if (createRes.isErr()) {
```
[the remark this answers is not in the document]
TBonnin: A drawback to accept in order to have performant creation of multiple tasks.

PR #5692, opened 2026-03-23
feat(scheduler): add queue backpressure monitoring

pfreixes:
Periodic daemon that detects groups where CREATED tasks exceed group_max_concurrency, emitting a DD gauge (nango.orch.queue.backpressure) tagged by groupKey and primitive for the top N
  offenders.

 

---

It queries queued `CREATED` tasks per `group_key` to identify top backpressure offenders, emits Datadog gauges for those groups, and allows configuration of the monitoring tick interval and top-N limit.

---


packages/scheduler/lib/daemons/monitoring/queue-depth-monitoring.daemon.ts
```diff
@@ -0,0 +1,54 @@
...
+    async run(): Promise<void> {
+        const topN = envs.ORCHESTRATOR_QUEUE_DEPTH_MONITORING_TOP_N;
+
+        const queueDepth = await tasks.getQueueDepth(this.db, { topN, threshold: this.threshold, groupKeyPattern: this.groupKeyPattern });
+        if (queueDepth.isErr()) {
+            logger.error(queueDepth.error);
+            return;
+        }
+
+        for (const { group_key, cnt } of queueDepth.value) {
```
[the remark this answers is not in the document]
pfreixes: Ill update the description, for now we will not be logging them since all info is in DD

packages/scheduler/lib/scheduler.ts
```diff
@@ -63,20 +66,28 @@ export class Scheduler {
             onError
         });
         this.cleaning = new CleaningDaemon({ db, abortSignal: this.ac.signal, onError });
+        this.queueDepthMonitors = [
```
rossmcewan: This section should probably be configurable from the environment similar to how the processors are configured:  see [here](https://github.com/NangoHQ/nango/blob/b8515093c2e432f22bfffe81e4eb688d5243cedd/packages/utils/lib/environment/parse.ts#L94) - then could get rid of the `ORCHESTRATOR_QUEUE_DEPTH_MONITORING_THRESHOLD_XX` vars.

packages/utils/lib/environment/parse.ts
```diff
@@ -81,6 +81,32 @@ export const ENVS = z.object({
...
+                    groupKeyPattern: z.string(),
+                    threshold: z.coerce.number()
+                })
+            )
+        )
+        .default([
+            { groupKeyPattern: 'sync*', threshold: 500 },
+            { groupKeyPattern: 'action*', threshold: 500 },
+            { groupKeyPattern: 'webhook*', threshold: 500 },
+            { groupKeyPattern: 'on-event*', threshold: 100 }
```
pfreixes: @rossmcewan @TBonnin just using what seems to be used the defaults for `max_concurrency` as an example this [one](https://github.com/NangoHQ/nango/blob/master/packages/scheduler/lib/daemons/scheduling/scheduling.daemon.ts#L68) for syncs (IIRC there was a way for increasing concurrency for a customer right? how?)

packages/scheduler/lib/scheduler.ts
pfreixes: done!

packages/utils/lib/environment/parse.ts
rossmcewan: I don't think we have a way to do that per customer - the defined limit is per environment and I recall us discussing allowing this value to be defined on the plan.

rossmcewan (approved):

packages/utils/lib/environment/parse.ts
TBonnin: there are several concurrency control mechanisms. The one this PR is monitoring is the processing/dequeuing mechanism that is not customer-based. It is in place to prevent processing starvation by one type of functions and load management at platform level.
What we want to monitor is the the per group queuing at the orchestrator level. `group` is the concept used in the orchestrator to queue task. A group value currently looks like `sync:environment:8725`
In order to monitor and graph executions that are delayed we simply need to run a query like 
```
SELECT count(*), group_key
FROM "nango_scheduler"."tasks"
WHERE state = 'CREATED'
  AND group_max_concurrency > 0
GROUP BY group_key
HAVING count(*) > max(group_max_concurrency)
ORDER BY 1 DESC
LIMIT 50;
```
to retrieve all groups that have more pending tasks that allowed by the group_max_concurrency

TBonnin (changes requested):

packages/utils/lib/environment/parse.ts
pfreixes: > The one this PR is monitoring is the processing/dequeuing mechanism that is not customer-based.

Well is not totally true this right? correct me if Im wrong, the current `dequeue` already considers the `group_max_concurrency `, so when a specific group key reaches this concurrency we start slowing down the consumption - moving from CREATED to STARTED - of tasks. The current PR introduces a global monitor which considers this scenario, first when using the general counter - tagged as `others` - and also for tracking specific group keys when the go beyond specific thresholds - this is more for fighting the DD cardinality issue, so it provides relevant information when the surge of the depth queue is general or when its attributed to a specific customer.

True that there is a chance that bc of how the threshold stuff works, we cold start reporting the queue depth for a specific group key much beyond to when the customer started to have backpressure - i.e group_max_concurrency could be configured at 100 and hour threshold for start annotating in DD could be configured to 500.

But I like the idea of having a datapoint explicit when back-pressure kicks-in

We can go for putting what you say like

```
  SELECT t.group_key, t.group_max_concurrency,
         count(*) FILTER (WHERE t.state = 'CREATED') as queued,
         count(*) FILTER (WHERE t.state = 'STARTED') as running
  FROM tasks t
  WHERE t.state IN ('CREATED', 'STARTED')
    AND t.group_max_concurrency > 0
  GROUP BY t.group_key, t.group_max_concurrency
  HAVING count(*) FILTER (WHERE t.state = 'STARTED') >= t.group_max_concurrency
     AND count(*) FILTER (WHERE t.state = 'CREATED') > 0
```

This would give us group keys under back-pressure + the total events that these group keys are being piled up in the queue, we should still limit this to a certain N customers for avoiding cardinality issues with DD.

WDYT?

pfreixes: Only caveat on going in that direction is that we will lose any visibility on the "overall" queue growth, we could keep at least the original intention for getting the overall queue depth.

TBonnin: It is a more complicated sql query, require a daemon/query per functions types, more complicated config that isn't tied to the real concurrency parameter used for dequeuing. I don't really see why we could not simply do a single `count by group limit N` query (no need for config, one query, one daemon) and report this to DD.

pfreixes: All changes applied, let me know if there is something else !

packages/scheduler/lib/daemons/monitoring/backpressure-monitoring.daemon.ts
```diff
@@ -0,0 +1,33 @@
...
+    async run(): Promise<void> {
+        const result = await tasks.getGroupsWithBackpressure(this.db, { limit: envs.ORCHESTRATOR_BACKPRESSURE_MONITORING_TOP_N });
+        if (result.isErr()) {
+            logger.error(result.error);
+            return;
+        }
+
+        for (const { group_key, queued } of result.value) {
+            const primitive = group_key.split(':')[0]!;
+            metrics.gauge(metrics.Types.ORCH_QUEUE_BACKPRESSURE, queued, { groupKey: group_key, primitive });
```
TBonnin, line 30: not sure we need to add an extra primitive dimension. That one more thing we would not have to give DD money for

TBonnin (approved):

packages/scheduler/lib/daemons/monitoring/backpressure-monitoring.daemon.ts
pfreixes: We have 4 primitives and we are capping the list of max offenders to be reported by default to 10, and we are expecting that during platform wide incidents the top N reported will be always the same (this is the use case which could cause some issues in terms of cardinality) so expecting that worst case scenario this will produce 40 custom metrics which Im expecting that should be peanuts.

PR #5693, opened 2026-03-23
fix(oauth): resolve flat dot-notation keys in token url interpolation

hassan254-prog:

- Resolve flat dot-notation keys in token url interpolation. This change relates to the Sentry public installation flow, which uses webhooks for token exchange. The installation UUID is stored in the connection config as `{ "installation.uuid": string }` and is required to generate the token URL `${connectionConfig.installation.uuid}`.

---

It also adds unit tests to cover flat-key interpolation, precedence between nested and flat keys, and missing-key behavior.

---

### DIFF
--- a/packages/shared/lib/utils/utils.ts
+++ b/packages/shared/lib/utils/utils.ts
@@ -291,7 +291,11 @@ export function interpolateStringFromObject(str: string, replacers: Record<strin
         if (b === 'endpoint') {
             return (replacers['endpoint'] as string | undefined) ?? '';
         }
-        const r = b.split('.').reduce((o: Record<string, any>, i: string) => o[i], replacers);
+        const r =
+            b
+                .split('.')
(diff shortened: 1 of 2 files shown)


packages/shared/lib/utils/utils.ts
```diff
@@ -291,7 +291,11 @@ export function interpolateStringFromObject(str: string, replacers: Record<strin
         if (b === 'endpoint') {
             return (replacers['endpoint'] as string | undefined) ?? '';
         }
-        const r = b.split('.').reduce((o: Record<string, any>, i: string) => o[i], replacers);
+        const r =
+            b
+                .split('.')
+                .reduce((o: Record<string, any> | undefined, i: string) => (o != null ? o[i] : undefined), replacers as Record<string, any> | undefined) ??
+            replacers[b];
```
TBonnin, line 300: a comment to explain the why of the fallback would be nice

TBonnin (approved):

PR #5694, opened 2026-03-23
perf(records): use ctid-based CTE to delete outdated records

TBonnin:
Replace the WHERE id IN (subquery) pattern with a MATERIALIZED CTE that selects ctids, then updates rows by physical address. On partitions with high churn and autovacuum lag, the previous approach triggered a bloated index scan to re-locate rows by id for the write step. The new approach bypasses the index entirely for the update

 

---

This change is specifically applied within `deleteOutdatedRecords` and preserves the existing transactional flow and advisory locking while switching the update mechanism to `ctid`-based selection.

---

### DIFF
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -918,35 +918,33 @@ export async function deleteOutdatedRecords({
             const batchResult = await retry(
                 () => {
                     return db.transaction(async (trx) => {
-                        const now = trx.fn.now(6);
                         // Lock to prevent concurrent modifications with upserts and deletes
                         await trx.raw(`SELECT pg_advisory_xact_lock(?) as lock_records_outdated`, [newLockId(connectionId, model)]);
 
-                        const res: { external_id: string; size_bytes: number; partition: string }[] = await trx
-                            .from<FormattedRecord>(RECORDS_TABLE)
-                            .whereIn('id', function (sub) {
-                                sub.select('id')
-                                    .from(RECORDS_TABLE)
-                                    .where({
-                                        connection_id: connectionId,
-                                        model,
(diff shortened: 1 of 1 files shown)


pfreixes (approved):
Nice, good finding and didn't know about ctid!

Seems that original issue is not having an efficient way to scan the ids, from my understanding this ```records_p169_connection_id_model_id_key ... (actual time=0.122..15957.956 rows=127409)``` shows that the planner decided to do a index scan and return all rows for later filtering the 5K over this 127K? so its not only about dead tuples but also about how we can target efficiently the ids batch.

Using the citd fixes this!

I still believe that using the cursor strategy for the inner query can help on the dead tuples too (we might want to gather some data about the cost of none page 0 queries), but lets go step by step, and this one might already solve the issues that we are observing Today!

TBonnin:
> Nice, good finding and didn't know about ctid!
> 
> Seems that original issue is not having an efficient way to scan the ids, from my understanding this `records_p169_connection_id_model_id_key ... (actual time=0.122..15957.956 rows=127409)` shows that the planner decided to do a index scan and return all rows for later filtering the 5K over this 127K? so its not only about dead tuples but also about how we can target efficiently the ids batch.
> 
> Using the citd fixes this!
> 
> I still believe that using the cursor strategy for the inner query can help on the dead tuples too (we might want to gather some data about the cost of none page 0 queries), but lets go step by step, and this one might already solve the issues that we are observing Today!

it indeed doesn't actually solve the root problem which is that `index-only scan` are not actually index only when pg has to go to the heap to ensure rows visibility. This change only remove one of the supposedly `index-only scan` to prevent the worst case scenario.

PR #5695, opened 2026-03-24
fix(pubsub): adding sns-sqs transport

rossmcewan:
---

**Add `sns-sqs` pubsub transport with migration mode and env validation**

This PR introduces a new `sns-sqs` pubsub transport backed by AWS SNS/SQS, including publishing with subject attributes, long-polling consumers, SNS envelope unwrapping, and delete-vs-retry handling based on deserialization and callback outcomes. A `Migration` transport is added to separate publisher and subscriber transports for migration scenarios, and the default transport selection now supports `activemq`, `sns-sqs`, `migration`, or `none` via `NANGO_PUBSUB_TRANSPORT`.

Environment parsing is extended to validate `NANGO_PUBSUB_SNS_SQS_CONFIG` along with polling limits, with unit tests added for both environment parsing and new transport behavior. Dependencies for AWS SDK clients and `zod` are added, and the example env file removes the old pubsub comment block.

---


packages/pubsub/lib/transport/combined.ts
```diff
@@ -0,0 +1,60 @@
+import { Err, Ok } from '@nangohq/utils';
+
+import type { Event } from '../event.js';
+import type { SubscribeProps, Transport } from './transport.js';
+import type { Result } from '@nangohq/utils';
+
+export class CombinedTransportError extends Error {
```
rossmcewan: This is used during the migration phase so that the subscriber side will listen to both queues

packages/pubsub/lib/transport/default.ts
```diff
@@ -15,6 +17,21 @@ export class DefaultTransport implements Transport {
     constructor() {
         if (envs.NANGO_PUBSUB_TRANSPORT === 'activemq') {
             this.transport = new ActiveMQ();
+        } else if (envs.NANGO_PUBSUB_TRANSPORT === 'sns-sqs') {
+            const cfg = envs.NANGO_PUBSUB_SNS_SQS_CONFIG;
+            this.transport = new SnsSqs({
+                topicArns: cfg.topicArns,
+                queueUrls: cfg.queueUrls
+            });
+        } else if (envs.NANGO_PUBSUB_TRANSPORT === 'migration') {
```
rossmcewan, line 26: This allows us to control the migration from the environment:
- Subscriber (i.e. metering) listens to both queues, publishers continue sending to activemq
- Change publisher to use sns-sqs
- When activemq is drained, change subscriber to use sns-sqs only
- Remove migration option

packages/pubsub/lib/transport/combined.ts
```diff
@@ -0,0 +1,60 @@
...
+        const results = await Promise.all(this.transports.map((t) => t.connect()));
+        return mergeResults('connect', results);
+    }
+
+    async disconnect(): Promise<Result<void>> {
+        const results = await Promise.all(this.transports.map((t) => t.disconnect()));
+        return mergeResults('disconnect', results);
+    }
+
+    async publish(event: Event): Promise<Result<void>> {
```
TBonnin: I don't think we should allow publishing to all transports. If subscribers downstream logic don't have dedup logic it will result in duplicated events

packages/pubsub/lib/transport/sns-sqs.ts
```diff
@@ -0,0 +1,245 @@
...
+const logger = getLogger('pubsub.sns-sqs');
+
+function subscriptionKey(consumerGroup: string, subject: string): string {
+    return `${consumerGroup}:${subject}`;
+}
+
+/** SNS→SQS subscriptions wrap the published payload in a JSON envelope unless raw delivery is enabled. */
+function unwrapSqsBody(body: string): string {
+    try {
+        const parsed = JSON.parse(body) as { Type?: string; Message?: string };
```
TBonnin: Cannot we validate with zod and/or import the real type from aws package instead of a fragile type casting?

packages/pubsub/lib/transport/sns-sqs.ts
```diff
@@ -0,0 +1,245 @@
...
+            return;
+        }
+        const abort = new AbortController();
+        this.pollerAbort.set(key, abort);
+        void this.pollQueue(queueUrl, props, abort.signal);
+    }
+
+    private async pollQueue<TSubject extends Event['subject']>(queueUrl: string, props: SubscribeProps<TSubject>, signal: AbortSignal): Promise<void> {
+        const { subject, callback } = props;
+        while (!signal.aborted) {
```
TBonnin, line 181: should this loop have a configurable `tickIntervalMs` wait

packages/pubsub/lib/transport/sns-sqs.ts
```diff
@@ -0,0 +1,245 @@
...
+        expectedSubject: TSubject,
+        callback: SubscribeProps<TSubject>['callback']
+    ): Promise<'delete' | 'retry'> {
+        const messageSubject = getSubjectMessageAttribute(body, messageAttributes);
+        if (messageSubject !== expectedSubject) {
+            report(new Error('SNS+SQS: message subject does not match subscriber subject'), {
+                expectedSubject,
+                messageSubject
+            });
+            return 'retry';
```
TBonnin, line 232: can it self-heal by retrying if it was delivered to the wrong subject in the first place?

packages/pubsub/lib/transport/sns-sqs.ts
```diff
@@ -0,0 +1,245 @@
...
+        }
+
+        const rawPayload = unwrapSqsBody(body);
+        let buf: Buffer;
+        try {
+            buf = Buffer.from(rawPayload, 'base64');
```
TBonnin: Can this ever throw?

packages/pubsub/lib/transport/sns-sqs.ts
```diff
@@ -0,0 +1,245 @@
...
+    } catch {
+        // non-JSON payload has no SNS envelope attributes
+    }
+
+    return undefined;
+}
+
+export interface SnsSqsProps {
+    /** Maps event subject (`user`, `team`, `usage`) to SNS topic ARN. Every published subject must have an entry. */
+    topicArns?: Record<string, string>;
```
TBonnin: `topicArns?: Record<event['subject'], string>` to make it impossible to configure subject that don't exist

packages/utils/lib/environment/parse.ts
```diff
@@ -379,7 +379,29 @@ export const ENVS = z.object({
...
+            try {
+                return JSON.parse(s) as unknown;
+            } catch {
+                ctx.addIssue(`Invalid JSON in NANGO_PUBSUB_SNS_SQS_CONFIG`);
+                return z.NEVER;
+            }
+        })
+        .pipe(
+            z.object({
+                topicArns: z.record(z.string(), z.string()).optional().default({}),
```
TBonnin: `topicArns: z.record(z.enum(['user', ...] satisfies Event['subject']), z.string()).optional().default({}),`
should we also validate `arn`?

packages/pubsub/lib/transport/sns-sqs.ts
```diff
@@ -0,0 +1,245 @@
...
+            report(new Error('SNS+SQS: no SQS queue URL for subscription'), { consumerGroup, subject, key });
+            return;
+        }
+
+        const existing = this.pollerAbort.get(key);
+        if (existing) {
+            existing.abort();
+            this.pollerAbort.delete(key);
+        }
+        this.activeSubscriptions.set(key, props);
```
TBonnin: this is `set` but never read

packages/pubsub/lib/transport/combined.ts
rossmcewan: From a configuration perspective, this wouldn't happen because publishers are not in the migration state - but I do agree that the code could still allow this to happen. I will rename the `Combined` transport to be a `Migration` transport and explicitly indicate the publisher and subscribers.

packages/pubsub/lib/transport/sns-sqs.ts
rossmcewan: I don't think so. It's a long poll with a wait time.

packages/pubsub/lib/transport/sns-sqs.ts
rossmcewan: It was so it will end up in DLQ vs getting deleted.

packages/pubsub/lib/transport/sns-sqs.ts
rossmcewan: Left over from something I was trying out. Removed.

packages/pubsub/lib/transport/sns-sqs.ts
rossmcewan: Yes! Have improved the validation of the queue urls too.

packages/pubsub/lib/transport/sns-sqs.ts
rossmcewan: Removed.

packages/pubsub/lib/transport/sns-sqs.ts
TBonnin: where is it waiting?

TBonnin (approved):

packages/pubsub/lib/transport/sns-sqs.ts
rossmcewan: `WaitTimeSeconds` on the `ReceiveMessageCommand`.

PR #5699, opened 2026-03-24
feat(connections): add   pausedSyncs to connection list API

agusayerza:
Add `pausedSyncs` field to the connections list API response.

We want to show which connections have paused syncs on the FE when listing connections. For each page of connections returned, we fetch their syncs from the DB and do a targeted lookup to the orchestrator using exact schedule names, bounded by page size (~20 connections) rather than the full environment.

- Add `pausedSyncs: string[]` to `ApiConnectionSimple` type
- Add `getSyncsByConnectionIds()` to sync service
- Use existing `Orchestrator.searchSchedules()` with exact names for the current page
- Graceful fallback: orchestrator errors log and return empty `pausedSyncs` rather than 500ing the entire list endpoint

---

This change also formalizes the enrichment flow across shared contracts and service boundaries, including safer failure handling and tighter validation on schedule-search inputs, so paused-sync visibility is added in a resilient, bounded way without compromising the stability of the connections listing path.

---


packages/server/lib/controllers/v1/connections/getConnections.ts
```diff
@@ -45,9 +48,24 @@ export const getConnections = asyncWrapper<GetConnections>(async (req, res) => {
...
+
     res.status(200).send({
         data: connections.map((data) => {
-            return connectionSimpleToApi({ data: data.connection, provider: data.provider, activeLog: data.active_logs, endUser: data.end_user });
+            return connectionSimpleToApi({
+                data: data.connection,
+                provider: data.provider,
+                activeLog: data.active_logs,
+                endUser: data.end_user,
+                hasPausedSyncs: pausedConnectionIds.has(data.connection.id)
```
marcindobry: Could we instead return a list of paused sync names and let the client figure out the boolean just based on .length > 0? We've done all the heavy lifting for it already at this point and so it's more information for free.

packages/scheduler/lib/db/migrations/20260324000000_schedules_paused_name_index.ts
```diff
@@ -0,0 +1,15 @@
+import type { Knex } from 'knex';
+
+import { SCHEDULES_TABLE } from '../../models/schedules.js';
+
+export const config = {
+    transaction: false
+};
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.raw(`CREATE INDEX CONCURRENTLY IF NOT EXISTS "idx_schedules_paused_name" ON ${SCHEDULES_TABLE} (name) WHERE state = 'PAUSED';`);
```
marcindobry: We're giving the ability to search by any state, but if someone unknowingly would add code to do search for syncs in other states, they might unknowingly run into a performance issue. Could we add a comment in the new functions that calling them with non-paused states might require adding an index like this?

marcindobry (changes requested):

packages/scheduler/lib/db/migrations/20260324000000_schedules_paused_name_index.ts
TBonnin: have you run an EXPLAIN on the search query? Given how small `schedules` table is, this index might not be even needed

packages/orchestrator/lib/routes/v1/schedules/postSearch.ts
```diff
@@ -14,6 +14,8 @@ type PostSearch = Endpoint<{
     Path: typeof path;
     Body: {
         names?: string[] | undefined;
+        namePrefix?: string | undefined;
```
TBonnin: is it better to add a new explicit param or to support wildcard in the existing `names` one?

packages/orchestrator/lib/routes/v1/schedules/postSearch.ts
```diff
@@ -23,6 +25,8 @@ type PostSearch = Endpoint<{
 const bodySchema = z
     .object({
         names: z.array(z.string().min(1)).optional(),
+        namePrefix: z.string().min(1).optional(),
+        state: z.enum(['STARTED', 'PAUSED', 'DELETED']).optional(),
```
TBonnin: recently found out you can use `satisfies ScheduleState[]` to prevent accidentally having an enum value that doesn't exist.

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -114,6 +123,19 @@ export class Orchestrator {
         return Ok(scheduleMap);
     }
 
+    async getPausedSyncsByEnvironment({ environmentId }: { environmentId: number }): Promise<Result<string[]>> {
+        const namePrefix = `environment:${environmentId}:sync:`;
```
TBonnin: let's expose the prefix from the `orchestrator.ts` to avoid duplication and make sure we only need to change it in one place if needed

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -114,6 +123,19 @@ export class Orchestrator {
...
+        const result = await this.client.searchSchedulesByPrefix({ namePrefix, state: 'PAUSED', limit: 10_000 });
```
TBonnin: this limit might quickly fall apart for Linux Foundation. That's also a huge amount of data to transfer when the only thing we are interested in is a boolean. 
What about a new orchestrator endpoint like `GET /stats?name_prefix=...` that return a count per state for now

packages/server/lib/controllers/v1/connections/getConnections.ts
agusayerza: Done

packages/shared/lib/clients/orchestrator.ts
agusayerza: No longer needed

packages/scheduler/lib/db/migrations/20260324000000_schedules_paused_name_index.ts
agusayerza: I had run an explain, with the old approach it would perform better, but its hard to tell if worth it. Now that we call it by page, it makes no sense so I dropped the index. Old approach seq scanned 74K rows in ~50ms without the index, for reference. The new approach hits `schedules_unique_name` and finishes in 0.1ms

packages/orchestrator/lib/routes/v1/schedules/postSearch.ts
agusayerza: No longer needed

packages/shared/lib/clients/orchestrator.ts
agusayerza: I changed the DB approach to only take into consideration the connections loaded on the current page, that way we have it scoped down and don't need to resolve every sync to display the icon

packages/server/lib/controllers/v1/connections/getConnections.ts
```diff
@@ -45,9 +48,32 @@ export const getConnections = asyncWrapper<GetConnections>(async (req, res) => {
         withError: queryString.withError
     });
 
+    const pausedSyncsByConnection = new Map<number, string[]>();
+    const connectionIds = connections.map((data) => data.connection.id);
+    const syncs = await getSyncsByConnectionIds({ connectionIds });
+    if (syncs.length > 0) {
+        const scheduleResult = await orchestrator.searchSchedules(syncs.map((s) => ({ syncId: s.id, environmentId: environment.id })));
+        if (scheduleResult.isErr()) {
+            res.status(500).send({ error: { code: 'server_error' } });
```
pfreixes: should we log the error or have it attached to the span somehow? otherwise will be tough to know what caused the error

packages/server/lib/controllers/v1/connections/getConnections.ts
```diff
@@ -45,9 +48,32 @@ export const getConnections = asyncWrapper<GetConnections>(async (req, res) => {
         withError: queryString.withError
     });
 
+    const pausedSyncsByConnection = new Map<number, string[]>();
+    const connectionIds = connections.map((data) => data.connection.id);
+    const syncs = await getSyncsByConnectionIds({ connectionIds });
+    if (syncs.length > 0) {
+        const scheduleResult = await orchestrator.searchSchedules(syncs.map((s) => ({ syncId: s.id, environmentId: environment.id })));
```
TBonnin: let's add some safeguard to orchestrator `postSearch` and max the length of names to a high value that is not infinite, so we don't happen to send a huge number by accident

packages/shared/lib/services/sync/sync.service.ts
```diff
@@ -225,6 +225,20 @@ export const getSyncs = async (
...
+    return db.knex.select('id', 'nango_connection_id').from<Sync>(TABLE).whereIn('id', syncIds).andWhere({ deleted: false });
+};
+
+export const getSyncsByConnectionIds = async ({ connectionIds }: { connectionIds: number[] }): Promise<Pick<Sync, 'id' | 'name' | 'nango_connection_id'>[]> => {
+    if (connectionIds.length === 0) {
+        return [];
+    }
+    return db.knex.select('id', 'name', 'nango_connection_id').from<Sync>(TABLE).whereIn('nango_connection_id', connectionIds).andWhere({ deleted: false });
+};
+
```
TBonnin, line 254: can we make all new functions that can fail return a `Result`?

marcindobry (approved):

PR #5700, opened 2026-03-25
fix(records): do not overwrite json if hash is identical

TBonnin:
pg is smart enough to only copy the toast pointer when updating a toasted value with the old tuple data

ex:
```
EXPLAIN (ANALYSE, BUFFERS) INSERT INTO toast_test (id, small_col, big_col)
VALUES (1, 'b', repeat('x', 100000))
ON CONFLICT (id)
DO UPDATE SET
    big_col   = toast_test.big_col -- old data
;

Insert on toast_test  (cost=0.00..0.01 rows=0 width=0) (actual time=0.803..0.810 rows=0 loops=1)
  ...
  Buffers: shared hit=4 dirtied=1
  ...
```

vs updating with the new identical data

```
EXPLAIN (ANALYSE, BUFFERS) INSERT INTO toast_test (id, small_col, big_col)
VALUES (1, 'a', repeat('x', 100000))
ON CONFLICT (id)
DO UPDATE SET
    big_col   = EXCLUDED.big_col; -- new value
;   
Insert on toast_test  (cost=0.00..0.01 rows=0 width=0) (actual time=2.173..2.174 rows=0 loops=1)
  ...
  Buffers: shared hit=264 dirtied=13
  ...

```

Note: we cannot prevent the creation of a new tuple for each record because column like sync_job_id are always updated. This commit is only an optimization that doesn't require to change the design of Nango records features

 

---

**Skip JSON overwrite on identical record hashes and update size delta tracking**

This PR adjusts `records` upsert behavior to avoid rewriting `json` when the incoming `data_hash` matches the stored value, reducing unnecessary TOAST writes. It also changes upsert result handling to compute and propagate `delta_size_bytes` for size accounting, and updates integration tests to assert decrypted payloads from the database via a new helper.

---

### DIFF
--- a/packages/records/lib/models/records.integration.test.ts
+++ b/packages/records/lib/models/records.integration.test.ts
@@ -8,6 +8,7 @@ import { db } from '../db/client.js';
 import { migrate } from '../db/migrate.js';
 import { formatRecords } from '../helpers/format.js';
 import * as Records from '../models/records.js';
+import { decryptRecordData } from '../utils/encryption.js';
 
 import type { FormattedRecord, UnencryptedRecordData, UpsertSummary } from '../types.js';
 import type { MergingStrategy, Result } from '@nangohq/types';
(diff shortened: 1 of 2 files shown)


packages/records/lib/models/records.integration.test.ts
```diff
@@ -104,11 +105,14 @@ describe('Records service', () => {
...
-        expect(after.find((r) => r.external_id === '3')?.sync_job_id).toBe(1);
-        expect(after.find((r) => r.external_id === '4')?.sync_job_id).toBe(1);
+        await expect(fromDb(connectionId, model, '1')).resolves.toMatchObject({ external_id: '1', sync_job_id: 2, decrypted: { id: '1', name: 'John Doe' } });
+        await expect(fromDb(connectionId, model, '2')).resolves.toMatchObject({
+            external_id: '2',
+            sync_job_id: 2,
+            decrypted: { id: '2', name: 'Jane Much Longer Name Doe' }
+        });
+        await expect(fromDb(connectionId, model, '3')).resolves.toMatchObject({ external_id: '3', sync_job_id: 1, decrypted: { id: '3', name: 'Max Doe' } });
+        await expect(fromDb(connectionId, model, '4')).resolves.toMatchObject({ external_id: '4', sync_job_id: 1, decrypted: { id: '4', name: 'Mike Doe' } });
```
TBonnin, line 115: Extending the tests to also check the persisted json

packages/records/lib/models/records.ts
```diff
@@ -36,8 +36,7 @@ interface UpsertResult {
     id: string;
     last_modified_at: string;
     previous_last_modified_at: string | null;
-    size_bytes: number;
-    previous_size_bytes: number | null;
+    delta_size_bytes: number;
```
TBonnin, line 39: calculating the size delta within the query is still problematic and forces pg to always load the json. I am gonna try to address this issue in next PR. In the meantime we can assess the gain provided by pg only copying the pointer when json is toasted

rossmcewan (approved):

PR #5701, opened 2026-03-25
feat(providers): agent-powered oauth2 scope discovery

hassan254-prog:

- This pr introduces an automated pipeline to maintain a catalogue of oauth2 scopes for every provider in Nango, and surfaces them as a searchable dropdown in the scopes input UI.

---

**`feat(providers)`: `OAuth2` scope catalog pipeline with API exposure and scope-picker UI**

This PR introduces a full provider-scope catalog system centered on a new `packages/providers/providers.scopes.yaml` dataset and a large new automation script at `scripts/validation/providers/sync-scopes.ts`. The script supports deterministic and agent-driven modes, batch processing, duplicate-key/YAML sanitization, comment preservation, fallback to `default_scopes`, validation guards, and CI check/write flows. It is wired into CI via `test:providers:scopes`, monthly automated sync in `.github/workflows/sync-provider-scopes.yaml`, and updates in `.github/workflows/validation.yaml`.

On the product side, provider APIs now include `availableScopes` through updates in `packages/types/lib/providers/api.ts`, `packages/server/lib/formatters/provider.ts`, and provider controllers. The web UI scope input was substantially refactored in `packages/webapp/src/components-v2/ScopesInput.tsx` to a chips+combobox interaction using new `packages/webapp/src/components-v2/ui/combobox.tsx`, with searchable suggestions, dedupe, paste/comma/enter parsing, and copy support, and integrated into `OAuthCreateForm`.

---


packages/providers/providers.scopes.yaml
hassan254-prog, line 1: chose a separate file instead of adding scopes directly into `providers.yaml`. The agent writes to this file, so keeping it isolated means the agent can never accidentally corrupt providers at config level. 
The merge happens at load time in `index.ts`.

packages/webapp/src/components-v2/ScopesInput.tsx
```diff
@@ -1,109 +1,189 @@
...
             setScopes(newScopes);
-            setInputValue('');
         } finally {
             setLoading(false);
         }
     };
 
-    const finalPlaceholder = placeholder ?? (isSharedCredentials ? 'Nango provided scopes' : 'Single scope or comma-separated list of scopes');
+    const copyScopes = () => {
+        void navigator.clipboard.writeText(scopes.join(','));
```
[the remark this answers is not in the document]
hassan254-prog: good catch, since this is a besteffort copy action with no user-facing feedback, the silent failure is acceptable here.

packages/providers/lib/index.ts
```diff
@@ -60,9 +66,38 @@ export function loadProvidersYaml(): Record<string, Provider> | undefined {
...
+        const providersScopesYamlPath = path.join(pkgRoot, 'providers.scopes.yaml');
+        if (!fs.existsSync(providersScopesYamlPath)) {
+            return undefined;
+        }
+
+        const scopesFileEntries = yaml.load(fs.readFileSync(providersScopesYamlPath).toString()) as Record<string, string[]> | null;
+        providerScopes = scopesFileEntries || {};
+        return providerScopes;
+    } catch (err) {
+        console.error('Failed to load providers.scopes.yaml', err);
```
TBonnin: I don't think we should use console anywhere

packages/providers/lib/index.ts
```diff
@@ -40,6 +45,7 @@ export function loadProvidersYaml(): Record<string, Provider> | undefined {
     try {
         const providersYamlPath = path.join(pkgRoot, 'providers.yaml');
         const fileEntries = yaml.load(fs.readFileSync(providersYamlPath).toString()) as Record<string, Provider | ProviderAlias>;
+        const scopes = loadProviderScopesYaml();
```
TBonnin: is providers also loaded by the runner? does the runner needs to load the scopes? does it matter?

scripts/validation/providers/sync-scopes.ts
```diff
@@ -0,0 +1,889 @@
...
+        if (line.startsWith('+++') || line.startsWith('---')) {
+            continue;
+        }
+
+        const content = line.slice(1);
+        if (content.startsWith(' ') || content.startsWith('\t')) {
+            continue;
+        }
+
+        const match = content.match(/^([a-z0-9][a-z0-9-]*):\s*$/);
```
TBonnin: not using the same regex as `removeDuplicateYamlKeys`. missing `_`

scripts/validation/providers/sync-scopes.ts
```diff
@@ -0,0 +1,889 @@
...
+const __dirname = path.dirname(__filename);
+const ROOT_DIR = path.resolve(__dirname, '../../..');
+
+config({ path: path.join(ROOT_DIR, '.env') });
+const providersPath = path.join(__dirname, '../../../packages/providers/providers.yaml');
+const providersScopesPath = path.join(__dirname, '../../../packages/providers/providers.scopes.yaml');
+const DEFAULT_MODEL = 'opencode/kimi-k2.5';
+const DEFAULT_PORT = 4096;
+const DEFAULT_POLL_INTERVAL_MS = 2000;
+const DEFAULT_SESSION_TIMEOUT_MINUTES = 2;
```
TBonnin, line 54: is 2mins enough for the agent to go through a batch?

TBonnin (commented):
I am not against leveraging llms but I open the generated yaml, click on the first link (brex docs) and it gave me a 404. I then checked a random provider (docusign) and there are some scopes missing. Maybe I was super unlucky but doesn't make me super confident.
One can argue it is better than having nothing but it is also a giant pile of data that we are gonna need to maintain

packages/providers/lib/index.ts
TBonnin: how long does it take to load?

hassan254-prog: Thanks for flagging this, no, this isn’t needed. I’ve updated it in the latest commit.

scripts/validation/providers/sync-scopes.ts
hassan254-prog: Updated with the latest commit.

packages/providers/lib/index.ts
hassan254-prog: I’ve removed all the ones that were present here in the latest commit.

hassan254-prog:
> I am not against leveraging llms but I open the generated yaml, click on the first link (brex docs) and it gave me a 404. I then checked a random provider (docusign) and there are some scopes missing. Maybe I was super unlucky but doesn't make me super confident. One can argue it is better than having nothing but it is also a giant pile of data that we are gonna need to maintain

Yes, I agree, it’s better than having nothing. We will also mention in the docs how we obtain this list to keep it transparent with our customers. I believe that, at the start, the list is not 💯  correct, but over time it will continue to improve through iterative updates run on a monthly basis.

scripts/validation/providers/sync-scopes.ts
hassan254-prog: It’s not enough for a batch. This is intended for single-provider runs, which works well. It’s currently set as a safeguard to prevent the agent from running indefinitely for a single provider. For batch updates, this is overridden in the ci to disable the timeout check entirely.

packages/providers/lib/index.ts
hassan254-prog: it takes roughly ~20–90ms depending on whether the file is already in the cache, but since it's only loaded once at startup and cached, it's a one-time cost.

scripts/validation/providers/sync-scopes.ts
```diff
@@ -0,0 +1,889 @@
...
+function chunkArray<T>(arr: T[], size: number): T[][] {
+    const chunks: T[][] = [];
+    for (let i = 0; i < arr.length; i += size) {
+        chunks.push(arr.slice(i, i + size));
+    }
+    return chunks;
+}
+
+function getOpencodeApiKey(model: string): string {
+    const [provider] = model.split('/');
```
marcindobry: Model name validation shouldn't be part of key retrieval.

scripts/validation/providers/sync-scopes.ts
```diff
@@ -0,0 +1,889 @@
...
+function freePort(port: number): void {
+    let pids: string;
+    try {
+        pids = execSync(`lsof -ti :${port}`, { encoding: 'utf-8' }).trim();
+    } catch {
+        return;
+    }
+    for (const pid of pids.split('\n').filter(Boolean)) {
+        try {
+            process.kill(Number(pid), 'SIGTERM');
```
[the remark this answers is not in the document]
marcindobry: Unresolving the comment because it is also something I noticed. We should fail fast and give instructions how to kill something on the port to do manually and not kill other processes.

scripts/validation/providers/sync-scopes.ts
```diff
@@ -0,0 +1,889 @@
...
+    }
+
+    const apiKey = getOpencodeApiKey(args.model);
+    freePort(args.port);
+
+    const opencodeConfig = {
+        model: args.model,
+        permission: {
+            external_directory: 'deny',
+            doom_loop: 'allow',
```
marcindobry, line 521: Why do we allow this?

scripts/validation/providers/sync-scopes.ts
```diff
@@ -0,0 +1,889 @@
...
+            read: { '*': 'allow' },
+            edit: { '*': 'allow' }
```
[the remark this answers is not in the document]
marcindobry: Unresolving because I also think this is legit. We could just allow it to edit the providers.scopes.yaml file specifically

scripts/validation/providers/sync-scopes.ts
```diff
@@ -0,0 +1,889 @@
...
+    await client.auth.set({
+        path: { id: 'opencode' },
+        body: { type: 'api', key: apiKey }
+    });
+
+    const baseRef = args.baseRef || getDefaultBaseRef();
+    const batches = chunkArray(oauth2TargetProviders, args.batchSize);
+    const totalProviders = oauth2TargetProviders.length;
+
+    console.log(`Processing ${totalProviders} providers in ${batches.length} batch(es) of up to ${args.batchSize}.`);
```
marcindobry, line 541: https://github.com/NangoHQ/nango/pull/5701#discussion_r3000806902

marcindobry: There are still a few in other parts of the file

marcindobry (changes requested):

scripts/validation/providers/sync-scopes.ts
hassan254-prog: The agent task involves writing scopes for multiple providers in batches. Each iteration reads from `providers.yaml`, fetches external docs, and writes to `providers.scopes.yaml`. We should allow this, as it can legitimately occur in this context.

scripts/validation/providers/sync-scopes.ts
hassan254-prog: It’s safe to include logging in this script since it’s a standalone cli script and not server side code.

scripts/validation/providers/sync-scopes.ts
```diff
@@ -0,0 +1,889 @@
...
+            external_directory: 'deny',
+            doom_loop: 'allow',
+            read: { '*': 'allow' },
+            edit: { [providersScopesPath]: 'allow' }
+        }
+    };
+
+    const { client, server } = await createOpencode({
+        port: args.port,
+        config: opencodeConfig as any
```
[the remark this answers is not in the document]
hassan254-prog: The sdk types only expose `ask | allow | deny` for edit/read, but the runtime api also accepts path keyed objects like `{ [path]: 'allow' }` ,  the types just don't cover that, hence the reason for casting.

scripts/validation/providers/sync-scopes.ts
hassan254-prog: The agent task involves writing scopes for multiple providers in batches. Each iteration reads from `providers.yaml`, fetches external docs, and writes to `providers.scopes.yaml`. We should allow this, as it legitimately occur in this context.

marcindobry (approved):

scripts/validation/providers/sync-scopes.ts
```diff
@@ -0,0 +1,889 @@
...
+    const match = output.stdout.match(/HEAD branch:\s+([^\n]+)/);
+    if (!match?.[1]) {
+        throw new Error('Unable to parse origin HEAD branch from `git remote show origin`.');
+    }
+
+    return `origin/${match[1].trim()}`;
+}
+
+function getChangedProviderNames(baseRef: string): string[] {
+    const output = runGit(['diff', '--unified=0', baseRef, '--', 'packages/providers/providers.yaml']);
```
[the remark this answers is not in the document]
hassan254-prog: This is intentional. `getChangedProviderNames` is designed to only trigger when a new top-level provider key is added to the `providers.yaml` file.

scripts/validation/providers/sync-scopes.ts
```diff
@@ -0,0 +1,889 @@
...
+    const allKeys = new Set([...Object.keys(before), ...Object.keys(after)]);
+    for (const key of allKeys) {
+        const a = dedupeAndSort(before[key] ?? []);
+        const b = dedupeAndSort(after[key] ?? []);
+        if (!sameScopes(a, b) && !allowedTouchedProviders.includes(key)) {
+            outOfScope.push(key);
+        }
+    }
+    if (outOfScope.length > 0) {
+        console.warn(`  Agent also updated out-of-scope provider(s) (keeping changes): ${outOfScope.sort().join(', ')}`);
```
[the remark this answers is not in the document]
hassan254-prog: This is intentional. The agent may legitimately touch adjacent provider entries during a run. Rejecting would make agent runs fragile. The warning is informational so reviewers can catch unexpected edits in the pr diff.

scripts/validation/providers/sync-scopes.ts
```diff
@@ -0,0 +1,889 @@
...
+if (args.verbose) {
+    console.log(`Loaded ${Object.keys(providers).length} providers`);
+    console.log(`Loaded ${Object.keys(providerScopes).length} scoped providers`);
+    console.log(`Detected ${oauth2Providers.length} OAuth2 providers`);
+    console.log(`Targeting ${oauth2TargetProviders.length} OAuth2 providers`);
+}
+
+if (oauth2TargetProviders.length === 0) {
+    console.log('No changed OAuth2 providers found. Nothing to do.');
+    process.exit(0);
```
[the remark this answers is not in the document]
hassan254-prog: `providers.scopes.yaml` is generated output managed by this script, it's not intended to be edited directly. The validation intentionally runs only when providers.yaml changes i.e when new providers are added.

packages/webapp/src/components-v2/ScopesInput.tsx
```diff
@@ -57,61 +68,179 @@ export const ScopesInput: React.FC<ScopesInputProps> = ({ scopesString, onChange
             await onChange?.(newScopes.join(','), countDifference);
```
kaposke, line 71: Should we move `setLoading(true)` to above this line instead of calling it everywhere before `onValueChange`? Looks like this is the only async work it does (

packages/webapp/src/components-v2/ScopesInput.tsx
```diff
@@ -57,61 +68,179 @@ export const ScopesInput: React.FC<ScopesInputProps> = ({ scopesString, onChange
...
-            <InputGroup className={cn(scopes && scopes.length > 0 && 'rounded-b-none')}>
-                <InputGroupInput
-                    value={inputValue}
-                    onChange={(e) => setInputValue(e.target.value)}
-                    onKeyDown={(e) => e.key === 'Enter' && onSubmit()}
-                    disabled={loading || isSharedCredentials || readOnly}
-                    placeholder={finalPlaceholder}
-                />
+    const deleteAllScopes = async () => {
+        setLoading(true);
```
kaposke: To reference my other comment, we would remove it from here and have it within `onValueChange`

packages/webapp/src/components-v2/ScopesInput.tsx
kaposke: Looks like it was here (within `onValueChange`, previously `onSubmit`) before, so maybe you had a reason to move it outside.

packages/webapp/src/components-v2/ScopesInput.tsx
```diff
@@ -57,61 +68,179 @@ export const ScopesInput: React.FC<ScopesInputProps> = ({ scopesString, onChange
...
-};
 
-export const ScopeItem: React.FC<{ scope: string; onDelete: () => void; isSharedCredentials?: boolean; readOnly?: boolean }> = ({
-    scope,
-    onDelete,
-    isSharedCredentials,
-    readOnly
-}) => {
+    if (isSharedCredentials || readOnly) {
+        return (
```
kaposke, line 120: We should probably still have the copy all button in this case

packages/webapp/src/components-v2/ScopesInput.tsx
```diff
@@ -57,61 +68,179 @@ export const ScopesInput: React.FC<ScopesInputProps> = ({ scopesString, onChange
...
-            {!isSharedCredentials && !readOnly && (
-                <Button variant="ghost" size="icon" onClick={onDelete}>
-                    <Trash2 />
-                </Button>
+        <div className="relative min-w-0">
+            {scopes.length > 0 && !loading && (
+                <div className="absolute -top-6 right-0 flex items-center gap-1">
+                    <Tooltip>
+                        <TooltipTrigger asChild>
+                            <button type="button" onClick={copyScopes} className="text-text-tertiary hover:text-text-primary flex items-center p-0.5">
```
kaposke: Can you try using the existing `CopyButton` component?

packages/webapp/src/components-v2/ScopesInput.tsx
```diff
@@ -57,61 +68,179 @@ export const ScopesInput: React.FC<ScopesInputProps> = ({ scopesString, onChange
...
+                                <Copy size={12} />
```
kaposke: Size is incorrect, but using the component will have the right size

packages/webapp/src/components-v2/ScopesInput.tsx
```diff
@@ -57,61 +68,179 @@ export const ScopesInput: React.FC<ScopesInputProps> = ({ scopesString, onChange
...
+                    </Tooltip>
+                    <Tooltip>
+                        <TooltipTrigger asChild>
+                            <button
+                                type="button"
+                                onClick={() => void deleteAllScopes()}
+                                className="text-text-tertiary hover:text-text-primary flex items-center p-0.5"
+                            >
+                                <Trash2 size={12} />
+                            </button>
```
kaposke: Use Button component:
```tsx
<Button size="icon" variant="ghost" onClick={() => void deleteAllScopes()}>
  <Trash />
</Button>
```

Icon size will be handled by the component and it should match the CopyButton component

packages/webapp/src/components-v2/ScopesInput.tsx
```diff
@@ -57,61 +68,179 @@ export const ScopesInput: React.FC<ScopesInputProps> = ({ scopesString, onChange
...
+                        ) : (
+                            <Tooltip>
+                                <TooltipTrigger asChild>
+                                    <button
+                                        type="button"
+                                        onClick={() => void addScopesFromText(inputValue)}
+                                        className="text-text-tertiary hover:text-text-primary flex items-center p-0.5"
+                                    >
+                                        <CornerDownLeft size={14} />
+                                    </button>
```
kaposke: Always use Button component (reference example from previous comment).
The component also has a "loading" that might replace this whole `loading ? ...` condition here

packages/webapp/src/components-v2/ScopesInput.tsx
```diff
@@ -57,61 +68,179 @@ export const ScopesInput: React.FC<ScopesInputProps> = ({ scopesString, onChange
...
+                        className="rounded-t-none shadow-none ring-0 border border-t-0 border-border-muted bg-bg-subtle flex flex-col"
+                    >
+                        {inputValue.trim() && (
+                            <button
+                                type="button"
+                                onClick={() => void addScopesFromText(inputValue)}
+                                className="flex w-full items-center gap-2 px-2 py-1.5 text-sm text-text-secondary hover:bg-dropdown-bg-hover hover:text-text-primary shrink-0"
+                            >
+                                <span className="inline-flex h-6 items-center gap-1 rounded-md bg-bg-elevated px-2 text-xs font-medium text-text-primary shrink-0">
+                                    <Plus size={11} />
```
kaposke: The design shows `14` as the size (figma can be confusing)
Image

Also, please use tailwind for sizes instead of lucide's size prop. In this case "size-3.5" (every tailwind unit is `4px`):

```suggestion
                                    <Plus className="size-3.5" />
```

You also had this in other places but in those I suggested using the `Button` component which handle icon sizing.

kaposke (changes requested):
[Review of the webapp changes]
I really like the new scopes component, works super well!

Just a few requests on component usage and icon sizing.

And I feel like the "add <custom>" item could scroll with the rest of the list like Bastien mentioned (unless you tried and it doesn't feel good?)
Image

scripts/validation/providers/sync-scopes.ts
```diff
@@ -0,0 +1,889 @@
...
+    if (timedOut) {
+        console.warn(`  Batch timed out. Missed: ${missing.sort().join(', ')}`);
+    }
+
+    return { completed, missing };
+}
+
+function chunkArray<T>(arr: T[], size: number): T[][] {
+    const chunks: T[][] = [];
+    for (let i = 0; i < arr.length; i += size) {
```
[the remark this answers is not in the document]
hassan254-prog: not validating for non-positive values since NaN/0 would produce 0 batches (no-op) rather than an infinite loop

hassan254-prog:
I’ve updated the “Add” item to scroll with the rest. I initially thought anchoring it at the top looked better, but we can revisit this later if needed.

packages/webapp/src/components-v2/ScopesInput.tsx
```diff
@@ -53,65 +66,178 @@ export const ScopesInput: React.FC<ScopesInputProps> = ({ scopesString, onChange
...
+                        <TooltipTrigger asChild>
+                            <span>
+                                <CopyButton text={scopes.join(',')} />
+                            </span>
+                        </TooltipTrigger>
+                        <TooltipContent side="top">Copy all</TooltipContent>
+                    </Tooltip>
+                    <Tooltip>
+                        <TooltipTrigger asChild>
+                            <Button type="button" size="icon" variant="ghost" onClick={() => void deleteAllScopes()}>
```
kaposke, line 161: I don't think you need this `type="button"` everywhere? 
```suggestion
                            <Button size="icon" variant="ghost" onClick={() => void deleteAllScopes()}>
```

kaposke (approved):

packages/webapp/src/components-v2/ScopesInput.tsx
hassan254-prog: The default type for this button is "submit", without `type="button"`, clicking this button would trigger form submission.

PR #5702, opened 2026-03-25
feat(auth): introspect refresh token JWT expiry for TWO_STEP providers

hassan254-prog:

- This triggers a proactive refresh before the refresh token itself expires, preventing auth failures. Uses the earliest expiry between access token and refresh token JWT `exp` claim, controlled by a new `refresh_token_introspect_expiry` flag in token_response.

---

**Add refresh token JWT expiry introspection for TWO_STEP credentials**

This PR updates `connectionService.parseRawCredentials` to derive `expires_at` from the refresh token JWT `exp` claim when it is earlier than the access token expiry, using `REFRESH_MARGIN_MS` to trigger earlier refresh. It also adds unit coverage for multiple refresh token shapes and updates a `TWO_STEP` provider configuration to include `token_expires_in_ms`.

---

### DIFF
--- a/packages/providers/providers.yaml
+++ b/packages/providers/providers.yaml
@@ -16909,7 +16909,6 @@ timify:
         token_expiration: expires
         token_expiration_strategy: expireIn
         refresh_token: refreshToken
-        refresh_token_introspect_expiry: true
     proxy:
         base_url: https://${connectionConfig.subdomain}.timify.com
         headers:
--- a/packages/providers/providers.yaml
+++ b/packages/providers/providers.yaml
@@ -15480,6 +15480,7 @@ shopworks:
         content-type: application/json
     token_response:
         token: id_token
+    token_expires_in_ms: 3600000
     proxy:
         base_url: https://${connectionConfig.hostname}
         headers:
@@ -16908,6 +16909,7 @@ timify:
         token_expiration: expires
         token_expiration_strategy: expireIn
         refresh_token: refreshToken
+        refresh_token_introspect_expiry: true
     proxy:
         base_url: https://${connectionConfig.subdomain}.timify.com
         headers:
--- a/packages/shared/lib/services/connection.service.ts
+++ b/packages/shared/lib/services/connection.service.ts
@@ -1048,7 +1048,7 @@ class ConnectionService {
                     expiresAt = new Date(Date.now() + DEFAULT_INFINITE_EXPIRES_AT_MS);
                 }
 
-                if (refreshToken && template.token_response.refresh_token_introspect_expiry) {
+                if (refreshToken) {
                     const decoded = jwtClient.decode(refreshToken);
                     if (decoded && typeof decoded['exp'] === 'number') {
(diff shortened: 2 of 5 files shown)


packages/providers/providers.yaml
```diff
@@ -15480,6 +15480,7 @@ shopworks:
         content-type: application/json
     token_response:
         token: id_token
+    token_expires_in_ms: 3600000
```
hassan254-prog, line 15483: unrelated to this pr, this was added untested, but I decoded the jwt and found that the tokens expire every hour.

packages/shared/lib/services/connection.service.ts
```diff
@@ -1047,6 +1048,16 @@ class ConnectionService {
                     expiresAt = new Date(Date.now() + DEFAULT_INFINITE_EXPIRES_AT_MS);
                 }
 
+                if (refreshToken && template.token_response.refresh_token_introspect_expiry) {
```
TBonnin: should we always introspect the jwt to check for the exp claim?

hassan254-prog: Yes, with Timify, we need to do that since the refresh_token `expire_at` value is not included in the token request response, and having a const value in the providers config means we need to keep track of when the refresh_token has changed so that we can now track the new token with the `refresh_token_expire_in_ms` const from providers config, because the `refresh_token` only changes when its near to expire.

TBonnin: I mean do we need yet another provider parameter instead of always checking the 'exp' claim?

hassan254-prog: Yes, this is an intentionally opt-in. Not all providers return refresh tokens as JWTs and require us to monitor their expiration. Currently, timify is the only provider that requires us to track both access and refresh token expirations.

TBonnin: how expensive would it be to do it anyway and check the exp claim if specified. With opt-in it requires to react and modify the provider definition. If we do it all the time we support jwt exp claim out-of-the box for all providers that specify it

hassan254-prog: I agree, we can drop the explicit check and just verify whether the refresh token has an `exp` claim and track that too. I’ve updated that in the latest [commit](https://github.com/NangoHQ/nango/pull/5702/commits/80bf30d8212d22d115e74e7c400c65f21e5da079).

packages/providers/providers.yaml
TBonnin, line 15483: do we not get this value from the exp claim anyway?

TBonnin (approved):

packages/providers/providers.yaml
hassan254-prog: Yes, I’ll follow up this pr with a cleanup to remove the `token_expires_in_ms` property from the once we can derive from the `exp` claim. The current pr only improves handling for `refresh_tokens`, not the access token.

PR #5703, opened 2026-03-25
fix(runner): add Redis-based abort polling for multi-replica runners

pfreixes:
When runners have replicas > 1, pods sit behind a K8s Service that load-balances requests. The HTTP-based tRPC abort call may hit the wrong pod, leaving the task running. This adds Redis polling for the abort flag (already set by Jobs) when RUNNER_CONFLICT_RESOLUTION_MODE is REDIS, matching the existing lambda-runner pattern.

 

---

It also centralizes the KV store instance and exposes flags to control abort polling in the runner server.

---

### DIFF
--- a/packages/runner/lib/server.ts
+++ b/packages/runner/lib/server.ts
@@ -7,10 +7,10 @@ import superjson from 'superjson';
 
 import { abort } from './abort.js';
 import { jobsClient } from './clients/jobs.js';
-import { heartbeatIntervalMs } from './env.js';
+import { abortCheckIntervalMs, heartbeatIntervalMs } from './env.js';
 import { exec } from './exec.js';
 import { logger } from './logger.js';
-import { abortControllers, locks, usage } from './state.js';
+import { abortControllers, abortViaRedis, kvStore, locks, usage } from './state.js';
 
 import type { NangoProps } from '@nangohq/types';
 import type { NextFunction, Request, Response } from 'express';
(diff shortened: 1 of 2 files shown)


rossmcewan (approved):

packages/runner/lib/server.ts
```diff
@@ -73,6 +73,22 @@ function startProcedure() {
...
+                              if (shouldAbort) {
+                                  logger.info('Aborting task via Redis poll', { taskId });
+                                  abortController.abort();
+                                  clearInterval(abortPoll!);
+                              }
+                          } catch (err) {
+                              logger.error('Error checking abort flag', { taskId, error: err });
+                          }
+                      }, abortCheckIntervalMs)
+                    : null;
```
TBonnin, line 90: we are adding a setInterval and redis query for each task every 1000ms. Can it be a problem resource-wise and redis-wise?

pfreixes: From my understanding this is the same interval that we use for [lambdas](https://github.com/NangoHQ/nango/blob/master/packages/lambda-runner/lib/index.ts#L6), so we already have a consistent workload coming from there with this interval, the server load seems pretty [low](https://us3.datadoghq.com/s/e94cfb60-2611-11ee-93ab-da7ad0900003/f4t-gag-u3s).

So, regardless if we start using this in the runners all of this workload will appear eventually from our lambdas (once we do the full migration)

Still this heartbeat is only enabled for runners that have more than one replica.

I was going to check how much extra load enabling would be coming from runners having replica > 1, ill try to gather some data.

TBonnin (approved):

PR #5705, opened 2026-03-25
fix: Fix not being able to see some templates

marcindobry:
Robin noticed that when he had a `hubspot/users` sync previously deployed (and disabled now), the new `hubspot/sync-users` template wasn't even available to turn on. This was the case for a few other syncs too.

The reason for that is we are filtering templates that sync models with the same name as any other previously deployed sync. Since syncs can't influence models from other syncs, this doesn't seem to make sense.

---

This PR updates flow template deduplication so templates are no longer filtered out solely due to sharing model names with previously deployed syncs, and adds an integration test to ensure renamed templates remain visible when legacy syncs exist.

---

### DIFF
--- a/packages/server/lib/controllers/v1/integrations/providerConfigKey/flows/getFlows.integration.test.ts
+++ b/packages/server/lib/controllers/v1/integrations/providerConfigKey/flows/getFlows.integration.test.ts
@@ -1,6 +1,6 @@
-import { afterAll, beforeAll, describe, expect, it } from 'vitest';
+import { afterAll, beforeAll, describe, expect, it, vi } from 'vitest';
 
-import { seeders } from '@nangohq/shared';
+import { flowService, seeders } from '@nangohq/shared';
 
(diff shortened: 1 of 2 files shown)


TBonnin (approved):

packages/server/lib/controllers/v1/integrations/providerConfigKey/flows/getFlows.integration.test.ts
```diff
@@ -124,4 +124,43 @@ describe(`GET ${route}`, () => {
         const scriptListFile = res.json.data.flows.find((value) => value.name === 'list-files');
         expect(scriptListFile).not.toBeUndefined();
     });
+
+    it('should keep renamed templates visible when an older sync with different endpoint exists', async () => {
```
kaposke: Isn't it supposed to be "when an older sync with the same model name exists"? (And same reflected on the test)

PR #5708, opened 2026-03-25
perf(records): avoid pg_column_size on unchanged records during upsert

TBonnin:
Since #5700, json is no longer overwritten when the data_hash is identical. 
However, pg_column_size(json) was still being computed for all records in both the existing and upsert CTEs, including unchanged records where the delta is always 0.

The fix pre-computes has_changed in CTE query so that previous_size_bytes is only computed for records whose json actually changed. Post-upsert sizes are then fetched in a separate query limited to inserted/changed records only. A separate query is required because we don't want to use pg_column_size in upsert CTE RETURNING clause where it would be applied to all rows. We cannot get the new size after the upsert CTE because of PG snapshot isolation. This will be avoidable with Postgres 17's RETURNING OLD/NEW syntax.

 

---

It also recalculates deltaSizeInBytes based on the newly fetched size map for inserted or changed records.

---

### DIFF
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -32,11 +32,12 @@ dayjs.extend(utc);
 const BATCH_SIZE = envs.RECORDS_BATCH_SIZE;
 
 interface UpsertResult {
+    partition: string;
     external_id: string;
     id: string;
     last_modified_at: string;
     previous_last_modified_at: string | null;
-    delta_size_bytes: number;
+    previous_size_bytes: number;
     status: 'inserted' | 'changed' | 'undeleted' | 'deleted' | 'unchanged';
 }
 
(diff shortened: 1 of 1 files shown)


packages/records/lib/models/records.ts
```diff
@@ -340,38 +341,37 @@ export async function upsert({
...
-                                    'deleted_at',
-                                    'updated_at',
-                                    trx.raw('pg_column_size(json) as size_bytes'),
-                                    trx.raw('tableoid::regclass as partition')
+                                    `${RECORDS_TABLE}.external_id`,
+                                    `${RECORDS_TABLE}.deleted_at`,
+                                    `${RECORDS_TABLE}.updated_at`,
+                                    trx.raw(`
+                                        CASE
+                                            WHEN incoming.data_hash IS DISTINCT FROM ${RECORDS_TABLE}.data_hash THEN pg_column_size(json)
```
TBonnin, line 359: The previous size is only computed for rows where the hash has changed

packages/records/lib/models/records.ts
```diff
@@ -407,18 +408,36 @@ export async function upsert({
...
+                        // Fetch post-upsert json sizes for inserted/changed records to compute delta_size_bytes.
+                        // BUT only for records that have actually changed (based on data_hash comparison)
+                        // to avoid unnecessary pg_column_size calls since they can be costly on large json payloads.
+                        // This must be a separate query since data-modifying CTEs use snapshot isolation.
+                        // Returning the delta from 'upsert' would be possible with Postgres17 since it supports old/new syntax but we are currently using Postgres16.
+                        const toSize = res.filter((r) => r.status === 'inserted' || r.status === 'changed').map((r) => r.external_id);
+                        const newSize = new Map<string, number>();
+                        if (toSize.length > 0) {
+                            const sizeRows = await trx
+                                .select<{ external_id: string; size_bytes: number }[]>('external_id', trx.raw('pg_column_size(json) as size_bytes'))
```
TBonnin, line 432: Second pg_column_size (size after the upsert) is also only applied on rows that changed/new. Avoiding costly operation when the json hasn't changed, especially when the json is toasted

marcindobry (approved):

PR #5711, opened 2026-03-25
chore: move authz types to shared package

kaposke:
Proposal to move authz types to the types package. Moves `permissions` to a new `authz` package.

The `authz` package is very small containing only the permission definitions (for now at least).

Q: why not moving `permissions` to `shared` or `utils`?
A: because they are not imported by the frontend, for good reasons (they are heavy on backend specific stuff).

Lmk if you have better ideas.

---

It also updates server imports to consume the new authz package and wires the new package into TypeScript project references and package dependencies across the server and webapp.

---

### DIFF
--- a/packages/webapp/package.json
+++ b/packages/webapp/package.json
@@ -28,6 +28,7 @@
         "@hookform/resolvers": "5.2.2",
         "@mantine/core": "7.12.1",
         "@mantine/prism": "5.10.5",
+        "@nangohq/authz": "file:../authz",
         "@nangohq/frontend": "0.69.46",
         "@nangohq/types": "file:../types",
         "@radix-ui/react-accordion": "1.2.2",
--- a/package-lock.json
+++ b/package-lock.json
@@ -4805,6 +4805,10 @@
             "resolved": "packages/account-usage",
             "link": true
         },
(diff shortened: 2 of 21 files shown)


rossmcewan (approved):

packages/webapp/package.json
```diff
@@ -28,6 +28,7 @@
         "@hookform/resolvers": "5.2.2",
         "@mantine/core": "7.12.1",
         "@mantine/prism": "5.10.5",
+        "@nangohq/authz": "file:../authz",
```
TBonnin, line 31: why do we need an authz package and not import the type from the `types` package directly?

kaposke: Types did move to the `types` package. The authz package currently holds exclusively `packages/authz/lib/permissions.ts` which is not a type. The types package currently [only emits type declarations](https://github.com/NangoHQ/nango/blob/415ad9e1dcf409fba85bbc8e3e64458b63a8101e/packages/types/tsconfig.json#L7), so I didn't want to break that rule for this. And the other shared packages (shared, utils) are backend heavy. 

I agree it seems overkill but I couldn't think of another way. Accepting suggestions

PR #5719, opened 2026-03-26
fix(auth): handle expired OAuth code in SSO callback gracefully

pfreixes:
When WorkOS returns an invalid_grant error (expired/invalid OAuth code), redirect users to the signin page with a friendly error message instead of showing raw JSON in the browser.

Example of the message that will be displayed to the user

Screenshot 2026-03-26 at 11 46 06

Users that might take too long to finish the flow, we have a few, will have the chance to start over the flow.

For invalid tokens - not expired -  the expectation is that this should happen only bc of bugs at workos side, but if we have concerns on swallowing issues in that specific area we can create a specific monitor which checks that we still have a healthy number of 200s

 

---

It also routes the failure through a sign-in error flag and ensures the sign-in page consumes that error parameter and clears it after display so the message does not persist on refresh.

---

### DIFF
--- a/packages/server/lib/controllers/v1/account/managed/getCallback.ts
+++ b/packages/server/lib/controllers/v1/account/managed/getCallback.ts
@@ -45,10 +45,23 @@ export const getManagedCallback = asyncWrapper<GetManagedCallback>(async (req, r
     const workos = getWorkOSClient();
 
     // Check the request against WorkOS
-    const { user: authorizedUser, organizationId } = await workos.userManagement.authenticateWithCode({
-        clientId: process.env['WORKOS_CLIENT_ID'] || '',
-        code: query.code
-    });
+    let authorizedUser;
+    let organizationId;
+    try {
+        const authResponse = await workos.userManagement.authenticateWithCode({
+            clientId: process.env['WORKOS_CLIENT_ID'] || '',
+            code: query.code
+        });
+        authorizedUser = authResponse.user;
+        organizationId = authResponse.organizationId;
+    } catch (err) {
+        const isInvalidGrant = err instanceof Error && (err as { error?: string }).error === 'invalid_grant';
+        if (isInvalidGrant) {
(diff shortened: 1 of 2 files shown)


packages/server/lib/controllers/v1/account/managed/getCallback.ts
```diff
@@ -45,10 +45,23 @@ export const getManagedCallback = asyncWrapper<GetManagedCallback>(async (req, r
...
     // Check the request against WorkOS
-    const { user: authorizedUser, organizationId } = await workos.userManagement.authenticateWithCode({
-        clientId: process.env['WORKOS_CLIENT_ID'] || '',
-        code: query.code
-    });
+    let authorizedUser;
+    let organizationId;
+    try {
+        const authResponse = await workos.userManagement.authenticateWithCode({
+            clientId: process.env['WORKOS_CLIENT_ID'] || '',
```
rossmcewan: Could this use `envs.WORKOS_CLIENT_ID` instead of `process.env`

rossmcewan (approved):

packages/server/lib/controllers/v1/account/managed/getCallback.ts
pfreixes: yeps, done!

packages/webapp/src/pages/Account/Signin.tsx
```diff
@@ -34,9 +34,25 @@ export const Signin: React.FC = () => {
     const navigate = useNavigate();
     const { toast } = useToast();
 
-    const [errorMessage, setServerErrorMessage] = useState('');
+    const [searchParams, setSearchParams] = useSearchParams();
+    const ssoError = searchParams.get('error');
```
kaposke: NIT would just call this error instead
```suggestion
    const error = searchParams.get('error');
```

kaposke (approved):

PR #5720, opened 2026-03-26
fix(lambda): added test-invocation schema

rossmcewan:
---

**Add Lambda readiness-check invocation schema and handling**

This PR introduces a readiness-check invocation type for the Lambda runner by expanding the schema to include a `readiness_check` request and updating the handler to short-circuit these events with an `{ ok: true }` response. It also adds a readiness-check invocation after alias creation in the Lambda job runner to validate new functions before registering them with the fleet.

---

### DIFF
--- a/packages/jobs/lib/runner/lambda.ts
+++ b/packages/jobs/lib/runner/lambda.ts
@@ -10,6 +10,7 @@ import {
     CreateAliasCommand,
     CreateFunctionCommand,
     DeleteFunctionCommand,
+    InvokeCommand,
     LambdaClient,
     PublishVersionCommand,
     PutFunctionEventInvokeConfigCommand,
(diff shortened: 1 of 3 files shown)


packages/lambda-runner/lib/schemas.ts
```diff
@@ -110,3 +110,13 @@ export const requestSchema = z
...
     })
     .and(z.union([inlineCodeSchema, refCodeSchema]));
+
+export const readinessCheckSchema = z.object({
+    type: z.literal('readiness_check')
+});
+
+export const lambdaInvocationSchema = z.union([requestSchema, readinessCheckSchema]);
+
+export type TaskRequest = z.infer<typeof requestSchema>;
```
TBonnin: can we rename requestSchema to `taskRequestSchema` or even maybe call them `FunctionExecutionRequest` and `functionExecutionSchema`?

packages/jobs/lib/runner/lambda.ts
```diff
@@ -156,6 +157,22 @@ class Lambda {
                         }
                     })
                 );
+                const command = new InvokeCommand({
+                    FunctionName: aResult.AliasArn,
+                    Payload: JSON.stringify({
+                        type: 'readiness_check'
```
TBonnin, line 163: I would make this a `LambdaRequestType` type so we don't have hardcoded string floating around

TBonnin (approved):

PR #5722, opened 2026-03-26
feat(webapp): RBAC UI

kaposke:
This adds RBAC gates to buttons/switches/pages. It still doesn't add role management or anything new.
The way to test it currently is to manually update your role in the database to one of:
- `administrator` -> full access
- `production_support` -> full access to dev envs, read access to non-sensitive prod data
- `development_full_access` -> full access to dev envs, zero access to prod

For the most part, this is disabling and adding a tooltip to specific components.
I've created a `ConditionalTooltip` component, which is a convenient way wrapping something in a tooltip given a condition, and it also handles nested tooltips (eg: you need a button to be disabled on 2 different situations, but show 2 different tooltips for each condition, but not both at the same time. This is well handled by `ConditionalTooltip`).
And to go further, `PermissionGate` is a wrapper around the `ConditionalTooltip` that adds the default message to the displayed tooltip, and also gives you a handle to use to disable the inner component, whatever that is. Example usage:
```tsx
<PermissionGate condition={canDeleteEnvironment}> {/** Conditionally wraps with the default "you can't do this" tooltip */}
  {(allowed) => // Handle to use to disable inner components and what not
    <Button disabled={!allowed} onClick={deleteEnv}>
      Delete environment
    </Button>
  }
</PermissionGate>
```

---

It also adds a centralized permissions hook and a route-level guard to enforce access beyond UI controls, extends credential/secret displays with read/edit gating, introduces an unauthorized environment page, and adjusts shared UI primitives to support permission-aware disabled behavior across the app.

---

### DIFF
--- a/packages/webapp/src/components-v2/AppSidebar/EnvironmentDropdown.tsx
+++ b/packages/webapp/src/components-v2/AppSidebar/EnvironmentDropdown.tsx
@@ -12,6 +12,7 @@ import { Tooltip, TooltipContent, TooltipTrigger } from '../ui/tooltip.js';
 import { LogoInverted } from '@/assets/LogoInverted';
 import { useEnvironment } from '@/hooks/useEnvironment';
 import { useMeta } from '@/hooks/useMeta';
+import { permissions } from '@/hooks/usePermissions.js';
 import { useStore } from '@/store';
 
 export const EnvironmentDropdown: React.FC = () => {
@@ -72,11 +73,7 @@ export const EnvironmentDropdown: React.FC = () => {
                     <DropdownMenuContent align="start" side="bottom" className="w-50 max-h-96 flex flex-col gap-2">
                         <div className="flex flex-col">
                             {meta?.environments.map((environment) => (
-                                <PermissionGate
-                                    key={environment.name}
-                                    bypass={!environment.is_production}
-                                    permission={{ action: 'read', resource: 'environment', scope: 'production' }}
-                                >
+                                <PermissionGate key={environment.name} bypass={!environment.is_production} permission={permissions.canAccessProdEnvironment}>
                                     {(allowed) => (
                                         <DropdownMenuItem
                                             disabled={!allowed}
@@ -94,7 +91,7 @@ export const EnvironmentDropdown: React.FC = () => {
                                 </PermissionGate>
                             ))}
                         </div>
-                        <PermissionGate permission={{ action: 'create', resource: 'environment', scope: 'global' }}>
+                        <PermissionGate permission={permissions.canCreateEnvironment}>
                             {(allowed) =>
                                 allowed ? (
                                     <Tooltip delayDuration={0}>
--- a/packages/webapp/src/components-v2/AppSidebar/EnvironmentDropdown.tsx
+++ b/packages/webapp/src/components-v2/AppSidebar/EnvironmentDropdown.tsx
@@ -4,15 +4,15 @@ import { useNavigate } from 'react-router-dom';
 
 import { StyledLink } from '../StyledLink.js';
 import { CreateEnvironmentDialog } from './CreateEnvironmentDialog.js';
+import { ConditionalTooltip } from '../ConditionalTooltip.js';
 import { PermissionGate } from '../PermissionGate.js';
 import { Button } from '../ui/button.js';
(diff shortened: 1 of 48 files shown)


packages/webapp/src/components-v2/AppSidebar/EnvironmentDropdown.tsx
```diff
@@ -71,53 +78,64 @@ export const EnvironmentDropdown: React.FC = () => {
...
+                                    condition={!!isMaxEnvironmentsReached}
+                                    content={
+                                        <>
+                                            Max number of environments reached.{' '}
+                                            {environment?.plan?.name.includes('legacy') ? (
+                                                <>Contact Nango to add more</>
+                                            ) : (
+                                                <>
+                                                    <StyledLink to={`/${env}/team/billing`} className="text-s">
+                                                        Upgrade
```
agusayerza, line 116: Should we show the button to the production support role?

agusayerza: Discussed in person, outermost permission gate will take precedence so this is not a problem

packages/webapp/src/pages/Environment/Settings/Backend.tsx
```diff
@@ -27,6 +31,13 @@ export const BackendSettings: React.FC = () => {
     const { mutateAsync: revertKeyAsync, isPending: isReverting } = useRevertKey(env);
     const { mutateAsync: activateKeyAsync, isPending: isActivating } = useActivateKey(env);
 
+    const isProdEnv = environmentAndAccount?.environment.is_production || false;
+
+    const { can } = usePermissions();
+    const canReadSecretKey = can(permissions.canReadProdSecretKey) || !isProdEnv;
+    const canGenerateNewSecretKey = canReadSecretKey;
```
agusayerza: I think we should have a permission to write the secret key instead of reusing read

`can(permissions.canWriteProdEnvironmentKeys) || !isProdEnv`

packages/webapp/src/components-v2/AppSidebar/ProfileDropdown.tsx
```diff
@@ -21,6 +24,9 @@ export const ProfileDropdown: React.FC = () => {
     const { user } = useUser();
     const showGettingStarted = useStore((state) => state.showGettingStarted);
 
+    const { can } = usePermissions();
+    const canManageBilling = can(permissions.canManageTeam);
```
agusayerza:
```suggestion
    const canManageBilling = can(permissions.canManageBilling);
```

packages/webapp/src/components/PermissionRoute.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+import type { Permission } from '@nangohq/types';
+
+export const PermissionRoute: React.FC<Permission> = (permission) => {
+    const { can } = usePermissions();
+    const { loading } = useUser();
+    const env = useStore((state) => state.env);
+
+    if (loading) return null;
+    if (!can(permission)) return <Navigate to={`/${env}`} replace />;
```
agusayerza, line 15: Should we show a toast on why they were redirected?

packages/webapp/src/pages/Integrations/providerConfigKey/Settings/components/GeneralSettings.tsx
```diff
@@ -27,6 +30,9 @@ export const GeneralSettings: React.FC<{ data: GetIntegration['Success']['data']
     const { confirm, DialogComponent } = useConfirmDialog();
     const { mutateAsync: patchIntegration } = usePatchIntegration(env, integration.unique_key);
 
+    const { can } = usePermissions();
+    const canEdit = !environment.is_production || can(permissions.canWriteProdIntegrations);
```
marcindobry, line 34: Why did you use this specific permission to check for the ability to edit settings?

kaposke: It was just the closes tbh

packages/webapp/src/components/PermissionRoute.tsx
kaposke: Not sure. Not a requirement though

agusayerza (approved):

PR #5724, opened 2026-03-26
fix(webhooks): use headers instead of the request body for webhookscript executions in g-calendar

hassan254-prog:

- Use headers instead of the request body for webhookscript executions in g-calendar, since the body is empty and this is also what we forward.

---

It also removes the unused body parameter from the webhook handler signature to align with the new header-based payload behavior.

---

### DIFF
--- a/packages/server/lib/webhook/google-calendar-webhook-routing.ts
+++ b/packages/server/lib/webhook/google-calendar-webhook-routing.ts
@@ -4,7 +4,7 @@ import { hashEmailAddress } from '../utils/pii.js';
 
 import type { WebhookHandler } from './types.js';
 
-const route: WebhookHandler = async (nango, headers, body) => {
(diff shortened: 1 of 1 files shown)


packages/server/lib/webhook/google-calendar-webhook-routing.ts
agusayerza, line 28: While we are at this, can we also update the function to first try to match with `X-Goog-Resource-ID` I think it will be much more reliable and enable webhooks for not-email owned calendars

agusayerza (approved):

packages/server/lib/webhook/google-calendar-webhook-routing.ts
hassan254-prog: I’ll deploy this as-is for now and continue the discussion on Slack to determine the best approach for identifying connections.

PR #5728, opened 2026-03-27
fix(webapp): conditionally load Stripe only when publishable key exists

### LINKED ISSUE #5640: publicStripeKey: "" crash in env.js for self-hosted instances.
# Self-hosted dashboard blank screen — Stripe.js crashes on empty `publicStripeKey`

## Description

The Nango dashboard renders a completely blank/black screen on self-hosted instances. The root cause is `stripe.js` attempting to initialize with an empty string, which throws an unhandled error that kills the entire React app.

## Console Error

```
Uncaught (in promise) IntegrationError: Please call Stripe() with your publishable key. You used an empty string.
    at stripe.js:1
```

Additionally:
```
The Content Security Policy directive 'upgrade-insecure-requests' is ignored when delivered in a report-only policy.
    at stripe.js:1
```

## Root Cause

The `/env.js` endpoint returns `"publicStripeKey": ""` for self-hosted instances (expected, since self-hosted doesn't use Stripe billing). However, the dashboard frontend code calls `Stripe("")` unconditionally without checking whether the key is a non-empty string first.

**`/env.js` output:**
```json
window._env = {
  "apiUrl": "https://nango.example.com",
  "publicStripeKey": "",
  "isCloud": false,
  "isHosted": true,
  "isEnterprise": false,
  "features": {
    "plan": false
  }
}
```

Despite `isCloud: false` and `features.plan: false`, the Stripe initialization still runs and crashes.

## Environment

- **Image:** `nangohq/nango-server:hosted` (also reproduced on `hosted-0.69.30`)
- **Version:** 0.69.40+ (git hash `5fc90c54d6b4efc23bddd6652fbbc087621ebd8d`)
- **Deployment:** Docker Compose via Coolify on Ubuntu 24
- **Self-hosted:** Yes, free tier

## Steps to Reproduce

1. Deploy `nangohq/nango-server:hosted` with standard self-hosted configuration per [[docs](https://nango.dev/docs/guides/platform/free-self-hosting/configuration)](https://nango.dev/docs/guides/platform/free-self-hosting/configuration)
2. Navigate to the dashboard URL in a browser
3. Observe blank/black screen
4. Open browser DevTools → Console → see `IntegrationError` from `stripe.js`

## Expected Behavior

The dashboard should load normally on self-hosted instances. Stripe initialization should be gated behind a check like:

```javascript
if (window._env.publicStripeKey && window._env.isCloud) {
  loadStripe(window._env.publicStripeKey);
}
```

## Workaround

The API itself is unaffected — `/health` returns `{"result":"ok"}` and all OAuth/proxy/connection endpoints work normally. The issue is isolated to the dashboard UI. Integrations can be managed via the API or `@nangohq/node` SDK as a workaround.

## Impact

This completely blocks dashboard access for all self-hosted users on the current `:hosted` tag. No configuration changes or documented environment variables resolve the issue.

cyphercodes:
## Problem
The Nango dashboard renders a completely blank/black screen on self-hosted instances because stripe.js attempts to initialize with an empty string, throwing an unhandled error that kills the entire React app.

Fixes #5640

## Changes
- Modified `packages/webapp/src/utils/stripe.ts` to conditionally call `loadStripe()` only when `publicStripeKey` is truthy
- Returns `null` instead of calling `loadStripe("")` when the key is empty

## Testing
- This fix prevents the `IntegrationError: Please call Stripe() with your publishable key. You used an empty string.` error
- Self-hosted instances will now load the dashboard properly
- Cloud instances with a valid Stripe key continue to work normally

---

This PR updates Stripe initialization to avoid calling `loadStripe()` with an empty publishable key. When `globalEnv.publicStripeKey` is falsy, `stripePromise` now resolves to `null`, preventing the self-hosted dashboard crash caused by Stripe's empty-key error.

---

### DIFF
--- a/packages/webapp/src/utils/stripe.ts
+++ b/packages/webapp/src/utils/stripe.ts
@@ -3,4 +3,4 @@ import { loadStripe } from '@stripe/stripe-js';
 import { globalEnv } from './env';
 
 const stripePublishableKey = globalEnv.publicStripeKey;
-export const stripePromise = stripePublishableKey ? loadStripe(stripePublishableKey) : null;
+export const stripePromise = stripePublishableKey ? loadStripe(stripePublishableKey) : Promise.resolve(null);
--- a/packages/webapp/src/utils/stripe.ts
+++ b/packages/webapp/src/utils/stripe.ts
@@ -3,4 +3,4 @@ import { loadStripe } from '@stripe/stripe-js';
 import { globalEnv } from './env';
 
 const stripePublishableKey = globalEnv.publicStripeKey;
-export const stripePromise = loadStripe(stripePublishableKey);
+export const stripePromise = stripePublishableKey ? loadStripe(stripePublishableKey) : null;


cyphercodes:
Updated the fix to keep astripePromisea consistently as a Promise to avoid null runtime errors. Changed the fallback from anulla to aPromise.resolve(null)a as suggested by propel-code-bot.

cyphercodes:
Fixed, thanks for the review!

PR #5729, opened 2026-03-27
fix(auth): attach WorkOS error details to DD span in SSO callback

pfreixes:
Add workos.request_id, workos.error_code and workos.error_message as span tags when authenticateWithCode fails, so we can see the full error context in Datadog APM traces without leaking sensitive data.

Also this allows us to provide any request id to Workos for pin point errors on their side

 

---

This also applies the tagging on the active Datadog span within the managed SSO callback when authenticateWithCode fails.

---

### DIFF
--- a/packages/server/lib/controllers/v1/account/managed/getCallback.ts
+++ b/packages/server/lib/controllers/v1/account/managed/getCallback.ts
@@ -1,3 +1,4 @@
+import tracer from 'dd-trace';
 import * as z from 'zod';
 
 import db from '@nangohq/database';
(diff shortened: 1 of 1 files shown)


rossmcewan (approved):

packages/server/lib/controllers/v1/account/managed/getCallback.ts
```diff
@@ -60,6 +61,18 @@ export const getManagedCallback = asyncWrapper<GetManagedCallback>(async (req, r
             res.redirect(`${basePublicUrl}/signin?error=sso_session_expired`);
             return;
         }
+
+        const workosErr = err as { rawData?: { code?: string; message?: string }; requestID?: string };
```
TBonnin, line 65: could we have imported the type from workos package instead of a blind type cast?

PR #5730, opened 2026-03-27
feat(website): replace webflow provider sync with sanity cms sync

hassan254-prog:

- replace webflow providers sync with sanity cms sync

---

**Replace Webflow provider sync with Sanity CMS sync pipeline**

This PR removes the Webflow-based provider sync and introduces a new Sanity CMS sync script that reads `providers.yaml` and provider `.mdx` docs, uploads logos, maps categories, and upserts/deletes Sanity `api` documents. The GitHub Actions workflow is updated to run the new Sanity sync on docs/provider changes, manual triggers, and a monthly schedule, while the scripts package adds the Sanity client dependency and removes `webflow-api`.

---


scripts/sanity-api-sync.ts
```diff
@@ -0,0 +1,176 @@
...
+            }
+
+            neededProviders[provider] = providers[provider];
+        }
+    }
+}
+
+console.log(`Found ${Object.keys(neededProviders).length} providers with docs`);
+
+const sanityCategories = await sanity.fetch<{ _id: string; slug: string }[]>(`*[_type == "apiCategory"]{ _id, "slug": slug }`);
```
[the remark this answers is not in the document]
hassan254-prog: The slug fields in both `api` and `apiCategory` schemas are defined as `type: 'string' `(plain strings), not Sanity's native `type: 'slug'`.

rossmcewan (commented):
I'm a little bit concerned about how this will play with the automated build of the site based on document changes. If forceUpdate is true then it will update each document - each of these operations will trigger a website build and deploy. We either need to disable the webhook to do this or come up with another solution for keeping the website and generated docs in sync.

hassan254-prog:
@rossmcewan I’ve pushed an update to improve this. We now disable the webhooks, run a batch update using the `sanity.transaction()` method, and then re-enable the webhooks at the end.

scripts/sanity-api-sync.ts
```diff
@@ -0,0 +1,282 @@
...
+        try {
+            const tx = sanity.transaction();
+            for (const doc of documents) {
+                tx.createOrReplace(doc);
+            }
+            for (const api of toDelete) {
+                tx.delete(api._id);
+            }
+            // remove all drafts — providers.yaml is the source of truth, draft content is discarded
+            for (const draft of draftApis) {
```
rossmcewan: Why do we delete the drafts? Is there no chance that someone is working on a doc and this would nuke it?

hassan254-prog: From my understanding, when we migrated from webflow, we had some providers that were left in draft mode. That’s why we were never able to publish the entire catalogue. The `delete draft` step will ensure we remove those drafts so that `providers.yaml` become the only source of truth.

The `draftApis` selection, using the `_type: "api"` field, ensures that we only pick draft APIs, which are currently updated exclusively by the publish script in the Nango repo.

rossmcewan (approved):

PR #5732, opened 2026-03-27
fix(webapp): allow for scopes to be defined when creating an oauth2_cc integration

hassan254-prog:

- Allow scopes to be defined when creating an `oauth2_cc` integration. This enables customers to specify scopes at the [integration](https://github.com/NangoHQ/nango/blob/master/packages/server/lib/controllers/oauth.controller.ts#L448) level, rather than only when creating a connect session.

---

**Add `OAUTH2_CC` scope configuration in integration settings and API**

This PR adds support for defining `OAUTH2_CC` scopes at the integration level across the web UI, server validation, and API types. It introduces a dedicated `OAuth2CCSettings` component with a `ScopesInput` control that persists scopes via PATCH, and wires this component into the auth-specific settings routing.

---

### DIFF
--- a/packages/webapp/src/components-v2/ScopesInput.tsx
+++ b/packages/webapp/src/components-v2/ScopesInput.tsx
@@ -47,6 +47,12 @@ export const ScopesInput: React.FC<ScopesInputProps> = ({ scopesString, onChange
         const newScopes = unique([...scopes, ...scopesToAdd]);
         const countDifference = newScopes.length - scopes.length;
 
+        if (countDifference === 0) {
+            setInputValue('');
+            setLoading(false);
+            return;
+        }
+
         try {
             await onChange?.(newScopes.join(','), countDifference);
             setScopes(newScopes);
--- a/packages/webapp/src/pages/Integrations/providerConfigKey/Settings/components/OAuthSettings.tsx
+++ b/packages/webapp/src/pages/Integrations/providerConfigKey/Settings/components/OAuthSettings.tsx
@@ -41,7 +41,7 @@ export const OAuthSettings: React.FC<{ data: GetIntegration['Success']['data']; 
     const onSave = async (field: Partial<PatchIntegration['Body']>, supressToast = false) => {
         try {
             await patchIntegration({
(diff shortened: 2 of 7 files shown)


packages/webapp/src/pages/Integrations/providerConfigKey/Settings/components/OAuthSettings.tsx
```diff
@@ -97,61 +97,68 @@ export const OAuthSettings: React.FC<{ data: GetIntegration['Success']['data'];
...
-                            secret={!canEdit}
-                        />
-                        {isEditingClientId && hasExistingClientId && (
-                            <Alert variant="warning">
-                                <AlertTriangle />
-                                <AlertDescription>
-                                    Updating the Client ID will invalidate token refreshes for all existing connections for this integration.
-                                </AlertDescription>
-                            </Alert>
+            {/* Callback URL (not applicable for OAUTH2_CC) */}
```
TBonnin: is there a cleaner way that doesn't require to filter `auth_mode !== 'OAUTH2_CC` cc @kaposke

hassan254-prog: I have updated this with the latest [commit](https://github.com/NangoHQ/nango/pull/5732/commits/e4837b37a8a2ba7a90307ad1432e6e6d171b38ff) to split it into its own separate settings component.

packages/webapp/src/components-v2/ScopesInput.tsx
```diff
@@ -47,6 +47,12 @@ export const ScopesInput: React.FC<ScopesInputProps> = ({ scopesString, onChange
         const newScopes = unique([...scopes, ...scopesToAdd]);
         const countDifference = newScopes.length - scopes.length;
 
+        if (countDifference === 0) {
```
hassan254-prog, line 50: This fixes a small bug where, if a user submitted a scope that already already existed, `countDifference` would be 0 due to deduplication via `unique()`. This incorrectly triggered the `else` branch in `handleScopesChange`, resulting in the message “Scope successfully removed.”

TBonnin (approved):

PR #5733, opened 2026-03-27
feat(server): track missing OAuth state cookies to assess CSRF fix re-enablement impact

agusayerza:
## Problem
We shipped OAuth2 callback state cookie validation as a CSRF fix (#5366, #5488, #5491) but had to revert it (#5493) because some customers' connectors proxy the OAuth redirect and silently drop cookies, breaking their auth flow.

Before re-enabling the hard validation, we need visibility into how many customers and which integrations would actually fail the check.

## Solution
Re-add the state cookie (set on OAuth request, checked on callback) but in observation-only mode (the check never blocks the flow). When the expected cookie is absent at callback time, we emit a `nango.server.auth.callback.state_cookie.missing` metric tagged with `account_id` and `auth_mode`

This covers all OAuth2 callbacks

NAN-5134

---

In addition to restoring observation-only state-cookie checks, this change standardizes that behavior across the relevant OAuth callback variants and adds the callback cookie-parsing/telemetry plumbing needed to reliably measure missing-cookie occurrences before any strict enforcement is reintroduced.

---

### DIFF
--- a/packages/server/lib/controllers/oauth.controller.ts
+++ b/packages/server/lib/controllers/oauth.controller.ts
@@ -1180,10 +1180,7 @@ class OAuthController {
 
             if (req.cookies[`oauth2-${session.id}`] !== '1') {
                 metrics.increment(metrics.Types.AUTH_CALLBACK_STATE_COOKIE_MISSING, 1, {
-                    provider: config.provider,
-                    integration: config.unique_key,
-                    account_id: account.id,
-                    auth_mode: session.authMode
+                    account_id: account.id
                 });
             }
 
--- a/packages/server/lib/controllers/oauth.controller.ts
+++ b/packages/server/lib/controllers/oauth.controller.ts
@@ -296,7 +296,16 @@ class OAuthController {
                 await this.mcpOauth2Request({ provider: provider as ProviderMcpOAUTH2, config, session, req, res, connectionConfig, callbackUrl, logCtx });
                 return;
             } else if (provider.auth_mode === 'MCP_OAUTH2_GENERIC') {
-                await this.mcpGenericRequest({ provider: provider as ProviderMcpOAuth2Generic, config, session, res, connectionConfig, callbackUrl, logCtx });
+                await this.mcpGenericRequest({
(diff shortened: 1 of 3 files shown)


packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -1151,6 +1174,15 @@ class OAuthController {
             const config = (await configService.getProviderConfig(session.providerConfigKey, session.environmentId))!;
             await logCtx.enrichOperation({ integrationId: config.id!, integrationName: config.unique_key, providerName: config.provider });
 
+            if (req.cookies[`oauth2-${session.id}`] !== '1') {
+                metrics.increment(metrics.Types.AUTH_CALLBACK_STATE_COOKIE_MISSING, 1, {
+                    provider: config.provider,
+                    integration: config.unique_key,
+                    account_id: account.id,
+                    auth_mode: session.authMode
+                });
```
TBonnin, line 1208: that's potentially a lot of dimensions, which given datadog pricing model can become very expensive very quickly. Can we only track the account id?

TBonnin (approved):
one comment about the dd metric dimensions that must be addressed but the rest looks good

packages/server/lib/controllers/oauth.controller.ts
agusayerza: Keeping only the `account_id`

packages/server/lib/controllers/oauth.controller.ts
```diff
@@ -1147,6 +1174,12 @@ class OAuthController {
             const config = (await configService.getProviderConfig(session.providerConfigKey, session.environmentId))!;
             await logCtx.enrichOperation({ integrationId: config.id!, integrationName: config.unique_key, providerName: config.provider });
 
+            if (req.cookies[`oauth2-${session.id}`] !== '1') {
+                metrics.increment(metrics.Types.AUTH_CALLBACK_STATE_COOKIE_MISSING, 1, {
```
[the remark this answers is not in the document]
agusayerza: Updated description

PR #5742, opened 2026-03-30
perf(records): writing records payload to dedicated table

TBonnin:
The majority of records being upserted haven't changed (ie: hash is the same as the one already stored). Unmodified records represents 90% of the upserts and 70% of the total size.
In order to prevent writing the payload for each upsert, we are now storing it in a dedicated table and only upsert the payload when it has changed. 
The changes made in this commit makes reading and writing modified records slower (double write and join in reads) but optimize for the most common scenario which is upserting unmodified records.
Old `json` column in records table is now set to null and new/changed records payload stored in the `records_data` table.
Logic takes into account that records payload might still exist in the records table and make sure to insert it in the new table when it is the case

Most of the complexity comes from keeping track of the records count/size deltas. I wish it could be done out-of-bound but refreshing count/size from scratch would take too long on big datasets. Any better idea is welcome though

---

It also updates read behavior to join against `records_data` with a legacy fallback, recalculates size/count deltas across operations, and adjusts delete/prune handling to align with the new storage model.

---


packages/records/lib/models/records.integration.test.ts
kaposke, line 1: Would love to see tests for the records.json row migration

kaposke (commented):
Couldn't spot any mistakes. Would feel a bit safer if the json migration including the size delta calculation was covered with some tests.

packages/records/lib/models/records.integration.test.ts
TBonnin: https://github.com/NangoHQ/nango/pull/5742/changes/559a4292e0c8e14f5d06ef244ef7c963a3907dec

marcindobry (commented):
So is the idea that we won't do an actual data migration but just let data retention do it's thing - eventually untouched data will get deleted and we can drop the old .json column and code paths that touch it?

TBonnin:
> So is the idea that we won't do an actual data migration but just let data retention do it's thing - eventually untouched data will get deleted and we can drop the old .json column and code paths that touch it?

Exactly. We have TB of records, a one time migration would be too slow/costly. Also hard to handle with self-hosted. 
With autopruning, in 30 days most of the records will be migrated or pruned.

agusayerza (approved):
LGTM. Autoprunning removes the record data even if the sync is still active, right? (The 60 days rule only applies for hard delete of sync data, correct?)

marcindobry (approved):

TBonnin:
> LGTM. Autoprunning removes the record data even if the sync is still active, right? (The 60 days rule only applies for hard delete of sync data, correct?)

yes. autopruning is only looking at the freshness of a record. autodelete looks at the last execution of the sync.

PR #5743, opened 2026-03-30
feat(integrations): add support for apple business manager

hassan254-prog:

- add support for apple business manager, untested.

---

**Add `apple-business-manager` integration with `JWT` assertion support in `TWO_STEP` auth**

This PR introduces a new `apple-business-manager` provider and extends the shared `TWO_STEP` authentication flow to support multiple assertion types. The previous SAML-only assertion path is generalized by renaming `packages/shared/lib/auth/samlAssertion.ts` to `packages/shared/lib/auth/assertion.ts`, splitting logic into `generateSamlAssertion()` and new `generateJwtAssertion()`, and routing assertion generation by `provider.assertion.type` in `packages/shared/lib/services/connection.service.ts`.

It also adds end-to-end integration documentation and assets for Apple Business Manager, including a new integration page, connect guide, generated snippets, docs navigation entry, and logo asset. Provider typing is expanded in `packages/types/lib/providers/provider.ts` to include assertion `type` and JWT-specific `header`/`payload` fields, and an existing SAML provider (`sap-success-factors`) is updated to set `assertion.type: saml` explicitly.

---


packages/shared/lib/auth/assertion.ts
```diff
@@ -317,3 +318,57 @@ export function generateAssertion({
...
+        const payload: Record<string, any> = {
+            iat: now,
+            exp: now + lifetimeInSeconds
+        };
+        for (const [k, v] of Object.entries(providerAssertion.payload ?? {})) {
+            payload[k] = interpolateString(v, interpolationValues);
+        }
+
+        const privateKey = generatePrivateKey(signingKey);
+        const token = signJWT({ payload, secretOrPrivateKey: privateKey, options: { header } });
```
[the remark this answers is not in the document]
hassan254-prog: this is already defined in the providers config.

docs/api-integrations/apple-business-manager/created_api.png
marcindobry, line 1: Does this come from some Apple guide or are we by accident exposing a businesses data?

packages/shared/lib/services/connection.service.ts
```diff
@@ -1352,12 +1352,21 @@ class ConnectionService {
...
 
-            const create = samlClient.generateAssertion({
-                provider,
-                dynamicCredentials: credentials,
-                connectionConfig,
-                ...(assertionOption && { assertionOption })
-            });
+            const assertionType = provider.assertion.type ?? 'saml';
+            const create =
+                assertionType === 'jwt'
```
marcindobry, line 1357: Don't other JWT integrations need this? Why is it specific to Apple Business Manager?

packages/shared/lib/services/connection.service.ts
```diff
@@ -1352,12 +1352,21 @@ class ConnectionService {
             const { assertionOption: assertionOptionValue, ...credentials } = dynamicCredentials;
             const assertionOption = assertionOptionValue as Record<string, any> | undefined;
 
-            const create = samlClient.generateAssertion({
-                provider,
-                dynamicCredentials: credentials,
-                connectionConfig,
-                ...(assertionOption && { assertionOption })
-            });
+            const assertionType = provider.assertion.type ?? 'saml';
```
marcindobry: Instead of this default, can we just add it explicitly in providers.yaml to all the providers that would fall under this category?

marcindobry (changes requested):

packages/shared/lib/services/connection.service.ts
hassan254-prog: Yes, I’ve updated it with the latest [commit](https://github.com/NangoHQ/nango/pull/5743/commits/411112c69042960a0f437922973ecdcbbd11e771). This is only used by SAP SuccessFactors.

packages/shared/lib/services/connection.service.ts
hassan254-prog: With other JWT providers, we generate the JWT, store it, and use it to authenticate API calls. However, for providers that use a two-step flow with assertion generation (i.e saml or jwt), we first generate an assertion, use it to request a token, and then use that token to authenticate API calls.

docs/api-integrations/apple-business-manager/created_api.png
hassan254-prog: Yes, this came from another provider’s public guide on how to generate these. I used it since we don’t have access to an account. I’ve also updated the images to hide any personal information.

marcindobry: Can we at least blur out the company name and account id in this one?

marcindobry: And in docs/api-integrations/apple-business-manager/api_option.png

hassan254-prog: thanks for flagging, I missed this one, it should be updated now with the latest commit.

marcindobry (approved):

PR #5745, opened 2026-03-31
feat(connectwise-psa): Support custom ConnectWise PSA hostnames

justonf:
## What changed

- Added an optional `hostname` connection config for `connectwise-psa`.
- Updated the provider base URL to prefer `hostname` and fall back to the existing `subdomain.myconnectwise.net` behavior.
- Added proxy URL unit coverage for both the custom-hostname path and the legacy subdomain fallback.

## Why

Some ConnectWise PSA tenants use custom hostnames instead of the standard `*.myconnectwise.net` pattern. The provider previously always built the cloud hostname from `subdomain`, which broke those tenants.

## Impact

- Existing connections that only have `subdomain` continue to use the existing URL shape.
- New or updated connections can provide `hostname` for custom-domain tenants.
- The ConnectWise config UI text now makes the custom-hostname vs cloud-subdomain behavior explicit.

## Validation

- Added targeted unit tests for hostname precedence and subdomain fallback.
- Manually validated against a local Nango instance with:
- a custom-hostname connection
- a standard cloud-subdomain connection

## Notes

- This change is provider-config only; it does not require a database migration.

---

In addition to the provider URL behavior update, this PR also includes a broader documentation realignment for ConnectWise PSA: content is moved into the `api-integrations` structure, an end-user connect guide is introduced, and related navigation/update references are adjusted so new onboarding flows emphasize `hostname` while legacy `subdomain` usage remains supported.

---


packages/providers/providers.yaml
```diff
@@ -4061,20 +4061,30 @@ connectwise-psa:
...
+        hostname:
+            type: string
+            title: Custom Hostname
+            description: Optional custom ConnectWise PSA hostname to use instead of the cloud subdomain. Enter the hostname only, without protocol or path.
+            example: psa.example.com
+            format: hostname
+            prefix: https://
+            suffix: /v4_6_release/apis/3.0
+            optional: true
         subdomain:
```
hassan254-prog: When deprecating one input in favor of another, we typically remove the previous input to prevent users from creating a connection without providing either. In this case, the hostname will be required, and the subdomain field will be removed from the connect ui.
This should also reflect in the connect ui docs, on how to get the hostname.

hassan254-prog (commented):
Thanks for the contribution, I just have one comment on how to handle this change.

packages/providers/providers.yaml
```diff
@@ -4061,20 +4061,30 @@ connectwise-psa:
...
     webhook_routing_script: connectwisePsaWebhookRouting
     proxy:
-        base_url: https://${connectionConfig.subdomain}.myconnectwise.net/v4_6_release/apis/3.0
+        base_url: https://${connectionConfig.hostname}/v4_6_release/apis/3.0 || https://${connectionConfig.subdomain}.myconnectwise.net/v4_6_release/apis/3.0
         headers:
             accept: application/vnd.connectwise.com+json; version=${connectionConfig.apiVersion}
             clientid: ${connectionConfig.clientId}
     docs: https://nango.dev/docs/integrations/all/connectwise-psa
     connection_config:
+        hostname:
```
hassan254-prog, line 4221: I would remove the subdomain input field from the connect ui and make the hostname a required field to prevent end users from creating a connection without providing either value, since both are currently optional.
This should also reflect in the connect ui docs on how to get the full hostname.

justonf: removed subdomain as an option after testing.

packages/providers/providers.yaml
```diff
@@ -4061,20 +4061,30 @@ connectwise-psa:
...
+            type: string
+            title: Hostname
+            description: The full ConnectWise PSA hostname. Enter the hostname only, without protocol or path. For example, use api-na.myconnectwise.net for standard cloud tenants or your custom PSA hostname if applicable.
+            example: api-na.myconnectwise.net
+            format: hostname
+            prefix: https://
```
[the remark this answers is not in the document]
justonf: not needed - former connections with subdomain only will fallback to hostname. going forward, future connections should only use hostname

docs/integrations/all/connectwise-psa-staging.mdx
```diff
@@ -43,7 +43,7 @@ _No setup guide yet._
 - To access the API documentation, you will need to log in to their developer [portal](https://developer.connectwise.com/).
 - ConnectWise PSA uses BASIC authentication. The username will always begin with the CompanyId followed by either the public key, integrator username, or MemberId. The password will be the private key, integrator password, or member hash.
 - When creating a new connection on Nango, you will need to include the `apiVersion` in the format `2019.1`. For more details, visit [Versioning](https://developer.connectwise.com/Products/ConnectWise_PSA/Developer_Guide#Versioning)
-- You will also need to provide the `subdomain` in the format `api-{region}`, with the region being specified (see [Authentication](https://developer.connectwise.com/Products/ConnectWise_PSA/Developer_Guide#Authentication))
+- You will also need to provide the full `hostname` for your ConnectWise PSA staging tenant. For standard staging tenants, this is usually `api-staging.connectwisedev.com`. If your staging account uses a custom domain, use that hostname instead.
```
hassan254-prog: This is not needed since you are only updating the `ConnectWise PSA` provider.

hassan254-prog (approved):
Changes looks good, one final comment on docs.

packages/providers/providers.yaml
```diff
@@ -4095,20 +4095,20 @@ connectwise-psa:
...
-        subdomain:
+        hostname:
             type: string
-            title: ConnectWise PSA subdomain
-            description: The subdomain to connect to ConnectWise PSA
-            pattern: '^api-(au|eu|na)$'
-            example: api-au
-            suffix: .myconnectwise.net
+            title: Hostname
+            description: The full ConnectWise PSA hostname. Enter the hostname only, without protocol or path. For example, use api-na.myconnectwise.net for standard cloud tenants or your custom PSA hostname if applicable.
```
hassan254-prog:
```suggestion
            description: The full ConnectWise PSA hostname.
```

The other info will be mentioned within the docs.

PR #5760, opened 2026-04-01
fix(locks): fix kv locks

rossmcewan:
---

**Fix `KVLocks` race conditions with atomic `KVStore` compare-and-mutate operations**

This PR is a significant lock correctness update focused on preventing stale-owner races in `KVLocks`. The lock implementation in `packages/runner/lib/sdk/locks.ts` is rewritten to use `KVStore` directly (instead of `Locking`), with stable logical lock keys (`runner:lock:` + hash of logical key), owner value checks, and companion owner-index keys to support safe acquire/refresh/release/release-all behavior.

To support this, the `KVStore` interface is extended with atomic compare-and-mutate APIs (including companion-key variants), implemented in both `packages/kvstore/lib/InMemoryStore.ts` and `packages/kvstore/lib/RedisStore.ts` (Redis via Lua `EVAL` scripts). Runtime wiring in `packages/lambda-runner/lib/index.ts` now passes `KVStore` to `KVLocks`, and tests were substantially expanded across lock/store unit and integration suites to validate concurrency and race regressions.

---


packages/runner/lib/sdk/locks.ts
```diff
@@ -20,57 +22,96 @@ export interface Locks {
...
+            return validation;
+        }
+
+        const sk = this.storageKey(key);
         try {
-            await this.locking.tryAcquire(lockKey, ttlMs, 1000);
+            const current = await this.store.get(sk);
+            if (current !== null) {
+                if (current === owner) {
+                    await this.store.set(sk, owner, { canOverride: true, ttlMs });
```
TBonnin: I think we have a race condition here where the lock could have been acquired elsewhere after the `get` on L61 and then overwritten on L64.

packages/runner/lib/sdk/locks.ts
```diff
@@ -20,57 +22,96 @@ export interface Locks {
...
     }
 
     public async releaseAllLocks({ owner }: { owner: string }): Promise<Result<void>> {
-        const lockKey = this.getLockKey(owner);
         try {
-            await this.locking.releaseAll(lockKey);
+            for await (const sk of this.store.scan(`${LOCK_STORAGE_PREFIX}*`)) {
+                const holder = await this.store.get(sk);
+                if (holder === owner) {
+                    await this.store.delete(sk);
```
TBonnin: same race condition between the get/delete here

TBonnin (changes requested):

packages/runner/lib/sdk/locks.ts
```diff
@@ -20,57 +22,96 @@ export interface Locks {
...
-            return Err(new Error(`Error releasing lock for key ${lockKey}`, { cause: err }));
+            return Err(new Error(`Error releasing lock for key ${key}`, { cause: err }));
         }
     }
 
     public async releaseAllLocks({ owner }: { owner: string }): Promise<Result<void>> {
-        const lockKey = this.getLockKey(owner);
         try {
-            await this.locking.releaseAll(lockKey);
+            for await (const sk of this.store.scan(`${LOCK_STORAGE_PREFIX}*`)) {
```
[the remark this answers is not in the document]
TBonnin: full scan can indeed be very expensive

packages/runner/lib/sdk/locks.ts
rossmcewan: Changed approach.

packages/runner/lib/sdk/locks.ts
rossmcewan: Changed approach.

packages/runner/lib/sdk/locks.ts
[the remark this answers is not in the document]
rossmcewan: Hard one to solve for. Could keep a secondary key that has a reference to the locks this owner has but complicates the writes. Given these aren't used much this is unlikely to be an issue but absolutely could be if usage were to increase. Thoughts?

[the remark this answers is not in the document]
rossmcewan: Maybe introduce environment id as part of the key to reduce the amount of items that would be scanned?

packages/kvstore/lib/RedisStore.ts
```diff
@@ -1,6 +1,31 @@
...
+const SET_IF_VALUE_EQUALS_LUA = `
+local v = redis.call('GET', KEYS[1])
+if v == false then
+  return 0
+end
+if v ~= ARGV[1] then
+  return 0
+end
+redis.call('SET', KEYS[1], ARGV[2], 'PX', tonumber(ARGV[3]))
+return 1
```
TBonnin, line 11: you don't need the second `if`, when the key doesn't exist GET returns false and `v ~= ARGV[1] will always be true
```
local v = redis.call('GET', KEYS[1])
if v ~= ARGV[1] then
  return 0
end
redis.call('SET', KEYS[1], ARGV[2], 'PX', tonumber(ARGV[3]))
return 1
```

packages/kvstore/lib/RedisStore.ts
```diff
@@ -1,6 +1,31 @@
 import type { KVStore } from './KVStore.js';
 import type { RedisClientType } from 'redis';
 
+/** Atomically refresh TTL only if the current value still matches (same-owner lock refresh). */
+const SET_IF_VALUE_EQUALS_LUA = `
```
TBonnin: `COMPARE_AND_SET`  (and `COMPARE_AND_DEL`) to keep the [CAS](https://en.wikipedia.org/wiki/Compare-and-swap) acronym? ;)

packages/kvstore/lib/InMemoryStore.ts
```diff
@@ -46,6 +46,30 @@ export class InMemoryKVStore implements KVStore {
...
 
+    public async setIfValueEquals(key: string, expectedValue: string, newValue: string, ttlMs: number): Promise<boolean> {
+        const res = this.store.get(key);
+        if (res === undefined || this.isExpired(res)) {
+            return Promise.resolve(false);
+        }
+        if (res.value !== expectedValue) {
+            return Promise.resolve(false);
+        }
+        this.store.set(key, { value: newValue, timestamp: Date.now(), ttlMs: ttlMs || 0 });
```
TBonnin: ttlMs is mandatory param so `|| 0` is dead code

packages/kvstore/lib/RedisStore.ts
```diff
@@ -32,6 +57,22 @@ export class RedisKVStore implements KVStore {
         }
     }
 
+    public async setIfValueEquals(key: string, expectedValue: string, newValue: string, ttlMs: number): Promise<boolean> {
+        const n = await this.client.eval(SET_IF_VALUE_EQUALS_LUA, {
```
TBonnin: does that mean the script is send to Redis every time it is being executed? Could it be a problem under heavy load?

packages/runner/lib/sdk/locks.ts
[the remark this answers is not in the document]
TBonnin: we can prefix by owner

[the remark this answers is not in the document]
rossmcewan: Including owner in the key was the problem we were trying to fix. How would prefixing by owner work here?

packages/kvstore/lib/RedisStore.ts
rossmcewan: I think we're very far away from redis being a problem.

packages/runner/lib/sdk/locks.ts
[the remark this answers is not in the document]
TBonnin: Sorry I was referring to your secondary key.
Can we add an optional owner argument to `compare_and_set`
if set write atomically another key `owner:$owner:$thekey` with the same ttl. Then the `releaseAll` can scan with prefix `owner:$owner:*`, extract the key and delete both of them.

That's 2 internal entries per lock but fast release

packages/runner/lib/sdk/locks.ts
```diff
@@ -57,26 +83,27 @@ export class KVLocks implements Locks {
         }
 
         const sk = this.storageKey(key);
+        const ik = this.ownerIndexKey(owner, key);
```
TBonnin: since they are related I would have a single function that return both keys: `const {sk, ik} = this.keys({key, owner})`

packages/kvstore/lib/RedisStore.ts
```diff
@@ -23,6 +23,45 @@ redis.call('DEL', KEYS[1])
 return 1
 `;
 
+const COMPARE_AND_SET_NX_WITH_COMPANION = `
+local ok = redis.call('SET', KEYS[1], ARGV[1], 'NX', 'PX', tonumber(ARGV[2]))
+if not ok then
+  return 0
+end
+local ok2 = redis.call('SET', KEYS[2], ARGV[3], 'PX', tonumber(ARGV[2]))
+if not ok2 then
```
TBonnin: can `ok2` ever be `falsy`?

packages/kvstore/lib/InMemoryStore.ts
```diff
@@ -46,6 +46,77 @@ export class InMemoryKVStore implements KVStore {
         return Promise.reject(new Error('set_key_already_exists'));
     }
 
+    public async setIfValueEquals(key: string, expectedValue: string, newValue: string, ttlMs: number): Promise<boolean> {
```
TBonnin, line 49: other functions have `ttlsMs` in an `opts` object

packages/kvstore/lib/InMemoryStore.ts
```diff
@@ -46,6 +46,77 @@ export class InMemoryKVStore implements KVStore {
...
+        return Promise.resolve(true);
+    }
+
+    public async setIfValueEqualsWithCompanion(
+        mainKey: string,
+        companionKey: string,
+        expectedValue: string,
+        newValue: string,
+        companionValue: string,
+        ttlMs: number
```
TBonnin: personal opinion: arg list longer than 2 makes signature hard to decipher when reading calling site: 
```
await this.store.setIfValueEqualsWithCompanion(sk, ik, owner, owner, LOCK_OWNER_INDEX_VALUE, ttlMs)
```
I would prefer an object param (with named fields) but it probably mean changing existing functions to be consistent

TBonnin (approved):

PR #5775, opened 2026-04-02
fix(proxy): add deny list for base url override in proxy

rossmcewan:
---

**Enforce `base-url-override` Denylist in Proxy With Redirect Validation and Normalization**

This PR adds server-side enforcement to block disallowed proxy `base-url-override` targets using a new environment variable, `NANGO_PROXY_BASE_URL_OVERRIDE_DENYLIST`. It introduces robust hostname normalization for denylist matching (case-insensitive hostnames, trailing dots, bracketed IPv6 handling, and URL/bare-entry normalization), applies checks early in `allPublicProxy`, and returns `400` with `error.code` `base_url_override_not_allowed` when blocked.

It also closes a redirect-based bypass by validating redirect destinations via `validateProxyRedirectUrl` in proxy request configuration, wiring this through shared proxy config/types, and surfacing denied redirects as the same `400` response. The PR adds observability with a new metric and warning logs, expands env parsing and docs for denylist configuration, and includes targeted unit tests for normalization, deny behavior, redirect validation, and error-chain handling.

---

### DIFF
--- a/packages/server/lib/controllers/proxy/allProxy.ts
+++ b/packages/server/lib/controllers/proxy/allProxy.ts
@@ -39,7 +39,6 @@ const MEMOIZED_CONNECTION_TTL = 60000;
 
 const logger = getLogger('Proxy.Controller');
 
-/** Parsed once at module load; `envs` comes from `process.env` at process start (restart to change denylist). */
 const baseUrlOverrideDenylist = normalizeDenylist(envs.NANGO_PROXY_BASE_URL_OVERRIDE_DENYLIST);
 
 const schemaHeaders = z.object({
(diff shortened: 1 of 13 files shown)


packages/server/lib/controllers/proxy/allProxy.ts
```diff
@@ -60,17 +64,40 @@ export const allPublicProxy = asyncWrapper<AllPublicProxy>(async (req, res, next
         res.status(400).send({ error: { code: 'invalid_headers', errors: zodErrorToHTTP(valHeaders.error) } });
         return;
     }
+    const parsedHeaders = valHeaders.data satisfies AllPublicProxy['Headers'];
     const { environment, account, plan } = res.locals;
 
+    const baseUrlOverride = parsedHeaders['base-url-override'];
+    if (baseUrlOverride && isBaseUrlOverrideDenied(baseUrlOverride, baseUrlOverrideDenylist)) {
+        metrics.increment(metrics.Types.PROXY_BASE_URL_OVERRIDE_DENIED, 1, { accountId: account.id });
```
TBonnin: do we need this metric?

packages/server/lib/controllers/proxy/baseUrlOverrideDenylist.ts
```diff
@@ -0,0 +1,62 @@
...
+    const seen = new Set<string>();
+    for (const e of denylist) {
+        const h = normalizeDenylistHost(e);
+        if (h && !seen.has(h)) {
+            seen.add(h);
+            out.push(h);
+        }
+    }
+    return out;
+}
```
TBonnin, line 53: let's return a set for O(1) lookup at runtime
```
export function normalizeDenylist(denylist: string[]): Set<string> {
   return new Set(denylist.map(normalizeDenylistHost).filter(Boolean));
}
```

packages/server/lib/controllers/proxy/baseUrlOverrideDenylist.ts
```diff
@@ -0,0 +1,62 @@
...
+ */
+export function canonicalizeHostnameForDenylist(host: string): string {
+    let h = host.trim().toLowerCase();
+    if (h.startsWith('[') && h.endsWith(']')) {
+        h = h.slice(1, -1);
+    }
+    while (h.endsWith('.')) {
+        h = h.slice(0, -1);
+    }
+    return h;
```
TBonnin, line 18: what about octal, hexa or other unusual format supported by node `URL`?

TBonnin (approved):

packages/server/lib/controllers/proxy/allProxy.ts
rossmcewan: I thought it would be worth possibly alerting on as it implies malicious intent.

PR #5787, opened 2026-04-03
feat(webhooks): add support for webhooks from multiple calendars

rbwest:
Contio requires Google Calendar webhooks from multiple calendars per connection, not just the primary calendar. This PR updates the webhook routing and associated docs to make this possible. It also adds the same routing support to google integration since Contio uses both google-calendar and google integrations for calendar syncing. 

[NAN-5160: Contio - Add support for webhooks from multiple calendars](https://linear.app/nango/issue/NAN-5160/contio-add-support-for-webhooks-from-multiple-calendars)

1. Create new google calendar connection
2. Update connection metadata as described in updated docs with appropriate `googleCalendarWatchResourceUris`
3. Subscribe to push notifications for those calendars as outlined in [current docs](https://nango.dev/docs/api-integrations/google-calendar/webhooks)
4. Create event on multiple calendars and see appropriate connections associated with each webhook

---

This change formalizes metadata-first webhook resolution by matching incoming `X-Goog-Resource-URI` values against configured watch resource URIs, while preserving backward compatibility for existing `google-calendar` primary-calendar flows through legacy matching behavior when metadata does not resolve a connection. It also establishes consistent routing support across both `google-calendar` and base `google` integrations so multi-calendar notifications can be reliably associated with the correct connection setup.

---


docs/api-integrations/google-calendar/webhooks.mdx
```diff
@@ -289,8 +293,37 @@ curl -X PATCH "https://api.nango.dev/connection/metadata" \
 
 To do this in bulk, iterate over the [list connections](/reference/api/connection/list) response and [update the metadata](/reference/api/connection/update-metadata) for each one.
 
+### Multiple calendars
+
+If you want to subscribe to notifications from multiple calendars, add the resource URIs for each calendar to **`metadata.googleCalendarWatchResourceUris`**:
```
agusayerza: Multiple calendars or any other calendar that isn't the primary, right?

agusayerza: Are we able to know which calendars we have access at the connection post script event? Would we be able to resolve them to store them in the connection config as well? (OOS for this PR, but its something to consider when thinking about the webhooks refactor)

docs/api-integrations/google-calendar/webhooks.mdx
```diff
@@ -289,8 +293,37 @@ curl -X PATCH "https://api.nango.dev/connection/metadata" \
...
+      "googleCalendarWatchResourceUris": [
+        "https://www.googleapis.com/calendar/v3/calendars/cal1@example.com/events?alt=json",
+        "https://www.googleapis.com/calendar/v3/calendars/cal2@example.com/events?alt=json"
+      ]
+    }
+  }'
+```
+This will associate both listed calendar resource URIs with the connection. Nango will match any incoming webhook with those exact `X-Goog-Resource-URI` values to this connection.
+
+To do this in bulk, iterate over the [list connections](/reference/api/connection/list) response and [update the metadata](/reference/api/connection/update-metadata) for each one.
```
agusayerza, line 322: Not sure about this line, it really depends on use-case.

docs/api-integrations/google-calendar/webhooks.mdx
```diff
@@ -289,8 +293,37 @@ curl -X PATCH "https://api.nango.dev/connection/metadata" \
...
+
+
 <Warning>
-If Nango cannot match the incoming webhook to a connection (because it can't find a matching hash/email in `connection_config` or `metadata`), the webhook will still be received but won't include a `connectionId` in the forwarded payload.
+If Nango cannot match the incoming webhook to a connection, the webhook will still be received but won't include a `connectionId` in the forwarded payload.
```
agusayerza:
```suggestion
If Nango cannot match the incoming webhook to a connection, the webhook will still be forwarded but won't include a `connectionId` in the payload.
```

docs/snippets/generated/salesforce-cc/PreBuiltUseCases.mdx
```diff
@@ -1,3 +1,3 @@
 _No pre-built syncs or actions available yet._
 
-<Tip>Not seeing the integration you need? [Build your own](/guides/primitives/functions) independently.</Tip>
+<Tip>Not seeing the integration you need? [Build your own](/guides/primitives/functions) independently.</Tip>
```
agusayerza: Maybe I'm blind, but whats the change here? Also, does something change for `salesforce-cc` ?

packages/server/lib/webhook/google-calendar-webhook-routing.unit.test.ts
```diff
@@ -28,36 +31,73 @@ describe('googleCalendarWebhookRouting', () => {
...
+
+        const headers = {
+            'x-goog-resource-uri': EXAMPLE_RESOURCE_URI,
+            'x-goog-resource-state': 'exists'
+        };
+
+        await GoogleCalendarWebhookRouting.default(nangoMock as unknown as InternalNango, headers as any, {}, '');
+
+        expect(mock).toHaveBeenCalledTimes(1);
+        expect(mock).toHaveBeenCalledWith(
```
marcindobry, line 96: Instead of just matching whether a mock has been called deep inside with an argument, how feasible is it to set up a test that actually has multiple connections with metadata and see what the result of the routing is?

packages/providers/providers.yaml
```diff
@@ -7475,6 +7475,7 @@ google:
             response_path: items
     docs: https://nango.dev/docs/api-integrations/google
     setup_guide_url: https://nango.dev/docs/api-integrations/google/how-to-register-your-own-google-api-oauth-app
+    webhook_routing_script: googleWebhookRouting
```
agusayerza, line 7478: We marked other google apis as now supporting webhooks (`google-ads`, `google-analytics`, `google-bigquery`, `google-chat`, etc) but we haven't added the routing for them? Do webhooks for those only work when connected through Google Platform? Or the docs update is wrong?

rbwest: Oh, sorry, I didn't take a close look at the auto generated updates. It will only work for `google` or `google-calendar`. How do I prevent the auto generated updates?

docs/api-integrations/google-calendar/webhooks.mdx
rbwest: I'm not following. What do you mean?

docs/api-integrations/google-calendar/webhooks.mdx
rbwest: > Multiple calendars or any other calendar that isn't the primary, right?

No, it's exclusive. You either match by the listed `googleCalendarWatchResourceUris`, or we will follow the previous path and assume you want the primary calendar. It's possible for Contio's use case that someone doesn't actually want to subscribe to their primary calendar, so you must explicitly list the URI for each calendar you want to do the routing for.


> Are we able to know which calendars we have access at the connection post script event? Would we be able to resolve them to store them in the connection config as well? (OOS for this PR, but its something to consider when thinking about the webhooks refactor)

So the google-calendar connection has access to all the calendars after connection, but we don't know which ones they want to subscribe to notifications for. Specifically for Contio, the user selects this in their application.

1. User connects to google-calendar or google
2. Contio fetches a list of all calendars
3. User selects which calendars from lists they want to sync

So blindly adding all calendars I think is the wrong call and just adds noise, but happy to discuss more.

docs/snippets/generated/salesforce-cc/PreBuiltUseCases.mdx
rbwest: I have no idea. I'm not super familiar with how these auto generated snippets are created. I assumed it was best practice to just commit them, but maybe not?

What should I be doing in these cases?

packages/server/lib/webhook/google-calendar-webhook-routing.unit.test.ts
rbwest: Not sure, let me take a look. I was just following the established pattern of unit tests for simplicity. Is this a blocker?

marcindobry: Nope, I’ll approve to unblock since it’s following an existing pattern but IMO it would be a worthy addition :)

marcindobry (approved):

docs/api-integrations/google-calendar/webhooks.mdx
agusayerza: For the first part I meant that if you want to subscribe to calendar "USA Holidays" instead of Ryan's primary (ryan@nango.dev) then you also need to follow this pattern, so not only multiple calendars but rather non-primary ones

agusayerza: Agreed on the second note, was just wondering

docs/api-integrations/google-calendar/webhooks.mdx
agusayerza: This is a VERY minor NIT, but I feel like the note is rather specific. Ideally most of the times you are already doing this on a post connect script, so this sounds more of a migration instruction than a setting up multiple calendars. Feel free to ignore the comment if you don't agree

packages/providers/providers.yaml
agusayerza: Did a deep dive on this: it’s showing up because of the `alias` value. I wouldn’t claim we support webhooks for those integrations if we don’t actually route them. The only case where `alias` is correct for webhook routing today is `confluence` inheriting from `jira`.

I’m planning to stop inheriting `webhook_routing_script` at alias merge time in a separate PR (as part of the webhooks groundwork), and only set routing explicitly where it’s real. For example adding `webhook_routing_script: jiraWebhookRouting` on `confluence` so behavior stays the same once merge stops inheriting.

For this PR, I’d revert the generated snippet / provider noise for integrations that don’t actually use this path so we don’t ship misleading docs. Do you agree?

docs/api-integrations/google-calendar/webhooks.mdx
rbwest: Ah, gotcha. Updated the title `Other calendars` and clarified this point in the docs.

docs/api-integrations/google-calendar/webhooks.mdx
rbwest: I copied this phrasing from the initial GCal webhook docs and before that from the Gmail docs you put together I believe 😄 

I think it's fine to leave since we call out doing this in the post connection script in step 4.

packages/providers/providers.yaml
rbwest: Agree. Reverted all of the changes for now, including google, since we only support webhooks there in a very specific capacity for Contio.

docs/snippets/generated/salesforce-cc/PreBuiltUseCases.mdx
rbwest: Reverted along with the other auto-generated snippets.

docs/api-integrations/google-calendar/webhooks.mdx
```diff
@@ -289,8 +293,37 @@ curl -X PATCH "https://api.nango.dev/connection/metadata" \
...
+    "provider_config_key": "google-calendar",
+    "metadata": {
+      "googleCalendarWatchResourceUris": [
+        "https://www.googleapis.com/calendar/v3/calendars/cal1@example.com/events?alt=json",
+        "https://www.googleapis.com/calendar/v3/calendars/cal2@example.com/events?alt=json"
+      ]
+    }
+  }'
+```
+This will associate both listed calendar resource URIs with the connection. Nango will match any incoming webhook with those exact `X-Goog-Resource-URI` values to this connection.
```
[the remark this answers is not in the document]
rbwest: No change necessary. In testing passing the unencoded @ matches as expected with the uri provided in the webhook.

packages/server/lib/webhook/google-calendar-webhook-routing.unit.test.ts
```diff
@@ -28,36 +31,73 @@ describe('googleCalendarWebhookRouting', () => {
...
         const headers = {
-            'x-goog-resource-uri': 'https://www.googleapis.com/calendar/v3/calendars/user%40example.com/events?alt=json',
+            'x-goog-resource-uri': EXAMPLE_RESOURCE_URI,
             'x-goog-resource-state': 'exists'
         };
 
         await GoogleCalendarWebhookRouting.default(nangoMock as unknown as InternalNango, headers as any, {}, '');
 
-        expect(mock).toHaveBeenCalledTimes(3);
+        expect(mock).toHaveBeenCalledTimes(4);
```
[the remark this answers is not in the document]
rbwest: Updated to add early exit for missing resource uri and associated test.

docs/api-integrations/google-calendar/webhooks.mdx
```diff
@@ -289,8 +293,37 @@ curl -X PATCH "https://api.nango.dev/connection/metadata" \
...
+### Other calendars
+
+If you want to subscribe to notifications from other calendars besides the primary calendar, add the resource URIs for each calendar to **`metadata.googleCalendarWatchResourceUris`**:
+
+- The value must be a JSON **`string[]`**. Store the **`X-Goog-Resource-URI` string exactly as Google sends it**: same encoding, path, and query string.
+- The lookup on **`metadata.googleCalendarWatchResourceUris`** takes priority over the primary calendar matching described [above for primary calendars](#primary-calendar), so include the URIs for all calendars you are subscribing to, including the primary calendar.
+
+**Duplicates:** If more than one connection lists the **same** URI in `googleCalendarWatchResourceUris`, **every** matching connection is included. Nango forwards the webhook **once per matching connection**, with appropriate connection id.
+
+```bash
```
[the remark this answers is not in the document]
rbwest: Fixed.

docs/api-integrations/google-calendar/webhooks.mdx
```diff
@@ -289,8 +293,37 @@ curl -X PATCH "https://api.nango.dev/connection/metadata" \
 
 To do this in bulk, iterate over the [list connections](/reference/api/connection/list) response and [update the metadata](/reference/api/connection/update-metadata) for each one.
 
+### Other calendars
+
+If you want to subscribe to notifications from other calendars besides the primary calendar, add the resource URIs for each calendar to **`metadata.googleCalendarWatchResourceUris`**:
+
+- The value must be a JSON **`string[]`**. Store the **`X-Goog-Resource-URI` string exactly as Google sends it**: same encoding, path, and query string.
+- The lookup on **`metadata.googleCalendarWatchResourceUris`** takes priority over the primary calendar matching described [above for primary calendars](#primary-calendar), so include the URIs for all calendars you are subscribing to, including the primary calendar.
```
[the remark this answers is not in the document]
rbwest: This is technically true, but I don't want users relying on two routing mechanisms, so I'm leaving the docs as is to push them towards just using one routing mechanism or the other.

Additionally, this routing will likely change in the near future as we change the way we handle webhook routing.

packages/server/lib/webhook/google-calendar-webhook-routing.ts
```diff
@@ -7,49 +7,68 @@ import type { WebhookHandler } from './types.js';
...
 
     const baseArgs = {
         body: headers,
         ...(headers['x-goog-resource-state'] && { webhookTypeValue: headers['x-goog-resource-state'] })
     };
 
-    const emailAddressHash = emailAddress ? hashEmailAddress(emailAddress) : undefined;
+    // First, try to match the resource URI to the googleCalendarWatchResourceUris (multiple calendar matching)
+    let response =
+        typeof resourceUri === 'string'
```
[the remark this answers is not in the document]
rbwest: Fixed.

packages/server/lib/webhook/google-calendar-webhook-routing.unit.test.ts
rbwest: I looked more at this Friday, but I didn't have time to implement. I'm going to hold off since Agustin is going to be looking at webhook improvements across the board and we may change how this routing is handled in the near future.

PR #5796, opened 2026-04-06
feat: redact connection credentials instead of deleting them

kaposke:
Went with a simple solution, curious on your thoughts.
Just recursively replace credentials values, regardless of which type it is. This way we don't need custom logic for each. Simple "NON_SENSITIVE_KEYS" to keep necessary `type` and any other non-sensitive key.

UI side of this was done already, so now that the UI knows which fields to render, this is what you get when you don't have the necessary permissions:
image
image

---

This PR also formalizes the behavior change so restricted responses preserve credential object shape with masked values rather than returning empty objects, and it aligns the UI by enforcing read-permission gating on credential refresh interactions. Together, these updates improve consistency for consumers while strengthening least-privilege handling across server responses and credential-management views.

---

### DIFF
--- a/packages/webapp/src/pages/Connection/components/AuthTab.tsx
+++ b/packages/webapp/src/pages/Connection/components/AuthTab.tsx
@@ -19,7 +19,7 @@ export const AuthTab = ({ connectionData, providerConfigKey }: { connectionData:
 
     return (
         <div className="flex w-full gap-11 justify-between">
-            <div className="flex flex-col gap-8 max-w-2xl">
+            <div className="flex flex-col gap-8 w-full max-w-2xl">
                 {errorLog && (
                     <Alert variant="error">
                         <Info />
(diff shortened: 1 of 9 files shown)


packages/webapp/src/pages/Connection/components/AuthTab.tsx
```diff
@@ -19,7 +19,7 @@ export const AuthTab = ({ connectionData, providerConfigKey }: { connectionData:
 
     return (
         <div className="flex w-full gap-11 justify-between">
-            <div className="flex flex-col gap-8 max-w-2xl">
+            <div className="flex flex-col gap-8 w-full max-w-2xl">
```
kaposke, line 22: Intentional, was getting some cases where it wouldn't fill the space appropriately.

pfreixes (approved):

PR #5798, opened 2026-04-06
feat: ingest usage events to clickhouse

TBonnin:
This commit introduce clickhouse and ingest usage events to raw table 
It doesn't ingest connections or records usage because they are not event-based and I haven't figure out how I am gonna model those metrics yet.
CLICKHOUSE_URL isn't set in staging/prod. Until then code is noop 
Next steps will be creating materialized views for the metrics.

 

---

This PR also establishes the foundational ingestion pipeline around ClickHouse by integrating migration and ingestion lifecycle handling into the metering flow and introducing buffered event processing so event-based analytics can be captured reliably without disrupting existing billing and usage-tracking paths. It formalizes the initial event-oriented ingestion scope as a staged rollout toward downstream metric modeling, while keeping behavior safely disabled in environments where ClickHouse is not configured.

---

### DIFF
--- a/packages/jobs/lib/execution/action.ts
+++ b/packages/jobs/lib/execution/action.ts
@@ -314,7 +314,7 @@ export async function handleActionSuccess({
                 functionName: nangoProps.syncConfig.sync_name,
                 success: true,
                 telemetryBag,
-                functionRuntime
+                runtime: functionRuntime
             }
         }
     });
@@ -500,7 +500,7 @@ function onFailure({
                     type: 'action',
                     success: false,
                     telemetryBag,
-                    functionRuntime
+                    runtime: functionRuntime
                 }
             }
         });
--- a/packages/metering/lib/crons/usage.ts
+++ b/packages/metering/lib/crons/usage.ts
@@ -81,6 +81,7 @@ const observability = {
                             }
                         }
                     ]);
+                    // TODO: ingest into clickhouse
                     metrics.gauge(metrics.Types.CONNECTIONS_COUNT, count, { accountId });
                 }
             } catch (err) {
@@ -155,6 +156,8 @@ const observability = {
                 if (toBilling.isErr()) {
                     logger.error(`Failed to ingest record billing events`);
                 }
+
+                // TODO ingest into clickhouse
 
                 // send to datadog
                 for (const {
--- a/packages/usage/lib/clickhouse/batcher.ts
+++ b/packages/usage/lib/clickhouse/batcher.ts
@@ -90,6 +90,7 @@ export class Batcher<T> {
             const batchToRetry = batch.filter((item) => item.retries < this.maxProcessingRetry).map((item) => ({ ...item, retries: item.retries + 1 }));
             const dropped = batch.length - batchToRetry.length;
             if (dropped > 0) {
+                // TODO: push metric
                 logger.error(`Clickhouse batcher: dropping ${dropped} items after exhausting retries.`);
             }
             this.queue.unshift(...batchToRetry);
--- a/packages/usage/lib/clickhouse/batcher.ts
+++ b/packages/usage/lib/clickhouse/batcher.ts
@@ -1,6 +1,6 @@
 import { Err, Ok } from '@nangohq/utils';
 
-import { logger } from './logger.js';
+import { logger } from '../logger.js';
 
 import type { Result } from '@nangohq/utils';
 
@@ -11,7 +11,6 @@ interface Item<T> {
 
 export class Batcher<T> {
     private readonly process: (events: T[]) => Promise<void>;
-    private readonly grouping: Grouping<T> | undefined;
     private readonly maxBatchSize: number;
     private readonly flushInterval: number;
     private queue: Item<T>[];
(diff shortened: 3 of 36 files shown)


packages/jobs/lib/execution/action.ts
```diff
@@ -314,7 +314,7 @@ export async function handleActionSuccess({
                 functionName: nangoProps.syncConfig.sync_name,
                 success: true,
                 telemetryBag,
-                functionRuntime
```
TBonnin, line 317: unrelated but functionRuntime properties was added to all events instead of being only a property of the `function_execution` one

packages/usage/lib/clickhouse/migrate.ts
```diff
@@ -0,0 +1,59 @@
...
+                continue;
+            }
+            const mod = (await import(path.join(migrationsDir, file))) as { sql: string };
+            logger.info(`Clickhouse migration: applying ${name}`);
+            await client.command({ query: mod.sql });
+            await client.insert({ table: migrationTable, values: [{ name }], format: 'JSONEachRow' });
+        }
+        logger.info(`Clickhouse migration: ${files.length > 0 ? `applied ${files.length} migration(s)` : ` no migrations`}`);
+        return Ok(undefined);
+    } catch (err) {
```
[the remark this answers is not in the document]
TBonnin: good point. will fix

packages/usage/lib/clickhouse/batcher.ts
```diff
@@ -0,0 +1,123 @@
...
+
+import type { Result } from '@nangohq/utils';
+
+interface Item<T> {
+    item: T;
+    retries: number;
+}
+
+export class Batcher<T> {
+    private readonly process: (events: T[]) => Promise<void>;
```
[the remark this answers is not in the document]
TBonnin: on purpose. Logic is slightly different (no grouping) and if the clickhouse pilot is sucessful the billing one will be removed

packages/billing/lib/batcher.ts
```diff
@@ -116,6 +116,10 @@ export class Batcher<T> {
             // Put failed items back at the front of the queue
             // unless they have been retried too many times
             const batchToRetry = aggregated.filter((item) => item.retries < this.maxProcessingRetry).map((item) => ({ ...item, retries: item.retries + 1 }));
+            const dropped = batch.length - batchToRetry.length;
+            if (dropped > 0) {
```
pfreixes, line 120: maybe have a metric for this? so we can add a monitor which tells us when we are losing billing events?

packages/usage/lib/clickhouse/migrations/20260406174100_create_raw_usage_events.ts
```diff
@@ -0,0 +1,15 @@
...
+        ts               DateTime64(3),
+        idempotency_key  String,
+        type             LowCardinality(String),
+        account_id       Int64,
+        value            Float64,
+        attributes       JSON
+    )
+    ENGINE = ReplacingMergeTree()
+    PARTITION BY toYYYYMM(ts)
+    ORDER BY (account_id, type, idempotency_key)
```
pfreixes, line 13: If Im not mistaken during query time we would be scanning all data within the prefix `account_id, type,` for query that might range between two timestamps (we will be making this kind of queries right?), so wondering if would be worth it to add also as part of the order by something like `(account_id, type, ts, idempotency_key)` considering that the `ts` related to a specific `idempotency_key` will be always the same too.

packages/usage/lib/clickhouse/batcher.ts
```diff
@@ -0,0 +1,131 @@
+import { Err, Ok } from '@nangohq/utils';
```
pfreixes, line 1: Should we add some tests for this one as we [have](https://github.com/NangoHQ/nango/blob/8037ded031ba2b428992d466cbfbe79c4ddbc28f/packages/billing/lib/batcher.unit.test.ts) for the current batcher?

packages/metering/lib/crons/usage.ts
```diff
@@ -81,6 +81,7 @@ const observability = {
                             }
                         }
                     ]);
+                    // TODO: ingest into clickhouse
                     metrics.gauge(metrics.Types.CONNECTIONS_COUNT, count, { accountId });
```
pfreixes, line 85: Maybe eventually we want to get rid of this and use Clickhouse as a source of truth, checking DD this metric is costing us around 300$ a month (could be peanuts, but its almost the 10% of our DD monthly costs

pfreixes (approved):
Left some comments, feel free to move on an merge but maybe would be worth it to discuss at least the way that we should store data and how much we can optimize it - if required - for retrieving it efficeintly.

packages/usage/lib/clickhouse/migrations/20260406174100_create_raw_usage_events.ts
pfreixes: Oks I see the counter part [here](https://github.com/NangoHQ/nango/pull/5804) so we already creating materialized views, so querys should the materialized views rather than the raw events, and materialized views will be updated with new events. So we are fine here.

packages/billing/lib/batcher.ts
TBonnin: Monitoring will be done as part of making CH production ready if we decide to go for it. I am gonna add a TODO in the meantime

packages/metering/lib/crons/usage.ts
TBonnin: definitely. Clickhouse can quickly become the source of truth for all the usage analytics

packages/usage/lib/clickhouse/migrations/20260406174100_create_raw_usage_events.ts
TBonnin: Exactly. the raw events table is not queried directly

packages/usage/lib/clickhouse/batcher.ts
TBonnin: good point. I forgot to add the test file. Thanks

PR #5817, opened 2026-04-08
feat(integrations): add support for bing-webmasters

naorhaziz:
## Summary
- Add Bing Webmasters as a new OAuth2 integration
- Uses Bing's own OAuth2 server (not Microsoft Entra ID)
- Includes provider config, docs, setup guide, and pre-built syncs placeholder

## Test plan
- [ ] Verify OAuth flow works with Bing Webmasters credentials
- [ ] Verify proxy requests route correctly to `https://www.bing.com/webmaster/api.svc/json`
- [ ] Verify token refresh works against the separate refresh endpoint


hassan254-prog (approved):
Thanks for the contribution, single comment otherwise looks good.

packages/providers/providers.yaml
```diff
@@ -2389,6 +2389,27 @@ bill:
...
 
+bing-webmasters:
+    display_name: Bing Webmasters
+    categories:
+        - dev-tools
+    auth_mode: OAUTH2
+    authorization_url: https://www.bing.com/webmasters/oauth/authorize
+    token_url: https://www.bing.com/webmasters/oauth/token
+    refresh_url: https://www.bing.com/webmasters/token
+    default_scopes:
```
hassan254-prog: I would omit this here, default_scopes are mainly for scopes that are needed for connections to work properly, i.e offline_access for Oauth2 providers.

Also, remember to include a template logo with a max width or height of 44px, placed inside a 64px square container.

naorhaziz: Fixed! Removed default_scopes and added the logo as well.

PR #5820, opened 2026-04-09
feat(integrations): add support for cisco duo admin

hassan254-prog:

- add support for cisco duo admin


packages/providers/providers.yaml
```diff
@@ -5425,6 +5425,43 @@ drata:
...
+cisco-duo-admin:
+    display_name: Cisco Duo Admin API
+    categories:
+        - other
+    auth_mode: BASIC
+    proxy:
+        base_url: https://${connectionConfig.hostname}
+        headers:
+            date: '${now:ddd, DD MMM YYYY HH:mm:ss ZZ}'
+            authorization: "Basic ${base64(${credentials.username}:${hmacSha1Hex(${now:ddd, DD MMM YYYY HH:mm:ss ZZ}\n${method}\n${connectionConfig.hostname}\n${path}\n${params}, ${credentials.password})})}"
```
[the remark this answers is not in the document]
hassan254-prog: The body format `form` is not needed at the provider level because the proxy passes through the request body as-is; callers are expected to set the correct Content-Type headers and serialize the body themselves.

packages/shared/lib/utils/utils.ts
```diff
@@ -206,6 +206,16 @@ function replaceSha256HexExpression(str: string, resolveInner: (inner: string) =
...
+function replaceHmacSha1HexExpression(str: string, resolveInner: (inner: string) => string): string {
+    return str.replace(/\${hmacSha1Hex\(([\s\S]*?)\)}/g, (match, inner) => {
+        const lastComma = inner.lastIndexOf(',');
+        if (lastComma === -1) return match;
+        const message = resolveInner(inner.slice(0, lastComma));
+        const key = resolveInner(inner.slice(lastComma + 1).trim());
+        return crypto.createHmac('sha1', key).update(message, 'utf8').digest('hex');
+    });
+}
+
```
TBonnin, line 218: I think it is time to move all the replace/interpolate functions to their own file (and out of utiles), with tests for all of them.

TBonnin (approved):

packages/shared/lib/utils/utils.ts
hassan254-prog: I agree, I have captured a ticket for this refractor.

PR #5822, opened 2026-04-09
feat(nan-5088): API key permissions with scoped customer keys

pfreixes:
API Key Permissions system

Services touched:
  - server — scope enforcement, auth middleware, routes, controllers
  - persist — internal secret auth path
  - jobs — renamed function calls
  - webapp — UI redesign

Data model:
- New `customer_keys` table with `key_type` column (api/webhook_signing)
- New `customer_keys_relations` table for generic entity mapping
- Cleanup trigger on environment delete
- Backfill migration copies existing api_secrets into customer_keys

Auth:
- Dual-read: public API auth checks customer_keys first, falls back to api_secrets
- Scope enforcement via withScope/withAnyScope middleware on all public API routes
- Wildcard scope matching (environment:*, environment:integrations:*, etc.)
- last_used_at tracking with 1-minute debounce

API key management:
- CRUD endpoints: list, create, patch scopes, delete (soft delete)
- RBAC: production_support can list (masked secrets) but not create/edit/delete on prod
- Secret masking for support role, full access for admin

Webhook signing key:
- Stored as key_type='webhook_signing' in customer_keys
- Displayed in Webhooks settings (read-only)
- RBAC redaction on production environments

UI:
- New "API Keys" tab in Environment Settings
- Scope selection with group-level checkboxes and presets
- Key detail/edit view with scope modification
- Delete confirmation requiring key name
- Relative time display for last_used_at
- Legacy secret key hidden from Backend settings

Documentation:
- New API keys reference page with all scopes
- Updated implementation guides: NANGO-SECRET-KEY -> NANGO-API-KEY
- Updated webhook verification docs to reference signing key

Tests:
- 17 API key CRUD + RBAC integration tests
- 36 scope enforcement tests covering all public API routes
- getEnvironment webhook_signing_key redaction test


packages/database/lib/migrations/20260420120000_create_customer_keys.cjs
```diff
@@ -0,0 +1,60 @@
...
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`
+        CREATE TABLE IF NOT EXISTS customer_keys (
+            id              SERIAL PRIMARY KEY,
+            account_id      INTEGER NOT NULL REFERENCES _nango_accounts(id),
+            key_type        VARCHAR(50) NOT NULL
+                            CONSTRAINT check_customer_keys_key_type CHECK (key_type IN ('api', 'webhook_signing')),
```
TBonnin, line 12: if I remember correctly modifying a constraint is quite cumbersome and not lock free. We could rely on typescript types and do it in code

packages/server/lib/controllers/v1/environment/deleteApiKey.ts
```diff
@@ -0,0 +1,31 @@
...
+    if (emptyQuery) {
+        res.status(400).send({ error: { code: 'invalid_query_params', errors: zodErrorToHTTP(emptyQuery.error) } });
+        return;
+    }
+
+    const keyId = Number(req.params['keyId']);
+    if (isNaN(keyId)) {
+        res.status(400).send({ error: { code: 'invalid_query_params', message: 'keyId must be a number' } });
+        return;
+    }
```
TBonnin, line 26: this should be done in zod

packages/server/lib/controllers/v1/environment/listApiKeys.ts
```diff
@@ -0,0 +1,42 @@
...
+export const listApiKeys = asyncWrapper<ListApiKeys>(async (req, res) => {
+    const emptyQuery = requireEmptyQuery(req, { withEnv: true });
+    if (emptyQuery) {
+        res.status(400).send({ error: { code: 'invalid_query_params', errors: zodErrorToHTTP(emptyQuery.error) } });
+        return;
+    }
+
+    const { environment } = res.locals;
+
+    const canReadSecret = !environment.is_production || (await resolve(res.locals, permissions.canReadProdSecretKey));
```
TBonnin: we see this pattern a couple of time, is permission allowed for given an environement and res. Maybe we could make a helper method in `permissions`

packages/server/lib/middleware/access.middleware.ts
```diff
@@ -359,12 +368,14 @@ export class AccessMiddleware {
...
-
                 res.locals['authType'] = 'secretKey';
                 res.locals['account'] = secretKeyResult.value.account;
                 res.locals['environment'] = secretKeyResult.value.environment;
                 res.locals['plan'] = secretKeyResult.value.plan;
-
+                res.locals['apiKeyScopes'] = secretKeyResult.value.auth?.scopes ?? [];
+                metrics.increment(metrics.Types.AUTH_GET_ENV_BY_SECRET_KEY_SOURCE, 1, {
+                    auth_source: secretKeyResult.value.auth?.source ?? 'env_var'
+                });
```
TBonnin, line 405: is this metric for the migration?

packages/server/lib/routes.private.ts
```diff
@@ -190,6 +194,12 @@ web.route('/environments/current').get(webAuth, can({ action: 'read', resource:
 web.route('/environments/webhook').patch(webAuth, can({ action: 'update', resource: 'webhook', scopedBy: envScope }), patchWebhook);
 web.route('/environments/variables').post(webAuth, can({ action: 'update', resource: 'environment_variable', scopedBy: envScope }), postEnvironmentVariables);
 
+// API Key management
+web.route('/environment/api-keys').get(webAuth, can({ action: 'read', resource: 'environment_key', scopedBy: envScope }), listApiKeys);
+web.route('/environment/api-keys').post(webAuth, can({ action: 'update', resource: 'environment_key', scopedBy: envScope }), createApiKey);
+web.route('/environment/api-keys/:keyId').patch(webAuth, can({ action: 'update', resource: 'environment_key', scopedBy: envScope }), patchApiKey);
+web.route('/environment/api-keys/:keyId').delete(webAuth, can({ action: 'update', resource: 'environment_key', scopedBy: envScope }), deleteApiKey);
```
TBonnin, line 210: personally I am not a super fan of `-` in url but up to you.

packages/shared/lib/services/account.service.ts
```diff
@@ -351,6 +357,141 @@ class AccountService {
...
+                    ON pending_secret.environment_id = _nango_environments.id
+                   AND pending_secret.is_default = false
+                LEFT JOIN plans ON plans.account_id = _nango_accounts.id
+                WHERE _nango_environments.deleted = false
+                LIMIT 1;
+            `,
+            [hash, hash]
+        );
+
+        const row = res.rows[0];
```
TBonnin: you can do `const [row] = await db.knex...` and save a line

packages/shared/lib/services/customerKey.service.ts
```diff
@@ -0,0 +1,315 @@
...
+    ): Promise<Result<DBCustomerKey>> {
+        try {
+            const plainText = providedSecret ?? uuid.v4();
+
+            const hashed = await this.hashSecret(plainText);
+            if (hashed.isErr()) {
+                throw hashed.error;
+            }
+
+            const created = await trx.transaction(async (innerTrx) => {
```
TBonnin, line 46: do we want an inner transaction or just run the following in a transaction if `trx` isn't already one?

packages/server/lib/controllers/v1/environment/patchApiKey.ts
```diff
@@ -0,0 +1,55 @@
...
+    if (!parsed.success) {
+        res.status(400).send({ error: { code: 'invalid_body', errors: parsed.error.issues } });
+        return;
+    }
+
+    if (parsed.data.display_name) {
+        const result = await customerKeyService.renameApiKey(db.knex, keyId, parsed.data.display_name, environment.id, account.id);
+        if (result.isErr()) {
+            const errType = (result.error as any).type ?? '';
+            const errMsg = result.error.message ?? '';
```
TBonnin: `const { type: errType = '', message: errMsg = '' } = result.error as { type?: string; message?: string };` if you want to avoid the any casting

packages/server/lib/controllers/v1/environment/patchApiKey.ts
```diff
@@ -0,0 +1,55 @@
...
+import type { Request, Response } from 'express';
+
+const bodySchema = z
+    .object({
+        scopes: z.array(z.string().min(1)).min(1).optional(),
+        display_name: z.string().min(1).max(255).optional()
+    })
+    .refine((data) => data.scopes || data.display_name, { message: 'At least one of scopes or display_name is required' });
+
+export const patchApiKey = async (req: Request, res: Response<any, RequestLocals>) => {
```
TBonnin: Missing `asyncWrapper` here I think

packages/server/lib/controllers/v1/environment/createApiKey.ts
```diff
@@ -0,0 +1,65 @@
...
+import { requireEmptyQuery, zodErrorToHTTP } from '@nangohq/utils';
+
+import { asyncWrapper } from '../../../utils/asyncWrapper.js';
+
+import type { CreateApiKey } from '@nangohq/types';
+
+const validationBody = z
+    .object({
+        display_name: z.string().min(1).max(255),
+        scopes: z.array(z.string()).optional()
```
TBonnin: should the scope values be validated to make sure they are real ones?

packages/webapp/src/pages/Environment/Settings/Backend.tsx
```diff
@@ -7,16 +7,12 @@ import { permissions } from '@nangohq/authz';
 
 import SettingsContent from './components/SettingsContent';
 import SettingsGroup from './components/SettingsGroup';
-import { useActivateKey, useEnvironment, usePatchEnvironment, useRevertKey, useRotateKey } from '../../../hooks/useEnvironment';
```
TBonnin, line 10: I imagine you are planning to remove the associated endpoints for legacy key rotation in followup PR?

packages/types/lib/api.endpoints.ts
```diff
@@ -176,6 +185,9 @@ export type PrivateApiEndpoints =
     | DeleteEnvironment
     | GetEnvironments
     | GetEnvironment
+    | ListApiKeys
+    | CreateApiKey
+    | DeleteApiKey
```
TBonnin, line 200: isn't the patch missing?

packages/database/lib/migrations/20260408120000_nan-5088_create_customer_keys.cjs
```diff
@@ -0,0 +1,60 @@
+exports.config = { transaction: false };
+
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    await knex.raw(`
+        CREATE TABLE IF NOT EXISTS customer_keys (
+            id              SERIAL PRIMARY KEY,
+            account_id      INTEGER NOT NULL REFERENCES _nango_accounts(id),
```
TBonnin: no `ON DELETE CASCADE`?

packages/webapp/src/pages/Environment/Settings/scope-logic.ts
```diff
@@ -0,0 +1,154 @@
...
+export const SCOPE_PRESETS: { label: string; description: string; scopes: string[] }[] = [
+    { label: 'Full access', description: 'All permissions', scopes: ['environment:*'] },
+    { label: 'Auth', description: 'Create connect sessions', scopes: ['environment:connect_sessions:write'] },
+    { label: 'CI/CD deploy', description: 'Deploy syncs and actions', scopes: ['environment:deploy'] },
+    {
+        label: 'Backend service',
+        description: 'Read connections, records, execute actions/syncs, proxy',
+        scopes: ['environment:connections:read', 'environment:records:read', 'environment:actions:execute', 'environment:syncs:execute', 'environment:proxy']
+    }
+];
```
TBonnin: doesn't seem to be used anywhere

packages/server/lib/routes.public.ts
```diff
@@ -152,83 +153,115 @@ publicAPI.route('/providers').get(connectSessionOrApiAuth, acceptLanguageMiddlew
...
 publicAPI.use('/connect', jsonContentTypeMiddleware);
-publicAPI.route('/connect/sessions').post(apiAuth, connectionCapping, postConnectSessions);
-publicAPI.route('/connect/sessions/reconnect').post(apiAuth, postConnectSessionsReconnect);
+publicAPI.route('/connect/sessions').post(apiAuth, withScope('environment:connect_sessions:write'), connectionCapping, postConnectSessions);
+publicAPI.route('/connect/sessions/reconnect').post(apiAuth, withScope('environment:connect_sessions:write'), postConnectSessionsReconnect);
 publicAPI.route('/connect/session').get(connectSessionAuth, getConnectSession);
 publicAPI.route('/connect/session').delete(connectSessionAuth, deleteConnectSession);
 publicAPI.route('/connect/telemetry').post(connectSessionAuthBody, postConnectTelemetry);
 
+// V1 passthrough (deprecated) — inherits action/records scopes
```
TBonnin: can you explain the comment. I am not sure I understand

packages/server/lib/controllers/v1/environment/createApiKey.ts
TBonnin: also no sure why we allow to pass no scopes, especially if it implicitly mean full access below `scopes ?? ['environment:*']`

packages/database/lib/migrations/20260420120000_create_customer_keys.cjs
pfreixes: Good point about the lock concern. However, modifying a CHECK constraint is a metadata-only change — it takes an exclusive lock but releases it immediately (no table rewrite). Since queries on this table are fast, the lock window is - or should be - negligible. We also took the same approach for user roles. The benefit is a storage-level safety net that prevents invalid values regardless of how data gets inserted (bugs, manual queries, etc.).

So here we need to make a call about what tradeoff we want to make, also would be interesting to follow some kind of pattern here. I'm happy to make the call to move all of these declarations in our code base for the sake of reducing to zero the disturbance of having to make changes in our schemas over having the storage-level safety.

Thoughts?

packages/server/lib/middleware/access.middleware.ts
pfreixes: Yes Im planning to remove this metric once we have finished the migration and there is no ore dual lookups, Ill use it for making sure that all keys used are coming from the new table before doing the cut off

packages/server/lib/routes.private.ts
pfreixes: Just following current convention, most of our URLs which require two or more words are using the `-`, Ill stick to it for consistency.

packages/shared/lib/services/customerKey.service.ts
pfreixes: From what I understood the knex type here can be either a plain connection or a connection which already uses a transaction. We have other places of the code where we are doing this, like [here](https://github.com/NangoHQ/nango/blob/master/packages/shared/lib/services/environment.service.ts#L112).

Adding an inner transaction guarantees the semantics, regardless if what is provided is a connection or a real transaction

packages/server/lib/controllers/v1/environment/patchApiKey.ts
pfreixes: good catch, fixed

packages/server/lib/routes.public.ts
```diff
@@ -152,83 +153,115 @@ publicAPI.route('/providers').get(connectSessionOrApiAuth, acceptLanguageMiddlew
...
 publicAPI.use('/v1', jsonContentTypeMiddleware);
 publicAPI.route('/v1/*splat').all(apiAuth, allPublicV1);
```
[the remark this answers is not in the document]
pfreixes: fixed!

packages/shared/lib/services/environment.service.ts
```diff
@@ -103,6 +104,23 @@ class EnvironmentService {
...
+            });
+            if (apiKey.isErr()) {
+                throw apiKey.error;
+            }
+
+            await customerKeyService.createWebhookSigningKey(trx, {
+                accountId: accountId,
+                environmentId: environment.id,
+                secret: secret.secret
+            });
```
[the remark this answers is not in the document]
pfreixes: fixed!

packages/webapp/src/pages/Environment/Settings/ApiKeys.tsx
```diff
@@ -0,0 +1,559 @@
...
+    const handleCreate = async () => {
+        if (!displayName.trim()) {
+            toast({ title: 'Display name is required', variant: 'error' });
+            return;
+        }
+        try {
+            await createApiKey({
+                display_name: displayName.trim(),
+                scopes: selectedScopes.length > 0 ? selectedScopes : undefined
+            });
```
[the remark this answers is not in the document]
pfreixes: fixed

packages/server/lib/controllers/v1/environment/deleteApiKey.ts
pfreixes: fixed!

packages/server/lib/controllers/v1/environment/listApiKeys.ts
pfreixes: fixed!

packages/shared/lib/services/account.service.ts
pfreixes: done!

packages/server/lib/controllers/v1/environment/patchApiKey.ts
pfreixes: done!

packages/server/lib/controllers/v1/environment/createApiKey.ts
pfreixes: good catch, I need to somehow update the why that I do prompts to Claude for making sure that it follows existing patterns, and I need to be more meticulous with reviewing the code produced by Claude

Added scope validation and not allowing empty scopes!

packages/webapp/src/pages/Environment/Settings/Backend.tsx
pfreixes: Yes, I need to come up with a final plan of everything ... stay tunned

packages/types/lib/api.endpoints.ts
pfreixes: yes good catch!

packages/database/lib/migrations/20260408120000_nan-5088_create_customer_keys.cjs
pfreixes: another good catch!

packages/webapp/src/pages/Environment/Settings/scope-logic.ts
pfreixes: removed!

packages/server/lib/routes.public.ts
pfreixes: Comments was fully wrong, Ive fixed this by also checking the scopes inline

packages/server/lib/controllers/v1/environment/createApiKey.ts
```diff
@@ -0,0 +1,63 @@
...
+import { requireEmptyQuery, zodErrorToHTTP } from '@nangohq/utils';
+
+import { asyncWrapper } from '../../../utils/asyncWrapper.js';
+
+import type { CreateApiKey } from '@nangohq/types';
+
+const validationBody = z
+    .object({
+        display_name: z.string().min(1).max(255),
+        scopes: z.array(z.string()).nonempty('At least one scope is required when scopes are provided').optional()
```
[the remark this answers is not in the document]
pfreixes: fixed

packages/shared/lib/services/environment.service.ts
```diff
@@ -103,6 +104,26 @@ class EnvironmentService {
             const secret = created.value;
             environment.secret_key = secret.secret;
             environment.pending_secret_key = null;
+
+            const apiKey = await customerKeyService.createApiKey(trx, {
+                accountId: accountId,
+                environmentId: environment.id,
+                displayName: 'Default - Full access',
+                secret: secret.secret
```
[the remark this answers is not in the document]
pfreixes: We are addressing the broken rotation key system, the user will be able to rotate the key by creating a new one under `customer_keys`

packages/database/lib/migrations/20260420120000_create_customer_keys.cjs
TBonnin: as you said it is a tradeoff. I have been bitten by metadata exclusive lock before. The recent records 2mins lockup was exactly that because even if there is no table rewrite the exclusive lock wait for all queries to complete and all coming queries piles up behind it. The assumption that all queries on this table are fast might be true at first but it might be different in 6 months. 

Up to you though.

packages/webapp/src/pages/Environment/Settings/Backend.tsx
pfreixes: Im gonna make finally the proposal to disable them in this PR (see new updates), looking and their [usage](https://us3.datadoghq.com/s/e94cfb60-2611-11ee-93ab-da7ad0900003/w65-kjy-xai) the chances of having someone doing a rotation during the deploy is barely impossible

packages/shared/lib/services/customerKey.service.ts
```diff
@@ -0,0 +1,315 @@
...
+                .whereExists(function () {
+                    void this.select(trx.raw('1'))
+                        .from(CUSTOMER_KEYS_RELATIONS_TABLE)
+                        .whereRaw(`${CUSTOMER_KEYS_RELATIONS_TABLE}.customer_key_id = ${CUSTOMER_KEYS_TABLE}.id`)
+                        .whereRaw(`${CUSTOMER_KEYS_RELATIONS_TABLE}.entity_type = ?`, ['environment'])
+                        .whereRaw(`${CUSTOMER_KEYS_RELATIONS_TABLE}.entity_id = ?`, [envId]);
+                })
+                .update({
+                    deleted_at: trx.fn.now() as unknown as Date
+                });
```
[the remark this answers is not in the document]
pfreixes: We will remove dual lookup, any ccompromise of the new default keys will not impact once removed the dual loockup

packages/shared/lib/services/environment.service.ts
```diff
@@ -103,6 +104,26 @@ class EnvironmentService {
...
+            });
+            if (apiKey.isErr()) {
+                throw apiKey.error;
+            }
+
+            const webhookKey = await customerKeyService.createWebhookSigningKey(trx, {
+                accountId: accountId,
+                environmentId: environment.id,
+                secret: secret.secret
```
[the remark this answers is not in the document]
pfreixes: This is what we had till Today as a follow up work once removed the dual lookup we will decouple the value and we will generate its own webhook value

packages/server/lib/routes.public.ts
```diff
@@ -152,83 +153,115 @@ publicAPI.route('/providers').get(connectSessionOrApiAuth, acceptLanguageMiddlew
 publicAPI.route('/providers/:provider').get(connectSessionOrApiAuth, acceptLanguageMiddleware, getPublicProvider);
 
 // @deprecated rollbacked for one customer, to delete asap
-publicAPI.route('/config/:providerConfigKey').get(apiAuth, configController.getProviderConfig.bind(configController));
+publicAPI
+    .route('/config/:providerConfigKey')
+    .get(
+        apiAuth,
+        withAnyScope('environment:integrations:read', 'environment:integrations:read_credentials'),
```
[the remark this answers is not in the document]
pfreixes: Good catch! fixed.

packages/shared/lib/services/account.service.ts
```diff
@@ -254,7 +274,11 @@ class AccountService {
             | { environmentId: number }
             | { environmentUuid: string }
             | { accountUuid: string; envName: string }
-    ): Promise<{ account: DBTeam; environment: DBEnvironment; secret: DBAPISecret; plan: DBPlan | null } | null> {
+    ): Promise<AccountContext | null> {
+        if ('secretKey' in opts) {
+            return this.getAccountContextByAnySecret(opts.secretKey);
```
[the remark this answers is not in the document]
pfreixes: Tehcnically persist is not reachable from outside, but its still worth it to fix this for not polluting the dual lookup, fixing it!

packages/server/lib/controllers/v1/environment/getEnvironment.ts
```diff
@@ -72,6 +73,14 @@ export const getEnvironment = asyncWrapper<GetEnvironment>(async (req, res) => {
 
     const webhookSettings = await externalWebhookService.get(environment.id);
 
+    let webhookSigningKey: string | null = null;
+    if (await canReadProdSecret(res.locals, environment)) {
+        const signingKeyResult = await customerKeyService.getWebhookSigningKeyForEnv(db.knex, environment.id);
+        if (signingKeyResult.isOk()) {
+            webhookSigningKey = signingKeyResult.value.secret;
```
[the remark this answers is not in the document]
pfreixes: We will make this change in a follow up PR, we do not require this right now since same secret is used, but we will require to do it before we allow to rotate webhooks or the internal keys (which is not possible and the legacy endpoints will be disabled)

packages/shared/lib/utils/encryption.manager.ts
```diff
@@ -281,6 +281,21 @@ export class EncryptionManager extends Encryption {
             await db.knex.from<DBEnvironmentVariable>(`_nango_environment_variables`).where({ id: environmentVariable.id }).update(environmentVariable);
         }
 
+        const customerKeys = await db.knex.select('*').from<DBCustomerKey>('customer_keys').whereNull('deleted_at');
```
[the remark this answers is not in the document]
pfreixes: fixed

packages/server/lib/routes.public.ts
```diff
@@ -152,83 +153,115 @@ publicAPI.route('/providers').get(connectSessionOrApiAuth, acceptLanguageMiddlew
...
 publicAPI.use('/connections', jsonContentTypeMiddleware);
-publicAPI.route('/connections').post(apiAuth, postPublicConnection);
-publicAPI.route('/connections').get(apiAuth, getPublicConnections);
-publicAPI.route('/connections/metadata').post(apiAuth, postPublicMetadata);
-publicAPI.route('/connections/metadata').patch(apiAuth, patchPublicMetadata);
-publicAPI.route('/connections/:connectionId').get(apiAuth, getPublicConnection);
-publicAPI.route('/connections/:connectionId').patch(apiAuth, patchPublicConnection);
-publicAPI.route('/connections/:connectionId').delete(apiAuth, deletePublicConnection);
+publicAPI.route('/connections').post(apiAuth, withScope('environment:connections:write'), postPublicConnection);
+publicAPI.route('/connections').get(apiAuth, withAnyScope('environment:connections:list', 'environment:connections:list_credentials'), getPublicConnections);
```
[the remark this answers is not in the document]
pfreixes: We need to see if in the future we want do add the credentials there, we will have the scope, right now list == list_credentials

packages/server/lib/routes.public.ts
```diff
@@ -152,83 +153,115 @@ publicAPI.route('/providers').get(connectSessionOrApiAuth, acceptLanguageMiddlew
...
+// Integrations
 publicAPI.use('/integrations', jsonContentTypeMiddleware);
-publicAPI.route('/integrations').get(connectSessionOrApiAuth, getPublicListIntegrations);
-publicAPI.route('/integrations').post(apiAuth, postPublicIntegration);
-publicAPI.route('/integrations/:uniqueKey').patch(apiAuth, patchPublicIntegration);
-publicAPI.route('/integrations/:uniqueKey').get(apiAuth, getPublicIntegration);
-publicAPI.route('/integrations/:uniqueKey').delete(apiAuth, deletePublicIntegration);
+publicAPI
+    .route('/integrations')
+    .get(connectSessionOrApiAuth, withAnyScope('environment:integrations:list', 'environment:integrations:list_credentials'), getPublicListIntegrations);
```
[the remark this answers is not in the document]
pfreixes: We need to see if in the future we want do add the credentials there, we will have the scope, right now list == list_credentials

packages/server/lib/routes.private.ts
```diff
@@ -194,22 +198,17 @@ web.route('/environments/current').get(webAuth, can({ action: 'read', resource:
...
-);
-web.route('/environment/activate-key').post(
-    webAuth,
-    can({ action: 'update', resource: 'environment_key', scopedBy: envScope }),
-    environmentController.activateKey.bind(environmentController)
-);
+// Legacy key rotation endpoints — disabled, replaced by customer_keys API (NAN-5088)
+// web.route('/environment/rotate-key').post(webAuth, can(...), environmentController.rotateKey)
+// web.route('/environment/revert-key').post(webAuth, can(...), environmentController.revertKey)
+// web.route('/environment/activate-key').post(webAuth, can(...), environmentController.activateKey)
```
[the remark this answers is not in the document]
pfreixes: we will not longer supporting this routes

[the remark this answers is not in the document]
pfreixes: we will not longer supporting this routes

packages/database/lib/migrations/20260408120000_nan-5088_create_customer_keys.cjs
```diff
@@ -0,0 +1,67 @@
...
+
+        -- Supports reverse lookups (entity → keys) used by the cleanup trigger on environment delete
+        CREATE INDEX IF NOT EXISTS idx_customer_keys_relations_entity
+            ON customer_keys_relations (entity_type, entity_id);
+
+        CREATE OR REPLACE FUNCTION cleanup_customer_key_relations()
+        RETURNS TRIGGER AS $$
+        BEGIN
+            DELETE FROM customer_keys_relations
+            WHERE entity_type = TG_ARGV[0] AND entity_id = OLD.id;
```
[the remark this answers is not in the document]
pfreixes: good catch!

docs/reference/api-keys.mdx
```diff
@@ -0,0 +1,192 @@
...
+
+### CI/CD Deploy
+
+For CI/CD pipelines deploying syncs and actions to production:
+
+| Scope |
+|-------|
+| `environment:deploy` |
+
+### Backend Service
```
bastienbeurier, line 92: This is sure if this is the right place, but it might be worth it to say that it's more secure if they never allow listing connections, so that they can treat connection IDs as a connection-specific secret key, meaning that a leaked API key will make it harder for an attacker to access customer-specific data.

docs/docs.json
```diff
@@ -994,6 +994,7 @@
           {
             "group": "Backend",
             "pages": [
+              "reference/api-keys",
```
bastienbeurier: @pfreixes can we put this one level down in the nav, under "HTTP API" please? 

Want to make sure that the top level navigation only contains the key APIs/SDKs.

docs/getting-started/quickstart.mdx
```diff
@@ -36,7 +36,7 @@ description: 'Authorize any API in minutes'
         <Tab title="cURL">
 
         ```bash
-        # <NANGO-SECRET-KEY>: Find in your Environment Settings
+        # <NANGO-SECRET-KEY>: Find in Environment Settings > API Keys
```
marcindobry: Should we rename this to NANGO-API-KEY?

packages/database/lib/migrations/20260420120000_create_customer_keys.cjs
```diff
@@ -0,0 +1,70 @@
...
+            tag             VARCHAR(255) NOT NULL DEFAULT '',
+            hashed          VARCHAR(255) NOT NULL,
+            last_used_at    TIMESTAMPTZ,
+            created_at      TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+            updated_at      TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+            deleted_at      TIMESTAMPTZ
+        );
+
+        -- Auth lookup by hash — unique per key_type to prevent ambiguous key resolution
+        CREATE UNIQUE INDEX IF NOT EXISTS idx_customer_keys_hashed_key_type
```
marcindobry, line 26: What do you think about adding account_id as the first part of the index? Won't we always want to add account_id anyways to prevent accidental cross-customer lookups?

docs/getting-started/quickstart.mdx
pfreixes: Yea I can do that, but Ill keep the `NANGO-SECRET-KEY` working just for the sake of keeping compatibility, but this touches a lot more files .. how do do yo feel doing this change in a following PR?

packages/database/lib/migrations/20260420120000_create_customer_keys.cjs
pfreixes: I would say that there is no risk Today considering that hashes can be considered unique, also we do not know the `account_id` at query time, we figure out the account later when once checked that the hash of the key has access to the environment

The legacy `api_secrets` does the same.

We could add also the `"idx_customer_keys_hashed_key_type" UNIQUE, btree (hashed, key_type)` as the third element, but not sure if this will really provide value unless Im missing something!

packages/server/lib/formatters/connection.ts
```diff
@@ -87,12 +87,14 @@ export function connectionFullToPublicApi({
     data,
     provider,
     activeLog,
-    endUser
+    endUser,
+    includeCredentials = true
```
rossmcewan: Why is `includeCredentials` always set to `true` by default? Why wouldn't callers be explicit?

packages/server/lib/middleware/access.middleware.ts
```diff
@@ -86,6 +91,29 @@ export class AccessMiddleware {
                 return;
             }
 
+            const isScript = req.get('Nango-Is-Script') === 'true';
```
rossmcewan, line 94: I am going to enter a ticket to look at how we can verify the caller more appropriately. Maybe a signed token that the function execution runtime can sign and can be verified by the middleware.

packages/server/lib/routes.private.ts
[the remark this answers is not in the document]
rossmcewan: Do we need to publish this as a breaking change in-case it is being used systematically by any customers? 
I wonder if returning a well-defined error from the endpoint will be better than deleting the route.

packages/shared/lib/services/customerKey.service.ts
```diff
@@ -0,0 +1,315 @@
...
+                    .join(CUSTOMER_KEYS_RELATIONS_TABLE, `${CUSTOMER_KEYS_RELATIONS_TABLE}.customer_key_id`, `${CUSTOMER_KEYS_TABLE}.id`)
+                    .where(`${CUSTOMER_KEYS_TABLE}.key_type`, 'api')
+                    .where(`${CUSTOMER_KEYS_TABLE}.display_name`, displayName)
+                    .where(`${CUSTOMER_KEYS_RELATIONS_TABLE}.entity_type`, 'environment')
+                    .where(`${CUSTOMER_KEYS_RELATIONS_TABLE}.entity_id`, environmentId)
+                    .whereNull(`${CUSTOMER_KEYS_TABLE}.deleted_at`)
+                    .first();
+
+                if (existing) {
+                    throw new NangoError('duplicate_api_secret', { display_name: displayName });
```
rossmcewan: Should this not be `duplicate_api_key`

packages/shared/lib/services/customerKey.service.ts
```diff
@@ -0,0 +1,315 @@
...
+                    .where(`${CUSTOMER_KEYS_TABLE}.key_type`, 'api')
+                    .where(`${CUSTOMER_KEYS_TABLE}.display_name`, displayName)
+                    .where(`${CUSTOMER_KEYS_RELATIONS_TABLE}.entity_type`, 'environment')
+                    .where(`${CUSTOMER_KEYS_RELATIONS_TABLE}.entity_id`, envId)
+                    .whereNull(`${CUSTOMER_KEYS_TABLE}.deleted_at`)
+                    .whereNot(`${CUSTOMER_KEYS_TABLE}.id`, keyId)
+                    .first();
+
+                if (existing) {
+                    throw new NangoError('duplicate_api_secret', { display_name: displayName });
```
rossmcewan: And here

packages/server/lib/formatters/connection.ts
pfreixes: no reason at all, done

packages/server/lib/middleware/access.middleware.ts
pfreixes: Im planning to address this issue as a follow up item (unless we have other urgent things to do)

packages/shared/lib/services/customerKey.service.ts
pfreixes: yeps we can, done!

packages/shared/lib/services/customerKey.service.ts
pfreixes: yeps we can, done!

packages/server/lib/formatters/connection.ts
pfreixes: changed, now the caller makes this explicit no default value

marcindobry (approved):

docs/getting-started/quickstart.mdx
pfreixes: Changed! my bad 🤦 this was only about changing the snippet which we can do it without any problem

PR #5825, opened 2026-04-09
feat(integrations): add support for stay-ai

hassan254-prog:

-  add support for stay-ai, untested


packages/providers/providers.yaml
```diff
@@ -1971,7 +1971,7 @@ aws:
             title: API Subdomain
             description: The API subdomain to the API you want to connect to
             example: idp.us-east-2
-            pattern: '^[a-z.-]+$'
```
hassan254-prog, line 1974: Not related to this pr, but this fixes a small regex pattern issue with aws.

marcindobry (approved):

PR #5833, opened 2026-04-10
fix: gate auto-idling trial behavior on auto_idle

kaposke:
There were cases where the `AutoIdlingBanner` would show even when the `auto_idle` was off (happens in our nango admin account, for example). I noticed that most of our checks relied on `trial_end_at` without really checking `auto_idle`, and upgrading plans didn't clear out trial dates. This PR makes it so the trial data is cleared when upgrading to paid plans, but also adds `auto_idle` checks in places where it should have impact in to support accounts that have it off and still have `auto_idle` off. This way we can still just toggle the flag from retool and have it all working.

### DIFF
--- a/packages/shared/lib/services/plans/plans.ts
+++ b/packages/shared/lib/services/plans/plans.ts
@@ -128,19 +128,6 @@ export async function getExpiredTrials(db: Knex): Promise<DBPlan[]> {
         .where('plans.auto_idle', true);
 }
 
-export function resetTrialStateOnPlanChange(): Pick<
-    DBPlan,
-    'trial_start_at' | 'trial_end_at' | 'trial_end_notified_at' | 'trial_extension_count' | 'trial_expired'
-> {
-    return {
-        trial_start_at: null,
-        trial_end_at: null,
-        trial_end_notified_at: null,
-        trial_extension_count: 0,
-        trial_expired: null
-    };
-}
-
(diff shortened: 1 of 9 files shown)


packages/shared/lib/services/plans/plans.ts
```diff
@@ -163,6 +165,15 @@ export async function handlePlanChanged(
...
+        ...(currentPlan.value.auto_idle && mergedFlags.auto_idle === false
+            ? {
+                  trial_start_at: null,
+                  trial_end_at: null,
+                  trial_end_notified_at: null,
+                  trial_extension_count: 0,
+                  trial_expired: null
+              }
+            : {}),
         ...mergedFlags
```
TBonnin, line 177: does the order matter here? ie: could mergeFlags override what's set above?

TBonnin (approved):

packages/shared/lib/services/plans/plans.ts
marcindobry: I think that's the point, the stuff above looks like defaults.

marcindobry (approved):

packages/shared/lib/services/plans/plans.ts
kaposke: Yeah these are not flags, just values we keep track of, so they won't me overriden

PR #5835, opened 2026-04-10
feat: allow customers to update invoicing details (frontend)

ErickRDev:
Customers currently have to reach out to support in order to have their billing invoicing details updated.

This PR gives customers the ability to self-serve on the invoicing details update process. It exposes UI components that render the user's current invoicing details and allows them to update them.

Go to the "Payment & Invoices" tab in the "Billing & usage" menu and edit away your invoicing information! :)


packages/webapp/src/pages/Team/Billing/components/InvoicingDetailsForm.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+import type { BillingCustomer } from '@nangohq/types';
+
+const addressSchema = z.object({
+    line1: z.string().nullable(),
+    line2: z.string().nullable(),
+    city: z.string().nullable(),
+    state: z.string().nullable(),
+    postalCode: z.string().nullable(),
+    country: z.string()
```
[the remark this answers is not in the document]
ErickRDev: Even though Orb's schema models `country` as an unconstrained string, they indeed respond with errors upon sending `country` as an empty string; in light of that, I've updated the code so that both forms (tax-id and billing-address) use the same country code structure to both validate and feed a select component.

packages/webapp/src/pages/Team/Billing/components/InvoicingDetailsForm.tsx
```diff
@@ -0,0 +1,151 @@
...
+const countryValues = countryCodes.map((c) => c.value) as [string, ...string[]];
+const taxIdTypeValues = taxIdTypes.map((t) => t.value) as [string, ...string[]];
+
+const addressSchema = z.object({
+    line1: z.string().nullable(),
+    line2: z.string().nullable(),
+    city: z.string().nullable(),
+    state: z.string().nullable(),
+    postalCode: z.string().nullable(),
+    country: z.enum(countryValues, { message: 'Valid country required' })
```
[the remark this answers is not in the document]
ErickRDev: IMO, it seems reasonable to use the tax-country list to feed the `Country` select on billing address. If we receive customer complaints or requests to introduce other countries, we can quickly iterate on this.

For context, the list is taken from Orb's documentation of "supported tax-id countries".

packages/webapp/src/pages/Team/Billing/components/InvoicingAddressFields.tsx
```diff
@@ -0,0 +1,140 @@
...
+    return (
+        <Card className="bg-bg-surface rounded border-2 border-border-disabled py-0 gap-0">
+            <CardHeader className="bg-bg-elevated h-10 flex flex-row items-center justify-between px-6">
+                <CardTitle className="text-text-primary !text-heading-sm flex items-center gap-2">
+                    Billing address
+                    <OptionalTag />
+                </CardTitle>
+                {address ? (
+                    <Button type="button" variant="ghost" size="icon" onClick={handleRemove}>
+                        <Trash2 className="size-4 text-text-tertiary" />
```
kaposke: These are the button defaults for `size="icon" variant="ghost"`
```suggestion
                        <Trash2 />
```

packages/webapp/src/pages/Team/Billing/components/InvoicingAddressFields.tsx
```diff
@@ -0,0 +1,140 @@
...
+                    </Button>
+                ) : (
+                    <Button type="button" variant="ghost" size="icon" onClick={handleAdd}>
+                        <ChevronDown className="size-4 text-text-tertiary" />
```
kaposke:
```suggestion
                        <ChevronDown />
```

kaposke: I've commented on figma to see what Gabi thinks, but I really think these should be a `+` icon instead. This chevron makes you think you're expading something that exists. It contrast with the "Trash" icon for the reverse action.

packages/webapp/src/pages/Team/Billing/components/InvoicingAddressFields.tsx
```diff
@@ -0,0 +1,140 @@
...
+                    </Button>
+                )}
+            </CardHeader>
+            {address && (
+                <CardContent className="grid grid-cols-2 gap-3 items-start px-6 py-9">
```
kaposke, line 45: Grid frontender detected. Imma fight ya with mi flexboxes!

packages/webapp/src/pages/Team/Billing/components/InvoicingDetailsForm.tsx
```diff
@@ -0,0 +1,151 @@
...
+import { Input } from '@/components-v2/ui/input';
+import { Skeleton } from '@/components-v2/ui/skeleton';
+import { usePutBillingInvoicingDetails } from '@/hooks/usePlan';
+import { useToast } from '@/hooks/useToast';
+import { useStore } from '@/store';
+
+import type { BillingCustomer } from '@nangohq/types';
+
+export const InvoicingInput = (props: React.ComponentProps<typeof Input>) => (
+    <Input className="data-[filled=true]:not-aria-invalid:border-border-muted" {...props} />
```
kaposke: Why? 

Oh I see the design examples with values don't have the usual border. Might be fine to keep the border though, figma has some inconsistencies every now and then but we can generally just use the same components.

Also valid to ask Gabi if she meant it (maybe she wants them all removed app-wide?)

packages/webapp/src/pages/Team/Billing/components/InvoicingTaxIdFields.tsx
```diff
@@ -0,0 +1,145 @@
...
+    return (
+        <Card className="bg-bg-surface rounded border-2 border-border-disabled py-0 gap-0">
+            <CardHeader className="bg-bg-elevated h-10 flex flex-row items-center justify-between px-6">
+                <CardTitle className="text-text-primary !text-heading-sm flex items-center gap-2">
+                    Tax ID
+                    <OptionalTag />
+                </CardTitle>
+                {taxId ? (
+                    <Button type="button" variant="ghost" size="icon" onClick={handleRemove}>
+                        <Trash2 className="size-4 text-text-tertiary" />
```
kaposke:
```suggestion
                        <Trash2 />
```

packages/webapp/src/pages/Team/Billing/components/InvoicingTaxIdFields.tsx
```diff
@@ -0,0 +1,145 @@
...
+                    </Button>
+                ) : (
+                    <Button type="button" variant="ghost" size="icon" onClick={handleAdd}>
+                        <ChevronDown className="size-4 text-text-tertiary" />
```
kaposke: (and same about using +)
```suggestion
                        <ChevronDown />
```

packages/webapp/src/pages/Team/Billing/invoicingConstants.ts
```diff
@@ -0,0 +1,228 @@
...
+    { value: 'UY', label: 'Uruguay' },
+    { value: 'UZ', label: 'Uzbekistan' },
+    { value: 'VE', label: 'Venezuela' },
+    { value: 'VN', label: 'Vietnam' },
+    { value: 'ZA', label: 'South Africa' },
+    { value: 'ZM', label: 'Zambia' },
+    { value: 'ZW', label: 'Zimbabwe' }
+];
+
+export const taxIdTypes: { value: string; label: string; placeholder: string }[] = [
```
kaposke: By chance I tried italy and it had no matching type. Is this list from orb/stripe somewhere?

kaposke (commented):
The select styles (when expanded) look different from the designs. Non-blocking but maybe a low-hanging fruit?

packages/webapp/src/pages/Team/Billing/components/InvoicingAddressFields.tsx
ErickRDev: Ops, didn't notice that! Awesome, thanks for pointing that out! 🙇🏼

packages/webapp/src/pages/Team/Billing/components/InvoicingAddressFields.tsx
ErickRDev: I agree with that! My first stab was to have a button (as a CTA) below the header, but that was way uglier 👹

packages/webapp/src/pages/Team/Billing/components/InvoicingAddressFields.tsx
ErickRDev: Not much of a frontender at all, to be honest with you! Show me the way! 🙏🏼

packages/webapp/src/pages/Team/Billing/components/InvoicingDetailsForm.tsx
ErickRDev: Yeah, I started with the non-overridden, regular borders, but Gabi's design used the `muted` variant, so I thought that was on purpose and went ahead with this approach.

packages/webapp/src/pages/Team/Billing/invoicingConstants.ts
ErickRDev: There's a 🐛 here; nice catch!

EU countries are not handled properly with the current logic; will commit a fix soon!

ErickRDev:
> The select styles (when expanded) look different from the designs. Non-blocking but maybe a low-hanging fruit?

Just checked this and I'm left wondering whether it might not be another case of Figma having inconsistencies? I'm using the <Select> component from components-v2, which is used elsewhere in the codebase (e.g. in the environment settings), so if the change is indeed intentional, maybe it's worth checking whether we shouldn't change the base component styling instead?

packages/webapp/src/pages/Team/Billing/components/InvoicingDetailsForm.tsx
ErickRDev: I'm removing this override and will check with Gabi tomorrow.

packages/webapp/src/pages/Team/Billing/components/InvoicingAddressFields.tsx
kaposke: Not me either. Grids are good as well :) I just flex box everything out of naiveness

kaposke (approved):

PR #5836, opened 2026-04-11
feat(server): add remote function public API

marcindobry:
Adds public /remote-function/compile, /remote-function/dryrun, and /remote-function/deploy endpoints with request validation, endpoint types, and E2B/local runtime clients for running Nango function code.
This intentionally keeps the broader agent-session/internal sandbox prompting APIs and UI work out of this PR.

N/A

- npm run ts-build
- npm run test:unit -- packages/server/lib/services/remote-function/compiler-client.unit.test.ts packages/server/lib/services/remote-function/shell.unit.test.ts
- npm run test:integration -- packages/server/lib/controllers/functions/remote-function.integration.test.ts

### DIFF
--- a/packages/server/lib/controllers/functions/compile/postCompile.ts
+++ b/packages/server/lib/controllers/functions/compile/postCompile.ts
@@ -49,7 +49,6 @@ export const postRemoteFunctionCompile = asyncWrapper<PostRemoteFunctionCompile>
     } catch (err) {
         sendStepError({
             res,
-            step: 'compilation',
             error: err,
             ...(err instanceof RemoteFunctionError ? {} : { status: 500 })
         });
--- a/packages/server/lib/controllers/functions/compile/postCompile.ts
+++ b/packages/server/lib/controllers/functions/compile/postCompile.ts
@@ -1,8 +1,8 @@
 import { configService } from '@nangohq/shared';
 import { requireEmptyQuery, zodErrorToHTTP } from '@nangohq/utils';
 
-import { CompilerError, invokeCompiler } from '../../../services/remote-function/compiler-client.js';
-import { sendStepError } from '../../../services/remote-function/helpers.js';
+import { invokeCompiler } from '../../../services/remote-function/compiler-client.js';
+import { RemoteFunctionError, sendStepError } from '../../../services/remote-function/helpers.js';
 import { asyncWrapper } from '../../../utils/asyncWrapper.js';
 import { remoteFunctionCompileBodySchema } from '../validation.js';
 
@@ -51,7 +51,7 @@ export const postRemoteFunctionCompile = asyncWrapper<PostRemoteFunctionCompile>
             res,
             step: 'compilation',
             error: err,
-            status: err instanceof CompilerError ? 400 : 500
+            ...(err instanceof RemoteFunctionError ? {} : { status: 500 })
         });
     }
 });
--- a/packages/server/lib/controllers/functions/compile/postCompile.ts
+++ b/packages/server/lib/controllers/functions/compile/postCompile.ts
@@ -1,0 +1,57 @@
+import { configService } from '@nangohq/shared';
+import { requireEmptyQuery, zodErrorToHTTP } from '@nangohq/utils';
+
+import { CompilerError, invokeCompiler } from '../../../services/remote-function/compiler-client.js';
+import { sendStepError } from '../../../services/remote-function/helpers.js';
+import { asyncWrapper } from '../../../utils/asyncWrapper.js';
+import { remoteFunctionCompileBodySchema } from '../validation.js';
+
+import type { PostRemoteFunctionCompile } from '@nangohq/types';
+
+export const postRemoteFunctionCompile = asyncWrapper<PostRemoteFunctionCompile>(async (req, res) => {
+    const emptyQuery = requireEmptyQuery(req);
+    if (emptyQuery) {
+        res.status(400).send({ error: { code: 'invalid_query_params', errors: zodErrorToHTTP(emptyQuery.error) } });
+        return;
+    }
+
+    const valBody = remoteFunctionCompileBodySchema.safeParse(req.body);
+    if (!valBody.success) {
+        res.status(400).send({ error: { code: 'invalid_body', errors: zodErrorToHTTP(valBody.error) } });
+        return;
+    }
+
+    const body = valBody.data;
+    const { environment } = res.locals;
+
+    const providerConfig = await configService.getProviderConfig(body.integration_id, environment.id);
+    if (!providerConfig) {
+        res.status(404).send({ error: { code: 'integration_not_found', message: `Integration '${body.integration_id}' was not found` } });
+        return;
+    }
+
+    try {
+        const result = await invokeCompiler({
+            integration_id: body.integration_id,
+            function_name: body.function_name,
+            function_type: body.function_type,
+            code: body.code
+        });
+
+        res.status(200).send({
+            integration_id: body.integration_id,
+            function_name: body.function_name,
+            function_type: body.function_type,
+            bundle_size_bytes: result.bundleSizeBytes,
+            bundled_js: result.bundledJs,
+            compiled_at: new Date().toISOString()
+        });
+    } catch (err) {
+        sendStepError({
+            res,
+            step: 'compilation',
+            error: err,
+            status: err instanceof CompilerError ? 400 : 500
+        });
+    }
+});
--- a/packages/server/lib/controllers/functions/deploy/postDeploy.ts
+++ b/packages/server/lib/controllers/functions/deploy/postDeploy.ts
@@ -1,0 +1,76 @@
+import db from '@nangohq/database';
+import { configService, getApiUrl, getSyncConfigRaw, secretService } from '@nangohq/shared';
+import { requireEmptyQuery, zodErrorToHTTP } from '@nangohq/utils';
+
+import { invokeDeploy } from '../../../services/remote-function/deploy-client.js';
+import { sendStepError } from '../../../services/remote-function/helpers.js';
+import { asyncWrapper } from '../../../utils/asyncWrapper.js';
+import { remoteFunctionDeployBodySchema } from '../validation.js';
+
+import type { PostRemoteFunctionDeploy } from '@nangohq/types';
+
+export const postRemoteFunctionDeploy = asyncWrapper<PostRemoteFunctionDeploy>(async (req, res) => {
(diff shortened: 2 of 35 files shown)


packages/server/lib/controllers/functions/deploy/postDeploy.ts
```diff
@@ -0,0 +1,76 @@
...
+    });
+    if (existingSyncConfig && (existingSyncConfig.is_public || existingSyncConfig.pre_built)) {
+        res.status(400).send({
+            error: {
+                code: 'invalid_request',
+                message: `Cannot overwrite pre-built function '${body.function_name}'`
+            }
+        });
+        return;
+    }
```
kaposke, line 48: I'd say it's now questionable wether we want to allow this or not

packages/server/lib/services/remote-function/helpers.ts
```diff
@@ -0,0 +1,43 @@
+import { stringifyError } from '@nangohq/utils';
+
+import type { FunctionErrorCode } from '@nangohq/types';
+import type { Response } from 'express';
+
+export type RemoteFunctionStep = 'compilation' | 'deployment' | 'lookup' | 'execution';
+
+export function sendStepError({ res, step: _step, error, status }: { res: Response; step: RemoteFunctionStep; error: unknown; status?: number }): void {
```
kaposke: Step is unused?

packages/server/lib/services/remote-function/helpers.ts
```diff
@@ -0,0 +1,43 @@
...
+        const status = typeof err['status'] === 'number' ? err['status'] : undefined;
+        return { message, ...(code ? { code } : {}), ...(payload !== undefined ? { payload } : {}), ...(status ? { status } : {}) };
+    }
+    return { message: sanitizeMessage(stringifyError(error)) };
+}
+
+function sanitizeMessage(message: string): string {
+    return message
+        .replace(/\/[^\s"']+\/[^\s"']*/g, '<path>')
+        .replace(/https?:\/\/(localhost|127\.0\.0\.1|10\.\d+\.\d+\.\d+|192\.168\.\d+\.\d+)(:\d+)?[^\s]*/g, '<internal-url>')
```
kaposke: Curious whats the case for this specific regex?

packages/server/lib/services/local/compiler-client.ts
```diff
@@ -0,0 +1,39 @@
+import { randomUUID } from 'node:crypto';
+
+import { execDockerFileAsync, getExecErrorOutput, readContainerFile, writeContainerFile } from './docker.js';
+import { CompilerError, buildIndexTs, getFilePaths } from '../remote-function/compiler-client.js';
+import { remoteFunctionLocalImage, remoteFunctionProjectPath } from '../remote-function/runtime.js';
+
+import type { CompileRequest, CompileResult } from '../remote-function/compiler-client.js';
+
+const compilerTimeoutMs = 3 * 60 * 1000;
```
kaposke: Maybe env var?

packages/server/lib/services/local/docker.ts
```diff
@@ -0,0 +1,47 @@
+import { execFile, spawn } from 'node:child_process';
+import path from 'node:path';
+import { promisify } from 'node:util';
+
+export const execDockerFileAsync = promisify(execFile);
```
kaposke: If it's just a promisified version of execFile, why is it "execDockerFile"

packages/server/lib/services/local/compiler-client.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export async function invokeLocalCompiler(request: CompileRequest): Promise<CompileResult> {
+    const containerName = `nango-compiler-${randomUUID().slice(0, 8)}`;
+
+    try {
+        await execDockerFileAsync('docker', ['run', '-d', '--name', containerName, remoteFunctionLocalImage, 'sleep', '300'], { timeout: 10_000 });
```
kaposke: There might be a more deterministic way of waiting for completeness other than `sleep 300`?

kaposke: Not to important I guess since it's just for the local flow

packages/server/lib/services/remote-function/deploy-client.ts
```diff
@@ -0,0 +1,77 @@
...
+            const result = await sandbox.commands.run(command, {
+                cwd: remoteFunctionProjectPath,
+                timeoutMs: deployTimeoutMs,
+                envs
+            });
+            return { output: result.stdout };
+        } catch (err) {
+            if (err instanceof CommandExitError) {
+                return { output: err.stdout || err.stderr || JSON.stringify(err) };
+            }
```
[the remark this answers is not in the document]
kaposke: @marcindobry sounds like it is a problem?

packages/server/lib/controllers/functions/validation.ts
```diff
@@ -0,0 +1,30 @@
+import { z } from 'zod';
```
agusayerza, line 1: NIT: Each body could be on the endpoint file

packages/server/lib/routes.public.ts
```diff
@@ -228,6 +231,13 @@ publicAPI.route('/connect/session').get(connectSessionAuth, getConnectSession);
 publicAPI.route('/connect/session').delete(connectSessionAuth, deleteConnectSession);
 publicAPI.route('/connect/telemetry').post(connectSessionAuthBody, postConnectTelemetry);
 
+if (envs.NODE_ENV !== 'production') {
+    publicAPI.use('/remote-function', jsonContentTypeMiddleware);
+    publicAPI.route('/remote-function/compile').post(apiAuth, postRemoteFunctionCompile);
+    publicAPI.route('/remote-function/dryrun').post(apiAuth, postRemoteFunctionDryrun);
+    publicAPI.route('/remote-function/deploy').post(apiAuth, postRemoteFunctionDeploy);
+}
```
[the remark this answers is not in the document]
agusayerza: I think this is completely intentional, right?

packages/server/lib/services/remote-function/dryrun-client.ts
```diff
@@ -0,0 +1,105 @@
...
+    checkpoint?: Record<string, unknown>;
+    last_sync_date?: string;
+}
+
+export interface DryrunResult {
+    output: string;
+}
+
+const compileTimeoutMs = 3 * 60 * 1000;
+const dryrunTimeoutMs = 5 * 60 * 1000;
```
kaposke: Maybe env vars as well? Could become env vars when needed tbf

packages/server/lib/services/remote-function/helpers.ts
kaposke: What is the intended purpose of these? Might deserve a jsdoc

packages/server/lib/controllers/functions/compile/postCompile.ts
```diff
@@ -0,0 +1,57 @@
...
+            function_type: body.function_type,
+            code: body.code
+        });
+
+        res.status(200).send({
+            integration_id: body.integration_id,
+            function_name: body.function_name,
+            function_type: body.function_type,
+            bundle_size_bytes: result.bundleSizeBytes,
+            bundled_js: result.bundledJs,
```
agusayerza, line 46: Do we need to return the js? In practice, we are only using compile to check that it wouldn't fail, right?

packages/server/lib/routes.public.ts
```diff
@@ -228,6 +231,13 @@ publicAPI.route('/connect/session').get(connectSessionAuth, getConnectSession);
 publicAPI.route('/connect/session').delete(connectSessionAuth, deleteConnectSession);
 publicAPI.route('/connect/telemetry').post(connectSessionAuthBody, postConnectTelemetry);
 
+if (envs.NODE_ENV !== 'production') {
+    publicAPI.use('/remote-function', jsonContentTypeMiddleware);
+    publicAPI.route('/remote-function/compile').post(apiAuth, postRemoteFunctionCompile);
+    publicAPI.route('/remote-function/dryrun').post(apiAuth, postRemoteFunctionDryrun);
+    publicAPI.route('/remote-function/deploy').post(apiAuth, postRemoteFunctionDeploy);
```
agusayerza: Should we be stricter about rate-limiting these endpoints?

packages/server/lib/routes.public.ts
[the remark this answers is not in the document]
marcindobry: Yes

packages/server/lib/controllers/functions/compile/postCompile.ts
marcindobry: Then this wouldn't really be a compile step, but a type-check step. We can add into the skill that the response returns a bundledJs and the code that the agent executes to do call the API should filter this out to avoid polluting the context.

packages/server/lib/controllers/functions/deploy/postDeploy.ts
marcindobry: I'd keep it for now until we know agents can reliably ask whether to replace it or not.

packages/server/lib/services/local/compiler-client.ts
marcindobry: Honestly I thought the same but apparently this is pretty standard way of checking if a container is ready.

packages/server/lib/services/local/docker.ts
marcindobry: That's a question to Agustin's agent :D 

I guess because it's used in a docker.ts with 'docker' as the main argument 🤷

marcindobry: I'll change it so that it uses 'docker' as the first argument every time for it to make more sense

packages/server/lib/routes.public.ts
marcindobry: I guess if we run into sandbox cost issues we can make them tighter. For now I imagine it's just going to cause a documentation headache for exceptions like this.

packages/server/lib/services/remote-function/helpers.ts
marcindobry: It's replacing internal networking/filesystem data from error messages / stacktraces.

packages/server/lib/services/local/compiler-client.ts
marcindobry: I want to avoid adding speculative env vars for now given this is still basically a prototype that will change a lot.

agusayerza (approved):

PR #5846, opened 2026-04-13
fix(providers): fix salesforce jwt to use authorization_url in connection config

hassan254-prog:

- Fix `salesforce-jwt` to use the `authorization_url` from the connection config. This allows users to provide their own authorization URL instead of being limited to `login.salesforce.com`.


hassan254-prog:
Checked the connections count in production, and it’s only one, from my test account.

TBonnin (approved):

PR #5852, opened 2026-04-14
fix(providers): update datadog application key regex pattern

hassan254-prog:

- update datadog application key regex pattern
- Migrate datadog docs to new location


packages/providers/providers.yaml
```diff
@@ -4791,28 +4772,29 @@ datadog:
...
             example: 'us5.datadoghq.com'
+            order: 1
             doc_section: '#step-1-finding-your-datadog-domain'
         applicationKey:
             type: string
             title: Application Key
-            description: The application key required for read data access
-            pattern: '^[a-f0-9]{40}$'
+            description: The application key required for read data access.
+            pattern: '^([a-f0-9]{40}|ddapp_[A-Za-z0-9]+)$'
```
TBonnin: can we be slightly more lenient so we don't have to update the pattern next time datadog decide to change the format?

TBonnin (approved):

packages/providers/providers.yaml
hassan254-prog: I agree, updated with the latest [commit](https://github.com/NangoHQ/nango/pull/5852/commits/45c7fbd7894adbed7c1fdd8ac36828c742bd1b32).

packages/providers/providers.yaml
```diff
@@ -4791,28 +4791,29 @@ datadog:
...
             example: 'us5.datadoghq.com'
+            order: 1
             doc_section: '#step-1-finding-your-datadog-domain'
         applicationKey:
             type: string
             title: Application Key
-            description: The application key required for read data access
-            pattern: '^[a-f0-9]{40}$'
+            description: The application key required for read data access.
+            pattern: '^[A-Za-z0-9_]+$'
```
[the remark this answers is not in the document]
hassan254-prog: Decided to make this more linnet to prevent such breaks from datadog.

PR #5855, opened 2026-04-14
feat(orchestrator): serialize webhooks admision

pfreixes:
## Summary

Serialize webhook task creation in the scheduler using PostgreSQL advisory locks (`pg_advisory_xact_lock`). When multiple concurrent webhook enqueue requests target the same group key, the advisory lock forces them to be serialized, preventing DB overload from unbounded parallel inserts.

  - Tasks with group keys starting with `webhook:` acquire a transaction-scoped advisory lock keyed on `enqueue:{groupKey}` before creation
  - The lock is automatically released on transaction commit/rollback
  - Non-webhook tasks are unaffected

## Benchmark Results

  Measured on a local testcontainers PostgreSQL 15.5 instance, 20s per scenario:

  | Scenario | Producers | Advisory Lock | Throughput (tasks/s) | DB CPU (max) |
  |----------|-----------|---------------|---------------------|--------------|
  | 1 | 1 | No | 722 | ~57% |
  | 2 | 1 | Yes | 654 | ~54% |
 | 3 | 2 | Yes | 764 | ~66% |
  | 4 | 2 | No | 1,176 | ~142% |
  | 5 | 4 | No | 1,670 | ~300% |

  **Key observations:**
  - **Lock overhead (single producer):** ~9.4% throughput reduction, ~5% less CPU — the pure cost of acquiring `pg_advisory_xact_lock` per task
  - **Without lock, CPU scales faster than throughput:** 2 producers → 1.6x throughput at 2.5x CPU; 4 producers → 2.3x throughput at 5.3x CPU
  - The lock caps concurrent webhook producers to a ~constant throughput and ~constant  CPU regardless of producer count, which is the intended behavior

## Test plan

  - [x] Integration test: happy-path test for webhook group key through create/dequeue/succeed lifecycle
  - [x] Benchmark test: `npm run test:benchmark -- --dir packages/scheduler` validates serialization effect
  - [x] All existing scheduler integration tests pass (66/66)


TBonnin (changes requested):
see my comment inline. I don't think this is the best way to address the problem

packages/scheduler/lib/scheduler.ts
```diff
@@ -202,6 +206,10 @@ export class Scheduler {
                 };
             }
 
+            if (shouldSerializeImmediateEnqueue(taskProps.groupKey)) {
+                await trx.raw(`SELECT pg_advisory_xact_lock(?) as "lock_immediate_enqueue"`, [stringToHash(`enqueue:${taskProps.groupKey}`)]);
```
TBonnin: A few things:
- I think all types of immediate tasks should be handled, not just webhooks. The same exact scenario can happen with actions
- This doesn't reduce the amount of queries. It is just distributing them over time, potentially slowing down any downstream execution.

I think a better solution would be to collapse the counting query (aka: caching promises) in order to transform all concurrency queries into a single one, at the cost of a little bit of memory.

pfreixes: > I think all types of immediate tasks should be handled, not just webhooks. The same exact scenario can happen with actions

Yes technically, still the conditions are not the same, webhooks are unrate limitted in our public API and we also have the fan out of creating N tasks per webhook which makes the webhook case much more concerning.

But yes, I would agree that any other primitive can also 

> This doesn't reduce the amount of queries. It is just distributing them over time, potentially slowing down any downstream execution.

Yes it slows down the downstream, providing less throughput since operations are serialized. But it basically reduces the max load that a single group key can inflict to our DB which is the main goal

> I think a better solution would be to collapse the counting query (aka: caching promises) in order to transform all concurrency queries into a single one, at the cost of a little bit of memory.

Yea good point, so you mean moving into a queue of promises all requests that are targeting the same groups? and once we have the result from the first one we just allow the others to move ahead or reject considering the initial counter received?

pfreixes: ill go for a solution which is not super complicated by just reusing the same counter returned by the first request

TBonnin: have a look at this PR for an example of this promise coalescing pattern https://github.com/NangoHQ/nango/pull/5090/changes. 
It is not a queue of promise, it is a single promise (per group key) that is shared across requests. Sometimes using node has some benefits 😉. So if 1000s of requests hit at the same time for the same webhook for instance, the `count` query can be a single promise that all requests await

packages/scheduler/lib/models/tasks.ts
```diff
@@ -179,7 +179,27 @@ export async function create(
...
 }
 
+// Coalesce concurrent queueSizes queries for the same group keys.
+// When multiple immediate() calls target the same group key concurrently,
+// they share a single DB query instead of each running their own count.
+// The result may be slightly stale but the cap is a safeguard, not a strict limit.
+const inflightQueueSizes = new Map<string, Promise<Result<Map<string, number>>>>();
+
 export async function queueSizes(db: knex.Knex, opts: { groupKeys?: string[] | undefined }): Promise<Result<Map<string, number>>> {
+    const cacheKey = opts.groupKeys ? opts.groupKeys.sort().join(',') : '*';
```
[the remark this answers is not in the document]
TBonnin: Valid point. group key currently cannot have a `,` but let's not rely on this assumption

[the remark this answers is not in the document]
pfreixes: yea good point

TBonnin (approved):

PR #5860, opened 2026-04-14
feat(server): add client IP tags to Datadog traces

marcindobry:
Adds Datadog client IP extraction from `X-Forwarded-For` for server traces and captures the raw `X-Forwarded-For` request header for trace inspection.

NAN-4843: https://linear.app/nango/issue/NAN-4843/create-sampling-rules-for-control-plane-api-actions


TBonnin (approved):
Not a huge fan of willingly adding PII into monitoring data but I am not the decider

PR #5861, opened 2026-04-14
feat(usage): add usage metric querying from clickhouse

TBonnin:
Adding `getUsage` function to query metrics for a given account. The UX of the metric dashboard isn't fully scoped yet so we might need to modify the input/output shapes at some point. 
For now I went with a signature that could be used to fetch capping values (single metric, no granularity, time interval) and also power our current usage dashboard (multi-metric, day granularity, time interval) with support for one dimension.
See [tests](https://github.com/NangoHQ/nango/pull/5861/changes#diff-2e53c8e9c2d47885e599c56f4a5abd5fbf90097aa147d36a57000c068194cef3R74) for examples

The function isn't used yet


packages/usage/lib/clickhouse/clickhouse.ts
TBonnin, line 1: renamed `usage.ts` to `clickhouse.ts`. only the `getUsage` and `flush` functions are new.

packages/usage/lib/clickhouse/migrations/20260407000005_create_daily_webhook_forward.ts
```diff
@@ -1,6 +1,6 @@
 export const sql = [
     `
-    CREATE TABLE IF NOT EXISTS usage.daily_webhook_forward
+    CREATE TABLE IF NOT EXISTS usage.daily_webhook_forwards
```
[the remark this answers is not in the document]
TBonnin: since the tables are not used yet, it is ok if we dropped data. I will just recreate the views manually.

packages/usage/lib/clickhouse/clickhouse.ts
```diff
@@ -0,0 +1,206 @@
...
+            return row && row.value > 0 ? [row] : [];
+        });
+        return this.batcher.add(...rows);
+    }
+
+    flush(): Promise<Result<void>> {
+        return this.batcher ? this.batcher.flush() : Promise.resolve(Ok(undefined));
+    }
+
+    async getUsage(query: GetUsageQuery): Promise<Result<GetUsageResult>> {
```
pfreixes, line 71: Maybe it would be ideal to add a full integration test. If I’m not mistaken, we already have a dockerized ClickHouse setup, and that would help exercise the full set of combinations here.

packages/usage/lib/clickhouse/clickhouse.ts
```diff
@@ -0,0 +1,206 @@
...
+
+                    const total = rows.reduce((sum, row) => sum + row.quantity, 0);
+
+                    usage.metrics[metric] = {
+                        series,
+                        total,
+                        view_mode: viewMode
+                    };
+                } else {
+                    logger.error(`Failed to execute Clickhouse query for metric ${Object.keys(query.metrics)[i]}: ${stringifyError(entry.reason)}`);
```
pfreixes, line 161: maybe a metric for queries failed for clickouse would be interesting to have

packages/usage/lib/clickhouse/clickhouse.ts
```diff
@@ -0,0 +1,206 @@
...
+                                SUM(${aggColumnForMetric(metric)}) AS quantity,
+                                ${startSelect(query)}
+                                ${dimensionSelect} AS dimension
+                            FROM ${this.database}.${tableForMetric(metric)}
+                            WHERE account_id = ${query.accountId}
+                            AND day >= toDate('${query.timeframe.start.toISOString().split('T')[0]}')
+                            AND day < toDate('${query.timeframe.end.toISOString().split('T')[0]}')
+                            GROUP BY account_id ${granularityGroupBy(query)} ${dimensionGroupBy}
+                            ORDER BY account_id ${granularityOrderBy(query)} ${dimensionOrderBy}
+                        `;
```
pfreixes, line 100: same here, maybe a metric for tracking latencies

packages/usage/lib/clickhouse/clickhouse.ts
```diff
@@ -0,0 +1,206 @@
...
+                        const res = await this.client?.query({ query: sql, format: 'JSONEachRow' });
+                        return {
+                            metric,
+                            dimension,
+                            rows: (await res?.json()) as {
+                                account_id: number;
```
pfreixes: do we really require to return this one? since will be the same for all of the rows and if Im not mistaken we do not bubble up this info so we might want to just not inject it

packages/usage/lib/clickhouse/clickhouse.ts
```diff
@@ -0,0 +1,206 @@
...
+            });
+            const results = await Promise.allSettled(qs);
+            const usage: GetUsageResult = {
+                accountId: query.accountId,
+                granularity: query.granularity,
+                metrics: {}
+            };
+            for (const [i, entry] of results.entries()) {
+                if (entry.status === 'fulfilled') {
+                    if (!entry.value) continue;
```
pfreixes, line 129: is this an expected case? or its just a safeguard?

pfreixes (approved):
left some comments, feel free to address them if you believe that are important

packages/usage/lib/clickhouse/clickhouse.ts
```diff
@@ -0,0 +1,206 @@
...
+                        const sql = `
+                            SELECT
+                                account_id,
+                                SUM(${aggColumnForMetric(metric)}) AS quantity,
+                                ${startSelect(query)}
+                                ${dimensionSelect} AS dimension
+                            FROM ${this.database}.${tableForMetric(metric)}
+                            WHERE account_id = ${query.accountId}
+                            AND day >= toDate('${query.timeframe.start.toISOString().split('T')[0]}')
+                            AND day < toDate('${query.timeframe.end.toISOString().split('T')[0]}')
```
pfreixes, line 97: Is this intended? asking bc seems that we are [supporting](https://github.com/NangoHQ/nango/blob/master/packages/server/lib/controllers/v1/plans/usage/getBillingUsage.ts#L16) full data time query (unless Im missing something here )

packages/usage/lib/clickhouse/clickhouse.ts
TBonnin: expected case. For instance some metrics like records or connection aren't supported and would returned `undefined`

packages/usage/lib/clickhouse/clickhouse.ts
TBonnin: https://github.com/NangoHQ/nango/pull/5861/changes#diff-2e53c8e9c2d47885e599c56f4a5abd5fbf90097aa147d36a57000c068194cef3 is a full integration tests suite with both ingestion and querying

packages/usage/lib/clickhouse/clickhouse.ts
TBonnin: The usage data is aggregated per `day` so yes it is intended. 
I known the API support full datetime but in reality the frontend is only asking for a month timeframe. https://github.com/NangoHQ/nango/blob/4619db7f35de611c4cc4665f840fb39f119b0190/packages/webapp/src/pages/Team/Billing/components/Usage.tsx?plain=1#L23-L24

packages/usage/lib/clickhouse/clickhouse.ts
TBonnin: yep we don't need it

packages/usage/lib/clickhouse/clickhouse.ts
TBonnin: we'll add full monitoring when/if going to prod

PR #5862, opened 2026-04-14
feat(webapp): Records list page

agusayerza:
## Summary
Screenshot 2026-04-14 at 6 10 55 PM
Screenshot 2026-04-14 at 6 11 08 PM
Screenshot 2026-04-14 at 6 11 15 PM

- Adds a new Records experience to connection pages, including a model list, variant-aware drilldown, paginated records table, and lazy-loaded payload viewer.
- Adds the private connection records API, shared types, and webapp hooks needed to list record models/variants and fetch records for a connection.
- Lays groundwork for the upcoming connection page redesigns by introducing the shared connection sidebar layout and moving the current connection pages onto that structure.
- Hardens records pagination/cursor handling and adds coverage for invalid cursor handling, pagination, and model/variant listing.
## Issue
- `NAN-5080`
## Testing
- Open a connection with records and verify the new Records tab:
  - lists models and variants correctly
  - paginates through records correctly
  - opens the payload dialog only when requested
- Verify the shared connection page layout still renders correctly for the existing connection tabs.
- `npm run test:unit -- packages/records/lib/cursor.unit.test.ts`
- `npm run test:integration -- packages/server/lib/controllers/v1/connections/connectionId/records/getModels.integration.test.ts packages/server/lib/controllers/v1/connections/connectionId/records/getRecords.integration.test.ts`
- `npx tsc -p packages/records/tsconfig.json --noEmit`
- `npx tsc -p packages/server/tsconfig.json --noEmit`
- `npx tsc -p packages/webapp/tsconfig.json --noEmit`


packages/records/lib/cursor.ts
```diff
@@ -3,14 +3,17 @@ export interface Cursor {
...
+
 export const Cursor = {
     new({ last_modified_at, id }: { last_modified_at: string; id: string }): string {
         return Buffer.from(`${last_modified_at}||${id}`).toString('base64');
     },
     from(encoded: string): Cursor | undefined {
         const decoded = Buffer.from(encoded, 'base64').toString('ascii');
         const [cursorSort, cursorId] = decoded.split('||');
-        if (cursorSort && cursorId) {
+        if (cursorSort && cursorId && ISO_TIMESTAMP_RE.test(cursorSort) && UUID_RE.test(cursorId)) {
```
TBonnin: would it make sense to use zod instead of regexes?

TBonnin (commented):
backend changes looks good 👍

packages/records/lib/models/records.ts
```diff
@@ -163,13 +167,16 @@ export async function getRecords({
...
+
         let query = dbRead
             .from<FormattedRecord>(RECORDS_TABLE)
             .timeout(60000) // timeout after 1 minute
             .where({ connection_id: connectionId, model })
             .orderBy([
-                { column: 'updated_at', order: 'asc' },
-                { column: 'id', order: 'asc' }
+                { column: 'updated_at', order: isDesc ? 'desc' : 'asc' },
+                { column: 'id', order: isDesc ? 'desc' : 'asc' }
```
TBonnin: you can assign sortOrder directly

packages/webapp/src/index.css
```diff
@@ -264,4 +264,21 @@
...
+
+.mantine-Prism-code::-webkit-scrollbar-thumb {
+    background-color: var(--border-strong);
+    border-radius: 5px;
+    border: 2px solid var(--surface-canvas);
+}
+
+.mantine-Prism-code::-webkit-scrollbar-track {
+    background-color: var(--surface-canvas);
 }
```
macko911: Do we want custom scrollbars here? It's the only custom scrollbar styling in the app (everything else is native), it applies to *every* `.mantine-Prism-code` block app-wide, and it conflicts with `RecordsTab`'s table — which *hides* its scrollbar. Suggest dropping it for native, or making it a shared design-system utility if we do want styled scrollbars.

macko911 (approved):

packages/webapp/src/index.css
agusayerza: Removing them

PR #5863, opened 2026-04-14
fix(webapp): show error message for suspended users

hassan254-prog:

- show error message for suspended users


kaposke (approved):

packages/shared/lib/services/user.service.ts
```diff
@@ -10,8 +10,13 @@ const VERIFICATION_EMAIL_EXPIRATION = 3 * 24 * 60 * 60 * 1000;
 const envs = parseEnvs(ENVS);
 
 class UserService {
-    async getUserById(id: number): Promise<DBUser | null> {
-        const result = await db.knex.select<DBUser>('*').from<DBUser>(`_nango_users`).where({ id, suspended: false }).first();
+    async getUserById(id: number, includeSuspended = false): Promise<DBUser | null> {
```
TBonnin, line 13: can we make the second object an `opts` object to make it clearer at the calling site
```
service.getUser(id, true) 
// 
service.getUser(id, { includesSuspended: true})
```

PR #5865, opened 2026-04-15
feat(orchestrator): Add dropped task metric for capped groups

pfreixes:
Adds a new metric for knowing when we are silently dropping tasks. This can be useful during incidents.

### DIFF
--- a/packages/scheduler/lib/models/tasks.ts
+++ b/packages/scheduler/lib/models/tasks.ts
@@ -1,6 +1,6 @@
 import { uuidv4, uuidv7 } from 'uuidv7';
 
-import { Err, Ok, stringToHash, stringifyError } from '@nangohq/utils';
+import { Err, Ok, metrics, stringToHash, stringifyError } from '@nangohq/utils';
 
 import { taskStates } from '../types.js';
 import { SCHEDULES_TABLE } from './schedules.js';
@@ -141,7 +141,7 @@ export async function create(
 
         const now = new Date();
         const toInsertPerGroup = new Map<string, Task[]>();
-        const cappedGroupKeys = new Set<string>();
+        const cappedGroupCounts = new Map<string, number>();
         for (const props of taskProps) {
             if (!toInsertPerGroup.has(props.groupKey)) {
                 toInsertPerGroup.set(props.groupKey, []);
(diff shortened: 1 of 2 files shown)


packages/scheduler/lib/models/tasks.ts
```diff
@@ -160,9 +160,17 @@ export async function create(
                     retryKey: props.retryKey || uuidv4()
                 });
             } else {
-                cappedGroupKeys.add(props.groupKey);
+                cappedGroupCounts.set(props.groupKey, (cappedGroupCounts.get(props.groupKey) ?? 0) + 1);
```
TBonnin, line 163: not sure of the value of this metric to be honest. Dropping tasks is usually the result of some other problem upstream that we have plenty of signal for.
If you really want it though just `increment` here, no need to aggregate, DD lib will do it for you

packages/scheduler/lib/models/tasks.ts
```diff
@@ -160,9 +160,17 @@ export async function create(
...
             }
         }
+        const droppedCountPerPrimitive = new Map<string, number>();
+        for (const [groupKey, droppedCount] of cappedGroupCounts.entries()) {
+            const primitive = groupKey.split(':')[0] || 'unknown';
```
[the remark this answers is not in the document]
pfreixes: number of primitives is low cardinality, but thanks for the heads up.

packages/scheduler/lib/models/tasks.ts
pfreixes: I could be wrong here, but at least i had no visibility on this during the latest incident, why we could suspicious that with some proxy metrics there was no an explicit datapoint for this (as said I could be wrong here so happy to learn what other signals we have for this)

Also its not only about incidents, and is similar to the max backpressure signal that we added for knowing when customers were reaching to their max concurrency threshold, with this new metric we can pin point specific time windows for knowing if one or multiple customers got impacted (we will still require the logging part for framing specific issues for specific customers)

Note: just for clarification, upstream -> dependencies of our services, downstream -> users of our services? or the other way arround? just asking bc Ive seen using them in both directions, and is worth it to get all of us aligned)

TBonnin: Dropped tasks are the result of tasks being created faster than being executed, which we already have several metrics for. Knowing and understanding why tasks are created too fast and/or not executed fast enough is the valuable piece of information. Dropped tasks is basically the same information reported as another metric.

pfreixes: > Dropped tasks are the result of tasks being created faster than being executed, which we already have several metrics for.

Sorry but I kinda disagree here, dropped tasks is the canonical source of truth, the other one is a proxy metric which you could infer that dropping is happening but without explicit datapoints - the ones provided by the dropped metric - its just an educated guess.

>. Knowing and understanding why tasks are created too fast and/or not executed fast enough is the valuable piece of information.

Related to the previous one, you know that a customer is sending data too fast or that the system is going to slow when data is being discarded, with the creation rate or execution rate you have proxy metrics about that.

Yea knowing what is causing is fundamental but its more a secondary step IMO, first you require data for knowing that this is happening.

pfreixes: Also, tasks.dropped gives us an explicit measure of discarded work volume. Creation and execution rates are only proxy metrics for pressure. With a direct dropped-tasks metric, we could even monitor absolute drops or dropped percentage relative to created tasks.

pfreixes: for having the % we would require to have a total counter, this would be IMO ideal for paths like the schedule, were we can emit total tried to schedule vs dropped, this can help you to quantify the impact of an incident.

Happy to add also this counter in this PR

TBonnin: feel free to add the metric. Just let ddtrace do the aggregation.

I disagree that dropping tasks is the canonical source of truth. With it alone you cannot even infer if this is a 'too many creations' or a 'not enough executions' problem.

rossmcewan (approved):

PR #5867, opened 2026-04-15
fix(cli): remove --ai variant from nango init command

rguldener:
The --ai option generated a claude.md (or Cursor rules) file with outdated instructions. Removes the option, its handler, and the bundled instructions template.

Solves NAN-5287


rguldener:
@marcindobry Looks reasonable at first glance.

I don't think claude tested the CLI because my machine isn't setup for local development

marcindobry (approved):

PR #5872, opened 2026-04-15
fix(clickhouse): compute_gbms to be calculated in mv

TBonnin:
compute_gbms was wrongly calculated because memory_gb was summed instead of being part of the aggregation group_by. Fixing this issue as well as pre-calculating compute_gbms

### DIFF
--- a/packages/usage/lib/clickhouse/migrations/20260407000003_create_daily_function_executions.ts
+++ b/packages/usage/lib/clickhouse/migrations/20260407000003_create_daily_function_executions.ts
@@ -13,11 +13,11 @@ export const sql = [
         runtime          LowCardinality(String),
         value            Int64,
         duration_ms      UInt64,
-        memory_gb        Float64,
+        compute_gbms     Float64,
         custom_logs      UInt64,
         proxy_calls      UInt64
     )
-    ENGINE = SummingMergeTree((value, duration_ms, memory_gb, custom_logs, proxy_calls))
+    ENGINE = SummingMergeTree((value, duration_ms, compute_gbms, custom_logs, proxy_calls))
     PARTITION BY toYYYYMM(day)
     ORDER BY (account_id, day, environment_id, integration_id, connection_id, function_type, function_name, success, runtime)
     TTL day + INTERVAL 24 MONTH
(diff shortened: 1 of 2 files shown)


packages/usage/lib/clickhouse/migrations/20260407000003_create_daily_function_executions.ts
```diff
@@ -37,7 +37,7 @@ export const sql = [
         attributes.runtime::String                                              AS runtime,
         sum(value)                                                              AS value,
         sum(coalesce(attributes.telemetryBag.durationMs::Nullable(UInt64), 0))  AS duration_ms,
-        sum(coalesce(attributes.telemetryBag.memoryGb::Nullable(Float64), 0.0))    AS memory_gb,
```
TBonnin, line 40: `memory_gb` was wrongly summed. We are now pre-calculating `compute_gbms` to fix the calculation and reduce the number of rows in the daily table

packages/usage/lib/clickhouse/migrations/20260407000003_create_daily_function_executions.ts
```diff
@@ -13,11 +13,11 @@ export const sql = [
         runtime          LowCardinality(String),
         value            Int64,
         duration_ms      UInt64,
-        memory_gb        Float64,
+        compute_gbms     Float64,
         custom_logs      UInt64,
         proxy_calls      UInt64
     )
-    ENGINE = SummingMergeTree((value, duration_ms, memory_gb, custom_logs, proxy_calls))
+    ENGINE = SummingMergeTree((value, duration_ms, compute_gbms, custom_logs, proxy_calls))
```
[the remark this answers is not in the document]
TBonnin: table and view aren't used yet so I will just replace them and drop the existing data

packages/usage/lib/clickhouse/migrations/20260407000003_create_daily_function_executions.ts
```diff
@@ -37,7 +37,7 @@ export const sql = [
         attributes.runtime::String                                              AS runtime,
         sum(value)                                                              AS value,
         sum(coalesce(attributes.telemetryBag.durationMs::Nullable(UInt64), 0))  AS duration_ms,
-        sum(coalesce(attributes.telemetryBag.memoryGb::Nullable(Float64), 0.0))    AS memory_gb,
+        sum(coalesce(attributes.telemetryBag.durationMs::Nullable(UInt64), 0) * coalesce(attributes.telemetryBag.memoryGb::Nullable(Float64), 0.0)) AS compute_gbms,
```
rossmcewan, line 40: I don't believe this is correctly set in the codebase. I will raise a PR to fix this for Lambda.

rossmcewan (approved):

PR #5873, opened 2026-04-15
feat(logs): add OpenSearch backend selectable via NANGO_LOGS_PROVIDER

rossmcewan:
Wire @opensearch-project/opensearch client, ISM policies, shared mappings, and isLogsNotFoundError for API handlers.

Made-with: Cursor


packages/logs/lib/storage/client.ts
```diff
@@ -0,0 +1,322 @@
...
+    async setupPolicies(client, policies) {
+        await client.ilm.putLifecycle(policies.messagesPolicy as any);
+        await client.ilm.putLifecycle(policies.operationsPolicy as any);
+    }
+};
+
+const openSearchAdapter: LogsStorageAdapter = {
+    provider: 'opensearch',
+    createRawClient(common) {
+        return new OpenSearchClient(common) as unknown as ElasticsearchClient;
```
TBonnin: casting into an ElasticsearchClient means typescript won't catch any future call that isn't covered by the mapper and it will crash at runtime.

packages/logs/lib/utils.ts
```diff
@@ -23,4 +25,33 @@ export const logLevelToLogger = {
...
+        } as any);
+    }
+
+    throw new esErrors.ResponseError({
+        statusCode: 404,
+        warnings: [],
+        body: {},
+        headers: {}
+    } as any);
+}
```
TBonnin, line 52: Not sure why we are not defining an error 
```
export class LogsNotFoundError extends Error {
    readonly statusCode = 404;
    constructor() { super('logs_not_found'); }
}

// then
throw new LogsNotFoundError
// and 
if(e instanceof LogsNotFoundError)
```

packages/logs/lib/storage/client.ts
```diff
@@ -0,0 +1,322 @@
+import { Client as ElasticsearchClient } from '@elastic/elasticsearch';
+import { Client as OpenSearchClient } from '@opensearch-project/opensearch';
+
+import { envs } from '../env.js';
+import { CircuitBreaker } from './circuitBreaker.js';
+import { putIsmPolicies } from '../opensearch/ismPolicies.js';
+
+/** Typed as Elasticsearch client; OpenSearch client is API-compatible for our usage. */
+export type LogsStorageClient = ElasticsearchClient;
```
TBonnin: not a huge fan (see my comment below around the forced casting)
I would definitely prefer to be explicit by defining a `LogsStorageClient` interface with the half-dozen methods we are using and implement the 2 versions explicitly.

```ts
export interface LogsStorageClient {
    search(params: SearchParams): Promise<SearchResult>;
    index(params: IndexParams): Promise<void>;
    updateByQuery(params: UpdateByQueryParams): Promise<void>;
    close(): Promise<void>;
...
}

export function createElasticsearchClient(config: ClientConfig): LogsStorageClient {
    const raw = new Client(config);
    return {
        search: (p) => raw.search(p),
        index:  (p) => raw.index(p),
        close: () => raw.close(),
...
    };
}

export function createOpenSearchClient(config: ClientConfig): LogsStorageClient {
    const raw = new Client(config);
    return {
        search: async (p) => {
            const { query, ..., ...rest } = p;
            const res = await raw.search({ ...rest, body: { query, ... } });
            return res.body; 
        },
        index: async (p) => {
            const { document, ...rest } = p;
            await raw.index({ ...rest, body: document });
        },
        close: () => raw.close(),
    };
}
``

**Comment by rossmcewan** [comment]
@codex review

File: packages/logs/lib/storage/circuitBreakerLogsClient.ts [hunk 4]
```diff
@@ -0,0 +1,103 @@
...
+export interface LogsStorageBackend extends LogsStorageClient {
+    healthCheck(): Promise<boolean>;
+    setupRetentionPolicies(policies: LogsStoragePolicies): Promise<void>;
+}
+
+function guard(circuitBreaker: CircuitBreaker): void {
+    if (circuitBreaker.isUnhealthy()) {
+        throw new Error('Logs storage circuit breaker is unhealthy - failing fast');
+    }
+}
```
TBonnin: I would move the guard function inside the client so it can be called without having to inject the circuit breaker every time

packages/design-system/tokens/tokens.json
TBonnin, line 1: doesn't belong to this PR 😄

packages/logs/lib/env.ts
```diff
@@ -1,9 +1,15 @@
...
+import type { z } from 'zod';
+
 // Do not require in community and enterprise right now
 const required = process.env['NANGO_LOGS_ENABLED'] === 'true';
 
-export const envs = parseEnvs(required ? ENVS.required({ NANGO_LOGS_ES_URL: true, NANGO_LOGS_ES_USER: true, NANGO_LOGS_ES_PWD: true }) : ENVS);
+export const envs = parseEnvs(required ? ENVS.required({ NANGO_LOGS_ES_URL: true, NANGO_LOGS_ES_USER: true, NANGO_LOGS_ES_PWD: true }) : ENVS) as z.infer<
+    typeof ENVS
+> & {
+    NANGO_LOGS_PROVIDER: 'elasticsearch' | 'opensearch';
```
TBonnin: why the casting? NANGO_LOGS_PROVIDER is correctly defined in `parse.ts`

packages/logs/lib/storage/types.ts
```diff
@@ -0,0 +1,128 @@
...
+    maxRetries: number;
+    auth: {
+        username: string;
+        password: string;
+    };
+}
+
+export interface LogsStoragePolicies {
+    messagesPolicy: unknown;
+    operationsPolicy: unknown;
```
TBonnin: is there no way to type them?

packages/logs/lib/storage/openSearchClient.ts
```diff
@@ -0,0 +1,237 @@
...
+        const body = unwrapOpenSearchResult<{ _index: string }>(res);
+        return { _index: body._index };
+    }
+
+    async update(params: LogsUpdateParams): Promise<void> {
+        await this.client.update(params);
+    }
+
+    async updateByQuery(params: LogsUpdateByQueryParams): Promise<void> {
+        await this.client.updateByQuery(splitUpdateByQueryParams(params) as unknown as Parameters<OpenSearchClient['updateByQuery']>[0]);
```
TBonnin: why is `splitUpdateByQueryParams` not returning the expected type?

packages/logs/lib/storage/openSearchClient.ts
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    readonly indices = {
+        existsIndexTemplate: async (params: { name: string }): Promise<boolean> => {
+            const res = await this.client.indices.existsIndexTemplate(params);
+            return unwrapOpenSearchResult(res);
+        },
+        putIndexTemplate: async (params: LogsPutIndexTemplateParams): Promise<void> => {
+            await this.client.indices.putIndexTemplate(
+                mapPutIndexTemplateParams(params) as unknown as Parameters<OpenSearchClient['indices']['putIndexTemplate']>[0]
```
TBonnin: same question about `mapPutIndexTemplateParams` not returning the expected type?

TBonnin (approved):

PR #5874, opened 2026-04-15
feat!: stop generating schema.ts / schema.json and stop accepting top-level json-schema

kaposke:
Stop generating schema.ts / schema.json as announced in https://nango.dev/docs/updates/dev#march-16-2026
~However, I also removed the handling for it in the server-side, which would break any CLI version before 0.64.42 (i believe). So ready to undo that if agreed to (though it would be a nice cleanup).~


TBonnin (approved):

packages/shared/lib/services/file/remote.service.ts
```diff
@@ -136,10 +136,6 @@ class RemoteFileService {
         }
     }
 
-    async getPublicTemplateJsonSchemaFile(integrationName: string): Promise<string | null> {
```
kaposke, line 139: Was unused

PR #5875, opened 2026-04-16
fix(lambda): track memory and duration correctly

rossmcewan:

### DIFF
--- a/packages/jobs/lib/invocations/lambda.processor.ts
+++ b/packages/jobs/lib/invocations/lambda.processor.ts
@@ -56,18 +56,30 @@ export class LambdaInvocationsProcessor {
 
         if (parsedMessage.responseContext.functionError === 'Unhandled') {
             const errorMessage = parsedMessage.responsePayload.errorMessage;
+            const errorType = lambdaErrorTypeFromMessage(errorMessage);
             await handle({
                 taskId: parsedMessage.requestPayload.taskId,
                 nangoProps: parsedMessage.requestPayload.nangoProps as unknown as NangoProps,
                 error: {
-                    type: lambdaErrorTypeFromMessage(errorMessage),
+                    type: errorType,
                     payload: { errorMessage },
                     status: parsedMessage.responseContext.statusCode
                 },
-                telemetryBag: { customLogs: 0, proxyCalls: 0, durationMs: 0, memoryGb: 0 },
+                telemetryBag: {
+                    customLogs: 0,
+                    proxyCalls: 0,
+                    durationMs: getDurationMsFromErrorType(errorType),
+                    memoryGb: envs.LAMBDA_DEFAULT_MEMORY_MB / 1024
+                },
                 functionRuntime: 'lambda',
                 checkpoints: null
             });
         }
     }
+}
+
+function getDurationMsFromErrorType(errorType: 'function_runtime_out_of_memory' | 'function_runtime_timed_out' | 'function_runtime_other'): number {
+    if (errorType === 'function_runtime_out_of_memory') return envs.LAMBDA_EXECUTION_TIMEOUT_SECS * 1000;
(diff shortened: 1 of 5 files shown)


packages/jobs/lib/invocations/lambda.processor.ts
```diff
@@ -56,18 +56,30 @@ export class LambdaInvocationsProcessor {
...
-                    type: lambdaErrorTypeFromMessage(errorMessage),
+                    type: errorType,
                     payload: { errorMessage },
                     status: parsedMessage.responseContext.statusCode
                 },
-                telemetryBag: { customLogs: 0, proxyCalls: 0, durationMs: 0, memoryGb: 0 },
+                telemetryBag: {
+                    customLogs: 0,
+                    proxyCalls: 0,
+                    durationMs: getDurationMsFromErrorType(errorType),
```
rossmcewan, line 71: We can't get accurate numbers here but the way it is at the moment, it means functions that exceed the Lambda timeout wouldn't accumulate any compute usage.

packages/jobs/lib/invocations/lambda.processor.ts
```diff
@@ -56,18 +56,30 @@ export class LambdaInvocationsProcessor {
...
+                    memoryGb: envs.LAMBDA_DEFAULT_MEMORY_MB / 1024
```
rossmcewan, line 72: Using the default as we don't know what the actual value was at this point.

packages/jobs/lib/invocations/lambda.processor.ts
```diff
@@ -56,18 +56,30 @@ export class LambdaInvocationsProcessor {
...
                 functionRuntime: 'lambda',
                 checkpoints: null
             });
         }
     }
 }
+
+function getDurationMsFromErrorType(errorType: 'function_runtime_out_of_memory' | 'function_runtime_timed_out' | 'function_runtime_other'): number {
+    if (errorType === 'function_runtime_timed_out') return envs.LAMBDA_EXECUTION_TIMEOUT_SECS * 1000;
+    if (errorType === 'function_runtime_out_of_memory') return envs.LAMBDA_DEFAULT_TIMEOUT_BILLING_SECS * 1000;
```
rossmcewan, line 83: use a new env var, default is 10s when out of memory occurs

packages/jobs/lib/invocations/lambda.processor.ts
```diff
@@ -56,18 +56,30 @@ export class LambdaInvocationsProcessor {
...
+                },
                 functionRuntime: 'lambda',
                 checkpoints: null
             });
         }
     }
 }
+
+function getDurationMsFromErrorType(errorType: 'function_runtime_out_of_memory' | 'function_runtime_timed_out' | 'function_runtime_other'): number {
+    if (errorType === 'function_runtime_timed_out') return envs.LAMBDA_EXECUTION_TIMEOUT_SECS * 1000;
```
rossmcewan, line 82: use default execution timeout as we don't know what the actual timeout was

packages/jobs/lib/invocations/lambda.processor.ts
TBonnin: we currently hard code it to `1` for the runner runtime so don't worry too much about the real value for now ;)

packages/jobs/lib/invocations/lambda.processor.ts
TBonnin: so we would account for 10s regardless of when the OOM happens?

TBonnin (approved):

packages/jobs/lib/invocations/lambda.processor.ts
rossmcewan: With this implementation yes - it's an env var so could default it to something else.

PR #5886, opened 2026-04-16
feat(clickhouse): add support for connections and records metrics

TBonnin:
- ingest connections and records snapshot data
- add materialized views for records and connections metrics
- add support for records and connections metric in clickhouse querying

Note: records and connections metrics are calculated the same way we do with Orb today, averaging all the ingested value across a day (assuming the ingestion interval is regular, given all ingested value the same weight)

### DIFF
--- a/packages/metering/lib/crons/usage.ts
+++ b/packages/metering/lib/crons/usage.ts
@@ -1,16 +1,19 @@
 import tracer from 'dd-trace';
 import * as cron from 'node-cron';
+import { uuidv7 } from 'uuidv7';
 
 import { billing as usageBilling } from '@nangohq/billing';
 import { getLocking } from '@nangohq/kvstore';
 import { records } from '@nangohq/records';
 import { connectionService } from '@nangohq/shared';
+import { Clickhouse } from '@nangohq/usage';
 import { flagHasUsage, getLogger, metrics } from '@nangohq/utils';
 
 import { envs } from '../env.js';
 
 import type { Lock } from '@nangohq/kvstore';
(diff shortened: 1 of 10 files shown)


packages/metering/lib/crons/usage.ts
```diff
@@ -57,10 +63,98 @@ export async function exec(): Promise<void> {
...
+
+                    const connectionIds = page.value.map((c) => c.connection.id);
+                    for await (const recordCounts of records.paginateCounts({ connectionIds })) {
+                        if (recordCounts.isErr()) {
+                            throw recordCounts.error;
+                        }
+                        for (const recordCount of recordCounts.value) {
+                            const connection = connectionsMap.get(recordCount.connection_id);
+                            if (connection) {
+                                const res = clickhouse.addRaw([
```
pfreixes, line 108: Should we store them into a list and batch after the loop?

pfreixes (commented):
Left one comment, sorry GH is down for reviewing so couldn't go beyond that file.

packages/metering/lib/crons/usage.ts
TBonnin: `clickhouse` has it own internal batcher and since each record count is independent and doesn't require any aggregation (unlike connections, the `record_counts` is itself an aggregation of the count and size per connection/model) we can just let `clickhouse` to do the batching

pfreixes: > we can just let clickhouse to do the batching

Then we are ok!

pfreixes (approved):

PR #5890, opened 2026-04-17
fix(server): Fix remote function sandbox host and diagnostics

marcindobry:
## Summary
- pass the public Nango API host into remote-function dryrun/deploy sandboxes in cloud
- preserve full command diagnostics for compile/dryrun/deploy failures
- make CLI dryrun setup failures exit non-zero and include deploy fetch causes

## Verification
- reproduced production compile/dryrun/deploy failures with the provided dev key/test connection
- verified valid dryrun and deploy via CLI with the public host
- verified valid dryrun and deploy through the local Docker sandbox wrapper
- npm run test:unit -- packages/server/lib/services/remote-function/runtime.unit.test.ts
- npx tsc -p packages/server/tsconfig.json --noEmit
- earlier focused remote-function unit/integration tests passed


packages/server/lib/services/remote-function/helpers.ts
```diff
@@ -77,10 +79,10 @@ function sanitizeMessage(message: string): string {
     });
 
     return messageWithUrlPlaceholders
-        .replace(/\/[^\s"']+/g, redactAbsolutePath)
+        .replace(/(^|[\s"'(])(\/[^\s"')]+)/g, (_, prefix: string, absolutePath: string) => `${prefix}${redactAbsolutePath(absolutePath)}`)
```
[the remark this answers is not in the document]
marcindobry: Not worth spending the time fixing for now

packages/server/lib/services/remote-function/command-output.ts
```diff
@@ -0,0 +1,128 @@
...
+        stderr: typeof err['stderr'] === 'string' ? err['stderr'] : undefined
+    });
+    if (output) {
+        return output;
+    }
+
+    return typeof err['message'] === 'string' && err['message'] ? err['message'] : fallback;
+}
+
+export function combineCommandOutput({ stdout, stderr }: { stdout?: string | undefined; stderr?: string | undefined }): string {
```
kaposke, line 23: I think these "hard to understand" functions deserve jsdocs (a description is enough) since the name doesn't really tell the full story.

This combines the two streams while checking if they're duplicates and only keeping stdout if so?

packages/server/lib/services/remote-function/command-output.ts
```diff
@@ -0,0 +1,132 @@
...
+    return [stdout, stderr]
+        .filter((value): value is string => typeof value === 'string' && value.length > 0)
+        .map((value) => value.trimEnd())
+        .filter((value, index, parts) => value && parts.indexOf(value) === index)
+        .join('\n');
+}
+
+export function isCompilationFailureOutput(output: string): boolean {
+    return /✗ Typechecking|Found \d+ errors?/i.test(output);
```
kaposke, line 32: Oh these are sooo prone to breaking

packages/server/lib/services/remote-function/command-output.ts
```diff
@@ -0,0 +1,128 @@
...
+}
+
+export function getDryrunCommandErrorOutput({ stdout, stderr }: { stdout?: string | undefined; stderr?: string | undefined }): string | undefined {
+    const output = combineCommandOutput({ stdout, stderr });
+    if (!output) {
+        return undefined;
+    }
+
+    if (/An error occurred during execution|Connection not found|No script matched/i.test(output)) {
```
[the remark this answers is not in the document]
kaposke: True, and prone to breaking as well

packages/server/lib/services/remote-function/command-output.unit.test.ts
```diff
@@ -0,0 +1,127 @@
...
+    getDryrunCommandSuccessOutput,
+    isCompilationFailureOutput,
+    parseDeploySuccessOutput,
+    parseDryrunSuccessOutput
+} from './command-output.js';
+
+describe('remote function command output helpers', () => {
+    it('preserves stdout and stderr when both contain useful diagnostics', () => {
+        expect(
+            combineCommandOutput({
```
kaposke: Maybe group these by tested methods? Currently have to look at test content to understand what it's testing.

packages/server/lib/services/remote-function/helpers.ts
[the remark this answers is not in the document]
kaposke: as long as we eventually do

packages/server/lib/services/remote-function/runtime.ts
```diff
@@ -13,3 +16,14 @@ const remoteFunctionSandboxTimeoutBufferMs = 30 * 1000;
 export const remoteFunctionCompilerSandboxTimeoutMs = remoteFunctionCompileTimeoutMs + remoteFunctionSandboxTimeoutBufferMs;
 export const remoteFunctionDeploySandboxTimeoutMs = remoteFunctionDeployTimeoutMs + remoteFunctionSandboxTimeoutBufferMs;
 export const remoteFunctionDryrunSandboxTimeoutMs = remoteFunctionCompileTimeoutMs + remoteFunctionDryrunTimeoutMs + remoteFunctionSandboxTimeoutBufferMs;
+
+export function getRemoteFunctionNangoHost(): string {
+    if (isCloud) {
+        return isStaging ? stagingHost : cloudHost;
```
kaposke, line 22: Should we support development?

kaposke (commented):
Main concern is the regex-based success/failure evaluation system and extraction of deployed functions. Considering we own the CLI, is it something we could do better?

packages/server/lib/services/remote-function/command-output.ts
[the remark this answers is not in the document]
marcindobry: The fact that they are mapped into a failure is intentional and also the fact that they are returned as success exit codes is what forces this error prone method to exist unfortunately :/ I didn't want to modify the CLI just for this experiment, it would be a borderline breaking change.

packages/server/lib/services/remote-function/command-output.ts
marcindobry: Yup but without changing the CLI itself it is hard to build something better. I didn't want to modify the CLI for this experiment, but it would definitely be something we should do if we productionize this

packages/server/lib/services/remote-function/runtime.ts
marcindobry: I don't think it's necessary at this point. We'll ship this to production and let people play around with it for now

packages/server/lib/controllers/functions/dryrun/postDryrun.ts
```diff
@@ -52,22 +54,23 @@ export const postRemoteFunctionDryrun = asyncWrapper<PostRemoteFunctionDryrun>(a
...
+        const dryrunOutput = parseDryrunSuccessOutput(result.output);
 
         res.status(200).send({
             integration_id: body.integration_id,
             function_name: body.function_name,
             function_type: body.function_type,
             execution_timeout_at: new Date(startedAt.getTime() + 5 * 60 * 1000).toISOString(),
             duration_ms: durationMs,
-            output: result.output
+            ...(dryrunOutput.hasResult ? { result: dryrunOutput.result } : {})
```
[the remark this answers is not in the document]
marcindobry: This is intentional, the output is just noise and we mapped the soft failures to errors

kaposke (approved):

PR #5899, opened 2026-04-17
feat: add and use `source` for tracking function source

kaposke:
Replacing `is_public` and `pre_built`, which are booleans used interchangeably (if you know the reason for `is_public` let me know, but as far as I've seen they always have the same value).

> [!NOTE]
> I'd like to split the migration from the logic but we need the values to be set correctly as soon as the column exists.
> For safety, I'm keeping the migration for removing `is_public` and `pre_built` for a second stage


packages/database/lib/migrations/20260417120000_sync_config_code_source.cjs
```diff
@@ -0,0 +1,34 @@
...
+            WHEN duplicate_object THEN
+                NULL;
+        END
+        $$`);
+    // Add column with default so the backfill step works without requiring existing rows to be touched first
+    await knex.raw(`ALTER TABLE _nango_sync_configs ADD COLUMN IF NOT EXISTS source sync_config_source NOT NULL DEFAULT 'repo'`);
+    // Backfill: rows deployed from the Nango catalog get 'catalog'; everything else keeps 'repo'
+    await knex.raw(`UPDATE _nango_sync_configs SET source = 'catalog' WHERE pre_built = true OR is_public = true`);
+    // Drop default — every insert path now sets source explicitly
+    await knex.raw(`ALTER TABLE _nango_sync_configs ALTER COLUMN source DROP DEFAULT`);
```
agusayerza: Can we leave destructive actions for a follow-up migration?

We will probably talk about it on tomorrow's incident review, but we should be a little bit more cautious about migrations that aren't easy to rollback

packages/server/lib/controllers/functions/deploy/postDeploy.ts
```diff
@@ -39,11 +39,11 @@ export const postRemoteFunctionDeploy = asyncWrapper<PostRemoteFunctionDeploy>(a
         name: body.function_name,
         isAction: body.function_type === 'action'
     });
-    if (existingSyncConfig && (existingSyncConfig.is_public || existingSyncConfig.pre_built)) {
+    if (existingSyncConfig && existingSyncConfig.source === 'catalog') {
```
agusayerza, line 42: is it worth it to extract to constants? Add auxiliary methods to check source?

packages/database/lib/migrations/20260417120000_sync_config_code_source.cjs
kaposke: This is just removing the default from the newly added field.
Sequence is:
- Add field with default, because we want it to be `NOT NULL`
- Backfill correct values
- Drop default value

Could be done with creating a nullable column, backfilling, then making it not nullable as well.

packages/server/lib/controllers/functions/deploy/postDeploy.ts
kaposke: Thought about this other approach:

```ts
const FunctionSources = {
  CATALOG: 'catalog',
  STANDALONE: 'standalone',
  REPO: 'repo',
} as const;

type FunctionSource = typeof FunctionSources[keyof typeof FunctionSources];

if (func.source === FunctionSources.CATALOG) { ... }
```

packages/database/lib/migrations/20260417120000_sync_config_code_source.cjs
agusayerza: I completely miss read the line, my bad

packages/server/lib/controllers/functions/deploy/postDeploy.ts
agusayerza: Yeah, I think having them as constant is better but it might be NIT

packages/database/lib/migrations/20260417120000_sync_config_code_source.cjs
agusayerza: Isn't there a risk for invalid inserts during rollout?

1. New server pod starts
2. It runs the migration (completes and drops the default)
3. Old server pod (not rolled out to yet) inserts a nango_sync_config
4. Source is null and insert fails

packages/lambda-runner/lib/schemas.ts
```diff
@@ -45,8 +45,7 @@ export const nangoPropsSchema = z.object({
         type: z.enum(['sync', 'action', 'on-event']),
         auto_start: z.boolean(),
         attributes: z.record(z.string(), z.any()),
-        pre_built: z.boolean(),
-        is_public: z.boolean(),
+        source: z.enum(['catalog', 'standalone', 'repo']),
```
agusayerza: Similar to the migration comment, I would be worried about this not being optional during deployment. [NangoProps are sent](https://github.com/NangoHQ/nango/blob/master/packages/lambda-runner/lib/index.ts#L187) back to jobs once the task is completed. There will be "legacy" actions/sync running during the deployment that will send a null source to and updated jobs pod, that [will revalidate](https://github.com/NangoHQ/nango/blob/master/packages/jobs/lib/routes/tasks/putTask.ts#L16) and fail.

In this case, I think the risk is very real

packages/jobs/lib/execution/onEvent.ts
```diff
@@ -90,10 +90,9 @@ export async function startOnEvent(task: TaskOnEvent): Promise<Result<void>> {
             webhook_subscriptions: [],
             attributes: {},
             input: null,
-            is_public: false,
+            source: 'repo',
```
agusayerza, line 93: Q: why do we need to hardcode 'repo' for on-events?  This is an ideal opportunity to start tracking their source as well, right? (In theory we could be providing on-event scripts as part of the catalog)

packages/database/lib/migrations/20260417120000_sync_config_code_source.cjs
marcindobry: Yup I think it's risky if not all code paths set it. We can drop it in a separate PR

packages/database/lib/migrations/20260417120000_sync_config_code_source.cjs
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * @param {import('knex').Knex} knex
+ */
+exports.up = async function (knex) {
+    // Create enum type
+    await knex.raw(`DO $$
+        BEGIN
+            CREATE TYPE sync_config_source AS ENUM (
+                'catalog',
+                'standalone',
```
marcindobry, line 12: Is that the final name? Tbh, I wouldn't know what it meant

packages/database/lib/migrations/20260417120000_sync_config_code_source.cjs
```diff
@@ -0,0 +1,34 @@
...
+            );
+        EXCEPTION
+            WHEN duplicate_object THEN
+                NULL;
+        END
+        $$`);
+    // Add column with default so the backfill step works without requiring existing rows to be touched first
+    await knex.raw(`ALTER TABLE _nango_sync_configs ADD COLUMN IF NOT EXISTS source sync_config_source NOT NULL DEFAULT 'repo'`);
+    // Backfill: rows deployed from the Nango catalog get 'catalog'; everything else keeps 'repo'
+    await knex.raw(`UPDATE _nango_sync_configs SET source = 'catalog' WHERE pre_built = true OR is_public = true`);
```
marcindobry, line 23: Should we also do this as a separate PR in case the weird pod rollout happens?

packages/database/lib/migrations/20260417120000_sync_config_code_source.cjs
kaposke: It's what came to my head for something unrelated to catalog or repo. Your "headless" suggestion is good too.

packages/jobs/lib/execution/onEvent.ts
kaposke: onEvent scripts currently don't currently have the prebuilt concept, and are only deployable through CLI. Definitely something we could do, I'd say it's OOS for now.

TBonnin:
> if you know the reason for is_public let me know, but as far as I've seen they always have the same value

A question for @khaliqgant. Maybe @rguldener remembers also. At some point we might have had pre-built templates that were not publicly visible

khaliqgant:
> > if you know the reason for is_public let me know, but as far as I've seen they always have the same value
> 
> A question for @khaliqgant. Maybe @rguldener remembers also. At some point we might have had pre-built templates that were not publicly visible

Yeah, at some point we prebuilt templates for particular customers that weren't public but were prebuilt just for that customer. With an admin key could push to a customers account.

packages/database/lib/migrations/20260417120000_sync_config_code_source.cjs
kaposke: The backfill should be fine, right? Just don't remove the other fields that might be depended on

packages/server/lib/controllers/functions/deploy/postDeploy.ts
marcindobry: Honestly, I prefer it the way it is now :P Typescript won't allow you to have anything other than the 3 string values anyways. You can always look up the type of `source` if you really want to see the options.

marcindobry (approved):

PR #5903, opened 2026-04-20
feat: classify `standalone` deploys

kaposke:
Assuming the "standalone" term I made up is accepted, classify functions deployed from the `/remote-function/deploy` endpoint as "standalone". Done by adding an optional `source` property to `postDeploy`'s body, and making so the CLI reads an optional `NANGO_DEPLOY_SOURCE` env var to send it as `standalone`. The sandbox is injected with that flag.

### DIFF
--- a/packages/server/lib/controllers/functions/deploy/postDeploy.ts
+++ b/packages/server/lib/controllers/functions/deploy/postDeploy.ts
@@ -39,7 +39,8 @@ export const postRemoteFunctionDeploy = asyncWrapper<PostRemoteFunctionDeploy>(a
         name: body.function_name,
         isAction: body.function_type === 'action'
     });
-    if (existingSyncConfig && existingSyncConfig.source === 'catalog') {
+
+    if (existingSyncConfig && existingSyncConfig.source !== 'standalone') {
         res.status(400).send({
             error: {
                 code: 'invalid_request',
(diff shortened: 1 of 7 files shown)


packages/server/lib/controllers/functions/deploy/postDeploy.ts
```diff
@@ -39,7 +39,8 @@ export const postRemoteFunctionDeploy = asyncWrapper<PostRemoteFunctionDeploy>(a
         name: body.function_name,
         isAction: body.function_type === 'action'
     });
-    if (existingSyncConfig && existingSyncConfig.source === 'catalog') {
+
+    if (existingSyncConfig && existingSyncConfig.source !== 'standalone') {
```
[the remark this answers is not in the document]
marcindobry: We don't care about them

marcindobry, line 43: Did we land on whether we should/shouldn't allow overwriting templates? Honestly I don't have any strong arguments in either way from an eng PoV, it seems like it's purely a product decision.

marcindobry (approved):

packages/server/lib/controllers/functions/deploy/postDeploy.ts
kaposke: won't be blocked by this for now

PR #5910, opened 2026-04-21
feat(feature-flags): add unleash openfeature client

agusayerza:
This PR adds `@nangohq/feature-flags`, a thin wrapper around the [OpenFeature](https://openfeature.dev/) server SDK with two providers:
- **`noop`** (default) — always returns the call-site default. Zero-config, safe to ship to any environment.
- **`unleash`** — talks to a self-hosted [Unleash](https://www.getunleash.io/) instance. Configured via env vars; only enabled where `NANGO_FLAG_PROVIDER=unleash` and `NANGO_UNLEASH_URL` are set (currently development only — see infra/env PRs).

NAN-5344

### DIFF
--- a/packages/feature-flags/lib/index.ts
+++ b/packages/feature-flags/lib/index.ts
@@ -1,0 +1,60 @@
+import { buildFeatureFlagsClient } from './client.js';
+import { envs } from './env.js';
+import { NoopProvider } from './providers/noop.js';
+import { UnleashProvider } from './providers/unleash.js';
+
+import type { FeatureFlagsClient } from './client.js';
+import type { Provider } from '@openfeature/server-sdk';
+
+export type { FeatureFlagsClient } from './client.js';
+export type { FlagContext } from './types.js';
+export { FLAGS, type FlagKey } from './registry.js';
+
+let clientPromise: Promise<FeatureFlagsClient> | undefined;
+let destroyPromise: Promise<void> | undefined;
+
+export async function getFeatureFlagsClient(): Promise<FeatureFlagsClient> {
+    if (destroyPromise) await destroyPromise;
+    if (clientPromise) return clientPromise;
+    clientPromise = createClient();
+    return clientPromise;
+}
+
+export async function destroy(): Promise<void> {
+    if (destroyPromise) return destroyPromise;
+    const promise = clientPromise;
+    if (!promise) return;
+    destroyPromise = (async () => {
+        try {
+            const client = await promise;
+            await client.destroy();
+        } finally {
+            clientPromise = undefined;
+            destroyPromise = undefined;
+        }
+    })();
+    return destroyPromise;
+}
+
+async function createClient(): Promise<FeatureFlagsClient> {
+    const provider = await buildProvider();
+    return buildFeatureFlagsClient(provider);
+}
+
+async function buildProvider(): Promise<Provider> {
+    if (envs.NANGO_FLAG_PROVIDER === 'unleash') {
+        if (!envs.NANGO_UNLEASH_URL) {
+            console.warn('NANGO_FLAG_PROVIDER=unleash but NANGO_UNLEASH_URL is unset; using noop provider');
+            return new NoopProvider();
+        }
+        const provider = new UnleashProvider({
+            url: envs.NANGO_UNLEASH_URL,
+            appName: envs.NANGO_UNLEASH_APP_NAME,
+            apiToken: envs.NANGO_UNLEASH_API_TOKEN,
+            refreshIntervalMs: envs.NANGO_UNLEASH_REFRESH_INTERVAL_MS
+        });
+        await provider.initialize();
+        return provider;
+    }
+    return new NoopProvider();
+}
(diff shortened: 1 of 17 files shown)


packages/feature-flags/lib/providers/unleash.ts
```diff
@@ -0,0 +1,125 @@
...
+        this.unleash.destroy();
+        return Promise.resolve();
+    }
+
+    resolveBooleanEvaluation(flagKey: string, defaultValue: boolean, context: EvaluationContext, _logger: Logger): Promise<ResolutionDetails<boolean>> {
+        try {
+            const value = this.unleash.isEnabled(flagKey, toUnleashContext(context), defaultValue);
+            return Promise.resolve({ value, reason: 'TARGETING_MATCH' });
+        } catch (err) {
+            return Promise.resolve({
```
marcindobry: Does this mean we're going to need to check for `reason = 'ERROR' or `errorCode` every time we get a flag value just in case?

Alternatives I can think of are 
- returning null/undefined in the value
- rejecting a Promise instead of resolving it.

Not sure what's the best solution, just curious what led you to make the current decision?

packages/feature-flags/lib/providers/unleash.ts
```diff
@@ -0,0 +1,125 @@
...
+    async initialize(): Promise<void> {
+        if (this.ready) return;
+        await new Promise<void>((resolve) => {
+            this.unleash.once('synchronized', () => resolve());
+            this.unleash.once('error', () => resolve());
+        });
+    }
+
+    onClose(): Promise<void> {
+        this.unleash.destroy();
```
TBonnin, line 150: is destroy synchronous?

packages/feature-flags/lib/providers/unleash.ts
```diff
@@ -0,0 +1,125 @@
...
+        this.unleash.on('error', (err: Error) => {
+            console.error(`Unleash error: ${err.message}`);
+        });
+    }
+
+    async initialize(): Promise<void> {
+        if (this.ready) return;
+        await new Promise<void>((resolve) => {
+            this.unleash.once('synchronized', () => resolve());
+            this.unleash.once('error', () => resolve());
```
TBonnin: I imagine synchronized and error cannot be triggered both together but just in case we could clean up listeners like 
```
   const done = () => { this.unleash.removeListener('synchronized', done); this.unleash.removeListener('error', done); resolve(); };
   this.unleash.once('synchronized', done);
   this.unleash.once('error', done);
```

packages/feature-flags/lib/providers/unleash.ts
TBonnin: my understanding is Provider functions are not called directly but used by the openfeature sdk which deal with either a rejected promise or an ERROR. Might be worth linking to the following docs though

https://openfeature.dev/docs/reference/concepts/provider/
> How should I handle error conditions?
> If an error condition is encountered in your provider, an error should be indicated by throwing or returning an error, as language idioms dictate. The SDK will ensure the default value is returned and expose the relevant error data.

packages/feature-flags/lib/index.ts
```diff
@@ -0,0 +1,60 @@
...
+
+async function createClient(): Promise<FeatureFlagsClient> {
+    const provider = await buildProvider();
+    return buildFeatureFlagsClient(provider);
+}
+
+async function buildProvider(): Promise<Provider> {
+    if (envs.NANGO_FLAG_PROVIDER === 'unleash') {
+        if (!envs.NANGO_UNLEASH_URL) {
+            console.warn('NANGO_FLAG_PROVIDER=unleash but NANGO_UNLEASH_URL is unset; using noop provider');
```
TBonnin: we should use `logger`, no?

packages/feature-flags/package.json
```diff
@@ -0,0 +1,28 @@
+{
+    "name": "@nangohq/feature-flags",
+    "version": "1.0.0",
+    "type": "module",
+    "main": "./dist/index.js",
+    "types": "./dist/index.js",
+    "private": true,
+    "scripts": {
+        "build": "rimraf ./dist && tsc"
```
TBonnin, line 9: rimraf is missing from the `devDependencies`

TBonnin (approved):

marcindobry (approved):

packages/server/lib/server.ts
```diff
@@ -92,6 +93,18 @@ if (NANGO_MIGRATE_AT_START === 'true') {
 // Preload providers
 getProviders();
 
+// [TEST] Exercise the feature-flags → Unleash chain on startup so the provider
+// initializes (and logs) without manual intervention. Remove before merge.
+void (async () => {
```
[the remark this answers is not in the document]
agusayerza: Reverted

packages/server/lib/server.ts
```diff
@@ -92,6 +93,18 @@ if (NANGO_MIGRATE_AT_START === 'true') {
 // Preload providers
 getProviders();
 
+// [TEST] Exercise the feature-flags → Unleash chain on startup so the provider
```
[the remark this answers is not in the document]
agusayerza: Reverted

PR #5922, opened 2026-04-23
refactor(nan-5088): remove dual lookup, unify auth paths

pfreixes:
## Why
- We are not having a single public API request which goes through the fallback, we did this solely for catching up data races during the migration. See [DD](https://us3.datadoghq.com/s/e94cfb60-2611-11ee-93ab-da7ad0900003/bie-tcr-qui) panel for this
- We can now remove the dual lookup

## Summary
- **Remove dual lookup**: `getAccountContextByCustomerKey` now only queries `customer_keys` (removed UNION ALL fallback to `api_secrets`). Customer API auth resolves exclusively through `customer_keys`.
- **Unify `validateApiKey`**: Single method in `access.middleware.ts` replaces the separate `isScript` branch and `validateSecretKey`, routing to `getAccountContextByApiKey({ secretKey | internalSecretKey })`.
- **Inject auth on internal keys**: `getAccountContextByInternalSecret` now returns `auth: { source: 'api_secret', scopes: ['environment:*'] }` — scopes are injected at the data layer rather than relying on middleware to check `isScript`.
- **Add `env_var` auth source**: Env-var-based keys return `auth: { source: 'env_var', scopes: ['environment:*'] }`.
- **Update tests**: Fallback tests updated to expect `null` (no dual lookup). Added internal script key tests using `Nango-Is-Script` header.

## Test plan
- [x] All 537 server integration tests pass (70 files)
- [x] All 22 persist integration tests pass
- [x] All 14 account service integration tests pass (including updated fallback tests)
- [x] Scope enforcement tests (47 tests) pass — including internal key access tests

### DIFF
--- a/packages/server/lib/middleware/access.middleware.ts
+++ b/packages/server/lib/middleware/access.middleware.ts
@@ -51,7 +51,7 @@ export class AccessMiddleware {
             secret: DBAPISecret;
             plan: DBPlan | null;
             auth?: {
-                source: string;
+                source: 'customer_key' | 'api_secret' | 'env_var';
                 scopes?: string[];
                 apiKeyId?: number;
             };
--- a/packages/server/lib/middleware/access.middleware.ts
+++ b/packages/server/lib/middleware/access.middleware.ts
@@ -69,7 +69,7 @@ export class AccessMiddleware {
             return Err('plan_not_found');
         }
 
-        return Ok(accountContext);
+        return Ok({ ...accountContext });
     }
 
     async secretKeyAuth(req: Request, res: Response<any, RequestLocals>, next: NextFunction) {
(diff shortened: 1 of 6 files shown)


marcindobry (approved):

packages/server/lib/middleware/access.middleware.ts
```diff
@@ -41,14 +41,17 @@ const ignoreEnvPaths = [
...
+        opts: { isScript: boolean }
+    ): Promise<
         Result<{
             account: DBTeam;
             environment: DBEnvironment;
             secret: DBAPISecret;
             plan: DBPlan | null;
             auth?: {
-                source: 'customer_key' | 'api_secret';
+                source: string;
```
rossmcewan: Should this not be `'customer_key' | 'api_secret' | 'env_var'`

rossmcewan (approved):

packages/server/lib/middleware/access.middleware.ts
pfreixes: yes!

PR #5924, opened 2026-04-23
feat(integrations): build correct authorization headers for aws-iam

hassan254-prog:

- Builds correct authorization headers for AWS IAM and migrates the documentation. Previously, this was not handled at the platform level when support was added, and users had to [implement](https://nango.dev/docs/integrations/all/aws-iam#api-gotchas) it themselves within their functions.


packages/shared/lib/services/proxy/utils.ts
```diff
@@ -321,6 +321,14 @@ export function buildProxyURL({ config, connection }: { config: ApplicationConst
     return url.toString();
 }
 
+function getRawBody(method: string, data: unknown): string {
+    if (!['POST', 'PUT', 'PATCH'].includes(method) || !data) return '';
+    if (typeof data === 'string') return data.startsWith('?') ? data.slice(1) : data;
+    if (Buffer.isBuffer(data)) return data.toString('utf8');
+    if (data instanceof URLSearchParams) return data.toString();
+    return '';
```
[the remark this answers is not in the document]
hassan254-prog: The AWS IAM only supports GET with query string params or POST with application/x-www-form-urlencoded, JSON bodies are not valid per the [AWS IAM API docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/programming.html).

docs/api-integrations/aws-iam.mdx
```diff
@@ -0,0 +1,82 @@
...
+        </Tab>
+
+        <Tab title="Node">
+
+        Install Nango's backend SDK with `npm i @nangohq/node`. Then run:
+
+        ```typescript
+        import { Nango } from '@nangohq/node';
+
+        const nango = new Nango({ secretKey: '<NANGO-SECRET-KEY>' });
```
pfreixes, line 39: Ideally we would like to rename `<NANGO-SECRET-KEY>` for `<NANGO-API-KEY>`, we already did for [some](https://github.com/search?q=repo%3ANangoHQ%2Fnango%20NANGO-API-KEY&type=code)

packages/shared/lib/services/proxy/utils.ts
```diff
@@ -321,6 +321,14 @@ export function buildProxyURL({ config, connection }: { config: ApplicationConst
...
+}
+
 // builds the canonical parameter string as required by the Duo API request signing spec.
```
rossmcewan, line 332: Why do these comments reference the Duo API?

packages/providers/providers.yaml
```diff
@@ -1984,14 +1984,17 @@ aws-iam:
     auth_mode: BASIC
     proxy:
         base_url: https://iam.amazonaws.com
+        headers:
+            authorization: '${awsSigV4(${credentials.username}, ${credentials.password}, ${connectionConfig.region}, iam)}'
```
pfreixes: I had some back and forth with codex when reviewing this PR, he seems determined to believe that for general services like IAM for the global endpoint `https://iam.amazonaws.com` we must use the `us-east-1` region [1].

Have we tried this, that for example list users works when we use a region that is not `us-east-1`?

PD: I asked codex to check what botocore does, and it show me how botocore behind the scenes uses the hardcoded `us-east-1` region

[1] https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_examples_aws_deny-requested-region.html

rossmcewan (approved):

pfreixes (commented):
Maybe something to get reviewed about the region used for interacting with IAM

rossmcewan:
> Maybe something to get reviewed about the region used for interacting with IAM

AWS maps all regions - except FIPS - to the global endpoint: https://docs.aws.amazon.com/general/latest/gr/iam-service.html

pfreixes (approved):

PR #5954, opened 2026-04-27
feat(webhooks): thread `ingressRequestId` through webhook handler chain (NAN-5338)

agusayerza:
Generates a server-side UUID at the top of postWebhook and carries it through `routeWebhook` into `InternalNango` as a constructor field. No behavioral change yet, the ID is consumed by the upcoming dispatch-queue publisher (`NAN-5340`) to build deterministic taskNames when no de-dupe key is available. It is also used to dedupe SQS redelivery.

### DIFF
--- a/packages/server/lib/controllers/webhook/environmentUuid/postWebhook.ts
+++ b/packages/server/lib/controllers/webhook/environmentUuid/postWebhook.ts
@@ -1,3 +1,5 @@
+import { randomUUID } from 'node:crypto';
+
 import tracer from 'dd-trace';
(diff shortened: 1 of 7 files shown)


TBonnin (commented):
the `logContextGetter` would get you a random id as well.

agusayerza:
> the `logContextGetter` would get you a random id as well.

Good point. We can use `activityLogId` instead. That makes this prep-only plumbing PR unnecessary, so I’m going to fold it into the dispatch-queue PR instead of landing it separately

PR #5955, opened 2026-04-27
feat(webhooks): add dispatch queue publisher (NAN-5339) 1/5

agusayerza:
- WebhookDispatchMessage shared type in `@nangohq/types`
- DispatchQueuePublisher in server package: batched SendMessageBatch, parallel batches, one inline retry for failed entries, partial-failure emits metric + log
- New WEBHOOK_DISPATCH_* metric types in `@nangohq/utils`
- Adds `@aws-sdk/client-sqs` depedency to server

No call sites yet, wired up in NAN-5340.

### DIFF
--- a/packages/server/lib/webhook/dispatch-queue/publisher.ts
+++ b/packages/server/lib/webhook/dispatch-queue/publisher.ts
@@ -1,0 +1,173 @@
+import { SendMessageBatchCommand } from '@aws-sdk/client-sqs';
+import tracer from 'dd-trace';
+
+import { metrics } from '@nangohq/utils';
+
+import { runWithConcurrencyLimit } from '../runWithConcurrencyLimit.js';
+
+import type { SQSClient, SendMessageBatchCommandOutput, SendMessageBatchRequestEntry } from '@aws-sdk/client-sqs';
+import type { WebhookDispatchMessage } from '@nangohq/types';
+
+const SQS_BATCH_MAX_ENTRIES = 10;
+const DEFAULT_PUBLISH_CONCURRENCY = 10;
+
+export interface DispatchQueuePublisherProps {
+    sqs: SQSClient;
+    queueUrl: string;
+    /** Max entries per SQS SendMessageBatch request. AWS caps this at 10. */
+    batchSize?: number;
+    /** Max concurrent SQS SendMessageBatch requests. Defaults to 10. */
+    publishConcurrency?: number;
+}
+
+export interface PublishResult {
+    enqueued: number;
+    failed: number;
+}
+
+interface BatchPublishResult extends PublishResult {
+    retriedEntries: number;
+}
+
+export class DispatchQueuePublisher {
+    private readonly sqs: SQSClient;
+    private readonly queueUrl: string;
+    private readonly batchSize: number;
+    private readonly publishConcurrency: number;
+
+    constructor(props: DispatchQueuePublisherProps) {
+        this.sqs = props.sqs;
+        this.queueUrl = props.queueUrl;
+        const configuredBatchSize = props.batchSize ?? SQS_BATCH_MAX_ENTRIES;
+        if (configuredBatchSize < 1) {
+            throw new RangeError(`batchSize must be > 0`);
+        }
+        this.batchSize = Math.min(configuredBatchSize, SQS_BATCH_MAX_ENTRIES);
+
+        const configuredPublishConcurrency = props.publishConcurrency ?? DEFAULT_PUBLISH_CONCURRENCY;
+        if (configuredPublishConcurrency < 1) {
+            throw new RangeError(`publishConcurrency must be > 0`);
+        }
+        this.publishConcurrency = configuredPublishConcurrency;
+    }
+
+    /**
+     * Publish a list of dispatch messages in batches. Batches are fired in parallel up to
+     * `publishConcurrency`; failed entries within a batch are retried once inline. Any
+     * entries still failing after the retry are counted as `failed`. Regular SQS send
+     * failures and partial failures are returned in the counts instead of throwing so the
+     * caller can treat them as a metric/trace concern, not an HTTP 500 (provider retries
+     * are worse).
+     */
+    async publish(messages: WebhookDispatchMessage[], messageGroupId: string): Promise<PublishResult> {
+        if (messages.length === 0) {
+            return { enqueued: 0, failed: 0 };
+        }
+
+        const activeSpan = tracer.scope().active();
+        const firstMessage = messages[0]!;
+        const batches = chunk(messages, this.batchSize);
+        const span = tracer.startSpan('webhook.dispatch.publish', {
+            ...(activeSpan ? { childOf: activeSpan } : {}),
+            tags: {
+                'nango.accountId': firstMessage.accountId,
+                'nango.environmentId': firstMessage.connection.environment_id,
+                'nango.integrationId': firstMessage.integrationId,
+                'nango.provider': firstMessage.provider,
+                'nango.providerConfigKey': firstMessage.connection.provider_config_key,
+                'nango.messageCount': messages.length,
+                'nango.batchCount': batches.length
+            }
+        });
+
+        return await tracer.scope().activate(span, async () => {
+            try {
+                const results = await runWithConcurrencyLimit(batches, this.publishConcurrency, async (batch) => {
+                    return await this.sendBatch(batch, messageGroupId);
+                });
+
+                const enqueued = results.reduce((sum, r) => sum + r.enqueued, 0);
+                const failed = results.reduce((sum, r) => sum + r.failed, 0);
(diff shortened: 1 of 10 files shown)


packages/server/lib/webhook/dispatch-queue/publisher.ts
```diff
@@ -0,0 +1,196 @@
...
+            tags: {
+                'nango.accountId': firstMessage.accountId,
+                'nango.environmentId': firstMessage.environmentId,
+                'nango.integrationId': firstMessage.integrationId,
+                'nango.provider': firstMessage.provider,
+                'nango.providerConfigKey': firstMessage.providerConfigKey,
+                'nango.messageCount': messages.length,
+                'nango.batchCount': batches.length,
+                'nango.publishConcurrency': this.publishConcurrency,
+                'nango.batchSize': this.batchSize
```
agusayerza: Will remove this last two dimensions

packages/server/lib/webhook/dispatch-queue/publisher.ts
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    private async trySend(entries: SendMessageBatchRequestEntry[]): Promise<{ failedIds: string[] }> {
+        if (entries.length === 0) return { failedIds: [] };
+        try {
+            const response: SendMessageBatchCommandOutput = await this.sqs.send(new SendMessageBatchCommand({ QueueUrl: this.queueUrl, Entries: entries }));
+            const failedIds = (response.Failed ?? []).map((f) => f.Id).filter((id): id is string => typeof id === 'string');
+            return { failedIds };
+        } catch (err) {
+            logger.error(`SendMessageBatchCommand threw`, { error: err });
```
agusayerza: Intetional log here as we don't have per-batch tracing

packages/types/lib/webhooks/dispatch.ts
```diff
@@ -0,0 +1,38 @@
...
+     * dedupe across separate ingress requests.
+     */
+    ingressRequestId: string;
+    accountId: number;
+    environmentId: number;
+    integrationId: number;
+    provider: string;
+    providerConfigKey: string;
+    parentSyncName: string;
+    connection: {
```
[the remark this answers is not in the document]
agusayerza: Done

packages/server/lib/webhook/dispatch-queue/publisher.ts
```diff
@@ -0,0 +1,191 @@
...
+            return { failedIds: entries.map((e) => e.Id!).filter((id): id is string => typeof id === 'string') };
+        }
+    }
+}
+
+function toEntry(message: WebhookDispatchMessage, index: number, messageGroupId: string): SendMessageBatchRequestEntry {
+    return {
+        Id: indexToEntryId(index),
+        MessageBody: JSON.stringify(message),
+        MessageGroupId: messageGroupId
```
TBonnin, line 155: no `MessageDeduplicationId`

packages/server/lib/webhook/dispatch-queue/publisher.ts
```diff
@@ -0,0 +1,191 @@
...
+        if (configuredPublishConcurrency < 1) {
+            throw new RangeError(`publishConcurrency must be > 0`);
+        }
+        this.publishConcurrency = configuredPublishConcurrency;
+    }
+
+    /**
+     * Publish a list of dispatch messages in batches. Batches are fired in parallel up to
+     * `publishConcurrency`; failed entries within a batch are retried once inline. Any
+     * entries still failing after the retry are counted as `failed`. Never throws — the
```
TBonnin: `Never throws` is contradicted by L112

packages/server/lib/webhook/dispatch-queue/publisher.ts
```diff
@@ -0,0 +1,191 @@
...
+                span.setTag('nango.enqueued', enqueued);
+                span.setTag('nango.failed', failed);
+                span.setTag('nango.retriedEntries', retriedEntries);
+                span.setTag('nango.retriedBatches', retriedBatches);
+
+                const provider = firstMessage.provider;
+
+                if (enqueued > 0) {
+                    metrics.increment(metrics.Types.WEBHOOK_DISPATCH_MESSAGES_ENQUEUED, enqueued, { provider });
+                    metrics.increment(metrics.Types.WEBHOOK_DISPATCH_PUBLISH_SUCCESS, enqueued, { provider });
```
TBonnin, line 102: Do we need both? We are counting the same things twice

packages/server/lib/webhook/dispatch-queue/publisher.unit.test.ts
```diff
@@ -0,0 +1,261 @@
...
+        tracerMocks.activate.mockClear();
+        tracerMocks.startSpan.mockClear();
+        tracerMocks.span.setTag.mockClear();
+        tracerMocks.span.finish.mockClear();
+        tracerMocks.activeSpan.setTag.mockClear();
+        tracerMocks.dogstatsd.increment.mockClear();
+        tracerMocks.dogstatsd.decrement.mockClear();
+        tracerMocks.dogstatsd.gauge.mockClear();
+        tracerMocks.dogstatsd.histogram.mockClear();
+        tracerMocks.dogstatsd.distribution.mockClear();
```
TBonnin, line 108: isn't `restoreAllMocks` already cleaning up the spies?

agusayerza: Most of these are hoisted `vi.fn()` mocks, not `vi.spyOns`, and `restoreAllMocks()` does not clear or reset the history of `vi.fn()`

packages/server/lib/webhook/dispatch-queue/publisher.ts
agusayerza: We are using a standard SQS queue with fair-queue behavior via `MessageGroupId`, not a FIFO queue. Unfortunately the caveat is that fair-queues do not support `MessageDeduplicationId`. We will be using the task name to dedupe on the orchestrator. I'm open to using a FIFO queue with `MessageDeduplicationId` but we lose the fairness across tenants. Even then orchestrator would still need deduping transport retries. (FIFO guarantees at-least-once delivery)

packages/server/lib/webhook/dispatch-queue/publisher.ts
agusayerza: Yeah, fixing it. I meant never throws on regular SQS operations.

packages/server/lib/webhook/dispatch-queue/publisher.ts
agusayerza: Agreed, `enqueued` ended up being redundant here and also reads more like queue depth, so I removed it and kept `publish.success` / `publish.failure`

packages/server/lib/webhook/dispatch-queue/publisher.ts
```diff
@@ -0,0 +1,172 @@
...
+    constructor(props: DispatchQueuePublisherProps) {
+        this.sqs = props.sqs;
+        this.queueUrl = props.queueUrl;
+        const configuredBatchSize = props.batchSize ?? SQS_BATCH_MAX_ENTRIES;
+        if (configuredBatchSize < 1) {
+            throw new RangeError(`batchSize must be > 0`);
+        }
+        this.batchSize = Math.min(configuredBatchSize, SQS_BATCH_MAX_ENTRIES);
+
+        const configuredPublishConcurrency = props.publishConcurrency ?? SQS_BATCH_MAX_ENTRIES;
```
marcindobry: Why are we using a constant for batching as a default for concurrency?

marcindobry (approved):

packages/utils/lib/telemetry/metrics.ts
```diff
@@ -67,6 +67,16 @@ export enum Types {
     WEBHOOK_ASYNC_ACTION_FAILED = 'nango.webhook.async_action.failed',
     WEBHOOK_INCOMING_PAYLOAD_SIZE_BYTES = 'nango.webhook.incoming.payloadSizeBytes',
 
+    WEBHOOK_DISPATCH_PUBLISH_SUCCESS = 'nango.webhook.dispatch_queue.publish.success',
+    WEBHOOK_DISPATCH_PUBLISH_FAILURE = 'nango.webhook.dispatch_queue.publish.failure',
```
TBonnin, line 72: I imagine the following metrics are declared for future use?

TBonnin (approved):

packages/server/lib/webhook/dispatch-queue/publisher.ts
agusayerza: Missed this, originally I had set it to the same value and forgot to split into new const

packages/utils/lib/telemetry/metrics.ts
agusayerza: Correct

PR #5959, opened 2026-04-27
feat(webhooks): add jobs dispatch queue consumer (NAN-5341) 2/5

agusayerza:
Consumer (`packages/jobs/lib/webhook/dispatch-queue/consumer.ts`)
- Long-polls the task-dispatch SQS queue and schedules webhook tasks via `orchestratorClient.executeWebhook()` using the deterministic `taskName` from the message.
- Deletes messages on successful scheduling and on duplicate-task detection. Leaves them visible on non-duplicate errors so SQS redelivers and eventually DLQs.
- Poison-pill messages (unparseable / schema-invalid) are deleted to avoid redelivery loops, with a dedicated metric.
- Starts one poll loop per `NANGO_TASK_DISPATCH_CONSUMER_CONCURRENCY`.

Scheduler (`packages/scheduler/lib/models/tasks.ts, packages/scheduler/lib/scheduler.ts`)
- Adds `DuplicateTaskNameError` so `scheduler.immediate()` can return a typed duplicate-name failure after checking for an existing task by name before insert.
- Preserves the existing `tasks.create()` behavior for other callers while still surfacing DB unique-constraint collisions as duplicate-name errors.

Orchestrator (`packages/orchestrator/lib/routes/v1/postImmediate.ts`)
- Returns structured duplicate metadata on `POST /v1/immediate` failures with `reason: 'duplicate_task_name'`, plus the conflicting `taskName` and existing `taskId` when available.

Tests
- Adds/updates unit and integration coverage for duplicate-name handling, consumer shutdown/delete behavior, and consumer concurrency.

### DIFF
--- a/packages/jobs/lib/app.ts
+++ b/packages/jobs/lib/app.ts
@@ -7,12 +7,14 @@ import { destroy as destroyLogs, otlp } from '@nangohq/logs';
 import { getOtlpRoutes } from '@nangohq/shared';
 import { getLogger, initSentry, once, report, stringifyError } from '@nangohq/utils';
 
+import { orchestratorClient } from './clients.js';
 import { envs } from './env.js';
 import { LambdaInvocationsProcessor } from './invocations/lambda.processor.js';
 import { Processor } from './processor/processor.js';
 import { getDefaultFleet, startFleets, stopFleets } from './runtime/runtimes.js';
 import { server } from './server.js';
 import { pubsub } from './utils/pubsub.js';
+import { DispatchQueueConsumer } from './webhook/dispatch-queue/consumer.js';
 
 const logger = getLogger('Jobs');
 
@@ -37,6 +39,18 @@ try {
     logger.info(`🚀 service ready at http://localhost:${port}`);
     const processor = new Processor(orchestratorUrl);
     const invocationsProcessor = new LambdaInvocationsProcessor();
+
+    const webhookDispatchConsumer = envs.NANGO_TASK_DISPATCH_QUEUE_URL
+        ? new DispatchQueueConsumer({
+              queueUrl: envs.NANGO_TASK_DISPATCH_QUEUE_URL,
+              orchestratorClient,
+              webhookMaxConcurrency: envs.WEBHOOK_ENVIRONMENT_MAX_CONCURRENCY,
+              consumerConcurrency: envs.NANGO_TASK_DISPATCH_CONSUMER_CONCURRENCY,
+              maxMessages: envs.NANGO_TASK_DISPATCH_MAX_MESSAGES,
+              waitTimeSeconds: envs.NANGO_TASK_DISPATCH_WAIT_TIME_SECONDS,
+              visibilityTimeoutSeconds: envs.NANGO_TASK_DISPATCH_VISIBILITY_TIMEOUT_SECONDS
+          })
+        : undefined;
 
     // We are using a setTimeout because we don't want overlapping setInterval if the DB is down
     let healthCheck: NodeJS.Timeout | undefined;
@@ -79,6 +93,9 @@ try {
             await db.readOnly.destroy();
             await destroyKvstore();
             await invocationsProcessor.stop();
+            if (webhookDispatchConsumer) {
+                await webhookDispatchConsumer.stop();
+            }
 
             console.info('Closed');
 
@@ -108,6 +125,11 @@ try {
     processor.start();
 
     invocationsProcessor.start();
+
+    if (webhookDispatchConsumer) {
+        webhookDispatchConsumer.start();
+        logger.info('webhook dispatch queue consumer started');
+    }
 
     void otlp.register(getOtlpRoutes);
 } catch (err) {
--- a/packages/jobs/lib/webhook/dispatch-queue/consumer.ts
+++ b/packages/jobs/lib/webhook/dispatch-queue/consumer.ts
@@ -1,0 +1,228 @@
+import { DeleteMessageCommand, ReceiveMessageCommand, SQSClient } from '@aws-sdk/client-sqs';
+import tracer from 'dd-trace';
+import * as z from 'zod';
+
+import { isDuplicateTaskNameClientError, jsonSchema } from '@nangohq/nango-orchestrator';
+import { Err, Ok, getLogger, metrics, report } from '@nangohq/utils';
+
+import { envs } from '../../env.js';
+
+import type { Message } from '@aws-sdk/client-sqs';
+import type { OrchestratorClient } from '@nangohq/nango-orchestrator';
+import type { WebhookDispatchMessage } from '@nangohq/types';
+import type { Result } from '@nangohq/utils';
+
+const logger = getLogger('jobs.webhook.dispatch-queue.consumer');
+
+const messageSchema: z.ZodType<WebhookDispatchMessage> = z.object({
+    version: z.literal(1),
+    kind: z.literal('webhook'),
+    taskName: z.string().min(1),
+    createdAt: z.string().min(1),
+    accountId: z.number(),
+    integrationId: z.number(),
+    provider: z.string(),
+    parentSyncName: z.string(),
+    activityLogId: z.string(),
+    webhookName: z.string(),
+    connection: z.object({
+        id: z.number(),
+        connection_id: z.string(),
+        provider_config_key: z.string(),
+        environment_id: z.number()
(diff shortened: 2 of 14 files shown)


packages/jobs/lib/webhook/dispatch-queue/consumer.ts
```diff
@@ -0,0 +1,233 @@
...
+
+import type { Message } from '@aws-sdk/client-sqs';
+import type { OrchestratorClient } from '@nangohq/nango-orchestrator';
+import type { WebhookDispatchMessage } from '@nangohq/types';
+
+const logger = getLogger('jobs.webhook.dispatch-queue.consumer');
+
+type ParseMessageResult = { success: true; message: WebhookDispatchMessage } | { success: false; reason: 'invalid_schema' | 'json_parse' };
+
+const getRegion = (): string => envs.AWS_REGION ?? 'us-west-2';
```
marcindobry: nit: Why does it have to be a function 🤔

packages/jobs/lib/webhook/dispatch-queue/consumer.ts
```diff
@@ -0,0 +1,233 @@
...
+                    span.setTag('error', true);
+                    span.setTag('error.type', scheduleRes.error.name);
+                    span.setTag('error.message', scheduleRes.error.message);
+
+                    const responsePayload = getClientErrorResponsePayload(scheduleRes.error);
+                    if (responsePayload) {
+                        span.setTag('error.details', responsePayload);
+                    }
+
+                    // Don't delete — let SQS redeliver and eventually DLQ.
```
marcindobry, line 181: How do we define max amount of retries? Is it some SQS queue parameter?

marcindobry (approved):

packages/jobs/lib/webhook/dispatch-queue/consumer.ts
```diff
@@ -0,0 +1,233 @@
...
+
+import { envs } from '../../env.js';
+
+import type { Message } from '@aws-sdk/client-sqs';
+import type { OrchestratorClient } from '@nangohq/nango-orchestrator';
+import type { WebhookDispatchMessage } from '@nangohq/types';
+
+const logger = getLogger('jobs.webhook.dispatch-queue.consumer');
+
+type ParseMessageResult = { success: true; message: WebhookDispatchMessage } | { success: false; reason: 'invalid_schema' | 'json_parse' };
```
TBonnin: Can we use `Result` instead of a defining a new type

packages/jobs/lib/webhook/dispatch-queue/consumer.ts
```diff
@@ -0,0 +1,233 @@
...
+                    }),
+                    { abortSignal: signal }
+                );
+
+                const messages = result.Messages ?? [];
+                if (messages.length === 0) continue;
+
+                // Process all messages in a batch concurrently. Each SQS receive is already
+                // capped at maxMessages (<=10) so no extra semaphore is needed.
+                await Promise.all(messages.map((msg) => this.processMessage(msg)));
```
TBonnin, line 109: we loose ordering here no? I am not sure it does matter for webhooks but we should make it part of the contract with customers

packages/jobs/lib/webhook/dispatch-queue/consumer.ts
```diff
@@ -0,0 +1,233 @@
...
+                        webhookName: message.webhookName,
+                        parentSyncName: message.parentSyncName,
+                        connection: message.connection,
+                        activityLogId: message.activityLogId,
+                        input: message.payload
+                    }
+                });
+
+                if (scheduleRes.isErr()) {
+                    if (isDuplicateTaskNameClientError(scheduleRes.error, message.taskName)) {
```
TBonnin: why is the task name required in the signature?

packages/jobs/lib/app.ts
```diff
@@ -38,6 +40,18 @@ try {
     const processor = new Processor(orchestratorUrl);
     const invocationsProcessor = new LambdaInvocationsProcessor();
 
+    const dispatchConsumer = envs.NANGO_TASK_DISPATCH_QUEUE_URL
```
TBonnin: should we call it something like `webhookDispatchConsumer`?

packages/jobs/lib/webhook/dispatch-queue/consumer.ts
```diff
@@ -0,0 +1,233 @@
...
+                    ...webhookTaskSchedulingSettings,
+                    args: {
+                        type: 'webhook',
+                        webhookName: message.webhookName,
+                        parentSyncName: message.parentSyncName,
+                        connection: message.connection,
+                        activityLogId: message.activityLogId,
+                        input: message.payload
+                    }
+                });
```
TBonnin, line 161: why call `immediate` instead of the `webhook` function?

packages/jobs/lib/webhook/dispatch-queue/consumer.ts
agusayerza: Its part of the SQS definition

packages/jobs/lib/webhook/dispatch-queue/consumer.ts
agusayerza: Setting it inline now, it used to have a much more complex behaviour

packages/jobs/lib/webhook/dispatch-queue/consumer.ts
agusayerza: Correct. Ordering is not guaranteed, this was a given as soon as we switched to queueing as dequeues and retrys can mess the order as well. Agreed that we should surface it on the contract with customers.

packages/jobs/lib/webhook/dispatch-queue/consumer.ts
agusayerza: Removed as discussed

packages/jobs/lib/webhook/dispatch-queue/consumer.ts
agusayerza: Switched to `executeWebhook` function

packages/orchestrator/lib/clients/client.ts
```diff
@@ -65,6 +65,16 @@ export class OrchestratorClient {
     public async immediate(props: ImmediateProps): Promise<Result<PostImmediate['Success'], ClientError>> {
         const res = await this.routeFetch(postImmediateRoute)({ body: props });
         if ('error' in res) {
+            const duplicatePayload = getDuplicateTaskNamePayload(res.error.payload);
+            if (duplicatePayload !== null) {
+                const taskName = duplicatePayload.taskName;
+                return Err({
+                    name: 'duplicate_task_name',
+                    message: duplicatePayload.message || (taskName ? `Task with name '${taskName}' already exists` : 'Task with this name already exists'),
+                    payload: taskName ? { taskName } : {}
```
TBonnin: can we simply forward the error from the orchestrator? why do we need a special case here?

packages/orchestrator/lib/routes/v1/postImmediate.ts
```diff
@@ -34,7 +39,7 @@ export type PostImmediate = Endpoint<{
         };
         args: JsonValue & { type: TaskType };
     };
-    Error: ApiError<'immediate_failed'>;
+    Error: ApiError<'immediate_failed' | 'duplicate_task_name', undefined, DuplicateTaskNamePayload>;
```
TBonnin: related to my question above. Do we need DuplicateTaskNamePayload?

packages/orchestrator/lib/clients/client.ts
agusayerza: We need to handle this as `routeFetch()` [(Ref)](https://github.com/NangoHQ/nango/blob/master/packages/utils/lib/express/route.ts#L94) wraps all non-2xx responses with `fetch_failed` and nests the real API error under payload. We special-case `duplicate_task_name` here so downstream callers (the consumer in this case), can handle it. If we want to remove this special case, we would need to change `routeFetch()` error propagation more broadly, which feels out of scope for this PR

packages/orchestrator/lib/routes/v1/postImmediate.ts
agusayerza: Removing the taskName on the error body as that is not used

TBonnin (approved):

PR #5970, opened 2026-04-28
feat(webhooks): switch internal-nango to publish via dispatch queue (NAN-5340) 3/5

agusayerza:
Behind `WEBHOOK_INGRESS_USE_DISPATCH_QUEUE` (default off), routes `InternalNango.executeScriptForWebhooks()` through the SQS dispatch queue instead of calling `orchestrator.triggerWebhook()` directly. When the flag is off, `triggerWebhook()` remains the fallback path.

When enabled, each (syncConfig × subscription × connection) triple becomes one `WebhookDispatchMessage`:
- `MessageGroupId` = `account:<accountId>:env:<environmentId>` for fair queueing
- Messages are batched into `SendMessageBatch` calls (up to 10 entries or 1 MiB per batch); large fan-out issues multiple batches in parallel

`connectionIds` are still returned so `forwardWebhook` keeps working.

Large fan-out (>200 messages) emits `WEBHOOK_DISPATCH_LARGE_FANOUT` for tracking.

A singleton `DispatchQueuePublisher` is cached in `dispatch-queue/client.ts` and gated on `NANGO_TASK_DISPATCH_QUEUE_URL` being set.

### DIFF
--- a/packages/server/lib/webhook/dispatch-queue/publisher.ts
+++ b/packages/server/lib/webhook/dispatch-queue/publisher.ts
@@ -1,7 +1,7 @@
 import { SendMessageBatchCommand } from '@aws-sdk/client-sqs';
 import tracer from 'dd-trace';
 
-import { metrics } from '@nangohq/utils';
+import { metrics, report } from '@nangohq/utils';
 
 import { runWithConcurrencyLimit } from '../runWithConcurrencyLimit.js';
 
@@ -23,6 +23,7 @@ export interface DispatchQueuePublisherProps {
 export interface PublishResult {
     enqueued: number;
     failed: number;
+    failedActivityLogIds: string[];
 }
 
 interface BatchPublishResult extends PublishResult {
@@ -61,7 +62,7 @@ export class DispatchQueuePublisher {
      */
     async publish(messages: WebhookDispatchMessage[], messageGroupId: string): Promise<PublishResult> {
         if (messages.length === 0) {
-            return { enqueued: 0, failed: 0 };
+            return { enqueued: 0, failed: 0, failedActivityLogIds: [] };
         }
 
         const activeSpan = tracer.scope().active();
@@ -88,6 +89,7 @@ export class DispatchQueuePublisher {
 
                 const enqueued = results.reduce((sum, r) => sum + r.enqueued, 0);
                 const failed = results.reduce((sum, r) => sum + r.failed, 0);
+                const failedActivityLogIds = results.flatMap((r) => r.failedActivityLogIds);
                 const retriedEntries = results.reduce((sum, r) => sum + r.retriedEntries, 0);
                 const retriedBatches = results.filter((r) => r.retriedEntries > 0).length;
 
@@ -109,7 +111,7 @@ export class DispatchQueuePublisher {
                     metrics.increment(metrics.Types.WEBHOOK_DISPATCH_PUBLISH_FAILURE, failed, { provider });
                 }
 
-                return { enqueued, failed };
+                return { enqueued, failed, failedActivityLogIds };
             } catch (err) {
                 span.setTag('error', err);
                 throw err;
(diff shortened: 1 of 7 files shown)


packages/server/lib/webhook/internal-nango.ts
```diff
@@ -140,20 +197,136 @@ export class InternalNango {
...
+                        parentSyncName: syncConfig.sync_name,
+                        activityLogId: logCtx.id,
+                        webhookName: webhook,
+                        connection: {
+                            id: connection.id,
+                            connection_id: connection.connection_id,
+                            provider_config_key: connection.provider_config_key,
+                            environment_id: connection.environment_id
+                        },
+                        payload: body
```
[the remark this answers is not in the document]
agusayerza: Implemented on #5976

packages/server/lib/webhook/internal-nango.ts
```diff
@@ -149,11 +285,127 @@ export class InternalNango {
...
+            });
+        }
+
+        const messageGroupId = `account:${this.team.id}:env:${this.environment.id}`;
+        const publishResult = await publisher.publish(messages, messageGroupId);
+        const failedActivityLogIds = new Set(publishResult.failedActivityLogIds);
+
+        for (const { message, logCtx } of queuedExecutions) {
+            if (!failedActivityLogIds.has(message.activityLogId)) {
+                void logCtx.info('The webhook was successfully queued for execution', {
```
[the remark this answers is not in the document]
marcindobry: This is true

packages/server/lib/webhook/internal-nango.ts
```diff
@@ -1,13 +1,75 @@
...
+    activityLogId
+}: {
+    environmentId: number;
+    providerConfigKey: string;
+    parentSyncName: string;
+    connectionId: number;
+    activityLogId: string;
+}): string {
+    const hash = createHash('sha256').update(`${environmentId}:${providerConfigKey}:${parentSyncName}:${connectionId}:${activityLogId}`).digest('hex');
+    return `webhook:${hash.slice(0, 32)}`;
```
TBonnin: it would be nice to have human readable prefix when debugging

packages/server/lib/webhook/internal-nango.ts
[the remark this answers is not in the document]
agusayerza: Fixed, but its highly unlikely to be a problem. That would mean AWS is returning an id we didn't send. Sounds like an AWS bug to me haha. I feel like this is being defensive for the sake of being defensive.

packages/server/lib/webhook/dispatch-queue/publisher.ts
```diff
@@ -160,8 +190,13 @@ function indexToEntryId(index: number): string {
...
 
-function entryIdToIndex(id: string): number {
-    return Number.parseInt(id.slice(1), 10);
+function entryIdToIndex(id: string): number | null {
+    if (!id.startsWith('m')) {
+        return null;
+    }
+
+    const index = Number.parseInt(id.slice(1), 10);
+    return Number.isNaN(index) ? null : index;
```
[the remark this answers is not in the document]
agusayerza: We're the ones setting these IDs via `indexToEntryId`, so an out-of-range or malformed ID coming back from SQS isn't a realistic scenario. I don't think this warrants a code change.

packages/server/lib/webhook/internal-nango.ts
```diff
@@ -140,7 +237,46 @@ export class InternalNango {
...
+
+                if (type === webhook || webhookHeaderValue === webhook || webhook === '*') {
+                    for (const connection of connections) {
+                        matchedExecutions.push({ syncConfig, webhook, connection });
+                    }
+
+                    matched = true;
                     if (webhook === '*') {
                         // Only trigger once since it will match all webhooks
                         break;
```
TBonnin: we don't need the break here anymore I think since matched = true, the next loop `if (matched) break;` will take care of the early return

packages/server/lib/webhook/internal-nango.ts
```diff
@@ -149,11 +285,136 @@ export class InternalNango {
...
+                            connection: {
+                                id: connection.id,
+                                connection_id: connection.connection_id,
+                                provider_config_key: connection.provider_config_key,
+                                environment_id: connection.environment_id
+                            },
+                            payload: body
+                        }
+                    };
+                } catch (err) {
```
TBonnin, line 342: we need to fail the logCtx here I think

packages/server/lib/webhook/internal-nango.ts
```diff
@@ -149,11 +285,136 @@ export class InternalNango {
...
+                taskName: message.taskName
+            });
+
+            await logCtx.error('The webhook failed to queue for execution', {
+                error,
+                webhook: message.webhookName,
+                connection: message.connection.connection_id,
+                integration: message.connection.provider_config_key
+            });
+            await logCtx.enrichOperation({ error });
```
TBonnin, line 425: that's redundant with the logctx.error() above

packages/server/lib/webhook/internal-nango.ts
```diff
@@ -149,11 +285,136 @@ export class InternalNango {
...
+        }
+
+        const messageGroupId = `account:${this.team.id}:env:${this.environment.id}`;
+        const publishResult = await publisher.publish(messages, messageGroupId);
+        const failedActivityLogIds = new Set(publishResult.failedActivityLogIds);
+        const unmappedFailureCount = publishResult.failed - failedActivityLogIds.size;
+
+        for (const { message, logCtx } of queuedExecutions) {
+            if (!failedActivityLogIds.has(message.activityLogId) && unmappedFailureCount === 0) {
+                void logCtx.info('The webhook was successfully queued for execution', {
```
TBonnin, line 406: the other logCtx operation are awaited

packages/server/lib/webhook/internal-nango.ts
agusayerza: Is it redudant? Will logging the error update the operation as well? I'm open to keeping one, but my intention was to log the error + update the operation itself

packages/server/lib/webhook/internal-nango.ts
agusayerza: This was intentional, as we are potentially dealing with fanning out to multiple connections and wouldn't want to wait on all successes. But I'm open to changing it to await if you think its best

packages/server/lib/webhook/internal-nango.ts
TBonnin: the `logCtx` below is updating the operation. Do we need to bubble up the error to the operation level?

packages/server/lib/webhook/internal-nango.ts
```diff
@@ -140,20 +237,222 @@ export class InternalNango {
...
+                                    syncConfigId: syncConfig.id,
+                                    syncName: syncConfig.sync_name,
+                                    webhook,
+                                    connectionId: connection.id,
+                                    connection: connection.connection_id,
+                                    integration: connection.provider_config_key
+                                });
+                            }
+                        }
+                    }
```
TBonnin, line 354: I don't understand why we need a for loop with anonymous function instead of calling `logCtx.error(), logCtx...` directly.

TBonnin (approved):

PR #5976, opened 2026-04-28
fix(webhooks): bypass SQS for oversize dispatch payloads 4/5

agusayerza:
Keep large-but-valid webhook messages on the dispatch queue while routing payloads above the 1 MiB SQS limit straight to the orchestrator so they are still delivered.

### DIFF
--- a/packages/server/lib/webhook/internal-nango.ts
+++ b/packages/server/lib/webhook/internal-nango.ts
@@ -10,11 +10,21 @@ import { envs } from '../env.js';
 import { runWithConcurrencyLimit } from './runWithConcurrencyLimit.js';
 import { getOrchestrator } from '../utils/utils.js';
 import { dispatchQueuePublisher } from './dispatch-queue/client.js';
+import { SQS_BATCH_MAX_BYTES } from './dispatch-queue/publisher.js';
 
-import type { DispatchQueuePublisher } from './dispatch-queue/publisher.js';
-import type { LogContextGetter } from '@nangohq/logs';
-import type { Config } from '@nangohq/shared';
-import type { ConnectionInternal, DBConnectionDecrypted, DBEnvironment, DBIntegrationDecrypted, DBPlan, DBSyncConfig, DBTeam, Metadata } from '@nangohq/types';
+import type { DispatchQueuePublisher, PreparedDispatchMessage } from './dispatch-queue/publisher.js';
+import type { LogContext, LogContextGetter } from '@nangohq/logs';
+import type {
+    ConnectionInternal,
+    DBConnectionDecrypted,
+    DBEnvironment,
+    DBIntegrationDecrypted,
+    DBPlan,
+    DBSyncConfig,
+    DBTeam,
+    Metadata,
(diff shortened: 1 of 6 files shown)


packages/server/lib/webhook/internal-nango.ts
```diff
@@ -379,33 +383,64 @@ export class InternalNango {
...
-                error,
-                webhook: message.webhookName,
-                connection: message.connection.connection_id,
-                integration: message.connection.provider_config_key
-            });
-            await logCtx.enrichOperation({ error });
-            await logCtx.failed();
+            const orchestrator = getOrchestrator();
+            for (const { syncConfig, webhook, connection, logCtx } of oversizedExecutions) {
+                await orchestrator.triggerWebhookWithLogContext({
```
marcindobry: Any reason not to use dispatchViaOrchestrator here instead?

packages/shared/lib/clients/orchestrator.ts
```diff
@@ -253,6 +253,36 @@ export class Orchestrator {
...
+                connection: { id: connection.id, name: connection.connection_id },
+                syncConfig: { id: syncConfig.id, name: syncConfig.sync_name }
+            }
+        );
+        logCtx.attachSpan(new OtlpSpan(logCtx.operation));
+
+        return await this.triggerWebhookWithLogContext({ connection, webhookName, syncConfig, input, maxConcurrency, logCtx });
+    }
+
+    async triggerWebhookWithLogContext<T = unknown>({
```
TBonnin: `triggerWebhook` is only used in one place. Let's replace `logContextGetter` param with `logCtx` and create the context at the calling site so we don't have to have 2 functions

TBonnin (commented):
do we know how often webhook payload is bigger than 1MiB?

packages/server/lib/webhook/internal-nango.ts
agusayerza: We needed to split it to a new function that receives the already created `logCtx`. Did that regardless so we can reuse the code. Good call.

agusayerza:
> do we know how often webhook payload is bigger than 1MiB?

Its not very often, but we have just deployed a metric to track this

TBonnin (approved):

packages/server/lib/webhook/internal-nango.ts
```diff
@@ -206,50 +212,83 @@ export class InternalNango {
...
+                        break;
                     }
                 }
             }
-        } finally {
-            metrics.increment(metrics.Types.WEBHOOK_DIRECT_TRIGGER_SUCCESS, scheduledCount, { provider: this.integration.provider });
         }
+
+        const dispatchResult = await this.dispatchExecutionsViaOrchestrator(executions, body);
+        metrics.increment(metrics.Types.WEBHOOK_DIRECT_TRIGGER_SUCCESS, dispatchResult.succeededCount, { provider: this.integration.provider });
```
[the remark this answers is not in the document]
agusayerza: By design

PR #5979, opened 2026-04-29
fix(shared): honor NANGO_SECRET_KEY_<ENV> for default API secret on self-hosted

kishenpateldotwork:
**Problem:** Self-hosted installs can set `NANGO_SECRET_KEY_<ENV_NAME>` for auth/dashboard, but code paths that load the default API secret from the DB did not apply that override—webhook HMAC and similar behavior could diverge.

**Solution:** In `getDefaultSecretForEnv`, on non-cloud, look up the env name and replace the decrypted secret with `NANGO_SECRET_KEY_<ENV_NAME>` when defined. Renamed `getInternalSecretForEnv` → `getDefaultSecretForEnv` and updated callers.

1. Self-hosted: set `NANGO_SECRET_KEY_<ENV>` for an environment and trigger a flow that uses the default API secret (e.g. webhook signing / dry-run that uses that secret). Confirm behavior matches when resolving the same env via secret key.
2. Run affected tests, e.g. `secret.service` integration tests (and any touched packages’ test suites).


packages/shared/lib/services/secret.service.ts
```diff
@@ -23,7 +23,21 @@ class SecretService {
...
             }
-            return Ok(encryptionManager.decryptAPISecret(secret)); // Callers expect unencrypted secret.
+            const decrypted = encryptionManager.decryptAPISecret(secret); // Callers expect unencrypted secret.
+
+            // Self-hosted operator override: NANGO_SECRET_KEY_<ENV_NAME> takes precedence,
+            // mirroring the override honored by account.service#getAccountContextBySecretKey
+            // and the dashboard's getEnvironment controller. Without this, webhook HMAC
+            // signing diverges from what callers expect when they configure the env var.
+            if (!isCloud) {
+                const env = await trx.select<Pick<DBEnvironment, 'name'>>('name').from<DBEnvironment>('_nango_environments').where({ id: envId }).first();
```
TBonnin: Making an additional query to the db for non cloud setup. 
All callers of `getDefaultSecretForEnv` seems to already hold the entire environment object. Let's pass it (or at least the id and name) to this function so this additional round-trip to the db can be avoided

packages/shared/lib/services/secret.service.ts
```diff
@@ -1,29 +1,42 @@
...
+                throw new NangoError('no_default_api_secret', { environment_id: env.id });
             }
-            return Ok(encryptionManager.decryptAPISecret(secret)); // Callers expect unencrypted secret.
+            const decrypted = encryptionManager.decryptAPISecret(secret); // Callers expect unencrypted secret.
+
+            // Self-hosted operator override: NANGO_SECRET_KEY_<ENV_NAME> takes precedence,
+            // mirroring the override honored by account.service#getAccountContextBySecretKey
+            // and the dashboard's getEnvironment controller. Without this, webhook HMAC
+            // signing diverges from what callers expect when they configure the env var.
+            if (!isCloud) {
```
TBonnin: we can condition the override on the presence of the env var alone. No need to check if not `isCloud` imho

packages/shared/lib/services/secret.service.ts
kishenpateldotwork: good catch! Updated -- thanks @TBonnin

packages/shared/lib/services/secret.service.ts
kishenpateldotwork: @TBonnin sounds good -- updated!

packages/shared/lib/services/secret.service.ts
```diff
@@ -5,25 +5,36 @@ import { Err, Ok } from '@nangohq/utils';
...
             }
-            return Ok(encryptionManager.decryptAPISecret(secret)); // Callers expect unencrypted secret.
+            const decrypted = encryptionManager.decryptAPISecret(secret); // Callers expect unencrypted secret.
+
+            // Self-hosted operator override: NANGO_SECRET_KEY_<ENV_NAME> takes precedence,
+            // mirroring the override honored by account.service#getAccountContextBySecretKey
+            // and the dashboard's getEnvironment controller. Without this, webhook HMAC
+            // signing diverges from what callers expect when they configure the env var.
+            const override = process.env[`NANGO_SECRET_KEY_${env.name.toUpperCase()}`];
+            if (override) {
```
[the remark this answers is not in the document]
kishenpateldotwork: @TBonnin 😅 should i add it back in?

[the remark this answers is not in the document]
TBonnin: no. I think the bot is over defensive

TBonnin (approved):

kishenpateldotwork:
@TBonnin thank you for getting this across the finish line for me!! 🙏

PR #6001, opened 2026-04-30
feat(billing): instrument UsageBillingClient.getUsage for Orb perf observability

pfreixes:
## Summary

The dashboard read path (`/plans/billing-usage` → Orb) currently has zero observability — no per-call latency, no cache-hit ratio, no error counter. This blocks any like-for-like comparison of the Orb-backed dashboard with alternatives (e.g. ClickHouse — see [iteration 1 report](https://linear.app/nango/document/clickhouse-metering-read-queries-72ca56b0ab7c) on NAN-5390).

This PR adds three metrics in `packages/usage/lib/billing.ts` (`UsageBillingClient.getUsage`):

| Metric | Tags | Captures |
|---|---|---|
| `BILLING_USAGE_CACHE` (counter) | `hit: true \| false` | Redis cache hit ratio for the 60-second Orb cache |
| `BILLING_USAGE_ORB_MS` (distribution) | `dashboard: true \| false` | Pure Orb call latency (inside throttle, around SDK call). The raw apples-to-apples comparison metric. |
| `BILLING_USAGE_ORB_ERRORS` (counter) | `dashboard: true \| false` | Orb call failures (`metrics.time` doesn't tag by `Result` success). |

The `dashboard` tag is derived from `opts.timeframe` presence — distinguishes the billing dashboard's per-day series queries from the left-column summary's revalidate calls (which share the same code path).

Uses the existing `metrics.time()` helper from `@nangohq/utils`, which uses `process.hrtime` under the hood (monotonic clock, not `Date.now()`).

## Why this specific timer placement

`BILLING_USAGE_ORB_MS` is **inside** the throttle callback, around `this.billingClient.getUsage(...)`. This excludes queue wait time and cache lookup — we want the pure Orb response time for backend-vs-backend comparison. Queue wait visibility was considered and dropped from this PR; we can revisit if rate-limit queueing becomes a concern.

The cache hit/miss counter is at the top before the throttle — captures every call, including ones that never hit Orb.

## Test plan

- [ ] Type-check passes (`npx tsc -b packages/usage`)
- [ ] After deploy: confirm metrics appear in DataDog under `nango.billing.usage.*`
- [ ] After deploy: validate `BILLING_USAGE_CACHE{hit:true}` ratio is meaningful (today the cache is supposed to absorb most dashboard traffic)
- [ ] After a week of data: pull p50/p95/p99 of `BILLING_USAGE_ORB_MS{dashboard:true}` to compare against the [ClickHouse iteration-1 numbers](https://linear.app/nango/document/clickhouse-metering-read-queries-72ca56b0ab7c)

## Related

NAN-5390 — Validate performance of metering read queries.


packages/usage/lib/billing.ts
```diff
@@ -38,26 +38,44 @@ export class UsageBillingClient {
...
-            }
-            return res;
-        });
+        const tags = { dashboard: opts?.timeframe ? 'true' : 'false' };
+
+        // Pure Orb call latency (no queue wait, no cache lookup) — apples-to-apples for any
+        // Orb-vs-alternative-backend comparison.
+        const callOrb = metrics.time(metrics.Types.BILLING_USAGE_ORB_MS, () => this.billingClient.getUsage(subscriptionId, opts), tags);
+
+        // Total cache-miss path (queue wait + Orb call + cache write). Subtracting BILLING_USAGE_ORB_MS
```
TBonnin: do we really need a metric for the cache/miss path though. The only piece not in our control (and that will be replaced) is the call to Orb?

pfreixes: We can live without it, queue time is not important right now, we can just compare end latency and also the overall traffic that we are not sending to Orb with the cache hits

packages/usage/lib/billing.ts
```diff
@@ -38,15 +38,22 @@ export class UsageBillingClient {
...
+        // Orb-vs-alternative-backend comparison.
+        const callOrb = metrics.time(metrics.Types.BILLING_USAGE_ORB_MS, () => this.billingClient.getUsage(subscriptionId, opts), tags);
 
-        // global throttling to avoid exceeding Orb usage endpoint rate limits
         return this.throttle('usage', async () => {
-            const res = await this.billingClient.getUsage(subscriptionId, opts);
+            const res = await callOrb();
             if (res.isOk()) {
                 try {
                     await this.redis.set(cacheKey, JSON.stringify(res.value), {
```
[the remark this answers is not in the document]
pfreixes: we will not be measuring it right now

marcindobry (approved):

PR #6004, opened 2026-04-30
feat(lambda): keep warm cron

rossmcewan:
- New cron that runs every x min (by default 0 so it doesn't run)
- Retrieves all environments created in the last 24 hours and publishes a message with env id, account id, and provisioned concurrency (default 1)
- these messages are picked up by a subscriber in jobs and the Lambda functions are pinged to wake up the required amount
- this is an attempt to resolve the cold start issues


packages/server/lib/crons/lambdaKeepWarm.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export async function exec(): Promise<void> {
+    const since = new Date(Date.now() - TWENTY_FOUR_H_MS);
+
+    const rows = await db.readOnly
+        .select<{ account_id: number; id: number }[]>('_nango_environments.account_id', '_nango_environments.id')
+        .from<DBEnvironment>('_nango_environments')
+        .join('plans', 'plans.account_id', '_nango_environments.account_id')
+        .where('_nango_environments.deleted', false)
+        .where('_nango_environments.created_at', '>=', since)
```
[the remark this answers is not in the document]
rossmcewan: This is by design for now

TBonnin (commented):
do we really need this? it doesn't cover customers that create an account and only used it days later. It doesn't cover customers where POC last for more than a day. It looks like a hacky bandaid for something we don't even know has an impact

packages/server/lib/crons/lambdaKeepWarm.ts
```diff
@@ -0,0 +1,71 @@
...
+        .where('plans.lambda_tenant_isolation', true);
+
+    for (const row of rows) {
+        const res = await pubsub.publisher.publish({
+            subject: 'lambda_keep_warm',
+            type: 'lambda_keep_warm.invoke',
+            payload: {
+                accountId: row.account_id,
+                environmentId: row.id,
+                provisionedConcurrency: 1
```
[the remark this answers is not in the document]
rossmcewan: This is just the starting point - we will expand provisioned concurrency when we have a better mechanism for determining it per tenant.

packages/shared/lib/services/environment.service.ts
```diff
@@ -83,46 +115,51 @@ class EnvironmentService {
...
                 throw webhookKey.error;
             }
 
-            return environment;
+            return env;
         });
+
+        if (environment) {
+            await onNewEnvironment(trx, { accountId, environmentId: environment.id });
+        }
```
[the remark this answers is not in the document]
rossmcewan: I don't think this is a problem. It's very cheap and doesn't create any new resources.

packages/shared/lib/services/environment.service.ts
```diff
@@ -83,46 +115,51 @@ class EnvironmentService {
...
             if (webhookKey.isErr()) {
                 throw webhookKey.error;
             }
 
-            return environment;
+            return env;
         });
+
+        if (environment) {
+            await onNewEnvironment(trx, { accountId, environmentId: environment.id });
```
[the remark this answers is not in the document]
rossmcewan: I don't think this is a problem. It's very cheap and doesn't create any new resources.

packages/server/lib/crons/lambdaKeepWarm.ts
```diff
@@ -0,0 +1,99 @@
...
+        try {
+            logger.info('Starting');
+
+            const ttlMs = cronMinutes * 60 * 1000;
+            const lockKey = 'lock:lambdaKeepWarm:cron';
+
+            try {
+                lock = await locking.acquire(lockKey, ttlMs);
+            } catch {
+                logger.info('Could not acquire lock, skipping');
```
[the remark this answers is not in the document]
rossmcewan: This isn't a failure though - skipping is fine - it means another pod has acquired the lock

packages/shared/lib/services/environment.service.ts
```diff
@@ -16,8 +19,41 @@ import type { DBEnvironment, DBEnvironmentVariable, SdkLogger } from '@nangohq/t
 
 const TABLE = '_nango_environments';
 
+const envs = parseEnvs(ENVS);
+
 export const defaultEnvironments = [PROD_ENVIRONMENT_NAME, 'dev'];
 
+/** After an environment row exists (outer transaction may still be open). Publishes keep-warm when Lambda + plan tenant isolation are enabled. */
```
TBonnin, line 24: > outer transaction may still be open

Is that ok to invoke the keep-warm function when environment is actually not created?

packages/jobs/lib/processors/lambdaKeepWarm.processor.ts
```diff
@@ -0,0 +1,95 @@
...
+
+    const tenantId = getLambdaTenantIdFromAccountEnv(accountId, environmentId);
+    const n = Math.max(1, Math.floor(provisionedConcurrency));
+
+    const results = await Promise.all(Array.from({ length: n }, () => invokeLambdaReadinessCheckEvent({ functionArn: url, tenantId })));
+
+    for (const r of results) {
+        if (r.isErr()) {
+            report(new Error('lambda_keep_warm_invoke_failed', { cause: r.error }), { accountId, environmentId });
+            return;
```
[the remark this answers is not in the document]
TBonnin: also do we care about reporting only the first error?

packages/jobs/lib/runner/lambda.ts
```diff
@@ -38,6 +38,63 @@ const cloudwatchLogsClient = new CloudWatchLogsClient();
...
+    };
+    const command = new InvokeCommand({
+        FunctionName: params.functionArn,
+        Payload: JSON.stringify(readinessCheckRequest),
+        InvocationType: 'Event',
+        ...(params.tenantId !== undefined && { TenantId: params.tenantId })
+    });
+    try {
+        const response = await lambdaClient.send(command);
+        if (response.StatusCode !== 202 && response.StatusCode !== 200) {
```
TBonnin: can an `Event` invocation ever return a `200`?

packages/shared/lib/services/environment.service.ts
```diff
@@ -16,8 +19,41 @@ import type { DBEnvironment, DBEnvironmentVariable, SdkLogger } from '@nangohq/t
...
+async function onNewEnvironment(trx: Knex, { accountId, environmentId }: { accountId: number; environmentId: number }): Promise<void> {
+    try {
+        if (!envs.LAMBDA_ENABLED) {
+            return;
+        }
+        const planRes = await getPlan(trx, { accountId });
```
TBonnin, line 30: is the plan guaranteed to be created before the environment for new account?

packages/server/lib/crons/lambdaKeepWarm.ts
```diff
@@ -0,0 +1,99 @@
...
+            } catch (err) {
+                report(new Error('cron_failed_lambda_keep_warm', { cause: err }));
+            } finally {
+                metrics.duration(metrics.Types.CRON_LAMBDA_KEEP_WARM, Date.now() - start);
+                logger.info('✅ done');
+            }
+        })().catch((err: unknown) => {
+            logger.error('Failed to execute lambdaKeepWarm cron job');
+            report(new Error('cron_failed_lambda_keep_warm', { cause: err }));
+        });
```
TBonnin, line 43: any reason for the double try/catch?

packages/jobs/lib/processors/lambdaKeepWarm.processor.ts
```diff
@@ -0,0 +1,95 @@
...
+                try {
+                    await processKeepWarm(event);
+                } catch (err) {
+                    report(new Error('lambda_keep_warm_handler_failed', { cause: err }), { event });
+                }
+            }
+        });
+    }
+}
+
```
TBonnin, line 49: do `lambdaKeepWarm` need a `stop` function?

packages/utils/lib/environment/parse.ts
```diff
@@ -470,6 +472,11 @@ export const ENVS = z.object({
     NANGO_ACTIVEMQ_CONNECT_TIMEOUT_MS: z.coerce.number().optional().default(10_000),
 
     // Lambda
+    LAMBDA_KEEP_WARM_ACCOUNT_AGE_MS: z.coerce
+        .number()
+        .optional()
+        .default(24 * 60 * 60 * 1000),
+    LAMBDA_KEEP_WARM_SUBSCRIBE_CONCURRENCY: z.coerce.number().min(1).max(10).optional().default(4),
```
TBonnin: no need for optional if using `default`

packages/pubsub/lib/transport/sns-sqs.ts
```diff
@@ -28,6 +28,14 @@ function subscriptionKey<TSubject extends Event['subject']>(consumerGroup: strin
     return `${consumerGroup}:${subject}`;
 }
 
+function subscribeConcurrency(concurrency: number | undefined): number {
```
TBonnin, line 31: can concurrency be undefined? Is it coming from an env var with default no?

packages/jobs/lib/utils/lambda.ts
```diff
@@ -23,22 +23,37 @@ export function getLambdaFunctionName(node: Node): string {
...
+    return getLambdaTenantIdFromAccountEnv(nangoProps.team.id, nangoProps.environmentId);
 }
 
-export function getRoutingId(params: { nangoProps: NangoProps; routingContext?: RoutingContext | undefined }): string {
-    const tshirtSize = getTShirtSize(params);
-    const prefix = params.routingContext?.plan?.fleet_node_routing_override || envs.LAMBDA_DEFAULT_PREFIX;
+interface PlanRoutingSlice {
+    fleet_node_routing_override: string | null;
+    lambda_tenant_isolation: boolean;
+}
```
TBonnin: Would be good to link this type with the Plan definition with something like `type PlanRoutingSlice = Pick<PlanDefinition, 'fleet_node_routing_override', ...>` to make the relationship explicit

and do it directly in `getRoutingIdFromPlan` since this is the only place it is being used

packages/jobs/lib/app.ts
```diff
@@ -126,6 +128,7 @@ try {
     processor.start();
 
     invocationsProcessor.start();
+    lambdaKeepWarmProcessor.start();
```
TBonnin, line 132: I should be stopped gracefully as well I think

packages/utils/lib/environment/parse.ts
TBonnin: why `4`? Isn't one subscriber per instance not enough?

TBonnin (approved):
I have already stated earlier, I don't think this PR should exist but from a code point of view it looks fine (minus a few comments)

packages/shared/lib/services/environment.service.ts
rossmcewan: Yes. We don't retrieve the environment information at a later time, it's just used to construct the tenant id. If the transaction fails and the environment is not created the Lambda would have been invoked but won't happen again.

packages/shared/lib/services/environment.service.ts
rossmcewan: Yes, the sequence was changed in the account service.

packages/server/lib/crons/lambdaKeepWarm.ts
rossmcewan: Same pattern as other cron processes.

packages/utils/lib/environment/parse.ts
rossmcewan: I was running a test in development which doesn't scale beyond one so this was a way to add multiple listeners - happy to make default 1.

packages/pubsub/lib/transport/sns-sqs.ts
rossmcewan: Only for one of the consumers not as a default for sns-sqs

packages/jobs/lib/app.ts
rossmcewan: Added transport.disconnect - this will handle this for the subscribers

packages/utils/lib/environment/parse.ts
TBonnin: unless 1 isn't fast enough, which I doubt, defaulting to 1 is good to start with

PR #6012, opened 2026-04-30
feat(logs): add log type for sync variant creation/deletion

rbwest:
Add log type for sync variant creation/deletion

Screenshot 2026-04-30 at 4 37 57 PM
 
 
 
 Screenshot 2026-04-30 at 4 38 16 PM
 
 
 
Screenshot 2026-04-30 at 4 38 40 PM
 
 
 

[NAN-5419: Add log type for sync variant creation/deletion](https://linear.app/nango/issue/NAN-5419/add-log-type-for-sync-variant-creationdeletion)

-Create/delete sync variant via API
-See new logs


packages/webapp/src/pages/Logs/constants.tsx
```diff
@@ -122,7 +122,9 @@ export const typesOptions = [
             { name: 'Full execution triggered', value: 'sync:request_run_full' },
             { name: 'Sync execution cancelled', value: 'sync:cancel' },
             { name: 'Sync schedule paused', value: 'sync:pause' },
-            { name: 'Sync schedule resumed', value: 'sync:unpause' }
+            { name: 'Sync schedule resumed', value: 'sync:unpause' },
+            { name: 'Sync variant created', value: 'sync:create_variant' },
+            { name: 'Sync variant deleted', value: 'sync:delete_variant' }
```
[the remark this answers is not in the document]
rbwest: Updated.

packages/server/lib/controllers/sync/deleteSyncVariant.ts
```diff
@@ -70,5 +71,32 @@ export const deleteSyncVariant = asyncWrapper<DeleteSyncVariant>(async (req, res
         res.status(500).send({ error: { code: 'failed_sync_variant_deletion' } });
         return;
     }
+
+    if (account) {
```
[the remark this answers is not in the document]
rbwest: Updated.

packages/server/lib/controllers/sync/deleteSyncVariant.ts
```diff
@@ -70,5 +71,30 @@ export const deleteSyncVariant = asyncWrapper<DeleteSyncVariant>(async (req, res
...
+        }
+    );
+    await logCtx.info(`Deleting sync variant '${params.variant}' for '${params.name}'`, {
+        syncName: params.name,
+        syncVariant: params.variant,
+        connection: connection.connection_id,
+        integration: providerConfig?.unique_key ?? body.provider_config_key,
+        syncId: sync.id
+    });
+    await logCtx.success();
```
TBonnin, line 106: the log operation/info should reflect the actual result of the deletion (failure or success)

TBonnin (changes requested):

packages/server/lib/controllers/sync/deleteSyncVariant.ts
rbwest: Hmm... the operation is always successful at this point since I do the log just at the end. I was originally thinking this was just to show a quick trail of successful operations, not show failed requests since they will see the failure mode in the response body immediately and then retry. 

What do you think? Should I render logs for all the failed requests as well with specific context on why the request failed?

TBonnin: I think failed deletion should also be reported. Don't you think?

rbwest: I've been going back and forth here tbh since it seems like extra noise and isn't something we do for other APIs currently, but I suppose you're right. I'll make the update and ping you when it's done.

TBonnin: > isn't something we do for other APIs currently

Which API endpoints are reporting successful operations but not failed ones?

rbwest: Sorry, I just meant not all API endpoints are represented in the logs (e.g. `GET /records`). So I wasn't initially sure if complete coverage of different error codes in the log output made sense, or if it was just noise. 

I agree with you though and I've updated to output specific messages and the correct log type based on what happens in the endpoint. Let me know if there's any additional context I should be attaching to the logs here though.

packages/server/lib/controllers/sync/deleteSyncVariant.ts
```diff
@@ -27,74 +28,77 @@ const paramsValidation = z
     .strict();
 
 export const deleteSyncVariant = asyncWrapper<DeleteSyncVariant>(async (req, res) => {
+    const { account, environment } = res.locals;
+    const logCtxData: OperationRowInsert = { operation: { type: 'sync', action: 'delete_variant' } };
+    const logCtx: LogContextOrigin = await logContextGetter.create(logCtxData, { account, environment });
+
```
TBonnin, line 32: we tend to inline the `OperationRowInsert` object when calling `create`. 
We also don't specify variable type explicitly unless it adds  important information. In general we rely on IDE

packages/server/lib/controllers/sync/deleteSyncVariant.ts
```diff
@@ -27,74 +28,77 @@ const paramsValidation = z
...
     }
 
     const emptyQuery = requireEmptyQuery(req);
     if (emptyQuery) {
-        res.status(400).send({ error: { code: 'invalid_query_params', errors: zodErrorToHTTP(emptyQuery.error) } });
+        const errResponse: ApiError<'invalid_query_params', ValidationError[]> = {
+            error: { code: 'invalid_query_params', errors: zodErrorToHTTP(emptyQuery.error) }
+        };
+        res.status(400).send(errResponse);
+        await logCtx.error(errResponse.error.message ?? 'Invalid query parameters', { errors: errResponse.error.errors });
```
TBonnin: might be less noisy to only have a variable for the errors
```
const errors: zodErrorToHTTP(emptyQuery.error)
res.status(400).send({ error: { code: 'invalid_query_params', errors} });
await logCtx.error('Invalid query parameters', { errors });
```

rbwest: Good call. Cleaned up and it's much more readable now. Thanks!

packages/server/lib/controllers/sync/deleteSyncVariant.ts
rbwest: Ah, thanks. Used to people being picky over explicit typing. Removed along with the cleanup below.

TBonnin (approved):
There is an overall question around why `create/delete variant` appears in the Nango logs while most of the other API operations don't but code lgtm

rbwest:
> There is an overall question around why `create/delete variant` appears in the Nango logs while most of the other API operations don't but code lgtm

This is mainly to give more visibility for support investigations. Full context from Kelvin [here](https://linear.app/nango/issue/NAN-5419/add-log-type-for-sync-variant-creationdeletion#comment-267b7ba7)

PR #6015, opened 2026-04-30
feat: `nango pull` command

kaposke:
Adds the `nango pull` command. Format is:
`nango pull <integration> <function> [-c/--catalog | -e/--env <env>] [-s/--sync | -a/--action/--on-event]`
`--catalog` and `--env` are mutually exclusive (only one can be provided). If none are provided, we enter the interactive mode to select an environment.

When `--catalog`, we fallback to the same behavior as `nango clone` (most of the code is extracted and reused). When `--env`(or env is selected via interactive mode), we use the new GET /functions/pull` endpoint, also added in this PR.


packages/cli/lib/index.ts
```diff
@@ -584,6 +585,81 @@ program
...
+                if (err instanceof MissingArgumentError) {
+                    this.outputHelp();
+                }
+                process.exit(1);
+            }
+            success = await pullFunction({ ...baseOptions, environmentName });
+        }
+
+        if (!success) {
+            process.exitCode = 1;
```
TBonnin, line 660: let's be consistent. `process.exitCode = 1;` vs `process.exit(1);` on L653

packages/server/lib/controllers/functions/pull/getPull.ts
```diff
@@ -0,0 +1,91 @@
...
+    if (useLocalFiles) {
+        const fileName = `${providerConfigKey}/${scriptTypeToFolder[syncConfig.type]}/${syncConfig.sync_name}.ts`;
+        const check = localFileService.checkForIntegrationSourceFile(fileName);
+        if (!check.result) {
+            return null;
+        }
+        return await fs.promises.readFile(check.path, 'utf8');
+    }
+
+    const dir = syncConfig.file_location.split('/').slice(0, -1).join('/');
```
TBonnin: can we use `path.dirname()` here?

packages/server/lib/controllers/integrations/uniqueKey/functions/getCode.ts
```diff
@@ -0,0 +1,91 @@
...
+
+import { asyncWrapper } from '../../../utils/asyncWrapper.js';
+
+import type { DBSyncConfig, GetFunctionPull, ScriptTypeLiteral } from '@nangohq/types';
+
+const scriptTypeToFolder: Record<ScriptTypeLiteral, 'syncs' | 'actions' | 'on-events'> = {
+    sync: 'syncs',
+    action: 'actions',
+    'on-event': 'on-events'
+};
```
TBonnin, line 18: already defined in `pull.service.ts`

packages/server/lib/controllers/functions/pull/getPull.ts
```diff
@@ -0,0 +1,91 @@
...
+
+import type { DBSyncConfig, GetFunctionPull, ScriptTypeLiteral } from '@nangohq/types';
+
+const scriptTypeToFolder: Record<ScriptTypeLiteral, 'syncs' | 'actions' | 'on-events'> = {
+    sync: 'syncs',
+    action: 'actions',
+    'on-event': 'on-events'
+};
+
+const useLocalFiles = isEnterprise ? !useS3 : isLocal || isTest;
```
TBonnin: related to our discussion the other day. `useS3` should be enough of a condition without checking isEnterprise

packages/server/lib/controllers/functions/pull/getPull.ts
```diff
@@ -0,0 +1,91 @@
...
+        report(err, { syncConfigId: syncConfig.id, providerConfigKey, syncName: syncConfig.sync_name });
+        return null;
+    }
+}
+
+const validationQuery = z
+    .object({
+        integrationId: z.string().min(1),
+        name: z.string().min(1),
+        env: z.string().min(1).optional(),
```
TBonnin: `env` is never used. Environment is inferred from locals below

packages/server/lib/routes.public.ts
```diff
@@ -283,6 +284,9 @@ publicAPI.route('/remote-function/compile').post(remoteFunctionAuth, postRemoteF
 publicAPI.route('/remote-function/dryrun').post(remoteFunctionAuth, postRemoteFunctionDryrun);
 publicAPI.route('/remote-function/deploy').post(remoteFunctionAuth, postRemoteFunctionDeploy);
 
+publicAPI.use('/functions', jsonContentTypeMiddleware);
+publicAPI.route('/functions/pull').get(apiAuth, withScope('environment:deploy'), getFunctionPull);
```
TBonnin: `environment:deploy` scope on a `GET` endpoint is weird. Should it be a `POST` since technically it is creating some entity

packages/cli/lib/services/pull.service.ts
```diff
@@ -0,0 +1,242 @@
...
+        );
+
+        const matches: { folder: 'syncs' | 'actions' | 'on-events'; candidatePath: string; content: string }[] = [];
+        for (const result of probeResults) {
+            if (result.status === 'fulfilled') {
+                matches.push(result.value);
+                contentCache.set(result.value.candidatePath, result.value.content);
+            } else if (!(result.reason instanceof GitHubNotFoundError)) {
+                throw result.reason;
+            }
```
TBonnin, line 167: what happen if reason is `GitHubNotFoundError`?

packages/cli/lib/services/pull.service.ts
```diff
@@ -0,0 +1,242 @@
...
+            return false;
+        }
+
+        let writtenCount = 0;
+        for (const file of files) {
+            if (filesToSkip.has(file.relativePath)) {
+                continue;
+            }
+
+            const content = contentCache.has(file.relativePath) ? contentCache.get(file.relativePath)! : await fetchFileContent(file.relativePath, debug);
```
TBonnin: can the `contentCache.has(file.relativePath)` condition ever be false?

packages/cli/lib/utils/githubTemplates.ts
```diff
@@ -0,0 +1,259 @@
...
+}
+
+export class GitHubNotFoundError extends Error {
+    constructor(path: string) {
+        super(`Template not found: ${path}`);
+        this.name = 'GitHubNotFoundError';
+    }
+}
+
+export function getGitHubHeaders(): Record<string, string> {
```
TBonnin: only used in this file it seems. no need to export

packages/server/lib/controllers/functions/pull/getPull.ts
```diff
@@ -0,0 +1,91 @@
+import fs from 'fs';
```
TBonnin:
```suggestion
import fs from 'node:fs';
```

TBonnin (approved):

packages/server/lib/routes.public.ts
kaposke: Not really creating an entity when pulling (not on the server anyway). But agree the scope could be more specific

packages/cli/lib/services/pull.service.ts
kaposke: It's looking for the filename in `syncs/` `actions/` and `on-events/`. `GitHubNotFoundError` is expected, just means it doesn't exist. It adds the ones that do exist to the `matches` array, then acts accordingly:
- If no matches are found, fail with a graceful message.
- If it only found one match, clone that.
- If it matched more than one type of function, consider it a conflict and prompt the user to disambiguate with `--action`/`--sync`/`--on-event`.

packages/server/lib/controllers/integrations/uniqueKey/functions/getCode.ts
kaposke: True but that's in the CLI, and this one is on the server. The CLI doesn't have access to shared packages

packages/cli/lib/services/pull.service.ts
kaposke: Good catch

packages/server/lib/routes.public.ts
kaposke: For the record: we entirely changed the endpoint.

packages/server/lib/routes.public.ts
```diff
@@ -191,6 +192,9 @@ publicAPI
     .route('/integrations/:uniqueKey')
     .get(apiAuth, withAnyScope('environment:integrations:read', 'environment:integrations:read_credentials'), getPublicIntegration);
 publicAPI.route('/integrations/:uniqueKey').delete(apiAuth, withScope('environment:integrations:write'), deletePublicIntegration);
+publicAPI
+    .route('/integrations/:uniqueKey/functions/:name/code')
+    .get(apiAuth, withAnyScope('environment:integrations:read', 'environment:integrations:read_credentials'), getFunctionCode);
```
TBonnin, line 197: missed this during my review: why does the endpoint needs `read_credentials` permission?

PR #6017, opened 2026-05-01
feat(persist): add endpoint to hard delete records cache 2/4

TBonnin:
This endpoint will be called by the runner to empty the records cache before running a full sync from empty cache.

Up to 50_000 records are deleted per call. I picked it as an arbitrary value so it is bounded for very large dataset. Only big dataset would require more than one round-trip between runner and persist.

### DIFF
--- a/packages/persist/lib/routes/environment/environmentId/connection/connectionId/sync/syncId/job/jobId/deleteHardRecords.ts
+++ b/packages/persist/lib/routes/environment/environmentId/connection/connectionId/sync/syncId/job/jobId/deleteHardRecords.ts
@@ -1,0 +1,99 @@
+import z from 'zod';
+
+import { records } from '@nangohq/records';
+import { connectionService } from '@nangohq/shared';
+import { validateRequest } from '@nangohq/utils';
+
+import { envs } from '../../../../../../../../../env.js';
+import { pubsub } from '../../../../../../../../../pubsub.js';
+
+import type { AuthLocals } from '../../../../../../../../../middleware/auth.middleware.js';
+import type { ApiError, DeleteHardAllRecordsSuccess, Endpoint } from '@nangohq/types';
+import type { EndpointRequest, EndpointResponse, Route, RouteHandler } from '@nangohq/utils';
+
+type DeleteHardRecords = Endpoint<{
+    Method: typeof method;
+    Path: typeof path;
+    Params: {
+        environmentId: number;
+        nangoConnectionId: number;
+        syncId: string;
+        syncJobId: number;
+    };
+    Body: {
+        model: string;
+    };
+    Error: ApiError<'hard_delete_records_failed'>;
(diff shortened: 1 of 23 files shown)


packages/persist/lib/routes/environment/environmentId/connection/connectionId/sync/syncId/job/jobId/deleteHardRecords.ts
```diff
@@ -0,0 +1,75 @@
...
+    })
+    .strict();
+
+const validate = validateRequest<DeleteHardRecords>({
+    parseBody: (data: unknown) => bodySchema.parse(data),
+    parseParams: (data: unknown) => paramsSchema.parse(data)
+});
+
+const handler = async (_req: EndpointRequest, res: EndpointResponse<DeleteHardRecords, AuthLocals>) => {
+    const { nangoConnectionId, environmentId } = res.locals.parsedParams;
```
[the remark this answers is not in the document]
TBonnin: I kept the sync/job as part of the url for consistency as well as debugging. 
Wrt model ownership, connection/model is unique so we will rely on the runner not trying to delete records for another connection or another model

marcindobry (approved):

packages/persist/lib/routes/environment/environmentId/connection/connectionId/sync/syncId/job/jobId/deleteHardRecords.ts
```diff
@@ -0,0 +1,99 @@
...
+                        integrationId: connection?.provider_config_key || 'unknown',
+                        connectionId: connection?.connection_id || 'unknown',
+                        syncId,
+                        model
+                    }
+                }
+            });
+        }
+
+        res.status(200).json({ deletedCount: result.value.count, hasMore: result.value.count === limit });
```
pfreixes, line 85: are we expecting the client to keep deleting the ones that were left due to the limit? if so, is it me or we are missing this logic here [1] https://github.com/NangoHQ/nango/pull/6017/changes#diff-3e464d09e879bc993cae055a51caec4b11cf007da0f798c5a2935f34df3a687bR227

or is this a follow up work?

pfreixes (commented):
Left one comment that might be interesting to clarify

packages/persist/lib/routes/environment/environmentId/connection/connectionId/sync/syncId/job/jobId/deleteHardRecords.ts
pfreixes: found it by myself [here](https://github.com/NangoHQ/nango/pull/6032/changes#diff-6fe99add29c61860a3604ba13d80f27bba263abdce608462c85b623ba84df284R634)

pfreixes (approved):

PR #6018, opened 2026-05-01
fix: validate jwt.decode() tenant ID before using in URL construction

calebcgates:
The OneDrive post-connection hook uses `jwt.decode()` to extract the tenant ID (`tid`) from the OAuth access token, then uses it to construct SharePoint and Microsoft login URLs.
  The decoded payload was unsafely cast via `as MicrosoftDecodedToken` without runtime validation that `tid` exists. If the token lacks a `tid` claim, `tenantId` would be `undefined`,
   causing the refresh token and client secret to be sent to `https://login.microsoftonline.com/undefined/oauth2/v2.0/token`.

  **Fix:** Remove the unsafe type assertion, use `jwt.decode()`'s native return type, and add a runtime check that `tid` exists and is a string before proceeding. Returns early
  (no-op) if validation fails, consistent with the existing guard pattern in the function.

  Ref: [`jwt.decode()` does not verify signatures](https://github.com/auth0/node-jsonwebtoken#jwtdecodetoken--options)

  **Testing:** `npm run ts-build` passes. ESLint passes (verified by pre-commit hook). The change is additive — only adds a guard clause that returns early.

### DIFF
--- a/packages/server/lib/hooks/connection/providers/one-drive/post-connection.ts
+++ b/packages/server/lib/hooks/connection/providers/one-drive/post-connection.ts
@@ -1,7 +1,7 @@
 import axios from 'axios';
 import jwt from 'jsonwebtoken';
 
-import type { MicrosoftDecodedToken, SharePointTokenResponse } from './types.js';
+import type { SharePointTokenResponse } from './types.js';
 import type { InternalNango as Nango } from '../../internal-nango.js';
 import type { OAuth2Credentials } from '@nangohq/types';
 
@@ -10,12 +10,15 @@ export default async function execute(nango: Nango) {
     const credentials = connection.credentials as OAuth2Credentials;
     const integration = await nango.getIntegration();
 
-    const decoded = jwt.decode(credentials.access_token) as MicrosoftDecodedToken;
-    if (!decoded || typeof decoded !== 'object') {
+    const decoded = jwt.decode(credentials.access_token);
+    if (!decoded || typeof decoded === 'string') {
         return;
     }
 
-    const tenantId = decoded.tid;
+    const tenantId = decoded['tid'];
+    if (!tenantId || typeof tenantId !== 'string') {
+        return;
+    }
 
     if (!integration || !integration.oauth_client_id || !integration.oauth_client_secret || !credentials.refresh_token) {
         return;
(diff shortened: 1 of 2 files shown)


calebcgates:
Apologies - I should have opened an issue first per your contributing guidelines. Happy to close this and resubmit after discussion if you'd prefer. The fix addresses a potential credential leakage path in the OneDrive post-connection hook where jwt.decode() result is used in URL construction without validating the tid claim.

packages/server/lib/hooks/connection/providers/one-drive/post-connection.ts
```diff
@@ -10,12 +10,15 @@ export default async function execute(nango: Nango) {
...
+    const decoded = jwt.decode(credentials.access_token);
+    if (!decoded || typeof decoded === 'string') {
         return;
     }
 
-    const tenantId = decoded.tid;
+    const tenantId = decoded['tid'];
+    if (!tenantId || typeof tenantId !== 'string') {
+        return;
+    }
```
TBonnin, line 27: using `zod` to validate the shape of the decoded token would be a clean solution imho

TBonnin (commented):
looks like we have the same issue with microsoft-teams provider. Can we also fix the issue there. 
Thank you very much for your contribution

calebcgates:
> looks like we have the same issue with microsoft-teams provider. Can we also fix the issue there. Thank you very much for your contribution

Thanks for the review @TBonnin — both points make sense. Pushed an update that:

1. Uses `safeParse` with a `z.object({ tid: z.string() })` schema in `one-drive/post-connection.ts`, matching the pattern in `xero/post-connection.ts`.
2. Applies the same `jwt.decode` → zod-validate pattern to `microsoft-teams/post-connection.ts`. The existing `as MicrosoftDecodedToken` cast on line 49 had the same shape issue, so the fix is symmetric. Also removed the inline 35-field `MicrosoftDecodedToken` interface that became dead code in that file once the cast was gone.

Schemas are inline per the xero pattern. If you'd prefer a shared `MicrosoftJWTPayloadSchema` for the two Microsoft providers, happy to refactor.

calebcgates:
@TBonnin A couple of related items I noticed while working on this — flagging here for visibility, happy to open separate follow-up PRs if you'd like:

- `sellsy/post-connection.ts:13` has the same `jwt.decode(token) as <Interface>` pattern (with `corpId` instead of `tid`). Same fix shape.
- The exported `MicrosoftDecodedToken` interface in `one-drive/types.ts` (lines 10–48) is now unused once the cast in `microsoft-teams/post-connection.ts` is gone.

Let me know if you'd like either of those as follow-ups.

TBonnin:
Thank you @calebcgates 

It would be great if you could open PRs for the follow-up items. Thank you again

TBonnin (approved):

PR #6025, opened 2026-05-04
feat: return CLI exit codes for remote functions

marcindobry:
Adds explicit CLI exit codes for compile, dryrun, and deploy failures so remote-functions can classify E2B/Docker command failures from exit codes instead of output regexes. Dryrun service methods now return Result values, and remote/local dryrun and deploy clients use CLI exit-code mapping while preserving stdout-only success parsing.

NAN-5373

- npm run test:unit -- packages/cli/lib/utils.unit.test.ts packages/cli/lib/services/dryrun.service.unit.test.ts packages/cli/lib/services/sdk.service.unit.test.ts packages/server/lib/services/remote-function/command-output.unit.test.ts packages/server/lib/services/remote-function/cli-exit-codes.unit.test.ts packages/server/lib/services/remote-function/dryrun-client.unit.test.ts packages/server/lib/services/remote-function/deploy-client.unit.test.ts
- npm run ts-build


packages/server/lib/services/remote-function/dryrun-client.ts
```diff
@@ -100,15 +101,11 @@ export async function invokeDryrun(request: DryrunRequest): Promise<DryrunResult
                 timeoutMs: remoteFunctionDryrunTimeoutMs,
                 envs
             });
-            const dryrunErrorOutput = getDryrunCommandErrorOutput({ stdout: result.stdout, stderr: result.stderr });
-            if (dryrunErrorOutput) {
-                throw new RemoteFunctionError({ code: 'dryrun_error', message: dryrunErrorOutput, status: 400 });
-            }
 
             return { output: getDryrunCommandSuccessOutput({ stdout: result.stdout, stderr: result.stderr }) };
```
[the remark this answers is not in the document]
marcindobry: This is experimental API, some breakage during deployment is fine.

packages/server/lib/services/remote-function/cli-exit-codes.ts
```diff
@@ -0,0 +1,53 @@
+import type { FunctionErrorCode } from '@nangohq/types';
+
+// Keep in sync with packages/cli/lib/exit-codes.ts.
+export const NangoCliExitCode = {
```
marcindobry, line 4: I couldn't find a good place to put this so that it's shared between the CLI and the server so I'm keeping it duplicated for now. Candidate packages were:

- `shared` - this isn't really "shared" between all packages but mostly backend packages
- `utils` - this isn't published so it would require publishing it or bundling the whole thing into the CLI. It has a lot of heavy stuff inside so not worth it for 3 constants.
- `types` - these are runtime constants, not types, so not a good fit
- `server` - Don't want to import all of server inside the CLI
- `runner-sdk` - The best existing candidate since the CLI already imports it, but the error codes are used by the server and cli, not runner-sdk

An alternative would be a new package but I don't think it's worth it for 3 constants.

kaposke: Always a bit of a pain. I think we need to create an actually shared package eventually. Webapp as the same problem requiring duplication

kaposke (approved):

PR #6027, opened 2026-05-04
fix(deploy): prevent duplicate active sync configs from concurrent deploys

hassan254-prog:

- prevent duplicate active sync configs from concurrent deploys

### DIFF
--- a/packages/server/lib/controllers/sync/deploy/postDeploy.ts
+++ b/packages/server/lib/controllers/sync/deploy/postDeploy.ts
@@ -29,15 +29,14 @@ export const postDeploy = asyncWrapper<PostDeploy>(async (req, res) => {
     const body: PostDeploy['Body'] = val.data;
     const { environment, account, plan } = res.locals;
 
-    // Serialize deploys per environment: queue this deploy until any in-progress deploy finishes.
+    // Prevent concurrent deploys per environment, fail immediately if another deploy is in flight.
     const locking = await getLocking();
-    const acquisitionTimeoutMs = 5 * 60 * 1000;
-    const ttlMs = 10 * 60 * 1000; // must exceed acquisitionTimeoutMs so the lock can't expire while a deploy is still running
+    const ttlMs = 10 * 60 * 1000; // max expected deploy duration
     const lockKey = `lock:deployService:deploy:${account.id}:${environment.id}`;
     let lock: Lock | undefined;
 
     try {
-        lock = await locking.tryAcquire(lockKey, ttlMs, acquisitionTimeoutMs);
+        lock = await locking.acquire(lockKey, ttlMs);
     } catch {
         const logCtx = await logContextGetter.create({ operation: { type: 'deploy', action: 'custom' } }, { account, environment });
         const error = new NangoError('concurrent_deployment');
--- a/packages/server/lib/controllers/sync/deploy/postDeploy.ts
+++ b/packages/server/lib/controllers/sync/deploy/postDeploy.ts
@@ -31,7 +31,7 @@ export const postDeploy = asyncWrapper<PostDeploy>(async (req, res) => {
 
     // Prevent concurrent deploys per environment, fail immediately if another deploy is in flight.
     const locking = await getLocking();
-    const ttlMs = 10 * 60 * 1000; // max expected deploy duration
+    const ttlMs = process.env['DEPLOY_LOCK_TTL_MS'] ? parseInt(process.env['DEPLOY_LOCK_TTL_MS']) : 10 * 60 * 1000; // max expected deploy duration
     const lockKey = `lock:deployService:deploy:${account.id}:${environment.id}`;
     let lock: Lock | undefined;
 
--- a/packages/server/lib/controllers/sync/deploy/postDeploy.ts
+++ b/packages/server/lib/controllers/sync/deploy/postDeploy.ts
@@ -29,15 +29,15 @@ export const postDeploy = asyncWrapper<PostDeploy>(async (req, res) => {
     const body: PostDeploy['Body'] = val.data;
     const { environment, account, plan } = res.locals;
 
-    // we don't allow concurrent deploys so we need to lock this
-    // and reject this deploy if there is already a deploy in progress
+    // Serialize deploys per environment: queue this deploy until any in-progress deploy finishes.
     const locking = await getLocking();
-    const ttlMs = 60 * 1000;
+    const acquisitionTimeoutMs = 5 * 60 * 1000;
+    const ttlMs = 10 * 60 * 1000; // must exceed acquisitionTimeoutMs so the lock can't expire while a deploy is still running
     const lockKey = `lock:deployService:deploy:${account.id}:${environment.id}`;
     let lock: Lock | undefined;
 
     try {
-        lock = await locking.acquire(lockKey, ttlMs);
+        lock = await locking.tryAcquire(lockKey, ttlMs, acquisitionTimeoutMs);
     } catch {
         const logCtx = await logContextGetter.create({ operation: { type: 'deploy', action: 'custom' } }, { account, environment });
         const error = new NangoError('concurrent_deployment');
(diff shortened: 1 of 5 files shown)


packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -158,9 +156,27 @@ export async function deploy({
...
+                    .from<DBSyncConfig>(TABLE)
+                    .where({
+                        environment_id: sc.environment_id,
+                        nango_config_id: sc.nango_config_id,
+                        sync_name: sc.sync_name,
+                        type: sc.type,
+                        active: true,
+                        deleted: false
+                    })
+                    .forUpdate()
```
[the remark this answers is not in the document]
TBonnin: that's a valid point. I think we should do 2 things:
- add a unique constraint so no duplicate can be inserted what so ever at the db level
- serialize sync config creation (or even deployment all together) to make it a better UX (instead of failing). Using FOR UPDATE is a nice idea but it only handles the case of the sync config already exists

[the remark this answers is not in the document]
pfreixes: > add a unique constraint so no duplicate can be inserted what so ever at the db level

If we can do that, then better. We would not even require the for update since the whole transaction should fail

TBonnin (commented):
I feel like queuing deployments for a given environment would be simpler. Allowing deployment to overlap could lead to other nondeterministic behaviors anyway and disallowing concurrent deployments would prevent them all.
wdyt @hassan254-prog @pfreixes?

packages/server/lib/controllers/sync/deploy/postDeploy.ts
```diff
@@ -29,15 +29,15 @@ export const postDeploy = asyncWrapper<PostDeploy>(async (req, res) => {
...
     const locking = await getLocking();
-    const ttlMs = 60 * 1000;
+    const acquisitionTimeoutMs = 5 * 60 * 1000;
+    const ttlMs = 10 * 60 * 1000; // must exceed acquisitionTimeoutMs so the lock can't expire while a deploy is still running
     const lockKey = `lock:deployService:deploy:${account.id}:${environment.id}`;
     let lock: Lock | undefined;
 
     try {
-        lock = await locking.acquire(lockKey, ttlMs);
+        lock = await locking.tryAcquire(lockKey, ttlMs, acquisitionTimeoutMs);
```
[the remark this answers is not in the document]
hassan254-prog: This is already handled [here](https://github.com/NangoHQ/nango/pull/6027/changes#diff-8c7501b889efa4daefff9f09f9dc4c85ec2d62750dc8fde42fbbc2146b215346R107).

packages/shared/lib/services/sync/config/deploy.service.unit.test.ts
```diff
@@ -243,7 +243,14 @@ describe('Sync config create', () => {
             return Promise.resolve([]);
         });
 
-        vi.spyOn(db.knex, 'from').mockRejectedValue(new Error());
+        vi.spyOn(db.knex, 'from').mockReturnValue({
+            where: vi.fn().mockReturnValue({
+                orderBy: vi.fn().mockReturnValue({
+                    first: vi.fn().mockResolvedValue(null)
+                })
```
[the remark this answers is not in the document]
hassan254-prog: This is already handled with this [commit](https://github.com/NangoHQ/nango/pull/6027/commits/fd1c8110e13d4e08b831fb868795992c24857c51).

hassan254-prog:
> I feel like queuing deployments for a given environment would be simpler. Allowing deployment to overlap could lead to other nondeterministic behaviors anyway and disallowing concurrent deployments would prevent them all. wdyt @hassan254-prog @pfreixes?

Agreed, I have updated the pr to reflect this where `tryAcquire` serializes deploys per environment, a second deploy waits (up to 5 mins) for the first to finish rather than running concurrently, and If it can't acquire the lock in time, it gets a 409 instead of overlapping.

packages/server/lib/controllers/sync/deploy/postDeploy.ts
```diff
@@ -29,15 +29,15 @@ export const postDeploy = asyncWrapper<PostDeploy>(async (req, res) => {
     const body: PostDeploy['Body'] = val.data;
     const { environment, account, plan } = res.locals;
 
-    // we don't allow concurrent deploys so we need to lock this
-    // and reject this deploy if there is already a deploy in progress
+    // Serialize deploys per environment: queue this deploy until any in-progress deploy finishes.
     const locking = await getLocking();
-    const ttlMs = 60 * 1000;
+    const acquisitionTimeoutMs = 5 * 60 * 1000;
```
TBonnin: Is it guaranteed a single deploy won't exceed this timeout

packages/server/lib/controllers/sync/deploy/postDeploy.ts
TBonnin: I don't think it is bad to fail immediately if another deploy is already in flight. Is there any advantage in queueing?

packages/shared/lib/services/sync/config/deploy.service.ts
```diff
@@ -300,7 +299,16 @@ async function compileDeployInfo({
     }
 
     const version = optionalVersion || bumpedVersion || '1';
-    const idsToMarkAsInactive: number[] = [];
+
+    // intentionally not filtered by source so a disabled sync stays disabled when switching sources (e.g. catalog → repo).
+    // select all active rows, not just .first(), so any duplicates left by prior races are also cleaned up.
```
TBonnin, line 304: should the same be done in template.ts?

hassan254-prog: `templates.ts` already uses `getSyncAndActionConfigsBySyncNameAndConfigId` which currently returns all active rows.

packages/server/lib/controllers/sync/deploy/postDeploy.ts
hassan254-prog: Agreed, no real advantage, I have switched from `tryAcquire` in the latest [commit](https://github.com/NangoHQ/nango/pull/6027/commits/1c39913b360deb2a3530bcc8a8622031784fe1dc), to `acquire` so it fails immediately with a 409 if another deploy is in progress.

packages/server/lib/controllers/sync/deploy/postDeploy.ts
hassan254-prog: good catch, not guaranteed, so I bumped `ttlMs` to 10 mins in the latest [commit](https://github.com/NangoHQ/nango/pull/6027/commits/1c39913b360deb2a3530bcc8a8622031784fe1dc) which should comfortably cover even large deploys.

TBonnin: do you already have a ticket to refactor the frequency update. That's the part of the deployment that is very un-optimized at the moment?

packages/server/lib/controllers/sync/deploy/postDeploy.ts
```diff
@@ -29,10 +29,9 @@ export const postDeploy = asyncWrapper<PostDeploy>(async (req, res) => {
     const body: PostDeploy['Body'] = val.data;
     const { environment, account, plan } = res.locals;
 
-    // we don't allow concurrent deploys so we need to lock this
-    // and reject this deploy if there is already a deploy in progress
+    // Prevent concurrent deploys per environment, fail immediately if another deploy is in flight.
     const locking = await getLocking();
-    const ttlMs = 60 * 1000;
+    const ttlMs = 10 * 60 * 1000; // max expected deploy duration
```
TBonnin: should we make it an env var in case we want to adjust it later?

TBonnin (approved):

packages/server/lib/controllers/sync/deploy/postDeploy.ts
hassan254-prog: Yes, I have updated this with the latest [commit](https://github.com/NangoHQ/nango/pull/6027/commits/71e7967b0d237e09a7efb68a11be7a4702ff669f).

packages/server/lib/controllers/sync/deploy/postDeploy.ts
```diff
@@ -29,10 +29,9 @@ export const postDeploy = asyncWrapper<PostDeploy>(async (req, res) => {
     const body: PostDeploy['Body'] = val.data;
     const { environment, account, plan } = res.locals;
 
-    // we don't allow concurrent deploys so we need to lock this
-    // and reject this deploy if there is already a deploy in progress
+    // Prevent concurrent deploys per environment, fail immediately if another deploy is in flight.
     const locking = await getLocking();
-    const ttlMs = 60 * 1000;
+    const ttlMs = process.env['DEPLOY_LOCK_TTL_MS'] ? parseInt(process.env['DEPLOY_LOCK_TTL_MS']) : 10 * 60 * 1000; // max expected deploy duration
```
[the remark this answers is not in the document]
hassan254-prog: the codebase doesn't validate numeric env vars elsewhere (e.g. `LOCAL_NANGO_USER_ID` uses the same bare parseInt pattern), and DEPLOY_LOCK_TTL_MS is an internal ops knob that no one will accidentally set to a garbage value.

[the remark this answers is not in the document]
TBonnin: we do validate all env vars. Please use https://github.com/NangoHQ/nango/blob/d8d6185e11021aca1660aa7d153090e0ade73578/packages/utils/lib/environment/parse.ts?plain=1#L100 instead of `process`. Thank you

PR #6046, opened 2026-05-05
feat: add records_batch table

TBonnin:
90% of records being batch saved are not being modified and have a payload identical to what's already stored.
However current logic updates every single row in order to update the sync_job_id that serves as a stamp to determine if track_deletes feature can soft-delete records that haven't been touched by a sync. 

The goal is to stop updating every single rows but instead only update records metadata/payload when hash/content is different and store the ids of all the records in the batch to be used by the track_deletes query at the end of the sync, replacing up to 1000 upserts per batch by a single insert.

This first commit adds the migration for the new table and insert batch entries in this new table. The upsert/update and track_deletes deletion isn't touched yet since it requires all batches to be inserted for a given sync. I am planning to have it run for more than 24h like this to ensure all batches for a given execution are being captured before modifying the rest of the logic.

next PR for context: https://github.com/NangoHQ/nango/pull/6049

### DIFF
--- a/packages/persist/lib/daemons/batchCleanup.daemon.ts
+++ b/packages/persist/lib/daemons/batchCleanup.daemon.ts
@@ -1,0 +1,36 @@
+import tracer from 'dd-trace';
+
+import { records } from '@nangohq/records';
+import { cancellableDaemon } from '@nangohq/utils';
+
+import { envs } from '../env.js';
+import { logger } from '../logger.js';
+
+/*
+ * Batch cleanup daemon
+ * Deletes records_batch entries older than PERSIST_BATCH_CLEANUP_MAX_AGE_MS.
+ */
+export function batchCleanupDaemon(): Awaited<ReturnType<typeof cancellableDaemon>> {
+    return cancellableDaemon({
+        tickIntervalMs: envs.PERSIST_BATCH_CLEANUP_INTERVAL_MS,
+        tick: async (): Promise<void> => {
+            return tracer.trace('nango.persist.daemon.batchCleanup', async (span) => {
+                try {
+                    const olderThan = new Date(Date.now() - envs.PERSIST_BATCH_CLEANUP_MAX_AGE_MS);
+                    const res = await records.deleteOldBatchEntries({ olderThan, limit: envs.PERSIST_BATCH_CLEANUP_LIMIT });
+                    if (res.isErr()) {
+                        span?.addTags({ error: res.error.message });
+                        logger.error(`[Batch cleanup] error deleting old batch entries: ${res.error.message}`);
+                        return;
+                    }
+                    span?.addTags({ deleted: res.value });
+                } catch (err) {
+                    span?.addTags({ error: (err as Error).message });
+                    logger.error(`[Batch cleanup] unexpected error: ${(err as Error).message}`);
+                } finally {
+                    span?.finish();
+                }
+            });
+        }
+    });
+}
--- a/packages/records/lib/db/migrations/20260505000100_create_records_batch.ts
+++ b/packages/records/lib/db/migrations/20260505000100_create_records_batch.ts
@@ -1,0 +1,20 @@
+import type { Knex } from 'knex';
+
+const TABLE = 'records_batch';
+
+export async function up(knex: Knex): Promise<void> {
+    await knex.raw(`
+        CREATE TABLE IF NOT EXISTS "${TABLE}" (
+            id             bigserial    PRIMARY KEY,
+            connection_id  integer      NOT NULL,
+            model          varchar(255) NOT NULL,
+            sync_job_id    integer      NOT NULL,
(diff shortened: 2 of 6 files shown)


packages/records/lib/db/migrations/20260505000100_create_records_batch.ts
```diff
@@ -0,0 +1,20 @@
...
+    await knex.raw(`
+        CREATE TABLE IF NOT EXISTS "${TABLE}" (
+            id             bigserial    PRIMARY KEY,
+            connection_id  integer      NOT NULL,
+            model          varchar(255) NOT NULL,
+            sync_job_id    integer      NOT NULL,
+            record_ids     uuid[]       NOT NULL,
+            created_at     timestamptz  NOT NULL DEFAULT CURRENT_TIMESTAMP
+        )
+    `);
```
TBonnin, line 15: this table is gonna be append only, so no update, no dead tuples and only one entry per batchSave.

packages/utils/lib/environment/parse.ts
```diff
@@ -95,6 +95,12 @@ export const ENVS = z.object({
         .number()
         .optional()
         .default(60 * 24 * 3600 * 1000), // 60 days
+    PERSIST_BATCH_CLEANUP_INTERVAL_MS: z.coerce.number().optional().default(5_000), // set to 0 to disable
+    PERSIST_BATCH_CLEANUP_LIMIT: z.coerce.number().optional().default(10_000),
+    PERSIST_BATCH_CLEANUP_MAX_AGE_MS: z.coerce
+        .number()
+        .optional()
+        .default(72 * 3600 * 1000), // 72 hours
```
TBonnin, line 103: in theory 24h would be enough. 72h for now to be safe. We can adjust later.

packages/persist/lib/daemons/batchCleanup.daemon.ts
```diff
@@ -0,0 +1,36 @@
+import tracer from 'dd-trace';
+
+import { records } from '@nangohq/records';
+import { cancellableDaemon } from '@nangohq/utils';
+
+import { envs } from '../env.js';
+import { logger } from '../logger.js';
+
```
TBonnin, line 8: background cleanup. I might decide to do inline cleanup when sync completes successfully but adding a background job to take care of the orphans in case of failures

packages/records/lib/models/records.ts
```diff
@@ -487,6 +500,13 @@ export async function upsert({
                             batchDeltaSizeInBytes = upsertDataResult.reduce((acc, r) => acc + r.new_size_bytes - r.prev_size_bytes, -totalLegacySizeInBytes);
                         }
 
+                        await insertBatchEntry(trx, {
+                            connectionId,
+                            model,
+                            syncJobId: chunk[0]!.sync_job_id,
+                            recordIds: upsertMetadata.map((r) => r.id)
+                        });
```
TBonnin, line 508: Note: if all functions were using the most recent `track_deletes` API in the runner sdk (trackDeletesStarts/Ends), batchSave could even carry some information that would allow us to skip saving the batch entry when a sync isn't using `track_deletes`

packages/records/lib/db/migrations/20260505000100_create_records_batch.ts
pfreixes: awesome

packages/utils/lib/environment/parse.ts
pfreixes: is it based on the maximum time frequency of 24h for sync? if so, is there any risk of a failed sync to be scheduled in the next cycle? i would say that no right?

packages/records/lib/db/migrations/20260505000100_create_records_batch.ts
```diff
@@ -0,0 +1,20 @@
+import type { Knex } from 'knex';
+
+const TABLE = 'records_batch';
```
pfreixes, line 3: what about `records_sync_seen` the current one is quite generic right?

pfreixes (approved):

packages/utils/lib/environment/parse.ts
TBonnin: yes. the goal is to bound the size of this table. 

> if so, is there any risk of a failed sync to be scheduled in the next cycle
I am not sure I understand.

packages/records/lib/db/migrations/20260505000100_create_records_batch.ts
TBonnin: I prefer 'batch' but 'records_sync_seen' wouldn't be a bad name

packages/utils/lib/environment/parse.ts
pfreixes: meaning why 24 hours is safe enough?

PR #6049, opened 2026-05-05
feat(persist): upsert and track_deletes logic refactor

TBonnin:
follow-up of https://github.com/NangoHQ/nango/pull/6046

This commit:
- uses the new records_batch table to determine which records must be soft-deleted in `
deleteOutdatedRecords`
- stop updating a record if unchanged, potentially rows churn of the
  records table by 90%

### DIFF
--- a/packages/records/lib/models/records.ts
+++ b/packages/records/lib/models/records.ts
@@ -386,7 +386,8 @@ export async function upsert({
                                     `${RECORDS_TABLE}.updated_at`,
                                     trx.raw(`incoming.data_hash IS DISTINCT FROM ${RECORDS_TABLE}.data_hash as has_changed`),
                                     trx.raw(`(${RECORDS_TABLE}.json IS NOT NULL OR ${RECORDS_TABLE}.pruned_at IS NOT NULL) as must_upsert_data`),
-                                    trx.raw(`COALESCE(pg_column_size(${RECORDS_TABLE}.json), 0) as legacy_size_bytes`)
+                                    trx.raw(`COALESCE(pg_column_size(${RECORDS_TABLE}.json), 0) as legacy_size_bytes`),
+                                    trx.raw(`${RECORDS_TABLE}.tableoid::regclass as partition`)
                                 )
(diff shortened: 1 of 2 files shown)


packages/records/lib/models/records.ts
```diff
@@ -420,12 +434,22 @@ export async function upsert({
                                     }
                                 }
                             })
```
TBonnin, line 436: records that are not updated (ie: skipped) are not returned by the upsert CTE anymore but we still need them as part of the result as unchanged

packages/records/lib/models/records.ts
```diff
@@ -1099,32 +1123,39 @@ export async function deleteOutdatedRecords({
...
                             await trx.raw(
-                                `WITH to_delete AS MATERIALIZED (
-                                    SELECT ctid, id
-                                    FROM ${RECORDS_TABLE}
-                                    WHERE connection_id = ?
-                                      AND model = ?
-                                      AND sync_job_id < ?
-                                      AND deleted_at IS NULL
-                                    LIMIT ?
+                                `WITH seen AS MATERIALIZED (
```
pfreixes, line 1126: any concerns on how this materialization of the seen ids would impact the performance of the DB? seems that we will be doing this over and over for each batch?

increasing the batch size will reduce the number of operations for sure!

worst case we could still throttle this loop (we have this freedom right? or the deleteOutdatedRecords records have to have happen as fast as we can?

Also not sure of the difference btw doing the unfolding at the very begining in a temporary table vs doing it inline as Today

pfreixes (approved):

packages/records/lib/models/records.ts
TBonnin: Materializing the seen ids could indeed be optimized. The priority isn't speed though but success.
Using a temp table could be a solution but we are not using session-based connections.
My objective is to go with the simplest solution (even if it means doing some unnecessary materialization) and optimize from here

kaposke (approved):

PR #6053, opened 2026-05-06
feat(integrations): add support for freshworks

hassan254-prog:

- add support for freshworks and few fixes for providers we added untested.


hassan254-prog:
> ### 💡 Codex Review
> https://github.com/NangoHQ/nango/blob/9202ee74e520bb71a7bc39eb58c11f8e789142db/packages/providers/providers.yaml#L6981-L6982
> 
> 
> **P1 Badge Permit Freshsales base path in Freshworks API hostname**
> The new `freshworks` provider hard-validates `connection_config.bundleAlias` as a `hostname`, which prevents users from entering Freshsales-style API roots that include a path segment (for example `/crm/sales`). For Freshworks CRM tenants this yields an incorrect proxy base URL (`https://<domain>`) and proxied `/api/...` calls will target the wrong endpoint instead of the product-specific path, so CRM API requests fail even with valid OAuth credentials.
> 
> ℹ️ About Codex in GitHub

This is okay as the user's can build the api `base_url` from the service hostname.

rbwest (approved):

PR #6055, opened 2026-05-06
fix(runner-sdk): url deny list in uncontrolledFetch

rossmcewan:

### DIFF
--- a/packages/runner-sdk/lib/action.ts
+++ b/packages/runner-sdk/lib/action.ts
@@ -43,6 +43,11 @@ import type * as z from 'zod';
 const MEMOIZED_CONNECTION_TTL = 60000;
 const MEMOIZED_INTEGRATION_TTL = 10 * 60 * 1000;
 
+/** Max hops when following redirects manually (denylist checked per hop). */
+const UNCONTROLLED_FETCH_MAX_REDIRECTS = 20;
+
+const REDIRECT_STATUS_CODES = new Set([301, 302, 303, 307, 308]);
+
 export type ProxyConfiguration = Omit<UserProvidedProxyConfiguration, 'files' | 'providerConfigKey' | 'connectionId'> & {
     providerConfigKey?: string;
     connectionId?: string;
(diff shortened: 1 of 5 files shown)


packages/runner-sdk/lib/action.ts
```diff
@@ -469,6 +470,14 @@ export abstract class NangoActionBase<
         headers?: Record<string, string> | undefined;
         body?: string | null;
     }): Promise<Response> {
+        const baseUrlOverrideDenylist = getBaseUrlOverrideDenylistFromEnv();
```
rossmcewan, line 480: What do we want to do about redirects here? We could do manual redirect handling - I have an implementation ready for that) or we just disable redirects on uncontrolledFetch and risk breaking existing functions.

TBonnin: >  have an implementation ready for that

How does it look like?

packages/runner-sdk/lib/baseUrlOverrideDenylist.ts
```diff
@@ -0,0 +1,121 @@
...
+    const lambdaRuntimeApi = typeof process !== 'undefined' ? process.env['AWS_LAMBDA_RUNTIME_API'] : undefined;
+    if (lambdaRuntimeApi) {
+        const normalized = normalizeDenylistHost(lambdaRuntimeApi);
+        if (normalized) {
+            denylist.add(normalized);
+        }
+    }
+
+    memoizedBaseUrlOverrideDenylist = denylist;
+    return memoizedBaseUrlOverrideDenylist;
```
TBonnin, line 120: don't we have the same code already somewhere?

rossmcewan: We do, but I wasn't sure if we explicitly kept shared out of the runner-sdk.

packages/runner-sdk/lib/action.ts
rossmcewan: Committed

packages/runner-sdk/lib/baseUrlOverrideDenylist.ts
TBonnin: we do want to keep shared or any other package out of `runner-sdk`. Duplication isn't ideal though

packages/runner-sdk/lib/action.ts
```diff
@@ -469,17 +477,115 @@ export abstract class NangoActionBase<
...
+            if (redirectsFollowed >= UNCONTROLLED_FETCH_MAX_REDIRECTS) {
+                void response.body?.cancel();
+                throw new this.ActionError({
+                    code: 'too_many_redirects',
+                    message: `Exceeded maximum of ${UNCONTROLLED_FETCH_MAX_REDIRECTS} redirects.`
+                });
+            }
+
+            // We're about to follow the redirect; we won't return this response, so cancel its body.
+            void response.body?.cancel();
```
TBonnin: could be moved before the if, avoiding the duplicate line in the if block

packages/runner-sdk/lib/action.ts
TBonnin: it sucks we have to re-implement a bit of an http client ourselves. 
There is this package that would give us a `beforeRedirect` hook https://www.npmjs.com/package/follow-redirects but it is adding another dependency. I don't like any of the solutions to be honest

TBonnin (approved):

packages/runner-sdk/lib/action.ts
```diff
@@ -469,17 +477,115 @@ export abstract class NangoActionBase<
...
+                });
+            }
         };
 
-        if (options.body) {
-            props.body = options.body;
-        }
+        throwIfDenied(options.url.toString());
+
+        let currentUrl = new URL(options.url.href);
```
TBonnin, line 493: isn't href already an url?

packages/runner-sdk/lib/action.ts
```diff
@@ -42,6 +43,13 @@ import type * as z from 'zod';
 const MEMOIZED_CONNECTION_TTL = 60000;
 const MEMOIZED_INTEGRATION_TTL = 10 * 60 * 1000;
 
+/** Max hops when following redirects manually (denylist checked per hop). */
+const UNCONTROLLED_FETCH_MAX_REDIRECTS = 20;
```
TBonnin: isn't 20 extremely generous?

packages/runner-sdk/lib/action.ts
rossmcewan: No. String.

packages/runner-sdk/lib/action.ts
rossmcewan: Probably. Will change to 5.

PR #6070, opened 2026-05-07
feat(metering): add ClickHouse -> S3 -> Orb billing events export cron

pfreixes:
## Summary

POC for [NAN-5315](https://linear.app/nango/issue/NAN-5315). Adds a cron in metering that exports daily Orb-shaped billing events from ClickHouse `daily_*` tables to S3 as NDJSON, where Orb's S3 sync picks them up. Pre-aggregated to one event per `(account, day)` per Orb `event_name`.

- **Disabled by default.** Dev wiring arrives via [nango-environments PR](https://github.com/NangoHQ/nango-environments/pull/new/pfreixes/nan-5315-billing-events-s3-export-dev) — events ship under `_poc` event_name suffix for parallel comparison vs the legacy HTTP path.
- **7 of 11 active Orb billable metrics covered**: proxy, function_executions (V2 + compute time + custom logs), webhook_forwards, records, billable_connections_v2, billable_actions, monthly_active_records. The 2 legacy metrics using `count(events)` and `max(count)` aggregations are fundamentally incompatible with pre-aggregated rows (documented inline + in the [Linear doc](https://linear.app/nango/document/orb-billable-metrics-2e0859635bc1)).
- **Role-based S3 auth** via the writer role provisioned in [NangoHQ/nango-infra#113](https://github.com/NangoHQ/nango-infra/pull/113). No static credentials.

## Scope

- 4 env vars (all default disabled/empty): `CRON_BILLING_EVENTS_S3_EXPORT_MINUTES`, `BILLING_EVENTS_S3_BUCKET`, `BILLING_EVENTS_S3_WRITER_ROLE_ARN`, `BILLING_EVENTS_S3_EVENT_NAME_SUFFIX`.
- New cron at `packages/metering/lib/crons/billingEventsS3Export.ts`.
- Exposes `clickhouseClient` from `@nangohq/usage` so the cron can run raw `INSERT INTO FUNCTION s3(...)` queries.
- Integration test (`*.integration.test.ts`) covers all 7 supported metrics + suffix threading + day filtering. Verified red-when-broken against an `environment_id` regression.

## Out of scope (productionise, tracked separately)

- Cutover from HTTP path to S3.
- Reconciliation, replay tooling, late-event handling.
- Migration of plans referencing `count(events)` / `max(count)` legacy Orb metrics.

## Test plan

- [x] `npm run test:integration -- --dir packages/metering` (9 tests, all pass locally)
- [x] After the nango-environments PR lands and dev image bumps, verify metering logs show the export cron firing and `Exporting proxy_poc for day=...` lines appear
- [x] Verify objects land in `s3://nango-billing-events-development/<yyyymmdd>/`
- [ ] After 24h of dev runs, compare `_poc` Orb billable metrics against the legacy HTTP-path metrics for parity


packages/metering/lib/crons/billingEventsS3Export.ts
```diff
@@ -0,0 +1,256 @@
...
+    await tracer.trace<Promise<void>>('nango.cron.billingEventsS3Export', async () => {
+        logger.info(`Starting`);
+        await withLock(async () => {
+            const client = clickhouseClient();
+            if (!client) {
+                logger.error(`Clickhouse client not configured`);
+                return;
+            }
+            try {
+                const day = yesterdayUTC();
```
pfreixes, line 169: We will for know publishing the same events every one hour, that is fine for the POC (the idempotent key will make the work )

packages/metering/lib/crons/billingEventsS3Export.ts
```diff
@@ -0,0 +1,256 @@
...
+function exportSql({ metric, day, runTimestamp, eventName }: { metric: MetricSpec; day: string; runTimestamp: string; eventName: string }): string {
+    const dayCompact = day.replace(/-/g, '');
+    const url = `https://${bucket}.s3.amazonaws.com/${dayCompact}/${runTimestamp}_${eventName}.jsonl`;
+
+    return `
+        INSERT INTO FUNCTION s3(
+            '${url}',
+            extra_credentials(role_arn = '${roleArn}'),
+            'JSONEachRow'
+        )
```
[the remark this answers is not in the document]
pfreixes: addressed!

packages/metering/lib/crons/billingEventsS3Export.ts
```diff
@@ -0,0 +1,256 @@
...
+                SUM(duration_ms) AS duration_ms,
+                SUM(custom_logs) AS custom_logs
+            FROM ${database}.daily_function_executions
+            WHERE day = toDate('${day}')
+            GROUP BY account_id, day
+        `,
+        extraProperties: [
+            { propertyName: 'telemetry.durationMs', columnName: 'duration_ms' },
+            { propertyName: 'telemetry.customLogs', columnName: 'custom_logs' }
+        ]
```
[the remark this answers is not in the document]
pfreixes: its not used in our events, but Ill add it just in case

packages/metering/lib/crons/billingEventsS3Export.ts
```diff
@@ -0,0 +1,256 @@
...
+    if (!bucket || !roleArn) {
+        logger.warning(`Skipping (BILLING_EVENTS_S3_BUCKET or BILLING_EVENTS_S3_WRITER_ROLE_ARN not set)`);
+        return;
+    }
+    if (!envs.CLICKHOUSE_URL) {
+        logger.warning(`Skipping (CLICKHOUSE_URL not set)`);
+        return;
+    }
+
+    cron.schedule(`*/${cronMinutes} * * * *`, () => {
```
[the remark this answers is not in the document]
pfreixes: done thanks!

packages/metering/lib/crons/billingEventsS3Export.ts
```diff
@@ -0,0 +1,249 @@
...
+    day,
+    eventName,
+    database = DEFAULT_DATABASE
+}: {
+    metric: MetricSpec;
+    day: string;
+    eventName: string;
+    database?: string;
+}): string {
+    const propertyEntries = [`'count', toString(count)`, ...(metric.extraProperties ?? []).map((p) => `'${p.propertyName}', toString(${p.columnName})`)];
```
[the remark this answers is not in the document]
pfreixes: good point, fixed

packages/metering/lib/crons/billingEventsS3Export.ts
```diff
@@ -0,0 +1,252 @@
...
+        canonicalEventName: 'records',
+        select: (day, database) => `
+            SELECT account_id, day, ROUND(SUM(avg_val)) AS count
+            FROM (
+                SELECT avgMerge(value) AS avg_val, account_id, day, environment_id, integration_id, connection_id, model
+                FROM ${database}.daily_records
+                WHERE day = toDate('${day}')
+                GROUP BY account_id, day, environment_id, integration_id, connection_id, model
+            )
+            GROUP BY account_id, day
```
TBonnin, line 117: do I understand the query correctly: we do the average of the values per model over the time interval then sum all the average values?

packages/metering/lib/crons/billingEventsS3Export.ts
TBonnin: can we partition per hour and export the data for the past hour (or a few hours for redundancy) to reduce the waste?

packages/metering/lib/crons/billingEventsS3Export.ts
```diff
@@ -0,0 +1,252 @@
...
+                SUM(custom_logs) AS custom_logs,
+                SUM(compute_gbms) AS compute_gbms
+            FROM ${database}.daily_function_executions
+            WHERE day = toDate('${day}')
+            GROUP BY account_id, day
+        `,
+        extraProperties: [
+            { propertyName: 'telemetry.durationMs', columnName: 'duration_ms' },
+            { propertyName: 'telemetry.customLogs', columnName: 'custom_logs' },
+            { propertyName: 'telemetry.compute', columnName: 'compute_gbms' }
```
TBonnin: the mapping could be done in the query directly by building a tuple. I even recall that Orb support dot notation so we might not event need to build a telemetry json object

packages/metering/lib/crons/billingEventsS3Export.ts
```diff
@@ -0,0 +1,252 @@
...
+        lock = await locking.acquire(LOCK_KEY, lockTtlMs);
+    } catch {
+        logger.info(`Could not acquire lock, skipping`);
+        return;
+    }
+    logger.info(`Lock acquired`);
+    try {
+        await fn();
+    } finally {
+        await locking.release(lock);
```
TBonnin, line 200: It is probably something we want to refactor but for other crons I think we don't release on success to prevent other instances to acquire the lock and run the job again. Might not be a problem for this export but I wanted to mention it

packages/utils/lib/environment/parse.ts
```diff
@@ -277,6 +278,9 @@ export const ENVS = z.object({
     BILLING_INGEST_BATCH_INTERVAL_MS: z.coerce.number().optional().default(5_000),
     BILLING_INGEST_MAX_QUEUE_SIZE: z.coerce.number().optional().default(100_000),
     BILLING_INGEST_MAX_RETRY: z.coerce.number().optional().default(3),
+    BILLING_EVENTS_S3_BUCKET: z.string().optional().default(''),
+    BILLING_EVENTS_S3_WRITER_ROLE_ARN: z.string().optional().default(''),
```
TBonnin: should we validate it looks like a ARN if provided

packages/utils/lib/environment/parse.ts
```diff
@@ -277,6 +278,9 @@ export const ENVS = z.object({
...
+    BILLING_EVENTS_S3_EVENT_NAME_SUFFIX: z.string().optional().default(''),
```
TBonnin: why the default `''`?

packages/utils/lib/environment/parse.ts
pfreixes: even if we validate it we do not know if it would be valid, so not sure how much value would add doing this. If the ARN is wrongly formatted, invalid, does not have the proper permissions, or whatever the cron will fail and once add the proper telemetry on top of that we will know it.

packages/utils/lib/environment/parse.ts
pfreixes: We can just default to undefined, let me change this.

packages/metering/lib/crons/billingEventsS3Export.ts
pfreixes: Yea, in our case we want to release it since we are interested on having new executions taking the look and finishing what the failing run didn't do it

packages/metering/lib/crons/billingEventsS3Export.ts
pfreixes: The plan is to not have this flow once we productionize this flow, so we will move from this POC which produces the same snapshot every hour within the same day to a system which only produces one, but uses the 1h cron for self-healing, so there should be no wasted files once we do that.

For testing I didn't want to implement all of this, considering that the number of events exported to S3 is just negligible and was enough for let me check if there was data parity btw the HTTP path and the S3 path (We will require more about this later but was just a first start)

We can not partition per hour neither we are using the daily materializations.

packages/metering/lib/crons/billingEventsS3Export.ts
pfreixes: Yes, in the inner query we avg all samples of the same "key" and later we do the sum of independent keys, otherwise we could be summing multiple rows of the same key.

But now thinking it through I can see a systemic issue with our ClickHouse pipeline that breaks down across multiple dimensions impacting averaged billable metrics. For records, when for different reasons — connection churn, connections without reporting records, etc. — we do the AVG using as a denominator the number of samples that we feed into ClickHouse. If we feed
  less — i.e. connection churn — the denominator would be < than the current path, which is 24 samples. So if we have today by contract that we will be averaging by day considering hourly
  samplings, this would basically break the existing contract. As said, this is not an issue with this feature, it's a systemic issue of the current base raw events; indeed the queries
  that we would be running for getting the overalls using ClickHouse would already drift (for the ones where we have fewer samples).

**Simple example.** Account with 1 connection, hourly cron over 24 firings. The connection is created at noon — so it's only alive for 12 firings (12:00–23:00), holding 1000 records the
  whole time.

  - **Legacy path:** ships 1000 at each of 12 firings, ships 0 at each of the other 12 (or in current code, just ships nothing — equivalent to 0). Orb's `average(count)` = `(1000 × 12 + 0
  × 12) / 24 = 500`.
  - **New path:** stores avgState over 12 snapshots of 1000 → avg = 1000. SUM across slices = 1000.

  **Legacy bills for 500. New path bills for 1000. 2× difference.**

We need to discuss this, but I would be leaning towards to try to keep the existing contract if possible

packages/metering/lib/crons/billingEventsS3Export.ts
pfreixes: > the mapping could be done in the query directly by building a tuple

Oks Ill to the refactor for having the map in the SQL query

> I even recall that Orb support dot notation so we might not event need to build a telemetry json object

Im just following the existing convention for not having to make any change in Orb, so we keep maintaining the current event contract. I would prefer to do so

packages/metering/lib/crons/billingEventsS3Export.ts
```diff
@@ -0,0 +1,259 @@
...
+        //        → daily average count, matching what Orb received from the legacy HTTP path.
+        select: (day, database) => `
+            SELECT
+                account_id, day,
+                map('count', toFloat64(ROUND(AVG(batch_val)))) AS properties
+            FROM (
+                SELECT SUM(value) AS batch_val, account_id, day, batch_id
+                FROM ${database}.daily_raw_records
+                WHERE day = toDate('${day}')
+                GROUP BY account_id, day, batch_id
```
TBonnin, line 115: we just need to send the sum for each account (per batchId). That's what is currently send to Orb. Orb then takes care of the average

TBonnin (approved):
one comment about the cumulative metrics where we only need to send the sum per account without doing the average. rest lgtm

PR #6074, opened 2026-05-07
docs: add custom llms indexes

bastienbeurier:
## Summary

Adds custom generated LLM docs indexes for Mintlify:

- `docs/llms.txt` stays focused on core Nango docs and points agents to the provider catalog instead of listing every provider page.
- `docs/llms-full.txt` includes full text for core docs only, excluding provider-specific API pages.
- `docs/api-catalog.txt` exposes compact provider slug metadata so agents can reconstruct provider docs URLs when needed.

Also wires the generator into `npm run docs:generate`, adds a standalone `docs:generate:llms` script, documents the workflow in `docs/AGENTS.md`, and adds a CI freshness check.

## Validation

- `npm run docs:generate:llms`
- `npx eslint scripts/docs-generate-llms.ts --quiet`
- `git diff --check`


.github/workflows/validation.yaml
```diff
@@ -26,12 +26,15 @@ jobs:
                   filters: |
                       docs:
                           - 'docs/**'
+                          - 'scripts/docs-generate-llms.ts'
+                          - 'package.json'
+                          - 'package-lock.json'
```
TBonnin: why package.json and package-lock.json?

docs/api-catalog.txt
TBonnin, line 1: I am not sure: do we want the outputs of the script be generated at website deploy time or have add them to the repo?

scripts/docs-generate-llms.ts
```diff
@@ -0,0 +1,418 @@
...
+
+    const data = yaml.load(match[1]) as Frontmatter | undefined;
+    return data ?? {};
+}
+
+function stripMdxBoilerplate(content: string): string {
+    return content
+        .replace(/^---\n[\s\S]*?\n---\n?/, '')
+        .split('\n')
+        .filter((line) => !line.trim().startsWith('import '))
```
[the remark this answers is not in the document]
TBonnin: why indeed is line starting with import filtered? It is removing all the import in the code examples.

TBonnin (commented):
the generated full file is 14k lines. are we expecting agent to load/read it?

bastienbeurier:
> I am not sure: do we want the outputs of the script be generated at website deploy time or have add them to the repo?

Not by default. `llms.txt` is the default compact entry point. `llms-full.txt` is for agents that explicitly want full core-docs context in one fetch. It intentionally excludes provider page bodies.

docs/api-catalog.txt
bastienbeurier: Mintlify requires a static file in the repo. We could serve this differently, but it's a bigger lift.

.github/workflows/validation.yaml
bastienbeurier: Indeed, that's too conservative.

scripts/docs-generate-llms.ts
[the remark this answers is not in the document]
bastienbeurier: The goal was to strip the imports at the top of the .mdx files that we're using to generate `llms.txt` and `llms-full.txt`, but this was indeed too aggressive and also removed imports from snippets. Should be fixed.

TBonnin (approved):

PR #6077, opened 2026-05-07
fix(utils): add additional retry network error codes from env vars

rossmcewan:


packages/utils/lib/retry.ts
```diff
@@ -98,7 +98,25 @@ export async function retryWithBackoff<T extends () => any>(fn: T, options?: Bac
...
+}
+
+export const networkError = [
+    'ECONNRESET',
+    'ETIMEDOUT',
+    'ECONNABORTED',
+    'ECONNREFUSED',
+    'EHOSTUNREACH',
+    'EAI_AGAIN',
+    ...parseCommaSeparatedEnvList(process.env['NANGO_RETRYABLE_NETWORK_ERRORS'])
```
TBonnin, line 118: Let's be consistent and move `NANGO_RETRYABLE_NETWORK_ERRORS` with the rest of the env vars and do the validation with zod

packages/utils/lib/retry.unit.test.ts
```diff
@@ -83,3 +95,47 @@ describe('retry', () => {
...
+    const envVarName = 'NANGO_RETRYABLE_NETWORK_ERRORS';
+    const originalEnvValue = process.env[envVarName];
+
+    afterEach(() => {
+        vi.resetModules();
+        if (originalEnvValue === undefined) {
+            Reflect.deleteProperty(process.env, envVarName);
+        } else {
+            process.env[envVarName] = originalEnvValue;
+        }
```
TBonnin: can we use`vi.stubEnv` in the tests instead?

TBonnin (approved):

